PR #127, opened 2017-08-23
Docs: Add footer

sawyerh:
### Internal

- Added footer to docs site with required links

@shashashasha @allypalanzi @cindyphan Does anyone have a strong preference for either of these? The light blue is what USWDS uses for footers, so that's what I currently have implemented:

localhost-3000-design-system-

localhost-3000-design-system- 1

### DIFF
--- a/packages/docs/src/scripts/Docs.jsx
+++ b/packages/docs/src/scripts/Docs.jsx
@@ -1,6 +1,7 @@
 /**
  * This is main template file for the documentation site.
  */
+import Footer from './components/Footer';
 import GitHubLinks from './components/GitHubLinks';
 import Header from './components/Header';
 import Nav from './components/Nav';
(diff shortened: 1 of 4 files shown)


sawyerh:
Here's an alternate grid column order

localhost-3000-design-system- 2

shashashasha:
I think I have a slight preference for having the footer CTA on the left, even though it takes up more columns than the left navigation. The other footer links are just more tangential (only the USWDS seems directly related).

I don't really have a preference on the light vs dark footer. Should we include the 'official federal government website' header?

allypalanzi:
+1 agree with Sha, the CTA on the left feels better to me because otherwise my eye gets drawn to the right which feels odd. I personally prefer the light footer but no strong opinion!

pwolfert:
Something that's not really visible in these screenshots is the fact that it's left-aligned. On larger screens that creates a lot of unbalanced dead space on the right. That's mirrored by the content, so maybe it's not bad, but USWDS centers its footer content, so I just thought I'd make note of it for the designers.

large-screen

sawyerh:
Thanks y'all. Updated the footer to span the entire screen width on extra large screens. Going to merge, and we can open up separate tickets for any other changes.

PR #128, opened 2017-08-23
Update alert styles to match new USWDS styles

hannaliebl:
### Changed
Updated the styles for the Alert component to match the new [USWDS styles](https://github.com/18F/web-design-standards/pull/2095).

Old:
screen shot 2017-08-23 at 4 59 15 pm

New:
screen shot 2017-08-23 at 4 57 55 pm

### DIFF
--- a/packages/core/src/components/Alert/Alert.scss
+++ b/packages/core/src/components/Alert/Alert.scss
@@ -22,25 +22,23 @@ Markup:
 Style guide: components.alert
 */
 
-$h-padding: $spacer-2;
-$bar-size: $spacer-1;
-$alert-icon-size: $spacer-7;
+$alert-padding: $spacer-2;
+$alert-bar-size: $spacer-1;
+$alert-icon-size: $spacer-5;
 
 .ds-c-alert {
   background-color: $color-primary-alt-lightest;
   background-image: url('#{$image-path}/info.png');
   background-image: url('#{$image-path}/info.svg');
-  background-position: $h-padding $h-padding;
+  background-position: $alert-padding $alert-padding;
   background-repeat: no-repeat;
   background-size: $alert-icon-size;
   box-sizing: border-box;
   color: $color-base;
   min-height: $alert-icon-size + $spacer-2;
-  padding-bottom: $h-padding * 0.7;
-  padding-left: $h-padding;
-  padding-right: $h-padding;
-  padding-top: $h-padding;
+  padding: $alert-padding;
   position: relative;
+  border-left: $alert-bar-size solid $color-primary-alt;
 
   a {
     color: $color-primary-darker;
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Alert/Alert.scss
```diff
@@ -55,10 +60,20 @@ $alert-icon-size: $spacer-5;
       margin-top: 0;
     }
   }
+
+  &:before {
```
sawyerh: 🤔 Could this just be a `border-left` property on `ds-c-alert`?

packages/core/src/components/Alert/Alert.scss
```diff
@@ -22,20 +22,25 @@ Markup:
 Style guide: components.alert
 */
 
-$alert-icon-size: $spacer-5;
+$h-padding: $spacer-2;
+$bar-size: $spacer-1;
+$alert-icon-size: $spacer-7;
```
sawyerh: This icon size feels a little too big, in relation to the size of the heading and body text. I have a slight preference for bumping this down. Thoughts on reverting it back to `$spacer-5`? That would align the icon with the first line of text:

image

image

packages/core/src/components/Alert/Alert.scss
```diff
@@ -22,20 +22,25 @@ Markup:
 Style guide: components.alert
 */
 
-$alert-icon-size: $spacer-5;
+$h-padding: $spacer-2;
+$bar-size: $spacer-1;
```
sawyerh: Nit: Could we could name these consistently (ie. namespacing w/ `$alert-`), so it's easier for people to theme and less likely to conflict with a project's existing Sass variables?

packages/core/src/components/Alert/Alert.scss
```diff
@@ -22,20 +22,25 @@ Markup:
...
-  background-position: $spacer-1 ($spacer-2 * 0.95);
+  background-position: $h-padding $h-padding;
   background-repeat: no-repeat;
   background-size: $alert-icon-size;
-  border: 1px solid $color-primary-alt-light;
   box-sizing: border-box;
   color: $color-base;
   min-height: $alert-icon-size + $spacer-2;
-  padding: $spacer-2;
+  padding-bottom: $h-padding * 0.7;
```
sawyerh: Could we make the top and bottom padding consistent (`padding: $h-padding`)? Visually, this feels a little off.

packages/core/src/components/Alert/Alert.scss
```diff
@@ -71,27 +86,37 @@ $alert-icon-size: $spacer-5;
...
+  }
 }
 
 .ds-c-alert--error {
   background-color: $color-error-lightest;
   background-image: url('#{$image-path}/error.png');
   background-image: url('#{$image-path}/error.svg');
-  border-color: $color-error-light;
+  &:before {
+    background-color: $color-secondary;
```
sawyerh: For the "error", "warn", and "success" modifiers, we should use their corresponding Sass variables: `$color-error`, `$color-warn`, `$color-success`. This would correspond to people's expectations if they override those variables.

packages/core/src/components/Alert/Alert.scss
hannaliebl: Agreed, I like that better.

hannaliebl:
PTAL - I also got rid of the only-child styles since I think it looks better with the smaller icon like that.

sawyerh:
Looks great! 👌 

Last remaining bit is some linting errors. You can test this locally by running `yarn test`:

image

PR #138, opened 2017-08-31
feature/spinner-a11y

1Copenut:
PR intends to add useful ARIA landmarks to our spinner graphic, so screenreaders pick up that the page is loading.

### Added
* `aria-valuetext="Loading"` and `role="progressbar"`

These additions allow screen readers to see the loading icon and read out Loading. This was tested in VoiceOver for Mac, and NVDA on Chrome and FF in a Win7 VM.

### Changed
* Spinner.jsx has had two props added in a destructured manner, and default props added for the aria-valuetext and role props. These were added as default props because they should be present on all spinners, whether used alone or embedded in buttons. Embedded spinners will not read out the spinner text, only what's in the button.
* Spinner.scss has had some a11y documentation added for considering how we can announce the spinner's arrival/departure, and changing content in the parent container.


packages/core/src/components/Spinner/Spinner.scss
```diff
@@ -113,11 +113,22 @@ Style guide: components.spinner
 
 - If the process takes a long time, use something else.
 
+## Accessibility Considerations
+
+- Spinners are useful to announce a page load sequence, and should be paired with a parent container that includes several ARIA attributes:
```
sawyerh: Should we add these ARIA attributes to the examples that show spinners within a parent container? (Button, "Filled", and the React component examples)

packages/core/src/components/Spinner/Spinner.scss
```diff
@@ -113,11 +113,22 @@ Style guide: components.spinner
...
+  * `aria-relevant="additions text"`
+  * `aria-live="polite"`
+  * `aria-atomic="false"`
+- aria-relevant may include additions, text, and removals. If a section is going to have items added or removed, use additions and removals. If there will be additions or text changes without explicit removals, use additions and text.
```
sawyerh: Nit: Can we format these HTML attributes and values as code? `aria-relevant`, `additions`, `text`, `removals`, etc

packages/core/src/components/Spinner/Spinner.scss
```diff
@@ -113,11 +113,22 @@ Style guide: components.spinner
 
 - If the process takes a long time, use something else.
 
+## Accessibility Considerations
```
sawyerh: For consistency with [other pages](https://design.cms.gov/components/tabs/#guidance), let's change this to "Accessibility"

packages/core/src/components/Spinner/Spinner.scss
```diff
@@ -23,9 +23,9 @@ Spinners signify that the application is waiting for an asynchronous operation t
 @status alpha
 
 Markup:
-<span class="ds-c-spinner ds-c-spinner--small"></span>
-<span class="ds-c-spinner"></span>
-<span class="ds-c-spinner ds-c-spinner--big"></span>
+<span class="ds-c-spinner ds-c-spinner--small" aria-valuetext="Loading" role="progressbar"></span>
```
sawyerh: I see these are documented in the React component, but we should probably expand on these in the "Accessibility" section of the Guidance section. For example, according to [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuetext_attribute) `aria-valuetext` should only be set when the rendered value can't be accurately represented as a number.

packages/core/src/components/Spinner/Spinner.scss
1Copenut: I think we should pass on adding these attributes to the buttons. It's more useful in context of a containing element like a div or table, that when updated, announces to users. I don't believe a button, even with the extra aria markup, will announce a change in text.

If you'd like to add the extra markup to the buttons, I'll do some further testing on that scenario.

packages/core/src/components/Spinner/Spinner.scss
1Copenut: Sure thing. Should these not be list items, and just inline code, or maybe marked up as a solid block of HTML that displays all of the contextual information?

1Copenut:
@sawyerh Appreciate the updated documentation. Branch has been merged into staging.

PR #139, opened 2017-09-02
Add support for site packages (AKA "themes")

sawyerh:
### Internal

This is one idea for developing, documenting, and previewing the HealthCare.gov "Site Package" NPM module. It allows a site package module to use the same tooling by adding support for a `themes` directory, which the site package can be nested within. The `themes` directory is gitignored, so it won't be included in this repo.

At a minimum, it expects a file structure like:

```
└── packages                              // existing packages directory
    ├── core                              // existing core package directory
    ├── themes
    │   └── ds-healthcare-gov             // site package can be git cloned here
    │       └── src
    │           └── index.scss            // design system variable overrides and new styles
    └── Other design system packages...
```

The biggest benefits of this approach are, with minimal effort we can:

1. Easily preview the effects of a site package's design system overrides
2. Generate the same design system documentation, with the addition of any site package documentation

#### TODO

- [x] Add CLI option for excluding themes directory
- [x] Support site package styling on React components. (This would be easier if the React components were also rendered in iframes like the CSS components)
- [x] Support overriding documentation pages like "Getting started"

#### Open questions

- Will the site package's documentation live on GitHub Pages? This will determine whether we want to add support for specifying where the docs are generated into.
- How do we handle CI testing and linting in the site package's repo, since the tooling won't be in the site package repo? There may need to be just a little bit of duplication here, but I don't think it's a deal breaker.

#### Known limitations

- ~This doesn't currently support multiple site packages, but this is solveable if the use case arises.~
  - Resolved by allowing you to specify the package name from the CLI. `yarn start:theme -- ds-healthcare-gov`
- The current implementation assumes a site package doesn't override design system declarations, and only overrides its variables or introduces new style declarations.


sawyerh:
@pwolfert When you have some time, could you try running through [these instructions](https://github.com/CMSgov/design-system/wiki/site-packages-and-themes) to preview the [HealthCare.gov Site Package](https://github.cms.gov/CMS-WDS/ds-healthcare-gov)? The main thing I'm looking for is:

1. Are there parts of the instructions that need clarified or expanded?
1. Are there any things or features I've missed for this MVP?

_Note_: Make sure you have this branch checked out before running through the instructions 😄

sawyerh:
@hannaliebl When you have a moment, could you also help test this out? See comment above.

hannaliebl:
@sawyerh it worked for me! I think the directions are clear. The only potentially confusing part is the symlink example. Maybe clarify that the example shown is specifically how to symlink. I'd never used `ln -s` before, so I assumed it was a way to create a symlink, but it might be better to be very clear there. Something like:

4. Then place or symlink your theme's directory into `packages/themes`. Here's an example of how to symlink your directory if you don't want to copy the files over:

packages/generator-cmsgov/generators/app/index.js
```diff
@@ -36,14 +36,25 @@ module.exports = class extends Generator {
         message: 'Generate React files too?',
         default: true
       }]);
-    }).then(answers => {
+    })
+    .then(answers => {
       this.props.slug = answers.slug;
       this.props.react = answers.react;
+
+      return this.prompt([{
```
pwolfert: I'm not sure how to get to this prompt. It doesn't show it for me.

tools/gulp/docs/__tests__/createRoutes.test.js
```diff
@@ -1,15 +1,27 @@
...
+function mockPage(slug, sections = [], customProps = {}) {
+  return Object.assign({
     header: slug,
     description: `${slug} description`,
+    reference: slug,
     referenceURI: slug,
     sections: sections
-  };
+  }, customProps);
 }
```
pwolfert: Did you plan on making use of this change? I think it's useful enough to keep, but it doesn't look like it was used.

packages/generator-cmsgov/generators/app/index.js
sawyerh: I think you might need to rerun `yarn install` for this new prompt to work

pwolfert: screen shot 2017-09-19 at 3 57 21 pm
Heh, I guess we've already started running into problems with node v8.x

sawyerh: Following up on this, you have to run `yarn upgrade generator-cmsgov`...which isn't ideal. Not sure if there's a better way around that.

PR #144, opened 2017-09-15
RTL Choice components

pwolfert:
There is now an option to put the label for a checkbox or radio button on the left side instead of the right.

### Added
- `ds-c-choice--left-label` puts the label on the left side of the checkbox/radio button

rtl-choice

### DIFF
--- a/packages/core/src/components/ChoiceList/Choice.jsx
+++ b/packages/core/src/components/ChoiceList/Choice.jsx
@@ -17,16 +17,18 @@ export const Choice = function(props) {
     className,
     id,
     inversed,
-    right,
+    inputPlacement,
+    inputClassName,
     ...inputProps
   } = props;
   /* eslint-enable prefer-const */
 
   const inputClasses = classNames(
+    inputClassName,
     'ds-c-choice',
     {
       'ds-c-choice--inverse': inversed,
-      'ds-c-choice--right': right
+      'ds-c-choice--right': inputPlacement === 'right'
     }
   );
 
@@ -65,6 +67,10 @@ Choice.propTypes = {
    */
   className: PropTypes.string,
   /**
+   * Additional classes to be added to the `input` element.
+   */
+  inputClassName: PropTypes.string,
+  /**
    * Sets the initial checked state. Use this for an uncontrolled component;
    * otherwise, use the `checked` property.
    */
@@ -79,9 +85,9 @@ Choice.propTypes = {
    */
   inversed: PropTypes.bool,
   /**
-   * Displays the checkbox on the right side of the label
+   * Placement of the input relative to the text label (`left` or `right`)
    */
-  right: PropTypes.bool,
+  inputPlacement: PropTypes.oneOf(['left', 'right'])
   /**
    * The `input` field's `name` attribute
    */
--- a/packages/core/src/components/ChoiceList/Choice.jsx
+++ b/packages/core/src/components/ChoiceList/Choice.jsx
@@ -49,7 +49,8 @@ export const Choice = function(props) {
 };
 
 Choice.defaultProps = {
-  type: 'checkbox'
+  type: 'checkbox',
+  inputPlacement: 'left'
 };
 
 Choice.propTypes = {
@@ -85,7 +86,7 @@ Choice.propTypes = {
    */
   inversed: PropTypes.bool,
   /**
-   * Placement of the input relative to the text label (`left` or `right`)
+   * Placement of the input relative to the text label
    */
   inputPlacement: PropTypes.oneOf(['left', 'right']),
   /**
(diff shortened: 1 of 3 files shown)


packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -161,6 +187,17 @@ Style guide: components.choice
   border-radius: 100%;
 }
 
+// Display label on the left side of the checkbox
+.ds-c-choice--left-label + label {
```
sawyerh: Thoughts on changing this modifier to `ds-c-choice--right`? Since the class is applied to the input, rather than the label, I think this still makes sense and is inline with other modifiers like `ds-u-float--right` and `ds-u-text-align--right`.

pwolfert: Yeah I think that's reasonable. I wasn't entirely sure what to call it, so I figured we'd discuss it in a review.

packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -161,6 +187,17 @@ Style guide: components.choice
...
+  padding-left: 0;
+  padding-right: $spacer-5;
```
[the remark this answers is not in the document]
pwolfert: This sounds like more of an intermediary value that we would calculate anyway. I feel like that could just go in the `Choice.scss` file to keep us from calculating twice, but that we should rather defined the gap between them (which happens to be `$spacer-1`) as a variable. The size+padding seems to be a roundabout value that nobody actually cares about, especially if we were to make multiple sizes of checkboxes later. It's the gap that people would care to modify.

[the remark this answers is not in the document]
pwolfert: Something like `$choice-label-checkbox-gap-x: $spacer-1;`

packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -79,9 +79,9 @@ Choice.propTypes = {
    */
   inversed: PropTypes.bool,
   /**
-   * Displays the label on the left side of the check
+   * Displays the checkbox on the right side of the label
    */
-  labelOnLeft: PropTypes.bool,
+  right: PropTypes.bool,
```
sawyerh: I'm torn on this name. `labelOnLeft` makes it more obvious to me that it expects a boolean value and makes clear what behavior is expected when the prop is present. (I understand why this change was made though).

A few other ideas:
- I'm wondering if we should instead support the option to pass in additional class names for the `input` (e.g. `inputClassName`), which isn't currently supported but might be useful, especially if/when we add another modifier for making the checkbox small. (Note: The Button component includes a `size` prop)
- Another idea: Having a prop like `inputPlacement` which accepts a value of either `left` (default) or `right`.

I think I'm leaning more for `inputClassName` since that seems like nice functionality to have either way, but not a strong preference.

pwolfert: I think we should have the `inputClassName` anyway, but I would like to also have a prop specifically for the label placement, because half the point of having these react classes is that we abstract away some of these long class names and give people a nicer interface for defining these things. I think `inputPlacement` as a string is easier than coming up with a good name for a boolean.

packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -79,9 +85,9 @@ Choice.propTypes = {
    */
   inversed: PropTypes.bool,
   /**
-   * Displays the checkbox on the right side of the label
+   * Placement of the input relative to the text label (`left` or `right`)
    */
-  right: PropTypes.bool,
+  inputPlacement: PropTypes.oneOf(['left', 'right'])
```
sawyerh: Tests are failing from a missing a trailing comma here

pwolfert:
Why is the full react summary with a parameter table not shown for the `Choice` component?

packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -79,9 +85,9 @@ Choice.propTypes = {
    */
   inversed: PropTypes.bool,
   /**
-   * Displays the checkbox on the right side of the label
+   * Placement of the input relative to the text label (`left` or `right`)
```
sawyerh: image

Nit: The documentation will automatically indicate the possible values based on what you set in `PropTypes.oneOf`, so we could get rid of that info from the description. 

Related: Can we set a `defaultProps` value for `inputPlacement`? Right now it's unclear to someone looking at the documentation or source code what the placement will be if it's left blank.

sawyerh:
> Why is the full react summary with a parameter table not shown for the Choice component?

@pwolfert Not sure what ya mean

pwolfert:
Ah, it's there now. It may have been an issue related to the syntax error and caching.

sawyerh:
@pwolfert Sorry, one more request, can you add a test to `Choice.test.jsx` to ensure these new props work?

PR #148, opened 2017-09-22
KSS looks in JavaScript files for documentation comments

pwolfert:
### Added
- KSS now parses JSX files as well when looking for documentation comments

### DIFF
--- a/tools/gulp/docs/index.js
+++ b/tools/gulp/docs/index.js
@@ -185,7 +185,7 @@ module.exports = (gulp, shared) => {
     );
 
     const packages = shared.packages.map(pkg => `packages/${pkg}/src/`);
-    const mask = /^(?!.*\.example\.jsx).*\.(css|less|sass|scss|jsx)$/;
+    const mask = /^(?!.*\.(example|test)).*\.(css|less|sass|scss|jsx)$/;
     return kss.traverse(packages, { mask })
       .then(styleguide =>
         /**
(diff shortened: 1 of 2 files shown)


tools/gulp/docs/index.js
```diff
@@ -185,7 +185,8 @@ module.exports = (gulp, shared) => {
     );
 
     const packages = shared.packages.map(pkg => `packages/${pkg}/src/`);
-    return kss.traverse(packages)
+    const mask = '*.css|*.less|*.sass|*.scss|*.jsx';
```
sawyerh: We also have `.jsx` files for examples and tests (ie. `Choice.example.jsx`, `Choice.test.jsx`), which this will traverse over. Not sure if that will have any noticeable performance impact, but if we could ignore those, that'd be ideal.

tools/gulp/watch.js
```diff
@@ -30,7 +30,7 @@ module.exports = (gulp, shared) => {
       `packages/${packages}/src/**/*.{js,jsx}`,
       `!packages/${packages}/src/**/*.example.{js,jsx}`,
       `!packages/${packages}/src/**/*.test.{js,jsx}`
-    ], ['lint:packages-scripts', 'docs:react']);
+    ], ['lint:packages-scripts', 'sass:process:docs', 'docs:generate-pages', 'docs:react']);
```
sawyerh: This is probably a little confusing based on how things are currently named, but `docs:generate-pages` will need to come after `docs:react`, since `docs:react` creates a file (`react-doc.json`) which the `docs:generate-pages` task depends upon.

Also, I don't think `sass:process:docs` needs called here or below. It only transforms the Sass to CSS, which wouldn't need done when a JSX file changes.

pwolfert: Ah, I assumed `sass:process:docs` read the docs out of the sass

PR #152, opened 2017-09-26
Adding requirementLabel prop to the FormLabel

pwolfert:
Added a `requirementLabel` prop to the `FormLabel` component to fulfill the requirements of https://jira.cms.gov/browse/HDSG-98. I also opened up the documentation page for the `FormLabel` with guidance copied from the forms guidance as a starting point for discussing what to include in this page.

### Added
- Added a `requirementLabel` prop to the `FormLabel` component
- Opened up a documentation page for `FormLabel`

### Changed
- `Choice` component now uses a `FormLabel` instead of just writing the label markup inline.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,6 @@
     "babel-core": "^6.26.0",
     "babel-jest": "^21.0.2",
     "babel-loader": "^7.1.2",
-    "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-react": "^6.24.1",
--- a/package.json
+++ b/package.json
@@ -23,6 +23,7 @@
     "babel-core": "^6.26.0",
     "babel-jest": "^21.0.2",
     "babel-loader": "^7.1.2",
+    "babel-plugin-transform-class-properties": "^6.24.1",
     "babel-plugin-transform-object-assign": "^6.22.0",
     "babel-preset-es2015": "^6.24.1",
     "babel-preset-react": "^6.24.1",
--- a/package.json
+++ b/package.json
@@ -95,11 +95,7 @@
       "<rootDir>/packages/generator-cmsgov/generators/app/templates/"
     ]
   },
-<<<<<<< HEAD
   "resolutions": {
     "gulp-sass/node-sass": ">=4.5.3"
   }
-=======
-  "dependencies": {}
->>>>>>> 573e4a8... Consolidated `optional` and `required` props into a `requirementLabel` prop that can be a string or node
 }
--- a/package.json
+++ b/package.json
@@ -95,7 +95,11 @@
       "<rootDir>/packages/generator-cmsgov/generators/app/templates/"
     ]
   },
+<<<<<<< HEAD
   "resolutions": {
     "gulp-sass/node-sass": ">=4.5.3"
   }
+=======
+  "dependencies": {}
+>>>>>>> 573e4a8... Consolidated `optional` and `required` props into a `requirementLabel` prop that can be a string or node
 }
--- a/packages/core/src/components/ChoiceList/ChoiceList.jsx
+++ b/packages/core/src/components/ChoiceList/ChoiceList.jsx
@@ -141,6 +141,8 @@ export class ChoiceList extends React.PureComponent {
           errorMessage={this.props.errorMessage}
           fieldId={this.id()}
           hint={this.props.hint}
+          optional={this.props.optional}
+          required={this.props.required}
           inversed={this.props.inversed}
         >
           {this.props.label}
@@ -178,6 +180,16 @@ ChoiceList.propTypes = {
    * Additional hint text to display
    */
   hint: PropTypes.node,
+  /**
+   * Set to `true` to display whether the field is optional or pass text to
+   * display as optional with additional text.
+   */
+  optional: PropTypes.oneOfType([PropTypes.bool, PropTypes.node]),
+  /**
+   * Set to `true` to display whether the field is required or pass text to
+   * display as required with additional text.
+   */
+  required: PropTypes.oneOfType([PropTypes.bool, PropTypes.node]),
   /**
    * Applies the "inverse" UI theme
    */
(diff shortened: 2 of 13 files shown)


packages/core/src/components/FormLabel/FormLabel.scss
```diff
@@ -4,7 +4,17 @@
 Form label/legend
```
sawyerh: Nit: Could we change this to "Label & Legend"?

packages/core/src/components/FormLabel/FormLabel.example.jsx
```diff
@@ -0,0 +1,14 @@
+/* eslint-disable react/display-name */
+import FormLabel from './FormLabel';
+import React from 'react';
+
+export default function() {
+  return (
+    <FormLabel
```
sawyerh: Can we add `requirementLabel` to this component to provide an example of what this looks like?

packages/core/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -179,6 +181,10 @@ ChoiceList.propTypes = {
    */
   hint: PropTypes.node,
   /**
+   * Text showing the requirement ("Required", "Optional", etc.). See []().
```
sawyerh: TODO: Add the link (or remove the "See..." part)

packages/core/src/components/FormLabel/FormLabel.jsx
```diff
@@ -43,6 +54,7 @@ export class FormLabel extends React.PureComponent {
         <span className={labelTextClasses}>{this.props.children}</span>
         {this.errorMessage()}
         {this.hint()}
+        {this.requirementLabel()}
```
sawyerh: I'm interpreting the design differently. I'm under the impression that the `requirementLabel` is the first child of the "hint" element, and is followed by any other hint text.

Pseudo code:

```jsx
<span>{this.props.requirementLabel} {this.props.hint}</span>
```

packages/core/src/components/FormLabel/FormLabel.scss
```diff
@@ -41,3 +51,38 @@ Hint text
...
+- Apply the `ds-c-field__hint` class to hint text.
+- Use hint text for supporting contextual help, which will always be shown.
+- Hint text should sit above a form field and below the label text.
+
+## Validation
+
+- Place inline validation messages within the field's `<label>` element.
+- Visually align inline validation messages with the input fields, so people using screen magnifiers can read them quickly.
+
+Style guide: components.form-label.guidance
```
sawyerh: Nice, I like having this info on this page. 

- [ ] We can now remove it from the generic "Forms" guidance page.

Some suggestions to bring this inline with other component guidance tabs:
> ## When to use
> **Labels**
> - Each form field should have a `<label>`. Never use a field's `placeholder` attribute as the primary way to label the field.

> **Legends**
> - Use a single `legend` for `fieldset` (this is required). One example of a common use of `fieldset` and `legend` is a question with radio button options for answers. The question text and radio buttons are wrapped in a `fieldset`, with the question itself being inside the `legend` tag.

> ## Guidance
> **Labels**
> - Apply the `ds-c-label` class to `<label>` elements.
> - Labels should have a `for` attribute, referencing the corresponding input's unique `id` attribute. Only one label can be associated to each unique form element.
> - Labels should be placed above their fields.
> - Label text should be short and in sentence case.
> - Avoid colons at the end of labels.

> **Hint text**
> - Place hint text within the field's `<label>` element.
> - Apply the `ds-c-field__hint` class to hint text.
> - Use hint text for supporting contextual help, which will always be shown.
> - Hint text should sit above a form field and below the label text.

> **Validation**
> - Place inline validation messages within the field's `<label>` element.
> - Visually align inline validation messages with the input fields, so people using screen magnifiers can read them quickly.
> ## Learn more
> - [Form Guidelines]({{root}}/guidelines/forms/)

packages/core/src/components/FormLabel/FormLabel.example.jsx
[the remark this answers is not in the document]
pwolfert: I replaced the `PropTypes.string` with
```js
function(props, propName, componentName) {
  if (props.errorMessage && props[propName] == null) {
    return new Error(`Prop \`${propName}\` is required when specifying an \`errorMessage\`. None specified for ${componentName}.`);
  }
  else {
    return PropTypes.node.apply(PropTypes.node, arguments);
  }
}
```
but it says `custom` in the doc page, which is unhelpful, isn't it?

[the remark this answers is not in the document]
sawyerh: Hmm, we can punt on this..I don't want it to block merging this PR's actual feature. 

I believe `fieldId` would also be required if `props.component` is `label` (the default), since it would be used as the label's `for` attribute value. I agree having it say `custom` on the doc page isn't ideal, but I think that's solveable.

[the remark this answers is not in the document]
pwolfert: What if we made it required, but if people really want to omit it, they could just say `fieldId=""`?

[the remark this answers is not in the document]
sawyerh: @pwolfert I think I'm cool with that (as well as supporting `null`). The one case where it wouldn't be required is when the component is a `legend` (like in `ChoiceList`).

packages/core/src/components/ChoiceList/ChoiceList.jsx
pwolfert: I replaced it with a link that should work, but it is replacing `{{root}}` with the wrong value for some reason. I've looked at the responsible functions, but I don't see any apparent issues.

package.json
```diff
@@ -23,6 +23,7 @@
     "babel-core": "^6.26.0",
     "babel-jest": "^21.0.2",
     "babel-loader": "^7.1.2",
+    "babel-plugin-transform-class-properties": "^6.24.1",
```
sawyerh: Just curious, where is this needed?

pwolfert: I was going to use it and ended up not. Nice to have, but I can remove it.

PR #161, opened 2017-10-06
Feature/a11y tabs

1Copenut:
### Added
Added the ability for Tabs component to change tab focus by pressing left and right arrows. This is in line with the latest [WAI-ARIA Authoring Practices 1.1 for Tabs.](https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel)

### Changed
* Updated Tabs.jsx to include a keydown method that listens for left and right arrow presses, without interfering with Tab or Enter keypresses
* Updated Tabs.jsx to include a ref prop for handling focus
* Updated Tab.jsx to include a ref prop for handling focus
* Updated Tabs.test.jsx to include a deepMount function that passes custom props and renders a full component using Enzyme's mount method
* Updated the Webpack.config.js file to add CoreJS method for Array.prototype.findIndex
* Updated the Webpack.config.js file to add CoreJS method for Object.getOwnPropertyNames

### Tested for a11y
* Verified proper keyboard handling in following browsers:
** Windows 10 - IE11, Edge, Chrome, Firefox
** OSX - Safari, Chrome, Firefox
* Verified tab usability in following screenreader combinations:
** Windows 10 - IE11 - JAWS
** Windows 10 - Firefox - NVDA
** OSX - Safari - VoiceOver


packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -74,7 +94,24 @@ export class Tabs extends React.PureComponent {
     this.replaceState(href);
   }
 
-  // Filter children and return only TabPanel components
+  handleTabKeyPress(evt, panelId) {
+    switch (evt.key) {
+      case LEFT_ARROW:
+        this.tabSwitchLeft(evt, panelId);
```
sawyerh: I believe we'd also want to call `evt.preventDefault()` here (and in the next `case`), otherwise the left and right arrow will still scroll the page horizontally if the page overflows the x-axis. Resize the page to `775px` to see what I mean.

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -74,7 +94,24 @@ export class Tabs extends React.PureComponent {
     this.replaceState(href);
   }
 
-  // Filter children and return only TabPanel components
+  handleTabKeyPress(evt, panelId) {
```
sawyerh: Nit: This handler gets called `onKeyDown`, not `onKeyPress`, so this method name isn't totally accurate.

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -4,6 +4,14 @@ import Tab from './Tab';
 import TabPanel from './TabPanel';
 import classnames from 'classnames';
 
+/** CONSTANTS
+ * Adding in the constant values for keycodes
+ * to handle onKeyDown events
+ */
+const LEFT_ARROW = 'ArrowLeft';
+const RIGHT_ARROW = 'ArrowRight';
+const TAB = 'Tab';
```
sawyerh: If `case TAB` is removed (see other comment), this line can then also be removed.

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -134,6 +175,75 @@ export class Tabs extends React.PureComponent {
...
+   * string instead. 
+   */
+  tabSwitchLeft(evt, curId) {
+    const tabs = this.tabsArrayLen();
+    const index = this.tabFindIndex(tabs, curId);
+
+    if (index === 0) {
+      const target = tabs[tabs.length - 1].props.id;
+      this.setState({ selectedId: target });
+      this.replaceState(`#${target}`);
```
sawyerh: We should use the Tab's `href` prop, which is returned as the forth argument of `onKeyDown`. It's not guaranteed to be a hash, so doing it this way won't always work. This also applies to the other lines where this is being done.

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -134,6 +175,75 @@ export class Tabs extends React.PureComponent {
...
+   * @param {String} ID prop of currently selected Tab
+   * 
+   * The index returned from tabFindIndex is used in a reduce by 1 logic
+   * tree to ensure the right tab[index] ID is targeted for updated state.
+   * 
+   * The replaceState function did not like the href argument passed into 
+   * handleTabClick, so passing an interpolated hashbang plus selected Tab
+   * string instead. 
+   */
+  tabSwitchLeft(evt, curId) {
```
sawyerh: We should try using consistent argument names here. It keeps things easier to follow, but also helps catch things like: The id being passed into this method right now is `panelId` — based on the description of this argument though, you might be expecting the tab id instead?

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -113,7 +152,9 @@ export class Tabs extends React.PureComponent {
           id={panelTabId(panel)}
           key={panel.key}
           onClick={this.handleTabClick}
+          onKeyDown={this.handleTabKeyPress}
           panelId={panel.props.id}
+          ref={(tab) => { this.tabs[panel.props.id] = tab; }}
```
sawyerh: I believe you want the tab `id` here instead? That would be `panelTabId(panel)`. You could create a `const` for that so you only need to call the method once.

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -134,6 +175,75 @@ export class Tabs extends React.PureComponent {
...
+  }
+
+  /**
+   * Find the current tab index on left or right arrow keypress
+   * @param {Array} React Tab components
+   * @param {String} ID prop of currently selected Tab
+   * @return {Number} Returns index of currently selected array item
+   */
+  tabFindIndex(tabsArr, findId) {
+    const index = tabsArr.findIndex(elem => elem.props.id === findId);
```
sawyerh: Could you use `this.tabs` here rather than passing it in as an argument?

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -134,6 +175,75 @@ export class Tabs extends React.PureComponent {
...
+    const index = this.tabFindIndex(tabs, curId);
+
+    if (index === 0) {
+      const target = tabs[tabs.length - 1].props.id;
+      this.setState({ selectedId: target });
+      this.replaceState(`#${target}`);
+    } else {
+      const target = tabs[index - 1].props.id;
+      this.setState({ selectedId: target });
+      this.replaceState(`#${target}`);
```
sawyerh: Rather than calling `setState` and `replaceState` multiple times in this method, you could move these to the bottom, outside of the conditional. The only unique line in this conditional is the assignment of `target` (and specifically the index calculation).

packages/docs/src/scripts/helpers/polyfills.js
```diff
@@ -5,6 +5,7 @@
 // - github.com/facebook/react/issues/8379#issuecomment-264867090
 require('core-js/fn/array/includes');
 require('core-js/fn/array/fill');
+require('core-js/fn/array/find-index');
```
sawyerh: Heads up: These polyfills are only loaded on the documentation site, so this will likely need moved into the relevant component file.

packages/core/src/components/Tabs/Tabs.jsx
1Copenut: @sawyerh I wondered about using the href attribute, but it was returning the wrong value. It's possible because I'm passing in the tab, not the tabPanel for ID that I'm getting the wrong value. I'll continue to review and clean up accordingly.

packages/core/src/components/Tabs/Tabs.jsx
1Copenut: @sawyerh I wondered about using the href attribute, but it was returning the wrong value. It's possible because I'm passing in the tab, not the tabPanel for ID that I'm getting the wrong value. I'll continue to review and clean up accordingly.

1Copenut:
@sawyerh Code is ready for second review.

packages/core/src/components/Tabs/Tab.jsx
```diff
@@ -65,6 +84,12 @@ Tab.propTypes = {
    */
   onClick: PropTypes.func,
   /**
+   * Click event called with the following arguments:
```
sawyerh: This isn't a "Click event" handler. Suggested edit:

> Called when the tab is selected and a `keydown` event is triggered. Called with the following arguments:

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -91,7 +119,7 @@ export class Tabs extends React.PureComponent {
           {
             selected: this.state.selectedId === child.props.id,
             tab: undefined,
-            tabHref: undefined,
+            tabHref: 'undefined',
```
sawyerh: Hm, why is this a string?

packages/core/src/components/Tabs/Tabs.jsx
sawyerh: Nit: Still curious about this. Looks like `tabFindIndex` is only ever called with `this.panelChildren()` as the first argument. We could dry up `tabSwitchLeft` and `tabSwitchRight` by moving that logic here and getting rid of the `tabsArr` argument.

packages/docs/src/scripts/helpers/polyfills.js
```diff
@@ -5,6 +5,7 @@
 // - github.com/facebook/react/issues/8379#issuecomment-264867090
 import 'core-js/fn/array/includes';
 import 'core-js/fn/array/fill';
+import 'core-js/fn/arrayfind-index';
```
sawyerh: We remove this from here, right? Looks like it was moved into Tabs.jsx. This also seems to break things...

image

packages/core/src/components/Tabs/Tab.jsx
```diff
@@ -65,6 +84,12 @@ Tab.propTypes = {
...
+   * [`SyntheticEvent`](https://facebook.github.io/react/docs/events.html),
+   * `id`, `panelId`
```
sawyerh: Just noticed I originally documented these arguments incorrectly (here and in the `onClick` documentation).

The order is incorrect and it's missing the `href` argument. This should actually be: `SyntheticEvent `, `panelId`, `id`, `href`

packages/core/src/components/Tabs/Tabs.test.jsx
```diff
@@ -22,6 +22,19 @@ function shallowRender(customProps = {}, children) {
   };
 }
 
+function deepMount(customProps = {}, children) {
```
sawyerh: Nit: Rather than creating a new method for using `mount` rather than `shallow`, we can follow the same [pattern we use in `VerticalNavItem.test.jsx`](https://github.com/CMSgov/design-system/blob/master/packages/core/src/components/VerticalNav/VerticalNavItem.test.jsx#L5-L16), where it supports passing in a `deep` boolean argument which determines whether `shallow` (default) or `mount` is used.

packages/docs/src/scripts/helpers/polyfills.js
1Copenut: Yes, we should be able to remove this here, since it was added to the Tabs.jsx component file.

1Copenut:
@sawyerh Changes are ready for review. I went ahead and refactored the tabSwitchLeft and tabSwitchRight into a single method called tabSwitcher, with an additional argument for pressed key. Looking at the code to handle key events, it made sense to structure the actual switcher the same way for future growth or change.

packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -190,37 +179,31 @@ export class Tabs extends React.PureComponent {
    * handleTabClick, so passing an interpolated hashbang plus selected Tab
    * string instead. 
    */
-  tabSwitchLeft(evt, panelId) {
+  tabSwitcher(pressed, evt, panelId) {
```
sawyerh: Seems like this could just become `handleTabKeyDown`, no?

sawyerh:
@1Copenut Looks good! I left one small comment above, but besides that, feel free to **Squash ad merge** when you're ready.

packages/core/src/components/Tabs/Tabs.jsx
1Copenut: I'm opting to rename it to handleTabKeyFocus, where handleTabKeyDown is already defined, but feel it does a good job explaining what the method does. Will push that change after testing, and re-run the build.

sawyerh: @1Copenut Sorry, to clarify: I was suggesting that what is currently in `handleTabKeyDown` seems redundant and `tabSwitcher` could essentially replace it, with a few minor changes (ie. moving the `evt.key` logic into this method). 

Let me know if it'd be easier to chat about this IRL.

1Copenut: @sawyerh Yes, after reading your comment, we could just make that one method easily. I'll follow up with you about a time to chat IRL.

packages/core/package.json
```diff
@@ -11,6 +11,7 @@
   "dependencies": {
     "@cmsgov/design-system-support": "^1.3.0",
     "classnames": "^2.2.5",
+    "core-js": "^2.5.1",
```
sawyerh: @1Copenut FYI. This needed added to the core package's `package.json` since each design system package can be installed separately, so any dependency imported into a core component need added.

packages/core/src/components/Tabs/Tabs.test.jsx
```diff
@@ -109,17 +109,48 @@ describe('Tabs', function() {
...
+        true
+      );
+      const panels = data.wrapper.find('TabPanel');
+      const tabs = data.wrapper.find('Tab');
+
+      tabs.first().simulate('keyDown', { key: 'ArrowRight' });
+
+      expect(panels.at(1).prop('selected')).toBe(true);
+      expect(tabs.at(1).prop('selected')).toBe(true);
+    });
```
sawyerh: @1Copenut I think I finally figured out why these new tests are failing. Looks like it's related to this: http://airbnb.io/enzyme/docs/guides/migration-from-2-to-3.html#for-mount-updates-are-sometimes-required-when-they-werent-before

TLDR: 

1. On line 132, we get the `panels` in their initial state (Panel 1 is selected)
2. On line 135, we change the state in the component. `panels` doesn't auto-update though, so it still thinks Panel 1 is selected, when in reality Panel 2 is now selected.

The solution appears to be doing the `data.wrapper.find...` again after the state is changed so that we get the components in their updated state.

sawyerh:
:shipit:

1Copenut:
Merging full left and right arrow functionality and passing test suite.

PR #163, opened 2017-10-09
Add support for React 16

sawyerh:
### Changed

- Add support for React 16.x to `design-system-core`

### Internal

- Fixed invalid `heading` tag (should have been `header`) on docs site 🙈 
- Major updates to React and Enzyme, which required updating some tests and changing how we render the docs when server-side rendering is used.

~There are still a few things I'm keeping an eye on before merging this in:~

- [x] [React 16 support in `react-hot-loader`](https://github.com/gaearon/react-hot-loader/pull/658)
- [x] [Lifecycle methods aren't called as expected when state changes](https://github.com/airbnb/enzyme/issues/1247)

### DIFF
--- a/package.json
+++ b/package.json
@@ -32,10 +32,13 @@
     "bytes": "^3.0.0",
     "cli-table": "^0.3.1",
     "colors": "^1.1.2",
+    "core-js": "^2.5.1",
     "cssnano": "^3.10.0",
     "cssstats": "^3.1.0",
     "del": "^3.0.0",
     "ejs": "^2.5.7",
+    "enzyme": "^3.1.0",
+    "enzyme-adapter-react-16": "^1.0.1",
     "eslint": "^4.8.0",
     "eslint-config-nava": "^2.1.0",
     "front-matter": "^2.2.0",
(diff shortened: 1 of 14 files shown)


sawyerh:
This was more painful than I expected, but I believe I've identified all the issues raised by upgrading to Enzyme v3.

pwolfert:
I haven't looked at the code yet, but I've just pulled it and installed the new deps and tried changing one of the examples, and it doesn't seem to do the hot reload like it's supposed to. I'll look at it more later. Tests run fine though. I haven't done the v16 upgrade yet on our other projects, and I'm not looking forward to the enzyme upgrade.

sawyerh:
> it doesn't seem to do the hot reload like it's supposed to

@pwolfert Good catch. Forgot to update the `react-hot-loader` dependency. This should be fixed now.

> I haven't done the v16 upgrade yet on our other projects, and I'm not looking forward to the enzyme upgrade.

Just to clarify, these changes won't require other projects to update right away. You can continue using React 15.

pwolfert:
> Just to clarify, these changes won't require other projects to update right away. You can continue using React 15.

I know; I just mean that you saying it was harder than you expected makes me wary of what's coming in my near future.

PR #165, opened 2017-10-12
Update button spacing to make top/left different from right/left

hannaliebl:
### Changed
Updated the padding on the Button component to `0.5em 1.25em`.

Before:
image

After:
image

----

For https://jira.cms.gov/browse/HDSG-151

### DIFF
--- a/packages/core/src/components/Button/Button.scss
+++ b/packages/core/src/components/Button/Button.scss
@@ -32,7 +32,7 @@ Style guide: components.button
   font-size: $base-font-size;
   font-weight: $font-bold;
   line-height: $input-line-height;
-  padding: $input-padding;
+  padding: 0.5em 1.25em;
   text-align: center;
   text-decoration: none;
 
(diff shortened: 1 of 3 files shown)


sawyerh:
So one thing I'm now realizing is that this now means the button will be a different height than an input field, breaking the ability to have a button to the side of a field. 

One option is to also reduce the padding of form fields (`input`, `select`, `textarea`) so that they remain consistent with the button padding. Another benefit of defining the `padding` using `em` units is that a designer can add a utility class to a field or button that changes the `font-size` and the padding would automatically adjust.

Here's a comparison of the current options:

image

And here is a comparison of a form:

image

@cindyphan any thoughts on this?

sawyerh:
Updated the input field to be consistent with the button spacing and switched over to `px` values. These button sizes are also now consistent with the sizes in the hcgov UI kit. cc @cindyphan @benkutil 

image

PR #171, opened 2017-10-23
Add a prop to expose a DOM ref on input to parent component

hannaliebl:
### Added
Added `fieldRef` prop to `TextField` to expose a ref on the input/textarea.


sawyerh:
@hannaliebl Looks good 👍  

What were the issues you mentioned having with unit tests? Anything I can help with?

hannaliebl:
Just not sure if a unit test should be added for this, and if so, how it would look. I guess there's not that much to actually test because the implementation can vary so much.

sawyerh:
@hannaliebl Ah got it. I think having a unit test to confirm this returns a reference to the expected component would be nice. One way we could do that is by setting the `defaultValue` prop and then checking the fieldRef's `value` to confirm it matches. That would verify that `fieldRef` returned a reference to the `input`.

packages/core/src/components/TextField/TextField.test.jsx
```diff
@@ -90,44 +91,59 @@ describe('TextField', function() {
...
     expect(data.wrapper.find('FormLabel').hasClass('bar')).toBe(true);
   });
 
+  it('returns reference to input field', () => {
+    let ref;
+    const data = render(
+      {
+        defaultValue: 'Yay',
+        fieldRef: el => {
+          ref = el;
```
sawyerh: @hannaliebl Here's how I got this to work. This is similar to how this prop would be used within a real world component, ie:

```jsx
<TextField fieldRef={el => { this.input = el } } />
```

PR #173, opened 2017-10-25
Add checkedChildren and uncheckedChildren props to Choice

sawyerh:
### Added

Added a `checkedChildren` and `uncheckedChildren` prop to the `Choice` component. These props conditionally display content below the Choice field when the field is checked or unchecked, respectively. 

No styling is add to the children, as there are many use cases for this, so I didn't want to add any constraints. Here's an example of one potential use case:

checkedchildren

This is a prerequisite for building out the [Exposed Within form pattern](https://jira.cms.gov/browse/HDSG-150).

----

### Note for reviewers

**I'd suggest viewing the diff with whitespaced ignore. You can [view that diff here](https://github.com/CMSgov/design-system/pull/173/files?w=1).**


packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -11,41 +11,88 @@ import uniqueId from 'lodash.uniqueid';
  * to the `input` element, so you can use this to set additional attributes if
  * necessary.
  */
-export const Choice = function(props) {
```
sawyerh: As I mention in the PR description, an easier way to view the diff for this file is to view the diff with whitespace ignored. You can view that diff here: https://github.com/CMSgov/design-system/pull/173/files?w=1

packages/core/src/components/ChoiceList/Choice.test.jsx
```diff
@@ -3,22 +3,32 @@ import FormLabel from '../FormLabel/FormLabel';
 import React from 'react';
 import { shallow } from 'enzyme';
 
-describe('Choice', () => {
-  const label = 'George Washington';
+const label = 'George Washington';
+
+function shallowRender(customProps = {}, children = label) {
```
sawyerh: Added this to make the other tests a lot more DRY.

packages/core/src/components/ChoiceList/Choice.test.jsx
```diff
@@ -31,200 +41,125 @@ describe('Choice', () => {
   });
 
   it('is not checked', () => {
-    const props = {
```
sawyerh: You can ignore these changes. This is just switching over to use the DRY'er `shallowRender` method. Skip down to the next comment where I point out what's actually new.

packages/core/src/components/ChoiceList/Choice.test.jsx
```diff
@@ -274,35 +209,167 @@ describe('Choice', () => {
     ).toBe(inputBId);
   });
 
+  describe('state', () => {
```
sawyerh: 👋 New tests below this line.

packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -11,41 +11,88 @@ import uniqueId from 'lodash.uniqueid';
...
+    }
+
+    if (!this.isControlled) {
+      this.setState({ checked: evt.target.checked });
+    }
+  }
+
+  render() {
+    /* eslint-disable prefer-const */
+    let {
```
pwolfert: It seems a shame for all of this to be declared with `let` when only `id` is changed.

packages/core/src/components/ChoiceList/Choice.test.jsx
```diff
@@ -3,22 +3,32 @@ import FormLabel from '../FormLabel/FormLabel';
...
-      name: 'presidents',
-      value: label
-    };
-    const wrapper = shallow(
-      <Choice {...props}>
-        <p>
-          <strong>Hello</strong> World
-        </p>
-      </Choice>
+    const data = shallowRender(
```
pwolfert: Could say `const { wrapper } = ` to keep the other lines the same.

sawyerh: oh nice, good point

PR #174, opened 2017-10-26
Fix:  Remove Chrome's default border-radius

sawyerh:
### Fixed

🤔 Seems like Chrome introduced a default `4px` `border-radius` on `button` elements. This removes that radius on the tabs and vertical nav items, which may be used on a `button` element.

| Before | After |
| ------ | ----- |
| image | image |

### DIFF
--- a/packages/core/src/components/Tabs/Tabs.scss
+++ b/packages/core/src/components/Tabs/Tabs.scss
@@ -88,9 +88,11 @@ Style guide: components.tabs.react-tabpanel
 Style guide: components.tabs.react-tab
 */
 .ds-c-tabs__item {
+  appearance: none;
   background-color: $color-background;
   border-bottom: 1px solid $border-color;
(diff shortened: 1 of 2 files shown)


packages/core/src/components/Tabs/Tabs.scss
```diff
@@ -131,7 +133,8 @@ Style guide: components.tabs.react-tab
     right: -1px;
     top: -1px;
     transform: scale3d(0, 1, 1);
-    transition: opacity $animation-speed-2 $ease-in-out-expo, transform $animation-speed-2 $ease-in-out-expo;
+    transition: opacity $animation-speed-2 $ease-in-out-expo,
```
sawyerh: (Prettier)

sawyerh:
Here's where this was introduced: https://www.chromestatus.com/features/5743649186906112

PR #181, opened 2017-10-27
Internal: Run prettier on Sass files

sawyerh:
### Internal

When I added prettier, I only ran it on `js` and `jsx` files, but not `scss` files. This does that, and also adds `stylelint-config-prettier` to disable any conflicting Stylelint rules.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,7 @@
 {
   "name": "design-system",
   "private": true,
-  "description":
-    "A shared set of design and code elements for creating accessible and consistent websites",
+  "description": "A shared set of design and code elements for creating accessible and consistent websites",
   "repository": "CMSgov/design-system",
   "scripts": {
     "build": "NODE_ENV=production gulp build --env=production",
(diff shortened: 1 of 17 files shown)


package.json
```diff
@@ -18,14 +17,11 @@
...
-      "eslint tools/ examples/**/src/ packages/**/src/ --ext .js --ext .jsx -f table && gulp lint --env=test --theme"
+    "lint": "eslint tools/ examples/**/src/ packages/**/src/ --ext .js --ext .jsx -f table && gulp lint --env=test --theme"
   },
   "devDependencies": {
-    "@cmsgov/eslint-config-design-system":
-      "file:./packages/eslint-config-design-system",
-    "@cmsgov/stylelint-config-design-system":
-      "file:./packages/stylelint-config-design-system",
+    "@cmsgov/eslint-config-design-system": "file:./packages/eslint-config-design-system",
+    "@cmsgov/stylelint-config-design-system": "file:./packages/stylelint-config-design-system",
```
sawyerh: Not exactly sure why Prettier sometimes formats these lines differently...

PR #182, opened 2017-10-28
Add modifier to hide alert icon

hannaliebl:
### Added
Added a modifier class for the alert component, `.ds-c-alert--hide-icon`, to hide the icon and adjust the left padding.

image

### DIFF
--- a/packages/core/src/components/Alert/Alert.scss
+++ b/packages/core/src/components/Alert/Alert.scss
@@ -69,7 +69,7 @@ $alert-icon-size: $spacer-5;
 }
 
 .ds-c-alert--hide-icon .ds-c-alert__body {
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Alert/Alert.scss
```diff
@@ -63,6 +68,10 @@ $alert-icon-size: $spacer-5;
   padding-left: $alert-icon-size + $alert-bar-size;
 }
 
+.ds-c-alert--hide-icon .ds-c-alert__body {
+  padding-left: $alert-padding;
```
sawyerh: I think we could remove the padding from this element completely so that the spacing is consistent on all sides:

image

hannaliebl:
PTAL

PR #185, opened 2017-11-05
Internal: Check for React component propTypes and description

sawyerh:
### Internal

- Fix issue where a React component without a `propTypes` definition would break the documentation page
- Check for a React component description before rendering its corresponding markup
- Add specs for `ReactContent`
- Update Enzyme


sawyerh:
@1Copenut This should fix the issue you were running into the other day

PR #196, opened 2017-11-28
MonthPicker

pwolfert:
A month picker for picking months:
month-picker

### Added
- `MonthPicker` component
- `labelClassName` prop for `FormLabel` component
- `ds-u-color--inherit` utility class


pwolfert:
@sawyerh it looks like Travis-CI doesn't have the English and Spanish locales installed. I think it's pretty safe to say that someone's browser has the locale a person is trying to use. Should I just remove those tests?

pwolfert:
retest this please

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+  }
+}
+
+MonthPicker.propTypes = {
+  /**
+   * The `input` field's `name` attribute
+   */
+  name: PropTypes.string.isRequired,
+  /**
+   * Locale string used to generate month names
```
sawyerh: ✍️ Copy suggestion: 

> A [BCP 47 language tag](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl#Locale_identification_and_negotiation) for month name localization. For example: Passing `es-US` as a value will render month names in Spanish.

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+  locale: PropTypes.string,
+  /**
+   * Applies the "inverse" UI theme
+   */
+  inversed: PropTypes.bool,
+  /**
+   * Array of 12 booleans representing 12 months, where a value of `true`
+   * means the corresponding month is disabled.
+   */
+  disabledMonths: PropTypes.arrayOf(PropTypes.bool),
```
sawyerh: (same comment as below, about passing month numbers)

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+   */
+  defaultSelectedMonths: PropTypes.arrayOf(PropTypes.bool),
+  onChange: PropTypes.func,
+  onSelectAll: PropTypes.func,
+  onClearAll: PropTypes.func,
+  /**
+   * For internationalization purposes, the text for the "Select all"
+   * button must be passed in as a prop.
+   */
+  selectAllText: PropTypes.string.isRequired,
```
sawyerh: We could set an initial value for this in `defaultProps` to support prototyping without i18n.  (See [`DateField.defaultProps`](https://github.com/CMSgov/design-system/blob/master/packages/core/src/components/DateField/DateField.jsx#L104-L114) for an example of how we're doing this in other components).

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+};
+
+export default MonthPicker;
+
+/**
+ * Generates an array of month names according to the given or default locale
+ *
+ * @param  {string} [locale] locale for generating month names
+ * @param  {boolean} [short] whether to return short month names
+ * @return {string[]}        array of short month names
```
sawyerh: tiny nit: It wouldn't return short names if `short` is `false` right?

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+          </Choice>
+        ))}
+      </ul>
+    );
+  }
+
+  renderButton(text, onClick) {
+    return (
+      <Button
+        variation="primary"
```
sawyerh: I'd push for keeping this as the default button style, at least by default. We could add support for setting what button variation gets used (ie. adding a `selectAllButtonVariation` and `clearAllButtonVariation` props). In general, it's frowned upon to have multiple primary call-to-actions on the same page and I'd like to avoid encouraging it. (For HealthCare.gov we actually use the "success" variation as our primary CTA, so I understand why this is like it is).

Here's with the default button style looks like, which I think I actually prefer since these actions don't seem like they deserve as much focus:

image

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+
+  renderMonths(selectedMonths) {
+    const disabledMonths = this.disabledMonths();
+    const { name, inversed } = this.props;
+    return (
+      <ul className="ds-c-list--bare ds-u-display--flex ds-u-justify-content--between ds-u-flex-wrap--wrap">
+        {this.months.map((month, i) => (
+          <Choice
+            name={name}
+            value={i}
```
sawyerh: This is a zero-based index, which might result in unexpected values for folks. For example, January's `value` would be submitted as `0`.

packages/core/src/components/MonthPicker/MonthPicker.scss
```diff
@@ -0,0 +1,17 @@
+@import '@cmsgov/design-system-support/src/index';
+
+/*
+Month Picker
+
+@uswds https://standards.usa.gov/components/form-controls
+
```
sawyerh: Let's add an HTML example? We could leave out the "Select/Clear all" buttons though since those require JS.

packages/core/src/components/MonthPicker/MonthPicker.scss
```diff
@@ -0,0 +1,17 @@
+@import '@cmsgov/design-system-support/src/index';
+
+/*
+Month Picker
+
+@uswds https://standards.usa.gov/components/form-controls
```
sawyerh: These links are meant to link to the related pattern in USWDS, but USWDS doesn't have a Month Picker component, so we can remove this.

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+      >
+        {text}
+      </Button>
+    );
+  }
+
+  render() {
+    const { selectAllText, clearAllText } = this.props;
+    return (
+      <div className="ds-c-month-picker ds-u-margin-y--3">
```
sawyerh: Should this component also render the field's label, like we do with `DateField`? Semantically, these choices would be wrapped in a `fieldset` and the label would be rendered as a `legend`.

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+    if (!this.isControlled) {
+      this.setState({ selectedMonths: this.monthsArray(false) });
+    }
+  }
+
+  renderMonths(selectedMonths) {
+    const disabledMonths = this.disabledMonths();
+    const { name, inversed } = this.props;
+    return (
+      <ul className="ds-c-list--bare ds-u-display--flex ds-u-justify-content--between ds-u-flex-wrap--wrap">
```
sawyerh: Nit: I think `ol` would be more semantic for this list since the order is meaningful.

Via [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol):

> The `<ol>` and `<ul>` both represent a list of items. They differ in the way that, with the `<ol>` element, the order is meaningful. As a rule of thumb to determine which one to use, try changing the order of the list items; if the meaning is changed, the `<ol>` element should be used, else the `<ul>` is adequate.

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+   */
+  selectedMonths: PropTypes.arrayOf(PropTypes.bool),
+  /**
+   * Array of 12 booleans representing 12 months, where a value of `true`
+   * means the corresponding month is selected by default.
+   * Sets the initial checked state for the 12 month checkboxes. Use this for
+   * an uncontrolled component; otherwise, use the `selectedMonths` property.
+   */
+  defaultSelectedMonths: PropTypes.arrayOf(PropTypes.bool),
+  onChange: PropTypes.func,
```
sawyerh: ✍️ Documentation suggestion for `onChange`:

> A callback function that's invoked when a month's `checked` state is changed. Note: This callback is not called when a month is selected or deselected via the "Select all" or "Clear all" buttons – use the `onSelectAll` and `onClearAll` event handlers for those instances.

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: Doh

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: They are passing in an array representing months, though, so I'd think they should expect a zero-based index.

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: @cindyphan what do you think about this visually?

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: Good idea 👍

packages/core/src/components/MonthPicker/MonthPicker.jsx
sawyerh: To clarify: The blue button style would still be possible if we want to do that on HealthCare.gov (using the props I suggest above), however what I'm advocating for here is to not set these buttons as primary buttons by default because sites implementing the design system may actually be using the blue button as their primary button style (since that's the standard).

packages/core/src/components/MonthPicker/MonthPicker.jsx
sawyerh: @pwolfert No one is passing in months though, and this is somewhat related to my other comment about sending an array of month numbers.

sawyerh:
> it looks like Travis-CI doesn't have the English and Spanish locales installed. I think it's pretty safe to say that someone's browser has the locale a person is trying to use. Should I just remove those tests?

@pwolfert This seems related to Node or JSDom (the tests fail locally too). I didn't dig too deep, but seems like it might be related to [what ICU data is embedded](https://nodejs.org/api/intl.html#intl_embed_a_limited_set_of_icu_data_small_icu) /shrug

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: @cindyphan, thoughts on having a visible legend/label?
@sawyerh @1Copenut significance of having a screen-reader-only legend?

sawyerh: You would need a label, no? Otherwise how would a user know what the months are being asked for? 😅

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+   * means the corresponding month is disabled.
+   */
+  disabledMonths: PropTypes.arrayOf(PropTypes.bool),
+  /**
+   * Array of 12 booleans representing 12 months, where a value of `true`
+   * means the corresponding month is selected.
+   * This will render a read-only field. If the field should be mutable,
+   * use `defaultSelectedMonths`.
+   */
+  selectedMonths: PropTypes.arrayOf(PropTypes.bool),
```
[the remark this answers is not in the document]
pwolfert: Okay, now I see it. Yeah, that does seem nicer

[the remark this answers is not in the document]
sawyerh: 🙏

packages/core/src/components/MonthPicker/MonthPicker.jsx
1Copenut: Instead of a legend, I would opt for a descriptive header that the UL is described by. Something like:
`<h4 id="month-checklist-header">Select months for your situation</h4>
  <ul
    aria-labelledby="month-checklist-header"
    className="ds-c-list--bare ds-u-display--flex ds-u-justify-content--between ds-u-flex-wrap--wrap"
  >`

This gives the list a nice header that is read out when the first list item is focused.

1Copenut: I lean more toward a `hN` than a `legend` because this may or may not be contained in a `fieldset`. The months grid could be in a div, or other component, so having a descriptive header that screen readers can hook into via the navigate by header or header rotor seems the best fit.

1Copenut: Reading this thread a little more closely, I'll qualify my previous comment. I'd still like to see a descriptive header, that could be a legend, presented visually. 

The header/legend could also contain some screen reader only text wrapped in a span, describing how to use the grid's select all and clear all features.

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: I didn't know we were doing this as it could cause users to forget that it has to be translated

sawyerh: We should definitely encourage and support localization, but I don't think we should require it. I think it's okay to have a sane default when we're able to.

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: Correct, and?

packages/core/src/components/MonthPicker/MonthPicker.jsx
sawyerh: Good discussion 👍 I mentioned wrapping it in a `fieldset` since I believe this is what you should typically use to wrap `input`'s and `label`'s ([MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset)). This is also [how the ChoiceList component is setup](https://github.com/CMSgov/design-system/blob/master/packages/core/src/components/ChoiceList/ChoiceList.jsx#L128-L143), and the `MonthPicker` is basically a `ChoiceList` that is layed out differently.

packages/core/src/components/MonthPicker/MonthPicker.jsx
sawyerh: ah, it says it returns "array of *short* month names". The tiniest of nits 🙈

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: Oh, I said "Doh" because when I first read this I thought we were talking about SubSteps. So don't make the same comment on that one until I change it.

packages/core/src/components/MonthPicker/MonthPicker.jsx
1Copenut: I did some quick screen reader testing with OSX + VoiceOver. Here's what I found:
* The screen reader only legend text is read every time a checkbox receives focus. This happens because the divs are treated as unique entities, and not part of a list.
* The legend is only read once if the checkboxes are grouped as part of a list, wrapped in list items like on the WDS example: https://standards.usa.gov/components/form-controls/#checkboxes

1Copenut: I haven't evaluated this for NVDA or JAWS, but my thinking is they will behave in same/similar fashion. The use of divs over a list for the rendered output wrappers might be a separate discussion; I only brought it up here because it affects how screen reader(s) read the legend text.

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -0,0 +1,220 @@
...
+          <Choice
+            name={name}
+            value={i}
+            checked={selectedMonths[i]}
+            key={month}
+            onChange={e => this.handleChange(e)}
+            className="ds-c-month-picker__month"
+            disabled={disabledMonths[i]}
+            inversed={inversed}
+            aria-label={this.monthsLong[i]}
```
1Copenut: This is a great call, having the aria-label. It's well tested, and lets us have short names for visual, and full names for aural experiences, respectively.

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: That's unfortunate, because that means we would have to modify `Choice` to get that effect. What if they were wrapped in a `span` instead of a `div`? Does that make a difference?

1Copenut: I'm not sure spans would make a difference; my guess is probably not. An (un)ordered list has an intrinsic relationship to its list items, and screen readers pick up on that. Non-semantic containers like divs and spans don't make that relationship on their own.

As I typed this response, I considered the prospect of adding `role="list"` and `role="listitem"` to the container div, and child divs. StackOverflow seems to support this possibility, and could give us the binding relationship we are looking for: https://stackoverflow.com/a/11131648

1Copenut: I will need to research further for the right combination of roles, tabindex, and placement for all screen reader combinations. There is promise in my previous comment, but it will be an enhancement when I have a bit more time to research across the major assistive devices.

packages/core/src/components/MonthPicker/MonthPicker.jsx
pwolfert: My confusion on this arose from the fact that the solid blue is the secondary color in HC.gov. They're not primary anymore, and I've added an optional `buttonVariation` prop.

packages/core/src/components/MonthPicker/MonthPicker.jsx
sawyerh: @pwolfert Is this the last piece preventing a merge?

Sounds like there are two threads here:

1. Whether `MonthPicker` should render the "label" (via a `legend`) for these checkboxes, like `ChoiceList`
2. and the HTML semantics related to the root container, checkbox containers, and label (if the answer to 1. is yes)

for 1, I'm struggling to think of a case where you wouldn't want a label for these fields. I'd be curious to see this component in the context of how y'all are using it in Tax Tools. I assume it has a label asking the user something like "Select the months you lived in X"?

pwolfert:
Okay, tests are passing again with new features. I'm submitting this for final review.

PR #197, opened 2017-11-28
StepList

pwolfert:
Added a `StepList` component and documentation.

### Added
- `StepList` and sub-components

### Changed
- Changed test "Components index: exports all components except ones we don't want to expose" to ignore a list of components. I don't want to expose the sub-components. They're not useful to anyone and would just require more documentation.

### Notes
- New errors in `react-docgen` for the sub-components because I import the `stepShape` object from another module instead of repeating it inline. This has no effect other than to annoy since the propTypes documentation for those components is never printed anywhere.
- It seems that eslint still checks some test files according to the old standards instead of with `prettier` standards


packages/core/src/components/StepList/Step.jsx
```diff
@@ -0,0 +1,82 @@
...
+export const Step = ({ step, ...props }) => {
+  const start = step.isNextStep;
+  const resume = step.started && !step.completed;
+  const className = classNames('ds-c-step', {
+    'ds-c-step--current': start || resume,
+    'ds-c-step--completed': step.completed
+  });
+  return (
+    <li className={className}>
+      <h2 className="ds-c-step__header ds-h3 ds-u-display--flex ds-u-justify-content--between">
```
1Copenut: https://cms.invisionapp.com/share/STEPOUPYB#/screens is the visual I'm comparing to, and it would seem H2 is the right tag, when thinking about it from a screen reader user who navigates via headers. `Get information to figure out your premium tax credit` feels like an H1, because it states the page's intent, and these steps would be the next logical hierarchical stopping point.

packages/core/src/components/StepList/StepLink.jsx
```diff
@@ -0,0 +1,39 @@
...
+
+  render() {
+    const { route, screenReaderText, className, children } = this.props;
+    const onClick = event => this.handleClick(event);
+    return (
+      <a href={route} onClick={onClick} className={className}>
+        {children}
+        <span className="ds-u-visibility--screen-reader">
+          {' '}
+          {screenReaderText}
```
1Copenut: I would expect this text to read out as "Edit a tax year" or "Edit household details". It might take some pruning and careful stemming of the existing text strings, but will be worth doing to give assistive device users a clear call to action with natural language.

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,291 @@
...
+    <li class="ds-c-step ds-c-step--completed">
+      <h2 class="ds-c-step__header ds-h3 ds-u-display--flex ds-u-justify-content--between">
+        <div class="ds-c-step__title">Choose a tax year</div>
+        <div>
+          <div class="ds-c-step__completed">
+            <span>Completed</span>
+            <a href="#step-1">
+              Edit
+              <span class="ds-u-visibility--screen-reader">
+                "Choose a tax year"
```
1Copenut: This sentence feels uneven. I'd like it to read "Edit tax year" for a clear call to action.

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,291 @@
...
+      </h2>
+    </li>
+    <li class="ds-c-step ds-c-step--current">
+      <h2 class="ds-c-step__header ds-h3 ds-u-display--flex ds-u-justify-content--between">
+        <div class="ds-c-step__title">Enter household details</div>
+        <div>
+          <a href="#step-2" class="ds-c-button ds-c-button--success">
+            Resume
+            <span class="ds-u-visibility--screen-reader">
+              "Enter household details"
```
1Copenut: Let's have this read "Resume `<span class="ds-u-visibility--screen-reader">household details</span>`.

packages/core/src/components/StepList/StepLink.jsx
pwolfert: Because of the differences in grammar between languages, I'm not comfortable composing sentences with smaller translated parts, so this would require us to pass multiple strings for each step, which is why I went with this method. We'd need to pass "Resume selecting a tax year", "Edit tax year", "Start selecting a tax year", "Select a tax year", which seems like a lot

pwolfert: Is there a simpler way we could do this without it being jarring for users?

packages/core/src/components/StepList/StepList.example.jsx
```diff
@@ -0,0 +1,73 @@
+/* eslint-disable react/display-name */
+import React from 'react';
+import StepList from './StepList';
+
+export default function() {
+  return (
+    <div style={{ maxWidth: '628px' }}>
+      <StepList
+        inverse
```
sawyerh: `inverse` probably shouldn't be here (it's not a supported prop)

packages/core/src/components/StepList/StepList.example.jsx
```diff
@@ -0,0 +1,73 @@
...
+        completedText="Completed"
+        editText="Edit"
+        resumeText="Resume"
+        startText="Start"
+        onEnterStep={() => {}}
```
sawyerh: This could be potentially confusing to folks. I'd rather have this do something, or not show it if it's an empty function.

packages/core/src/components/StepList/StepLink.jsx
```diff
@@ -0,0 +1,39 @@
...
+    );
+  }
+}
+
+StepLink.propTypes = {
+  /**
+   * Label text or HTML.
+   */
+  children: PropTypes.node.isRequired,
+  route: PropTypes.string.isRequired,
```
sawyerh: Could we rename this to `url` or `href` to make it more explicit?

packages/core/src/components/StepList/StepLink.jsx
```diff
@@ -0,0 +1,39 @@
...
+  stepId: PropTypes.string.isRequired,
```
sawyerh: What's the id used for? Does it have to be something that's required?

packages/core/src/components/StepList/StepLink.jsx
```diff
@@ -0,0 +1,39 @@
...
+        <span className="ds-u-visibility--screen-reader">
+          {' '}
+          {screenReaderText}
+        </span>
+      </a>
+    );
+  }
+}
+
+StepLink.propTypes = {
```
sawyerh: I'd love to see these documented on the docs page – I think some people (including myself) would find it valuable.

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,291 @@
+@import "@cmsgov/design-system-support/src/settings/index";
+
+/*
+Step List
```
sawyerh: One thing currently not implemented are the responsive styles. Here was the original design for that:



The font sizes in the mobile version are:

| Element | Size |
| -------- | ---- |
| Step number | `$small-font-size` | 
| Title | `$h5-font-size` |
| Description | `$small-font-size` | 
| Resume/Completed/Edit | `$small-font-size` |

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,291 @@
...
+Style guide: components.step-list.react
+*/
+
+/*
+Step object
+
+A `step` object contains all information needed to render the step, including
+text, progress, and routing information.
+
+Step object properties:
```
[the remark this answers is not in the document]
pwolfert: For none of the sub-components though is the `step` object part of the root props object, so this won't make a difference.

packages/core/src/components/StepList/Step.jsx
```diff
@@ -0,0 +1,82 @@
...
+        <div className="ds-c-step__title">{step.title}</div>
+        <div>
```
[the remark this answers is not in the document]
pwolfert: I was thinking it would be beneficial for screen readers to have this as part of the heading, but I'll defer to @1Copenut on that.

packages/core/src/components/StepList/Step.jsx
```diff
@@ -0,0 +1,82 @@
...
+          {step.completed && (
+            <div className="ds-c-step__completed">
+              <span>{props.completedText}</span>
+              {!step.steps && (
+                <StepLink
+                  route={step.route}
+                  stepId={step.id}
+                  screenReaderText={`"${step.title}"`}
```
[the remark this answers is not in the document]
pwolfert: This is because of the discussion above with Trevor. I didn't want to have to send in five strings representing the different sentences, so I thought it would be easier to have the screen reader read 'Resume "Choose a tax year"'

packages/core/src/components/StepList/StepLink.jsx
```diff
@@ -0,0 +1,39 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export class StepLink extends React.PureComponent {
+  handleClick(event) {
+    if (this.props.onEnterStep) {
```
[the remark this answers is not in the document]
pwolfert: This was a decision made after discussions and much deliberations. The problem with `onClick` is that it's ambiguous. What's being clicked? And this function is only for routing that is triggered by clicking certain elements. I'm willing to budge on this, but you'll have to argue your case.

packages/core/src/components/StepList/StepLink.jsx
pwolfert: I can see `href` making it more clear.

packages/core/src/components/StepList/StepLink.jsx
pwolfert: Really it should be required in the `stepShape` because it's used as a key in the lists. We know that if it is a required part of the step that it's always going to be passed to this component by the components that use it (because we control the only components that use it, and I purposely forced the package not to expose those sub-components).

packages/core/src/components/StepList/StepList.example.jsx
pwolfert: whoops

pwolfert: left over from early tests before talking to Cindy about inverse support

packages/core/src/components/StepList/Step.jsx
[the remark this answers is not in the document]
1Copenut: I concur with @pwolfert about not wanting to pass in multiple strings. When he explained the logistics of it, I realized I was asking a lot for contextual differences. A string like "Resume Choose a tax year" or "Edit Choose a tax year", is pretty clear, and more actionable than just Resume, even if it's not a completely custom block of micro-content.

pwolfert:
Tests are passing

packages/core/src/components/StepList/StepList.scss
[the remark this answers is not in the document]
sawyerh: Oh hm, maybe I'm misunderstanding how this is all connected. I thought these were props of `Step.jsx`. I guess that would be the case if `step` wasn't a prop on `<Step>` and instead, the properties of the `step` object were top-level props on the `<Step>` component? Was there a reason for setting it up that way, rather than destructuring it like below?

```jsx
<Step onEnterStep={blah} resumeText="Resume" {...step} />
```

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,356 @@
...
+    <li class="ds-c-step">
+      <h2 class="ds-c-step__header">
+        <div class="ds-c-step__title">View premium results</div>
+        <div class="ds-c-step__actions" />
+      </h2>
+    </li>
+  </ol>
+</div>
+
+Style guide: components.step-list
```
sawyerh: Cross posting from Slack...

image

This would mean swapping `components` with `patterns` here.

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,356 @@
...
+form. It serves as a table of contents and a way to quickly see where they are
+and what they should be working on next.
+
+Markup:
+<div style="max-width: 628px">
+  <ol class="ds-c-step-list ds-u-margin-top--4">
+    <li class="ds-c-step ds-c-step--completed">
+      <h2 class="ds-c-step__header">
+        <div class="ds-c-step__title">Choose a tax year</div>
+        <div class="ds-c-step__actions">
```
sawyerh: Depending on the conclusion of [the discussion around the placement of these actions on mobile and screen readers](https://teamwds.slack.com/archives/C7S6GJKC2/p1512149247000013), we may need to adjust the markup of this section. Based on the current discussion, it sounds like we want these actions read after the title and description, so this markup would need moved after those. I think this actually makes life easier for us when it comes to how this adjusts for mobile and desktop. We could change the markup so that it's two columns, which collapse to one column for small screens. Here's some pseudo code of what I had in mind when working on the design of this...

```html
<li class="ds-c-step">
  <!-- Left column on desktop -->
  <div class="ds-c-step__body">
    <h2 class="ds-c-step__header">Foo</h2>
    <p>Description text</p>
    <ol class="ds-c-step__substeps">...</ol>
  </div>

  <!-- Right column on desktop -->
  <div class="ds-c-step__actions">
     <a class="ds-c-button" href="#">Start</a>
  </div>
</li>
```

The only tricky thing this introduces is the alignment of the substep "Edit" links, but I think this is easier to solve with absolute positioning (`right: 0`).

packages/core/src/components/StepList/StepList.scss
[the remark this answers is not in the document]
pwolfert: I wanted to keep the `step` information all in its own object so it could be passed down to children in the hierarchy without getting muddied by the individual components' props—because not all components that take a step need the same props

packages/core/src/components/StepList/StepList.scss
1Copenut: I pulled Patrick's code base, and mocked up some HTML by hacking on the example view. This moves the primary actions to the bottom of a `<Step />`, beneath the description and the sub-actions. It also adds header 3's for the description, sub-actions, and actions, that are available only to screen readers. These could easily be updated via className to be visible if wanted.

```
<div style="max-width: 628px">
  <ol class="ds-c-step-list ds-u-margin-top--4">
    <li class="ds-c-step ds-c-step--completed">
      <h2 class="ds-c-step__header">
        <span class="ds-c-step__title">Choose a tax year</span>
      </h2>
      <div class="ds-c-step__actions" aria-labelledby="step-1-actions">
        <h3 class="ds-u-visibility--screen-reader" id="step-1-actions">Primary actions for Choose a Tax Year</h3>
        <div class="ds-c-step__completed-text">
          <span class="ds-u-visibility--screen-reader">"Choose a tax year"</span> Completed
        </div>
        <a href="#step-1">
          Edit<span class="ds-u-visibility--screen-reader">
            {" "}
            "Choose a tax year"
          </span>
        </a>
      </div>
    </li>
    <li class="ds-c-step ds-c-step--current">
      <h2 class="ds-c-step__header">
        <span class="ds-c-step__title">Enter household details</span>
      </h2>
      <div aria-labelledby="step-2-desc" class="ds-c-step__description">
        <h3 class="ds-u-visibility--screen-reader" id="step-2-desc">
          Step 2 Description
        </h3>
        <p>
          Answer questions about who in your household qualifies for a premium tax
          credit and information on each person, including date of birth,
          location(s) they lived in for the year, and months of marketplace
          coverage.
        </p>
      </div>
      <div aria-describedby="step-2-sub-actions" class="ds-c-step__substep">
        <h3 class="ds-u-visibility--screen-reader" id="step-2-sub-actions" >Secondary actions for Household Details</h3>
        <ol class="ds-c-step__substeps">
          <li class="ds-c-substep">
            <div class="ds-c-substep__line">
              <div class="ds-c-substep__title">Overall household</div>
              <a href="#step-2a">
                Edit<span class="ds-u-visibility--screen-reader">
                  {" "}
                  "Overall household"
                </span>
              </a>
            </div>
          </li>
          <li class="ds-c-substep">
            <div class="ds-c-substep__line">
              <div class="ds-c-substep__title">Bob's information</div>
            </div>
          </li>
          <li class="ds-c-substep">
            <div class="ds-c-substep__line">
              <div class="ds-c-substep__title">Barb's information</div>
            </div>
          </li>
        </ol>
      </div>
      <div aria-labelledby="step-2-actions" class="ds-c-step__actions">
        <h3 class="ds-u-visibility--screen-reader" id="step-2-actions">Primary actions for Household Details</h3>
        <a href="#step-2" class="ds-c-button ds-c-button--primary">
          Resume<span class="ds-u-visibility--screen-reader">
            {" "}
            "Enter household details"
          </span>
        </a>
      </div>
    </li>
    <li class="ds-c-step">
      <h2 class="ds-c-step__header">
        <div class="ds-c-step__title">Review your information</div>
        <div class="ds-c-step__actions">
      </div></h2>
    </li>
    <li class="ds-c-step">
      <h2 class="ds-c-step__header">
        <div class="ds-c-step__title">View premium results</div>
        <div class="ds-c-step__actions">
      </div></h2>
    </li>
  </ol>
</div>
```

pwolfert: @1Copenut I'm working on implementing your headings and described-bys and such. Do you have examples of sub-action headings and description headings that reuse the step titles? Should we be saying "Description for "Enter household details""? I haven't been passing in the step number, but I can do that if you'd prefer it to be "Step 2 Description"

packages/core/src/components/StepList/StepLink.jsx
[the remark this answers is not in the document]
pwolfert: I didn't realize you originally just meant this one sub-component. That makes sense in this context.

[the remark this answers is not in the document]
pwolfert: At first I didn't want something long-winded because of code formatting, but all the declaration of StepLinks are multi-line anyway. `onStepLinkClick` makes sense. It also wasn't' originally called `StepLink`

packages/core/src/components/StepList/Step.jsx
```diff
@@ -0,0 +1,85 @@
...
+  });
+  const contentClassName = classNames('ds-c-step__content', {
+    'ds-c-step__content--with-content': step.description || step.steps
+  });
+  // const actionsId = `step-${step.id}-actions`;
+  // const subActionsId = `step-${step.id}-sub-actions`;
+  return (
+    <li className={className}>
+      <div className={contentClassName}>
+        <h2 className="ds-c-step__title">{step.title}</h2>
```
sawyerh: @pwolfert In our design system sync today, @cindyphan @1Copenut and I chatted and decided that from an accessibility and usability standpoint, we'd like to also link this title to the same place as the primary Start/Resume/Edit button, when the step is completed or in progress. That would also mean that this title would be the default link style (blue and underlined) for both of these states.

pwolfert: I'll upload the change soon. The titles are so ugly now, but I guess that's the price we pay for usability.

pwolfert: I know it's not as consistent, but the link for completed titles might almost be bearable to my eyes if it were black.

sawyerh: Hm..I don't find it to be that ugly, but this is probably one of those subjective things and influenced by how you've become accustomed to seeing it. I think the styling we use for links should be as consistent as possible. Here's how it looked in the original design comps for App 3:

image

sawyerh:
> New errors in `react-docgen` for the sub-components because I import the `stepShape` object from another module instead of repeating it inline. This has no effect other than to annoy since the `propTypes` documentation for those components is never printed anywhere.

Dug into this a little bit. Appears to be related to this: https://github.com/reactjs/react-docgen/issues/150

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,413 @@
...
+    font-size: $base-font-size;
+    text-align: right;
+  }
+}
+
+.ds-c-step__completed-text {
+  display: inline-block;
+  line-height: $step-button-height - 12px;
+  margin-right: $spacer-2;
+  padding-top: 6px;
```
sawyerh: Last nit, I promise: When the actions are collapsed, this padding is causing too much space to be above. We'd only want this padding when the actions aren't collapsed — since that centers the completed text with the title, which is nice.

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,413 @@
...
+links.
+
+Visually "Start" and "Resume" are buttons, but all three are implemented
+with `<a>` tags. In the React component, the links' `href` properties are
+determined by the `step` object's `href` property. One can also
+optionally pass an `onStepLinkClick` function that will cancel the default
+link behavior and call `onStepLinkClick` with `href` as a parameter for
+apps that handle routing with JavaScript.
+
+When a step has substeps and is incomplete, the `href` property of should
```
sawyerh: Typo?

> the href property **of should be set**

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,413 @@
...
+}
+
+/*
+Start, Resume, and Edit links
+
+A user interacts with the steps through "Start", "Resume", and "Edit"
+links.
+
+Visually "Start" and "Resume" are buttons, but all three are implemented
+with `<a>` tags. In the React component, the links' `href` properties are
```
sawyerh: ✍️ We could cut this first sentence. These are details people can inspect if they're really curious, but I think for the majority of people they're not going to care much about what HTML elements we use.

packages/core/src/components/StepList/StepList.jsx
```diff
@@ -0,0 +1,80 @@
...
+  <ol className="ds-c-step-list ds-u-margin-top--4">
+    {steps.map(step => <Step step={step} key={step.id} {...props} />)}
+  </ol>
+);
+
+// Define the shape of a single step so we can recursively define the shape
+// as well as reuse it in multiple components' prop-type definitions. This
+// has to be in this file or else the docs generator will break.
+export const stepShape = {
+  id: PropTypes.string.isRequired,
```
sawyerh: I think I left a comment on this before, but just flagging I'm still not sure why an `id` is a required prop

packages/core/src/components/StepList/StepList.jsx
```diff
@@ -0,0 +1,80 @@
...
+  descriptionLabelText: 'Description for %{step}',
+  substepsLabelText: 'Secondary actions for %{step}'
+};
+
+StepList.propTypes = {
+  /**
+   * An array of step objects that contain information needed to render
+   * them like text, state, and link/button URLs.
+   * See [Start, Resume, and Edit links]({{root}}/components/step-list/#components.step-list.buttons)
+   * and [Step object]({{root}}/components/step-list/#components.step-list.step-object)
```
sawyerh: These links 404 after we've moved the page to `/patterns/`

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,413 @@
...
+these substeps span multiple pages and have their own unique URLs. If, for
+example, the user completes the first page of the `household > overall` substep
+where they list the household members but has not completed the second page
+where they define those members' relationships to each other, we want the
+"Resume" button to take them back to the relationships page and not the first
+page where they entered their names. This, of course, requires an extra
+steps-building process to update a top-level steps' `href` property by
+traversing the substep tree to find the first incomplete step. See the
+[Completed, started, and isNextStep](#completed-started-and-isnextstep) section
+below an for an example JavaScript function that can change the `href` of steps
```
sawyerh: Typo?

> section below **an for an**

packages/core/src/components/StepList/StepList.scss
```diff
@@ -0,0 +1,413 @@
...
+- **title** (`string`) - Text to display as the step title
+- **description** (`string`) - Additional text to display under (only rendered for top-level steps)
+
+Style guide: patterns.step-list.step-object
+*/
+
+/*
+---
+
+## Creating steps
```
sawyerh: ✍️ We might want to frame this as "Advanced usage: Maintaining state of complex lists" or something. This tab usually provides basic guidance that anyone implementing a pattern should follow, but while this content is useful, I'd be wary of framing it as something that applies to every implementation.

packages/core/src/components/StepList/StepList.scss
pwolfert: For the record, visual details aren't a nit to me. I like things pixel-perfect.

packages/core/src/components/StepList/StepList.jsx
pwolfert: Because it's used as the key when listing steps

packages/core/src/components/StepList/StepList.scss
pwolfert: I feel like the `hrefs` for "Resume" corresponding to the current step is important for user experience. It is not needed when there are no substeps, but I don't think it's advanced. I like the idea for changing the title to mention state management. What about "Managing list state". I get how it could be "advanced" usage, but it's really stuff that should be observed for any list

pwolfert: I'm just pushing back a little bit on this. I think it's important

packages/core/src/components/StepList/StepList.jsx
sawyerh: Does this need to be explicitly defined or can we generate it? Seems tedious to require it if we can generate it. We use `lodash.uniqueid` in a few other components to generate ID's for a11y purposes. Could we do that for these?

packages/core/src/components/StepList/StepList.scss
sawyerh: Yeah, agreed that it's important that "Resume" takes you to the current step. "Managing list state" works for me. My main concern was that "Creating steps" could be potentially confusing for simple lists.

packages/core/src/components/StepList/StepList.jsx
pwolfert: I mean, really I could have it use the array index. That would be really simple, and the order of these should not be changing. I was just avoiding it because I already had it as a required field because they were necessary for our app. I'll have it use the id if it's there and the index if it's not

pwolfert: I just didn't want to make my map-function code any larger

PR #198, opened 2017-11-30
Add CSS and HTML for Dialog component

sawyerh:
_Note_: I'm breaking this component out into several separate PR's to make it easier to review: This first one which includes just the HTML and CSS, a second one which will be coming either tomorrow or next week which includes the JS/React, and likely a third one which includes more detailed guidance.

### Added

Added a Dialog CSS component (`.ds-c-dialog`). It's structure is similar to other dialog components:

1. A header with a close button and optional title.
2. A body area that takes any HTML you throw at it.
3. An optional actions bar at the bottom. You can pass this any HTML, so the buttons visible here is up to you.
4. An "underlay" which dims the underlying page.

The component also has three modifiers that can be used to modify the width of the modal (`narrow`, `wide`, `full`).

image


sawyerh:
@asarazor Here's the first pass at the modal component (see my note in the PR description about future work coming in different PR's)

package.json
```diff
@@ -41,13 +41,13 @@
     "cssstats": "^3.1.0",
     "del": "^3.0.0",
     "ejs": "^2.5.7",
-    "enzyme": "^3.1.1",
```
sawyerh: Ignore these...I've been using this branch to test https://github.com/CMSgov/design-system/pull/199

packages/core/src/components/Dialog/Dialog.scss
```diff
@@ -0,0 +1,120 @@
...
+The dialog component can be used to focus a user's attention on a single piece of content, without taking them to a new screen. Please use with caution; view our guidance for more details.
+
+.ds-c-dialog--narrow - A dialog with a narrow measure
+.ds-c-dialog--wide - A dialog with a wide measure
+.ds-c-dialog--full - A dialog spanning a normal page width
+
+Markup:
+<div data-demo="This div is for demo purposes only" style="min-height: 400px;">
+  <div class="ds-c-dialog-wrap">
+    <div class="ds-c-dialog {{modifier}}" role="dialog" aria-labelledby="dialog-title">
```
sawyerh: cc @1Copenut for a11y review

packages/core/src/components/Dialog/Dialog.scss
```diff
@@ -0,0 +1,120 @@
...
+- [Alert]({{root}}/components/alert)
+
+## Learn more
+
+- [Making an accessible dialog box](https://www.nczonline.net/blog/2013/02/12/making-an-accessible-dialog-box/)
+- [GOV.UK modal dialog boxes discussion](https://paper.dropbox.com/doc/Modal-dialog-boxes-jbsTPoITg37IIc6ybjetM)
+- [Overuse of Overlays: How to Avoid Misusing Lightboxes](https://www.nngroup.com/articles/overuse-of-overlays/)
+- [Using ARIA role=dialog to implement a modal dialog box](https://www.w3.org/WAI/GL/wiki/Using_ARIA_role%3Ddialog_to_implement_a_modal_dialog_box)
+
+Style guide: components.dialog.guidance
```
sawyerh: Additional guidance will be added in a separate PR

packages/support/src/tools/_mixins.button.scss
```diff
@@ -0,0 +1,7 @@
+@mixin base-button {
```
pwolfert: I'm not seeing this used anywhere else. Do you have a plan for this?

sawyerh: Good catch – this is a leftover from how I was approaching the close button originally. Removed 👍

asarazor:
Looks great, @sawyerh 

I took the liberty of playing with this: I stripped down the visual presentation even further and beefed up the white-space, just as an experiment. The intent is to minimize distractions as much as possible so the content is the focus. Much of this may contravene other conventions in the design system (e.g. key-line above action buttons, blue for _all_ links?), so take it as just another viewpoint. 

image

Thanks for keeping us in the loop, and looking forward to your React work on this.

sawyerh:
@asarazor Was literally about to ping you in Slack to get your thoughts on this 😄 

Yeah, I've been debating the whitespace and top border on the actions. I think your approach is good here, and I can make these changes before merging this in 👍 The buttons at the bottom are up to the implementor, so those can be any style you choose. For the close button, I'd prefer keeping it blue for now to stay consistent with the link/button styles, but happy to discuss that more if you feel strongly.

sawyerh:
Merged! Will continue iterating in the React branch: https://github.com/cmsgov/design-system/tree/sawyerh/modal%23193

PR #201, opened 2017-12-06
Feature/autocomplete field

1Copenut:
### WIP Added
Built a basic autocomplete using Downshift. The tests are not ready at all, and the documentation is copy/pasted from another component. I'm interested in getting your feedback on the basic component and any insights into how Downshift works that I might be overlooking. Tests and docs will be cleaned up in the next few days.

### Improved 12/19
* Added handling for `Loading...` and `No results` states
* Removed the filtering function because the component will largely deal with async data calls
* Abstracted the list item creation out to its own function for better control
* Added a `loading` Boolean prop for custom messaging when data has not come back

### Downshift Resources
* https://github.com/paypal/downshift
* https://blog.kentcdodds.com/introducing-downshift-for-react-b1de3fca0817


1Copenut:
This branch is ready for full PR when reviewers have time.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,175 @@
...
+        onStateChange={onStateChange}
+        render={({
+          clearSelection,
+          getInputProps,
+          getItemProps,
+          highlightedIndex,
+          isOpen,
+          inputValue
+        }) => (
+          <div className="ds-u-display--flex ds-u-flex-wrap--wrap ds-c-autocomplete">
```
sawyerh: Were these flexbox classes leftovers from a previous implementation of this? Doesn't seem to be anything currently in here that requires this (unless I'm missing something).

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,175 @@
...
+            <input
+              {...getInputProps({
+                className: 'ds-c-field ds-c-autocomplete__input',
+                disabled: disabled,
+                id: this.id,
+                ref: fieldRef,
+                ...fieldProps
+              })}
+            />
+
```
sawyerh: Would it be better if this component accepted a `TextField` component as a child instead? If it's possible, I personally favor reusing an existing component, and then we can remove all the label and input props from `Autocomplete`, and someone could potentially later add the autocomplete functionality to an existing form field by wrapping it with the `Autocomplete` component:

```jsx
<Autocomplete items=["A", "B", "C"]>
  <TextField defaultValue="Foo" label ="Enter a zipcode" hint="For example: 12345" />
</Autocomplete>
```

You can check out `Tabs.jsx` to see an example of how you might manipulate the `children` prop.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,175 @@
...
+AutocompleteField.defaultProps = {
+  clearAriaLabel: 'Clear typeahead and search again',
+  clearInputText: 'Search again',
+  itemToString: noop,
+  labelHint:
+    'This is an autocomplete field. Begin typing to search for relevant information. The number of results will be updated as you type.',
+  labelText: 'Generic autocomplete label',
+  onChange: noop,
+  onStateChange: noop
```
sawyerh: Do we need this to be `noop` or could these be `undefined` (no default) until the user defines them?

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,108 @@
...
+  &__input[disabled] {
+    + .ds-c-autocomplete__button {
+      background: transparent;
+      border: transparent;
+      color: $color-gray-dark;
+      pointer-events: none;
+    }
+  }
+
+  &__list {
```
sawyerh: Should this list be absolute positioned so it doesn't push other page content down as the user is typing? 

autocomplete

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: The Flexbox classes were used to force the layout I wanted, with the clear link always to the lower right. I could try floats and clearfix on the parent div, if there's a strong feeling to move in that direction.

sawyerh: Ah ok, got it. I was just missing where it was being used.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,175 @@
...
+                        {...getItemProps({ item })}
+                      >
+                        {itemToString(item)}
+                      </li>
+                    ))}
+                </ul>
+              </div>
+            ) : null}
+
+            <Button
```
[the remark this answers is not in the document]
1Copenut: I feel a clear function is in order. Users should have a visual cue that this is not a standard input box. It would be a simple update to reset the `selectedItem` control prop on input focus, and maybe we should be doing that anyway. I still feel a secondary clear function is in order, otherwise we assume too much about users understanding this is a unique UI element.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: @pwolfert pointed me to Downshift source line 649 ( https://github.com/paypal/downshift/blob/master/src/downshift.js#L649 ) and 612, where `getInputProps` is defined. This method applies the key role, aria, and autocomplete attributes, and is what makes Downshift fully accessible. 

We need access to Downshift's `getInputProps` and those can’t be directly applied to `TextField` children unless we add a generic `inputProps` prop to `TextField` and pass them down to the `<input />`.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: I don't have strong feelings either way. I have used `noop` in the past so I can make function props like `onChange` required, test more easily. I'll remove them, and update the tests.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
sawyerh: This isn't documented anywhere (right now), but `TextField` already accepts any prop and they get passed into the `input` field: https://github.com/CMSgov/design-system/blob/d556a674939f892c8c83e77c3c822ff321b4ee07/packages/core/src/components/TextField/TextField.jsx#L68

```jsx
const textFieldChild = this.props.children; // we'd want to filter this to get <TextField>
const inputProps = getInputProps();
const field = React.cloneElement(textFieldChild, {...inputProps});
```

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
sawyerh: I think if it's required in order to work, I'd argue it's better not to have a default so that a warning is logged to the console when a developer is implementing it. Also one less dependency if we don't need `noop` 😛

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: I'll give it a try, and see what happens.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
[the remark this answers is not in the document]
sawyerh: > It would be a simple update to reset the `selectedItem` control prop on input focus, and maybe we should be doing that anyway.

Do you mean on `blur`? I agree with that. We probably shouldn't rely on the user clearing the field by clicking the "Clear" button. Some users might erase the value and proceed with the rest of the form, so we wouldn't want a hidden value getting submitted to the backend in those scenarios.

> Users should have a visual cue that this is not a standard input box.

I think the autocomplete list that gets displayed is a pretty good visual cue already that the field has additional functionality, no? 

Your comment got me thinking though: There are two ways I can think of an autocomplete list being used, and I think one of these ways requires a stronger visual cue than the other:

1. A **constrained** autocomplete input, where the user *must* select from a constrained list of options. An example of a constrained autocomplete input is one where a user is asked to enter their zip code and they're required to select their county from the autocomplete list. For this category of autocomplete, it's really important that we make it clear to the user that they need to pick an option from the list. 

    Here's one idea, where we could use a label within the autocomplete list to indicate to the user that they need to select from the list. The exact phrasing we use might need some finessing.
  image
2. An **unconstrained** autocomplete input, where the autocomplete list is there to help the user fill out the field quicker, but whatever value they enter is submitted to the backend. An example of an unconstrained autocomplete is a search field, where an autocomplete list could surface common search queries, but the field accepts any search query the user types. For this category, the autocomplete functionality is an enhancement and I think the autocomplete list itself is enough of a visual cue that the field has additional functionality.
  image

Curious to hear @cindyphan's thoughts on this 👂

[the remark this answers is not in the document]
1Copenut: My example component is definitely a constrained one, but I envision this being used in an unconstrained way, with asynchronous data calls returning 0 to N results on the fly. When users select an option from the suggested list, the input should remember and retain that choice until the user opts to change it. This could be initiated by clicking the "Clear typeahead" link, or focusing the input again.

If my option were cleared out of the input field on blur, I would feel something was broken, and try searching again.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: That actually works better, setting to undefined. I'll use that for all patterns in the future.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
1Copenut: Yes, using position: absolute is a good call. This will be included in the next PR.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: I opted to go back to a simple float, and clearfix on the containing div. Will be included in the next PR push.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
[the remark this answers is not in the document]
cindyphan: I think there is a use for both of those autocompletes, things like ZIP Code search used in WindowShop/Screeners/FLH would need a constrained one, and I think adding the label "Select which county" or some other phrasing helps to clarify what needs to happen since we are using the autocomplete as the selection point. I do think we need some kind of clear button, "Search again" or "Clear". We could do the micro-blue-outline for it. 

Then the unconstrained is useful for more general searches, like help. (This is something we don't need right away however.)

I think something we'll need to account for as well is:
1. I typed in a ZIP Code - 60647
2. Selected a county from the dropdown list - Chicago, IL 60647
3. Purposely or mistakenly deleted from letters from my selection - Chicago, IL 6064
4. Now this selection is technically invalid

This is how Find Local Help is currently approaching it, will connect with Trevor about this more:
image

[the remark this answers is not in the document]
1Copenut: I added a conditional Boolean for a constrained list header. By passing `constrainedList` to `Autocomplete`, you will see the header in the typeahead list. It has a prop for updating the header string as well.

I'll keep looking at Downshift to see if we can pass a conditional "No results found" message.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,198 @@
...
+   */
+  constrainedList: PropTypes.bool,
+  /**
+   * Text string for the `<h5>` header added to the`<div.ds-c-autocomplete__list>` when prop `constrainedList`is passed to `<Autocomplete>`
+   */
+  constrainedListText: PropTypes.string,
+  /**
+   * Passes prop `disabled` to the child `<Textfield>` component
+   */
+  isDisabled: PropTypes.bool,
```
pwolfert: Is there a technical reason this can't be called `disabled`?

pwolfert: Sorry to tack something on at the last minute. Just was looking over the new props and noticed that.

1Copenut: Don't believe so. I should be able to change that easily.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,198 @@
...
+      if (isTextField(child)) {
+        return React.cloneElement(
+          child,
+          getInputProps({
+            className: isDisabled ? 'ds-c-autocomplete__inner--disabled' : '',
+            disabled: isDisabled,
+            fieldClassName: 'ds-c-autocomplete__input',
+            id: this.id,
+            inputId: this.id,
+            labelClassName: 'ds-u-margin-top--0 ds-c-autocomplete__label'
```
sawyerh: 1. Removing the label's margin is going to cause issues with spacing when this component is used in the real world. If someone needs to remove the margin, they can do that themself through the `TextField` child's props.
2. `ds-c-autocomplete__label` doesn't appear have any styles associated with it, so we can remove it unless you have an intended usage for it?

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,198 @@
...
+  /**
+   * Adds a conditional header to `<div.ds-c-autocomplete__list>`
+   */
+  constrainedList: PropTypes.bool,
+  /**
+   * Text string for the `<h5>` header added to the`<div.ds-c-autocomplete__list>` when prop `constrainedList`is passed to `<Autocomplete>`
+   */
+  constrainedListText: PropTypes.string,
+  /**
+   * Passes prop `disabled` to the child `<Textfield>` component
```
sawyerh: This is a little confusing to me. Could we remove the need for this prop at this level, or rephrase the documentation around it (ie. "Disables autocomplete functionality and overrides the `TextField`'s disabled state")? 

I could be wrong, but I think the use case of having a disabled autocomplete field is low, and if someone wanted that, they could render `<Textfield disabled />` (without nesting it in `Autocomplete`)

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,198 @@
...
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.string,
+      name: PropTypes.name
+    })
+  ).isRequired,
+  /**
+   * https://github.com/paypal/downshift#onchange
+   *
+   * Called when the user selects an item and the selected item has changed. Called with the item that was selected and the new state of `downshift`.
```
sawyerh: I like how you're referencing Downshift's documentation. I have a small nit: Could we flip the order, so that the link comes last? For example:

> Blah blah blah description

> Also see: https://github...

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,198 @@
...
+   */
+  onChange: PropTypes.func,
+  /**
+   * https://github.com/paypal/downshift#onstatechange
+   *
+   * This function is called anytime the internal state changes. This can be useful if you're using downshift as a "controlled" component, where you manage some or all of the state (e.g. `isOpen, selectedItem, highlightedIndex`, etc) and then pass it as props, rather than letting downshift control all its state itself.
```
sawyerh: This documentation is confusing to me, since I'm not sure how one would use Downshift as a "controlled" component through the `<Autocomplete>` component. Do we need this prop?

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,198 @@
...
+                ) : null}
+
+                <ul
+                  aria-labelledby={
+                    constrainedList ? 'constrained-list-header' : null
+                  }
+                  className="ds-c-list--bare"
+                  role="listbox"
+                >
+                  {items
```
sawyerh: This component currently renders an empty list if `items` is empty. Can we display a (customizable) "No results" message? Also, how would this work if you have to query an API..is there a loading state?

image

image

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,198 @@
...
+                        item.name
+                          .toLowerCase()
+                          .includes(inputValue.toLowerCase())
+                    )
+                    .map((item, index) => (
+                      <li
+                        aria-selected={highlightedIndex === index}
+                        className={
+                          highlightedIndex === index
+                            ? 'ds-u-padding--1 ds-c-autocomplete__list-item ds-c-autocomplete__list-item--active'
```
sawyerh: We could place the padding within the `ds-c-autocomplete__list-item` CSS declaration to remove the need for a utility class here

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+@import '@cmsgov/design-system-support/src/settings/index';
+
+/*
+Autocomplete field
+
+@paypal https://github.com/paypal/downshift#itemtostring
+
+Autocomplete fields allow users to enter any combination of letters, numbers, or symbols of their choosing (unless otherwise restricted), and receive one or more suggested matches in a list below the input.
+
+Markup:
```
sawyerh: I'll rarely say this 😅  but we could remove this HTML example – I'm not sure how useful it is since there's no way to preview the autocomplete list.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
+// Form-specific variables can be found in _variables.forms
+@import '@cmsgov/design-system-support/src/settings/index';
+
+/*
+Autocomplete field
+
+@paypal https://github.com/paypal/downshift#itemtostring
```
sawyerh: This isn't going to render anything in the docs since it's not a supported "flag". We use `@uswds` for USWDS links since we do want to render those on the page.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+
+Style guide: components.autocomplete-field
+*/
+
+// autocomplete component
+.ds-c-autocomplete {
+  position: relative;
+  width: 460px;
+
+  &__inner--disabled {
```
sawyerh: I'd push us to avoid this Sass syntax since it reduces the legibility of our declarations and makes searching for class names impossible (ie. Doing a search for `ds-c-autocomplete__inner--disabled` wouldn't find this).

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+  <a aria-label="Clear typeahead and search again" class="ds-c-button ds-c-button--transparent ds-c-button--small ds-u-float--right ds-u-padding-right--0 ds-c-autocomplete__button" href="javascript:void(0);">Search again</a>
+</div>
+
+Style guide: components.autocomplete-field
+*/
+
+// autocomplete component
+.ds-c-autocomplete {
+  position: relative;
+  width: 460px;
```
sawyerh: This should be `max-width` so that it remains responsive, and we can use the `$input-max-width` variable so that it's consistent with the field's max-width

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+
+  &__inner--disabled {
+    + .ds-c-autocomplete__button {
+      background: transparent;
+      border: transparent;
```
sawyerh: The `background` and `border` properties aren't needed here since these are already set to `transparent` when using the `ds-c-button--transparent` modifier.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+      padding: 0;
+      position: absolute;
+      width: 1px;
+      word-wrap: normal;
+    }
+  }
+
+  &__list {
+    background-color: $color-white;
+    max-width: 442px;
```
sawyerh: I think this should use the `$input-max-width` value, but I'm not sure why it's smaller than the field's max-width?

But also, if we're setting its `width` to `100%`, then I don't think we need `max-width`, since the `max-width` on the parent `ds-c-autocomplete` would handle this for us.

packages/core/src/components/AutocompleteField/AutocompleteField.example.jsx
```diff
@@ -0,0 +1,109 @@
...
+      }}
+    >
+      <TextField
+        hint="This is an autocomplete field. Begin typing to search for relevant information. The number of results will be updated as you type."
+        label="Unconstrained list"
+        name="Downshift_autocomplete"
+      />
+    </AutocompleteField>
+
+    <AutocompleteField
```
sawyerh: Up to you, but I'd be cool not having a `disabled` example for the sake of simplifying this example markup.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,198 @@
...
+  clearInputText: 'Search again',
+  constrainedListText: 'Select from the options below:'
+};
+
+AutocompleteField.propTypes = {
+  children: PropTypes.node,
+  /**
+   * Screenreader-specific label for the Clear input link. Intended to provide a longer, more descriptive explanation of the link's behavior.
+   */
+  clearAriaLabel: PropTypes.string,
```
sawyerh: This goes back to my original comment about "Search again" being confusing/deceiving. If the button's text said "Clear", would we need an ARIA label?

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+                      </li>
+                    ))}
+                </ul>
+              </div>
+            ) : null}
+
+            <Button
+              aria-label={clearAriaLabel}
+              className="ds-u-float--right ds-u-padding-right--0 ds-c-autocomplete__button"
+              href="javascript:void(0);"
```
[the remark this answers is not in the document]
1Copenut: Yes, forcing it to render an <a> tag. It was either a link, or a lot of overriding to get the button to not have a background when disabled. Will change as needed.

package.json
```diff
@@ -40,6 +40,7 @@
     "cssnano": "^3.10.0",
     "cssstats": "^3.1.0",
     "del": "^3.0.0",
+    "downshift": "^1.22.1",
```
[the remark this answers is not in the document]
1Copenut: Done. Moved to the /packages/core/package.json file.

package.json
```diff
@@ -72,6 +73,7 @@
     "mkdir-recursive": "^0.3.0",
     "mz": "^2.7.0",
     "node-notifier": "^5.1.2",
+    "nooop": "^1.0.0",
```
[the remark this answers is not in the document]
1Copenut: Done, and removed from package.json.

packages/core/src/components/AutocompleteField/AutocompleteField.example.jsx
1Copenut: Done. Removed the disabled example because we might not need a `disabled` prop after all.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+ * Determine if a React component is a TextField
+ * @param {React.Node} child - a React component
+ * @return {Boolean} Is this a TextField component?
+ */
+function isTextField(child) {
+  return child != null && child.type === TextField;
+}
+
+/**
+ * An `AutocompleteField` component renders a typeahead input field, form label, and a clear input link.
```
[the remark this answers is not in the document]
1Copenut: Done. Updated the descriptive text.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+ */
+export class AutocompleteField extends React.PureComponent {
```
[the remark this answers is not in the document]
1Copenut: Done. All files renamed and not breaking on build. Will re-run and modify tests before committing.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+        }) => (
+          <div className="ds-u-clearfix ds-c-autocomplete">
+            {this.renderChildren(getInputProps)}
+
+            {isOpen ? (
+              <div className="ds-u-border--1 ds-u-padding--1 ds-c-autocomplete__list">
+                {constrainedList ? (
+                  <h5
+                    className="ds-u-margin--0 ds-u-padding--1"
+                    id="constrained-list-header"
```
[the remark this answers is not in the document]
1Copenut: Done. Using unique_id and matching up the `ul[aria-labelledby]` to use that same class attribute.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+AutocompleteField.propTypes = {
+  children: PropTypes.node,
+  /**
+   * Screenreader-specific label for the Clear input link. Intended to provide a longer, more descriptive explanation of the link's behavior.
+   */
+  clearAriaLabel: PropTypes.string,
+  /**
+   * Clear link text that will appear on the page as part of the rendered component
+   */
+  clearInputText: PropTypes.string,
```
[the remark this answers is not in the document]
1Copenut: Done. Changed to PropTypes.node.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Adds a conditional header to `<div.ds-c-autocomplete__list>`
+   */
+  constrainedList: PropTypes.bool,
+  /**
+   * Text string for the `<h5>` header added to the`<div.ds-c-autocomplete__list>` when prop `constrainedList`is passed to `<Autocomplete>`
+   */
+  constrainedListText: PropTypes.string,
```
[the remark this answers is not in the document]
1Copenut: Done. Used the `label` markup suggested in Slack conversation, and PropType set to `node`.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
1Copenut: Done. Removed from docs.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: Done. Added the padding to `.autocomplete__list-item` using `$spacer-2` variable.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: I'm opting to leave the `aria-label` as a secondary item. If a screen reader settles on the link and says "Clear, link" I immediately think "Clear what?" It's not enough context by itself in my opinion. I'll change the default visual text to clear, to make it shorter.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+                        {itemToString(item)}
+                      </li>
+                    ))}
+                </ul>
+              </div>
+            ) : null}
+
+            <Button
+              aria-label={clearAriaLabel}
+              className="ds-u-float--right ds-u-padding-right--0 ds-c-autocomplete__button"
```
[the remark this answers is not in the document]
1Copenut: Done. Removed the `disabled` prop and updated the CSS to remove the extra class.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: Done. Removed the `disabled` prop. If we hear from users that having a way to disable the Clear search link is needed, we can refactor at that time.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: Since we have the TextField as a child now, and can manage our API calls using its `onChange` prop, I think we can remove this.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
1Copenut: Done. Removed the HTML example, because yeah, we won't be able to show the typeahead without React.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+  /**
+   * Screenreader-specific label for the Clear input link. Intended to provide a longer, more descriptive explanation of the link's behavior.
+   */
+  clearAriaLabel: PropTypes.string,
+  /**
+   * Clear link text that will appear on the page as part of the rendered component
+   */
+  clearInputText: PropTypes.string,
+  /**
+   * Adds a conditional header to `<div.ds-c-autocomplete__list>`
```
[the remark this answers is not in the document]
1Copenut: Done. Updated the documentation to be more general, using your suggested text above.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
1Copenut: Done. Changed the order and added Also see... text

packages/core/src/components/AutocompleteField/AutocompleteField.scss
1Copenut: Done. Removed the BEM shorthand from SCSS file and added fully qualified class names.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
1Copenut: Done. Removed this section entirely because we're not disabling the `Autocomplete` component.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+    + .ds-c-autocomplete__button {
+      background: transparent;
+      border: transparent;
+      color: $color-gray-dark;
+      pointer-events: none;
+    }
+  }
+  
+  &__label {
+    > .ds-c-field__hint {
```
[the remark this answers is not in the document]
1Copenut: It felt like context information that might be nice for screen readers, but not needed for sighted users. I didn't want to depend solely on ARIA markup for assistive devices. I can make the hint shown for all users, though designs call for no hint shown. It was a judgment call; either way works for me.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+## When to consider alternatives
+
+- When users are choosing from a specific set of options. Consider [checkboxes, radio buttons]({{root}}/components/choice), or a [select menu]({{root}}/components/select) in these cases.
+
+## Guidance
+
+- `<Autocomplete>` makes use of the `<Downshift>` component from Paypal: https://github.com/paypal/downshift
+- Don't use placeholder text in autocomplete fields. Try to write a descriptive label that identifies what the user is searching for. People who have cognitive or visual disabilities have additional problems with placeholder text.
+- The length of the text field provides a hint to users as to how much text to write. Do not require users to write paragraphs of text into a single-line `input` box; use a `textarea` instead.
+- Custom props can be added to the `<input>` by passing them to the `<Autocomplete>` component
```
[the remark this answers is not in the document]
1Copenut: Done. Removed that line so users won't be trying to pass custom props directly to `Autocomplete`.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+- Group each set of thematically related controls in a `fieldset` element. Use the `legend` element to offer a label within each `fieldset`. The `fieldset` and `legend` elements make it easier for screen reader users to navigate the form.
+- Keep your form blocks in a vertical pattern. This approach is ideal, from an accessibility standpoint, because of limited vision that makes it hard to scan from left to right.
+
+## Learn more
+
+- [Form Guidelines]({{root}}/guidelines/forms/)
+- ["Placeholders in Form Fields Are Harmful"](https://www.nngroup.com/articles/form-design-placeholders/)
+- [Asking for a date of birth](https://designnotes.blog.gov.uk/2013/12/05/asking-for-a-date-of-birth/)
+- [GOV.UK text boxes discussion](https://paper.dropbox.com/doc/Text-boxes-8NLlgz9tjR8OzTz2N9wE3)
+- [Four steps for choosing form elements on the Web (PDF)](http://www.formsthatwork.com/files/Articles/dropdown.pdf)
```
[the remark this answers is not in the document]
1Copenut: Done. Removed them all since they were indeed leftovers.

packages/core/src/components/TextField/TextField.jsx
```diff
@@ -61,7 +62,7 @@ export class TextField extends React.PureComponent {
         </FormLabel>
         <FieldComponent
           className={fieldClasses}
-          id={this.id}
+          id={inputId || this.id}
```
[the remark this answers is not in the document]
1Copenut: Agreed. Changed that to the `constructor`. Much cleaner.

packages/core/src/components/TextField/TextField.jsx
```diff
@@ -100,6 +101,11 @@ TextField.propTypes = {
    * Additional hint text to display
    */
   hint: PropTypes.node,
+  /**
+   * Optional String ID, useful if you are trying to pass an ID into
+   * `<FormLabel>` and `<FieldComponent>` from a parent like `<Autocomplete>`
+   */
+  inputId: PropTypes.string,
```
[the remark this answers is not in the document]
1Copenut: Done. Made the change to documentation and in `Autocomplete` cloned props.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+
+## Accessibility
+
+- The `<Autocomplete>` component has taken special care to ensure accessibility for screenreader devices. It announces the number of results based on `items` matches with the `inputValue` string. The component also reads out the name of each list item when users arrow up or down.
+- `<Autocomplete>` has visually hidden hint text nested in the `<label>` element. This hint `<span>` is being used to provide contextual instructions for the typeahead, and can be overriden by passing a custom `String` into the `labelHint` prop.
+- `<Autocomplete>` has a link to clear the search, and refocus the `<input>` element. This resets the `selectedItem` prop to `null`, and will re-read the label and screenreader hint text.
+- Group each set of thematically related controls in a `fieldset` element. Use the `legend` element to offer a label within each `fieldset`. The `fieldset` and `legend` elements make it easier for screen reader users to navigate the form.
+- Keep your form blocks in a vertical pattern. This approach is ideal, from an accessibility standpoint, because of limited vision that makes it hard to scan from left to right.
```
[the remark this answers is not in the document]
1Copenut: Done. Removed from the documentation.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
1Copenut: Done. Changed to max-width and added `$input-max-width` variable.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
1Copenut: Done. Added a `box-sizing: border-box;` declaration to force the list container to honor the right width and not be wider than the input.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
```diff
@@ -0,0 +1,117 @@
...
+- Don't use placeholder text in autocomplete fields. Try to write a descriptive label that identifies what the user is searching for. People who have cognitive or visual disabilities have additional problems with placeholder text.
+- The length of the text field provides a hint to users as to how much text to write. Do not require users to write paragraphs of text into a single-line `input` box; use a `textarea` instead.
+- Custom props can be added to the `<input>` by passing them to the `<Autocomplete>` component
+
+**[View the "Forms" guidelines for additional guidance and best practices.]({{root}}/guidelines/forms/)**
+
+## Accessibility
+
+- The `<Autocomplete>` component has taken special care to ensure accessibility for screenreader devices. It announces the number of results based on `items` matches with the `inputValue` string. The component also reads out the name of each list item when users arrow up or down.
+- `<Autocomplete>` has visually hidden hint text nested in the `<label>` element. This hint `<span>` is being used to provide contextual instructions for the typeahead, and can be overriden by passing a custom `String` into the `labelHint` prop.
```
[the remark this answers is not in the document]
1Copenut: On second thought, I'll leave the hint shown, and let teams add their own domain-specific screen reader classes to hide it as needed.

1Copenut:
@sawyerh All changes are complete and ready for review. Thank you.

packages/core/src/components/AutocompleteField/AutocompleteField.jsx
```diff
@@ -0,0 +1,201 @@
...
+    return React.Children.map(this.props.children, child => {
+      const { isDisabled } = this.props;
+
+      if (isTextField(child)) {
+        return React.cloneElement(
+          child,
+          getInputProps({
+            className: isDisabled ? 'ds-c-autocomplete__inner--disabled' : '',
+            disabled: isDisabled,
+            fieldClassName: 'ds-c-autocomplete__input',
```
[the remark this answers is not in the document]
sawyerh: Just flagging this again. @1Copenut do we need this class name for anything?

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+          getInputProps,
+          getItemProps,
+          highlightedIndex,
+          isOpen,
+          inputValue
+        }) => (
+          <div className="ds-u-clearfix ds-c-autocomplete">
+            {this.renderChildren(getInputProps)}
+
+            {isOpen ? (
```
sawyerh: We should handle the case where there are no results. 

From what I'm seeing elsewhere, other implementations hide the autocomplete list container when there aren't any results, so that could be one way we handle this scenario.

A few examples:

- http://ianmcburnie.github.io/mindpatterns/input/autocomplete/
- https://ant.design/components/auto-complete/

Alternatively, if we want to provide feedback to the user that there are no results, we could provide a message in the autocomplete container. This could be an optional prop, which renders the message when present and causes the container to not render when it's not set. 

For example:

```jsx
// This would render a message in the dropdown when there are no results:
<Autocomplete emptyText="No results">...</Autocomplete>

// This would result in no dropdown when there are no results:
<Autocomplete>...</Autocomplete>
```

Here's what things currently look like when the list is empty, which I believe would be confusing to users:

image

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+                  </h5>
+                )}
+
+                <ul
+                  aria-labelledby={label ? this.labelId : null}
+                  className="ds-c-list--bare"
+                  role="listbox"
+                >
+                  {items
+                    .filter(
```
sawyerh: How would this work if you have to query an API..is there a loading state? 

@pwolfert Does this implementation work for how y'all are thinking about the zipcode + county picker? I assume you would be doing the filtering on the backend and only returning the counties that match?

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+                      </li>
+                    ))}
+                </ul>
+              </div>
+            ) : null}
+
+            <Button
+              aria-label={clearAriaLabel}
+              className="ds-u-float--right ds-u-padding-right--0"
+              href="javascript:void(0);"
```
sawyerh: @1Copenut In a previous comment you said:

>  forcing it to render an `<a>` tag. It was either a link, or a lot of overriding to get the button to not have a background when disabled. Will change as needed.

Now that we're no longer passing in the disabled state into `<Autocomplete>`, I'd push for us to remove the `href` and use the more semantically correct `<button>` tag for this action. 

(I'd argue that anywhere you're using `javascript:void(0)` as the `href`, then the element should be a `button`).

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+   * Also see: https://github.com/paypal/downshift#itemtostring
+   */
+  itemToString: PropTypes.func,
+  /**
+   * Array of objects used to populate the suggestion list that appears below the input as users type. This array of objects is intended for an async data callback, and should conform to the prescribed shape to avoid errors.
+   */
+  items: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.string,
+      name: PropTypes.name
```
sawyerh: This isn't a valid `PropType`, I think you meant `string`? Not sure why React isn't throwing an error about this though.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+    })
+  ).isRequired,
+  /**
+   * Adds a heading to the top of the autocomplete list. This is used to convey to the user that they're required to select an option from the autocomplete list.
```
sawyerh: ✍️  Content suggestion:

Change "This **is** used..." to "This **can be** used", since this is just one way the `label` prop can be used now.

packages/core/src/components/AutocompleteField/AutocompleteField.scss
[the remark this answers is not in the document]
sawyerh: @1Copenut The hint text element is something available on any text field, and can be used by apps to provide additional help content related to the field, so we don't want to hide it. If you're considering adding additional text here for assistive devices, we should place that somewhere other than in `ds-c-field__hint`.

Examples of how hint text is used on some fields:

image


image

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
```diff
@@ -0,0 +1,64 @@
...
+        {
+          id: 'mZfKcGf9',
+          name: 'Cook County, TN'
+        }
+      ]}
+      onChange={selectedItem => console.log(selectedItem)}
+    >
+      <TextField
+        hint="This is an autocomplete field. Begin typing to search for relevant information. The number of results will be updated as you type."
+        label="Constrained list"
```
sawyerh: We don't explain anywhere what "Constrained" or "Unconstrained" means, so we could rephrase these field labels to something maybe more clear, like: "Labeled list" and "Simple list".

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
```diff
@@ -0,0 +1,64 @@
...
+        },
+        {
+          id: 'mZfKcGf9',
+          name: 'Cook County, TN'
+        }
+      ]}
+      onChange={selectedItem => console.log(selectedItem)}
+    >
+      <TextField
+        hint="This is an autocomplete field. Begin typing to search for relevant information. The number of results will be updated as you type."
```
sawyerh: Could we provide a hint to help the user understand what they can type in order to see autocomplete results? Right now you have to look at the code to understand that you should type `Co..`.  

✍️  Hint suggestion:

> Type `c` then use ARROW keys to change options, ENTER key to make a selection, ESC to dismiss.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+  clearAriaLabel: 'Clear typeahead and search again',
+  clearInputText: 'Clear search'
+};
+
+Autocomplete.propTypes = {
+  children: PropTypes.node,
+  /**
+   * Screenreader-specific label for the Clear input link. Intended to provide a longer, more descriptive explanation of the link's behavior.
+   */
+  clearAriaLabel: PropTypes.string,
```
sawyerh: A convention we've been following for `aria` prop names is to prefix it with `aria`. So, this would be `ariaClearLabel`.

`VerticalNavItem` and `VerticalNavItemLabel` are two components now where we follow this naming convention, and `Dialog` will be another one.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+
+  renderChildren(getInputProps) {
+    // Extend props on the TextField, by passing them through
+    // Downshift's `getInputProps` method
+    return React.Children.map(this.props.children, child => {
+      if (isTextField(child)) {
+        return React.cloneElement(
+          child,
+          getInputProps({
+            fieldClassName: 'ds-c-autocomplete__input',
```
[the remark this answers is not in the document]
1Copenut: Done. Removed the fieldClassName.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+            id: this.id,
+            labelClassName: 'ds-u-margin-top--0 ds-c-autocomplete__label'
```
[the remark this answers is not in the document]
1Copenut: Done, removed the margin and `ds-c-autocomplete__label` class.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Done. Converted to a button. This is better semantics for sure.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: @sawyerh The React docs list `PropTypes.shape` as a valid type: https://reactjs.org/docs/typechecking-with-proptypes.html

I felt it could be a good way to get back an object with the necessary attributes. If there's a better approach, I'll switch this up.

```
// An object taking on a particular shape
  optionalObjectWithShape: PropTypes.shape({
    color: PropTypes.string,
    fontSize: PropTypes.number
  })
```

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Done. Changed to can be used... Reads better now.

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
1Copenut: Done. Changed the example hints to suggested text.

packages/core/src/components/Autocomplete/Autocomplete.jsx
sawyerh: @1Copenut Sorry, to clarify: I'm referring to `PropTypes.name` being invalid. Did you mean to have `.string` rather than `.name`?

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
1Copenut: Done. Labels changed to suggested "Labeled" and "Simple" lists.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: @sawyerh Yes, that was meant to be `.string`. Corrected locally, will be part of the next code push.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Done. Changed the label to match convention.

packages/core/src/components/Autocomplete/Autocomplete.jsx
pwolfert: For our case the filtering would be unnecessary, yeah. We'd only be giving items to the component that we want it to render.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Added handling for Loading... and No results.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Removed the filtering, per @pwolfert comment to it not being needed.

1Copenut:
@pwolfert @sawyerh Latest code is added, and ready for PR. This encompasses the handling for Loading... and No Results, plus adds a few tests against the new `loading` prop.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,196 @@
...
+  filterItems(
+    items,
+    inputValue,
+    getInputProps,
+    getItemProps,
+    highlightedIndex
+  ) {
+    if (this.props.loading) {
+      return (
+        <li className="ds-c-autocomplete__list-item--message">Loading...</li>
```
sawyerh: We'll need to localize this string, so will need a prop (`PropTypes.node`) for defining it (but this could be it's default).

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,196 @@
...
+          {...getItemProps({ item })}
+        >
+          {this.props.itemToString(item)}
+        </li>
+      ));
+    }
+
+    return (
+      <li className="ds-c-autocomplete__list-item--message">
+        No results found
```
sawyerh: Same here. We'll need to localize this string, so will need a prop (`PropTypes.node`) for defining it (but this could be it's default).

packages/core/src/components/Autocomplete/Autocomplete.scss
```diff
@@ -0,0 +1,78 @@
...
+  padding: $spacer-1 $spacer-2;
+}
+
+.ds-c-autocomplete__list-item--active {
+  background-color: $color-primary-alt-darkest;
+  color: $color-white;
+}
+
+.ds-c-autocomplete__list-item--message {
+  color: $color-gray-medium;
```
sawyerh: Can use `$color-muted` here so it uses whatever the theme's muted color is. This is what we use for other things like form field hint text.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,196 @@
...
+    PropTypes.shape({
+      id: PropTypes.string,
+      name: PropTypes.string
+    })
+  ).isRequired,
+  /**
+   * Adds a heading to the top of the autocomplete list. This can be used to convey to the user that they're required to select an option from the autocomplete list.
+   */
+  label: PropTypes.node,
+  loading: PropTypes.bool,
```
sawyerh: This `loading` prop could benefit from a short description 👼

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
```diff
@@ -0,0 +1,94 @@
...
+    >
+      <TextField
+        hint="Type c then use ARROW keys to change options, ENTER key to make a selection, ESC to dismiss."
+        label="Simple list"
+        name="Downshift_autocomplete"
+      />
+    </Autocomplete>
+
+    <Autocomplete
+      getA11yStatusMessage={() => console.log('Yep, triggered')}
```
sawyerh: Is `getA11yStatusMessage` meant to be here? It doesn't appear to be documented or implemented anywhere

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,196 @@
...
+  /**
+   * Clear link text that will appear on the page as part of the rendered component
+   */
+  clearInputText: PropTypes.node,
+  /**
+   * Used to determine the string value for the selected item (which is used to compute the `inputValue`.
+   *
+   * Also see: https://github.com/paypal/downshift#itemtostring
+   */
+  itemToString: PropTypes.func,
```
sawyerh: This appears to be a required prop based on the current implementation, though could we make it's default the following:

```jsx
(item) => (item ? item.name : '')
```

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
1Copenut: Nope, that was an oversight. Was experimenting, didn't like how it was turning out. Will remove on PR.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Done. Added a `PropType.node` and default string of `Loading...`.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Done. Added a `PropType.node` and default string of `No results` to match the screen reader output.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Done. Made `itemToString` a required function, added your example for default.

packages/core/src/components/Autocomplete/Autocomplete.scss
1Copenut: Done. Switched to `$color-muted`.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Done. Added descriptions to `loading` and the two new `PropType.node` props for handling Loading and No results language.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -0,0 +1,174 @@
...
+
+/**
+ * The `Autocomplete` component is a parent component that adds autocomplete functionality to a `TextField` component.
+ */
+export class Autocomplete extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.id = uniqueId('autocomplete_');
+    this.labelId = uniqueId('constrained-list-header_');
```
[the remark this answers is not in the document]
1Copenut: Missed this the first time. Just pushed a fix for it.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -170,7 +175,7 @@ Autocomplete.propTypes = {
    *
    * Also see: https://github.com/paypal/downshift#itemtostring
    */
-  itemToString: PropTypes.func,
+  itemToString: PropTypes.func.isRequired,
```
sawyerh: Since we have a default set for this, we don't need it to be required. This would also mean we could remove `itemToString` from the example components.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Removed the `.isRequired` attribute per Slack conversatoin.

PR #202, opened 2017-12-07
Add Dialog React component

### LINKED ISSUE #193: Dialogs / Modal component
## Description

We have a need for a general purpose dialog component in our products at HHS OIG. [Here's a gov.uk example for reference](http://rural-payments-styleguide.herokuapp.com/dialogs/) of what we are considering. Does the core team have a dialog component on the roadmap?

If not, we will look at implementing our own.

sawyerh:
### Added

Added a new `Dialog` React component that internally uses `react-aria-modal` to render an accessible modal dialog with the design system's styling.

Resolves #193


sawyerh:
@asarazor

sawyerh:
There's an [open PR that adds support for stateful React examples](https://github.com/CMSgov/design-system/pull/203). Once that's merged, I can add an example for this Dialog component.

asarazor:
@sawyerh This is great!

May I suggest a little extra head-room for the actions: `.ds-c-dialog__actions { margin-top:32px }`.

Will give you this:
screen shot 2017-12-11 at 4 51 24 pm

Pinging @5arias the FE lead in our team at HHS OIG to see if he has any thoughts on the React side of things.

pwolfert:
I like that it doesn't try to maintain its own state. That part is just how I would have done it 👍

sawyerh:
Just cleaned up this PR and merged in the documentation updates. This is ready for final review. cc @pwolfert @1Copenut @hannaliebl.

@asarazor Thanks for the feedback! I've bumped up the top margin of the actions to be more consistent with the spacing between the header and body.

packages/core/src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,107 @@
...
+    <AriaModal
+      dialogClass="ds-c-dialog"
+      escapeExits={!escapeExitDisabled}
+      includeDefaultStyles={false}
+      onExit={onExit}
+      titleId="dialog-title"
+      underlayClass="ds-c-dialog-wrap"
+      {...modalProps}
+    >
+      <header className="ds-c-dialog__header">
```
1Copenut: This needs a `div` wrapping the header, children, and footer buttons. The div should have attribute `role="document"` on it, otherwise NVDA will not let its users navigate by up and down arrow keypress. The `Dialog` will have its normal role as defined by the react-aria-dialog, but we are adding a secondary role to force NVDA into the right browsing mode.

packages/core/src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,107 @@
...
+          </h1>
+        )}
+        <button
+          className="ds-c-button ds-c-button--transparent ds-c-dialog__close"
+          onClick={onExit}
+        >
+          Close
+        </button>
+      </header>
+      {children}
```
1Copenut: I'd like to consider wrapping the `{children}` declaration in a `<main>` component. This gives another important landmark for screen readers, and follows suit with the `<header>` you're already using. Let's add role="main" as well, for redundant compatibility.

packages/core/src/components/Dialog/Dialog.jsx
1Copenut: Let's add a `role="banner"` to this header element, for backward compatibility.

packages/core/src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,107 @@
...
+      {actions && <div className="ds-c-dialog__actions">{actions}</div>}
```
1Copenut: I think these action buttons could be wrapped in a `<footer role="contentinfo"` tag, so the modal has 3 distinct landmark regions.

packages/core/src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,107 @@
...
+      {...modalProps}
+    >
+      <header className="ds-c-dialog__header">
+        {title && (
+          <h1 className="ds-h2" id="dialog-title">
+            {title}
+          </h1>
+        )}
+        <button
+          className="ds-c-button ds-c-button--transparent ds-c-dialog__close"
```
1Copenut: What do you think about adding an optional `aria-label` so we can provide a richer label to screen readers for the action buttons in the modal. I'd advise an aria-label for at least the Close button, and the Cancel button, because those could effectively say the same thing with one prop, unless they are intended to do different things.

I feel the extra context is good for assistive devices because if the amount of information being read is long, or time has lapsed since the user opened the modal, they may not remember what they are closing.

1Copenut:
I just noticed the Close text was darker blue when it had keyboard `:focus` than the X beside it.

screen shot 2017-12-15 at 12 34 40 pm

packages/core/src/components/Dialog/Dialog.jsx
sawyerh: @1Copenut Is it better to add `role="banner"` or change this element to a normal `div`?

packages/core/src/components/Dialog/Dialog.jsx
sawyerh: Good idea 👍

packages/core/src/components/Dialog/Dialog.jsx
sawyerh: Should this be `contentinfo` or `complementary`? These would be primary actions related to the `main` content. For example, `main` might include form fields and the actions might include a submit button.

packages/core/src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,120 @@
...
+            aria-label={ariaCloseLabel}
+            className="ds-c-button ds-c-button--transparent ds-c-dialog__close"
+            onClick={onExit}
+          >
+            Close
+          </button>
+        </header>
+        <main role="main">{children}</main>
+        {actions && (
+          <footer className="ds-c-dialog__actions" role="contentinfo">
```
sawyerh: > I think these action buttons could be wrapped in a `<footer role="contentinfo"` tag, so the modal has 3 distinct landmark regions.

@1Copenut  Should this be `contentinfo` or `complementary`? These would be primary actions related to the `main` content. For example, `main` might include form fields and the actions might include a submit button.

packages/core/src/components/Dialog/Dialog.jsx
```diff
@@ -0,0 +1,120 @@
...
+      dialogClass="ds-c-dialog"
+      escapeExits={!escapeExitDisabled}
+      includeDefaultStyles={false}
+      onExit={onExit}
+      titleId="dialog-title"
+      underlayClass="ds-c-dialog-wrap"
+      {...modalProps}
+    >
+      <div role="document">
+        <header className="ds-c-dialog__header" role="banner">
```
sawyerh: > Let's add a `role="banner"` to this header element, for backward compatibility.

@1Copenut Is it better to add `role="banner"` or change this element to a normal `div`?

sawyerh:
@1Copenut Thanks for the feedback! I implemented your suggestions, but have a few questions (see above).

packages/core/src/components/Dialog/Dialog.jsx
1Copenut: You raise a good point. `aside role="complementary"` feels better. After a bit more digging, this Treehouse article supports your thinking: http://blog.teamtreehouse.com/use-html5-sectioning-elements

```
The <footer> element is used to represent information about a section such as the author, copyright information, or links to related web pages.
```

Let's change it to an `aside` with role `complementary`. Good call.

packages/core/src/components/Dialog/Dialog.jsx
1Copenut: Let's leave the `role="banner"` because it's what's needed for backward compatibility. NVDA and VoiceOver announce the landmarks, and VoiceOver allows users to navigate inside the modal via landmarks, so having these are a nice feature.

1Copenut:
@sawyerh I answered your two questions inline, and did a quick analysis to confirm the `ESC` key was dismissing the modal dialog. https://github.com/gdkraus/accessible-modal-dialog offers a nice checklist of things I use as guidelines for accessible modals. Your modal checks these boxes and will serve teams well.

PR #203, opened 2017-12-08
Docs: Render React examples within iframes

sawyerh:
### Internal

This is something I've been wanting to do for a looooong time, but haven't had a reason for until now. Rendering these React examples in an iframe:

1. Allows us to separate the examples away from any custom code needed for the documentation. This helps ensure the previews we're showing are actually accurate.
2. Provides support for examples showing stateful React components, like the new `<Dialog>` component.
3. Now means all HTML and React examples have their own URL, which will make visual regression and accessibility testing easier when we decide to tackle that.
4. Actually means the code is cleaner and lighter, now that we're not bundling every single example in one file 🙈 

#### Testing

To help me test this, please run the documentation locally and poke around. I've already tested this pretty extensively, but I'd like to confirm the examples and code snippets show correctly for someone else.

#### What has changed

The most relevant change for contributors is: React examples no longer export a function that returns a presentational component. Instead, the example files should now use `ReactDOM.render` and render to `#js-example`.  Everything you put in this example file, besides the `import` statements, will now show in the markup example in the documentation:

> image


tools/gulp/docs/parseReactFile.js
```diff
@@ -0,0 +1,89 @@
...
+    file.path = gutil.replaceExtension(file.path, '.json');
+    return cb(null, file);
+  });
+};
+
+/**
+ * For an example file, we only need a reference to its uncompiled source code
+ * @param {Object} file
+ */
+function parseExample(file) {
```
sawyerh: ℹ️ This method returns the source code that's shown in the "Code snippet" section:

image

tools/gulp/docs/index.js
```diff
@@ -82,12 +92,6 @@ module.exports = (gulp, shared) => {
    * @return {Promise<Array>}
    */
   function generateDocPages(pages) {
-    // It's important to require the generatePage method here since the
-    // react-doc.json data file is a dependency. This require() would fail if
-    // it was called before the file existed.
-    // TODO(sawyer): Would it be better if we passed in the relevant React
-    // documentation as a prop, rather than pulling it from the JSON file?
```
sawyerh: ℹ️ Answer: yes, this is way cleaner and removes a lot of hacky stuff we were doing. A React component's documentation is now passed in as props (`reactComponentDocs`, `reactExampleSource`, etc). The biggest benefit here is it relates a lot of code bloat.

tools/gulp/docs/generateReactExample.js
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Creates an HTML page using the React component and no additional UI.
+ * This can then be viewed in a browser, or rendered in an iFrame in
+ * the documentation.
+ * @param {Object} page
+ * @param {String} rootPath - Root docs site path
+ * @return {Promise}
+ */
+function generateReactExample(page, rootPath) {
```
sawyerh: ℹ️ We use Webpack to bundle and compile each React example, then place the output as an inline `script` at the bottom of the generated HTML page for that example.

sawyerh:
🙏

PR #209, opened 2017-12-18
Primary button border color from vars

pwolfert:
### Fixed
- Primary button border color is also determined by the `$button-primary-*` variables now


packages/core/src/components/Button/Button.scss
```diff
@@ -61,6 +61,7 @@ Style guide: components.button
 
 .ds-c-button--primary {
   background-color: $button-primary-bg;
+  border-color: $button-primary-bg;
```
sawyerh: This would need set for the various states as well (`:focus`, `:hover`). On second glance at this though, can we remove the border which is being inherited from the default `ds-c-button` styling? Then we wouldn't need to worry about the border colors. If we go that route, I think we'll just want to add a comment explaining the logic behind it, ie:

```
// Remove the default border to support easier theming
```

packages/core/src/components/Button/Button.scss
```diff
@@ -61,6 +61,7 @@ Style guide: components.button
 
 .ds-c-button--primary {
   background-color: $button-primary-bg;
+  border: none; // Remove the default border to support easier theming
```
sawyerh: I think the linter wants `0` for this instead of `none`, but otherwise LGTM

PR #210, opened 2017-12-21
[AUTOCOMPLETE] Hotfix to add onInputValueChange prop to component

1Copenut:
### Added
* `onInputValueChange` prop was added to `Autocomplete` for gathering child input value
* documentation for new method
* `listboxId` class attribute
* `aria-controls=this.listboxId` to the `getInputProps` method for building out TextField
* child cloned `onBlur, onChange, onKeyDown` events
* undocumented Autocomplete props can be passed to Downshift

### Removed
* hard-coded `aria-controls` attribute in examples

### DIFF
--- a/packages/core/src/components/Autocomplete/Autocomplete.example.jsx
+++ b/packages/core/src/components/Autocomplete/Autocomplete.example.jsx
@@ -42,8 +42,10 @@ ReactDOM.render(
       ]}
       label="Select from the options below:"
       onChange={selectedItem => console.log(selectedItem)}
+      onInputValueChange={inputVal => console.log('[INPUT_VALUE]: ' + inputVal)}
     >
       <TextField
+        aria-controls="owned_listbox"
         hint="Type c then use ARROW keys to change options, ENTER key to make a selection, ESC to dismiss."
         label="Labeled list"
         name="Downshift_autocomplete"
@@ -68,6 +70,7 @@ ReactDOM.render(
       onChange={selectedItem => console.log(selectedItem)}
     >
       <TextField
+        aria-controls="owned_listbox"
         hint="Type c then use ARROW keys to change options, ENTER key to make a selection, ESC to dismiss."
         label="Simple list"
         name="Downshift_autocomplete"
@@ -81,6 +84,7 @@ ReactDOM.render(
       onChange={selectedItem => console.log(selectedItem)}
     >
       <TextField
+        aria-controls="owned_listbox"
         hint="List should return string Loading to simulate async data call."
         label="Loading message"
         name="Downshift_autocomplete"
@@ -89,6 +93,7 @@ ReactDOM.render(
 
     <Autocomplete items={[]}>
       <TextField
+        aria-controls="owned_listbox"
         hint="List should return string No results found."
         label="No results message"
         name="Downshift_autocomplete"
--- a/packages/core/src/components/Autocomplete/Autocomplete.example.jsx
+++ b/packages/core/src/components/Autocomplete/Autocomplete.example.jsx
@@ -42,12 +42,17 @@ ReactDOM.render(
       ]}
       label="Select from the options below:"
       onChange={selectedItem => console.log(selectedItem)}
-      onInputValueChange={inputVal => console.log('[INPUT_VALUE]: ' + inputVal)}
+      onInputValueChange={inputVal =>
+        console.log('[AUTOCOMPLETE_VALUE]: ' + inputVal)
+      }
     >
       <TextField
         hint="Type c then use ARROW keys to change options, ENTER key to make a selection, ESC to dismiss."
         label="Labeled list"
         name="Downshift_autocomplete"
+        onBlur={() => console.log('Blur event')}
+        onChange={e => console.log(e)}
+        onKeyDown={() => console.log('Keydown event')}
       />
     </Autocomplete>
 
@@ -72,6 +77,9 @@ ReactDOM.render(
         hint="Type c then use ARROW keys to change options, ENTER key to make a selection, ESC to dismiss."
         label="Simple list"
         name="Downshift_autocomplete"
+        onBlur={() => console.log('Blur event')}
+        onChange={e => console.log(e)}
+        onKeyDown={() => console.log('Keydown event')}
       />
     </Autocomplete>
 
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -90,13 +92,15 @@ export class Autocomplete extends React.PureComponent {
...
-      onChange
+      onChange,
+      onInputValueChange
     } = this.props;
 
     return (
       <Downshift
         itemToString={itemToString}
         onChange={onChange}
+        onInputValueChange={onInputValueChange}
```
pwolfert: Are we also going to watch out for `onChange` props on children and apply them here if they exist?

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -90,13 +92,15 @@ export class Autocomplete extends React.PureComponent {
       itemToString,
       label,
       loading,
-      onChange
+      onChange,
+      onInputValueChange
```
pwolfert: Are we going to do the thing where we pass any remaining, undocumented props to Downshift?

1Copenut: Added `child, {...autocompleteProps}` to our Downshift prop declaration block.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -70,12 +71,15 @@ export class Autocomplete extends React.PureComponent {
     // Downshift's `getInputProps` method
     return React.Children.map(this.props.children, child => {
       if (isTextField(child)) {
-        return React.cloneElement(
-          child,
-          getInputProps({
-            id: this.id
-          })
-        );
+        const propOverrides = {
```
1Copenut: We opted to break the child props being overridden out to their own `const` for better readability. These functions are being composed into Downshift here: https://github.com/paypal/downshift/blob/master/src/downshift.js#L609

Hat tip to @pwolfert for explaining how these are working.

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
```diff
@@ -42,11 +42,17 @@ ReactDOM.render(
       ]}
       label="Select from the options below:"
       onChange={selectedItem => console.log(selectedItem)}
+      onInputValueChange={inputVal =>
+        console.log('[AUTOCOMPLETE_VALUE]: ' + inputVal)
```
pwolfert: Maybe just make these logs consistent with what kind of stuff they're printing out

1Copenut: Agreed. Standardizing on [Component]: Event Type || Value for strings, and plain objects.

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
```diff
@@ -42,11 +42,17 @@ ReactDOM.render(
...
+        console.log('[Autocomplete]: ' + inputVal)
+      }
     >
       <TextField
         hint="Type c then use ARROW keys to change options, ENTER key to make a selection, ESC to dismiss."
         label="Labeled list"
         name="Downshift_autocomplete"
+        onBlur={() => console.log('[TextField]: Blur event')}
+        onChange={e => console.log(e)}
+        onKeyDown={() => console.log('[TextField]: Keydown event')}
```
sawyerh: I think all of these placeholder event listeners are nice for debugging, but as a developer implementing the design system, I would be confused about which one I actually need to include. Can we reduce this to just one event handler that someone would need to be aware of in order to make an API call to get an updated list of items? Is it `onInputValueChange` that would be best to use?

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -90,13 +94,17 @@ export class Autocomplete extends React.PureComponent {
       itemToString,
       label,
       loading,
-      onChange
+      onChange,
+      onInputValueChange,
+      children,
+      ...autocompleteProps
```
sawyerh: Since we're passing in `autocompleteProps` now, is there a reason we need to create a `const` for 
`itemToString`, `onInputValueChange`, and `onChange`?

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -194,19 +204,25 @@ Autocomplete.propTypes = {
    */
   loading: PropTypes.bool,
   /**
-   * Message users will see when the `loading` prop is passed to `<Autcomplete />`.
+   * Message users will see when the `loading` prop is passed to `Autcomplete`.
```
sawyerh: ✍️ Small typo here: `Autcomplete`

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -194,19 +204,25 @@ Autocomplete.propTypes = {
...
    */
   loadingMessage: PropTypes.node,
   /**
    * Message users will see when the `items` array returns empty and the `loading` prop is passed to `<Autocomplete />`.
    */
   noResultsMessage: PropTypes.node,
   /**
-   * Called when the user selects an item and the selected item has changed. Called with the item that was selected and the new state of `downshift`.
+   * Called when the user selects an item and the selected item has changed. Called with the item that was selected and the new state of `Downshift`.
```
sawyerh: We don't really describe to the user what `Downshift` is, so this last sentence is likely confusing. Can we rephrase to:

> Called with the item that was selected and the new state.

packages/core/src/components/Autocomplete/Autocomplete.example.jsx
1Copenut: I'm split on how much to include, and how much to leave out. The Autocomplete `onInputValueChange` prop is probably the highest level, easiest way to make an API call. I can remove these extra event listeners from the examples, but feel they're worth leaving in the cloned component. Do you both concur?

sawyerh: yea, my comment is only directed towards the example code we put in the docs. I don't think our intent is for a developer to add all of these event handlers, so wouldn't want to advocate doing that 😛

PR #211, opened 2017-12-21
Docs: Retain console.log in React examples

sawyerh:
### Internal

- Update our UglifyJS configs so that `console.log` is preserved in the React examples since we use these for demo purposes.
- Re-enable mangling on the docs JS now that React examples are compiled using a separate process.


sawyerh:
Merging for v1.8 docs

PR #216, opened 2017-12-22
DateField adding onComponentBlur

pwolfert:
### Added
- `onComponentBlur` hook for `<DateField>` component

### DIFF
--- a/packages/core/src/components/DateField/DateField.example.jsx
+++ b/packages/core/src/components/DateField/DateField.example.jsx
@@ -2,11 +2,26 @@ import DateField from './DateField';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
-ReactDOM.render(
-  <DateField
-    monthDefaultValue={10}
-    dayDefaultValue="31"
-    yearDefaultValue="2020"
-  />,
-  document.getElementById('js-example')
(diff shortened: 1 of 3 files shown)


packages/core/src/components/DateField/DateField.example.jsx
```diff
@@ -2,21 +2,36 @@ import DateField from './DateField';
...
+  render() {
+    return (
+      <React.Fragment>
+        <DateField
+          monthDefaultValue={10}
+          dayDefaultValue="31"
+          yearDefaultValue="2020"
+          hint={this.state.hint}
+          onComponentBlur={() =>
+            this.setState({ hint: '[DateField]: Component lost focus' })
```
sawyerh: Hm, was this added for debugging purposes, or do we want to show this in the documentation? This might be a little unexpected if someone isn't sure how to interpret the code example. If we're keeping this, could we change it do a `console.log`, like we do for the `Autocomplete` example's event handlers?

pwolfert: Yes, I can backpedal there. I came up with this solution after I thought it was swallowing my console logs. (Turned out my code wasn't working.)

PR #222, opened 2018-01-11
Internal: Fix server-side rendering by defining process.env.root outside of Webpack

sawyerh:


tools/gulp/docs/generateDocPage.js
```diff
@@ -16,6 +16,11 @@ function generateDocPage(routes, page, docsPath, rootPath) {
       return '';
     }
 
+    // On the client-side the "root" variable is defined via Webpack,
+    // but we also need to define it here for "server-side" rendering
+    process.env = Object.assign(process.env, {
```
pwolfert: Can this be just `process.env.root = rootPath`?

sawyerh: Derp, yup. Fixed

PR #226, opened 2018-01-19
Adding Array.find polyfill to support IE11 fully.

1Copenut:
### Added
* Array.find polyfill for full IE11 support in cases where the method hadn't been polyfilled previously

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,8 +1,7 @@
 {
   "name": "design-system",
   "private": true,
-  "description":
-    "A shared set of design and code elements for creating accessible and consistent websites",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -49,6 +45,7 @@
     "enzyme-to-json": "^3.3.0",
     "eslint": "^4.13.1",
     "eslint-config-nava": "^2.1.0",
+    "event-source-polyfill": "^0.0.12",
```
sawyerh: This doesn't appear to be used

1Copenut: You're right. I must have overlooked removing it from the package.json file while troubleshooting on IE11. I'll create a new branch and remove it later today.

PR #228, opened 2018-01-23
Add ds-u-border--dark utility class

sawyerh:
### Added

- New `ds-u-border--dark` utility class to set the border to a dark gray.
- New `$border-color-dark` Sass variable

These are useful for situations where the background isn't white, and you need a darker neutral border color.


sawyerh:
Merging this now in order to open a PR for the hc.gov Header, but please leave any comments/feedback if you have any.

PR #230, opened 2018-01-24
Internal: Add support for theme's including design system Sass imports + Theme React files

sawyerh:
### Internal

- Fixed `postcss-image-inliner` configuration
- Added support for previewing themes which `@import` the design system's files directly within the theme's Sass file. This doesn't introduce duplicate declarations thanks to [cssnano](http://cssnano.co/optimisations/discardduplicates/).
- Added support for compiling a theme's JS files


sawyerh:
Merging this now in order to open a PR for the hc.gov Header, but please leave any comments/feedback if you have any.

PR #231, opened 2018-01-29
[HDSG-186] Updating MonthPicker to better serve assistive devices, remove aXe error

1Copenut:
### Added
* `<h4>` and `<p>` for a title structure, that will be related to by buttons, and grid of checkboxes
* Two class IDs to anchor `aria-describedby` attributes
* `aria-desribedby` attributes, to make the buttons and checkboxes relate back to title structure

### Changed
* `<FormLabel />` hint has been set to a `null` value. This is because the hint is being surfaced as part of the title structure, and referenced with `aria-describedby`. It creates a better screen reader experience, reading out the `props.label.title` on each checkbox focus, and `props.label.hint` if the user pauses on a checkbox.

### DIFF
--- a/packages/core/src/components/MonthPicker/MonthPicker.example.jsx
+++ b/packages/core/src/components/MonthPicker/MonthPicker.example.jsx
@@ -24,13 +24,6 @@ ReactDOM.render(
       label="Select the months you were in cryosleep during 2046"
       disabledMonths={[7, 8, 9, 10, 11, 12]}
     />
-    <MonthPicker
-      name="cryosleep-months"
-      label="Select the months you were in cryosleep during 2046"
-      disabledMonths={[7, 8, 9, 10, 11, 12]}
-      inversed
-      headingType="3"
-    />
   </div>,
   document.getElementById('js-example')
 );
--- a/packages/core/src/components/MonthPicker/MonthPicker.example.jsx
+++ b/packages/core/src/components/MonthPicker/MonthPicker.example.jsx
@@ -24,6 +24,13 @@ ReactDOM.render(
       label="Select the months you were in cryosleep during 2046"
       disabledMonths={[7, 8, 9, 10, 11, 12]}
     />
+    <MonthPicker
+      name="cryosleep-months"
+      label="Select the months you were in cryosleep during 2046"
+      disabledMonths={[7, 8, 9, 10, 11, 12]}
+      inversed
+      headingType="3"
+    />
   </div>,
   document.getElementById('js-example')
 );
--- a/packages/core/src/components/MonthPicker/MonthPicker.jsx
+++ b/packages/core/src/components/MonthPicker/MonthPicker.jsx
@@ -113,7 +113,7 @@ export class MonthPicker extends React.PureComponent {
         {this.months.map((month, i) => (
           <li key={month}>
             <Choice
-              aria-describedby={this.hintId}
+              aria-describedby={this.props.hint ? this.hintId : null}
               aria-label={this.monthsLong[i]}
               checked={selectedMonths.includes(i + 1)}
               className="ds-c-month-picker__month"
--- a/packages/core/src/components/MonthPicker/MonthPicker.jsx
+++ b/packages/core/src/components/MonthPicker/MonthPicker.jsx
@@ -157,7 +157,6 @@ export class MonthPicker extends React.PureComponent {
         labelClassName={classes}
         component="legend"
         errorMessage={this.props.errorMessage}
-        hint={null}
         requirementLabel={this.props.requirementLabel}
         inversed={this.props.inversed}
       >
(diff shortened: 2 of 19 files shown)


packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -168,14 +175,24 @@ export class MonthPicker extends React.PureComponent {
       this.props.className
     );
     return (
-      <fieldset className={classes}>
-        {this.renderLabel()}
+      <div className={classes}>
+        <div className="ds-c-label">
+          <h4 className="ds-u-margin--0" id={this.labelId}>
```
sawyerh: Setting this as an `h4` means it takes on the browser's default styling, like being bold by default. Is there another way to approach this to retain the label styling we use elsewhere? Could we place `ds-c-label` directly on the element instead?

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -168,14 +175,24 @@ export class MonthPicker extends React.PureComponent {
...
+            {this.props.hint}
+          </p>
+        </div>
         <div className="ds-u-margin-y--3">
           {this.renderButton(selectAllText, () => this.handleSelectAll())}
           {this.renderButton(clearAllText, () => this.handleClearAll())}
         </div>
-        <div className="ds-c-month-picker__months">{this.renderMonths()}</div>
-      </fieldset>
+        <fieldset className={classes}>
```
sawyerh: Do we need `classes` on this element, as well as the containing `div`? Could we choose one or the other, or split the classes up?

packages/core/src/components/MonthPicker/MonthPicker.test.jsx
```diff
@@ -48,7 +48,7 @@ describe('MonthPicker', () => {
     const { wrapper, props } = renderMonthPicker({
       labelClassName: 'ds-u-color--primary',
       errorMessage: 'Error!',
-      hint: 'Tips and tricks'
```
sawyerh: I think we'd still want to test that the hint gets rendered, no?

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -147,10 +153,11 @@ export class MonthPicker extends React.PureComponent {
     );
     return (
       <FormLabel
+        className="ds-u-visibility--screen-reader"
         labelClassName={classes}
         component="legend"
         errorMessage={this.props.errorMessage}
-        hint={this.props.hint}
+        hint={null}
```
sawyerh: Can we remove this prop completely or do we need it to be `null`?

packages/core/src/components/MonthPicker/MonthPicker.jsx
1Copenut: I refactored classes and kept the `<fieldset>` dynamic, and hard coded the container.

1Copenut:
@sawyerh Added a few tests to handle the title structure I created, ability to set heading level through an optional prop, and the class changes you recommended.

packages/core/src/components/MonthPicker/MonthPicker.example.jsx
```diff
@@ -24,6 +24,13 @@ ReactDOM.render(
       label="Select the months you were in cryosleep during 2046"
       disabledMonths={[7, 8, 9, 10, 11, 12]}
     />
+    <MonthPicker
+      name="cryosleep-months"
+      label="Select the months you were in cryosleep during 2046"
+      disabledMonths={[7, 8, 9, 10, 11, 12]}
+      inversed
```
sawyerh: This is looking pretty funky, do we need this example?

image

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -168,21 +167,31 @@ export class MonthPicker extends React.PureComponent {
...
-      'ds-c-month-picker',
       'ds-c-fieldset',
       'ds-u-margin-y--3',
       this.props.className
     );
     return (
-      <div className={classes}>
-        <div className="ds-c-label">
-          <h4 className="ds-u-margin--0" id={this.labelId}>
+      <div className="ds-c-month-picker ds-u-margin-y--3">
```
sawyerh: `this.props.className` would need added here too, to align with its intended behavior:

> Additional classes to be added to the root element.

Another thing to note is that this root element and the `fieldset` both have a `y` margin, but only the root element has the option for overriding that margin. So, I think we'd want the margin only applied to the root element so the user can customize it if they'd want.

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -238,6 +247,10 @@ MonthPicker.propTypes = {
    * Additional hint text to display
    */
   hint: PropTypes.node,
+  /**
+   * Heading type to override default `<h4>` in title block
+   */
+  headingType: PropTypes.string,
```
sawyerh: Nit: I think `headingLevel` and `PropTypes.number` is more accurate for the value we're expecting here.

However, do we need to provide an option for setting the heading level?

packages/core/src/components/MonthPicker/MonthPicker.test.jsx
```diff
@@ -44,17 +44,40 @@ describe('MonthPicker', () => {
     expect(onClearAll).toHaveBeenCalled();
   });
 
+  it('renders a title block with hint', () => {
+    const { wrapper } = renderMonthPicker({
+      label: 'Select a month',
+      hint: 'Tips and tricks'
+    });
+    const title = wrapper.find('h4.ds-c-label');
+    const hint = wrapper.find('p.ds-u-margin--0.ds-c-field__hint');
```
sawyerh: This is a really specific selector. Would `ds-c-field__hint` suffice?

packages/core/src/components/MonthPicker/MonthPicker.jsx
1Copenut: Yes, we need a way to set the heading level. This component could be places where peer headings are not `<h4>`'s, so having that override is important for semantic reasons. Screen reader users often navigate by skimming headings; I'd like the option to match peer headings through prop override when necessary.

1Copenut: A number makes more sense, so I'll update that.

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -161,21 +167,41 @@ export class MonthPicker extends React.PureComponent {
...
     const classes = classNames(
-      'ds-c-month-picker',
       'ds-c-fieldset',
       'ds-u-margin-y--3',
       this.props.className
     );
     return (
-      <fieldset className={classes}>
-        {this.renderLabel()}
+      <div className="ds-c-month-picker ds-u-margin-y--3">
```
sawyerh: (This comment was posted previously, but I think it got lost somewhere)

`this.props.className` would need added here too, to align with its intended behavior:

> Additional classes to be added to the root element.

Another thing to note is that this root element and the `fieldset` both have a `y` margin, but only the root element has the option for overriding that margin. So, I think we'd want the margin only applied to the root element so the user can customize it if they'd want.

packages/core/src/components/MonthPicker/MonthPicker.jsx
sawyerh: @1Copenut Thoughts on rephrasing to `headingLevel`? This is the typical vocabulary when referring to the heading options

PR #235, opened 2018-02-02
[HDSG-175] Adding aria-pressed to Select All and Clear All buttons for screen reader context

1Copenut:
### Added
* `aria-pressed` prop to `<Button />` components being used to generate Select All, Clear All
* `Boolean` state keys `selectAllPressed` and `clearAllPressed`
* logic to update `aria-pressed` attribute on select all, clear all, checkbox change
* new tests to ensure attributes are updated on click

### Tested
* Win10 JAWS + IE11
* Win10 NVDA + Firefox
* OSX High Sierra Safari + VoiceOver


packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -167,6 +191,7 @@ export class MonthPicker extends React.PureComponent {
 
   render() {
     const { selectAllText, clearAllText } = this.props;
+    const { selectAllPressed, clearAllPressed } = this.state;
```
pwolfert: I just realized that this will only set the aria labels if it's an uncontrolled component, but it can also be a controlled component. In the case of it being a controlled component, we need to derive these two booleans like you do in `handleAriaPressed`. Something like
```
if it's a controlled component:
  selectAllPressed = selectedMonths.length === 12
  clearAllPressed = selectedMonths.length === 0
else:
  get them from the state
```

pwolfert: And in fact we use it as a controlled component in taxtools

1Copenut: @pwolfert Thank you for the code snippet. Updates pushed and ready for review.

pwolfert: Perfect 👍

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -76,7 +78,19 @@ export class MonthPicker extends React.PureComponent {
...
       this.setState({ selectedMonths });
+
+      this.handleAriaPressed(selectedMonths);
+    }
+  }
+
+  handleAriaPressed(selectedMonths) {
+    if (!this.isControlled) {
+      this.setState({
+        selectAllPressed: selectedMonths.length === 12,
```
sawyerh: What happens when some months are disabled? For example, if `4` months are disabled, should `selectAllPressed` be `true` when `8` are selected? Instead of hard-coding `12` here and elsewhere, should this instead reference the number of active checkboxes?

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -167,6 +190,19 @@ export class MonthPicker extends React.PureComponent {
 
   render() {
     const { selectAllText, clearAllText } = this.props;
+    let { selectAllPressed, clearAllPressed } = this.state;
+
+    if (this.isControlled) {
```
sawyerh: Could we add a test to ensure this works as intended?

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -76,7 +78,19 @@ export class MonthPicker extends React.PureComponent {
       } else {
         selectedMonths.push(month);
       }
+
       this.setState({ selectedMonths });
+
+      this.handleAriaPressed(selectedMonths);
```
sawyerh: Nit: the `handle...` naming convention is used to indicate that a method is used as an event listener. That's not totally the case here, and I'm wondering if this would be more accurately described as `updatedAriaPressed`?

packages/core/dist/components/MonthPicker/MonthPicker.js
```diff
@@ -142,7 +143,11 @@ var MonthPicker = exports.MonthPicker = function (_React$PureComponent) {
         var selectedMonths = monthNumbers.filter(function (m) {
           return !disabledMonths.includes(m);
         });
-        this.setState({ selectedMonths: selectedMonths });
+        this.setState({
+          selectedMonths: selectedMonths,
+          selectAllPressed: true,
+          clearAllPressed: false
```
pwolfert: Forgot to revert these

packages/core/dist/components/MonthPicker/MonthPicker.js
```diff
@@ -153,7 +158,11 @@ var MonthPicker = exports.MonthPicker = function (_React$PureComponent) {
       }
 
       if (!this.isControlled) {
-        this.setState({ selectedMonths: [] });
+        this.setState({
+          selectedMonths: [],
+          selectAllPressed: false,
+          clearAllPressed: true
```
pwolfert: and these

packages/core/dist/components/MonthPicker/MonthPicker.js
```diff
@@ -90,8 +90,6 @@ var MonthPicker = exports.MonthPicker = function (_React$PureComponent) {
       // Since this isn't a controlled component, we need a way
       // to track when the value has changed.
       _this.state = {
```
sawyerh: Totally fine that you've been checking these in, but in case it's helpful: We run the build process right before the release process, so if you want, you can skip that step for PR's.

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -80,17 +78,6 @@ export class MonthPicker extends React.PureComponent {
       }
 
       this.setState({ selectedMonths });
-
-      this.handleAriaPressed(selectedMonths);
-    }
-  }
-
-  handleAriaPressed(selectedMonths) {
```
sawyerh: Nice work simplifying this 👌

packages/core/dist/components/MonthPicker/MonthPicker.js
1Copenut: I'd like to skip having these rendered files in PRs to be honest. I'll avoid adding them in the future.

packages/core/dist/components/MonthPicker/MonthPicker.js
1Copenut: Not sure why it isn't showing the latest version I committed last night. The file is correct in Sawyer's comment below: https://github.com/CMSgov/design-system/pull/235/files/1433a682fef50adfa8cc35c69dfafb8c04e048f2..9647e024ca0d412a12b45839bade5d704c08bf06#diff-6c5bf996b9de0d202168b671faa4c0da

packages/core/dist/components/MonthPicker/MonthPicker.js
1Copenut: Same as previous comment. I won't check in the rendered files anymore: https://github.com/CMSgov/design-system/pull/235/files/1433a682fef50adfa8cc35c69dfafb8c04e048f2..9647e024ca0d412a12b45839bade5d704c08bf06#diff-6c5bf996b9de0d202168b671faa4c0da

packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -90,7 +91,11 @@ export class MonthPicker extends React.PureComponent {
       const selectedMonths = monthNumbers.filter(
         m => !disabledMonths.includes(m)
       );
-      this.setState({ selectedMonths });
+      this.setState({
+        selectedMonths,
+        selectAllPressed: true,
+        clearAllPressed: false
```
sawyerh: > Forgot to revert these

@1Copenut @pwolfert Bringing the convo down to this (uncompiled) file. Are these two lines (`selectAllPressed`, `clearAllPressed`) supposed to still be here or no?

1Copenut: No, they should not. I missed those on code cleanup, and will remove them now. Thanks all.

1Copenut:
@sawyerh @pwolfert The final remnants of `selectAllPressed` and `clearAllPressed` are out of state logic. Ready for review.

PR #237, opened 2018-02-12
[HDSG-166] Review component

hannaliebl:
### Added
Added Review component.
image

My main question has to do with whether this should be in the pattern section or the component section. For now I put it into the component section. I thought initially this would go in the site package, but I think it's generic enough to be its own component here. I based the `ReviewLink` component on the `StepLink` component to handle the edit link in a generic way.

Also, I used all existing design system classes to do this - no new classes were added with any styles attached to them. I added some classes that I use in unit tests and there might be a use for them down the line, but I wasn't totally sure how to handle this in general: the combination of utility classes creates all the styles I think I need, but it might be worth encapsulating them into specific `.ds-c-review` classes, and it does make the tests a bit easier to write if I do include classes specific to the component.


packages/core/src/components/Review/ReviewLink.jsx
```diff
@@ -0,0 +1,32 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export class ReviewLink extends React.PureComponent {
+  handleClick(event) {
+    if (this.props.onClick) {
+      event.preventDefault();
+      this.props.onClick(this.props.href);
```
sawyerh: To be consistent with other component's event handlers, I think we'd want to send the `event` as the first argument here. This would be consistent with the `Button` and `Tab` component (amongst others), and would allow a dev to do more advanced things if they wanted to.

packages/core/src/components/Review/ReviewLink.jsx
```diff
@@ -0,0 +1,32 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+export class ReviewLink extends React.PureComponent {
+  handleClick(event) {
+    if (this.props.onClick) {
+      event.preventDefault();
```
sawyerh: Should we be calling this here or should it be up to the user? They would be able to do this if we pass in the `event` as the first argument below.

packages/core/src/components/Review/Review.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+/**
+ * The `Review` component is used for listing a summary of information entered
+ * by a user. Its content includes a heading, value, and edit link.
+ */
+export class Review extends React.PureComponent {
+  heading() {
+    if (this.props.heading) {
+      return (
+        <h3 className="ds-c-review__heading ds-text ds-u-margin-bottom--0 ds-u-font-weight--bold ds-u-display--inline-block">
```
sawyerh: Will we need a way to customize the heading level used here? This is something @1Copenut did for the `MonthPicker` component by adding a `headingLevel` prop.

packages/core/src/components/Review/Review.jsx
```diff
@@ -0,0 +1,69 @@
...
+Review.defaultProps = {
+  alignTop: true,
+  linkText: 'Edit'
+};
+
+Review.propTypes = {
+  /**
+   * Set to false to vertically center the edit link.
+   */
+  alignTop: PropTypes.bool,
```
sawyerh: Do y'all use this prop anywhere currently? I don't think we have a need for it on App 3, and from a design standpoint, I'd almost prefer not offering this option so directly. If we add support for a `className` prop, someone could pass in `ds-u-align-items--center` if they really wanted it.

For reference though, in case we do want to keep this: in other components a boolean props presence means `true`. (ie. `alignCenter` here instead). This is so its consistent with other HTML attributes like `selected`, `required`, and `checked`.

packages/core/src/components/Review/Review.jsx
```diff
@@ -0,0 +1,69 @@
...
+    );
+  }
+}
+
+Review.defaultProps = {
+  alignTop: true,
+  linkText: 'Edit'
+};
+
+Review.propTypes = {
```
sawyerh: Could we add support for a `className` prop so someone can pass in additional classes to be added to the root element? In your example, `ds-u-margin-top--6` could then be added directly to the `Review` component.

packages/core/src/components/Review/Review.jsx
```diff
@@ -0,0 +1,69 @@
...
+   * Set to false to vertically center the edit link.
+   */
+  alignTop: PropTypes.bool,
+  children: PropTypes.node.isRequired,
+  /**
+   * An optional function that is executed on link click. The href value is
+   * passed to this function.
+   */
+  onClick: PropTypes.func,
+  heading: PropTypes.string,
```
sawyerh: We might want to support `PropTypes.node` here, for a case where someone wants to pass in a linked header, or a heading with additional HTML tags in it.

packages/core/src/components/Review/Review.jsx
```diff
@@ -0,0 +1,69 @@
...
+  href: PropTypes.string.isRequired,
```
sawyerh: Could we document this prop to make it super clear what type of `href` we're looking for? 🙏

packages/core/src/components/Review/ReviewLink.jsx
```diff
@@ -0,0 +1,32 @@
...
+      this.props.onClick(this.props.href);
+    }
+  }
+
+  render() {
+    const { href, className, children } = this.props;
+    const onClick = event => this.handleClick(event);
+    return (
+      <div className="ds-u-margin--0">
+        <a href={href} onClick={onClick} className={className}>
```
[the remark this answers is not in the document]
1Copenut: I'd like this to have an `aria-label` attribute/prop that could be used to read out longer, more descriptive text to screen readers. This has been well-tested across screenreaders, and can be omitted when there's no longer label required.

packages/core/src/components/Review/Review.jsx
hannaliebl: There was a vets.gov screenshot on the original ticket with the edit link vertically centered, but in tax tools it is aligned to the top. I'm happy to remove that prop and just have the user add that via their own class.

sawyerh: Ah, that's my fault, sorry for the confusion. Yeah, let's get rid of this for now. We're not vertically centering this anywhere right now. I'd still advocate for adding support for a `className` prop though 😄

packages/core/src/components/Review/Review.jsx
hannaliebl: I added this: "The href should be a URL that allows the user to modify whatever content the row displays." But that is a bit awkward and may not be what you mean. Do you have thoughts?

sawyerh:
@hannaliebl I think we'll eventually want to add a "Remove" option to this component, though that can happen in a separate PR in the future. For context, we'll likely need that option for the Income/Deductions section in App 3. 

With that in mind, I think we'll want to change the name of some props to avoid confusion in the future. What do you think of the suggested renames below?

`href` -> `editHref` (this also probably removes the need for the docs description)
`linkText` -> `editText`
`onClick` -> `onEditClick`

hannaliebl:
PTAL, I also changed the prop names as you suggested.

packages/core/src/components/Review/Review.jsx
```diff
@@ -0,0 +1,66 @@
...
+import React from 'react';
+import ReviewLink from './ReviewLink';
+import classNames from 'classnames';
+
+/**
+ * The `Review` component wraps up the styles and markup required to make a
+ * single reviewable row. They are designed to be used either alone, as a single
+ * row, or multiple rows can be combined together one after the other to present
+ * more information. The component can take text or HTML as row content.
+ */
```
sawyerh: Great documentation 👌

packages/core/src/components/Review/Review.jsx
```diff
@@ -0,0 +1,66 @@
...
+  heading: PropTypes.node,
+  /**
+   * Heading type to override default `<h3>`.
+   */
+  headingLevel: PropTypes.number,
+  editHref: PropTypes.string.isRequired,
+  editText: PropTypes.node.isRequired,
+  /**
+   * An optional function that is executed on edit link click. The event and
+   * props.href value are passed to this function.
```
sawyerh: Minor thing: This should be `props.editHref`. ProTip: you can wrap that with markdown code backticks to add code formatting to it too.

packages/core/src/components/Review/ReviewLink.jsx
```diff
@@ -0,0 +1,41 @@
...
+    if (this.props.onClick) {
+      this.props.onClick(event, this.props.href);
+    }
+  }
+
+  render() {
+    const { href, className, children, ariaLabel } = this.props;
+    const onClick = event => this.handleClick(event);
+    return (
+      <div className="ds-u-margin--0">
```
sawyerh: Is this margin necessary? Disabling it doesn't seem to have an affect on anything from what I can tell.

PR #238, opened 2018-02-13
 Update example text of DateField

sawyerh:
### Changed

- Removed slashes from the date field's example based on guidance from content team. This reverts things back to how they were, and is consistent with the USWDS.


pwolfert:
Ha! This is the exact opposite of what the tax tools product team wanted. I just changed it from no slashes to slashes there.

sawyerh:
@pwolfert @heyitsolivia  FWIW, I think regardless of what we set as the default here, we'll still need to manually set it in HealthCare.gov, because Spanish.

pwolfert:
That's true

PR #240, opened 2018-02-16
Change: Remove relative positioning from grid columns

sawyerh:
### Changed

- Removed `relative` positioning from grid columns so a developer can position a child element relative to an ancestor farther up the chain. For context: We need to do that for the help drawer in App 3. Based on my testing in Chrome, Firefox, and IE 11, this property doesn't seem like it was necessary and may have been an artifact from older browser support we no longer need. 

### Internal

- Snuck in the ability to pass in a `--port` argument to `yarn start`. This lets me run the design system locally alongside another app running on port `3000`. For example, you can now run `yarn start --port 1337`

### DIFF
--- a/packages/layout/src/_col.scss
+++ b/packages/layout/src/_col.scss
@@ -8,7 +8,6 @@
   min-width: 0; // Resize columns as expected (https://css-tricks.com/flexbox-truncated-text/)
   padding-left: ($grid-gutter-width / 2);
   padding-right: ($grid-gutter-width / 2);
-  position: relative;
   width: 100%;
 }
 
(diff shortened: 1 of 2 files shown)


sawyerh:
Yup, I'd rather rip that bandaid off now if it's not actually necessary. It seems like a thing that would cause more pain than actual good.  @pwolfert Would you consider this a breaking change? 🤔

pwolfert:
@sawyerh, that's a tough question. I don't know if I want _this_ to be what brings us to `v2.0`. It could possibly be a breaking change, but how many apps are using this right now? It's also not going to cause errors, though I know just the behavior difference could be enough to call it breaking.

PR #241, opened 2018-02-20
Add theming and Sass variable documentation 

sawyerh:


packages/core/src/utilities/font-size.scss
```diff
@@ -10,9 +10,10 @@ Use the font size utility to make the text larger or smaller.
 
 Markup:
 <% ['display','title','h1','h2','h3', 'h4', 'lead','base','small'].forEach(size => { -%>
-<article class="ds-u-font-size--<%= size %> ds-u-leading--heading ds-u-margin-bottom--3 ds-u-truncate">
+<article class="ds-u-font-size--<%= size %> ds-u-leading--heading ds-u-margin-bottom--3 ds-u-truncate ds-u-border-bottom--1">
```
pwolfert: When I first looked at this, the bottom border actually confused me about which set of variables belonged to which one. If you look at the top of the list you can tell, but it actually seemed like underlining a heading when my eye first hit it, which made me think the variables below it might belong to a heading. I think maybe an `<hr>` might be more appropriate because there would be some padding between it and the header before it.

packages/docs/src/pages/guidelines/themes.md
```diff
@@ -0,0 +1,48 @@
+---
+title: Theming
+weight: 10
+---
+
+The design system supports the ability to "theme" certain aspects like its color palette, type scale, and spacing. There are a few ways to accomplish this.
+
+## Sass variables
+
+All design system styles are available as either Sass or CSS files. [Sass](https://sass-lang.com/) is a CSS preprocessor and its biggest benefit is the ability to define [variables](https://sass-lang.com/documentation/file.SASS_REFERENCE.html#variables_). Similar to other languages, like JavaScript, you can define a variable once (ie. `$color-primary`) and then reference that variable throughout your CSS declarations.
```
pwolfert: > preprocessor  and its biggest benefit is the ability to define variables
Nit: comma before and

packages/core/src/utilities/font-size.scss
sawyerh: Ah, thanks for flagging. That was the opposite of what I was going for 😅  

I played around some more with this, how's this feel?

image

pwolfert: That looks perfect! Not confusing at all 👍

PR #242, opened 2018-02-23
Adds size and additional className props to Dialog component

pwolfert:
### Added
- `size` prop to `<Dialog>` component
- `className`, `actionsClassName`, and `headerClassName` props to `<Dialog>` component
- Snapshot unit test

### DIFF
--- a/packages/core/src/components/Dialog/Dialog.jsx
+++ b/packages/core/src/components/Dialog/Dialog.jsx
@@ -1,4 +1,5 @@
 import AriaModal from 'react-aria-modal';
+import Button from '../Button/Button';
 import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
@@ -10,6 +11,8 @@ export const Dialog = function(props) {
     ariaCloseLabel,
     children,
     className,
+    closeButtonVariation,
+    closeText,
     escapeExitDisabled,
     headerClassName,
     onExit,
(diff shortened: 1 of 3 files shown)


pwolfert:
Proposition:
We render the button as a `Button` component and add a prop `closeButtonVariation` that defaults to `"transparent"`. Oh, and we also need to translate "Close".
```jsx
<Button
  aria-label={ariaCloseLabel}
  className="ds-c-dialog__close"
  onClick={onExit}
  variation={variation}
>
 {closeText}
</Button>
```

packages/core/src/components/Dialog/Dialog.jsx
```diff
@@ -83,6 +103,10 @@ Dialog.propTypes = {
    * at the bottom of the dialog.
    */
   actions: PropTypes.node,
+  /**
+   * Additional classes to be added to the actions element.
```
sawyerh: ✍️ Nit: Replace `element` with `container` since the class is applied to the node that wraps whatever is passed in as the `actions` prop

packages/core/src/components/Dialog/Dialog.jsx
```diff
@@ -102,12 +136,17 @@ Dialog.propTypes = {
    * it (e.g. `document.getElementById(..)`)
    */
   getApplicationNode: PropTypes.func,
+  /**
+   * Additional classes to be added to the header element.
```
sawyerh: I wonder if "header element" is confusing here, since we don't mention "header" anywhere else for this component. Should we clarify this by saying...

> Additional classes to be added to the header, which wraps the title and close button.

PR #249, opened 2018-03-05
Add currency field component/text field mask prop

sawyerh:
### Added

- Added a "Currency field" component. This extends the default text field component and applies additional styling to indicate the expected value (by default: a dollar amount). To use in React, add a `mask="currency"` prop to the `TextField` component.

   Note: this only includes styling at the moment, and additional masking/validation functionality may come later.
   
   image

### DIFF
--- a/packages/core/src/components/TextField/TextField.example.jsx
+++ b/packages/core/src/components/TextField/TextField.example.jsx
@@ -12,12 +12,6 @@ ReactDOM.render(
       requirementLabel="Optional"
     />
     <TextField
-      ariaLabel="Enter amount in dollars"
-      label="Currency input"
-      mask="dollar"
-      name="currency_example"
-    />
-    <TextField
       errorMessage="Error message example"
       hint="Example of a multiline field with an error"
       label="Multiline"
--- a/packages/core/src/components/TextField/TextField.example.jsx
+++ b/packages/core/src/components/TextField/TextField.example.jsx
@@ -18,7 +18,12 @@ ReactDOM.render(
       multiline
       name="multiline_example"
     />
-
+    <TextField
+      label="Disabled field"
+      disabled
+      name="disabled_example"
+      value="Example value"
+    />
     <TextField label="Password field" name="disabled_example" type="password" />
   </div>,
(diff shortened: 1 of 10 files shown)


packages/core/src/components/TextField/TextField.example.jsx
```diff
@@ -18,12 +18,7 @@ ReactDOM.render(
       multiline
       name="multiline_example"
     />
-    <TextField
```
hannaliebl: Why remove the disabled example?

sawyerh: Ah, I originally replaced it with the currency example before I created a separate page for it. Just added the disabled example back.

packages/core/src/components/TextField/__snapshots__/TextField.test.jsx.snap
```diff
@@ -0,0 +1,30 @@
...
+  <div
+    className="ds-c-field-mask ds-c-field-mask--currency"
+  >
+    <div
+      className="ds-c-field__before ds-c-field__before--currency"
+    >
+      $
+    </div>
+    <input
+      aria-label="Enter amount in dollars"
```
1Copenut: @sawyerh I'd like this `aria-label` to include the printed label + more descriptive information. Screenreader users who are moving around via the Tab key might settle on the input, and in doing so, miss the `<label>` being read out. So all they'll hear is "Enter amount in dollars."  but it's not clear what they should enter dollars for.

sawyerh: PTAL

sawyerh:
@pwolfert Right – This only includes the styling (**$** icon). I'll take a closer look at masking and manipulating the value in a separate PR: https://jira.cms.gov/browse/HDSG-179

sawyerh:
@1Copenut I believe I've addressed your requested changes. Going to merge this, but let me know if I missed anything.

1Copenut:
@sawyerh Yes, this last update captures the change I commented on. Thank you.

PR #250, opened 2018-03-05
[HDSG-193] Add tiny and medium modifiers for input widths

hannaliebl:
### Added
Added `ds-c-field--tiny` and `ds-c-field--medium` classes.

Should I do something like this for `select` as well?

image


packages/core/src/components/TextField/TextField.scss
```diff
@@ -6,11 +6,15 @@ Text field
 
 @uswds https://standards.usa.gov/components/form-controls/#text-input
 
-Text fields allow users to enter any combination of letters, numbers, or symbols of their choosing (unless otherwise restricted). Text fields can span single or multiple lines.
+Text fields allow users to enter any combination of letters, numbers, or symbols of their choosing (unless otherwise restricted). Text fields can span single or multiple lines. Use the `ds-c-field--tiny` or `ds-c-field--medium` classes to constrain the width of the input.
```
sawyerh: Should we include this in the description, or are the examples and code snippets enough for this?

packages/core/src/components/TextField/TextField.scss
```diff
@@ -64,6 +68,14 @@ Style guide: components.text-field
   }
 }
 
+.ds-c-field--tiny {
+  max-width: 6em;
```
sawyerh: Should we define these as variables instead, in case a team wants to override them? We could put them in `_variables.forms.scss`

packages/core/src/components/TextField/TextField.scss
```diff
@@ -64,6 +68,14 @@ Style guide: components.text-field
   }
 }
 
+.ds-c-field--tiny {
```
sawyerh: We use `small` in other places..should we use that to be consistent? I.e. `ds-c-choice--small`, `ds-c-button--small`

pwolfert: Or do we need even more 😮values? `tiny`, `small`, `medium`, `large`, `child-size`

sawyerh: 🙅‍♂️

pwolfert: > `child-size`

If someone doesn't respond with a gif in the next hour, I'll have to

pwolfert: That's horrifying 😱 I was going more for 
200w_d
 😛

pwolfert: Wow, sorry @hannaliebl for vandalizing your PR. The code looks good to me 👍

sawyerh: I was going more for this, but couldn't find a version in GIF format. http://nymag.com/selectall/2018/02/baby-steals-food-from-bird-in-viral-gif.html

Sorry, not sorry @hannaliebl

hannaliebl: That was quite a journey. 😄

hannaliebl:
PTAL. Should I add the size-modified examples to `Select` and `ChoiceList` like I did on `TextField`?

packages/core/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -81,6 +81,10 @@ export class ChoiceList extends React.PureComponent {
    * @param {array} options - <option> components
    */
   select(selectProps, options) {
+    const sizeClass = this.props.size
+      ? `ds-c-field--${this.props.size}`
+      : undefined;
+    const classes = classNames(sizeClass);
```
sawyerh: Do we need the `classNames` line here? Looks like we might be able to get away with just the conditional above since there's one class name possible.

packages/core/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -202,6 +207,10 @@ ChoiceList.propTypes = {
   name: PropTypes.string.isRequired,
   onBlur: PropTypes.func,
   onChange: PropTypes.func,
+  /**
+   * If the component renders a select, set the max-width of the input either to `'small'` (maps to `ds-c-field--small` class) or `'medium'` (maps to `ds-c-field--medium` class)
+   */
+  size: PropTypes.string,
```
sawyerh: Should this be `PropTypes.oneOf(['small', 'medium'])`?

packages/core/src/components/ChoiceList/Select.jsx
```diff
@@ -17,14 +17,18 @@ export const Select = function(props) {
...
   /* eslint-enable prefer-const */
 
+  const sizeClass = size ? `ds-c-field--${size}` : undefined;
+
   const classes = classNames(
     'ds-c-field',
     { 'ds-c-field--inverse': inversed },
-    className
+    className,
+    sizeClass
```
sawyerh: fwiw, in place of this you could also do:

```jsx
size && `ds-c-field--${size}`
```

packages/core/src/components/ChoiceList/Select.jsx
```diff
@@ -82,6 +84,10 @@ Select.propTypes = {
   name: PropTypes.string.isRequired,
   onBlur: PropTypes.func,
   onChange: PropTypes.func,
+  /**
+   * Set the max-width of the input either to `'small'` (maps to `ds-c-field--small` class) or `'medium'` (maps to `ds-c-field--medium` class)
+   */
+  size: PropTypes.string,
```
sawyerh: Should this be PropTypes.oneOf(['small', 'medium'])?

packages/core/src/components/ChoiceList/Select.jsx
```diff
@@ -82,6 +84,10 @@ Select.propTypes = {
   name: PropTypes.string.isRequired,
   onBlur: PropTypes.func,
   onChange: PropTypes.func,
+  /**
+   * Set the max-width of the input either to `'small'` (maps to `ds-c-field--small` class) or `'medium'` (maps to `ds-c-field--medium` class)
```
sawyerh: I dunno if we need to specify what class these match up to, but I don't feel strongly either way.

packages/core/src/components/TextField/TextField.jsx
```diff
@@ -209,6 +212,10 @@ TextField.propTypes = {
    * applicable if this is a multiline field.
    */
   rows: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  /**
+   * Set the max-width of the input either to `'small'` (maps to `ds-c-field--small` class) or `'medium'` (maps to `ds-c-field--medium` class)
+   */
+  size: PropTypes.string,
```
sawyerh: Should this be `PropTypes.oneOf(['small', 'medium'])`?

hannaliebl:
PTAL

packages/core/src/components/ChoiceList/Select.example.jsx
```diff
@@ -3,14 +3,36 @@ import ReactDOM from 'react-dom';
...
+      <option value="1">Option 1</option>
+      <option value="2">Option 2</option>
+      <option value="3">Option 3</option>
+      <option value="4">Option 4</option>
+      <option value="5">Option 5</option>
+      <option value="6">Option 6</option>
+      <option value="7">Option 7</option>
+    </Select>
+    <h4>Small size modifier</h4>
+    <Select defaultValue="2" size="small" name="small-select-demo">
```
sawyerh: Hm, this is actually cutting off the full label. Should we change these options to reflect a "Suffix" list (i.e. `Sr.`, `Jr.`) so the text doesn't get cut off? 

image

packages/core/src/components/ChoiceList/ChoiceList.example.jsx
```diff
@@ -82,5 +91,6 @@ function choices() {
 function options() {
   const options = generateChoices(8);
   options[1].defaultChecked = true;
+  console.log(options);
```
sawyerh: 👀

hannaliebl: Ugh sorry. I will reopen and fix.

PR #251, opened 2018-03-06
Add Currency mask formatting onBlur

sawyerh:
### Changed

- The currency field has its `value` formatted `onBlur`. Formatting includes adding commas and adding a trailing zero if the number only includes a single decimal number.

currency-mask

### DIFF
--- a/packages/core/src/components/TextField/Mask.jsx
+++ b/packages/core/src/components/TextField/Mask.jsx
@@ -73,7 +73,7 @@ export class Mask extends React.PureComponent {
       // Format number with commas. If the number includes a decimal,
       // ensure it includes two decimal points
       value = this.toNumber(value);
-      value = this.stringWithFixedDigits(value.toLocaleString());
+      value = this.stringWithFixedDigits(value.toLocaleString('en-US'));
     }
 
     return value;
--- a/packages/core/src/components/TextField/Mask.jsx
+++ b/packages/core/src/components/TextField/Mask.jsx
@@ -16,6 +16,15 @@ export class Mask extends React.PureComponent {
     };
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (!prevState.debouncedOnBlurEvent && this.state.debouncedOnBlurEvent) {
+      this.field.props.onBlur(this.state.debouncedOnBlurEvent);
+
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({ debouncedOnBlurEvent: null });
+    }
+  }
+
   /**
    * @param {String} value
    * @returns {Number}
@@ -77,11 +86,20 @@ export class Mask extends React.PureComponent {
    * @param {Object} evt
    */
   handleBlur(evt) {
-    this.setState({ value: this.maskedValue(evt.target.value) });
+    const debounceOnBlur = typeof this.field.props.onBlur === 'function';
 
-    if (typeof this.field.props.onBlur === 'function') {
-      this.field.props.onBlur(evt);
+    // We need to retain a reference to the event after the callback
+    // has been called. We pass this onto the consuming app's onBlur
+    // only after the value has been manipulated – this way, the
+    // value returned by event.target.value is the value after masking
+    if (debounceOnBlur) {
+      evt.persist();
     }
+
+    this.setState({
+      value: this.maskedValue(evt.target.value),
+      debouncedOnBlurEvent: debounceOnBlur ? evt : null
+    });
   }
 
   handleChange(evt) {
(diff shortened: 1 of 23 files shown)


packages/core/src/components/TextField/Mask.jsx
```diff
@@ -0,0 +1,115 @@
...
+  maskedValue(value) {
+    value = value.trim();
+
+    if (value === '') return value;
+
+    if (this.props.mask === 'currency') {
+      // Format number with commas. If the number includes a decimal,
+      // ensure it includes two decimal points
+      value = this.toNumber(value);
+      value = this.stringWithFixedDigits(value.toLocaleString());
```
pwolfert: Do you think we should be explicit in the locale used to create the locale string? The `toNumber` function assumes a specific locale that will output it in the format `12,345.67`, but if the client has a European locale, for example, the value could be formatted to `12.345,67`, which would break that function.

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -0,0 +1,115 @@
...
+    return value;
+  }
+
+  /**
+   * To avoid a jarring experience for screen readers, we only
+   * add/remove characters after the field has been blurred,
+   * rather than when the user is typing in the field
+   * @param {Object} evt
+   */
+  handleBlur(evt) {
```
pwolfert: This is a great solution. It avoids all the issues we were experiencing with our old currency input with different browsers handling the cursor positioning differently.

pwolfert:
Two things I found when testing it with a stateful example:
1. `undefined` or empty-string values won't work for the initial value of the `TextField`
2. It doesn't behave as expected when the `TextField` is a controlled component
  - `onChange` is called with intermediate values like `1,0002`. If this is expected, that's fine
  - Fortunately (or by design) if an unmasked value gets passed it as the value, it doesn't mess up masked rendering down the line. For example, if the input currently shows `12,300`, and the component updates it to `123009`, it still shows `12,3009` as you're typing as expected rather than stripping the commas.

sawyerh:
@pwolfert PTAL. Just pushed a small change that modifies how the `onBlur` is called when it's present. The purpose of this is for scenarios where a consuming app (like App 3) does validation onBlur and references the field's value (`event.target.value`). We want the validation to happen on the formatted value, otherwise there's a small chance that the value that gets submitted would be different than what got validated. 

Demo showing this new behavior (console showing `event.target.value`):

currency-mask-onblur

sawyerh:
oops, I think we posted at the same time. Taking a look at your other questions.

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -16,6 +16,15 @@ export class Mask extends React.PureComponent {
     };
   }
 
+  componentDidUpdate(prevProps, prevState) {
+    if (!prevState.debouncedOnBlurEvent && this.state.debouncedOnBlurEvent) {
+      this.field.props.onBlur(this.state.debouncedOnBlurEvent);
+
+      // eslint-disable-next-line react/no-did-update-set-state
+      this.setState({ debouncedOnBlurEvent: null });
```
pwolfert: Do we actually need `debouncedOnBlurEvent` to be in the state? Since we know it's already going to update because of the change in value, could we just attach it to the component instance? (`this.debouncedOnBlurEvent = `)

sawyerh: Hmm, I don't have strong feelings on that so am happy to change. I'm not sure what the performance impacts are of either option tbh.

pwolfert: There's a chance `state.value` could be updating for a different reason, but that's why we still check whether `this.debouncedOnBlurEvent` exists before using it/clearing it.

sawyerh: I think you're right though, since changing the state would require checking if the component needs re-rendered, whereas I don't think that's the case if it's attached to the component instance.

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -0,0 +1,133 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+
+/**
+ * A `Mask` component renders a controlled input field. When the
+ * field is blurred, it applies formatting to improve the readability
+ * of the value. When the field is focused, formatting is removed
```
hannaliebl: I don't see formatting being removed on focus (I assume that the commas are meant to be removed, right?)

sawyerh: Ah good catch. This is actually the expected behavior and this comment is incorrect.

sawyerh:
Thanks @pwolfert @hannaliebl 🙏

PR #253, opened 2018-03-07
Moved react and react-dom to peerDependencies

pwolfert:
### Changed
- Moved `react` and `react-dom` to `peerDependencies` for our core package so webpack can automatically deduplicate in target applications


sawyerh:
Merging to get this into the next release 👍

PR #256, opened 2018-03-12
Add zip, ssn, and phone masks

sawyerh:
### Changed

- Added new mask values to `TextField`: `phone`, `ssn`, and `zip`

image


packages/core/src/components/TextField/Mask.jsx
```diff
@@ -153,11 +202,15 @@ Mask.propTypes = {
...
 export function unmask(value, mask) {
-  if (!value) return value;
+  if (!value || typeof value !== 'string') return value;
+
+  value = value.trim();
 
   if (mask === 'currency') {
     // Preserve only digits, decimal point, or negative symbol
     value = value.match(/^-|[\d.]/g).join('');
+  } else if (['phone', 'ssn'].indexOf(mask) >= 0) {
```
sawyerh: ℹ️ This currently doesn't unmask the `zip` value, though I think it should. App 3 currently expects a hyphen when it validates the value, but this could be updated or the call to `unmaskValue` could be skipped for zip code values. Thoughts, @sansumbrella @allypalanzi?

packages/core/src/components/TextField/Mask.example.jsx
```diff
@@ -0,0 +1,50 @@
+import TextField, { unmaskValue } from './TextField';
+import React from 'react';
+import ReactDOM from 'react-dom';
+// import { unmaskValue } from '@cmsgov/design-system-core';
+
+function handleBlur(evt, mask) {
+  console.log('Unmasked value:', unmaskValue(evt.target.value, mask));
```
hannaliebl: `console.log` left in

hannaliebl: Actually, this in intentional, right? Nevermind.

hannaliebl: But is the comment above `handleBlur` intentional for the example?

sawyerh: Yeah, these are all hacks to demonstrate how someone might implement this 😅 Open to better ideas!

packages/core/src/components/TextField/Mask.jsx
sawyerh: Update: I'm going to support unmasking the zip code value in this method and consuming apps can skip calling `unmaskValue` if they rely on the value being masked, like App 3 currently. 

https://github.com/CMSgov/design-system/pull/256/commits/884739febf010904cb72cad4862f769bb24f615e

packages/core/src/components/TextField/Mask.example.jsx
```diff
@@ -0,0 +1,50 @@
...
+        type="tel"
+        value="1234567890"
+      />
+
+      <TextField
+        label="Social security number (SSN)"
+        mask="ssn"
+        name="ssn_example"
+        onBlur={evt => handleBlur(evt, 'ssn')}
+        value="123456789"
```
pwolfert: Would it be out of scope to make these clip extra characters at the end? Like I shouldn't be able to input `123-45-678910111213`, and it would be nice if this feature handled that part as well as the additive formatting.

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -69,13 +101,30 @@ export class Mask extends React.PureComponent {
...
+      const { mask } = this.props;
       value = value.trim();
 
-      if (this.props.mask === 'currency') {
+      if (mask === 'currency') {
         // Format number with commas. If the number includes a decimal,
         // ensure it includes two decimal points
         value = this.toNumber(value);
         value = this.stringWithFixedDigits(value.toLocaleString('en-US'));
+      } else if (['phone', 'ssn', 'zip'].indexOf(mask) >= 0) {
```
pwolfert: We do most of the work for the `currency` branch in another function. If we move this transformation function to someplace external with a persistent `maskRegex` object, we could just reference `Object.keys(maskRegex).includes(mask)` rather than listing the keys again. What if we also moved `stringWithFixedDigits` outside the class since we never reference `this` inside it?

pwolfert: It's just that I feel like these string transformations that are `string => string` don't need to be inlined here, and it would decrease the complexity of this function. I like how you did it with the currency block.

packages/core/src/components/TextField/Mask.example.jsx
sawyerh: It actually used to do that but I added support for unexpectedly long values. I think in these cases the consuming app should surface an error message rather than the masking functionality removing characters and potentially resulting in an incorrect value being submitted.

sawyerh:
@pwolfert Thanks for the feedback! PTAL when you have a chance – https://github.com/CMSgov/design-system/pull/256/commits/00b0286f62235183ec2f6bcbcf43eb8a81dc6907

PR #258, opened 2018-03-20
Aria-describedby attributes for the DateField fields

pwolfert:
Fulfills HSDG-199 by adding the id and aria-describedby attributes to `<DateField>`.

### Added
- Added the id and aria-describedby fields to `<DateField>`
- Added an optional `id` prop to `<FormLabel>`

### Changed
- 9 snapshots
    - I could have avoided changing all these snapshots but figured it was better to change those than add extra code to the component

Tested with VoiceOver

### DIFF
--- a/packages/core/src/components/FormLabel/FormLabel.jsx
+++ b/packages/core/src/components/FormLabel/FormLabel.jsx
@@ -50,13 +50,14 @@ export class FormLabel extends React.PureComponent {
   }
 
   render() {
+    const { fieldId, id, children } = this.props;
     const ComponentType = this.props.component;
     const labelTextClasses = classNames(this.props.labelClassName);
     const classes = classNames('ds-c-label', this.props.className);
 
     return (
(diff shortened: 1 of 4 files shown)


packages/core/src/components/FormLabel/FormLabel.jsx
```diff
@@ -67,6 +68,11 @@ export class FormLabel extends React.PureComponent {
 FormLabel.defaultProps = { component: 'label' };
 FormLabel.propTypes = {
   children: PropTypes.node.isRequired,
+  /**
+   * A unique `id` for the label element. Useful for referencing the label from
+   * other components with aria-describedby.
```
sawyerh: nit: wrap `aria-describedby` with Markdown code formatting for prettier docs?

sawyerh:
> we supply the ids to FormLabel rather than FormLabel generating its own.

Oh oops, I missed that!

PR #260, opened 2018-03-22
[WIP] Visual regression testing

pwolfert:
This is a work in progress, but I want it up so people can look at it. It works by running `yarn test --e2e`, but it only works on node versions older than `9`.

### Questions
- The `posttest` linting is annoying for this. I could make it another task, but we should talk about how we're structuring these now that we have a separation between unit tests, end-to-end tests, and linting. One "test" task that combines all and then other tasks for individual types of tests?
- Should we put work into automating this so it determines what all the examples are and runs through them rather than manually creating tests for each component? Problem with that is that it would kill our ability to run the tests in parallel.

### Added
- Visual regression testing with `Puppeteer`/`Jest`/`jest-image-snapshot`

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
     "start": "NODE_ENV=development gulp watch --env=development",
     "start:theme": "yarn start --theme",
     "test": "yarn test:unit && yarn test:e2e && yarn lint",
-    "test:unit": "sh scripts/jest.sh",
-    "test:e2e": "sh scripts/jest.sh --e2e",
-    "update-snapshots": "NODE_ENV=test sh scripts/jest.sh -u",
+    "test:unit": "bash scripts/jest.sh",
+    "test:e2e": "bash scripts/jest.sh --e2e",
+    "update-snapshots": "NODE_ENV=test bash scripts/jest.sh -u",
     "lint":
       "eslint tools/ examples/**/src/ packages/**/src/ --ext .js --ext .jsx -f table && gulp lint --env=test --theme"
   },
--- a/scripts/jest.sh
+++ b/scripts/jest.sh
@@ -34,9 +34,9 @@ while [[ $# -gt 0 ]]; do
 done
 
 if [ "$E2E" = true ]; then
-  OPTS="--config=jest.e2e.config.js"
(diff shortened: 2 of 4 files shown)


pwolfert:
Uh, I don't think Travis knows how to `bash`

sawyerh:
Should `__diff_output__` be ignored so image diffs aren't being committed or is there a reason we might want those in the repo?

> Should we put work into automating this so it determines what all the examples are and runs through them rather than manually creating tests for each component? Problem with that is that it would kill our ability to run the tests in parallel.

I like the idea of automating it if that's not too much trouble. Some of the `docs:generate-pages` code could be reused for generating the data we'd need for identifying the URL's. We could use the same data that [gets passed into `generateMarkupPages` now](https://github.com/CMSgov/design-system/blob/master/tools/gulp/docs/index.js#L238), or extend that method to output an array of URL's that get generated. 

I don't know what the performance will be like if the tests no longer run in parallel, but what if we generated the test suites and run them the same way it does now so they do remain in parallel?

sawyerh:
Re. Travis and `bash`, you might need to do something like this? https://docs.travis-ci.com/user/deployment/script/

jest.config.js
```diff
@@ -0,0 +1,14 @@
+module.exports = {
```
sawyerh: Nit: Should we put these Jest config files in `tools/jest` to keep the root clean'ish?

pwolfert: Sure, I didn't think we had a cleanish root, but I guess all those dotfiles are normally hidden 😛

pwolfert:
@sawyerh, it's failing on Travis CI, but I can't view the diff files. Do you know of a way to view the workspace besides using this plugin to upload artifacts to S3 https://docs.travis-ci.com/user/uploading-artifacts/ ?

sawyerh:
@pwolfert Hmm, I'm not super familiar with Travis CI tbh. The S3 solution might be the quickest way to debug things, or there also seems to be a way to SSH into a build: https://docs.travis-ci.com/user/running-build-in-debug-mode/

PR #261, opened 2018-03-29
[HDSG-195] Top-align multiline checkbox radio

hannaliebl:
### Changed
Made input labels top-aligned if they are more than one line by using flexbox to position the label and its contents.

Before:
image

After:
image


packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -57,22 +60,22 @@ Markup:
...
     <legend class="ds-c-label">Select a historical figure</legend>
     <input class="ds-c-choice ds-c-choice--inverse" id="truth-4" type="radio" name="historical-figures-4" value="truth" checked>
-    <label for="truth-4">Sojourner Truth</label>
+    <label for="truth-4" class="ds-u-flex-wrap--wrap ds-u-align-items--center">Sojourner Truth</label>
     <input class="ds-c-choice ds-c-choice--inverse" id="douglass-4" type="radio" name="historical-figures-4" value="douglass">
-    <label for="douglass-4">Frederick Douglass</label>
+    <label for="douglass-4" class="ds-u-flex-wrap--wrap ds-u-align-items--center">Frederick Douglass</label>
     <input class="ds-c-choice ds-c-choice--inverse" id="washington-4" type="radio" name="historical-figures-4" value="washington">
-    <label for="washington-4">Booker T. Washington</label>
+    <label for="washington-4" class="ds-u-flex-wrap--wrap ds-u-align-items--center">Booker T. Washington</label>
```
sawyerh: Since we have a style declaration block for this element (`.ds-c-choice + label`) could we include these styles there to simplify implementation?

packages/core/src/components/FormLabel/FormLabel.jsx
```diff
@@ -52,8 +52,16 @@ export class FormLabel extends React.PureComponent {
   render() {
     const { fieldId, id, children } = this.props;
     const ComponentType = this.props.component;
-    const labelTextClasses = classNames(this.props.labelClassName);
-    const classes = classNames('ds-c-label', this.props.className);
+    const labelTextClasses = classNames(
+      'ds-c-label-content__container',
```
sawyerh: This is a good solution for the choice labels, but since it affects _all_ labels, I wonder if an alternative could be adding `flex-basis: 100%` to `.ds-c-choice + label .ds-c-field-hint` instead so we can selectively apply this style only when it's needed?

packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -27,7 +27,10 @@ Markup:
   <legend class="ds-c-label">Historical figures</legend>
   <span class="ds-c-field__hint">Select all that apply</span>
   <input class="ds-c-choice" id="truth-1" type="checkbox" name="historical-figures-1" value="truth" checked>
-  <label for="truth-1">Sojourner Truth</label>
+  <label for="truth-1">
+    Sojourner Truth
+    <span class="ds-c-field__hint">Example hint text</span>
```
hannaliebl: @sawyerh, I hadn't realized I had added this - is this hint text (or technically requirement text) in the CSS example ok, or should I remove it? The React example does include hint/requirement text to demonstrate how to pass it in via props, but these examples didn't originally include them.

PR #262, opened 2018-03-30
[HDSG-203] Fix hint+requirementLabel bug in FormLabel component

pwolfert:
Fixed bug in FormLabel where the hint text would render incorrectly if it is a node and there is a `requirementLabel`. Updated snapshots may look like they're wrong, but it renders correctly when it goes into the DOM.

### Added
- Added to the FormLabel test to make sure we handle the case when those props are nodes.

### Changed
- Had to separate out the hint destructuring to a const because of the linter even though it doesn't look as good

### Fixed
- HDSG-203


sawyerh:
Thank youuuuu 🙏  Merging and releasing so we can get this into App 3 ASAP.

PR #265, opened 2018-06-22
Improving Tabs & Supporting Preact

somepedro:
### Added
Preact support for tabs.
Disabled tab support.

* Some tests fail prior to this PR changes


packages/core/src/components/Tabs/Tabs.jsx
```diff
@@ -46,7 +46,14 @@ function panelTabId(panel) {
  * @return {Boolean} Is this a TabPanel component?
  */
 function isTabPanel(child) {
-  return child != null && child.type === TabPanel;
+  // Preact doesn't support child.type, React doesn't support child.nodeName
+  return (
+    child != null &&
+    (child.type === TabPanel ||
+      (child.nodeName &&
```
sawyerh: Could we check `child.nodeName === TabPanel` here for Preact?

packages/core/src/components/Tabs/Tab.jsx
```diff
@@ -33,12 +33,13 @@ export class Tab extends React.PureComponent {
...
         aria-selected={String(this.props.selected)}
+        aria-disabled={String(this.props.disabled)}
         aria-controls={this.props.panelId}
         className={classes}
         href={this.href}
         id={this.props.id}
-        onClick={this.handleClick}
-        onKeyDown={this.handleKeyDown}
+        onClick={!this.props.disabled && this.handleClick}
+        onKeyDown={!this.props.disabled && this.handleKeyDown}
```
sawyerh: Could we add a unit test to confirm these events aren't fired when `disabled` is `true`? Check out the `'calls onClick'` test in `Tab.test.jsx` for an example.

packages/core/src/components/Tabs/Tab.jsx
```diff
@@ -33,12 +33,13 @@ export class Tab extends React.PureComponent {
     return (
       <a
         aria-selected={String(this.props.selected)}
+        aria-disabled={String(this.props.disabled)}
```
sawyerh: What's the visual styling of this in the app you're using this in? I think we'll want some visual indication that the tab is disabled. 

I'm also, mainly out of curiosity, interested to learn more about the use case you're using this for? This would help inform the guidance that goes alongside the component.

packages/core/src/components/Tabs/Tabs.jsx
somepedro: This expression never returns true with Preact.

packages/core/src/components/Tabs/Tab.jsx
somepedro: Since we switched qpp-style to use this repo instead of react-tabs, we wanted to not break the support for disabled tabs:
screen shot 2018-06-25 at 3 30 02 pm

sawyerh: Could we add similar visual styling into this? We could use the same styling as the [disabled button](https://design.cms.gov/example/components.button.disabled/). This would also probably save y'all from needing to write as much custom styling. Let me know if you'd like me to jump in for the visuals here.

PR #266, opened 2018-07-03
Fixing Tabs in Prod and Tabs Accessibility

somepedro:
One of the latest tab changes broke the prod version when using with Preact. This PR fixes it. It also addresses accessibility issues.


packages/core/src/components/Tabs/Tab.jsx
```diff
@@ -29,25 +29,35 @@ export class Tab extends React.PureComponent {
...
-          this.tab = tab;
-        }}
-      >
-        {this.props.children}
-      </a>
-    );
+    if (!this.props.disabled) {
+      return (
+        <a
+          role="tab"
```
sawyerh: Could we move `role="tab"` to the `sharedTabProps`? You'll probably get yelled at by ESLint, but you can disable the error by placing the following above the `<a` 

```
// eslint-disable-next-line jsx-a11y/role-supports-aria-props
```

PR #270, opened 2018-07-09
HDSG-206: Allow for Review components without an edit link

pwolfert:
### Background
In App3 there are `<Review>` components that do not have an edit link for one reason or another. We don't need to recommend uneditable review sections as a pattern, but it _would_ be handy for App3 if we didn't need to hack around this requirement to get the desired behavior.

### Changed
- `<Review>` component will not render a `<ReviewLink>` if the `editHref` prop is not provided


packages/core/src/components/Review/Review.test.jsx
```diff
@@ -45,12 +59,10 @@ describe('Review', function() {
 
     expect($p.length).toBe(1);
```
hannaliebl: To keep the style consistent in here, can you remove the `$` from in front of the `p` as well?

pwolfert: I think it was originally prefixed with `$` because it was a [Cheerio wrapper](http://airbnb.io/enzyme/docs/api/ShallowWrapper/render.html). In this test it is still a Cheerio wrapper, and it's still a pattern in tests in other components to prefix these Cheerio wrappers. (I took out Cheerio rendering in the others because it was unnecessary to go beyond shallow rendering to get access to our...aaaaand I just tried removing it in this test as well, and it worked, so I'll make it like the others.

PR #274, opened 2018-08-15
[APPIII-2106] Separate DateField inputs with slashes

scitech:
## Changed
This puts some slashes in between the inputs in a date field.

https://jira.cms.gov/browse/APPIII-2106

### Before:
datefield-before

### After:
datefield-slash-examples

Small & zoomed:
datefield-slash-zoomed


packages/core/src/components/DateField/DateField.scss
```diff
@@ -46,6 +48,11 @@ Style guide: components.date-field
   width: 6ch; // provide additional spacing for browser UI
 }
 
+.ds-c-datefield__separator {
+  font-size: $h2-font-size;
+  line-height: $reset-line-height;
+  margin-bottom: 12px; // visually align with adjacent inputs
```
scitech: I don't love a magic pixel number here, but the slash character in our font isn't flush with its line height. So with that, combined with the way the input/label groups are marked up, it seemed impractical to align this using a less manual method.

packages/core/src/components/DateField/DateField.jsx
```diff
@@ -106,6 +106,9 @@ export class DateField extends React.PureComponent {
             value={this.props.monthValue}
             aria-describedby={labelId}
           />
+          <span role="presentation" className="ds-c-datefield__separator">
```
pwolfert: Nice 👌 `role="presentation"`

packages/core/src/components/DateField/DateField.scss
pwolfert: Proposed solution to the magic: We can give it the same height and line-height and vertical margin as the text inputs, since we know this is always supposed to fit between the text inputs. Users of the design system can change the height and margins of the inputs, and that would throw off this magic number.

packages/core/src/components/DateField/DateField.jsx
1Copenut: Please add `aria-hidden="true"` to these `<span>` tags also. I've been testing with the big three screen readers and their preferred browsers. The "slash" being read out feels not quite right between the text labels.

scitech: @1Copenut Updated, thanks for catching this.

I agree that the slash should not be read out. I was hoping `role="presentation"` would do the trick.

packages/core/src/components/DateField/DateField.scss
scitech: @pwolfert -- something like the most recent change?

PR #276, opened 2018-09-11
[HDSG-215] Fix errors thrown when unmasking field with no numbers

pwolfert:
### Fixed
- Fixed errors thrown when attempting to unmask a field that has only non-numeric characters. This never came up before because the example only calls the `unmask` function `onBlur`

## Steps to reproduce
1. Modify the `Mask.example.jsx` and change all `onBlur`s to `onChange`.
2. Enter values without numbers into the fields like "banana".
3. With the fix, there should be no errors in the console


packages/core/src/components/TextField/Mask.jsx
```diff
@@ -75,6 +75,7 @@ function toDigitsAndAsterisks(value) {
  */
 function toNumber(value) {
   if (typeof value !== 'string') return value;
+  if (!value.match(/\d/)) return undefined;
```
hannaliebl: This can still cause errors when entering things like `11.` or `11. ...` or `11.bbb`, etc. Because we split on `.` later on, it causes an error when periods are included along with invalid info later, because the `parts[1].match(digitsRegex)` becomes `null` in that case and `join` can't be called on it.

Here is one solution (I think...) that works for that edge case:
```
  let b = '';
  if (parts.length >= 2 && parts[1].match(digitsRegex)) {
    b = parts[1].match(digitsRegex).join('');
  }
```

packages/core/dist/components/TextField/Mask.js
```diff
@@ -99,6 +99,7 @@ function toDigitsAndAsterisks(value) {
  */
 function toNumber(value) {
   if (typeof value !== 'string') return value;
+  if (!value.match(/\d/)) return undefined;
```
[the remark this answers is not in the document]
pwolfert: Whoops, no!

packages/core/src/components/TextField/Mask.jsx
pwolfert: Ugh, this whole thing seems more complicated than it needs to be. Maybe I should rethink all these masking functions. That ugh isn't at you, and I think your fix will help. Thanks for reviewing this

pwolfert:
Whoops, I forgot about this. I'll see if I can get to this today.

pwolfert:
Okay, this came up again in our app, so that lit a fire under me. I've addressed the problems you found @hannaliebl. PTAL, and thanks!

packages/core/src/components/TextField/Mask.test.jsx
```diff
@@ -298,6 +298,13 @@ describe('unmask', () => {
     expect(unmask(null)).toBeNull();
   });
 
+  it('returns empty string when there are no numeric characters in the value', () => {
+    expect(unmask('banana', 'currency')).toBe('');
+    expect(unmask('banana', 'zip')).toBe('');
+    expect(unmask('banana', 'ssn')).toBe('');
+    expect(unmask('banana', 'phone')).toBe('');
+  });
+
```
sethetter: What do you think about adding some tests to check for proper rendering of both non-numeric, and numeric with extra junk characters, just to ensure there's no regression? Up further in this file in the `Currency` describe block.

PR #279, opened 2018-09-22
Allow custom edit content in <Review>

bernardwang:
### Background
Healthcare.gov needs the ability to have 2 buttons in `<Review>`. While this can be achieved through the `editText` prop, this wraps the content with an `<a>`, which breaks tabbing & screenreader experiences. 

screen shot 2018-09-18 at 4 26 02 pm

### Added
This PR adds an optional `editContent` prop to `<Review>` to provide functionality for custom edit links. 
Tests and an example were also added. Happy to change the documentation copy or naming.

Example using new `editContent` prop
screen shot 2018-09-21 at 9 38 19 pm

### DIFF
--- a/packages/core/src/components/Review/Review.example.jsx
+++ b/packages/core/src/components/Review/Review.example.jsx
@@ -2,25 +2,38 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import Review from './Review';
 
+const editButtonHref = 'javascript:void(0);';
 ReactDOM.render(
   <div>
-    <Review heading="A single Review component" editHref="javascript:void(0);">
+    <Review heading="A single Review component" editHref={editButtonHref}>
       This is an example of a single React Review component.
     </Review>
     <Review
       className="ds-u-padding-top--6"
       heading="Multiple Review components"
-      editHref="javascript:void(0);"
+      editHref={editButtonHref}
     >
       Multiple Review components can be combined together one after another.
     </Review>
-    <Review heading="Full Name" editHref="javascript:void(0);">
+    <Review
+      heading="A Review component with edit content"
+      editContent={
+        <div>
+          <a href={editButtonHref}>Edit</a>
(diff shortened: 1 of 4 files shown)


packages/core/src/components/Review/Review.example.jsx
```diff
@@ -2,33 +2,46 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import Review from './Review';
 
+const editButtonHref = 'javascript:void(0);';
```
bernardwang: I added `editButtonHref` because `<a href="javascript:void(0);">` would fail eslint, because it requires a valid, navigable address. Open to suggestions on how to better solve this!

pwolfert: I wonder why eslint didn't pick up on this before. Does it only run on staged files?

packages/core/dist/components/Review/Review.js
```diff
@@ -115,7 +117,11 @@ Review.propTypes = {
    * An optional function that is executed on edit link click. The event and
    * props.editHref value are passed to this function.
    */
-  onEditClick: _propTypes2.default.func
+  onEditClick: _propTypes2.default.func,
+  /**
+   * An optional node in place of the edit link. If this defined, no edit link will be shown.
+   */
+  editContent: _propTypes2.default.node
```
pwolfert: You don't need to include the dist files in your PR because they will be built when a release is created.

packages/core/src/components/Review/Review.example.jsx
bernardwang: My guess is because all the other javascript voids were saved to the `editHref` prop, while my example saved it directly to `href`

packages/core/dist/components/Review/Review.js
bernardwang: gotcha!

PR #280, opened 2018-09-28
Handle non-numeric values in currency mask

sethetter:
### Changed
- `<TextField mask="currency" />` will now convert non-numerical values to `0`. 

### Fixed
- `<TextField mask="currency" />` was throwing an error on non-numeric inputs, `Cannot read property 'join' of null`, because it expected the value to be numeric. The above functional change resolves this error.

### DIFF
--- a/packages/core/dist/components/TextField/Mask.js
+++ b/packages/core/dist/components/TextField/Mask.js
@@ -103,6 +103,10 @@ function toNumber(value) {
   // 0 = number, 1 = decimals
   var parts = value.split('.');
   var digitsRegex = /^-|\d/g; // include a check for a beginning "-" for negative numbers
+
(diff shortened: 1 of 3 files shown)


scitech:
This was mentioned in Slack but for visibility here, @pwolfert had something in progress https://github.com/CMSgov/design-system/pull/276

The two have a functional difference -- #276 converts non-numbers to `''` whereas this PR converts them to `0`. Personally I think converting to `''` makes more sense. A user might enter `"don't know"` (which I've seen in production...) or `"ten million dollars"` and it would be inaccurate to say they're attesting to `0`.

sethetter:
@scitech that's a solid point, and also my mistake for missing the already open PR! I'll close this in favor of #276

PR #281, opened 2018-09-28
Masked TextFields to behave as controlled component when value prop is used

pwolfert:
Previously the `Mask` component feigned to be controllable, but it was in actuality always a stateful, uncontrolled component. This fixes it and allows us to use it as a controlled component if we wish.

### Background

In our app, we treat all our inputs as controlled components. We treated masked `<TextField>`s as controlled components too, using its `value` prop. However, we discovered recently that that's not how it behaves, and if we take back control and try to clear out the value after the user has already typed things into it, the `<Mask>` component wouldn't listen. It stubbornly continued rendering what was stored in its local state.

This is what happens currently when we use it as a controlled component:
```
Parent           | <Mask>
-----------------|-----------
value={1000}     | 
       \         |
         ------> | state.value = 1,000
                 | renders 1,000
                 |
                 | user inputs (1,0000):
                 |   state.value = 1,0000
                 |   onChange
                 |       /
unmask value and | <----
store as 10000:  |
  value={10000}  |
         \       |
           ----> | no change.
                 |
                 | loses focus:
                 |   state.value = 10,000
                 |   onBlur
```

We really don't want to be bothered by the masked version of its value in our form state, so it's not actually convenient to keep track of the mask for that component—otherwise we'd just use a normal `<TextField>` component and write all the masking code in the parent component that renders it. That being said, if we were ever to update the `value` prop to, say, `null`, it would still show `10,000`. This is a problem because we (I) assume it will behave as a controlled component even though it's clearly not by its behavior.

### Changed
- `Mask` example now uses `defaultValue` since the example is of an uncontrolled component

### Fixed
- `Mask` now behaves as a controlled component when the `value` prop is used

### DIFF
--- a/packages/core/src/components/TextField/Mask.jsx
+++ b/packages/core/src/components/TextField/Mask.jsx
@@ -105,7 +105,7 @@ Style guide: components.masked-field.react
  * field is blurred, it applies formatting to improve the readability
  * of the value.
  */
-class Mask extends React.PureComponent {
+class _Mask extends React.PureComponent {
   static getDerivedStateFromProps(props, state) {
     const fieldProps = React.Children.only(props.children).props;
     const isControlled = fieldProps.value !== undefined;
@@ -207,7 +207,7 @@ class Mask extends React.PureComponent {
   }
 }
 
-Mask.propTypes = {
+_Mask.propTypes = {
   /** Pass the input as the child */
   children: PropTypes.node.isRequired,
   mask: PropTypes.string.isRequired
@@ -258,7 +258,7 @@ export function unmask(value, mask) {
   return value;
 }
 
-const PolyfilledMask = polyfill(Mask);
+const Mask = polyfill(_Mask);
 
-export { PolyfilledMask as Mask };
-export default PolyfilledMask;
+export { Mask };
+export default Mask;
--- a/packages/core/src/components/TextField/Mask.jsx
+++ b/packages/core/src/components/TextField/Mask.jsx
@@ -8,6 +8,7 @@ Style guide: components.masked-field
 */
 import PropTypes from 'prop-types';
 import React from 'react';
+import { polyfill } from 'react-lifecycles-compat';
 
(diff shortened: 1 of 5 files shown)


tbthorpe:
I didnt have time to really test it, but I dig this. Also, thanks for digging into this. More discussion to come next week

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -104,12 +105,30 @@ Style guide: components.masked-field.react
  * field is blurred, it applies formatting to improve the readability
```
bernardwang: Small thing, could you update the comment here to not say `controlled`?

bernardwang: I just noticed 94 is outdated, `unmaskValue` should be `unmask` I think.

bernardwang:
This LGTM! Only suggestion I have is that now that `maskValue` is pulled out of the component, it might be nice to have some tests.

packages/core/src/components/TextField/Mask.jsx
pwolfert: > A Mask component renders a controlled input field
is still a true statement, because it always uses its internal state to provide a value to the input itself. `Mask` is kind of a confusing component because it is a child of the `TextField` and yet it renders the actual HTML `input` element. And this input element is always controlled by `Mask` if `Mask` is rendered.

pwolfert: And line 94 is also still accurate because it's exported as `unmaskValue` here: https://github.com/CMSgov/design-system/blob/master/packages/core/src/components/TextField/TextField.jsx#L8-L9

pwolfert: Thanks for paying attention to these details though

bernardwang: gotcha thanks for clarifying! 👍

PR #283, opened 2018-10-08
Allow Tabs to detect TabPanels in React

### LINKED ISSUE #271: Tabs does not detect TabPanels in React
Attempts to use the example for Tabs results in a seemingly empty div.
### JSX
```
                <Tabs>
                    <TabPanel id="interventions" tab="Interventions">
                        someother stuff here
                    </TabPanel>
                    <TabPanel id="implementation" tab="Implementation Costs">
                        some stuffh ere
                    </TabPanel>
                </Tabs>
```
### html
```
<div>
    <div class="ds-c-tabs" role="tablist"></div>
    <div aria-hidden="true" aria-disabled="false" class="ds-c-tabs__panel" id="interventions" role="tabpanel">someother stuff here</div>
    <div aria-hidden="true" aria-disabled="false" class="ds-c-tabs__panel" id="implementation" role="tabpanel">some stuffh ere</div>
</div>
```

Traced it to the line that detects `TabPanels` (`child.type === TabPanel`); see screenshot

screen shot 2018-07-16 at 2 57 38 pm

Inquiring to change to `child.props && child.props.tab`.

### Environment:
```
"@cmsgov/design-system-core": "^1.23.0",
"react": "16.3.0",
```
Components retrieved via individual exports, i.e.
```
import { TabPanel } from '@cmsgov/design-system-core/dist/components/Tabs/TabPanel';
import { Tabs } from '@cmsgov/design-system-core/dist/components/Tabs/Tabs'; 
```

erinrep:
### Changed
Changed the isTabPanel function to check for child.props.tab instead of child.attributes.tab. This will still support Preact, but will also account for wrapped TabPanel React components where the type is not TabPanel.

### Fixed
Fixes #271

### DIFF
--- a/packages/core/dist/components/Tabs/Tabs.js
+++ b/packages/core/dist/components/Tabs/Tabs.js
@@ -78,8 +78,8 @@ function panelTabId(panel) {
  * @return {Boolean} Is this a TabPanel component?
  */
 function isTabPanel(child) {
-  // Preact doesn't support child.type
-  return child && (child.type === _TabPanel2.default || child.props && child.props.tab);
+  // Preact doesn't support child.type, React doesn't support child.attributes
+  return child && (child.type === _TabPanel2.default || child.attributes && child.attributes.tab);
 }
 
 /**
--- a/packages/core/dist/components/Tabs/Tabs.js
+++ b/packages/core/dist/components/Tabs/Tabs.js
@@ -78,8 +78,8 @@ function panelTabId(panel) {
  * @return {Boolean} Is this a TabPanel component?
  */
 function isTabPanel(child) {
-  // Preact doesn't support child.type, React doesn't support child.attributes
-  return child && (child.type === _TabPanel2.default || child.attributes && child.attributes.tab);
+  // Preact doesn't support child.type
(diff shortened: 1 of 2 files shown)


pwolfert:
So the scenario here is that you have `<TabPanel>` components that are wrapped, so `type === TabPanel` will be false? As for `child.attributes`, I don't even know what that is or why it isn't already `child.props`.

You may exclude the `dist/*` files from the diff.

erinrep:
Yep, we have a component that wraps the TabPanel, although it looks like even if it were unwrapped, the transpiled version of TabPanel still might not pass that type comparison. I think child.attributes was added as Preact support, but I'm not sure why child.props wasn't used.

I'll update the PR to not include the dist file change. Thanks!

Also, what is the process to get a version bump?

pwolfert:
I don't know anything about Preact, so I'll defer to you. Once you say your code is working, I'll go ahead and approve this. The current change makes sense to me, but it sounds like it's not working.

Once this is merged, I can release a new version.

erinrep:
Cool, yep our code is working with this change. Thanks!

PR #286, opened 2018-10-24
Fix to include the theme - if present - into the docs.css for local devs

Andrew565:
### Fixed
docs.scss now imports the theme's .scss file(s), when building with `yarn start:theme` or `yarn build:theme`.

### Changed
Also, updated the yarn lockfile.


pwolfert:
The tests are failing because they're not configured to run on other repositories, so I'm going to merge this despite the failing tests. I've run the tests locally, and they're fine.

PR #290, opened 2018-10-26
Fix errors on Travis CI

pwolfert:
Sometime in the last week, our unit tests on Travis CI started failing. I started by fixing the `SecurityError: localStorage is not available for opaque origins` error by changing Jest's `testURL` to localhost. That revealed some more errors, one of which was some strange behavior of the `Mask` component's lifecycle methods when rendered with Enzyme and jsdom. Rather than spending days fixing those issues when it worked perfectly in real life, I removed the polyfill I had been including to give us one of the new lifecycle methods and implemented the behavior with `componentDidUpdate` instead. This also meant I had to suppress an error that eslint was giving for having `setState` in `componentDidUpdate`. It's not wrong to put a `setState` in `componentDidUpdate`, but it's not something that should be encouraged generally. I did upgrade Enzyme to the latest version.

### Fixed
- Fixed the `SecurityError: localStorage is not available for opaque origins` error for our automated tests on Travis CI
- Fixed other unit-test errors that were most likely due to recent dependency updates

Note that I'm including the compiled version of `Mask.js` now because apparently some of the tests on CI rely on the dist copies, and the dist copy from master tries to require `react-lifecycles-compat`, which I removed 🤷‍♂️


pwolfert:
It looks like Travis CI is also getting a slightly different version of Jest or one of its dependencies, which fails some tests. I think it would be good to upgrade Jest to v23, but first I'll fix the current errors.

pwolfert:
Thanks, @dcloud! About the node update: there's probably a lot of updating we need to do at this point 😬

PR #296, opened 2018-12-11
removed up/down arrows from datefield via css; scroll capability stil…

phiden:
### Changed
- removed increment arrows from date field inputs per https://jira.cms.gov/browse/HDSG-234

The ultimate goal was the visual fix of the arrows; thus scrollwheel and keyboard incremental functionality is retained. 

Before:
screen shot 2018-12-11 at 1 32 43 pm

After:
screen shot 2018-12-11 at 1 32 51 pm

### DIFF
--- a/packages/core/src/components/DateField/DateField.scss
+++ b/packages/core/src/components/DateField/DateField.scss
@@ -34,15 +34,6 @@ Markup:
 Style guide: components.date-field
 */
 
-input[type="number"]::-webkit-outer-spin-button, input[type="number"]::-webkit-inner-spin-button {
-  -webkit-appearance: none;
-  margin: 0;
-}
-
-input[type="number"] {
-  -moz-appearance: textfield;
-}
-
 .ds-c-field--month,
 .ds-c-field--day,
 .ds-c-field--year {
--- a/packages/core/src/components/DateField/DateField.scss
+++ b/packages/core/src/components/DateField/DateField.scss
@@ -33,6 +33,16 @@ Markup:
 
 Style guide: components.date-field
 */
+
+input[type="number"]::-webkit-outer-spin-button, input[type="number"]::-webkit-inner-spin-button {
+  -webkit-appearance: none;
+  margin: 0;
+}
+
+input[type="number"] {
+  -moz-appearance: textfield;
+}
+
 .ds-c-field--month,
 .ds-c-field--day,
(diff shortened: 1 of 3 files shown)


packages/core/src/components/DateField/DateField.scss
```diff
@@ -33,6 +33,16 @@ Markup:
 
 Style guide: components.date-field
 */
+
+input[type="number"]::-webkit-outer-spin-button, input[type="number"]::-webkit-inner-spin-button {
```
pwolfert: These selectors will apply to all text fields of the number type rather than just the date-field ones. Is that your intention?

pwolfert: And I see that the ticket requests that the fields be turned into normal text fields. I like your solution better, but I just thought I'd bring that up.

phiden: @pwolfert it wasn't, but thinking about it I think it's all right that all number fields don't have the increment arrows. They're a confusing UI and they don't always work as expected, which was the genesis for getting rid of them in the date field. 

RE: turning into normal text fields, discussed that with Mel and we agreed that retaining the number fields and removing the arrow UI was a good solution -- the problem is not the functionality of the number field per se, but of the increment arrows themselves. (the original AC is too prescriptive in this case)

pwolfert: Cool, sounds good to me. In this case I'd suggest we move this into https://github.com/CMSgov/design-system/blob/master/packages/core/src/components/TextField/TextField.scss#L45

pwolfert: Yeah, that's the linter. It's rather opinionated 😕. I'll pull down your branch and try to fix them. You mind if I commit to your branch?

Also, none of our existing CSS has ever selected on the bare `input`s. I think it would be more in line with the existing pattern if the selector were `.ds-c-field[type="number"]...etc` that way we don't overstep our bounds.

phiden: No, please go ahead! Appreciate the help.

pwolfert: I'll try to get to it this morning, but if I don't make any updates in the next two hours, yell at me.

phiden:
@pwolfert looks like this is good to go?

PR #299, opened 2018-12-26
[TAXTOOLS-556] Add Help Drawer component

hannaliebl:
### Added
Added `HelpDrawer` and `HelpDrawerToggle` components (based on App 3 components.)

help-drawer-gif
help-drawer

### Open Questions
1. Will this work for App 3 implementation? I based all of this on the existing components in App 3, but the state management/id handling is a bit different, and I'm not sure if this is too simplified or if there is more I should do with that to support how App 3 uses these components.
2. `componentWillUpdate` is almost depreciated and will be in future versions of React - I assume I should find a better way to handle this, but this was how it was done in App 3 so I copied it here for now.
3. I'd love help with more documentation and perhaps better examples (no "lorem ipsum.") Feel free to add to this branch if that's easiest with better documentation and examples.


packages/core/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -0,0 +1,74 @@
...
+          <div className="ds-u-fill--gray-lightest ds-u-padding--2 ds-u-display--flex ds-u-align-items--start">
+            <h3
+              ref={el => (this.titleRef = el)}
+              tabIndex="0"
+              className="ds-u-text--lead ds-u-margin-y--0 ds-u-margin-right--2"
+            >
+              {title}
+            </h3>
+            <Button
+              className="ds-u-margin-left--auto"
```
1Copenut: I'd like to have an `aria-label` attribute added here, with a default string of "Close help drawer". This will probably need English and Spanish translations, and could be a `HelpDrawer` default prop. Screen reader users don't get a lot of context when "Close, button" is announced on focus. The suggested text would read "Close help drawer, button" which is more descriptive and doesn't change the visual design.

packages/core/src/components/HelpDrawer/HelpDrawerToggle.jsx
```diff
@@ -0,0 +1,67 @@
...
+/**
+ * A link that triggers the visibility of a help drawer
+ */
+export class HelpDrawerToggle extends React.PureComponent {
+  constructor(props) {
+    super(props);
+    // Use a unique ID to avoid scenarios where multiple toggles on the same
+    // page trigger the same help drawer, which could result in multiple instances
+    // of the drawer being rendered if these weren't unique:
+    this.id = uniqueId('helpDrawer');
```
pwolfert: I think all this `id` business might be specific to the App 3 implementation. Looking at the `.example.js` component, I don't see how multiple toggles changing the same bit of component state would cause the component to render multiple help drawers because there's only one in the render function. And if none of that's necessary, I don't think this `HelpDrawerToggle` component is necessary to create either, since it basically becomes a link. That being said, there's a lot of boilerplate to make this link behave like a button, so maybe it's good to keep around.

This also brings up your question about being able to use this in App 3 after it's released. I think we'll just have to use some sort of adapter for the time being, but I don't think that's a bad thing.

hannaliebl: I'm ok with keeping the toggle a separate component. I removed the id handling from the toggle - do you think the way it works now would work for App3?

packages/core/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -0,0 +1,74 @@
...
+            */}
+          <div className="ds-u-fill--gray-lightest ds-u-padding--2 ds-u-display--flex ds-u-align-items--start">
+            <h3
+              ref={el => (this.titleRef = el)}
+              tabIndex="0"
+              className="ds-u-text--lead ds-u-margin-y--0 ds-u-margin-right--2"
+            >
+              {title}
+            </h3>
+            <Button
```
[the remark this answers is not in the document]
hannaliebl: I updated this, let me know if there's anything else you see that can be improved!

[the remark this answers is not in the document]
hannaliebl: Oh, and, yes, I made it a default prop. We should be able to pass in translations ourselves when using this - we don't provide translations or keys for the default components here, though, correct?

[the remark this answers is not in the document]
1Copenut: I don't believe we provide translations or keys for defaults.

packages/core/src/components/HelpDrawer/HelpDrawerToggle.jsx
pwolfert: Yeah, it should still work 👍Looks great

packages/core/src/components/HelpDrawer/HelpDrawerToggle.jsx
```diff
@@ -0,0 +1,46 @@
...
+          {this.props.children}
+        </a>
+      </span>
+    );
+  }
+}
+
+/* eslint-disable react/no-unused-prop-types */
+HelpDrawerToggle.propTypes = {
+  /** Whether or not the Help Drawer controlled by this toggle is open or closed. This value is used to re-focus the toggle that opened the drawer when the drawer closes. */
```
pwolfert: What's the intended behavior with regards to focusing? When you open the help drawer, is it supposed to move focus back to the button rather than focus on the help drawer's title?

packages/core/src/components/HelpDrawer/HelpDrawer.jsx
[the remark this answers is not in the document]
1Copenut: @hannaliebl I just re-checked locally with these changes. No regressions and the Close button action is clear when it's read out. No further changes from me. Thank you much!

packages/core/src/components/HelpDrawer/HelpDrawerToggle.jsx
hannaliebl: It should focus the title of the help drawer on open (I'm not entirely sure why, this was behavior that I duplicated from the original - @1Copenut can give a better answer, but maybe because the title is read out by screen readers and it makes more sense for it to first read the title rather than immediately reading the close button?) and on close it should return focus to the toggle that opened it (so you're back to your original spot in the DOM before it opened.)

packages/core/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -0,0 +1,79 @@
...
+        <div className="ds-c-help-drawer__body ds-u-md-font-size--small ds-u-lg-font-size--base ds-u-padding--2">
+          {children}
+        </div>
+        <div className="ds-c-help-drawer__footer ds-u-fill--primary-alt-lightest ds-u-md-font-size--small ds-u-lg-font-size--base ds-u-padding--2">
+          <h4 className="ds-text ds-u-margin--0">{footerTitle}</h4>
+          {/* dangerouslySetInnerHTML since footerBody can include HTML tags */}
+          <p
+            className="ds-text ds-u-margin--0"
+            dangerouslySetInnerHTML={{ __html: footerBody }}
+          />
```
pwolfert: What if we leave the danger up to the developer who uses the component? That is, we just expect a renderable node, and if the developer wants to render a string, they can turn it into html themselves. So we'd have:
```jsx
<div className="ds-text ds-u-margin--0">
  {footerBody}
</div>
```
and they could pass stuff like
```jsx
<HelpDrawer footerBody="just some text" />
<HelpDrawer footerBody={<span>some <em>special</em> stuff </span>} />
<HelpDrawer footerBody={<span dangerouslySetInnerHTML={{ __html: stuff }} />} />
```

packages/core/src/components/HelpDrawer/HelpDrawerToggle.jsx
1Copenut: @hannaliebl @pwolfert 
You're exactly right. Settling focus on the heading when the Help Drawer is opened causes it to be read out immediately to the big three screen readers. NVDA doesn't always read out content when we focus a container or non-text node, so I'm trying to rally around heading focus for programmatic UI changes where possible, and fall back to non-text nodes where necessary.

Here we had a chance to hook into the heading, and it's maximally useful to all assistive devices I tested against.

hannaliebl:
@pwolfert, PTAL, I made the footerBody no longer use`dangerouslySetInnerHTML`.

Also, I was going to get better example text, but I couldn't think of any, and I realize we use lorem ipsum in other spots, so from my perspective this is ready to be merged now as-is, unless someone has any other examples (or other feedback) they'd like to give.

PR #303, opened 2019-01-09
Datefield: replace For example: 4 28 1986 with For example: 4 / 28 / 1986 

phiden:
### Changed
In Datefield, added `/` to the help text.

screen shot 2019-01-09 at 11 05 15 am

### DIFF
--- a/packages/core/dist/components/DateField/DateField.js
+++ b/packages/core/dist/components/DateField/DateField.js
@@ -217,7 +217,7 @@ var DateField = exports.DateField = function (_React$PureComponent) {
 
 DateField.defaultProps = {
   label: 'Date',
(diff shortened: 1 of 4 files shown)


1Copenut:
@phiden LGTM. I'd say it might be worth adjusting lines 21 and 26 to reflect our changed spans in the DateField.scss, but that won't require another review. Thank you!

* https://github.com/CMSgov/design-system/blob/7f8d71892823f4cca41fe6c5c31c28dc47a8e2d5/packages/core/src/components/DateField/DateField.scss#L21
* https://github.com/CMSgov/design-system/blob/7f8d71892823f4cca41fe6c5c31c28dc47a8e2d5/packages/core/src/components/DateField/DateField.scss#L26

PR #306, opened 2019-01-15
Update docs

phiden:
### Changed
Restructured 'getting started' section by splitting sections into individual files, which triggered more easily found side navigation. Also updated some copy. 

Please note: this PR will have a broken link until https://github.com/CMSgov/design-system/pull/298 is merged.


packages/docs/src/pages/index.md
```diff
@@ -24,4 +24,4 @@ In order to meet these goals we are:
 - Reducing the overall footprint of websites by writing [scalable CSS]({{root}}/guidelines/code-conventions).
 - Supporting rapid development with [utility classes]({{root}}/utilities) and easy-to-use components with best practices baked in.
 - Reducing the number of decisions needed when iterating on UI by providing a solid foundation of [typography]({{root}}/style/typography), [layout]({{root}}/layout/grid), and [color]({{root}}/style/color).
-- Distributing resources as [NPM packages]({{root}}/getting-started) which can be easily installed, updated, and imported.
\ No newline at end of file
+- Distributing resources as [NPM packages]({{root}}/getting-started) which can be easily installed, updated, and imported.
```
hannaliebl: I think this should link to `/startup/installation/` now

packages/docs/src/pages/startup/examples.md
```diff
@@ -0,0 +1,14 @@
+---
+title: Examples
+weight: 4
+---
+
+<h2 id="examples" class="ds-h2 ds-u-color--primary-darker">Examples</h2>
```
hannaliebl: The headings here are not done the same way as they are in other parts of the docs. I would propose removing all the top-level h2s entirely from these new pages, since they are duplicated in the main page title (since each section that used to be in the old single page Getting Started guide is now its own page.) This is what I mean visually with that, basically, the h2 is redundant here:

screen shot 2019-01-17 at 3 42 47 pm

I would also suggest maybe just using markdown syntax for all the headings. So `## Need Help` instead of the H2 tag below. I think that will standardize the colors with the rest of the docs (unless there is some reason I don't know about that the headings for these pages should be blue.)

packages/docs/src/pages/startup/installation.md
```diff
@@ -0,0 +1,41 @@
...
+* Responsive flexbox grid framework
+
+```
+npm install --save @cmsgov/design-system-layout
+```
+
+<h2 id="need-help" class="ds-h2 ds-u-color--primary-darker">Need more information on NPM?</h2>
+<p class="ds-c-alert__text">Learn more about using packages and NPM at <a href="https://www.npmjs.com/">npmjs.org</a>.</p>
+
+<h3>Usage</h3>
```
hannaliebl: I think this should be an h2 (it seems like it's the same level as the "Need more information on NPM" h2 above it)

packages/docs/src/pages/startup/examples.md
```diff
@@ -0,0 +1,14 @@
...
+
+Additional examples of the design system in use can be viewed on GitHub. These projects demonstrate the various ways you can incorporate the design system into your development process and various use cases.
+
+<a href="https://github.com/CMSgov/design-system/tree/master/examples/" class="ds-c-button">Browse example projects</a>
+
+<h2 id="need-help" class="ds-h2 ds-u-color--primary-darker">Need help or ran into an issue?</h2>
```
hannaliebl: Thoughts on having this same Need Help heading on every page of the startup? I think it definitely makes sense on the first section (Installation), but I'm not sure it needs to be on every page. I don't know what a best practice is here, though.

packages/docs/src/pages/startup/react.md
```diff
@@ -1,55 +1,9 @@
...
-
-You can manually copy these directories, or you could integrate this step into your build process. [Here's an example of how this step could be accomplished using a Gulp task](https://github.com/CMSgov/design-system/blob/master/examples/react-app/Gulpfile.js).
-
-You can change the default paths by overriding the following Sass variables:
-
-* `$font-path`
-* `$image-path`
-
-<h2 id="usage" class="ds-h2 ds-u-color--primary-darker">Usage</h2>
+<h2 id="installation" class="ds-h2 ds-u-color--primary-darker">Usage</h2>
```
hannaliebl: This page has the Using Sass/CSS and Examples sections included in it still as well as the Usage section from the Installation page (I just think those need to be cleared out, leaving only the React Components section here.)

packages/docs/src/pages/startup/examples.md
phiden: I'd like to keep it for now -- I'm hearing some feedback that finding help is a little tough, and there's no harm in having it repeated here.

packages/docs/src/pages/startup/examples.md
phiden: Removed the second, redundant header. I think the blue h2 was an attempt to hang on to typographical styles, but that doesn't mean we need to continue that.

packages/docs/src/pages/startup/installation.md
phiden: Agree. Bumped it up and bumped the NPM header down one. Interestingly, Markdown's not rendering, so leaving it as full html markup.

PR #310, opened 2019-02-03
ChoiceList: add `ariaLabel` prop for select component type

amistein:
### Added
* A new prop - `ariaLabel` - to the `ChoiceList` component. `ChoiceList` will pass the value of this prop as the `aria-label` attribute when it renders a `Select` component. This addition is necessary for 508 compliance in some instances. See https://jira.cms.gov/browse/WNMGAPPIII-201.

### DIFF
--- a/packages/core/src/components/ChoiceList/ChoiceList.jsx
+++ b/packages/core/src/components/ChoiceList/ChoiceList.jsx
@@ -91,6 +91,7 @@ export class ChoiceList extends React.PureComponent {
         onBlur={this.props.onBlur}
         onChange={this.props.onChange}
         className={classes}
+        aria-label={this.props.ariaLabel}
         {...selectProps}
       >
         {options}
@@ -216,7 +217,11 @@ ChoiceList.propTypes = {
    * `checkbox` fields will be rendered. If less than 10 choices are passed in,
    * then `radio` buttons will be rendered.
    */
-  type: PropTypes.oneOf(['checkbox', 'radio', 'select'])
+  type: PropTypes.oneOf(['checkbox', 'radio', 'select']),
+  /**
+   * Adds `aria-label` attribute if component renders a select
+   */
+  ariaLabel: PropTypes.string
 };
 
 export default ChoiceList;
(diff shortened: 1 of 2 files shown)


packages/core/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -216,7 +217,11 @@ ChoiceList.propTypes = {
    * `checkbox` fields will be rendered. If less than 10 choices are passed in,
    * then `radio` buttons will be rendered.
    */
-  type: PropTypes.oneOf(['checkbox', 'radio', 'select'])
+  type: PropTypes.oneOf(['checkbox', 'radio', 'select']),
+  /**
+   * Adds `aria-label` attribute if component renders a select
```
bernardwang: Given this is specific to `Select`, could you add this to `Select.propTypes` in `Select.jsx` as well?

amistein: Good point. Fixed in a985582340118222c7f5df9b92c1ffc4d104f3b2.

packages/core/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -91,6 +91,7 @@ export class ChoiceList extends React.PureComponent {
         onBlur={this.props.onBlur}
         onChange={this.props.onChange}
         className={classes}
+        aria-label={this.props.ariaLabel}
```
1Copenut: @amistein Looking at this again in terms of @bernardwang comment, we may need to add this prop to the `Select` component also. We can render an HTML `<select>` from ChoiceList, but also the standalone Select. Should be a one line add, and give us the flexibility to use proper FormLabel components as the best practice, or `aria-label` for cases where that's not possible.

amistein: It seems like we are currently able to use `aria-label` with the `<Select>` component directly, since in `<Select.jsx>` we pass `...selectProps` to the HTML `<select>`. Do you want me to add it as an explicit prop?

1Copenut: Yeah, you're right @amistein. Since we can pass undocumented props directly into the `<Select />` it's okay to pass on adding `aria-label`. Good call.

PR #314, opened 2019-02-08
Fix broken link to the measure utility

line47:
### Fixed
Broken link to the measure utility from the alerts component page. 
Before fix link was: https://design.cms.gov/components/alert/%7B%7Broot%7D%7D/utilities/measure
After fix link is: https://design.cms.gov/utilities/measure/


line47:
Is this ok for me to merge or how does the process work?

pwolfert:
Yep, go ahead and _squash and merge_!

PR #315, opened 2019-02-08
Add ARIA roles to child content container

line47:
### Added
- `aria-live="polite" and aria-relevant="additions text" and aria-atomic="false"` to choice when  `checkedChildren` prop is used. 
- Example of `checkedChildren` to `choice` example for both radios and checkboxes 

### Looking to Fix
This is working to fix an issue where the choice element is not to announcing the content in `checkedChildren` when it is revealed.

### To do
- [x] Check against screen readers @1Copenut 
- [x] Add styles for the child content container. This should be a follow on issue

### DIFF
--- a/packages/core/src/components/ChoiceList/Choice.example.jsx
+++ b/packages/core/src/components/ChoiceList/Choice.example.jsx
@@ -66,16 +66,16 @@ ReactDOM.render(
     <div className="ds-u-margin-top--4">
       <fieldset className="ds-c-fieldset">
         <legend className="ds-c-label">Radios with children</legend>
-        <Choice name="radio_choice" type="radio" value="a">
-          Radio A
+        <Choice name="radio_choice_children" type="radio" value="c">
+          Radio C
         </Choice>
         <Choice
-          name="radio_choice"
+          name="radio_choice_children"
           type="radio"
-          value="b"
+          value="d"
           checkedChildren={childSelect}
         >
-          Radio B - with children
+          Radio D - with children
         </Choice>
       </fieldset>
     </div>
--- a/packages/core/src/components/ChoiceList/Choice.example.jsx
+++ b/packages/core/src/components/ChoiceList/Choice.example.jsx
@@ -1,6 +1,26 @@
 import Choice from './Choice';
+import ChoiceList from './ChoiceList';
 import React from 'react';
 import ReactDOM from 'react-dom';
+
+const selectChoices = [
+  { label: 'A', value: 'A' },
+  { defaultChecked: true, label: 'B', value: 'B' },
+  { label: 'C', value: 'C' },
+  { label: 'D', value: 'D' },
+  { label: 'E', value: 'E' },
+  { label: 'F', value: 'F' },
+  { label: 'G', value: 'G' },
+  { label: 'H', value: 'H' }
+];
+
+const childSelect = (
+  <ChoiceList
+    choices={selectChoices}
+    label="Select example"
+    name="select_choices_field"
+  />
+);
 
 ReactDOM.render(
   <div>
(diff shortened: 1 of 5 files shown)


line47:
@pwolfert or @hannaliebl if you can help and have thoughts on why this PR is failing, your help would be appreciated.

packages/core/src/components/ChoiceList/Choice.example.jsx
```diff
@@ -20,13 +40,45 @@ ReactDOM.render(
...
+        <Choice name="radio_choice" value="b" checkedChildren={childSelect}>
+          Checkbox B - with children
+        </Choice>
+        <Choice name="checkbox_choice" value="c">
+          Checkbox C
+        </Choice>
+      </fieldset>
+    </div>
+
+    <div className="ds-u-margin-top--4">
```
1Copenut: We will want to wrap each one of these groupings in a `<fieldset>` with a proper `<legend>` and update the name attributes to match on each group. This will help prevent errors when we add an e2e test that runs axe on this component.

packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -104,7 +104,12 @@ export class Choice extends React.PureComponent {
     if (inputProps.onChange) delete inputProps.onChange;
 
     return (
-      <div className={className}>
+      <div
+        className={className}
+        aria-live={checkedChildren ? 'polite' : null}
```
1Copenut: These are working well on the following combinations:

* Win10 + IE11 + JAWS
* Win10 + IE11 + Chrome
* Win10 + Firefox + NVDA
* MacOS Mojave + Safari + VoiceOver

It's not announcing properly on iOS + Safari + VoiceOver. That's not necessarily a failure point, but something to be aware of.

packages/core/src/components/ChoiceList/Choice.example.jsx
```diff
@@ -1,7 +1,27 @@
...
+  { label: 'D', value: 'D' },
+  { label: 'E', value: 'E' },
+  { label: 'F', value: 'F' },
+  { label: 'G', value: 'G' },
+  { label: 'H', value: 'H' }
+];
+
+const childSelect = (
+  <ChoiceList
+    choices={selectChoices}
```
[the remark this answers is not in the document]
line47: It looks like classes are being applied already to the select -are there other classes that should be added?
screen shot 2019-02-11 at 4 49 29 pm

line47:
Thanks for the review @1Copenut - pushed some updates

packages/core/src/components/ChoiceList/Choice.example.jsx
line47: Updated

packages/core/src/components/MonthPicker/MonthPicker.test.jsx
```diff
@@ -237,7 +237,7 @@ describe('MonthPicker', () => {
 
   it('renders a snapshot', () => {
     const tree = renderer
-      .create(<MonthPicker name="months" label="Months" />)
+      .create(<MonthPicker name="months" label="Months" locale="en" />)
```
line47: Do you think we should do follow on work for other tests?

pwolfert: If there are other components that use the `toLocaleString` function, then yes. You can audit it by doing a search for that string. I'm fairly certain we're not using it anywhere else

line47: Thanks @pwolfert it looks like Mask is also using `toLocaleString` https://github.com/CMSgov/design-system/search?q=toLocaleString&unscoped_q=toLocaleString should we change there too?

line47:
@1Copenut can you take another look at this, I've made some updates and think we are in good shape.

packages/core/src/components/MonthPicker/MonthPicker.test.jsx
pwolfert: Oh, and it's hard-coded. This ones a bit tricker. I think we should just circle back to it when it becomes a problem. It's not an actual issue in the real world—just in testing.

packages/core/src/components/ChoiceList/Choice.example.jsx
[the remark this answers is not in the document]
1Copenut: Going to nix this comment and issue the CSS updates as a separate PR per our pair-code discussion.

PR #330, opened 2019-02-15
[WNMGDS-6/checked-child-css] Adding CSS support for Choice checked children

1Copenut:
### UPDATE
* Fixed the CSS and class placement
* Verified the inverse theme is working correctly
* Updated docs to call out `checkedChild` handling

### Added
* CSS support for the `<Choice />` checked children prop to match the older Healthcare.gov site package expose within pattern
* Documentation for proper use of the expose within pattern

---
screen shot 2019-02-18 at 2 20 41 pm

### DIFF
--- a/packages/core/src/components/ChoiceList/Choice.scss
+++ b/packages/core/src/components/ChoiceList/Choice.scss
@@ -172,6 +172,22 @@ Style guide: components.choice.inversed
   border-radius: 100%;
 }
 
+// Checked child container
+$ds-c-inset-border-width: $spacer-half;
+
+.ds-c-choice__checkedChild {
+  border-left: $ds-c-inset-border-width solid $color-primary;
+  margin-bottom: $spacer-2;
+  margin-left: ($choice-size / 2) - ($ds-c-inset-border-width / 2);
+  padding: $spacer-2;
+}
+
+.ds-c-choice--inverse {
+  .ds-c-choice__checkedChild {
+    border-left-color: $color-white;
+  }
+}
+
 /*
 Right-to-Left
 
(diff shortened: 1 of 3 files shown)


packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -316,6 +333,19 @@ Style guide: components.choice.react
 
 **[View the "Forms" guidelines for additional guidance and best practices.]({{root}}/guidelines/forms/)**
 
+## Checked children and the expose within pattern
```
line47: This is great info though it seems a bit disjointed from the props on the usage tab. Is there a way to link to the content from the props?

1Copenut: I don't believe so reliably, no. We're using basic anchor links `#usage` and `#guidance` to set the tabs correctly, but without any routing logic. I'd need to have an ID reference for the heading to deep-link to it properly. 

I can add a mention in the `checkedChild` props to see this section in the Guidance tab for further instructions.

line47:
Reverified that this looks good, thanks @1Copenut 🚀

packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -82,12 +82,26 @@ Markup:
...
 // create our own custom version
 .ds-c-choice {
   left: -999em;
   opacity: 0;
   position: absolute;
+
+  // Checked children container
+  &__checkedChild {
+    border-left: $ds-c-inset-border-width solid $color-primary;
+    margin-bottom: $spacer-2;
```
bernardwang: I noticed the bottom margin is 16px while the margin on the top was 8px. Just want to double check this is the intended design

1Copenut: Great question. This margin matches up with the HCGov site package [exposed within](https://github.cms.gov/pages/CMS-WDS/ds-healthcare-gov/patterns/exposed-within/) pattern. I matched the two up with screenshots, so this is the standard we want to style toward.

bernardwang:
This is definitely out of scope for this PR, but it looks like `Frame.jsx` doesn't handle the `checkedChildren` example [that well](http://recordit.co/KOrNGbwrn0). It's fixed when you resize the page, but it might be worth fixing in the future

1Copenut:
@bernardwang Agreed. It'd be good to fix that in the near term. I'll log an issue with reference to your screen cap. I noticed this late too, after I'd pushed the PR. I usually work off the new tab, so I can just see the markup I want, and evaluate that in a silo.

Issue #337, opened 2019-02-19
Fix design for checkbox/radio checked children examples


line47:
## Description
@bernardwang reported this design issue when using the checked children on small checkboxes.
The expanding line does not line up. 

<img width="360" alt="slack-imgs" src="https://user-images.githubusercontent.com/1449852/53049470-29415700-3465-11e9-8219-45feadad5a03.png">

## To do
- [ ] Update design so small checkbox and radio checked children examples line up properly

jenniferthibault:
Is this checkbox that's followed by a field (pending the user's action) a component somewhere that's part of the design system? @hbillings and I have been looking for this exact interaction pattern! 

Would appreciate if you could point us to it in the design system, or to @bernardwang's use of it if it's in an open source project.

bernardwang:
@jenniferthibault Examples for it were just added! You can find it by searching for `checkedChildren` [here](https://design.cms.gov/components/choice/). 

I believe this functionality was around for a little while, but recently it was worked on to improve a11y, standardize stylings, and some other improvements!

jenniferthibault:
This just made my day!

I'm still relatively new to using the system, so thanks for your patience in pointing it out to me. One thing that might make it easier for other folks to spot in the future might be to display that option in the documentation page as selected by default, so the state with the child field is displayed by default?

Thanks again!

bernardwang:
@jenniferthibault Funny that you say that, `1.31.1` which was just published 45 min ago has that change in the docs, its now selected by default. You ask and you shall receive 😆

jenniferthibault:
:HAPPY-DANCE:!!!

Issue #333, opened 2019-02-18
Update the Choice example frame to handle expanding content on design.cms.gov


1Copenut:
## Description
[Release 1.31.0](https://github.com/CMSgov/design-system/releases/tag/v1.31.0) includes a new pattern that expands the `<Choice />` component to reveal nested elements. The frame around this content has to be resized to see the entire block as shown in this screen capture: http://recordit.co/KOrNGbwrn0

1Copenut:
Adding a prop `defaultChecked` to the `<Choice />` components with children that are exposed on check forces the frame to be the correct height on load. If users uncheck the nested inputs, the frame will be too tall until the page is resized, but this seems less an issue.

Will get a fix in for this tomorrow.

PR #341, opened 2019-02-20
Adding small and right inputs, fixing frame height on Choice.

1Copenut:
### Added
* CSS for proper handling of checked children with small inputs
* CSS for checked children with right to left order
* `<Choice />` component examples to support the previous two items
* Documentation for using these examples

### Changed
* Fixed the choice frame example by setting all children open with `defaultChecked`

### Design considerations
I made a few judgment calls on the inputs to the right of labels, with regard to padding and margin. Please advise if these need tweaked. The full-width didn't feel right, so I capped width, and sight-adjusted, dipping back into UI design for a moment.

### Related Issues
* Closes https://github.com/CMSgov/design-system/issues/333
* Closes https://github.com/CMSgov/design-system/issues/337

---
screen shot 2019-02-21 at 9 03 59 am

### DIFF
--- a/packages/core/src/components/ChoiceList/Choice.example.jsx
+++ b/packages/core/src/components/ChoiceList/Choice.example.jsx
@@ -26,7 +26,7 @@ const childSelect = (
 ReactDOM.render(
   <div>
     <fieldset className="ds-c-fieldset">
-      <legend className="ds-c-label">Checkbox example</legend>
+      <legend className="ds-c-label">Checkboxes</legend>
       <Choice
         defaultChecked
         hint="Checkbox A hint"
@@ -44,16 +44,12 @@ ReactDOM.render(
     </fieldset>
 
     <fieldset className="ds-c-fieldset">
-      <legend className="ds-c-label">Checkbox example with children</legend>
-      <Choice
-        hint="Checkbox A hint"
-        defaultChecked
-        name="checkbox_choice_children"
-        value="a"
-      >
+      <legend className="ds-c-label">Checkboxes with children</legend>
+      <Choice hint="Checkbox A hint" name="checkbox_choice_children" value="a">
         Checkbox A
       </Choice>
       <Choice
+        defaultChecked
         name="checkbox_choice_children"
         value="b"
         checkedChildren={
@@ -67,9 +63,11 @@ ReactDOM.render(
       </Choice>
     </fieldset>
 
+    <hr className="ds-u-margin-top--2" />
+
     <fieldset className="ds-c-fieldset">
-      <legend className="ds-c-label">Radio example</legend>
-      <Choice name="radio_choice" type="radio" value="a">
+      <legend className="ds-c-label">Radio buttons</legend>
+      <Choice defaultChecked name="radio_choice" type="radio" value="a">
         Radio A
       </Choice>
       <Choice name="radio_choice" type="radio" value="b">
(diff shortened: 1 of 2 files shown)


line47:
I'm wondering if we should have that`Right checkboxes with children` example. That seems like a pattern we would not want to promote and maybe even discourage. 

Taking that one step further, the should we have a `Right-to-Left` option for checkboxes at all? Again this seems like it makes the checkboxes harder to use.

1Copenut:
@line47 I was wondering that myself. I'm not thrilled with that inputs to the right pattern, but feel we need to account for it, given it's been in the design system for some time, maybe since the beginning.

Now, we could recommend users **not** include `checkedChildren` when using that pattern and remove the styling overrides for it, but I think that has risk too.

melwoodard:
Agreed with @line47. At minimum I'd prefer to exclude the checked children functionality from the `Right-to-Left` checkbox pattern. We should also discuss (in a new issue) whether the `Right-to-Left` pattern is something we want to include at all, and what steps we may need to take to properly deprecate it in case it's being used somewhere.

1Copenut:
Issue's posted for discussing deprecation. References this PR for context. I'll update the PR and docs later today.

packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -98,9 +98,29 @@ $ds-c-inset-border-width: $spacer-half;
     margin-left: ($choice-size / 2) - ($ds-c-inset-border-width / 2);
     padding: $spacer-2;
 
+    // Checked children container on dark background
     &--inverse {
       border-left-color: $color-white;
     }
+
+    // Inputs to the right of labels
```
line47: Recommend removing styling for right aligned checked children

packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -98,9 +98,29 @@ $ds-c-inset-border-width: $spacer-half;
...
+    &--right {
+      border-left: 0;
+      border-right: $ds-c-inset-border-width solid $color-primary;
+      margin-left: 0;
+      margin-right: ($choice-size / 2) - ($ds-c-inset-border-width / 2);
+      padding: $spacer-1 0 $spacer-1 ($spacer-2 * 0.75);
+    }
+
+    // Max recommended parent width is $ds-u-measure--narrow (12em)
```
line47: Recommend removing styling for width on right aligned checked children

packages/core/src/components/ChoiceList/Choice.example.jsx
```diff
@@ -94,15 +93,74 @@ ReactDOM.render(
...
+        size="small"
+      >
+        Checkbox B - with children
+      </Choice>
+      <Choice name="checkbox_choice_children_small" size="small" value="c">
+        Checkbox C
+      </Choice>
+    </fieldset>
+
+    <div className="ds-u-measure--narrow ds-u-margin-top--4 ds-u-margin-bottom--5 ds-u-padding--2 ds-u-border--2">
```
line47: Suggest we remove this Right checkboxes with children

packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -334,21 +340,23 @@ Style guide: components.choice.react
 
 - The `<Choice>` component includes a `checkedChildren` prop that can expose hidden text information or form elements. This **expose within** pattern is especially useful if you need to collect data from follow up questions or give just-in-time feedback.
 - Checked children can be exposed by checking the parent checkbox or radio button
-- Checked children will be announced to screen readers when they are exposed. They have been tested with the following devices:
-  - Windows 10 + Internet Explorer 11 + JAWS screen reader
-  - Windows 10 + Chrome + JAWS
-  - Windows 10 + Firefox + NVDA
-  - MacOS Mojave + Safari + VoiceOver
 - The checkedChildren prop should return one or more items wrapped in a `div` with the following className: `ds-c-choice__checkedChild`. This class sets the spacing and border color for the exposed elements.
```
bernardwang: Small content nitpick, I noticed other instances of `checkedChildren` are styled as code, should this one be too?

1Copenut: Yes it should. I'll update now.

PR #345, opened 2019-02-22
Fixing the broken internal links

1Copenut:
### Changed
* All instances of `{{root}}` were removed from SCSS files. This had the effect of creating local URLs instead of having that string encoded for HTML

### Tested
* Verified each link resolved on localhost by opening and closing in turn


line47:
There is a link on the guidelines/themes page that is broken ` import the design system's Sass files` is going to `/getting-started/` which is not a page

1Copenut:
@line47 Changes are in, should be able to pull the branch locally and double-verify. I added an ID to the Sass file to drop directly to the right heading.

PR #359, opened 2019-02-27
[WIP] Consider moving article, form, and responsive examples into the design system

line47:
These don't seem to be used for anything so I think they are safe to remove

### Article example
screen shot 2019-02-27 at 3 41 00 pm

### Form example
screen shot 2019-02-27 at 3 41 10 pm

### Responsive example
screen shot 2019-02-27 at 3 41 20 pm


phiden:
I'm not sure we can say this isn't being used for anything -- would want to check in with consumers of the system to make sure it's not somehow a reference. We should probably at some point talk about deprecation procedures (meaning developing them; we don't have any that I'm aware of) -- another thing or two for the roadmap!

line47:
Should examples like this be part of the design system site/docs?
I'm unsure how helpful these particular examples are sitting where they are in the repo structure 🤷‍♂️

melwoodard:
Huh, didn't know these were here. I think these are useful, the form example in particular is very much in play in App3. That said, if they don't have representation on design.cms.gov, I doubt people are finding them.

line47:
## Article example
I think we could move the article example under patterns - similarly to the documentation page. 
Though I don't know how much it would get used or referenced, as it's mostly headings and paragraphs.

## Form example
I'm thinking this could be updated and moved to the patterns section and we should do something similar to https://v2.designsystem.digital.gov/components/form-templates/ the `form templates` here and set up templates that we are actually using in applications instead of this generic form.

## Responsive example
I think we should create a new section for responsive and include an example similar to this. Right now we have a responsive design guideline page which feels out of place. We also have a Display/Visibility utility page which also includes responsive information. Maybe it would fit better under getting started or adding a utility page about Responsive options?

line47:
@phiden I'd be curious about your thoughts here?

phiden:
LGTM. For what it's worth, there's a way to demonstrate responsive examples in line with each component...but of course _how_ to do that is not super clear!

line47:
Are you referring to the responsive grid in this screenshot?
Screen Shot 2019-03-28 at 4 12 12 PM

phiden:
That’s the one!

line47:
@bernardwang do you have thoughts here?

PR #382, opened 2019-03-01
Add Propose new item issue template

line47:
Adding template for proposing a new item to the design system.
The template is geared to be simple and not offer too many restrictions for folks that want to submit an idea for a new item.


line47:
Presented this to the CMSWDS Design Check in today and got 👍 from folks. 
Can you review this so we can get it published?

line47:
☝️ @phiden

PR #389, opened 2019-03-19
Update WRITING-DOCUMENTATION.md

line47:
Reviewing and updating how to add and document components in the CMS Design system.

### Added
- Required to the sections of documentation that are required
- More details and examples to the `Flags` section
- More detailed examples
- Explanation of what the weight front matter property does

### Changed
- Header structure to better fit the flow of the page.
- Statues to reflect the component maturity - Draft, Work in progress, Ready, and Deprecated

### Removed
- Extra wording that was not helping with context or clarification
- EJS example description as that is not in use for any examples of the design system.
- Inline example HTML recommendation in favor of using HTML example files. This will lead to cleaner example files and allow them to be checked easier against linters, code inspectors, and evaluation tools.


guides/WRITING-DOCUMENTATION.md
```diff
@@ -15,176 +15,147 @@ For a component named "Foo", you would have the following:
...
+        └── Foo.example.html    HTML example
 ```
 
-# CSS components
+## CSS components
 
 [KSS](https://github.com/kss-node/kss-node) is the primary way documentation is written. They establish what pages get created and the content on those pages. We use KSS to parse CSS comment blocks to extract documentation, and it expects your comment block to follow this particular format:
 
-```css
+```scss
```
phiden: why change this to scss?

guides/WRITING-DOCUMENTATION.md
```diff
@@ -15,176 +15,147 @@ For a component named "Foo", you would have the following:
...
 /*
 [Heading]
 
 [Description]
 
-[@flags]
-
-Markup:
-[Inline HTML/EJS or filename]
+[Flags]
```
phiden: I'd keep this as it was -- the `@` is what triggers the KSS

guides/WRITING-DOCUMENTATION.md
```diff
@@ -15,176 +15,147 @@ For a component named "Foo", you would have the following:
...
 
 [Modifiers]
 
+Markup: [filename]
```
phiden: the way the system builds now, all the markup is inline. I'd keep this as it was unless we intend to rewrite all our files to follow filename-only conventions.

guides/WRITING-DOCUMENTATION.md
```diff
@@ -15,176 +15,147 @@ For a component named "Foo", you would have the following:
...
+
 Style guide: [Reference]
+
+[Documentation and guidance]
```
phiden: is KSS looking for this?

guides/WRITING-DOCUMENTATION.md
```diff
@@ -15,176 +15,147 @@ For a component named "Foo", you would have the following:
...
 
-This is a required field, but if you want to create a subsection without a heading, you can enter `---`.
+**note:** If you want to create a subsection without a heading, add `---` after the opening `/*`.
+
+### Description
+
+The description should describe what pattern does or should be used for in plain language.
 
-### @flags
+### Flags
```
phiden: I know the `@flags` looks weird, but keeping that formatting consistent reduces confusion that it could be some other kind of flag. QPP for example uses "flags" to turn features on and off.

guides/WRITING-DOCUMENTATION.md
```diff
@@ -15,176 +15,147 @@ For a component named "Foo", you would have the following:
...
-<summary>Inline HTML example</summary>
-
-```
-Markup:
-<button class="ds-c-button {{modifier}}">Button label</button>
-```
-</details>
+If the item you are documenting has multiple states or styles depending on added classes or pseudo-classes, you should document them in the modifiers section. [More info on modifiers from the KSS documentation site](https://warpspire.com/kss/syntax/).
 
+### Markup
```
phiden: In theory I support this change, but I'd keep the original in the docs as well until we have time (or the will?) to rewrite the way our own system looks. Walk our talk, etc.

guides/WRITING-DOCUMENTATION.md
```diff
@@ -15,176 +15,147 @@ For a component named "Foo", you would have the following:
...
-
-- List any related components or utility classes that
-  connect with the component, whether as a parent, sibling,
-  or child within the pattern's taxonomy.
-
-## Learn more
-
-- List any links that served as inspiration, references,
-  or research related to this component.
+Content here
```
phiden: ?

guides/WRITING-DOCUMENTATION.md
```diff
@@ -234,25 +208,23 @@ If a React component expects any `props`, you should document them within the co
 Button.propTypes = {
   label: React.PropTypes.node.isRequired,
   /**
-   * Overwrite the button's `class` attribute by providing your own
-   */
+  Overwrite the button's `class` attribute by providing your own
```
phiden: I think React will choke on this formatting? -- it's weirdly specific sometimes, but I can't remember off the top of my head if it requires that `*` in the comment block

guides/WRITING-DOCUMENTATION.md
line47: I chose SCSS as the files we use in the design system are `.SCSS`

guides/WRITING-DOCUMENTATION.md
line47: I was thinking it would be easier to validate and view as a separate file with just the HTML instead of inlining the markup. Maybe this could be some follow on work to extract the inline code to separate files?

guides/WRITING-DOCUMENTATION.md
line47: When documenting a component there are 2 pieces that are needed one is `Style guide: components.alert.guidance`

The other is the actual guidance content which is what the `[Documentation and guidance]` is supposed to represent here.

guides/WRITING-DOCUMENTATION.md
line47: That makes sense - I'll add a follow on issue for that work and update the language until we are ready.

guides/WRITING-DOCUMENTATION.md
line47: I added documentation guidance below and pulled it out of the code block. We can add it back if you think that makes more sense. I was thinking it would be easier to understand as guidance to follow outside of the code block.

line47:
Thanks for the review 🎉

guides/WRITING-DOCUMENTATION.md
phiden: Ok. Maybe add a comment to that effect inline -- it's pretty unclear

guides/WRITING-DOCUMENTATION.md
phiden: Let's find out!

PR #277, opened 2018-09-12
Begins downshift 2.2.1 upgrade - in progress

savannahmillion:
### Added
This PR doesn't add anything new.

### Changed
Downshift gets upgraded to 2.2.1. The PR is currently unfinished because the Downshift upgrade dropped the prop `render` and only supports `children`, breaking some of the existing syntax.

### Deprecated
This PR doesn't deprecate anything.

### Removed
This PR doesn't remove anything.

### Fixed
This PR (currently) doesn't fix any bugs.

PR #407, opened 2019-04-05
[WNMGDS-66] Updating Autocomplete to use Downshift v3.

1Copenut:
### Updates
This PR is ready for review. After careful consideration, I have built the autocomplete to the legacy [ARIA 1.0 combobox](https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html) with list autocomplete example. The new 1.1 version that renders a containing div with `role="combobox"` was causing some confusing and unexpected behaviors with multiple screen readers. I felt this was a serious enough regression to stick with the previous markup pattern. See [Differences between ARIA 1.0 and 1.1: Changes to role=”combobox”](https://www.levelaccess.com/differences-aria-1-0-1-1-changes-rolecombobox/) for an in-depth list of the differences.

**More additions:**
* Changed the first child returned by Downshift from a `<div>` to a custom component and overrode the default ARIA using `getRootProps()` This was done to prevent Downshift from assigning attributes I did not want added. See [How to give rendering control to users with prop getters](https://kentcdodds.com/blog/how-to-give-rendering-control-to-users-with-prop-getters) for more explanation of this technique.
* Added unit tests for removing the Clear search button with opt-in `noClearInput` prop
* Added e2e tests to mock mouse and keyboard interaction with the rendered component

**A11y testing:**
* Confirmed existing behavior in JAWS, NVDA, and VoiceOver in their preferred browsers
* Confirmed existing keyboard navigation behavior
* Confirmed zero aXe errors in browser testing and e2e test suite

### Changed
* Bumped [Downshift](https://github.com/downshift-js/downshift) to v 3.2.7
* Refactored our `Autocomplete` to use Downshift with render props
* Added a label ID to the `TextField` as an optional prop for Downshift's `aria-labelledby` pointers

I'd like the team to review when they've got a second. This is a WIP, draft PR. I need to test this in IE11 and other Windows browsers, as well as screen readers, before merging.

closes https://github.com/CMSgov/design-system/pull/277


packages/core/src/components/TextField/TextField.jsx
```diff
@@ -201,6 +204,10 @@ TextField.propTypes = {
    * Additional classes to be added to the label
    */
   labelClassName: PropTypes.string,
+  /**
+   * A unique `id` to be used on the label field.
+   */
+  labelId: PropTypes.string,
```
pwolfert: 👍

pwolfert:
Linking this so we remember to close it if this gets merged: https://github.com/CMSgov/design-system/pull/277

1Copenut:
@pwolfert @rdimartino 
I've refactored this quite a bit and gone back to version 1.0 of the ARIA specification. The combined combobox was causing all kinds of problems with screen readers, and the previous version of the spec is still valid. See https://www.levelaccess.com/differences-aria-1-0-1-1-changes-rolecombobox/ for more details on that front.

I do need a bit of help to get the custom loading message working correctly with screen readers. There's an advanced prop in Downshift called `getA11yStatusMessage` but I'm not sure how to implement it. Direct link here: https://github.com/downshift-js/downshift#geta11ystatusmessage

I'll follow up later today or tomorrow on this item.

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -101,42 +136,63 @@ export class Autocomplete extends React.PureComponent {
...
           clearSelection,
           getInputProps,
           getItemProps,
+          getRootProps,
           highlightedIndex,
           inputValue,
           isOpen
         }) => (
-          <div className={rootClassName}>
+          <ComponentDiv
```
pwolfert: I don't entirely understand what's going on here, but would it be possible to put these props directly on an inline `<div ...` declaration rather than creating the separate `ComponentDiv` function above? If it's a matter of pulling out the `innerRef` property, we could change this render function a bit to do something like
```js
(stuff) => {
  const rootProps = getRootProps({ ...blahblahblah });
  return (
    <div {...rootProps} ref={rootProps.innerRef}>
    </div>
  );
}
```
If what `refKey: 'innerRef'` is doing is telling Downshift what to call the property, could we say `refKey: 'ref'` and not have to do anything special to apply it to the `ref` prop?

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -237,6 +298,10 @@ Autocomplete.propTypes = {
    * Message users will see when the `loading` prop is passed to `Autocomplete`.
    */
   loadingMessage: PropTypes.node,
+  /**
+   * Do not render the Clear search `<button>` when prop is passed
+   */
+  noClearInput: PropTypes.bool,
```
pwolfert: An alternative to prefixing with `no` could be to make it an optional prop that has a defaultValue of true. That works too 👍

1Copenut: I like this much better. I changed it to `clearSearchButton` and added a default `true`. Opting out by passing `clearSearchButton={false}` feels more direct and clear.

packages/core/src/components/Autocomplete/Autocomplete.jsx
1Copenut: Downshift looks for a `<div>` as the container element, and applies a lot of ARIA attributes to make the finished component fall in-line with the ARIA 1.1 specification for comboboxes. I discovered in screen reader testing this is unpredictable and introduces a number of challenges for usability. 

I opted to use the previous pattern ( https://www.w3.org/TR/wai-aria-practices-1.1/examples/combobox/aria1.0pattern/combobox-autocomplete-list.html ) and needed a way to suppress those ARIA attributes. Using `getRootProps` looked like the best way, and I modeled the `<MyDiv />` "component" after prior art on the Downshift issues page. ( https://github.com/downshift-js/downshift/issues/235)

line47:
What was your questions here @pwolfert and did you get it answered?

pwolfert:
@line47, yeah, it was the first resolved conversation on my review. We're good here. I know Trevor wanted to do some more testing on it, though, before merging.

1Copenut:
@pwolfert @line47 
I've added a check that uses Selenium's `getAttribute()` method to return `outerHTML` strings for comparison: https://seleniumhq.github.io/selenium/docs/api/javascript/module/selenium-webdriver/lib/webdriver_exports_WebElement.html#getAttribute

Would appreciate eyes on the lines using this to make sure that's a good way to test that an element has been removed from the DOM:
* https://github.com/CMSgov/design-system/blob/8c7d7f5facb642c159af8826a1e8fb64fb037157/packages/core/src/components/Autocomplete/Autocomplete.e2e.test.js#L98
* https://github.com/CMSgov/design-system/blob/8c7d7f5facb642c159af8826a1e8fb64fb037157/packages/core/src/components/Autocomplete/Autocomplete.e2e.test.js#L106

packages/core/src/components/Autocomplete/Autocomplete.e2e.test.js
```diff
@@ -1,19 +1,113 @@
...
+    expect(el).toEqual('');
+  });
+
+  it('Closes the listbox when ESC is pressed', async() => {
+    await driver.get(rootURL);
+
+    let container = await getElementByXPath(
+      '//*[@id="js-example"]/div/div[1]/div'
+    );
+    container = await container.getAttribute('outerHTML');
```
pwolfert: To your question about using the outerHTML to test the presence of the element, I think it would be better to use `findElements`. Something like

```js
const autocompleteField = await getElementByXPath('//*[@id="autocomplete_1"]');

// Make sure the list shows when we start typing
autocompleteField.click();
await autocompleteField.sendKeys('c');
let listbox = await driver.findElements(by.css('the-listbox-selector'))
expect(listbox.length).toEqual(1);

// Make sure it disappears when we hit escape
await autocompleteField.sendKeys(key.ESCAPE);
listbox = await driver.findElements(by.css('the-listbox-selector'))
expect(listbox.length).toEqual(0);
```

pwolfert: I also think it would make the tests easier to follow if you use the variable name to describe what the element is rather than reusing the `el` variable.

1Copenut: @pwolfert I took your advice and named the test `elem` more specifically. They read a lot better now.

I also found a way to assert keyboard focus effectively. It could use for a refactor to a helper function, but I'll do that as an enhancement since `Autocomplete` is the only one doing this style of testing atm.

pwolfert: They do read a lot easier 👍 LGTM

PR #420, opened 2019-05-15
Updating the definition of site package and theme

line47:
Update the definition of site package to be more clear about what a site package is and add a definition for what a theme is.


line47:
@phiden can you take a look at this?

PR #422, opened 2019-05-20
[WNMGDS-81] Updating measure variables to use ex instead of em.

line47:
### Added

### Changed
- Changed the variables for `base`, `narrow` and `wide` from `em` values to `ex`. `Ex` is a better path than `ch` which was what was originally proposed. 
  - https://www.w3schools.com/cssref/css_units.asp
  - https://v2.designsystem.digital.gov/design-tokens/typesetting/measure/
- Updated the measure utility to have a separate HTML example and it's own folder (similar to components) There could be some follow on work to placing utilities inside folders to follow a consistent approach. 
- Showing each example `base`, `narrow` and `wide` in its own code snippet instead of combined. This follows the convention of the components which I think is stronger than grouping everything into one code snippet box. This also shows the modifier and seems easier to read and get information about the style you are looking for without having to comb through the code box. 

## Screenshot before
Screen Shot 2019-05-20 at 12 21 36 PM

## Screenshot after
Screen Shot 2019-05-20 at 12 21 48 PM

### DIFF
--- a/packages/core/src/utilities/index.scss
+++ b/packages/core/src/utilities/index.scss
@@ -29,7 +29,7 @@ Style guide: utilities
 @import 'font-weight';
 @import 'line-height';
 @import 'margin';
-@import 'measure';
+@import 'measure/measure';
 @import 'overflow';
 @import 'padding';
 @import 'text-align';
--- a/packages/core/src/utilities/measure/measure.example.html
+++ b/packages/core/src/utilities/measure/measure.example.html
@@ -1,0 +1,1 @@
+<div class="{{modifier}}">{{ lorem-l }}</div>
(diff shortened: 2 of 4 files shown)


line47:
@phiden This PR is only addressing measure and not font size.

I'm not sure what you mean by consistent, are you referring to the font-size variables here or something else? https://github.com/CMSgov/design-system/blob/master/packages/support/src/settings/_override.uswds.scss

phiden:
@line47 yeah -- they're measured in px / em / rem; it seems like we should use a consistent unit of measure across the board. That is, if `measure` is in `ex`, font sizes should also be given in `ex`. Thinking about possible confusion for future users.

line47:
Currently updating the measure from `em` to `ex` results in using a length that is a better estimate for the length of measure.

I think there could definitely be follow on work to look into the best approach for font sizing. We could add something to the backlog to define that work more explicitly.

line47:
@phiden - I'd like to merge this work and add a backlog item for the font consistency you mentioned. Does that work for you?

PR #425, opened 2019-05-29
Give solid buttons transparent borders rather than no borders

pwolfert:
### Changed
- Solid-colored buttons now have transparent borders (which fills the space with the background color) rather than having a border width of 0, which makes those buttons smaller than the default buttons. 
  size-difference


line47:
This looks good to me and keeps the ways buttons are built more consistent, thanks for this add!

PR #430, opened 2019-06-07
Fix vulnerabilities for tunnel-agent, js-yaml, and handlebars

line47:
### Fixed
Upgrading to appropriate versions of tunnel-agent, js-yaml, and handlebars to fix security vulnerabilities.


line47:
I'm no longer seeing the vulnerabilities reported under the security alerts and am wondering if we can close this PR?  https://github.com/CMSgov/design-system/network/alerts

pwolfert:
The only problem I see with doing it this way is that adding them to the dependency list could create confusion for us down the line as to which dependencies we're actually using, like if we get rid of one of the packages that requires one of these. I like the dependencies in package.json to be just the direct dependencies. Does it work if you revert the changes in `package.json` but keep the `yarn.lock` changes?

line47:
Seems to be OK to go that route @pwolfert care to take another look?

pwolfert:
When I `yarn install`, it tries to clear out those entries 😞 
I'm looking at the way Dependabot did it, and it looks like the secret might be to keep the versions in the headers in yarn.lock the same bu change the `version "x.x.x"` field below them. Are these changes that we can get the bot to do instead?

line47:
I created some automated security fixed for those - however, handlebars is no longer showing as an issue here https://github.com/CMSgov/design-system/network/alerts

bernardwang:
https://gyazo.com/3df69b506798952b31f3683ab60ec541
I'm also seeing things deleted in the yarn.lock. @line47 how did you upgrade these? was it with `yarn upgrade`?

line47:
I believe so - it's been a while since I made this PR I'll need to look back into it or maybe we close it for now?

line47:
Closing this PR and revisiting at a later date

PR #432, opened 2019-06-12
MCT1802: update aria text on autocomplete to match visual

SparrowSpit:
### Fixed
As put forth in [the new 2.1 criteria](https://www.w3.org/WAI/WCAG21/Understanding/label-in-name.html), the aria label must match or start with the same text as the visual text. Given this, I edited the aria so that the: 
- beginning of the aria label of the autocomplete component is the same as the visual label.


line47:
👋 Hey @SparrowSpit I think you need to run `yarn update-snapshots` to update the Jest snapshots they this should no longer be failing the checks. Let me know if you need help

SparrowSpit:
I ran `yarn update-snapshots` but not sure it fixed the errors. Let me know what you think @line47

line47:
Closing as #448 accomplishes this

PR #434, opened 2019-06-14
[WNMGDS-111] Create dropdown component

bernardwang:
### Added
New `<Dropdown>` component and documentation
Screen Shot 2019-06-14 at 2 14 13 PM

### Changed
- `<Dropdown>` is factored out of the `<Select>` and `<ChoiceList>` component, and modeled after other independent field components like `<TextField>`. 
- The Guidance section in the documentation has been updated to match USWDS. Most of this was the same info just reworded, but there are a few small additions and deletions.
- Tests are also updated


bernardwang:
This is still WIP, but if any of you want to take an early look feel free!

packages/core/src/components/Dropdown/Dropdown.example.html
```diff
@@ -0,0 +1,25 @@
+<label class="ds-c-label ds-u-margin-top--0" for="options">Dropdown label</label>
```
bernardwang: This is the same html example as the original one for <Select>, I just moved it to it’s own file for organization

packages/core/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,169 @@
...
+   */
+  labelClassName: PropTypes.string,
+  /**
+   * The field's `name` attribute
+   */
+  name: PropTypes.string.isRequired,
+  /**
+   * The list of options to be rendered.
+   */
+  options: PropTypes.arrayOf(
```
bernardwang: I was considering renaming this prop from “choices” to “options” to differentiate the Dropdown prop from the Choicelist one. I was thinking it might be useful to do this because the props are slightly different, and “options” relates to the actually html element used in this component. That being said, it might be better to just keep the names the same too make it easier to convert between Choicelist and Dropdown. Do you have any thoughts? @line47 @phiden

line47: I don't have strong feelings either way - I think this is smart to think about. How much simpler does it make it by not switching the name from `choices` to `options`?

bernardwang: The main impact I was thinking was if you were converting the old `ChoiceList` to the new `Dropdown`, you could avoid renaming that prop. However, you would still have to change the props like this: 
```
<ChoiceList choices={[{
   value: 1,
   label: 'first choice',
   checked: true,
}]}/>
```
becomes ->
```
<Dropdown choices={[{
   value: 1,
   label: 'first choice',
}]} value={1} />
```

The main reason why `Dropdown` has a new a `value/defaultValue` prop and has a slightly modified `choices/options` prop is because this prevents multiple options from being selected. 
I was considering keeping `choices` exactly the same and just throwing an error if multiple choices were selected, but I thought it would be better to just design the props so that only 1 selected option is allowed.

bernardwang: The main consideration I was thinking of was if you were converting the old `ChoiceList` to the new `Dropdown`, you could avoid renaming that prop. However, you would still have to change the props like this: 
```
<ChoiceList choices={[{
   value: 1,
   label: 'first choice',
   checked: true,
}]}/>
```
becomes ->
```
<Dropdown choices={[{
   value: 1,
   label: 'first choice',
}]} value={1} />
```

The main reason why `Dropdown` has a new a `value/defaultValue` prop and has a slightly modified `choices/options` prop is because this prevents multiple options from being selected. 
I was considering keeping `choices` exactly the same and just throwing an error if multiple choices were selected, but I thought it would be better to just design the props so that only 1 selected option is allowed.

bernardwang: The main consideration I was thinking of was if you were converting the old `ChoiceList` to the new `Dropdown`, you could avoid renaming that prop. However, you would still have to change the props like this: 
```
<ChoiceList choices={[{
   value: 1,
   label: 'first choice',
   checked: true,
}]}/>
```
becomes ->
```
<Dropdown choices={[{
   value: 1,
   label: 'first choice',
}]} value={1} />
```

The main reason why `Dropdown` has a new a `value/defaultValue` prop and has a slightly modified `choices/options` prop is because this prevents multiple options from being selected. 
I was considering keeping `choices` exactly the same and just throwing an error if multiple choices were selected, but I thought it would be better to just design the props so that only 1 selected option is allowed.

packages/core/src/components/Dropdown/Dropdown.scss
```diff
@@ -67,30 +42,29 @@ Style guide: components.dropdown.react
...
 
-- Use sparingly — only when a user needs to choose from about 7 to 15 possible options and you have limited space to display the options.
+- **Use sparingly**. Use the dropdown component only when a user needs to choose from about seven to 15 possible options and you have limited space to display the options.
 
 ### When to consider alternatives
 
-- If the list of options is short. Use [radio buttons or checkboxes](/components/choice) instead.
-- If the list of options is very long. Let users type the same information into a [text input](/components/text-field) that suggests possible options instead. Long lists can sometimes be broken into multiple shorter lists by asking a few questions up-front to filter out options.
-- If you need to allow users to select more than one option at once. Users often don’t understand how to select multiple items from dropdowns. Use checkboxes instead.
-- If the data is highly familiar to users, such as the day, month, or year of their birth.
```
bernardwang: This was not present in the new USWDS docs

packages/core/src/components/Dropdown/Dropdown.scss
```diff
@@ -67,30 +42,29 @@ Style guide: components.dropdown.react
...
+- **Fewer than seven options.** Use [radio buttons or checkboxes](/components/choice) instead.
+- **More than 15 options.** If the list of options is very long. Let users type the same information into a text input that suggests possible options instead.
+- **Multi-select.** If you need to allow users to select more than one option at once. Users often don’t understand how to select multiple items from dropdowns. Use [checkboxes](/components/choice) instead.
+- **Site navigation.** Use the [navigation components](/components/vertical-nav/) instead.
 
 ## Guidance
 
-- Choose a sensible order for the options to enable visual scanning of the lists.
```
bernardwang: same with this one

bernardwang:
Just FYI, this PR should be ready now!

packages/core/src/components/Dropdown/Dropdown.example.jsx
```diff
@@ -0,0 +1,48 @@
+import Dropdown from './Dropdown';
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+const dropdownOptions = [
+  { label: 'A', value: 'A' },
```
line47: I'd like to change these to `Option 1`, `Option 2`to keep the examples consistent

packages/core/src/components/Dropdown/Dropdown.jsx
line47: That sounds great, thanks!

line47:
I'm wondering about the Checkboxes with children and radios with children examples on the Choice page https://design.cms.gov/components/choice/

Could we still accomplish something similar with the new dropdown?

bernardwang:
@line47 I was thinking about the checked children and I think it might not make as much sense on the `<Dropdown>`. Unlike `<Choice>` which has multiple radio/checkboxes, in `<Dropdown>` there is always the single `select` as the only parent element. To me, this reduces the value of the expose within pattern on `<Dropdown>` as you can just hide/unhide an additional question depending on the selected value instead of "nesting" another question.

line47:
@bernardwang I was thinking in the context of this example. https://github.com/CMSgov/design-system/blob/master/packages/core/src/components/ChoiceList/Choice.example.jsx#L51 Would we be able to use the dropdown component where `childSelect` is used?

line47:
It looks like using the dropdown in the `Choice.example.jsx` file works just fine 👍 Not sure if we want to update the examples to use the dropdown or not?

bernardwang:
Ah I see what you mean, just swapping out `<Select>` with `<Dropdown>` in the `<Choicelist>` checkedChildren example. I can do that!

packages/core/src/components/Dropdown/Dropdown.jsx
1Copenut: I concur with @line47 about the prop name change. `options` is a clear prop to me, esp. since the rendered HTML is a series of `<option>` tags and we want to prevent multiple selection.

bernardwang:
@1Copenut This should be ready for a 508 audit!

line47:
I think this is ready for some testing @1Copenut 🎉 
Also @pwolfert your eyes would be helpful here too

packages/core/src/components/ChoiceList/Choice.example.jsx
```diff
@@ -14,12 +14,13 @@ const selectChoices = [
   { label: 'H', value: 'H' }
 ];
 
-const childSelect = (
-  <ChoiceList
-    choices={selectChoices}
-    label="Select example"
+const childDropdown = (
+  <Dropdown
```
pwolfert: 👌

packages/core/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,169 @@
...
+      // Cache the ID so we're not regenerating it on each method call
+      this._id = uniqueId(`select_${this.props.name}_`);
+    }
+    return this._id;
+  }
+
+  render() {
+    /* eslint-disable prefer-const */
+    let {
+      // Using let rather than const since we sometimes rewrite id
```
pwolfert: Looks like this doesn't apply anymore. I don't see `id` in this list or any of these variables being mutated.

packages/core/src/components/Dropdown/Dropdown.jsx
```diff
@@ -0,0 +1,169 @@
...
+          requirementLabel={requirementLabel}
+          inversed={inversed}
+        >
+          {label}
+        </FormLabel>
+        <select
+          className={fieldClasses}
+          id={this.id()}
+          /* eslint-disable no-return-assign */
+          ref={focusTrigger ? loader => (this.loader = loader) : fieldRef}
```
pwolfert: Could it be a problem to make focus management and letting the parent specify a ref function be mutually exclusive? I think it might cause unexpected behavior for the user who expects his or her `fieldRef` function to be executed. Something like
```js
ref => {
  if (focusTrigger) {
    this.selectRef = ref;
  }
  this.props.fieldRef(ref);
}
```

packages/core/src/components/Dropdown/Dropdown.test.jsx
```diff
@@ -0,0 +1,181 @@
...
+}
+
+describe('Dropdown', () => {
+  it('renders a select menu', () => {
+    const data = render();
+
+    expect(data.wrapper.find('FormLabel').length).toBe(1);
+    expect(data.wrapper.find('select').length).toBe(1);
+    expect(data.wrapper.find('select').children('option').length).toBe(1);
+    expect(data.wrapper).toMatchSnapshot();
```
pwolfert: Might be good to have one of these assign a `value` or `defaultValue` so we can see how that looks on the final output.

packages/core/src/components/Dropdown/Dropdown.jsx
1Copenut: I agree. This looks like a better approach on the whole. I wrote the ternary logic nearly a year ago to fix a specific problem. Now that the design system is gaining wider adoption, defensive patterns like this are more sustainable.

line47:
@bernardwang there are some comments Patrick left can you take a look?

packages/core/src/components/Dropdown/Dropdown.scss
```diff
@@ -0,0 +1,82 @@
...
+- **Fewer than seven options.** Use [radio buttons or checkboxes](/components/choice) instead.
+- **More than 15 options.** If the list of options is very long. Let users type the same information into a text input that suggests possible options instead.
+- **Multi-select.** If you need to allow users to select more than one option at once. Users often don’t understand how to select multiple items from dropdowns. Use [checkboxes](/components/choice) instead.
+- **Site navigation.** Use the [navigation components](/components/vertical-nav/) instead.
+
+## Guidance
+
+- **Make sure to test.** Test dropdowns thoroughly with members of your target audience. Several usability experts suggest they should be the “UI of last resort.” Many users find them confusing and difficult to use.
+- **Avoid dependent options.** Avoid making options in one dropdown menu change based on the input to another. Users often don’t understand how selecting an item in one impacts another.
+- **Use a good default.** When most users will (or should) pick a particular option, make it the default: ``<option selected="selected">Default</option>``
```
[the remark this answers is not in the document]
bernardwang: Hows this?
```
**Use a good default.** When most users will (or should) pick a particular option, make it the default. For an uncontrolled component this can be done with the `defaultValue` prop.
```

bernardwang:
@pwolfert @line47 PR feedback should be addressed here: https://github.com/CMSgov/design-system/pull/434/commits/2c5798774d7674fb9967f3756523c34043251d48

line47:
Hey @bernardwang OK If I merge this now?

PR #435, opened 2019-06-19
[WNMGDS-62] Return original value in Mask component when unable to mask

bernardwang:
### Changed
`<Mask>` now will always return the originally entered `value` when `value` has no numeric characters and cannot be masked. Previously this component would return either empty string or the original `value`. This PR also renames `unmask` to the intended name `unmaskValue`.

### Deprecated
`unmask` in `Mask.jsx` has been renamed to `unmaskValue`. This was exported but never really publicly documented, so this should be fine.

### Fixed
`maskValue` and `unmaskValue` functions will only ever return the original `value`, or the masked value.


packages/core/src/components/TextField/Mask.jsx
```diff
@@ -84,9 +83,6 @@ function toDigits(value) {
  * @returns {Number}
  */
 function toNumber(value) {
-  if (typeof value !== 'string') return value;
-  if (!value.match(/\d/)) return undefined;
```
bernardwang: I removed these 2 lines because the checks are made prior to calling this function, and because all the other helper functions (i.e. `toDigits`, `toNumber`, `toDigitsAndAsterisk`) do not make any checks either. I figured it would be easier to handle the errors in one place in `maskValue()`.

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -109,19 +105,24 @@ function toNumber(value) {
...
+      value = '';
+    } else {
+      if (mask === 'currency') {
+        // Format number with commas. If the number includes a decimal,
+        // ensure it includes two decimal points
+        const number = toNumber(value);
+        if (number !== undefined) {
+          value = stringWithFixedDigits(number.toLocaleString('en-US'));
+        }
+      } else if (deliminatedMaskRegex[mask]) {
```
bernardwang: Changed `Object.keys(deliminatedMaskRegex).includes(mask)` to `deliminatedMaskRegex[mask]`

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -266,23 +270,24 @@ Mask.propTypes = {
  * @returns {String}
  */
 export function unmask(value, mask) {
-  if (!value || typeof value !== 'string') return value;
```
bernardwang: Previously `unmask` would only return the original value only if `mask` was an invalid option (not currency, phone number, ssn or zip). This rarely happens though, and most of the time the `unmask` function would work as expected, or it would return `''` when a non numeric value was given with a `currency` mask. 

This change makes `maskValue` and `unmask` behave consistently by returning when we cant unmask. This isnt really good either, but it is most similar to the existing behavior while also simplifying the logic.

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -260,32 +277,28 @@ Mask.propTypes = {
...
-  } else if (Object.keys(deliminatedMaskRegex).includes(mask)) {
-    // Remove the deliminators and revert to single ungrouped string
-    value = toDigitsAndAsterisks(value);
-  } else {
-    return rawValue;
   }
 
   return value;
 }
+// TODO: Remove deprecated naming
```
bernardwang: I renamed from `unmask` to `unmaskValue` because [this line](https://github.com/CMSgov/design-system/pull/435/files#diff-bc403a3448d8591da046ffb97228eab1R147) suggests `unmaskValue` is the intended name. Also this makes it consistent with `maskValue`.

packages/core/src/components/TextField/Mask.jsx
```diff
@@ -279,7 +282,7 @@ Mask.propTypes = {
  * @param {String} mask
  * @returns {String}
  */
-export function unmask(value, mask) {
+export function unmaskValue(value, mask) {
```
bernardwang: I renamed from unmask to unmaskValue to make it consistent with maskValue. It looks like the only reason why it was unmask in Mask.jsx in the first place was because Textfield.jsx imports the function, and then reexports it as unmaskValue. Since the docs already reference the function as unmaskValue, I thought we might as well use the same name everywhere

packages/core/src/components/TextField/Mask.test.jsx
```diff
@@ -332,72 +332,72 @@ describe('Mask', function() {
...
+    expect(unmaskValue()).toBeUndefined();
+    expect(unmaskValue(null)).toBeNull();
   });
 
-  it('returns empty string when there are no numeric characters in the value', () => {
-    expect(unmask('banana', 'currency')).toBe('');
-    expect(unmask('banana', 'zip')).toBe('');
-    expect(unmask('banana', 'ssn')).toBe('');
-    expect(unmask('banana', 'phone')).toBe('');
+  it('returns same string back when there are no numeric characters in the value', () => {
```
bernardwang: This is the only test case that changed in behavior

pwolfert:
Can this be merged now?

PR #436, opened 2019-06-19
Add info about examples

line47:
### Added
Added descriptions for the various examples that we have in the GitHub repo to help people find them easier. 

I'm not sure how helpful the form and article examples are and am thinking that the responsive examples could be explained in the context of the https://design.cms.gov/utilities/display-visibility/ page

### DIFF
--- a/packages/docs/src/pages/startup/examples.md
+++ b/packages/docs/src/pages/startup/examples.md
@@ -3,7 +3,7 @@ title: Examples
 weight: 4
 ---
 
-Additional examples of the design system in use can be viewed on GitHub. These projects demonstrate the various ways you can incorporate the design system into your development process and various use cases.
+Additional examples of the design system in use can be viewed on [GitHub](https://github.com/CMSgov/design-system). These projects demonstrate the various ways you can incorporate the design system into your development process and various use cases.
 
 ## HTML/CSS examples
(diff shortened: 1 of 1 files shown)


packages/docs/src/pages/startup/examples.md
```diff
@@ -5,8 +5,11 @@ weight: 4
 
 Additional examples of the design system in use can be viewed on GitHub. These projects demonstrate the various ways you can incorporate the design system into your development process and various use cases.
 
-<a href="https://github.com/CMSgov/design-system/tree/master/examples/" class="ds-c-button">Browse example projects</a>
-
-<h2 id="need-help" class="ds-h2 ds-u-color--primary-darker">Need help or ran into an issue?</h2>
-
-If you're having trouble installing or setting up the design system, or if you think you've found a bug, feel free to [open an issue on GitHub](https://github.com/CMSgov/design-system/issues).
```
bernardwang: Is it found elsewhere? or is it just not needed because theres already a github link?

line47: It's only at the bottom of getting started pages, I have a new issue I created to make this need help information part of the footer and sitewide rather than only on a few pages. 

Maybe that PR should include the removal here. 
I'm going to add back this content for now.

PR #439, opened 2019-06-21
moved old sketch UI file into an archive to reduce confusion

phiden:
Added V2 to master already; moving the original into an archive


line47:
I'm a bit unsure about naming the Sketch file V2. I think it's easier to keep the name as it is. 

I'm recommending that we remove the V2 from the file name and keep it as `CMS Design System UI kit.sketch` and don't keep an archived file at all.

phiden:
Happy to rename it, but want to keep the old one for the record.

line47:
Github will keep it in history for us so we don't have to 👍 
For example - https://github.com/CMSgov/design-system/commits/master/design-assets

line47:
I'm so excited this work is happening! 🚀

phiden:
> Github will keep it in history for us so we don't have to 👍
> For example - https://github.com/CMSgov/design-system/commits/master/design-assets

Right, but not everyone is familiar with Github, and this may make it harder to find if someone is interested.

line47:
I'm unsure as to why someone would want to use the old version - if you feel strongly about keeping it that works for me?

phiden:
> I'm unsure as to why someone would want to use the old version - if you feel strongly about keeping it that works for me?

Yeah, I do. It's happened before. We can revisit it at a later point.

PR #441, opened 2019-06-24
[MEA-846] Enable manual focusing of Button components.

scitech:
I wanted to focus a `<Button>` in https://jira.cms.gov/browse/MEA-846 but realized we didn't expose the ref callback in the design system.

### Added

- Allow components using `<Button>` to grab a reference to the underlying DOM element using a new prop called `buttonRef`.

### DIFF
--- a/packages/core/src/components/Button/Button.jsx
+++ b/packages/core/src/components/Button/Button.jsx
@@ -29,6 +29,7 @@ export class Button extends React.PureComponent {
       onClick,
       size,
       variation,
+      buttonRef,
       ...props
     } = this.props;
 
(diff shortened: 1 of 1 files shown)


line47:
Thanks for this contribution @scitech 🎉 
Is this update in your local code base already?

packages/core/src/components/Button/Button.jsx
```diff
@@ -117,6 +122,10 @@ Button.propTypes = {
    * Not called when the button is disabled.
    */
   onClick: PropTypes.func,
+  /**
+   * Access a reference to the `button` or `a` element
+   */
+  buttonRef: PropTypes.func,
```
bernardwang: I think these are supposed to be alphabetical, could you move this proptype up?

packages/core/src/components/Button/Button.jsx
```diff
@@ -87,7 +88,11 @@ export class Button extends React.PureComponent {
       delete attrs.type;
     }
 
-    return <ComponentType {...attrs}>{this.props.children}</ComponentType>;
+    return (
+      <ComponentType ref={this.props.buttonRef} {...attrs}>
```
bernardwang: since you already defined `buttonRef` above, just for sake of using it, can you remove the `this.props` here?

PR #443, opened 2019-06-26
[MCT-1695] Add aria and id to alert

SparrowSpit:
**Added**

- added `aria-labelledby` to the outer `div` of the Alert. Also, added this to the options in the component documentation
- added the option of `region` to `role` within prop and component documentation
- added a default `id` with the ability to write a custom `id`


line47:
Hey @SparrowSpit I created an additional PR with the updated snapshots and I'm thinking this could be closed. See https://github.com/CMSgov/design-system/pull/447

PR #445, opened 2019-06-29
[WNMGWINSHOP-196] Fix broken autocomplete in stateful tree

pwolfert:
### Fixed
- Fixed a bug where if the `Autocomplete` component were to be rendered in any kind of stateful tree (stateful parent or ancestor) and that state were to change (even if it didn't directly relate to one of the Autocomplete's props), the component would not be able to maintain its browser focus and would start to behave erratically.
  - The root cause of this was the `MyDiv` that functional component that was being defined inside the render function. It was not able to persist the instance of `MyDiv` between renders because the component function itself could not persist between renders. Moving it out of the render function fixed it.
  - I wanted to define it at the top of the `Autocomplete.jsx` module, but that generated too many eslint errors that I was not able to suppress with a reasonable number of _eslint-disable-line_ comments. (It didn't like multiple components being defined in the same module, components without propTypes, etc.)

### How to test this fix
1. Check out 61e5fe8f1f2e5e762f131810b2cae52ed3cbe45d, which has a modified, stateful example and the bug. 
2. Run the example and try typing into the first Autocomplete on the Autocomplete documentation page. Notice the erratic behavior of the input.
3. Switch to 6157c2a16c899910a305500cd6354bcea299d916, which has the modified, stateful example but a fix for the bug, and notice that it behaves as expected

### DIFF
--- a/packages/core/src/components/Autocomplete/Autocomplete.example.jsx
+++ b/packages/core/src/components/Autocomplete/Autocomplete.example.jsx
@@ -3,107 +3,121 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import TextField from '../TextField/TextField';
 
-ReactDOM.render(
-  <div>
-    <Autocomplete
-      items={[
-        {
-          id: 'kRf6c2fY',
-          name: 'Cook County, IL'
-        },
-        {
-          id: 'lYf5cGfM',
-          name: 'Cook County, MD'
-        },
-        {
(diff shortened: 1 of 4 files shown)


pwolfert:
I have also tried manually making this change in the node_modules of the project I'm working on, and it fixes all my problems.

line47:
Thanks @pwolfert and @1Copenut for tackling this!

PR #446, opened 2019-07-01
[WNMGWINSHOP-196] Upgrade Downshift from 3.0.0 to 3.2.10

pwolfert:
### Changed
- Upgrade the version of Downshift we use from `3.0.0` to `3.2.10` to get various bugfixes, including IE bugfixes

https://github.com/downshift-js/downshift/releases

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -12,7 +12,7 @@
     "@cmsgov/design-system-support": "^2.0.2",
     "classnames": "^2.2.5",
     "core-js": "^2.5.3",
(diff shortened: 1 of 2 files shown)


1Copenut:
@pwolfert I went back to my 2.0.0 release commit and checked the version of Downshift being pulled into the build. The yarn.lock file showed `3.2.7` so this bump wasn't as big as I first thought.

pwolfert:
Oh, interesting. I assumed it was 3.0.0 since that was what was in the package file. Thanks for regression testing!

PR #447, opened 2019-07-01
[MCT-1695] Add aria and id to alert

line47:
**Added**

- added `aria-labelledby` to the outer `div` of the Alert. Also, added this to the options in the component documentation
- added the option of `region` to `role` within prop and component documentation
- added a default `id` with the ability to write a custom `id`


line47:
There is 1 error preventing this that I'm not sure how to fix @pwolfert or @bernardwang maybe you have a thought? 

/design-system/packages/core/src/components/Alert/Alert.jsx
║ 10       │ 27       │ error    │ 'headingId' is missing in props validation             │ react/prop-types     ║

packages/core/src/components/Alert/Alert.jsx
```diff
@@ -49,6 +49,7 @@ Alert.propTypes = {
   ariaLabelledBy: PropTypes.string,
   children: PropTypes.node.isRequired,
   heading: PropTypes.string,
+  headingId: PropTypes.string,
```
line47: I think I fixed the props error with this

pwolfert: Yep, that's what it wanted since it was being referenced in the constructor

packages/core/src/components/Alert/Alert.jsx
```diff
@@ -1,10 +1,23 @@
...
 export class Alert extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    this.headingId = this.props.headingId || uniqueId('alert_');
+  }
+
   heading() {
-    if (this.props.heading) {
+    if (typeof id !== 'undefined') {
```
1Copenut: We're defining `this.headingId` in the constructor so we will always have the string *alert_N* available to create unique IDs. Users can override it by passing a prop `headingId` into their component when declared in JSX. I think we can return a heading without the if/else statement, maybe like this:

```javascript
heading() {
  return (
    <h3 className="ds-c-alert__heading" id={this.headingId}>
      {this.props.heading}
    </h3>
  );
}
```

packages/core/src/components/Alert/Alert.jsx
```diff
@@ -18,7 +31,11 @@ export class Alert extends React.PureComponent {
     );
 
     return (
-      <div className={classes} role={this.props.role}>
+      <div
+        className={classes}
+        role={this.props.role}
+        aria-labelledby={this.props.ariaLabelledBy}
```
1Copenut: Let's have the aria-labelledby reference the same string as the heading:

`aria-labelledby={this.headingId}`

packages/core/src/components/Alert/Alert.jsx
```diff
@@ -29,11 +46,13 @@ export class Alert extends React.PureComponent {
 }
 
 Alert.propTypes = {
+  ariaLabelledBy: PropTypes.string,
```
1Copenut: You can remove this prop if you choose to use `this.headingId` for the ID and aria-labelledby references.

packages/core/src/components/Alert/Alert.jsx
```diff
@@ -29,11 +46,13 @@ export class Alert extends React.PureComponent {
 }
```
1Copenut: I've been spot checking locally, and wanted to get your thoughts on adding a default role of "region" to the Alert component. Users can always override it with `role="alert"` or `role="alertdialog"` but having a default would ensure it always gets into the Landmarks menu for screen readers. Pairing this role with the always available ID + aria-labelledby makes it a named region that can be skipped to easily.

```javascript
Alert.defaultProps = {
  role: 'region'
};
```

line47:
Hey @SparrowSpit can you take a look at @1Copenut comments here

packages/core/src/components/Alert/Alert.jsx
line47: I think this is a smart default

line47:
This is ready for a re-review folks 👍 
CC @1Copenut

sawyerh:
@line47 I see this got marked as a breaking change in the release notes. Can you clarify what it might break?

line47:
Hey @sawyerh we flagged this because of the defaults we are adding as something to be mindful of for folks who might have customized the Alert component.

sawyerh:
@line47 Got it, thanks!

PR #451, opened 2019-07-05
update sketch with m.gov assets - https://jira.cms.gov/browse/WNMGDS-197

phiden:
### Added
Added m.gov-specific icons


line47:
This icon work is great and so necessary! 🎉 

I'm wondering if the M.gov icons should be part of the design system sketch file at all though? I feel like it muddies the waters a bit and sets the expectation of other icons to be added for other site packages. 

I'm wondering if we could select some images that are consistent between M.gov and healthcare.gov to add to the CMSDS, like we are doing with the component addition process only add stuff that is used in multiple instances.

PR #452, opened 2019-07-08
[WNMGDS-220] - Add browser support page and update getting started section

line47:
### Added
- Added new page for browser support information 

### Changed
- Changed `Getting started`  left nav item to `For developers` as all of the documentation was geared for developers and not designers. 
- Changed the order of left nav items under `For developers` to organize them a bit more than they were 
- Used sentence case for left nav items under `For developers`


packages/docs/src/pages/index.md
```diff
@@ -14,7 +14,11 @@ The design system is a set of open source design and front-end development resou
 
 It is currently being applied to [HealthCare.gov](https://www.healthcare.gov/). It is open source and freely available to use by anyone.
 
-<a href="{{root}}/startup/installation/" class="ds-c-button ds-c-button--primary ds-c-button--big">Learn how to get started</a>
```
line47: Added a section for developers and designers to the homepage. 👍

packages/core/src/__tests__/smoke.e2e.test.js
```diff
@@ -11,17 +11,17 @@ describe('CMS Design System smoke test', () => {
       .setRect(1024, 768, 0, 0);
   });
 
-  it('Should click Learn how to get started and open startup/installation/', async() => {
+  it('Should click Learn how to get started and open for-developers/installation/', async() => {
```
phiden: I'd rather we not rename this to "for developers" -- we've had use cases in the past where designers or content people have needed the information contained here and this label can be off-putting.

packages/docs/src/pages/index.md
phiden: I like this idea, but again, we've had cases where non-devs need the dev stuff, and non-designers need the design stuff. These labels can be off-putting, and I don't think we actually need to separate the content by role.

line47:
This is ready for a re-review @phiden 👍

PR #454, opened 2019-07-09
[WNMGDS-71] - Update table to be consistent with USWDS standards

line47:
### Changed
- Changed padding on cells to use `$spacer` variable instead of hard coded value

### Removed
- Removes `min-width: 100%` from tables so it doesn't take up the whole width for smaller data. 

### Screenshot of update
Screen Shot 2019-07-09 at 11 48 23 AM


line47:
This is ready for 👀 @phiden

PR #457, opened 2019-07-12
[WNMGDS-193] - Combine pattern lifecycle and contributing doc 

line47:
The pattern lifecycle document and contributing doc had some overlap and it makes sense to move the proposal process into the contributing doc. 

### Added
- Added pattern lifecycle contents to contribution doc

### Removed
- Removed the pattern lifecycle page


CONTRIBUTING.md
```diff
@@ -9,39 +9,102 @@ Please take a look at our [Code of Conduct](CODE_OF_CONDUCT.md) to learn more.
 
 If you are interested in running this site locally, please take a look at [setting up your local development environment](/README.md#running-locally).
 
-Use [our guides](https://github.com/CMSgov/design-system/tree/master/guides) to find additional information like:
-
-* Pattern proposal process
-* Coding guidelines
-* Guiding principles
-* How to write documentation
+Use [our guides](https://github.com/CMSgov/design-system/tree/master/guides) to find additional information like: Guiding principles, Coding guidelines, and How to write documentation.
```
phiden: Cognitively, a running list like this is harder to read. I think that's why we'd formatted it as a bulleted list, which I don't see the harm in retaining.

PR #458, opened 2019-07-12
[WNMGDS-110] - upgrade Lodash to 4.17.14 (latest)

line47:
### Changed
- Updated lodash to latest version


line47:
@pwolfert if you can take a look here too it would be appreciated

PR #459, opened 2019-07-12
[WNMGDS-67](https://jira.cms.gov/browse/WNMGDS-67) Vis reg testing

phiden:
### Added
- Installed [backstop.js](https://github.com/garris/BackstopJS )for visual regression testing
- Configured backstop for our responsive screen sizes and 77 test cases -- one for each unique URL
- Ran the test and established a baseline, which is checked in.


line47:
This looks like a good start and I am wondering a few things.

- Can we move the `backstop_data` directory under `tools` to help organize the repo?

- Should we incorporate testing scripts into our package.json https://github.com/garris/BackstopJS#using-npm-run-scripts

- I tried to run `backstop test` and it didn't work. Do I need to install to test the functionality, I ran `yarn install` and that didn't change anything.

phiden:
I don't think it belongs in our automated testing scripts -- we don't need visual regression run _every single time_ we run those tests, and they take a few minutes. Maybe we revisit at a later point but for now they should be manual. 

Good call on adding backstop to the yarn file; I'll do that and see if I can move the data directory without breaking the tool.

phiden:
Closing in favor of the other test PR

PR #460, opened 2019-07-12
[WNMGDS-203] - Update and document yeoman component generator

line47:
### Added
- Added instructions to the README so the user knows what to expect when using the generator
- Added an index.example.html file to the templating 

### Changed
- Updated yeoman to 2.0.6 
- Updated the default content when creating a component 

to test - run `yarn install` then run `yo cmsgov` see the readme for more info https://github.com/CMSgov/design-system/blob/d36869470045fd6b53521491b049c301537a6e03/packages/generator-cmsgov/README.md


phiden:
`yo cmsgov` returns `yo: command not found`

PR #474, opened 2019-08-01
[WNMGPCII-261] StepList steps to have optional onClick handler

pwolfert:
### Added
- `StepList` `step` objects can now have their own `onClick` property specific to that step

### Verifying
- You can verify that it works by modifying the steps in the example and including an `onClick: () => alert("hey!")` in some of them.


packages/core/src/components/StepList/Step.jsx
```diff
@@ -19,6 +19,21 @@ export const Step = ({ step, ...props }) => {
...
   const descriptionLabel = descriptionLabelText.replace('%{step}', step.title);
+
+  let linkLabel;
+  if (step.completed && !step.steps) {
+    linkLabel = step.linkText || props.editText;
+  } else if (start) {
+    linkLabel = step.linkText || props.startText;
+  } else if (resume) {
+    linkLabel = step.linkText || props.resumeText;
+  }
```
austin-martinez: :+1:

PR #486, opened 2019-08-09
[WNMGDS-67] Set up visual regression testing with BackstopJS

line47:


line47:
I think we need to remove the `tota11y-toolbar`  from running during `yarn start` and instead have a separate yarn command we can run to enable the toolbar. https://khan.github.io/tota11y/

It's causing visual regression false positives when comparing local files to the live site. 
@1Copenut @bernardwang you may have some thoughts here.

### Screenshot (lower right is where the toolbar shows)
Screen Shot 2019-08-16 at 9 17 14 AM

1Copenut:
@line47 I think you're exactly right. The Tota11y toolbar has it's own JS file, and is probably being loaded async, so sometimes it's on the page when the screenshot is taken, sometimes it's not. 

I wonder if we can pass a `NODE_ENV` variable to the visual regression run, and wrap the Tota11y JS load in a conditional like:

```javascript
if ('NODE_ENV_VIZ' !== true) {
  <script src="./path/to/tota11y.js"></script>
};
```

line47:
I think i can use the removeSelectors variable to remove the toolbar during testing https://github.com/garris/BackstopJS#removing-selectors

I'm going to give that a go

line47:
Instructions for running backstop JS are now in the README https://github.com/CMSgov/design-system/tree/sw-visual-regression#visual-regression-testing

line47:
Hey @bernardwang can you give this another pass - I think it's ready to go

bernardwang:
So I'm seeing 3 tests that are failing because it is comparing a gif, I'm not sure how we can fix that one. The other 8 tests I'm seeing fail are all from the page not finishing loading. I tried bumping up the `delay` property in `backstop.json` but it doesnt seem to help much. Are you able to get those tests to work? @line47

line47:
I was having issues with those as well and I think what the issue is that the saved image it is testing against is the loading/half loaded page.

See this screenshot - 
https://github.com/CMSgov/design-system/blob/sw-visual-regression/tools/backstop_data/bitmaps_reference/backstop_default_CMSDS_-_Utilities_-_Border_Radius_0_document_2_desktop.png

line47:
Also, I'm unsure how to replace one or a few reference screenshots that maybe didn't capture correctly the first time. 🤔

bernardwang:
also, can we add `tools/backstop_data/html_report/` to the `.gitignore`?

bernardwang:
@line47  I got the tests to pass everything except for 1 test. I did have to reset the reference files to get this result though. When you got a chance to confirm, then I think maybe this can be merged!

line47:
Thanks @bernardwang - going to merge this!

PR #490, opened 2019-09-16
[WNMGDS-234] Fix error message visibility

bernardwang:
### Added
New "header" description for the Form Label & Legend page. I noticed pretty every page had a short 1 or 2 sentence description for each component so I thought I should add one. 

### Changed
Error messages now come after the hint to be closer to the input. Html examples have been updated with this new ordering


line47:
Should this also be updated in the Date field example?
- https://design.cms.gov/components/date-field/

bernardwang:
@line47 I'm seeing the date field with the updated error message position. Is this what you mean?
https://gyazo.com/4d29c4e1fc354bb84193d99d8ad50f7c

line47:
Yep that's what I was referring to, looks good :shipit:

PR #492, opened 2019-09-20
[WNMGDS-238] Move HTML examples from inline to stand alone files

line47:
Moving inline example code to stand alone example files. 🎉


bernardwang:
This looks really good to me, all the examples seem to be working. I was wondering about the naming pattern for the html examples though, do you have a particular reason for doing the undercase and dash naming pattern?

packages/core/src/components/Button/Button.scss
```diff
@@ -150,13 +150,12 @@ Style guide: components.button
 
 /*
 ---
+Inverse buttons
```
bernardwang: Other similar labels are `h2`s, I removed this markdown to be consistent

packages/core/src/components/TextField/TextField.scss
```diff
@@ -63,6 +63,7 @@ Style guide: components.text-field
 
 /*
 ---
+Inverse theme
```
bernardwang: Button is the only other component with a separate code block for inverse theme. Later maybe we should consider combining the inverse examples with the main examples like in `<Dropdown>`?

Issue #343, opened 2019-02-20
Consider deprecating the inputs to right of label pattern on Choice component


1Copenut:
## Description
The `Choice` component allows for inputs to be placed to the right of labels, which can create ragged spacing and readability issues. We should discuss if this is a pattern we want to maintain going forward, or if it would be better deprecated. This discussion will include identifying any uses in the wild, and ways to notify and remediate any breaking changes.

This came about as part of a discussion on https://github.com/CMSgov/design-system/pull/341, specifically @melwoodard comment https://github.com/CMSgov/design-system/pull/341#issuecomment-465618775

Including screen shot from that PR showing an example right to left layout with `checkedChildren`

## Additional information [optional]
![screen shot 2019-02-20 at 7 24 45 am](https://user-images.githubusercontent.com/934879/53103593-8be13400-34f3-11e9-8c58-2a184050a203.png)

line47:
I would suggest expanding this work to deprecate `Right-to-Left` checkboxes and radios as well
![screen shot 2019-02-20 at 10 53 52 am](https://user-images.githubusercontent.com/1449852/53104707-dd8ebc00-34fd-11e9-9908-0f3dd889f6f8.png)
.

line47:
I think we should add a component maturity status for deprecated items

line47:
@phiden and @melwoodard What are your thoughts here?

phiden:
Let's also add this change to the roadmap so that we can prioritize it with the other changes.

bernardwang:
@line47 Should we be removing this from the codebase, or should we leave it and just add a "Deprecated" label?

line47:
The current documentation says something about leaving the pattern available for at least a month after being deprecated.
https://github.com/CMSgov/design-system/blob/master/CONTRIBUTING.md#retiring-patterns

There is also the `deprected` label we can apply to that component or section. 
https://design.cms.gov/guidelines/component-maturity/

bernardwang:
Sounds good. Should I also add the `deprecated` label to the `Select` component as well?

line47:
I think that makes sense 👍 You should be able to use `@status Deprecated` 
https://github.com/CMSgov/design-system/blob/master/guides/WRITING-DOCUMENTATION.md#flags

bernardwang:
These patterns have been labeled as deprecated, but they still need to be removed from the code in the future

PR #494, opened 2019-09-23
[WNMGDS-112] Deprecate Right-to-Left and <Select>

bernardwang:
closes https://github.com/CMSgov/design-system/issues/343
https://jira.cms.gov/browse/WNMGDS-112

### Deprecated
Right-to-Left choice/checkboxes
`<Select>` component


packages/core/src/components/ChoiceList/Choice.scss
```diff
@@ -195,6 +195,8 @@ $ds-c-inset-border-width: $spacer-half;
 /*
 Right-to-Left
 
+<span class="ds-c-badge ds-u-margin-right--1 ds-u-text-transform--capitalize">Deprecated</span>
```
bernardwang: @status Deprecated doesn't work here I believe.

line47: You are right it only works in the KSS flags portion of the documentation

PR #495, opened 2019-09-24
[WNMGDS-239] Separate out inline styles

bernardwang:
Our current KSS files mixed documentation with inline CSS. Now that we have removed the inline React/HTML code we should separate out the CSS too. Ideally we wouldnt have files named `TextField.docs.css` and we could have `TextField.md` instead, but in the meantime this will hopefully make it easier to understand how our doc site is built

## Changed
Split most scss files into `*.docs.scss` for documentation related comments and `*.scss` for component source scss. 
### Added
Added new `Mask.doc.scss` by factoring out inline documentation comments in `Mask.jsx`. Also separated out `Mask.scss` from `CurrencyField.scss` and `TextField.scss`.

Also added `Dropdown.doc.scss` to `index.scss` because I forgot it earlier.


packages/core/src/components/TextField/Mask.scss
```diff
@@ -1,5 +1,16 @@
 @import '@cmsgov/design-system-support/src/settings/index';
 
+.ds-c-field-mask {
```
pwolfert: 👍

bernardwang:
@pwolfert Yep those utilities should be separated out as well!

PR #496, opened 2019-09-25
[WNMGDS-222] Add onComponentBlur to <ChoiceList>

bernardwang:
https://jira.cms.gov/browse/WNMGDS-222

### Added
`inputRef` prop to `<Choice>`
`onComponentBlur` prop to `<ChoiceList>`

### DIFF
--- a/packages/core/src/components/ChoiceList/Choice.jsx
+++ b/packages/core/src/components/ChoiceList/Choice.jsx
@@ -90,6 +90,7 @@ export class Choice extends React.PureComponent {
       requirementLabel,
       size,
       uncheckedChildren,
+      inputRef,
       ...inputProps
     } = this.props;
 
@@ -116,6 +117,9 @@ export class Choice extends React.PureComponent {
           onChange={this.handleChange}
           ref={input => {
             this.input = input;
+            if (inputRef) {
+              inputRef(input);
+            }
           }}
           {...inputProps}
         />
@@ -170,6 +174,10 @@ Choice.propTypes = {
    * otherwise, use the `checked` property.
    */
   defaultChecked: PropTypes.bool,
+  /**
+   * Access a reference to the `input` element
+   */
+  inputRef: PropTypes.func,
   /**
    * Additional hint text to display below the choice's label
    */
--- a/packages/core/src/components/ChoiceList/ChoiceList.jsx
+++ b/packages/core/src/components/ChoiceList/ChoiceList.jsx
@@ -7,7 +7,7 @@ import classNames from 'classnames';
 import uniqueId from 'lodash.uniqueid';
 
 /**
- * A `ChoiceList` component can be used to render a select menu, radio
+ * A `ChoiceList` component can be used to render a radio
  * button group, or checkbox group.
  *
  * By default the component determines the type of field for you, taking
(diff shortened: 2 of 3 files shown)


packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -170,6 +174,10 @@ Choice.propTypes = {
    * otherwise, use the `checked` property.
    */
   defaultChecked: PropTypes.bool,
+  /**
+   * Access a reference to the `input` element
+   */
+  inputRef: PropTypes.func,
```
bernardwang: I filed a ticket to standardize the naming for all these `ref` props across all the components

https://jira.cms.gov/projects/WNMGDS/issues/WNMGDS-242

packages/core/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -122,6 +132,27 @@ export class ChoiceList extends React.PureComponent {
...
+    if (this.props.onBlur) {
+      this.props.onBlur(evt);
+    }
+
+    if (this.props.onComponentBlur) {
+      this.handleComponentBlur(evt);
+    }
+  }
+
+  handleComponentBlur(evt) {
```
bernardwang: This logic is modeled from `onComponentBlur` in DateField

packages/core/src/components/ChoiceList/ChoiceList.test.jsx
```diff
@@ -166,18 +166,37 @@ describe('ChoiceList', () => {
...
+      const data = shallowRender({ onBlur, onComponentBlur });
       data.wrapper
         .find('Choice')
         .first()
         .simulate('blur');
 
-      expect(onBlur.mock.calls.length).toBe(1);
+      expect(onBlur).toHaveBeenCalled();
+      setTimeout(() => {
+        expect(onComponentBlur).not.toHaveBeenCalled();
```
pwolfert: I don't actually understand why this one is working and feel like it must be that when we simulate blur in Enzyme, it focuses on the next one automatically. It works when I test it out for real though.

bernardwang: When we blur the first choice, it does go to the next choice automatically, but we dont want it to fire `onComponentBlur` until the last choice is blurred. Does that answer your question?

pwolfert: My question is whether this is an implicit feature of Enzyme. If we're all okay with this as expected behavior of Enzyme, then that's good enough for me. Might be good to add a comment that says that if we simulate blur that it will automatically focus on the next element.

bernardwang: Ah I see, I couldnt find much info on the enzyme docs page to confirm, i just assumed it was expected cuz thats what the browser does. I can add a comment though for sure

packages/core/src/components/ChoiceList/Choice.jsx
bernardwang: @pwolfert @line47 Do either of you have ideas for how to standardize these `ref` props? `fieldRef` is ok, but not great considering a button isn't a field. Naming after the actual html element doesn't work great either because this is often dynamic (i.e. <Textfield> could be `textarea` or `input`). In my head it would be nice to be able to use a single naming standard for accessing the base html element of a component, my top two are `inputRef` or `elemRef`

pwolfert: Yeah, I'm not sure. If we really want it standardized, `inputRef` is reasonable even if it's a textarea, whereas `elemRef` is kind of an abiguous name because there are several elements in most of these components

bernardwang: `inputRef` it is!

PR #497, opened 2019-10-01
[WNMGDS-244][WNMGWINSHOP-277] Fill for default buttons

pwolfert:
### Changed
- Made the default button have background fill (with inverse fill for inverse buttons). This maintains proper contrast levels when default buttons are on various backgrounds. 
  Screen Shot 2019-10-01 at 11 45 27 AM

### DIFF
--- a/packages/core/src/components/Button/Button.scss
+++ b/packages/core/src/components/Button/Button.scss
@@ -162,10 +162,6 @@ Inverse buttons
   }
 }
 
-.ds-c-button--transparent-inverse {
-  background-color: transparent;
-}
-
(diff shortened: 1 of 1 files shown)


packages/core/src/components/Button/Button.scss
```diff
@@ -160,6 +162,10 @@ Inverse buttons
   }
 }
 
+.ds-c-button--transparent-inverse {
```
bernardwang: Isn't this already styled above on line 71?

pwolfert: Doh!

PR #498, opened 2019-10-01
[WNMGDS-174] Define pattern template and add to docs

bernardwang:
### Added
Overview page under "Patterns". Content is basically copy and pasted from [va.gov](https://design.va.gov/patterns/). I would really appreciate more input here!

### Changed
Utilities page under "Utilities" was changed to Overview. Due to changes in implementation, small changes happened from this.
1. The `url` went from `/utilities` to `/utilities/overview`. I don't think this is ideal, but maybe can be addressed in a future PR.
2. The `View Source File` link is removed from the header. I think this makes sense as there's no source code to Overview pages, and this is consistent with the other markdown based pages like the ones under "Guidelines"

### Screenshots of changes
Screen Shot 2019-10-01 at 2 51 28 PM
Screen Shot 2019-10-01 at 2 51 34 PM


guides/WRITING-DOCUMENTATION.md
```diff
@@ -224,16 +224,23 @@ The format of these markdown pages follow the same format as Jekyll pages:
...
 weight: 100
 ---
 ```
 
+#### Label (optional)
+The `label` property is used for the nav label for a page. If it is not provided, the title property is used instead
+
+#### Weight (optional)
+The `weight` property controls the order of the page in the left navigation. `0` would be first followed by `1` then `2` and so on.
+
```
bernardwang: Down the line if we remove KSS altogether, we would want more detailed docs on all the frontmatter arguments in general

tools/gulp/docs/index.js
```diff
@@ -63,6 +63,12 @@ function addTopLevelPages(kssSections) {
         sections: [],
         weight: 7
       },
+      {
```
bernardwang: I think this was missing before because Utilities was the only top level section that had a "Overview" page (it was labeled "Utilities"). As a result the top level section and it's corresponding nav label was kind of built dynamically from the KSS (I think?). So basically I made the top level section explicitly defined here instead.

tools/gulp/docs/processMarkdownPage.js
```diff
@@ -51,9 +51,11 @@ function processMarkdownPage(filePath, body, rootPath = '') {
     referenceURI = path.join(rootPath, referenceURI);
   }
 
+  const header = parts.attributes.title || 'Untitled';
   let page = {
     depth: depth,
-    header: parts.attributes.title || 'Untitled',
+    label: parts.attributes.label || header,
```
bernardwang: `label` is the new optional front matter argument that allows the nav label to be different than the actual page title. I couldn't figure out how to add this support to KSS, so I only added the functionality to the markdown files. I'm hoping we can eventually move away from KSS altogether so this won't matter. 

At the very least, now its consistent where only the component & . utilities source pages use KSS, and everything else is markdown.

packages/docs/src/pages/patterns/overview.md
```diff
@@ -0,0 +1,15 @@
+---
+label: Overview
+title: Patterns
+weight: -1
```
bernardwang: It looks like both the markdown and KSS pages have a default of `weight: 0`. I couldnt figure out how to change the default for both to `1`, so I did this kinda hacky thing instead haha... open to suggestions here

packages/docs/src/pages/patterns/overview.md
```diff
@@ -0,0 +1,15 @@
...
+---
+Patterns are solutions and researched best practices that solve recurring user interface design problems:
+
+  * How do you build complex, multi-part web forms?
+  * How do you provide navigational context?
+  * How do you handle progressive disclosure?
+
+## How are they different from components?
+A pattern is more than the sum of its parts. Patterns are solutions, whereas a component can be considered a UI chunk. Patterns consist of design, content strategy, reusable components, and accessibility put together to solve common problems on CMS.gov.
```
melwoodard: Suggest replacing `CMS.gov` with `CMS websites`

packages/docs/src/pages/patterns/overview.md
line47: I think this is where the default is set to `0` https://github.com/CMSgov/design-system/blob/5675f0b8c1e6bf297bd71a37d539bc0753916d9c/tools/gulp/docs/processMarkdownPage.js#L64

packages/docs/src/pages/patterns/overview.md
line47: Could we remove `on CMS.gov` and end the sentence after `problems` ?

packages/docs/src/pages/patterns/overview.md
bernardwang: Yeah that was the default for the markdown pages but I couldnt seem to find where the default is set for KSS.

bernardwang:
@melwoodard The copy changes you mentioned are in now, lmk if this looks good to go!

PR #500, opened 2019-10-02
Fix visual regressions

bernardwang:
With backstop I discovered 2 regressions introduced by the separating out inline CSS/HTML work. This PR fixes them before the release

### Fixed
1. Date field css is changed
https://gyazo.com/ca29a5ed8363562539f9d5dd4f3b3609
2. Documentation for <Tab> is no longer there https://gyazo.com/56a87571e1af402778258009bcd893d7

### DIFF
--- a/packages/core/src/components/_index.scss
+++ b/packages/core/src/components/_index.scss
@@ -5,7 +5,6 @@
 @import 'Button/Button';
 @import 'ChoiceList/Choice';
 @import 'ChoiceList/Select';
-@import 'DateField/DateField';
 @import 'Dialog/Dialog';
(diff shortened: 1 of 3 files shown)


packages/core/src/components/_index.scss
```diff
@@ -17,8 +16,11 @@
 @import 'StepList/StepList';
 @import 'Table/Table';
 @import 'Tabs/Tabs';
+// Mask and Date field styles need imported
+// after TextField since they modify it
```
pwolfert: Nice find! This seems a little fragile, but I'm not sure what the best solution is right now. Right now just fixing it is probably the best solution. Thanks for the comment

PR #506, opened 2019-10-04
[WNMGDS-109] - Remove guidance tab and add guidance anchor link

### LINKED ISSUE #395: Can't toggle between tabs when URL contains an #anchor to the second tab, and the second tab is missing active state
**Describe the bug**
I can't toggle between the usage and guidance tabs when I land on a page via a URL that has the #anchor linking me to the second tab on the page.  

**To Reproduce**
1. Go to https://design.cms.gov/components/form-label/#guidance
2. Try to toggle between the Usage and Guidance tabs

**Expected behavior**
Second tab should have active state (not first), and should be able to toggle between usage and guidance

 - OS: Mac OS 10.14.3
 - Chrome v72.0.3626.121

**Additional context**
Works as expected if I go to the page via https://design.cms.gov/components/form-label/#usage, so thinking it has something to do w/ linking to the second tab?

### LINKED ISSUE #398: Remove usage and guidance tabs - use single column instead
I'd like to recommend we remove the `usage` and `guidance` tabs in favor of having one page without tabs.

- Place the guidance documentation at the bottom of the page. 
- Replace the USWDS link in the header with an anchor link to take the user down in the page to the guidance. 

## Screenshot example/comparison

Screen Shot 2019-03-26 at 9 08 14 AM

this loosely relates to #395 as well

line47:
### Added
- Added a `view guidance` anchor link to the page header to link the user down to the content. This link shows based on if there are guidance sections written for that particular page or not.

### Changed
- Changed some of the headers below the guidance header from H2 to H3 to help with the outlining of the page. 
- Changes some of the headers that said `theming` when talking about SCSS variables to use the header `Customization` instead as that was already more prevalent in the documentation and i think makes more sense. 
- Changed redundant `Guidance` headers in the content to `Usage`

### Before screenshot
Screen Shot 2019-10-04 at 5 04 20 PM

### After screenshot
Screen Shot 2019-10-04 at 5 04 26 PM

### Fixed
fixes #395
closes #398


packages/core/src/base/typography.scss
```diff
@@ -217,7 +217,7 @@ Style guide: style.typography.responsive
 */
 
 /*
-Customizing type sizes
```
bernardwang: Just curious, whats the reasoning for this change? Do other pages with similar docs also call it "Customization"?

packages/docs/src/scripts/components/Page.jsx
```diff
@@ -21,9 +20,7 @@ class Page extends React.PureComponent {
       default: 3,
       sm: 6
     };
-    this.tabPanelClasses = `ds-u-border--0 ds-u-padding-x--${this
-      .horizontalPadding.default} ds-u-sm-padding-x--${this.horizontalPadding
-      .sm} ds-u-padding-y--0`;
+    this.tabPanelClasses = `ds-u-border--0 ds-u-padding-x--${this.horizontalPadding.default} ds-u-sm-padding-x--${this.horizontalPadding.sm} ds-u-padding-y--0`;
```
bernardwang: Is this used anywhere?

packages/docs/src/scripts/components/Page.jsx
```diff
@@ -58,48 +55,46 @@ class Page extends React.PureComponent {
     return <PageBlock {...this.props} hideHeader />;
   }
 
+  renderGuidanceTabPanel() {
```
bernardwang: Can we rename this to be `renderGuidanceSection`

packages/docs/src/scripts/components/PageHeader.jsx
```diff
@@ -26,6 +26,16 @@ class PageHeader extends React.PureComponent {
     }
   }
 
+  guidanceLink() {
+    if (this.props.showGuidanceLink) {
+      return (
+        <a className="ds-u-margin-right--2" href="#guidance">
+          View guidance
```
bernardwang: I noticed the other two links are to external sites. I wonder if we want to distinguish this link more? like maybe changing the content to `Jump to guidance` or adding a spacer somehow? Do you have any thoughts on this @melwoodard?

packages/core/src/base/typography.scss
line47: Yeah there was a mix between `theming` and `customization` I'm open to suggestions too. 
This is where SASS variables will be displayed to help with styling

packages/docs/src/scripts/components/Page.jsx
line47: Nope, I'm removing it

packages/docs/src/scripts/components/PageHeader.jsx
line47: Updated to `jump to guidance`

packages/docs/src/scripts/components/Page.jsx
bernardwang: @line47 I think this is the last thing for me

line47: Updated 👍

bernardwang:
@line47  Something else to consider, other pages without guidance still have a tab. Should we have a single tab even though there's no guidance? Or should we remove the tab everywhere (although we also talked about adding a react tab haha)
https://gyazo.com/da1ec68090aaf7c3df44b5c5d14faf45

line47:
I'm not seeing a tab on the color page when running the branch locally. There should be no tabs on 
Screen Shot 2019-10-10 at 10 20 52 AM

line47:
@davidakennedy can you take a look at this PR and weigh in?

PR #507, opened 2019-10-07
[WNMGDS-262] Remove build files from Github

bernardwang:
### Summary
Having build files in GH adds an extra step to our release process and makes development a tiny bit more annoying. Generally build files arent included in source control, so this PR removes it.
CBJ runs `yarn build` when deploying to design.cms.gov, and our npm package process just publishes whatever local code you have, so it should be safe to remove. Only changes were to documentation so people can still consume the build files directly

### Changed
Exclude ece helper files from build files

### Removed
Remove dist folders from build script


bernardwang:
@line47 it looks like you need to run `yarn build` before running `yarn test` in order for them to pass... we might need to update the automated tests to get it to work

line47:
When I run `yarn install` and `yarn test` locally I get 1 fail.
``` 
FAIL  packages/core/src/components/Dialog/Dialog.e2e.test.js (6.34s)
```

Why is Travis reporting others?

.travis.yml
```diff
@@ -3,7 +3,11 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.1.0
   - export PATH="$HOME/.yarn/bin:$PATH"
   - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 --no-sandbox --disable-dev-shm-usage http://localhost &
-branches:
+install:
+  - yarn install
+  - yarn build
+  - yarn test
+ranches:
```
pwolfert: 🤠 typo?

.travis.yml
```diff
@@ -3,7 +3,11 @@ before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.1.0
   - export PATH="$HOME/.yarn/bin:$PATH"
   - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 --no-sandbox --disable-dev-shm-usage http://localhost &
-branches:
+install:
```
bernardwang: My understanding is that `assertNoAxeViolations` in e2e requires build files to work, as they check actual pages in `localhost:3001`. Previously build files were in the branch, so this config file adds the extra `yarn build` step and fixes that.

additional note, since the build files were previously only updated on releases, I think the old config here was a bit broken so its nice to fix that

tools/gulp/build.js
```diff
@@ -29,7 +29,8 @@ module.exports = (gulp, shared) => {
           `packages/${pkg}/src/**/*.{js,jsx}`,
           `!packages/${pkg}/src/**/{__mocks__,__tests__}/*.{js,jsx}`,
           `!packages/${pkg}/src/**/*.example.{js,jsx}`,
-          `!packages/${pkg}/src/**/*.test.{js,jsx}`
+          `!packages/${pkg}/src/**/*.test.{js,jsx}`,
+          `!packages/${pkg}/src/helpers/**/*.{js,jsx}`
```
bernardwang: Files from `helpers/e2e` were being included in the build, this removes them because the code was related to testing. 

I wasn't entirely sure if I should have ignored only `e2e` or the entire `helpers` folder. I.e., if there's code in helpers and that's being used by a component, does it need to be included here or is it automatically included?

packages/docs/src/pages/startup/installation.md
```diff
@@ -3,7 +3,7 @@ title: Install with NPM
 weight: 0
 ---
 
-The design system is available as NPM packages or via a <a href="https://github.com/CMSgov/design-system/archive/v1.30.0.zip">.zip download</a>.
```
bernardwang: Old page directly linked to a release version. I updated to our normal release page

.travis.yml
bernardwang: yeehaw

packages/docs/src/pages/startup/sass-and-css.md
```diff
@@ -7,7 +7,8 @@ weight: 1
 
 The easiest way to add the design system's styles to your site is by referencing its minified CSS.
 
-1. Copy the design system's `dist/css` folder into a relevant place in your code base — likely a directory where you keep third-party libraries. In the example below, our directory is `css/vendor`.
+1. Download the zip file from the latest CMS design system release and open that file.
+1. Copy the `packages/core/dist/index.css` file into a relevant place in your code base — likely a directory where you keep third-party libraries. In the example below, our directory is `css/vendor`.
```
pwolfert: 👍

tools/gulp/build.js
pwolfert: I think since we've labeled the `helpers/e2e` folder as an _e2e_ thing by calling it that, we're saying that that part is for e2e tests and the rest of `helpers` is for whatever else. That is, we're not prescribing a use for all of `helpers/` but we are for `helpers/e2e/`, so maybe we should only ignore that part. It's possible that we put other things in there in the future.

pwolfert: It would also be possible to overload our use of the `utilities` folder by putting javascript in there, but that could get messy. What do you think?

pwolfert:
My local unit tests have a lot of failures because the modules for component unit tests reference the dist version of components when importing. We should probably change all the references in the unit test modules to import the src versions, since that's what they're testing anyway. (Alternative would be to have it build the dist version before unit testing, but that's yucky.)

tools/gulp/build.js
bernardwang: I'm fine with just only ignoring `helpers/e2e` if that's ok with you!

bernardwang:
Update: I'm a bit stuck right now trying to get tests to pass without having to run `yarn build`. A few doc site component tests are failing on the lines importing from the design system, not sure why yet

bernardwang:
Yay, thanks to @pwolfert's help tests are passing!

.gitignore
```diff
@@ -1,6 +1,10 @@
 # Documentation is generated into this directory
 /docs
 
+# Do not commit package build files to github
+/packages/core/dist
+/packages/layout/dist
```
bernardwang: Maybe want to consider adding the themes folder too?

PR #508, opened 2019-10-08
[WNMGDS-245 WNMGDS-249] Add deprecated warnings and remove select example from <ChoiceList>

bernardwang:
### Added
Deprecated warnings to <Select> and right to left in <Choice> 

### Changed
Removed Select example from <ChoiceList>. The <Select> example is still present.


packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -19,6 +19,12 @@ export class Choice extends React.PureComponent {
   constructor(props) {
     super(props);
 
+    if (props['inputPlacement'] === 'right') {
+      console.error(
+        `[Deprecated]: 'inputPlacement' is no longer supported and will be removed in a future release, please remove right-to-left <Choice>`
```
line47: I'm wondering if we should also update the text in the prop table so folks know it's deprecated.
https://github.com/CMSgov/design-system/pull/508/files#diff-7c1a7d4da2bdf438dffdc0cd5254e7edR205

I'm also wondering if in the future we remove the `inputPlacement` prop altogether

line47: Tested this and it's working 👍 
I'm wondering if we should add text about it being a React property something like 
```
[Deprecated]: Please remove the React property 'inputPlacement' for the <Choice> component. It is no longer supported and will be removed in a future release.```

bernardwang: that sounds good!

bernardwang: yeah I can add it to the prop table too. Yes, we should remove the prop in the future for sure

bernardwang:
@line47 i just pushed changes, lmk what you link!

PR #510, opened 2019-10-10
[WNMGDS-256] Add focus styles across the DS

davidakennedy:
* This isn't commit-ready, but more for reviewing and talking about design direction.

**What type of feedback am I looking for?**

It would good to think about these questions:

* Do any of these feel like they fit with the current system?
* Should we explore a lighter color for the `outline` color?
* Are there any other techniques we could explore here that would give us focus states that don't rely on color alone?
* Should we alter regular link and/or form focus styles at all as a result of this work?

## Added
* Added five different classes to help prototype button focus styles.

## Changed
N/A.

### Deprecated
N/A.

### Removed
N/A.

### Fixed
N/A.


packages/core/src/components/Button/Button.scss
```diff
@@ -185,3 +185,119 @@ Button inline with field
 .ds-c-button > svg {
   @include inline-icon;
 }
+
+/* Show focus styles in demo. */
+.ds-c-button--focus-1 {
```
bernardwang: Personally I think the plain box focus style works. In other sites I have seen an even larger outline width (like 4px) but idk how others feel about that. Also I'm curious to know what other colors would look like... Are there other options that have good contrast?

bernardwang:
Also to confirm, would this focus styles apply to other elements too? I saw you named the styles specifically to buttons but I was under the assumption this change would override the default focus style for all html elements

line47:
I like the styling gov.uk uses 
https://design-system.service.gov.uk/get-started/focus-states/

> Focus states in the GOV.UK Design System use a combination of yellow and black to make sure  they meet Web Content Accessibility Guidelines (WCAG) 2.1 level AA non-text contrast on any background colour used on GOV.UK.

> The yellow has a high contrast with dark backgrounds and the thick black border has a high contrast against light backgrounds.

I feel this type of styling works on another level other than color by adding a thick black border under focused items. 

**Should we alter regular link and/or form focus styles at all as a result of this work?**
Yes!

packages/core/src/components/Button/Button.scss
davidakennedy: > Also I'm curious to know what other colors would look like... Are there other options that have good contrast?

Yeah, we can definitely try other colors. None of our other secondary blues meet the proper color contrast. We could use some of our primary colors, but most would be even darker than the existing one.

davidakennedy:
> Also to confirm, would this focus styles apply to other elements too? I saw you named the styles specifically to buttons but I was under the assumption this change would override the default focus style for all html elements

@bernardwang Yes, we could do that. I wanted to start here and nail down the style, the expand outward.

> I feel this type of styling works on another level other than color by adding a thick black border under focused items.

@line47 Agreed. We could explore that a bit in our system.

melwoodard:
Echoing sentiments above -- It'd be worth looking at something closer to gov.uk and USWDS's approaches to focus state. Bright colors and thick borders give a great amount of contrast to make it super clear what's in focus, and tend to stand out well regardless of the surrounding background color. Also agree we should apply (mostly) the same focus state to anything that has it - links, buttons, form controls, etc.

davidakennedy:
Based on the feedback here, and from showing this in a UX meeting, I wanted to share a few screenshots with focus styles leaning more in the direction of the gov.uk focus states.

For buttons and links, I'm using our `$color-primary-alt-lightest` as a background, and our `$color-primary-darker` to color the bottom border.

For form fields, I'm keeping our same box shadow, but also setting an inset shadow to mimic a thicker border without changing border widths, and making content jump.

Some screenshots:

Default:

Default button

Primary:

Primary button

Transparent:

Transparent button

Danger:

Here, I wasn't feeling great about the big color change on focus, so I've included another option.

Danger button

Success:

Same with the danger button, I've included another option (pay no mind to the repeated default state in the screenshot).

Success button

Link:

Link


Inverse:

This shows focused states on a dark background.

Inverse example

Form field:

This last field is focused, the first one is not.

Form field

Some examples of it used it the wild:

Window shop page

Window shop start

Window shop household

Window shop you

Browse plans

Some questions:

* How does this feel as a direction to pursue?
* Any worries about this clashing with site packs?
* Anything else pop to mind?

melwoodard:
I like the visual harmony of these examples, but interestingly the focus state for the buttons in particular now looks less visually prominent than the default state. That last example with the plan cards also points out some contrast issues. I would think that the light blue and grey would be indistinguishable to users with low vision.

I do think the highlight and bottom border for text links works well. For buttons I think we could get away with a thick border similar to [USWDS](https://designsystem.digital.gov/components/button/), which we could then also apply to text fields, etc.

Since we don't have anything particularly bright in our color palette, I'd open it up to consideration of adding a new color that we only use for focus states, similar to gov.uk's obnoxiously bright yellow or uswds's slightly brighter blue.

line47:
I also like the visual consistency across different components like links, buttons and form elements and I think the border thickness is working well.

I would like to see a bolder color used for the background that stands out more - I'm thinking something in the yellow/orange range. 

I like the idea @melwoodard has about adding a focus state color to the current palette. This will also allow for easy overriding.

davidakennedy:
I [filed an issue](https://github.com/CMSgov/design-system/issues/558) to go along with this PR.

I also have a [demo here](https://davidakennedy.github.io/focus-styles-demo/).

In my mind, there are a few outstanding things we need to work through, beyond a review:

- [ ] Do we deprecate the `$focus-shadow-inverse` variable. In this PR, I left it, and changed the value.
- [ ] There's some content jumping on the buttons when focused at the different size variants because of padding. I need to fix that. You can see this on the clear button in the Autocomplete component.
- [ ] I'm not wild about [doing this in the Tabs CSS](https://github.com/CMSgov/design-system/pull/510/files#diff-4d4681b4bd0e838d679de19c0174e208R132), but needed to because of the high specificity. There's probably a better way.
- [ ] Any other feedback you all have.

cc: @melwoodard @line47 @bernardwang

davidakennedy:
I pushed a few commits from yesterday and today that fixed a few small issues. Also, I wanted to highlight a few other questions/challenges I found while doing a bunch of cross-browser testing today:

Autocomplete clear button

**Clear button in Autocomplete**: The right padding here on the Autocomplete clear button means it's flush with the edge when it's transparent, but then when focused, it lacks padding on the right, looking awkward. I'm thinking we add the padding back for this, but open to other options.

**Tab focus styles and borders**: I'm not wild about how the border width changes are so visible when you tab through these. I'm thinking a subtle animation might help.

**Buttons and padding**: With how I had to change the padding on focus for buttons because of the border width change (prevents content jump), if you change top or bottom padding on buttons/links styled as buttons, things get messy. There may be a better way to handle this, and I'm open to ideas. Or we could say it's up to authors.

davidakennedy:
@melwoodard Thanks for the feedback!

> The dark bottom border on buttons gets lost on an inverse background. Not a big deal, but it does make the button look shorter compared to the ones next to it that aren't focused. Is there a way we can solve for that?

I thought about this too, as I worked through things, but liked the overall consistency of staying with yellow and black in all cases. That said, the "shorter" appearance is a good point. I'll work through some options for the border on dark backgrounds.

> On focus and hover on an inverse background, the labels on the default and transparent buttons look pretty low contrast. Is that an accessibility concern?

Good catch! I'm not too concerned here since it's unlikely someone could focus and hover at the same time. But I do suppose someone could use the `ds-c-button--focus` style for something else. So I'll tighten that up to prevent it.

> The focus styles for tabs and vertical nav feel very visually prominent. Is there a way to downplay them, like highlighting just the text instead of the entire container?

Also thought about this as I worked through it. I was trying to not be too heavy handed with changes to the structure and styles. GDS does some nice trickery with [pseudo elements in its tabs](https://design-system.service.gov.uk/components/tabs/) to make this happen. I'll investigate and also discuss with @line47 and @bernardwang.

> I noticed when I opened a help drawer that the help drawer title immediately receives focus, with the browser default blue halo styling. Is that an element we'd also want to consider for custom focus styling? And if so, do we have any other focus management type situations where we point focus to a non-interactive element that we should accommodate?

Another good catch! I was doing research on this today, talking to Trevor, Jen and Dan to figure out what we should account for. I have some ideas and will handle these cases in an update.

I also added these points to the [list above](https://github.com/CMSgov/design-system/pull/510#issuecomment-557638358).

melwoodard:
@davidakennedy thanks! Re: the clear button comment you posted earlier, I wonder if only the text in transparent buttons should be highlighted, vs. the whole area? I don't know how many real-world usage examples we have for transparent buttons, but I wonder if they'd run into similar situations where the button fits well in the UI its transparent state, but giving it a solid background on focus causes some visual conflict.

line47:
I think it would be best to update the transparent button on the autocomplete component to remove the right side padding class `ds-u-padding-right--0` and its place use  `ds-u-margin-right--0`. This will achieve the same visual spacing and also allow the focus state to work consistently.

davidakennedy:
> I think it would be best to update the transparent button on the autocomplete component to remove the right side padding class ds-u-padding-right--0 on the clear search button and its place use ds-u-margin-right--0.

@line47 After looking at this, I don't think this will work, unless I'm misunderstanding. The padding on the buttons provide the internal space, and there isn't any margins on the button.

line47:
Maybe I didn't explain it well. The `ds-u-padding-right--0` class on the button removes the right side padding on the button and that gives us the focus highlight right on the edge of the text. By removing `ds-u-padding-right--0` the original padding is restored on the button so there is space for the focus background-color. 

Adding `ds-u-margin-right--0` to the button overrides the  default 10px margin on the button. 

Here are two screenshots, hopefully, this helps. 👍 

### Original button with focus bg right against text
Screen Shot 2019-12-03 at 9 51 00 AM

### Button with focus after proposed changes
Screen Shot 2019-12-03 at 9 51 19 AM

line47:
I'm wondering if we should update the styling on the autocomplete for items in the dropdown 🤔 

Something like this?
Screen Shot 2019-12-03 at 2 05 35 PM

davidakennedy:
> I'm wondering if we should update the styling on the autocomplete for items in the dropdown

@line47 I like that idea. Right now, it doesn't look like the autocomplete component differentiates between hover and focus states. The results are list items that can't be natively focused. So we'd have to add a way to items to be focused, or keep the hover and focus states the same.

davidakennedy:
Regarding these two bits of feedback from @melwoodard:

> The dark bottom border on buttons gets lost on an inverse background. Not a big deal, but it does make the button look shorter compared to the ones next to it that aren't focused. Is there a way we can solve for that?

Changing the border color is probably the easiest thing – only changing it on inverse backgrounds. Something like:

Set of buttons on dark background

> The focus styles for tabs and vertical nav feel very visually prominent. Is there a way to downplay them, like highlighting just the text instead of the entire container?

I worked up a demo for this. It's still a bit rough... especially around how the text shadow that works as the bottom border overlaps the container. Also, I'm not sure the hover color makes sense anymore when the links/buttons not being full width. However, you can still click link or button as though it was full width.

vertical-nav

I can do something similar with the tabs, so that the focus style is more on the text, and not the whole container.

Let me know what you all think.

cc: @line47

line47:
Thanks for your work on this DK! 🎉 

I think the focused inverse button looks good here and would recommend going with what you have in the screenshot. 

I'm also thinking the focus states for the vertical nav look better when the whole section is highlighted and not just the text. It follows the same pattern as the hover and I think the consistency in length looks better than the ragged edge where only the text is highlighted.

melwoodard:
> I'm also thinking the focus states for the vertical nav look better when the whole section is highlighted and not just the text. It follows the same pattern as the hover and I think the consistency in length looks better than the ragged edge where only the text is highlighted.

That's a good point. I think what's tripping me up is that when I select a link in that vertical nav on the demo page, the focus state is essentially the selected state until you focus on another element. But I assume it wouldn't necessarily work that way in a real scenario - when you select a nav link most likely the page would reload right?

line47:
@melwoodard is this what you are referring to? 
In this GIF I click the `current page` link and then move my mouse up to the top of the screen. 
The focus style is applied until I click somewhere else. 

ezgif com-video-to-gif (1)

melwoodard:
@line47 yes, that's it. I'm assuming that in practice, when I select a link in the vertical nav it'll trigger a page load to load the page I just selected, so the focus state would be pretty momentary. I'm not even sure if anyone's _using_ the vertical nav component currently, so either way my thoughts are trivial here.

davidakennedy:
> I'm assuming that in practice, when I select a link in the vertical nav it'll trigger a page load to load the page I just selected, so the focus state would be pretty momentary.

@melwoodard That's true, yes. But there are other situations where the focus style would show for longer. Toggle buttons come to mind, like ones that toggle a menu or filters. There's a CSS feature coming up that will help with this: focus-visible. It makes it easier to only show focus when navigating via keyboard. But browser support isn't great.. yet.

davidakennedy:
I added the style to a few places in our apps for context:

Screen Shot 2019-12-11 at 1 11 44 PM
Screen Shot 2019-12-11 at 1 19 26 PM
Screen Shot 2019-12-11 at 1 21 45 PM
Screen Shot 2019-12-11 at 1 22 01 PM
Screen Shot 2019-12-11 at 1 25 20 PM
Screen Shot 2019-12-11 at 1 28 32 PM
Screen Shot 2019-12-11 at 12 25 03 PM
Screen Shot 2019-12-11 at 12 34 11 PM
Screen Shot 2019-12-11 at 12 39 59 PM

bernardwang:
Quick thing, I noticed the error color on the textfield disappears on focus. https://gyazo.com/4d1f2ffeaf64bb5f83030cb79d99ac5f

Another small style question, I notice that the border radius of textfields disappear on focus. I like how the focus highlight looks without a border radius, but I wonder if its possible to keep the default border radius on the form element when focusing.

davidakennedy:
Thanks for the feedback, @bernardwang! Great questions.

> What do you think about keeping the border radius consistent on the focus highlight? Buttons have a rounded black underline when focused, but links and buttons have a normal square black underline.

We could do that for sure, if we feel like it's important. It would soften it up a bit. I left it because the other two elements that have rounded corners always have borders there, so it makes a bit more sense.

> Related to this, I noticed that the border radius of textfields changes on focus. Would it be possible to keep keep the default border radius on the form element when focusing?

We could do that. Right now, I use the outline property for that outer most outline. It can't do border radius, but we could do a double box shadow to make it work, and have the same look. I liked the subtle change because it was another small way to communicate a change of state.

> I noticed the focus style on tabs can have 2 lines (top or bottom) and includes an animation, I wonder if this could be simplified so that the line is always at the top. I also wonder if we should introduce animation considering there isn't much animation anywhere else in the DS

Yeah, I was a bit torn here. Originally, I did the focus styles on the tabs with borders and without animation. The transition from regular to focused state looked too janky. So I went with a pseudo element, similar to how the current tab looks with the top border. And it didn't feel right to have one animated but not the other.

I also thought about having the line be at the top when focused too. I'm not against it, as it might make better sense with this component. I put it at the bottom since all the other focused state lines are at the bottom, and I thought the consistency was good. However, having the two lines does make it look heavier, so totally open to going with one.

> Similar to the tab issue above, the vertical nav focus styles also has the potential for 2 different lines (bottom and left). I feel it might be simpler to just have the black line always be on the left

Same as above here, I put the line at the bottom for consistency, but am open to changing it for this component.

> I noticed the error color on the textfield disappears on focus, is this because of the focus styles?

Right. I chose to change the focus styles even if a field is in an error state or success state. The [GDS design system](https://design-system.service.gov.uk/components/error-message/) does this as well. It makes sense to me since the states are different. But we could also leave the red or green borders and make s different focus state for the error or success state.

How do they work now? As in when do these states go away?

bernardwang:
For the Tabs and Vertical nav, I don't think it necessarily has to be on the bottom for it to be consistent. If we stuck to that rule then [this case](https://gyazo.com/60a9827ad8b4665b4311efba3396b39b) should also have a bottom line, which I think could be overemphasized. Personally, I think it makes sense to consistently have 1 focus outline, on the top for Tabs, on the left for Vertical nav, and bottom everywhere else. I'm interested in hearing what others think of this!

As far as the animation, I actually didn't realize that Tab already had an animation for the line. Now I understand why you added it, but I wonder if maybe that's something the core team should first deal with. It seems weird to me to have an animation there when animation doesn't appear anywhere else in the DS like in the Vertical nav. I know there has been talks for adding animation and guidance in Slack, but I'm not sure if we are prepared to support that yet. @line47 


Everything else you said about the focus styles in error/success state makes sense to me, and I also dont have strong thoughts on the other border radius stuff. Thanks for the hard work on this @davidakennedy!

line47:
I noticed on the month picker there is a link in the example when I tab to that particular link, which has a class of `ds-u-color--inherit` the text color does not shift to base. 

See the screenshot
Screen Shot 2019-12-18 at 2 38 49 PM

line47:
For the tabs and vertical nav - I think it's Ok to follow the convention that is already in place. 
Tabs - line at the top
Nav - line at the left 

I like the idea of removing the animation and revisiting - see this PR https://github.com/CMSgov/design-system/pull/581

For the error states, I think it's fine for the focus state to take precedence over the current state (error or success)

davidakennedy:
> I noticed on the month picker there is a link in the example when I tab to that particular link, which has a class of ds-u-color--inherit the text color does not shift to base.

Good catch! To address this, @line47, I added this, and something similar to all the color utilities:

```.ds-u-color--inherit:not(:focus) {
  color: inherit !important;
}```

Let me know if that's too heavy handed.

**Comment by davidakennedy** [comment]
I took some notes to help when we release this, and to summarize things since this PR has a lot of changes, and this thread is long:

# Major changes in this PR

## Components

**Links**:
- Adds focus style with background color and bottom box shadow to links.
- Also includes a transparent outline which gets converted to a visible color in Windows High Contrast mode. Box shadow colors are ignored by Windows High Contrast mode.

**Buttons**:
- Similarily, adds focus style with background color and bottom box shadow to buttons.
- Also includes a transparent outline which gets converted to a visible color in Windows High Contrast mode.
- The outline offset helps make sure that the outline isn't bumping into the button when focused and in Windows High Contrast mode.
- Adds "focused" style button example for each button type in docs.
- Removes inner outline user agent styles in Firefox.

**ChoiceList**:
- Uses a combination of borders and box shadow to show the focus styles.
- Also includes a transparent outline which gets converted to a visible color in Windows High Contrast mode.

**Dialog**:
- Uses button styles to show the focus styles, no major changes needed.
- Show the black SVG close icon to match styles.

**Dropdown**:
- Removes inner outline user agent styles in Firefox.

**Help Drawer**:
- Adds a focus style to a heading that focus programmatically shifts to via a mixin.

**Skip Nav**:
- Removes background color so that new link focus styles can work.

**Tabs**:
- Uses the pseudo-element also used when a tab is current for the top border on focused tabs. It's kept on top, not the bottom, to match the current styles.

**TextField**:
- Uses a combination of borders and box shadow to show the focus styles.
- The slight border radius on fields in squared off on focus.

**Vertical Nav**:
- Uses the left border (like on the parent nav item) to show the focus styles.

### Other changes

Because of changes to the above components, other components have visual changes to the focus styles, including:

- Autocomplete.
- Date field
- Masked field
- Month picker

## Variables

- Adds a `$color-focus` variable, the bright yellow.
- Changes the `$focus-shadow` to an inset shadow, mostly used on form fields.
- Adds a `$focus-shadow-link` variable, used in links and a mixin for adding focus styles to other elements.
- Changes the `$color-focus-inverse` color to the same bright yellow as `$color-focus` variable.
- Changes `$focus-shadow-inverse` shadow properties.
- Adds a `$focus-border-inverse` variable, used on inverse backgrounds at the bottom of buttons and links.
- Adds a `$focus-shadow-link-inverse` used on inverse backgrounds with links.

## Mixins

- Adds a `@mixin focus-text` for adding focus styles to other elements if needed. Currently, this is used to add focus styles to a heading element in the help drawer after focus is shifted to it.

## Utilities

**Color**:
- Adds a `:not(:focus)` rule to ensure that color changes made with color utility classes don't override focus styles, and create unintended or inaccessible color combinations. 

## Why

In this pull request, we add focus styles in the same way that [exist in the Gov.uk design system](https://design-system.service.gov.uk/get-started/focus-states/). The bright yellow background with black text and black border stands out on light backgrounds, and the border makes sure that we don't rely on color alone.

On darker backgrounds, the bright yellow background still works well, but we change the border color to a more gold yellow tone. It maintains proper contrast while still standing out against the dark background.

This gives the design system a foundation for consistent focus styles across all elements and components without relying on browser styles alone.

**Comment by bernardwang** [comment]
@davidakennedy One issue I ran into is that with the way our CSS is setup, `focus` always overrides `active`, meaning we effective don't have an `active` style anymore. Ill try and work on a solution next week

**Comment by davidakennedy** [comment]
> One issue I ran into is that with the way our CSS is setup, focus always overrides active, meaning we effective don't have an active style anymore. Ill try and work on a solution next week

@bernardwang I worked on this yesterday and today... I did some slight refactoring:

- Reordered pseudo classes to work better in most cases: visited, hover, focus, active
- Added :active:focus styles to make sure :active state can be viewed in Chrome when an element is also focused.

It seems that only Chrome shows focus styles at the same time as active ones. Firefox and Safari don't do that. I need to test some more in Edge.

Take a look, and let me know what you think. Happy to take another approach if there's a cleaner way.

**Comment by bernardwang** [comment]
@davidakennedy Those changes seem to be working well for me on Chrome, I'll be able to finish testing tomorrow!
One additional thing we should handle is the focus style on inverse text.
https://gyazo.com/076da52d912857ea6afc4f7f2fd8fd2c

**Comment by davidakennedy** [comment]
> One additional thing we should handle is the focus style on inverse text.

@bernardwang Added a fix for this, using an additional mixin.

File: packages/support/src/tools/_mixins.focus-text.scss [hunk 2]
```diff
@@ -4,3 +4,10 @@
     color: $color-base;
     outline: 3px solid transparent
 }
+
+@mixin focus-text-inverse {
```
bernardwang: Ah I just noticed this mixin file. Is the intention to allow others to use these mixins to achieve focus styles? Because if this is just to promote code reuse, I feel like the they are more appropriate as [sass placeholders](https://sass-lang.com/documentation/style-rules/placeholder-selectors)

davidakennedy: @bernardwang Oh cool. I haven't used placeholders much. I don't think you can use variables in placeholders, so that would be a blocker. I do like that the CSS doesn't output unless it's used.

The use case here is mostly for those times, like in the Help Drawer, when you need another element to take focus that normally wouldn't. In the Help Drawer, it's the heading after the link is clicked to open it.

bernardwang: Ah I gotcha, if its intended to be used in other cases then I think a mixin is appropriate, I just wasnt aware of other use cases. 

Re: placeholders, I think they do allow sass variables, you just cant pass variables in for dynamic styles. Placeholder is also slightly more efficient with generated css, preventing duplication of CSS.

packages/core/src/components/VerticalNav/VerticalNav.scss
```diff
@@ -1,5 +1,19 @@
 @import '@cmsgov/design-system-support/src/index';
 
+%nav-item-hover {
```
bernardwang: Previously the hover would override the focus, but thats not consistent with how focus works on `Tab` and other components. This commit makes the focus work the same across those components

packages/core/src/components/Button/button.example.html
```diff
@@ -68,19 +68,19 @@ <h6 class="preview__label">Success button</h6>
...
+  <span class="ds-c-button--transparent-text">Hover</span>
 </button>
 <button type="button" class="ds-c-button ds-c-button--transparent ds-c-button--active">
-  Active
+  <span class="ds-c-button--transparent-text">Active</span>
 </button>
 <button type="button" class="ds-c-button ds-c-button--transparent ds-c-button--focus">
-  Focus
+  <span class="ds-c-button--transparent-text">Focus</span>
 </button>
```
davidakennedy: @bernardwang Seems to be causing some display differences in both Firefox and Safari (MacOS). Haven't tested on Windows.

Yellow background and black text on button with the yellow background not extending the full height and width.

bernardwang: Which differences are you seeing? I took a look at FF/safari/IE11 and they look similar to the screenshot you have above

davidakennedy: @bernardwang Sorry if I wasn't clear. See how in my screenshot above, the background doesn't any padding on the right or left. But if other browsers, it does.

Is it displaying different for you?

line47: I'm seeing the padding just fine in Chrome, Safari, Firefox and IE11

bernardwang: @line47 @davidakennedy To clarify, the padding in the screenshot above is the same for every button. The only difference with the transparent button is that there is an additional span that is styled to have a highlighted background. https://gyazo.com/0e7d6bf55d32b93d39bc2ae02585dd1b

bernardwang: I originally changed the transparent button focus because the transparent button is styled like a link, I thought it made sense for the focus styles to also be styled like a link. This brings up the point of why we provide a button that looks like a link in the first place, but my rationale is that if we do provide it, it should be consistent.

I'm willing to go back to the original styling if this doesnt make sense however. Would it make sense to get design input on this?

Before: 
ff63e1cb0766e100892176c6510852e6
After: 
01cff3c090a8a24698ffc579b97d59ca

bernardwang:
I originally changed the transparent button focus because the transparent button is styled like a link, I thought it made sense for the focus styles to also be styled like a link. This brings up the point of why we provide a button that looks like a link in the first place, but my rationale is that if we do provide it, it should be consistent.

I'm willing to go back to the original styling if this doesnt make sense however. Would it make sense to get design input on this?

Before: 
ff63e1cb0766e100892176c6510852e6
After: 
01cff3c090a8a24698ffc579b97d59ca

line47:
@melwoodard can you weigh in on the above?  https://github.com/CMSgov/design-system/pull/510#issuecomment-577406960

packages/support/src/settings/_variables.color.scss
```diff
@@ -74,8 +74,14 @@ $color-base-inverse: $color-white !default;
...
 // Focus
-$color-focus: #3e94cf !default;
-$color-focus-inverse: #59bcff !default;
+$color-focus: #ffde11 !default;
+$color-focus-inverse: #ffde11 !default;
+$focus-border-inverse: #8f7c00 !default;
+$focus-shadow: inset 0 0 0 1px $color-base;
+$focus-shadow-inverse: inset 0 0 0 1px $color-base !default;
+$focus-shadow-link: 0 3px $color-base;
+$focus-shadow-link-inverse: 0 3px $focus-border-inverse !default;
```
line47: I'm thinking we should use the following format for focus styles
- $focus-NAME-NAME

For example, 
- $focus-color
- $focus-color-inverse
- $focus-border-color-inverse
- $focus-shadow
- etc...

bernardwang: i like this

davidakennedy: @line47 @bernardwang 
> I'm thinking we should use the following format for focus styles

I went ahead and [made this change here](https://github.com/CMSgov/design-system/pull/510/commits/9635e951a46cbad3ba5c1ae3700fcdd5047affdc).

bernardwang: @line47 I think these changes have been made, could you update your review?

line47:
Added focus colors to the color palette page. Feel free to make edits, suggestions, etc...

Screen Shot 2020-02-03 at 4 08 21 PM

packages/core/src/utilities/background-color/background-color.scss
```diff
@@ -260,3 +260,14 @@
 .ds-u-fill--transparent {
   background-color: transparent !important;
 }
+
+// Focus colors
+.ds-u-fill--focus-color {
```
bernardwang: @line47 Do you think we need to add fill utilities for focus? I feel like its not necessary to provide utilities for every color, and I cant think of many use cases for this utility

packages/support/src/settings/_variables.color.docs.scss
```diff
@@ -117,6 +117,36 @@ Markup:
...
+
+These colors are used when an item on the page has focus.
+
+@hide-markup
+
+Markup:
+<div class="">
+  <article class="ds-u-margin-bottom--1">
+    <div class="c-swatch__preview c-swatch__preview--condensed ds-u-radius--circle ds-u-fill--focus-color"></div>
+    <code>focus-color</code>
```
bernardwang: these css variables are missing `$`

packages/support/src/settings/_variables.color.docs.scss
```diff
@@ -117,6 +117,36 @@ Markup:
 Style guide: style.color.status
 */
 
+/*
+Focus colors
```
bernardwang: What do you think about just putting the focus colors in the "Additional colors" section rather than its own? I'm thinking how there are other state colors like `$color-visited` which should also fall into the same category as the focus colors, and wouldn't necessarily deserve their own section

packages/core/src/utilities/background-color/background-color.scss
line47: Currently, that's how the palette is generated. I'm open to suggestions for changing it

bernardwang: hm I see. If thats the case we should also include it in the background color utilities page.

bernardwang:
@line47 I pushed changes that made both the [color palette](https://gyazo.com/cb944303fc9c4451f45019430668b9df) and [background color](https://gyazo.com/af4516c4f2f33490bdc5de01f790c753) page show the new focus colors.  I think this is good for now and should not block the PR (in retrospect my own feedback could have waited as well).

One thought I had about colors going forward, we should eventually rethink which utilities we want to provide vs and how we organize our color palette vs background color page. Especially considering `focus-color` and `focus-color-inverse` are the same. I think some future discovery work into design tokens would help us clarify and better organize context specific styles (i.e. `$color-error`, `$focus-color`) vs base css variables (i.e. `$color-red`, `$color-muted-yellow-highlight`).

packages/core/src/components/Button/Button.example.jsx
```diff
@@ -22,7 +22,7 @@ ReactDOM.render(
     <Button className="ds-u-margin-right--1" disabled>
       Button with `disabled` prop
     </Button>
-    <Button className="ds-u-margin-right--1" href="javascript:void(0);">
+    <Button className="ds-u-margin-right--1" variation="transparent" href="javascript:void(0);">
```
bernardwang, line 25: was it necessary to add the transparent variation?

PR #511, opened 2019-10-11
Update footer to add contribution links 

line47:
Updated the footer to add GitHub contribution links for 
- reporting bugs
- requesting changes
- or proposing something new

I also needed to add styles for when links are used on the `.ds-base--inverse ` background. 

### Screenshot
localhost_3000_ (1)


mliupaeter:
This looks good to me. Thanks for adding.

melwoodard:
@line47 to give this section a bit more prominence and reduce the muddiness between the grey and light blue backgrounds, what about giving the contribution section a bold blue background like `primary` or `primary-darker` and reversing the text to white?

line47:
@melwoodard here are screenshots of `primary` and `primary darker`. 

### primary
localhost_3000_

### primary darker
localhost_3000_ (2)

benkutil:
Something to consider: does trying to combine a new color with the light blue & dark blue already at the bottom cause a challenge with highlighting the "contribution" links.

line47:
Update to combine the bottom links footer and very bottom tagline to use the same background. 
Looks like we need to add some styling for links to be included in the inverse bg. 😄 

Screen Shot 2019-10-15 at 2 57 46 PM

line47:
Updates to the footer

Screen Shot 2019-10-16 at 10 27 35 AM

packages/core/src/base/body.scss
```diff
@@ -37,5 +37,15 @@ Style guide: style.base
   @extend %ds-base;
   background-color: $color-background-inverse;
   color: $color-base-inverse;
```
line47: I needed to add styles for links that are used in a `.ds-base--inverse` section. 
I thought this was the best place to add them unless folks think these styles should be in with the `typography` styles

packages/docs/src/scripts/components/Footer.jsx
```diff
@@ -32,63 +32,102 @@ function renderLinks(links) {
...
+              className="ds-c-button"
+              href={githubUrl(
+                'issues/new?assignees=&labels=&template=general-issue.md&title='
+              )}
+              target="_blank"
+            >
+              Request a change
+            </a>
+          </div>
+          <div className="ds-l-col--12 ds-l-sm-col--4 ds-l-lg-col--4">
```
bernardwang: The other 2 buttons both have `ds-u-margin-bottom--2`. Should this one also have the class, or can we remove the margin for all the buttons?

packages/docs/src/scripts/components/Footer.jsx
```diff
@@ -32,63 +32,102 @@ function renderLinks(links) {
...
+      <section className="ds-base--inverse">
+        <div className="ds-l-container ds-u-padding-y--4">
+          <div className="ds-l-row">
+            <div className="ds-l-col--12 ds-l-sm-col--4 ds-l-lg-col--4 ds-u-margin-bottom--4 ds-u-margin-top--2">
+              <h6 className="ds-h4">Contact Us</h6>
+              <p className="ds-u-font-size--small">
+                Need help with the CMS Design System?
+                <br />
+                <a href="https://forms.cms.gov/cms-wds-design-system-contact-form/responses/new">
+                  Drop us a line{' '}
```
bernardwang: Whats the `{' '}` for?

packages/docs/src/scripts/components/Footer.jsx
```diff
@@ -32,63 +32,102 @@ function renderLinks(links) {
...
+                </a>
+              </p>
+            </div>
+            <dl className="ds-l-col--12 ds-l-sm-col--4 ds-l-lg-col--4 ds-u-margin-bottom--4">
```
bernardwang: I'm looking at the `ds-u-margin-bottom--4` styling for these `dl` elements and I'm also wondering if we should only have the margin present when the lists are collapsed on smaller breakpoints. I know that this is the existing style, but you can see in this screenshot there's quite a bit of margin now that that the address is moved to the bottom

https://gyazo.com/d100b47b8d66bdaddfa4b11b5207a089

packages/docs/src/scripts/components/Footer.jsx
```diff
@@ -32,63 +32,102 @@ function renderLinks(links) {
...
+      </dd>
     );
   });
 }
 
 const Footer = () => {
   return (
-    <footer className="ds-u-fill--primary-alt-lightest ds-u-padding-y--6 ds-u-md-padding-bottom--3">
-      <div className="ds-u-padding-x--3 ds-u-margin-x--0">
+    <footer className="ds-u-fill--gray-lightest">
```
bernardwang: Removing the padding means that the menu button covers up some text, maybe we should add the padding back for whatever breakpoint the menu button appears.
Screen Shot 2019-10-16 at 4 54 54 PM

Its also worth noting the menu now has the same color as the footer and theres no contrast difference at all. I'm not sure if we should maybe change the menu button color? For reference heres how it looked originally
Screen Shot 2019-10-16 at 4 54 46 PM
Screen Shot 2019-10-16 at 4 56 35 PM

bernardwang:
I added @davidakennedy as a reviewer, theres some markup changes that might be good to review!

bernardwang:
oops ignore that accidental close

packages/docs/src/scripts/components/Footer.jsx
line47: This is not needed, removing...

packages/docs/src/scripts/components/Footer.jsx
line47: I've updated the padding/margin to better accomodate for the spacing

packages/docs/src/scripts/components/Footer.jsx
line47: I've adjusted the footer spacing and changed the menu button color

PR #512, opened 2019-10-11
[WNMGDS-242] Standardize ref props

bernardwang:
This PR establishes `inputRef` as the general name for accessing HTML elements of form components (i.e. Button, TextField etc).

The notable exceptions are `DateField` (which have 3 refs `dayFieldRef`, `monthFieldRef`, `yearFieldRef`), `ChoiceList` and `MonthPicker`. I considered updating `DateField` props to `dayInputRef`, `monthInputRef` and `yearInputRef` but I wasn't sure if I should. `ChoiceList` doesn't have a ref prop, but `Choice` does. I was considering adding an `inputRefs` prop that allows someone to keep track of a list of `Choice` refs but I wasn't sure if that was too complicated. 

### Changed
`Button`, `TextField` and `Dropdown` now have `inputRef` props, and the old versions have been deprecated

### How to test
Add `inputRef={(elem) => console.log(elem)}` to the various component to ensure they are working. Also check if the old name causes a console error to show


packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -45,7 +45,6 @@ export class Autocomplete extends React.PureComponent {
     this.listboxId = uniqueId('autocomplete_owned_listbox_');
     this.listboxContainerId = uniqueId('autocomplete_owned_container_');
     this.listboxHeadingId = uniqueId('autocomplete_header_');
-    this.loader = null;
```
bernardwang: unused

packages/core/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -114,6 +113,7 @@ export class Autocomplete extends React.PureComponent {
           autoComplete: this.props.autoCompleteLabel,
           focusTrigger: this.props.focusTrigger,
           id: this.id,
+          inputRef: this.props.inputRef,
```
bernardwang: `Autocomplete` just passes certain props to the child `TextField`.

packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -25,6 +25,7 @@ export class Choice extends React.PureComponent {
       );
     }
 
+    this.input = null;
```
bernardwang: `this.input` was already used internally in `Choice`, I just defined it here for extra clarity

bernardwang:
@1Copenut @line47 
Another discussion I think is worth having is deciding which components should have a `focusTrigger` prop. Right now most form components have it, but it's also not clear to me which components should have it. I.e., `Alert` sounds like it should have a `focusTrigger`

packages/core/src/components/Button/Button.jsx
```diff
@@ -13,6 +13,16 @@ import classNames from 'classnames';
  * you could pass in a `target` prop to pass to the rendered anchor element.
  */
 export class Button extends React.PureComponent {
+  constructor(props) {
+    super(props);
+
+    if (props['buttonRef']) {
+      console.error(
+        `[Deprecated]: Please remove the React property 'buttonRef' for the <Button> component. It is no longer supported and will be removed in a future release, use 'inputRef' instead`
```
pwolfert: Nice 👌

packages/core/src/components/ChoiceList/Choice.jsx
```diff
@@ -121,10 +122,10 @@ export class Choice extends React.PureComponent {
           className={inputClasses}
           id={this.id}
           onChange={this.handleChange}
-          ref={input => {
-            this.input = input;
+          ref={ref => {
```
pwolfert: 👍

packages/core/src/components/Dropdown/Dropdown.jsx
```diff
@@ -78,14 +89,9 @@ export class Dropdown extends React.PureComponent {
...
-          ref={ref => {
-            if (focusTrigger) {
-              this.selectRef = ref;
-            }
-            if (fieldRef) {
-              fieldRef(ref);
-            }
-          }}
+          ref={
+            focusTrigger ? ref => (this.focusRef = ref) : inputRef || fieldRef
```
pwolfert: This line is more confusing to read than what was there before. There's a ternary, an anonymous function with an assignment statement for its return value as well as an OR operator. My eyes can't easily resolve what's happening.

packages/core/src/components/TextField/TextField.jsx
```diff
@@ -119,7 +124,7 @@ export class TextField extends React.PureComponent {
         className={fieldClasses}
         id={this.id}
         /* eslint-disable no-return-assign */
-        ref={focusTrigger ? loader => (this.loader = loader) : fieldRef}
+        ref={focusTrigger ? ref => (this.focusRef = ref) : inputRef || fieldRef}
```
pwolfert: Whatever is done to the address the readability of this property in `Dropdown` could be done here as well

packages/core/src/components/Dropdown/Dropdown.jsx
bernardwang: Good point, hows this instead?
```
          ref={ref => {
            if (focusTrigger) {
              this.focusRef = ref;
            } else {
              if (inputRef) {
                inputRef(ref);
              }
              if (fieldRef) {
                fieldRef(ref);
              }
            }
          }}
```

pwolfert: Yep, that looks good to me 👍 It's certainly longer, but the logic is clearer to me.

PR #515, opened 2019-10-14
[WNMGDS-250] Remove max and min from TextField prop documentation

bernardwang:
### Summary
We pass down any custom TextField props down to the input element, including `max`, `min`, `maxLength`, `minLength` etc. Since we don't do anything special to these attributes there's no reason to specifically call out `min` and `max` in our documentation. This PR removes them.

Note: we should consider adding extra documentation on the recommended way to do validation and user input requirements with our React form components

### Removed
`max` and `min` prop documentation


line47:
Do we need to explain this in the documentation or no?  
> We pass down any custom TextField props down to the input element, including max, min, maxLength, minLength etc

bernardwang:
@line47 The only reason why I didnt is because thats a default React behavior, and it also applies to all the other form components we have like `ChoiceList`. I feel it would be better to include guidance around validation, and mentioning how most of our form components accept native html validation attributes, but since they are abstractions above html elements it's not recommended to use them. 

Since that guidance isnt available yet, I'm still open to adding something small in the meantime! I'm just not sure where it would go...

line47:
That makes sense and I agree about the validation guidance

packages/core/src/components/TextField/TextField.jsx
```diff
@@ -234,17 +233,6 @@ TextField.propTypes = {
...
-   * `max` HTML input attribute
-   */
-  max: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
-  /**
-   * `min` HTML input attribute
-   */
-  min: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
-  /**
-   * Whether or not the text field is a multiline text field
-   */
```
dcloud: Did you intend to remove the docstring for the multiline prop as well?

bernardwang: oh oops! thanks for the catch

PR #517, opened 2019-10-15
[WNMGDS-158] Update site package documentation

line47:
Updated site package documentation to explain how to take a site package from a separate repo and place it in the CMS Design System to get the doc site to build. 😓


packages/docs/src/pages/startup/site-package.md
```diff
@@ -70,11 +70,30 @@ The documentation site should load at the address: http://localhost:3000/
 
 The CMS Design System supports the ability to preview your project's theme in the context of the design system's documentation site. This can be a convenient way to preview how your project's Sass affects the existing component styles.
 
-* Run **`yarn start:theme`** to generate documentation and preview your theme's styles. A `docs` sub-directory will be placed in your theme's directory.
-* Run **`yarn build:theme`** to compile the documentation site for your theme. A `docs` sub-directory will be placed in your theme's directory.
-* **Note**:
-  - If you have multiple themes inside of `packages/themes`, you can specify which theme to run by passing the name of the folder. For example, `yarn start:theme my-theme-folder-name`
-  - If your documentation site will be uploaded to a subdirectory like `example.com/design-system`, you can set the root path by passing the `--root` option. For example, `yarn build:theme --root design-system`
+1. Run `yarn start:theme` to generate documentation and preview your theme's styles. A `docs` sub-?directory will be placed in your theme's directory.
```
bernardwang: sub?-directory looks like typo

packages/docs/src/pages/startup/site-package.md
```diff
@@ -70,11 +70,30 @@ The documentation site should load at the address: http://localhost:3000/
...
+1. Run `yarn build:theme` to compile the documentation site for your theme. A `docs` sub-directory will be placed in your theme's directory.
+
+**Note**:
+* If you have multiple themes inside of `packages/themes`, you can specify which theme to run by passing the name of the folder. For example, `yarn start:theme my-theme-folder-name`
+* If your documentation site will be uploaded to a subdirectory like `example.com/design-system`, you can set the root path by passing the `--root` option. For example, `yarn build:theme --root design-system`
+
+### Preview a site package from another GitHub repo
```
bernardwang: this is just supposed to be the instructions for how to run a site package from a different repo right? i feel like the language of previewing another site package is a little confusing to me. i wonder if instead of adding another section about how to copy over a site package and run it, we should just make it more clear that site packages need to be in the themes folder to be built and run. i think right now the idea of previewing another site package sounds like a different step or activity, even really it’s the same as running ur own site package

line47: Yeah it is - thanks for the thoughts. I'll rework this section a bit.

line47: Updated to remove a bunch of these steps and add an additional note to the `Previewing your site package` section

packages/docs/src/pages/startup/site-package.md
```diff
@@ -68,13 +68,15 @@ The documentation site should load at the address: http://localhost:3000/
...
-  - If your documentation site will be uploaded to a subdirectory like `example.com/design-system`, you can set the root path by passing the `--root` option. For example, `yarn build:theme --root design-system`
+1. Run `yarn start:theme` to generate documentation and preview your theme's styles. A `docs` sub-directory will be placed in your theme's directory.
+1. Run `yarn build:theme` to compile the documentation site for your theme. A `docs` sub-directory will be placed in your theme's directory.
+
+**Note**:
+* If you have multiple themes inside of `packages/themes`, you can specify which theme to run by passing the name of the folder. For example, `yarn start:theme my-theme-folder-name`
+* If your documentation site will be uploaded to a subdirectory like `example.com/design-system`, you can set the root path by passing the `--root` option. For example, `yarn build:theme --root design-system`
+* If you are wanting to preview your theme from another GitHub repo copy the site package files into the packages > themes directory.
```
bernardwang: Sorry, quick reword, how about this:

If you want to preview your a theme from another Github repo, copy the site package files into the `packages/themes` directory

PR #518, opened 2019-10-15
Add word-break utility class

hannaliebl:
### Added
Added a word break utility class, `.ds-u-word-break`.

I wrote a style like this for our shared component library for plancompare/windowshop (long drug names on plan cards, for example, won't wrap properly without a class like this.) And then I thought maybe a utility class in the design system would be useful.

I am open to suggestions about whether this belongs here, whether it should be called something different, and if I'm describing this accurately.

Here it is applied in Chrome:
Screen Shot 2019-10-15 at 7 10 46 PM

Here it is un-applied in Chrome:
Screen Shot 2019-10-15 at 7 10 29 PM

For what it's worth, [Bootstrap calls this type of utility word break but the class text break](https://getbootstrap.com/docs/4.3/utilities/text/#word-break). They use `word-break: break-word`, but this actually doesn't work in IE 11. (there's an outstanding [PR in Bootstrap](https://github.com/twbs/bootstrap/pull/29325) to fix this.) So I think this is the way to make this work across browsers.


line47:
Thanks for adding this @hannaliebl !

PR #524, opened 2019-10-18
[WNMGDS-266] Copy fonts and img in core build

bernardwang:
https://jira.cms.gov/browse/WNMGDS-266
Currently the dist folder generated for design-system-core doesn't include assets. Especially if we want to make the design system available as simply a zip of dist files, we should process and include them

### Changed
Added gulp task for copying over font and images from `src` to `dist` in the `core` package.


bernardwang:
I got this to work by overridding the default path for images/fonts in the docs folder. This is because the `docs` folder has a different folder structure than `core`. I still need to test this with themes

bernardwang:
@line47 After struggling a bit with this ticket, I feel like future build process improvement work might be better spent on separating the build process rather than trying to incrementally improve it. The changes I made here where mainly so that I can complete https://jira.cms.gov/browse/WNMGDS-267. However, I think the changes also just add more complexity to the build situation which ideally would be redone anyway. For example, if we redid the doc site with jekyll the build process would be easily handled by jekyll. If we decided to get rid of `support` and `layout` packages it would simplify the gulp tasks for building the design system package.

line47:
@bernardwang I think that makes sense and I agree. 
Is this ready for review?

bernardwang:
@line47 yeah this is ready to go!

line47:
It looks like it is working however for copying fonts and images when running `yarn build` to build the core

bernardwang:
I actually need to revisit this. The fonts are loading correctly but I think the icons are not

bernardwang:
Backlogged this ticket

PR #525, opened 2019-10-21
[WNMGDS] - Change doc site block quote styling

line47:
### Changed
Changes the styling of the blockquote to look more like a typical blockquote styling and less like an alert. Made the text italic as well as adding in a quote icon via the content property

#### Before screenshot
Screen Shot 2019-10-21 at 12 22 33 PM

#### After screenshot
Screen Shot 2019-10-21 at 12 29 41 PM


bernardwang:
I like the increase in the left border width. About the grey bg, I wonder if it's sufficient to just have a transparent bg. This would deemphasize the block quote so its less "alerty" without introducing a new grey bg color. I also don't have a strong opinion on the quote styling

line47:
Updated screenshot 
Screen Shot 2019-10-23 at 11 52 35 AM

bernardwang:
I like that! Could we also use the primary blue on it?

line47:
Which blue 😂 ? https://design.cms.gov/style/color/

melwoodard:
I like the grey, it's pretty typical of block quote styling

line47:
I prefer grey to blue as well 

1. We don't use grey in alert styling 
2. It's not too intrusive to the text flow of the page

packages/docs/src/styles/base/_typography.scss
```diff
@@ -1,8 +1,23 @@
...
+  border-left: 4px solid $color-gray-light;
+  font-family: $font-sans;
+  font-style: italic;
   margin-left: 0;
   padding: $spacer-2;
+  padding-left: $spacer-6;
+  position: relative;
+
+
+  &::before {
```
bernardwang: This is a bit tricky because screen readers will read out the quotation mark and it will be an additional element to move through before reaching the quote. Normally something like this could be put in a `span` with aria hidden, but since this is just css im not sure if its possible to hide it. 
@davidakennedy 
Screen Shot 2019-10-24 at 11 09 58 AM

davidakennedy: @line47 @bernardwang Yeah, most modern screen reader and browser pairings will announce this. While it's not ideal, it's not horrible either.

It would be great to hide it, but the only reliable way I can think of is to put the generated content on a `<span aria-hidden="true">` like that. But that can be a pain if you're just writing markdown files.

I think it's fine to leave it as is, but we should be mindful of how much we use CSS generated content in the design system and on docs so we don't overdo it.

line47:
I'm good with leaving it as is for now to get this merged @bernardwang ok with you?

bernardwang:
@line47 yep if its fine to announce then its good with me!

PR #526, opened 2019-10-21
[WNMGDS-240] - Adding linting for Markdown and HTML that happens on lint-staged

line47:
### Added
Added linting to Markdown and HTML files that happens during lint-staged when committing automagically.

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,6 +109,10 @@
       "eslint --fix",
       "git add"
     ],
+    "*.{md,html}": [
+      "prettier --write",
+      "git add"
+    ],
     "*.scss": [
       "stylelint",
       "stylelint --syntax=scss"
--- a/packages/core/src/components/Button/button-inverse.example.html
+++ b/packages/core/src/components/Button/button-inverse.example.html
@@ -4,5 +4,7 @@
   <button class="ds-c-button ds-c-button--primary">Button</button>
   <button class="ds-c-button ds-c-button--danger">Button</button>
   <button class="ds-c-button ds-c-button--success">Button</button>
-  <button disabled class="ds-c-button ds-c-button--disabled-inverse">Button</button>
+  <button disabled class="ds-c-button ds-c-button--disabled-inverse">
+    Button
+  </button>
 </div>
--- a/packages/core/src/components/Dropdown/Dropdown.example.html
+++ b/packages/core/src/components/Dropdown/Dropdown.example.html
@@ -1,4 +1,6 @@
-<label class="ds-c-label ds-u-margin-top--0" for="options">Dropdown label</label>
+<label class="ds-c-label ds-u-margin-top--0" for="options"
+  >Dropdown label</label
+>
 <select class="ds-c-field" name="options" id="options">
   <option value="1">Option 1</option>
   <option value="2">Option 2</option>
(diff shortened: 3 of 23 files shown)


package.json
```diff
@@ -109,6 +109,10 @@
       "eslint --fix",
```
bernardwang: While testing this PR I found out that there's some conflicts with the rules between `prettier` and `eslint` for JS. I.e., `prettier` will add spaces after function definitions and then `eslint` will remove it right after. We might want to consider adjusting our rules in the future
https://stackoverflow.com/questions/44690308/whats-the-difference-between-prettier-eslint-eslint-plugin-prettier-and-eslint

packages/core/src/components/Review/review.example.html
```diff
@@ -20,16 +28,21 @@ <h3 class="ds-text ds-u-margin-bottom--0 ds-u-font-weight--bold ds-u-display--in
...
+      class="ds-text ds-u-margin-bottom--0 ds-u-font-weight--bold ds-u-display--inline-block"
+    >
       States Lived
     </h3>
     <div>
       <ul>
         <li>Ohio</li>
         <li>Idaho</li>
         <li>Iowa</li>
+      </ul>
```
bernardwang: This was a missing `ul` caught by prettier :)

packages/core/src/components/Button/button-inverse.example.html
```diff
@@ -4,5 +4,7 @@
   <button class="ds-c-button ds-c-button--primary">Button</button>
   <button class="ds-c-button ds-c-button--danger">Button</button>
   <button class="ds-c-button ds-c-button--success">Button</button>
-  <button disabled class="ds-c-button ds-c-button--disabled-inverse">Button</button>
+  <button disabled class="ds-c-button ds-c-button--disabled-inverse">
+    Button
+  </button>
```
line47: I wonder why this button is treated differently than the rest here?

bernardwang: I think its the length of the classes. Prettier will add line breaks if the line is too long

packages/core/src/components/HelpDrawer/help-drawer.example.html
```diff
@@ -1,24 +1,65 @@
...
     <a href="javascript:void(0);">Toggle the help drawer.</a>
   </span>
   <div class="ds-c-help-drawer">
     <div class="ds-c-help-drawer__header">
-      <div class="ds-u-fill--gray-lightest ds-u-padding--2 ds-u-display--flex ds-u-align-items--start">
-        <h3 tabindex="0" class="ds-u-text--lead ds-u-margin-y--0 ds-u-margin-right--2">Help Drawer Title</h3>
-        <button class="ds-c-button ds-c-button--secondary ds-c-button--small ds-u-margin-left--auto" type="button">Close</button>
+      <div
+        class="ds-u-fill--gray-lightest ds-u-padding--2 ds-u-display--flex ds-u-align-items--start"
+      >
```
line47: I personally find this breakdown where the div is broken into 3 lines harder to read, I'm wondering if we can adjust the formatting a bit in a future issue

line47:
This looks good to me 👍

packages/core/src/components/HelpDrawer/help-drawer.example.html
bernardwang: We can adjust it here: https://prettier.io/docs/en/options.html#print-width
would you want to include it in this PR?

line47: We can address it later on

packages/core/src/components/Dropdown/Dropdown.example.html
```diff
@@ -1,4 +1,6 @@
-<label class="ds-c-label ds-u-margin-top--0" for="options">Dropdown label</label>
+<label class="ds-c-label ds-u-margin-top--0" for="options"
+  >Dropdown label</label
```
bernardwang: This is another weird thing that prettier does... I've never seen this formatting before

bernardwang: I think it might be fixed with some new lines though. If that works I'll commit it

packages/core/src/components/HelpDrawer/help-drawer.example.html
bernardwang: Would you be opposed to handling it now if its quick? I'm just thinking it will be touching all of these same files again anyway if we bump up the print width to 100 for example

line47: I don't know that the print width will really solve the issue. I'm thinking it would crop up again anyways. We can leave it as is for now.

Issue #529, opened 2019-10-23
Links embedded in a fieldset > legend not reading correctly with JAWS + IE11


eppixdev:
**Describe the bug**
When passing in a link to `hint` as a prop in `ChoiceList`, `TextField`, or `FormLabel` (maybe some others?) JAWS + IE11 won't read the link text because it's nested in the legend element.

Related:
https://www.powermapper.com/tests/screen-readers/labelling/fieldset-links/

**To Reproduce**
1. Pass link as a prop to one of the above Components

**Expected behavior**
Screen readers to read the link text

**Screenshots**
**Current (`ChoiceList`):** 
```html
<fieldset>
  <legend>
    <span>
      Some hint text here
    </span>
    <a href="link">Click here to learn more</a>
  </legend>
  ... choices
</fieldset>
```

**Possible Fix:**
Moving the link out of the legends fixes the issue AFAIK.

```html
<fieldset>
  <legend>
    <span>
      Some hint text here
    </span>
  </legend>
  <a href="link">Click here to learn more</a>
  ... choices
</fieldset>
```

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**
Originally reported by @1Copenut here: https://jira.cms.gov/browse/MEA-215

1Copenut:
@eppixdev Yes, moving the `<a>` out of the `<legend>` and into its own space, possibly in a div that sits between the legend and the div containing your checkboxes / radio buttons should work well. You may have to add `aria-labelledby` attributes to the inputs themselves, and IDs to each label and the legend to get all screen reader pairings to respond correctly.

Recommend testing this solution with several pairings:
* IE11 + JAWS (Win10)
* Chrome + JAWS (Win10)
* Firefox + NVDA (Win10)
* Safari + VoiceOver (MacOS)
* Mobile Safari + VoiceOver (iOS, ideally)

```html
<form>
  <legend id="unique-legend-id">Your legend text</legend>
  <div>
    <a href="#">Trigger to open the help drawer</a>
  </div>
  <div>
    <input
      aria-labelledby="unique-legend-id unique-label-id-1"
      id="checkbox-1"
      name="option-1"
      type="checkbox" 
    />
    <label for="checkbox-1" id="unique-label-id-1">Checkbox one label</label>
    ...
  </div>
</form>
```

bernardwang:
Tracked here: https://jira.cms.gov/browse/WNMGDS-281

[...]

PR #533, opened 2019-10-28
[WIP WNMGDS-281] Add initial custom form label

bernardwang:
From this discussion: https://adhoc.slack.com/archives/C645GLLNN/p1571929043180500
We needed a way to pass a value to `FormLabel`'s `labelClassName` prop while using `ChoiceList`. Rather than add more props `ChoiceList` to pass through to `FormLabel`, we allow a custom label element by adding a `formLabel` prop

### Added
`formLabel` prop for supplying a custom label element

### DIFF
--- a/packages/core/src/components/ChoiceList/ChoiceList.example.jsx
+++ b/packages/core/src/components/ChoiceList/ChoiceList.example.jsx
@@ -1,5 +1,6 @@
+import React, { Fragment } from 'react';
 import ChoiceList from './ChoiceList';
-import React from 'react';
+import FormLabel from '../FormLabel/FormLabel';
 import ReactDOM from 'react-dom';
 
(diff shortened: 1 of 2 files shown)


line47:
While this does solve the problem I feel like it makes working with `formlabel` more confusing because we are using it to render a legend and some additional link text. 

I think it would make sense to add some props for things like `legend` and `legend hint` in this context as it is more direct as to how they will be applied. 

See my related comment here https://github.com/CMSgov/design-system/issues/529#issuecomment-547487771

bernardwang:
Closing because we will be going with this solution instead: https://jira.cms.gov/secure/RapidBoard.jspa?rapidView=2762&projectKey=WNMGDS&view=planning&selectedIssue=WNMGDS-291

PR #534, opened 2019-10-29
[WNMGDS-230] Add Tooltip component

hannaliebl:
### Added
Added a `<Tooltip />` component. This is essentially the tooltip that is used in PC2 and Windowshop. I've made a few changes to make it work with the design system naming conventions  and added an `aria-labelledby` attribute to the tooltip content based on a recommendation from https://jira.cms.gov/browse/WNMGPCII-430

The design/UX is based on https://jira.cms.gov/browse/WNMGWINSHOP-30. It's worth reading through the Tooltips Blueprint.docx file on that ticket to see what the expected behavior is/why I did some things the way I did. Let me know if you need access to it in a different way!

### TODO:

There are a lot of TODOs/questions left with this. Some of my motivation for moving this from our internal shared repo for PC2/Windowshop is to get more eyes on this to improve it, and to see if there need to be changes to support other teams who might use this.

#### Documentation
I didn't fill any of the documentation out that would be included with a new component. I'm not sure I'm the best person to write this, but it's probably something that should come at the end of the process once this is finalized.

#### Tests
I didn't include any tests! Again, also something that should be added once we have a finalized component.

#### Accessibility
There's a ticket in the PC2 backlog with some accessibility recommendations, specifically around screenreaders: https://jira.cms.gov/browse/WNMGPCII-430. The main actionable one here is ensuring that the tooltip content is in the DOM on page load instead of adding/removing it from the DOM when we hide/show it. There's also a question about whether or not we should even do the whole focus trap thing: instead, it might be better to just recommend that no actionable content should ever be inside a tooltip.

The tooltip as a child of the button recommendation is perhaps tough to do, and I hope my addition of the `aria-labelledby` fixes the issue well enough, but in any case, I chose to use a portal instead of making the tooltip the child of as I ran into positioning issues in IE 11 in some contexts - like tooltips in modals, for example. Also, a lot of the tooltip libraries I looked at for reference (bootstrap, material-ui) all append to the bottom of the body to avoid potential gnarly style issues, I think.

#### Mobile Behavior
On mobile, you need to click the tooltip icon itself to open/close it, which is maybe unexpected/odd. There's definitely more testing and thought to be done around mobile behavior, I think.

#### Dependencies
The way this is done now uses a library for positioning (`react-popper`), `react-transition-group` for the transition styles, as well as `react-focus-trap` for keyboard navigation when the tooltip content includes links/actionable content.

React/React DOM also needs to be version 16.3 or above due to `react-transition-group` requiring the context API, so I *think* I've properly updated the dependencies.

This opens up discussion about: how many other libraries we want to add to support a single component, whether or not there's a way to make this work for other React versions.


davidakennedy:
Nice work, @hannaliebl! This works well in my accessibility testing so far. There are some improvements I'd like to see though:

**Have the tooltip containers in the DOM**. It looks like you've marked everything up nicely, but the tooltip content isn't being read in Safari with VoiceOver. I think this is because the container isn't in the DOM until it's hovered/button pressed. That probably means we should just `display: none;`
it until it comes into view.

**Provide a way to close the tooltip via the escape key**. This is especially true for tooltip containers that have interactive content in them.

 I can also start writing the accessibility docs for this component.

DanielMiller-agile6:
Regarding your question about whether or not to support interactive content and include the focus trap, I have a few thoughts... we recently ran into this at QPP as well :).

We had a need for "glossary info tips" that help users see inline definitions for frequently used policy terms.  Our version also implemented a focus trap for interactive content like this, but we had similar reservations.  I'm of the opinion that a "Tooltip" pattern should never include interactive content, as described by WAI-ARIA authoring practices:

https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip

For "tooltips" that require interactive content, the recommendations seem to point to the "non-modal dialog" pattern, but only the "modal" version of this is defined:

https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal

Deque calls this pattern a "Tooltip Dialog," which seems like a better name for the interactive tooltip pattern and they offer some implementation examples:

https://dequeuniversity.com/library/aria/sf-basic-patterns/sf-tooltip-dialog

Maybe I'm just getting hung up on naming & terminology, but would it make more sense to offer 2 separate components for each of these patterns?  This would help to avoid overloading too much functionality into a generic "Tooltip" implementation, but may also cause some duplication in behavior if these were built separately.

packages/core/src/components/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,214 @@
...
+                    // Can't directly modify style, so copy and add styles from props
+                    const newStyle = {
+                      ...style,
+                      ...defaultTransitionStyle,
+                      ...transitionStyles[transitionState]
+                    };
+                    newStyle.maxWidth = this.props.tooltipMaxWidth || '300px';
+                    newStyle.zIndex = this.props.tooltipZIndex || '1';
+                    return (
+                      <div
```
DanielMiller-agile6: We may need to add a `role=tooltip` attribute here for the `hasInteractiveContent=false` case.  For Tooltip dialogs with interactive content, I think the recommendation is to use `role=dialog`.

packages/core/src/components/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,214 @@
...
+    this.setState({ showTooltip: true });
+  }
+
+  render() {
+    const body = document.querySelector('body');
+    return (
+      <Manager>
+        <Reference>
+          {({ ref }) => (
+            <button
```
DanielMiller-agile6: Have you thought about allowing this component to support alternate use cases for the trigger element (like `input`, `a`, or `span` elements)?

davidakennedy:
These are really good points, both on accessibility and pattern type.

To the original focus trap thoughts from @hannaliebl 

> There's also a question about whether or not we should even do the whole focus trap thing: instead, it might be better to just recommend that no actionable content should ever be inside a tooltip.

And also @DanielMiller-agile6 

> I'm of the opinion that a "Tooltip" pattern should never include interactive content, as described by WAI-ARIA authoring practices

I agree with this. We should steer clear of interactive content in the tooltip, and I do think that's a good reason for exploring another, similar pattern. The Tooltip Dialog pattern from Deque seems like a viable solution for interactive content.

bernardwang:
Thank you @hannaliebl for taking this on, I'll be helping reviewing and updating this PR today! 

I also agree with restricting the Tooltip to non interactive content. However, I also wanted to clarify, is an `<a>` element considered interactive?

DanielMiller-agile6:
I think an `<a>` is considered interactive.  Not totally only can it be clicked/tapped on by a user to navigate, but it also requires access via keyboard (which necessitates the focus trap and keyboard escape).

That said, the pattern is still in the  draft stages, so I don't believe there are a solid set of standards surrounding this:

https://github.com/w3c/aria-practices/issues/85

packages/core/src/components/Tooltip/Tooltip.jsx
bernardwang: The learn team uses [this tooltip](
https://gyazo.com/b67509066b6baab737067bd399122716) so I think its a good idea! However, I'm not sure whats the best way to implement these alternate use cases. 

I feel like allowing entirely custom tooltip triggers might not be ideal, i.e. I'm not sure if we want `input` or `a` elements that act like tooltips. Maybe instead of allowing custom markup we can just provide an optional `triggerLabel` prop that will replace the icon?

bernardwang:
@davidakennedy @hannaliebl 
I was testing this in VO, and I am also not able to get it to read. I tried changing it so the tooltip container is always in the DOM and to use `display:none` when it's hidden, but that didnt seem to fix it. However, when I moved the container next to the button instead of next to the `body`. I understand there is styling issues with this though, so I'm not entirely sure what's a good solution

Also, should we add the ability to exit out of the tooltip with the `esc` key for both interactive and non interactive tooltips? My intuition is that we should only exit on `esc` for interactive tooltips as its similar to a modal. Another consideration is what happens if the mouse is hovering on the button and someone presses `esc`?

packages/core/src/components/Tooltip/Tooltip.jsx
DanielMiller-agile6: The screenshot you shared seems pretty similar to the design QPP came up with for their glossary tooltips as well (dotted-underlined text with tooltip dialog on hover).  Between the "icon tip" and the "label tip", those seem like the only 2 scenarios we've had so far for tooltips and tooltip dialogs so baking those two cases into the component props makes sense.

bernardwang:
Some other misc feedback:

1. Instead of providing a `tooltipBodyInverse` prop, what do you think about providing a `tooltipBodyClasses` prop instead? I think it would be nice to only have 1 `inverse` prop to keep things simple
2. I think if the user defines`triggerContent`, we should not be using the default trigger icon.
3. Should we add the ability to exit out of the tooltip with the esc key for both interactive and non interactive tooltips? My intuition is that we should only exit on esc for interactive tooltips as its similar to a modal. Another consideration is what happens if the mouse is hovering on the button and someone presses esc?

DanielMiller-agile6:
Regarding the ESC key for tooltips, I think I agree... they aren't as necessary for "tooltips" as they are for "tooltip dialogs".  wai-aria practices don't mention ESC for the [tooltip](https://www.w3.org/TR/wai-aria-practices-1.1/#tooltip) pattern, but it is called out for the draft of [tooltip dialogs](https://rawgit.com/w3c/aria-practices/master/aria-practices-DeletedSectionsArchive.html#dialog_tooltip) pattern.

There's also [some discussion of this topic](https://github.com/w3c/aria-practices/issues/128) for the basic tooltip pattern... most seem to agree with your intuition @bernardwang 😄.

davidakennedy:
> I was testing this in VO, and I am also not able to get it to read. I tried changing it so the tooltip container is always in the DOM and to use display:none when it's hidden, but that didnt seem to fix it.

Yeah, there isn't a way for the tooltip itself to connect with the tooltip content. We solved this in PC2 and WS by putting the content in an `aria-label`. But that breaks down if/when you have interactive element in the content.

We could use a pattern like this:

```html
<button id="tooltip1" aria-describedby="tip1" class="tooltip">
<div class="tooltip-content" id="tip1">Some content.</div>
```

The thing I need to do some research on is how `aria-label` would interact with `aria-describedby` when it comes to screen readers.

The other thing not working here is the keyboard navigation. I can't get into the tooltip content if it has an interactive element.

I'll experiment some more here.

DanielMiller-agile6:
I hadn't looked at the `reach-ui` project in a while, but looks like they might be dealing with some similar issues as well regarding tooltips.  Could be useful as a reference :)

https://reacttraining.com/reach-ui/tooltip/

bernardwang:
@hannaliebl @davidakennedy @line47  What do yall think about proceeding with this PR by merging with interactive tooltips disabled. This way we can get this Tooltip out there as a `beta` component, and we can file a ticket for supporting interactive tooltips

bernardwang:
> Yeah, there isn't a way for the tooltip itself to connect with the tooltip content. We solved this in PC2 and WS by putting the content in an `aria-label`. But that breaks down if/when you have interactive element in the content.

@davidakennedy In my last commit I tried adding `aria-describedby` in addition to using `aria-label`, and I also added `role`. Unfortunately I'm still having trouble getting inside the tooltip with VO. I might be navigating something wrong but its still tricky for me

davidakennedy:
@bernardwang –

> In my last commit I tried adding aria-describedby in addition to using aria-label, and I also added role.

These additions are working well.

> Unfortunately I'm still having trouble getting inside the tooltip with VO. I might be navigating something wrong but its still tricky for me

I suspect you're having trouble with this because of the location where the tooltip content gets inserted. It's at the end of the dom, so if you're trying to go in order with a screen reader, you won't get to it for awhile. That may need some rethinking. Ideally, the container with the content comes after the tooltip button, but we may be doing that for easier positioning.

bernardwang:
I saw this [tooltip today](https://ussba.github.io/certify-design-system-documentation/components/tooltips.html), and it made me wonder if we should consider simplifying the design of this component. The default "i" icon uses a lot of drop shadows and the circle that appears on hover adds a bit of complexity to the component. I wonder if this is suited for the CMSDS, especially when many use cases would use the tooltip with a different icon or element altogether. 
Screen Shot 2020-01-31 at 12 59 07 PM
Any thoughts from designers? @melwoodard

bernardwang:
@hannaliebl @line47 Last week I found out that this component came from [this repo](https://github.cms.gov/CMS-WDS/pet-shared-components), and I'm thinking that it would make more sense to have this Tooltip initially be included in the [hc.gov design system](https://github.cms.gov/CMS-WDS/ds-healthcare-gov). This is because m.gov has their own Tooltip implementation that looks and functions differently:
Screen Shot 2020-02-03 at 10 54 29 AM

If we move this to the hc.gov design system we can keep iterating on it without having to consider M.gov's implementation (for now). Ideally, it would be really great to move more relevant components from that `pet-shared-components` repo to the hc design system in the future!

melwoodard:
I do think it would be valuable for CMSDS to have _some_ kind of tooltip implementation, mainly for the benefit of providing guidance on accessibility as well as when to use it and when to consider something else. Since tooltips are so often implemented poorly or inappropriately, it would be good for CMSDS to point people in the right direction.

cc @savannahmillion

davidakennedy:
> What do yall think about proceeding with this PR by merging with interactive tooltips disabled. This way we can get this Tooltip out there as a beta component, and we can file a ticket for supporting interactive tooltips

I tested this some more today... I like the idea of moving forward with the simpler tooltip first, and going from there. But I'd want to see some changes there first. See these comments from @1Copenut in the original ticket @hannaliebl linked to:

> - The tooltip text needs to exist in the DOM on first render, instead of be added dynamically. We can hide it with CSS like display: none and reveal it on :hover and :focus.
> - Tooltip text should either be a child of the <button> or have a unique ID and the button have an aria-labelledby that references said unique ID.

This next piece related to having interactive content:

> - Tooltips with close buttons and linkable text are problematic. Can we remove this branch of logic, or do we have approved designs that call for this type of display? It will require additional testing for full accessibility and may need to be moved into a help drawer component.

Those first two are important to having a solid experience all around, even in beta.

packages/core/src/components/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,260 @@
...
+              };
+              return (
+                <div
+                  id={`tooltip-${id}`}
+                  ref={ref}
+                  className={classNames('ds-c-tooltip__container', tooltipBodyClassName, {
+                    'inverse-tooltip-body': inverse
+                  })}
+                  style={newStyle}
+                  onMouseEnter={() => this.showTooltip()}
```
austin-martinez: This isn't a hard requirement but in the past PC2 logged analytic events to track how often tooltips were used by listening to mouseenter and I think that information could still be valuable as we evaluate when to use tooltips vs. other UX patterns

Ian-Lowrie:
Hello from QualityNet land! We are working with our prime to implement accessible tooltips with hover on the new Healthcare Quality Reporting site. This looks like it would absolutely fit the bill. I wanted to share our user story for this tool, and see if there was any movement on this pull request. 

**User Story: Tool Tips with Hover**

**Description**
The usability of in-context help on Hospital Quality Reporting would be improved by adding a hover state.  Currently, the ADO has added Tool Tips without Hover as the Hover text is not readable by screen readers.

The goal is to expand the functionality of the Tool Tip to include a Hover.  The User would then be able to accidentally activate the Tool Tip, improving the discoverability of in-context help. Further, hover is more usable than click for quick nomenclature lookups, a major use case for in-context help on HQR.

**User**
The User is a hospital administrator who tends to be older and less tech savvy.  They are mandated to submit certain medical procedure measures in order to receive compensation for their organization from CMS. They are very dependent upon the Specific Rules and often need guidance as to how to execute certain measures.

By allowing them the ability to access the information without clicking, they will be able to quickly intuit how the help system works.  

**Precondition**
The CMS Tool Tip is already implemented.  This would be an update to the current CMS Design System.

**Story**
As a hospital administrator, I want to be able to find help easily so I can submit my measures in a timely manner.

**Acceptance criteria includes:**
A User can see the Help prompt by hovering over the Tool Tip icon
A User can see the Help prompt by clicking on the Tool Tip icon
A User can keep the Help prompt open for a short period of time by clicking on the Tool Tip icon
A User can click on links in the Tool Tip

Ian-Lowrie:
Hello from QualityNet land! We are working with our prime to implement accessible tooltips with hover on the new Healthcare Quality Reporting site. This looks like it would absolutely fit the bill. I wanted to share our user story for this tool, and see if there was any movement on this pull request. 

**User Story: Tool Tips with Hover**

**Description**
The usability of in-context help on Hospital Quality Reporting would be improved by adding a hover state.  Currently, the ADO has added Tool Tips without Hover as the Hover text is not readable by screen readers.

The goal is to expand the functionality of the Tool Tip to include a Hover.  The User would then be able to accidentally activate the Tool Tip, improving the discoverability of in-context help. Further, hover is more usable than click for quick nomenclature lookups, a major use case for in-context help on HQR.

**User**
The User is a hospital administrator who tends to be older and less tech savvy.  They are mandated to submit certain medical procedure measures in order to receive compensation for their organization from CMS. They are very dependent upon the Specific Rules and often need guidance as to how to execute certain measures.

By allowing them the ability to access the information without clicking, they will be able to quickly intuit how the help system works.  

**Precondition**
The CMS Tool Tip is already implemented.  This would be an update to the current CMS Design System.

**Story**
As a hospital administrator, I want to be able to find help easily so I can submit my measures in a timely manner.

**Acceptance criteria includes:**
A User can see the Help prompt by hovering over the Tool Tip icon
A User can see the Help prompt by clicking on the Tool Tip icon
A User can keep the Help prompt open for a short period of time by clicking on the Tool Tip icon
A User can click on links in the Tool Tip

nichia:
Thank you @Ian-Lowrie for the user story - we're picking this up. Stay posted

davidakennedy:
I retested this some today after more perspective, thanks to advising another team on how to create accessible tooltips. 😄 

I'll emphasize some of the feedback from Trevor in the Jira ticket linked up by Hanna. I've reordered this feedback from most important to least.

**The tooltip text needs to exist in the DOM on first render, instead of being added dynamically. We can hide it with CSS like display: none and reveal it on :hover and :focus.** This is probably the most important thing to change from the current implementation. Without it, associating the tooltip content with the tooltip itself can be buggy (see my testing notes) in some browser and screen reader combinations.

**Tooltips with close buttons and linkable text are problematic. Can we remove this branch of logic, or do we have approved designs that call for this type of display?** This is true, and can be more challenging for assistive technologies. We know that we have a current need for this based on the tooltips in production on healthcare.gov (and likely others sites as well). A few ideas related to this:
- When tooltips contain no interactive content (buttons, links, inputs), they're less problematic because you don't need to manage focus. That means moving focus when the tooltip is open, and maybe trapping it while open. Hana points to this above. We can't move focus on hover because it's unexpected from a user point of view and also a WCAG violation (The current implementation does this, and we should remove it). So we could have two types of tooltips, and you use a different one depending on whether the content is interactive. But this seems problematic because they would work differently. That's not great UX, of course.
- Or we could make all tooltips work on touch/click only. This would enable all the tooltips to be able to work the same, regardless of the type of content. That would make things technically easier from an accessibility (and mobile) standpoint, but I don't know how strong the need would be to keep things working with hover as well. This would be a good thing to user test.

**Tooltip text should either be a child of the <button> or have a unique ID and the button have an aria-labelledby that references said unique ID.** Having the tooltip text be a child of the tooltip would be ideal, but this can be mitigated using aria-labelledby or aria-describedby. So I don't see an issue here when looking at the current implementation, especially as we tackle the above two issues.

Testing notes:

As Travor pointed out in his Jira ticket, screen reader virtual cursors are screen reader specific and do not always follow browser :focus behavior. I found that in my testing.

JAWS + Edge: Does not open when the virtual cursor hits control. Needs to be opened with the spacebar or enter. Does announce the tooltip content when tabbed to with focus.

JAWS + IE11: Does not open when the virtual cursor hits control. Needs to be opened with the spacebar or enter. Also, the tooltip content is never announced. Doesn't announce the tooltip content when tabbed to with focus.

JAWS + Chrome: Does not open when the virtual cursor hits control. Needs to be opened with the spacebar or enter. Does announce the tooltip content when tabbed to with focus.

NVDA + Firefox: Opens when the virtual cursor hits control, but does not announce tooltip content. Does announce the tooltip content when tabbed to with focus.

NVDA + Chrome: Opens when the virtual cursor hits control, but does not announce tooltip content. Does announce the tooltip content when tabbed to with focus.

VoiceOver + Safari: Works well overall, but for the version that has interactive content in it, focus is shifted, but VoiceOver's virtual cursor seems to get stuck on the tooltip button, and doesn't shift to the first interactive element inside the tooltip content.

packages/design-system/src/components/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,260 @@
...
+                  id={`tooltip-${id}`}
+                  ref={ref}
+                  className={classNames('ds-c-tooltip__container', tooltipBodyClassName, {
+                    'inverse-tooltip-body': inverse,
+                  })}
+                  style={newStyle}
+                  onMouseEnter={() => this.showTooltip()}
+                  onMouseLeave={() => this.hideTooltip()}
+                  data-placement={placement}
+                  modifiers={{ offset: TOOLTIP_OFFSET }}
```
bernardwang, line 183: @hannaliebl do you know why this `TOOLTIP_OFFSET` is needed? I tried changing it's value and removing it from this `modifiers` and nothing seemed to change

bernardwang:
> **The tooltip text needs to exist in the DOM on first render, instead of being added dynamically. We can hide it with CSS like display: none and reveal it on :hover and :focus.** This is probably the most important thing to change from the current implementation. Without it, associating the tooltip content with the tooltip itself can be buggy (see my testing notes) in some browser and screen reader combinations.

@davidakennedy I have been unable to get this to work with `display: none` unfortunately, as it doesn't play well with transitions. I tried hiding it visually (`opacity: 0`) and from the screenreader (`aria-hidden`), but this breaks for interactive tooltips, which has elements that are still accessible via tab (fyi this is the approach USDWS takes but they dont really support interactive tooltips). 

I'm considering testing a 3rd option where we use `aria-hidden` again, but we have the tooltip body located at the end of the `body` element using a React Portal. However, I think there's still the issue of tabable elements in interactive tooltips, and im not sure if we can get around that even if the elements are at the end of the `body` element

bernardwang:
@davidakennedy @nichia @line47 
This is ready for a first initial review, especially on the a11y side of things! I would also love to get some feedback on how the component usage is designed. Originally I was going to allow users to pass in their own trigger elements to allow for full customization. However, because there are a good number of attributes we need to apply (i.e. `aria-label`, `aria-describedby`, and event listeners) as well as the need to generally speaking restrict trigger elements to `<button>` and `<a>`, I opted for a different approach. 

Instead, users can choose the element type similar to the `component` prop in `FormLabel`, via the `triggerComponent` prop. We also provide `triggerContent`, `triggerClassName`, `triggerHref` to pass along to the trigger element. The benefit to this approach is that the user can still pass along a custom element like an icon, but we can insure the icon will be inside either a `<button>` and `<a>`.

On the otherhand, it's also a bit of a pain to have a bunch of props that are basically just passed along. A potentially simpler option could be to allow the user to pass in the trigger element directly, and then we just clone it and append all the necessary attributes and event listeners. This gives us a bit less control, i.e. its possible the user will pass in a `<span>` or something, but in this case we could also add console warnings.

What do yall think?

davidakennedy:
Hey @bernardwang, thanks for all the work on this. It's great to see this progressing...

> Instead, users can choose the element type similar to the component prop in FormLabel, via the triggerComponent prop. We also provide triggerContent, triggerClassName, triggerHref to pass along to the trigger element. The benefit to this approach is that the user can still pass along a custom element like an icon, but we can insure the icon will be inside either a <button> and <a>.

This seems like a good balance. Have users understood the similar approach with the form label?

> On the otherhand, it's also a bit of a pain to have a bunch of props that are basically just passed along. A potentially simpler option could be to allow the user to pass in the trigger element directly, and then we just clone it and append all the necessary attributes and event listeners. This gives us a bit less control, i.e. its possible the user will pass in a <span> or something, but in this case we could also add console warnings.

I'm okay with this approach too. We could check if the trigger is a link, button or input and if not, return and pass an error.

-----

I also tested this some more, and wanted to share my notes. We talked about most of these already, Bernard:

# Tooltip PR Review Notes

Notes are separated by types of testing.

## Overall Notes

- I wonder if we should have a set of recommendations for properties, depending on the type of tooltip you want? Because if the mix and match nature of the component properties, that might be the best way to help people create accessible tooltips.
- Can we make it so text is copyable via a mouse? See this WCAG 2.1 requirement: https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html
- Why is the `tabindex="-1"` on the content container?

## Keyboard

- When interactive content is included, a keyboard trap happens. You can tab into the tooltip content but can't get out in any way. To resolve this, we could:
    - Put a close button, coupled with an ESC key to close content.
    - Not trap focus (only really works if the tooltip content is after tooltip trigger in source). I think this is my preference given the current setup and trying to catch all the needs across the DS ecosystem.

## Screen Reader

- Let's not use `aria-label` by default on examples. It's too heavy handed. You only need this if an button is the trigger with an icon.
- I wasn't able to test in all screen readers because of the keyboard trap.

bernardwang:
@line47 @nichia @davidakennedy This PR has been updated, and is ready to review with the exception of one bug (closing the on click tooltip by clicking the trigger).

Changes since the last review:
- Only tooltips using the `onClick` prop will have a focus trap, and they will be only activated and deactivated by click. 
- The bug where the hover tooltips would flash on and off @nichia encountered has been fixed
- The basic interactive tooltip example has no focus trap, and can be tabbed through normally.

packages/design-system-docs/src/pages/components/Tooltip/_Tooltip.docs.scss
```diff
@@ -0,0 +1,19 @@
...
+
+/*
+`<Tooltip>`
+
+@react-example Tooltip.example.jsx
+
+@react-props Tooltip.jsx
+
+Style guide: components.tooltip.tooltip
+*/
```
line47, line 19: This is a start to guidance for using Tooltips 

```suggestion
*/
/*
---

### When to use
- Use for helpful, non-critical information to strengthen an existing message.
- Use to increase certainty and enhance confidence about an interaction.
- Use with brief descriptions as tooltips perform best with succinct text.
- Tooltips are useful as a last resort for space-constrained UI. Explore other options for keeping content visible without a tooltip.

## When to consider alternatives
- Don’t hide information necessary for completing a task behind a tooltip interaction.
- Tooltips are microcopy and should be brief. Don't use a tooltip if you need a lot of text.
- Don’t use a tooltip when its content is repetitive or if usability is obvious.
- If content can fit outside a tooltip, don't use a tooltip.

Style guide: components.tooltip.guidance
*/
```

davidakennedy (changes requested):
Overall, changes look good @bernardwang. A few minor notes that hopefully are quick fixes:

- The close button on the click example doesn't seem to work. The ESC does.
- The example file isn't rendering in IE 11 for me.

Also, I see this in Safari + VoiceOver. Note how there's the extra tab stop happening where you can see I'm able with VoiceOver to focus on the div container. Why is that? I don't think this is an issue for merging, just wondering if it's something we could improve.

Screen Cast 2020-09-16 at 3 27 47 PM

bernardwang:
> * The close button on the click example doesn't seem to work. The ESC does.

I updated the example to not have a close button, as its misleading. It was supposed to just be example content.

> * The example file isn't rendering in IE 11 for me.

My guess is that this is from the libraries we are using for the tooltip. I will investigate further but this might be unfortunately have to be handled in a follow up. Would that work for you?

Same with the extra tab stop, I think I can iterate on that and fix it in a follow up.

packages/design-system/src/components/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,293 @@
...
+      <TriggerComponent
+        id={triggerId}
+        type={TriggerComponent === 'button' ? 'button' : undefined}
+        onTouchStart={() => this.setTooltipActive(!this.state.active)}
+        onFocus={() => (dialog ? null : this.setTooltipActive(true))}
+        onBlur={() => (dialog ? null : this.handleBlur())}
+        onMouseEnter={() => (dialog ? null : this.setTooltipActive(true))}
+        onMouseLeave={() => (dialog ? null : this.setTooltipActive(false))}
+        onClick={() => this.setTooltipActive(!this.state.active)}
+        aria-label={ariaLabel || ''}
```
nichia, line 120:
```suggestion
        aria-label={ariaLabel || undefined}
```

packages/design-system/src/components/Tooltip/Tooltip.jsx
```diff
@@ -0,0 +1,293 @@
...
+    const triggerClasses = classNames('ds-c-tooltip__trigger', 'ds-base', triggerClassName, {
+      [triggerActiveClassName]: this.state.active,
+      'ds-c-tooltip__trigger--click-only-active': this.props.dialog && this.state.active,
+    });
+
+    return (
+      <TriggerComponent
+        id={triggerId}
+        type={TriggerComponent === 'button' ? 'button' : undefined}
+        onTouchStart={() => this.setTooltipActive(!this.state.active)}
```
nichia, line 114: Nice use of `onTouchStart`, first time to see this event. [MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ontouchstart#Browser_compatibility) list this as an experimental technology, I'm curious to continue trying this feature out on various browsers.

nichia (approved):
What a well structured and well written piece of code! Totally enjoyed reviewing this component. 
The tooltips LGTM and I've just added a minor suggestion.

davidakennedy:
@bernardwang 

> My guess is that this is from the libraries we are using for the tooltip. I will investigate further but this might be unfortunately have to be handled in a follow up. Would that work for you?

That works. Note that I haven't tested the latest in IE 11 with JAWS, but I wouldn't expect issues from an accessibility perspective.

> Actually, I think the tab stop might be coming from `role="tooltip"` on the container of the tooltip content? Is that expected?

Hmm... I removed that, and it still happens. I didn't think about this, but VoiceOver announces it as a group, so something is making that happen. Going to do a bit of research. I don't think this is an showstopper, but I want to see what's causing it so I know why and if it's expected. So we could address in a followup, if needed.

davidakennedy (approved):

davidakennedy:
@bernardwang I did some more testing and research around that tab stop in the tooltip content...

It's coming from the `aria-labelledby` attribute. That group is where the aria-labelledby attribute is announced, which in this case is the tooltip trigger’s accessible name. So the extra tab stop makes sense. I don't think we need it though since it's announcing the trigger name again, shortly after it's announced the first time. So it's overkill. We can remove it.

bernardwang:
@davidakennedy I pushed some updates to the branch, it should be working in IE11 now. 

I also went ahead and removed the `aria-labelledby` attribute, but some reason in my testing I couldnt reproduce that extra tab stop again, which is a bit weird. 

I also noticed 2 additional screenreader behaviors we could improve on:
1. Dialog tooltips announce the tooltip content even if the user never clicked on the button, it is announcing the contents as if the tooltip is activating on focus/hover.
2. Should we indicate when the dialog tooltips are open or not? Or give some sort of prompt on a dialog tooltip trigger that you should click the trigger to open the tooltip?
2. Currently basic hover/focus tooltips still listen to click events, and will turn on and off if the user clicks on them. Is this making things too complicated? I could see this impacting the screenreader experience for hover/focus tooltips if clicking the tooltip trigger (which is a button) unexpectedly toggles the tooltip.

bernardwang:
@davidakennedy I've been calling the tooltip that activates on click only (and includes focus trap) as a "[tooltip dialog](https://github.com/w3c/aria-practices/issues/85)". Should I consider renaming to "[toggletips](https://inclusive-components.design/tooltips-toggletips/#inclusivetoggletips)"?

bernardwang:
I merged this branch to get the Draft component out into master, and I will continue to address the feedback here in follow up PRs

davidakennedy:
@bernardwang 

> Dialog tooltips announce the tooltip content even if the user never clicked on the button, it is announcing the contents as if the tooltip is activating on focus/hover.
Good catch. Will do some research here to see if there's a better way to handle that so it doesn't occur. I agree it's confusing.

> Should we indicate when the dialog tooltips are open or not? Or give some sort of prompt on a dialog tooltip trigger that you should click the trigger to open the tooltip?
Good question! That's handled for screen reader and keyboard users by the focus management we do. I don't think we need to be verbose with the wording anywhere.

> Currently basic hover/focus tooltips still listen to click events, and will turn on and off if the user clicks on them. Is this making things too complicated? I could see this impacting the screenreader experience for hover/focus tooltips if clicking the tooltip trigger (which is a button) unexpectedly toggles the tooltip.

We need that to make them work well on mobile, unless I'm misunderstanding. What other way would are you thinking?

> I've been calling the tooltip that activates on click only (and includes focus trap) as a "tooltip dialog". Should I consider renaming to "toggletips"?

I like tooltip dialog because it's clearer, and less jargon-y to me.

Issue #499, opened 2019-10-02
Third-Party Link Disclosure 


line47:
Please add your examples of where you have notified users they are leaving a CMS site and going to a third-party site. 

Screenshots and links would be helpful!

melwoodard:
This is not an example, but noting CMS's stated policy on external links: https://www.cms.gov/About-CMS/Agency-Information/Aboutwebsite/PolicyforLinkingtoOutsideWebsites.html

melwoodard:
healthcare.gov link disclosure: https://www.healthcare.gov/privacy/#links

medicare.gov link disclosure: https://www.medicare.gov/privacy-policy#other-sites

SparrowSpit:
On MCT, we are hoping to implement something similar to the above healthcare . gov link that Mel posted.

bernardwang:
The example Melissa gave us was [this alert on blue button](https://www.medicare.gov/index.php/manage-your-health/medicares-blue-button-blue-button-20/blue-button-apps?label=connecttoresearch)

I've heard several ideas thrown around for this, including an interstitial page, external link icons, and a dialog. From what I can tell the dialog was the most commonly agreed upon solution, and it is very similar to the JS alert we have in that example. 

How do people feel about adding a Pattern to the docs that uses the [CMSDS Dialog](https://gyazo.com/d59fedb19607358645b7dc0522fbc989)?

melwoodard:
Are there any strong benefits to using the CMSDS dialog over the JS alert? Thinking particularly about mobile, being able to leverage a native browser modal will likely be a better experience.

line47:
I think the strongest benefit is UI cohesiveness and I feel like the default DOM alert doesn't look great. However, I'm unsure how the mobile experience would be when for example you are on a long page and invoke the CMSDS Modal. I think there is some testing to do here still.

melwoodard:
My concern is that both hc.gov and m.gov seem to be receiving guidance from CMS to use the default DOM alert, so for CMSDS to give alternate guidance is a bit confusing.

melwoodard:
My concern is that both hc.gov and m.gov seem to be receiving guidance from CMS to use the default DOM alert, so for CMSDS to give alternate guidance is a bit confusing. But either way I think if we add documentation around this, we should specify a recommended max character count to ensure it works well on mobile.

PR #535, opened 2019-10-30
[WNMGDS-264] - Add pattern for third party external links 

line47:
### Added
Added a react example and a docs page for third party external links. 

#### Screenshots 
Screen Shot 2019-10-30 at 3 11 59 PM
Screen Shot 2019-10-30 at 3 12 04 PM
Screen Shot 2019-10-30 at 3 12 10 PM


packages/core/src/components/Dialog/_ThirdPartyLink.docs.scss
```diff
@@ -0,0 +1,32 @@
...
+*/
+
+/*
+---
+
+Use this pattern when the user is in the middle of a task or application to let them know that the action they are taking will move them away from the current site and or process. This serves as a warning where the user can then decide to remain where they are in the application or process.
+
+### Learn more
+
+- [Making an accessible dialog box](https://www.nczonline.net/blog/2013/02/12/making-an-accessible-dialog-box/)
```
bernardwang: are these the same resources on the dialog page? maybe we could link those cms privacy guidance pages?

line47: Yep, they are the same. I'm not sure which links you are referring to?

bernardwang: Maybe we could replace the resources on dialogs with the links Mel posted on this page:
https://github.com/CMSgov/design-system/issues/499#issuecomment-537659356

line47: Ready for another pass

mliupaeter:
This looks great! Thanks guys!

PR #536, opened 2019-10-30
[WNMGDS-260] Consolidate code examples

bernardwang:
### Summary
This PR consolidates and standardizes component code examples. Major changes on `ChoiceList`, `Alert`, `TextField`, `Button`, `Badge`. In general, code examples were that were once separate sections (inverse examples, modifier examples) were combined into a single block. 

### Added
- New example page styles, `.example--wrapper` and `.example--inverse`
- 3 new button classes used in the code examples (See https://github.com/CMSgov/design-system/pull/536#discussion_r340839931)

### Changed
- `ChoiceList` examples were combined but copy was also changed to be "self describing"
- `TextField` examples were modified based off the existing React examples, also changed to be "self describing"


packages/core/src/components/Button/Button.scss
```diff
@@ -17,12 +17,15 @@
   text-decoration: none;
 
   &:focus,
-  &:hover {
+  &:hover,
+  &.ds-c-button--focus,
```
bernardwang: Added 3 new button classes used in the code examples. `ds-c-button--focus`, `ds-c-button--hover`, and `ds-c-button--active`. 
`ds-c-button--focus` isnt used right now but once we implement focus we can add it to the code examples

bernardwang: It also might also be worth refactoring `ds-c-button--disabled` to work similarly to those new classes. Could be included in this ticket: https://jira.cms.gov/browse/WNMGDS-290

packages/core/src/components/Alert/Alert.example.jsx
```diff
@@ -3,19 +3,16 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 ReactDOM.render(
-  <div>
+  <div className="example--wrapper">
```
line47: Is `example--wrapper` to be used on all examples or only specific ones?

packages/core/src/components/Badge/badge.example.html
```diff
@@ -1,3 +1,35 @@
-<span class="ds-c-badge {{modifier}}">5</span>
-<span class="ds-c-badge {{modifier}}">Badge label</span>
-<span class="ds-c-badge {{modifier}} ds-u-font-size--base">Badge label</span>
+<div class="example--wrapper">
```
line47: Wondering if we should add labels above the badges like we're doing for buttons?

line47:
This looks good, left a few comments but nothing that would stop a merge 🚀

packages/core/src/components/Alert/Alert.example.jsx
bernardwang: It only works for certain examples like Badge. This is because components like `TextField` are composed of 2 adjacent elements, and `example--wrapper` will put the margin on both elements unless we wrap the elements with a div, and I didnt want to clutter up our code examples.

Theres other cases too where theres already margin from a label or some other child element, so it doesnt make sense to use `example--wrapper`.

This definetely isnt perfect... I wish the margins on our components were consistently on the outer wrapper so that these examples could be spaced perfectly

packages/core/src/components/Badge/badge.example.html
bernardwang: The only reason why I didnt is because the Badges already have the name of the color/type of Badge (i.e. Primary, Warning etc). You cant do this with buttons because the labels refer to hover, focus, active etc

bernardwang:
@line47 I forgot to mention, the last component that I don't know how to handle is the Modal Dialog. It uses the old "modifier" format, but it's a really big example so I'm not sure if its possible to combine. Do you think it's fine the way it is, or if I should try and combine them?

line47:
Because of the way the modal examples work I think it would be easier to keep them as separate example blocks. There is also a lot of code for one example so I think combining them would make it harder to read.

PR #537, opened 2019-10-31
[WNMGDS-275] - Add links to the CMSDS

line47:
### Added
Added a page to the doc site for links 

### Changed
I needed to modify some other component styles to ensure the visited link styling would not conflict with the component.


packages/core/src/base/body.scss
```diff
@@ -40,6 +40,7 @@ Style guide: style.base
     a {
       color: $color-base-inverse;
 
+      &:visited,
```
line47: Kept the color the same for visited links on inverse bg

packages/core/src/base/typography.scss
```diff
@@ -53,26 +53,7 @@ Markup:
...
-    &:hover,
-    &:focus {
-      color: $color-primary-darker;
-    }
-
-    &:active {
-      color: $color-primary-darkest;
-    }
-  }
-}
```
line47: Moved this under links.scss

packages/core/src/components/Button/Button.scss
```diff
@@ -40,7 +40,7 @@
   padding: $spacer-half $spacer-1;
 }
 
-.ds-c-button--primary {
+.ds-c-button--primary, .ds-c-button--primary:visited {
```
line47: Needed to add some visited styles here to prevent the buttons for getting the visited link text color

packages/core/src/components/Links/_Links.docs.scss
```diff
@@ -0,0 +1,51 @@
...
+Never use a link to say “click here.” A nondescript link forces users to backtrack and read the surrounding text for more context. This is even more problematic for those who rely on screen readers, which can list links for quicker navigation. A list of “click here” isn’t helpful for anyone.
+
+### Accessibility
+
+- Users should be able to tab to navigate between links.
+- Users should be able to activate a link when pressing ‘Enter’ on their keyboard.
+- Users should be able to identify links without relying on color alone.
+- Users should be able to activate hover and focus states with both a mouse and a keyboard.
+
+Style guide: components.links.guidance
```
line47: Would enjoy feedback on the content here @davidakennedy or @melwoodard

packages/core/src/components/Links/_Links.docs.scss
```diff
@@ -0,0 +1,51 @@
+/*
```
bernardwang: I'm not sure if it makes sense to group Links under components, as its just an `<a>` tag. Maybe we should consider making the stylings a utility class and moving it under utilities?

bernardwang: Actually, I'd even be ok with just having a default styling and not having an additional documentation page. Since its just a native HTML element I dont think we need much guidance.

packages/core/src/components/Links/_Links.docs.scss
```diff
@@ -0,0 +1,51 @@
...
+Links
+
+Links connect users to a different page or further information.
+
+@uswds https://standards.usa.gov/components/labels/
```
bernardwang: As I understand, USWDS doesnt have a link component, this links to https://designsystem.digital.gov/components/tag/, which is entirely different. Much of the guidance on this page is also from Tags

packages/core/src/components/Links/_Links.docs.scss
line47: I was thinking it would be nice to have to show what the styling looks like. Maybe under components is not the best place. This brings up the definition of component thing for me and does component mean React component or general HTML component?

packages/core/src/components/Links/_Links.docs.scss
line47: USWDS places their link information under typography https://designsystem.digital.gov/components/typography/#links

packages/core/src/components/Links/_Links.docs.scss
bernardwang: I think it makes sense if we included it in the typography section! As far as components go, I don't think if its react or html, as long as it's showcasing something that you use in the DS like a set of classes or certain markup.

packages/core/src/base/index.scss
```diff
@@ -1,3 +1,3 @@
 @import 'body';
-@import 'typography';
+@import 'typography/typography';
```
line47: Moved typography to its own folder due to splitting up styles and examples into their own files

packages/core/src/base/typography/_typography.docs.scss
```diff
@@ -0,0 +1,106 @@
...
+*/
+
+/*
+Links
+
+Links connect users to a different page or further information.
+
+The only base HTML element the design system applies styling to is the `<a>` element. To prevent this, you can override the `$ds-include-base-html-rulesets` Sass variable and set it to `false`.
+
+Markup: links.example.html
```
line47: I wish the markup was inserted where it appears in the file instead of at the bottom, this makes the page not look so great

packages/core/src/base/typography/typography.example.html
```diff
@@ -0,0 +1,11 @@
+<h1 class="ds-display">Display</h1>
+<h1 class="ds-title">Title</h1>
+<h1 class="ds-h1">Heading 1</h1>
+<h1 class="ds-h2">Heading 2</h1>
+<h1 class="ds-h3">Heading 3</h1>
+<h1 class="ds-h4">Heading 4</h1>
+<h1 class="ds-h5">Heading 5</h1>
+<h1 class="ds-h6">Heading 6</h1>
```
line47: separated the example and remove some of the extra redundant paragraphs in the example

packages/core/src/components/_index.scss
```diff
@@ -8,6 +8,7 @@
 @import 'Dropdown/Dropdown';
 @import 'FormLabel/FormLabel';
 @import 'HelpDrawer/HelpDrawer';
+@import 'typography/Links';
```
line47: updated the links scss reference

packages/core/src/components/Links/_Links.docs.scss
```diff
@@ -0,0 +1,51 @@
...
+
+Style guide: components.links
+*/
+
+/*
+---
+
+### When to use
+
+- Navigating to a new page or view in your application
```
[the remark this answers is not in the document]
davidakennedy: One way to tighten the technical boundaries is: If the URL changes, use a link, otherwise use a button.

There may be exceptions to that, but I can't think of any right now.

packages/core/src/components/_index.scss
```diff
@@ -8,6 +8,7 @@
 @import 'Dropdown/Dropdown';
 @import 'FormLabel/FormLabel';
 @import 'HelpDrawer/HelpDrawer';
+@import '../base/typography/Links';
```
bernardwang: Could you move the import to `base/index.scss`?

packages/core/src/components/Links/_Links.docs.scss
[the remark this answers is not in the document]
line47: Updated

packages/core/src/components/Links/_Links.docs.scss
```diff
@@ -0,0 +1,51 @@
...
+- Navigating to a different web page, e.g. external documentation
+
+### When to consider alternatives
+
+- Opening or closing a modal or dialog
+- Triggering a dropdown menu
+- Submitting data to the server
+
+### Usage
```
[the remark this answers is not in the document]
line47: This content has been removed, thanks!

bernardwang:
Last thing, currently we don't have a visited link color for inverse, should we?

line47:
We do, however, the styles are under body.scss in base here
https://github.com/CMSgov/design-system/blob/sw-add-link/packages/core/src/base/body.scss#L40

bernardwang:
Hm, yeah its super weird, I see the style in the example but it doesnt seem to be working for some reason

bernardwang:
It's because opacity doesnt work for visited links, maybe we should consider changing the color to whatever `#ffffff` at 0.8 opacity is. It's not a perfect translation (cuz opacity depends on the background so "converting" 0.8 opacity to hex wont work perfectly), but I think its an improvement.

packages/core/src/components/Button/Button.scss
bernardwang: Should this be added to all the button variations? Not just the main class, primary, and inverse? Alternatively, can we get away with just adding it to the base class `ds-c-button`? That should be on every button right?

bernardwang: Oopf, I just realized something. If we want an `<a>` element to be styled as a button, we don't want it to change its styling after being clicked. That's why we add `:visited` to all these classes. This gets a bit weird when we also a transparent button variation which is styled like a link, because then you would expect it to have a `:visited` style, even though its a button. This relates to the larger "should you have links styled as buttons and vice versa" discussion.

packages/core/src/base/body.scss
```diff
@@ -44,7 +44,7 @@ Style guide: style.base
       &:active,
       &:focus,
       &:hover {
-        color: $color-base-inverse;
+        color: $color-muted-inverse;
```
bernardwang: I just did the calculation, the exact color of white at 80% opacity on the inverse dark bg is `#cfd5dc`. I wonder if we should use that instead, considering we use `$color-muted-inverse` as the disabled transparent button styling.

PR #538, opened 2019-10-31
[WNMGDS-285] Remove select page and right to left choice sections from doc site

### LINKED ISSUE #360: Propose remove select examples from Checkbox & Radio component page
## Description
On the Checkbox & Radio page https://design.cms.gov/components/choice/ we have select components but we also have a separate select page. 

I think we should remove the select examples from the Checkbox & Radios page and keep them on the select page.

line47:
### Removed
- Removed the select page from the doc site, however, the component will still function if used. 
- Removed the right to left choice examples from the documentation 

This pr closes #360


bernardwang:
I see there's a few references left to "select menus" in `forms.md`, `Textfield.docs.scss`, `ChoiceList.docs.scss`. Can those also be updated to use dropdown?

packages/core/src/components/ChoiceList/_ChoiceList.docs.scss
```diff
@@ -11,13 +11,19 @@ Style guide: components.choice
 */
 
 /*
```
bernardwang: These changes right here look like merge conflicts, they should be safe to delete

PR #540, opened 2019-11-01
Add concise modal example with javascript

bernardwang:
This issue (https://github.com/CMSgov/design-system/pull/536#issuecomment-548584972) was bugging me a bit, so I tried fixing it with js
[Gif of new example](https://gyazo.com/b2ad3abdda62f1e3f031b69657c4d783)

### Changed
Replaced "modifier" pattern for modal sizes with single code example that combines all sizes by using javascript. Each size modal is opened with a button. 

This required me to add JS to the `example.js` file, even though it is only relevant on this page.


line47:
I think this is a good change! Nice :)

PR #547, opened 2019-11-04
[WNMGDS-290] - Update disabled styling for buttons

line47:
Currently, the disabled button styling is the same for all non-inverse button styles. We should change the default disabled to match the default outline style, similar to USWDS

### Screenshot

Screen Shot 2019-11-04 at 4 32 37 PM

### DIFF
--- a/packages/core/src/components/Button/Button.scss
+++ b/packages/core/src/components/Button/Button.scss
@@ -24,6 +24,13 @@
     color: $color-primary-darker;
   }
 
+  &:disabled {
+    background-color: $color-white;
+    border-color: $color-gray-lighter;
+    color: $color-gray-lighter;
+    pointer-events: none;
+  }
+
   &:active,
   &.ds-c-button--active {
     border-color: $color-primary-darkest;
@@ -48,6 +55,13 @@
   border-color: transparent;
   color: $color-white;
 
+  &:disabled {
+    background-color: $color-gray-lighter;
+    border-color: transparent;
+    color: $color-gray-dark;
+    pointer-events: none;
+  }
+
   &:focus,
   &:hover,
   &:active,
@@ -58,7 +72,7 @@
     color: $color-white;
   }
 
-  &:focus,   
+  &:focus,
   &.ds-c-button--focus {
     background-color: $button-primary-bg--focus;
   }
@@ -80,6 +94,13 @@
   border-color: transparent;
   text-decoration: underline;
 
+  &:disabled {
+    background-color: transparent;
+    border-color: transparent;
+    color: $color-gray-lighter;
+    pointer-events: none;
+  }
+
   &:focus,
   &:hover,
   &:active,
@@ -94,6 +115,13 @@
   background-color: $color-error;
   border-color: transparent;
   color: $color-white;
+
+  &:disabled {
+    background-color: $color-gray-lighter;
+    border-color: transparent;
+    color: $color-gray-dark;
+    pointer-events: none;
+  }
 
   &:focus,
   &:hover,
@@ -117,6 +145,13 @@
   border-color: transparent;
   color: $color-white;
 
+  &:disabled {
+    background-color: $color-gray-lighter;
+    border-color: transparent;
+    color: $color-gray-dark;
+    pointer-events: none;
+  }
+
   &:focus,
   &:hover,
   &:active,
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Button/Button.scss
```diff
@@ -24,6 +24,13 @@
     color: $color-primary-darker;
   }
 
+  &:disabled {
+    background-color: $color-white;
+    border-color: $color-gray-lighter;
```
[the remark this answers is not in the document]
line47: I don't think disabled buttons needs to pass color contrast, @davidakennedy ?

packages/core/src/components/Button/Button.scss
```diff
@@ -55,7 +56,8 @@
   border-color: transparent;
   color: $color-white;
 
-  &:disabled {
+  &:disabled,
+  &.ds-c-button--disabled {
     background-color: $color-gray-lighter;
```
line47: Do we need to have the `ds-c-button--disabled` class for each button variation? I was thinking using the disabled HTML attribute would be a better way to go if folks want to disable a button. 

>If you do disable a button, make sure it receives the disabled styling. A button element will automatically be styled as a disabled button when it has the disabled HTML attribute, but an a element will need to have the disabled modifier class applied to it.

bernardwang: Yes, we would still recommend to use the `disabled` attribute for buttons, but we should still provide both options for link elements styled as a button. Adding the `ds-c-button--disabled` also makes sure the styling is consistent regardless if you are using the classname or HTML attribute. This is also consistent with `.ds-c-button--focus`, `.ds-c-button--hover` and `.ds-c-button--active`.

packages/core/src/components/Button/Button.scss
```diff
@@ -161,6 +200,14 @@ Inverse buttons
   background-color: $color-background-inverse;
   border-color: $border-color-inverse;
 
+  &:disabled,
```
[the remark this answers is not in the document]
bernardwang: If we wanted to follow BEM exactly, I _think_ we should have classes like this:
```
ds-c-button ds-c-button--primary ds-c-button--primary-disabled
ds-c-button ds-c-button--primary ds-c-button--primary-inverse
ds-c-button ds-c-button--primary ds-c-button--primary-active
```
but since that's a bit verbose and can get confusing with the HTML attributes and multiple button states (think inverse + disabled), I think it makes more sense to do this:
```
ds-c-button ds-c-button--primary ds-c-button--disabled
ds-c-button ds-c-button--primary ds-c-button--inverse
ds-c-button ds-c-button--primary ds-c-button--active
```

bernardwang:
@line47 I pushed a commit illustrating what I suggested here:
https://github.com/CMSgov/design-system/pull/547#discussion_r342678171

Let me know what you think about this!

packages/core/src/components/Button/Button.scss
```diff
@@ -174,38 +150,17 @@
   }
 }
 
-.ds-c-button--disabled {
```
bernardwang: Because we added `.ds-c-button--disabled` to all the `&:disabled`, we can remove this separate class definition

packages/core/src/components/Button/Button.scss
```diff
@@ -238,21 +196,31 @@ Inverse buttons
...
   &.ds-c-button--active {
-    background-color: $color-background-inverse;
-    border-color: $color-background-inverse;
+    border-color: $color-base-inverse;
+  }
+}
+
+.ds-c-button--transparent-inverse {
+  &:disabled,
+  &.ds-c-button--disabled {
```
bernardwang: Previously the transparent inverse button didn't have a disabled state

packages/core/src/components/Button/Button.scss
```diff
@@ -238,21 +196,31 @@ Inverse buttons
   }
 }
 
-.ds-c-button--disabled-inverse,
```
bernardwang: This class has been deleted

line47:
This looks good, thanks @bernardwang

bernardwang:
Awesome just waiting on this:
https://github.com/CMSgov/design-system/pull/547/files#r342655126

packages/core/src/components/Button/Button.scss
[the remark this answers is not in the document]
davidakennedy: They don't. From:

> User Interface Components
> Visual information required to identify user interface components and states, except for inactive components or where the appearance of the component is determined by the user agent and not modified by the author;

https://www.w3.org/TR/WCAG21/#x1-4-11-non-text-contrast

Key part there is: **except for inactive components**.

PR #550, opened 2019-11-08
[WNMGDS-92] Add react badge component

bernardwang:
### Added
<Badge> react component and updated react examples.

### DIFF
--- a/packages/core/src/components/Badge/Badge.example.jsx
+++ b/packages/core/src/components/Badge/Badge.example.jsx
@@ -1,5 +1,17 @@
+import React, { Fragment } from 'react';
 import Badge from './Badge';
-import React from 'react';
 import ReactDOM from 'react-dom';
 
-ReactDOM.render(<Badge>Today</Badge>, document.getElementById('js-example'));
+ReactDOM.render(
+  <Fragment>
+    <Badge>Default badge</Badge>
+    <Badge variation="info">Variant badge</Badge>
+    <Badge variation="success">
+      Badge with <b>nested</b>
+    </Badge>
+    <Badge className="ds-u-font-size--base" variation="alert">
+      Large badge
+    </Badge>
+  </Fragment>,
+  document.getElementById('js-example')
+);
--- a/packages/core/src/components/Badge/Badge.example.jsx
+++ b/packages/core/src/components/Badge/Badge.example.jsx
@@ -8,7 +8,7 @@ ReactDOM.render(
     <Badge variation="info">
       Badge with <code>variation</code> prop
     </Badge>
-    <Badge size="medium">
+    <Badge size="big">
       Badge with <code>size</code> prop
     </Badge>
   </Fragment>,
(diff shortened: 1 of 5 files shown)


packages/core/src/components/Badge/Badge.example.jsx
```diff
@@ -5,12 +5,13 @@ import ReactDOM from 'react-dom';
 ReactDOM.render(
   <Fragment>
     <Badge>Default badge</Badge>
-    <Badge variation="info">Variant badge</Badge>
+    <Badge variation="info">Info badge</Badge>
     <Badge variation="success">
-      Badge with <b>nested</b>
+      Success Badge with <b>bold text</b>
```
bernardwang: Hm, the only thought I have is that this is supposed to demonstrate an example with more complex `children` or multiple nested elements. Whether its bold or something else doesnt actually matter

packages/core/src/components/Badge/Badge.example.jsx
```diff
@@ -5,12 +5,13 @@ import ReactDOM from 'react-dom';
 ReactDOM.render(
   <Fragment>
     <Badge>Default badge</Badge>
-    <Badge variation="info">Variant badge</Badge>
+    <Badge variation="info">Info badge</Badge>
```
bernardwang: Same with this, I would say this example is demonstrating the `variation` prop. Maybe something like "Badge with `variation` prop"?

packages/core/src/components/Badge/Badge.jsx
```diff
@@ -17,7 +17,7 @@ export const Badge = props => {
 
 Badge.propTypes = {
   /**
-   * Additional classes to be added to the root button element.
```
bernardwang: Nice catch!

bernardwang:
@line47 Referring to the naming, the most important thing for me is that the naming describes the case/functionality we want to highlight. In the HTML examples this was pretty straightforward because you want to show all the `classes`. However in the React examples, I feel its more important to showcase the React props and the different ways to use the React component. So that's why I chose the names that I did, in order to describe the React specific use cases rather than the styles, which is already covered in the HTML examples. What do you think about that?

packages/core/src/components/Badge/Badge.example.jsx
line47: Ah, I see what you are saying. Makes sense to leave it as is

bernardwang:
@line47 If you want to take a look again, I added a size prop and updated the examples!

packages/core/src/components/Badge/Badge.jsx
```diff
@@ -1,16 +1,40 @@
...
    * In most cases this will be the badge's label, but you could also use this
    * to nest more advanced JSX.
-  */
-  children: PropTypes.oneOfType([PropTypes.node, PropTypes.string]).isRequired
+   */
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.string]).isRequired,
+  /**
+   * Sets the font size of the Badge
+   */
+  size: PropTypes.oneOfType(['big']),
```
line47: I think the size propType should be `oneOf`. Right now the prop table shows `custom` where I think it should be showing `big` similar to the way the button component props are set up.
Screen Shot 2019-11-12 at 10 04 21 AM

bernardwang: thanks for the catch!

Issue #530, opened 2019-10-23
ReviewLink accepts ariaLabel as a prop but Review doesn't provide it


eppixdev:
**Describe the bug**
`Review` renders the `ReviewLink` component which accepts `ariaLabel` as a prop but `Review` doesn't pass the prop down.

**To Reproduce**
1. Pass `ariaLabel` to `Review`
2. `aria-label` not added to links

**Expected behavior**
Expect aria-label's to be added to edit and remove links

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

bernardwang:
Tracked here: https://jira.cms.gov/browse/WNMGDS-282

PR #551, opened 2019-11-08
[WNMGDS-282] Add editAriaLabel to Review

bernardwang:
### Summary
<ReviewLink> takes an `ariaLabel` prop to add extra info for screenreaders, but <Review> did not use this prop.

### Added
This PR adds `editAriaLabel` to <Review>, and passes that along to <ReviewLink>

### DIFF
--- a/packages/core/src/components/Review/Review.example.jsx
+++ b/packages/core/src/components/Review/Review.example.jsx
@@ -16,7 +16,7 @@ ReactDOM.render(
       Multiple Review components can be combined together one after another.
     </Review>
     <Review
-      heading="A Review component with edit content"
+      heading="A Review component with custom edit content"
       editContent={
         <div>
           <a href={editButtonHref}>Edit</a>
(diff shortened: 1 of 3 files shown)


bernardwang:
This PR closes https://github.com/CMSgov/design-system/issues/530

line47:
Hey @bernardwang I was thinking it would be good to show the `ReviewLink` component and props on the review pattern page in a similar way that we do with `ChoiceList and `Choice`.

_Ignore the content of the screenshot_
Screen Shot 2019-11-13 at 9 32 49 AM


I think we might also need to add a prop for something like `removeAriaLabel` as the remove link was also mentioned in issue #530

bernardwang:
@line47 I'm not sure if its necessary to add `ReviewLink` since I cant imagine it being used outside of the internal `Review` implementation because it's basically just a link arguably doesn't have to be separated into its own file. This is different from a component like  `Choice` which has unique styling/functionality and is used for custom form components. 

As far as the `ariaLabel`, my understanding is that `Review` and `ReviewLink` currently only support a single button (using `editHref`, `editText` & `onEditClick`). If someone wants an edit button and a remove button, they have to provide their own custom markup via `editContent`, and in that case they are responsible for supplying the right aria-labels.

line47:
Gotcha, I wasn't sure if it would be helpful to show the props for `ReviewLink` on the same page as review or not. So it's OK to not show those then yeah?

PR #553, opened 2019-11-14
Release v3.4.0

line47:


bernardwang:
This looks good, we can also try using `backstop approve` on this PR, with that command we should be able to update reference files without waiting to deploy to prod

PR #554, opened 2019-11-14
[WIP WNMGDS-287] Reorganize utilities pages and examples

bernardwang:
###
This is a very large PR so you will have to go through each page in order to compare the differences.

### Added
- Added **Responsive format** standard, every utility with breakpoints will also have this.

### Changed
- General content/example changes, attempt to have more consistent spacing with `example--wrapper` and `example--list`.
- Moved "Clearfix" page to be a section under "Float"
- Removed bulletpoints that show "Format", opt for bracketed list instead. 
   i.e.`Format: ds-u-justify-content--[start|end|center|between|around]`
- Extend `preview__grid` styling to be used in more examples rather than just the Grid examples. Use this styling with `preview__element` generally whenever we want to see an element's boundaries like in "Display/Visibility", "Float", and "Flexbox"

### Removed
- Removed `Roboto Mono` from doc page and opt to use `monospace` instead. This is because the `code` styling in the examples was different than the rest of the site. This also has the added benefit of removing a font asset request


packages/core/src/utilities/font-family/font-family.scss
```diff
@@ -1,38 +1,10 @@
 /* stylelint-disable declaration-no-important */
 @import '@cmsgov/design-system-support/src/settings/index';
 
-/*
-Font family
```
line47: I'm wondering if we can combine font family, style, size, and weight into one font page?

packages/core/src/utilities/line-height/_line-height.docs.scss
```diff
@@ -3,17 +3,9 @@ Line height
...
-Markup:
-<h2 class="preview__label">ds-u-leading--lead</h2>
-<div class="ds-u-leading--lead">{{lorem-l}}</div>
-<h2 class="preview__label">ds-u-leading--base</h2>
-<div class="ds-u-leading--base ds-u-font-size--base">{{lorem-l}}</div>
-<h2 class="preview__label">ds-u-leading--heading</h2>
-<div class="ds-u-leading--heading">{{lorem-l}}</div>
-<h2 class="preview__label">ds-u-leading--reset</h2>
-<div class="ds-u-leading--reset">{{lorem-l}}</div>
+**Format**: `ds-u-leading--[lead|base|heading|reset]`
```
line47: I think it would be helpful to show the spacing value for each leading, for example, base is `1.5`

packages/core/src/utilities/text-decoration/text-decoration.example.html
```diff
@@ -0,0 +1,16 @@
...
+    <code class="ds-margin-bottom--1">ds-u-text-decoration--underline</code>
+    <div class="ds-u-text-decoration--underline">Underlined text</div>
+  </article>
+  <article>
+    <code class="ds-margin-bottom--1">ds-u-text-decoration--line-through</code>
+    <div class="ds-u-text-decoration--line-through">line-through text</div>
+  </article>
+  <article>
+    <code class="ds-margin-bottom--1">ds-u-text-decoration--none</code>
+    <div class="ds-u-text-decoration--none">
```
line47: the `ds-u-text-decoration--none` class needs to be on the a tag

packages/core/src/utilities/word-break/word-break.example.html
```diff
@@ -0,0 +1,12 @@
+<section class="ds-l-container preview__grid">
+  <div class="ds-l-row">
+    <div class="ds-l-col--3 ds-u-word-break">
```
line47: I think changing the block from `ds-l-col--3` to `ds-l-col--2` better shows how the word break is working.
Maybe also consider showing them on separate lines?

packages/core/src/utilities/text-decoration/_text-decoration.docs.scss
```diff
@@ -3,15 +3,9 @@ Text decoration
 
 Set whether text is underlined or not using the `text-decoration` utility.
 
-**Format**: `ds-u-text-decoration--[VALUE]`
```
line47: Maybe consider grouping text alignment, decoration, and transform and possible word break under a new `text styles` page?

packages/core/src/utilities/background-color/_background-color.docs.scss
```diff
@@ -7,32 +7,7 @@ Use the background color utility to change the default background color of an el
 
 @hide-markup
 
-Markup:
-<% var groups = [
-  ['primary', 'primary-darker', 'primary-darkest'],
-  ['primary-alt', 'primary-alt-dark', 'primary-alt-darkest', 'primary-alt-light', 'primary-alt-lightest'],
-  ['gray', 'gray-dark', 'gray-light', 'gray-lighter', 'gray-lightest'],
-  ['gold', 'gold-light', 'gold-lighter', 'gold-lightest'],
-  ['warn', 'warn-light', 'warn-lighter', 'warn-lightest'],
```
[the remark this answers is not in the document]
bernardwang: yeah i like this

packages/core/src/utilities/border/_border.docs.scss
```diff
@@ -9,14 +9,7 @@ The border utility class sets the `border-color` and `border-size`, with support
 
 @hide-markup
 
-Markup:
```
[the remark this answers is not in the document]
bernardwang: ah yeah, the issue with stacking vertically is that it becomes a lot more verbose (allot more empty white space) and you lose the subgrouping (all borders, vertical, horizontal). I’m a bit divided on this, because if we add pixels we should add to everything, but this one kinda works without it, especially considering the pixels is just 1,2,3px in this case

packages/core/src/utilities/color/_color.docs.scss
```diff
@@ -3,10 +3,10 @@ Color
 
 Use the color utility to change an element's text color.
 
-**Format**: `ds-u-color--[COLOR NAME]-[SHADE]`
-
 The options below offer color palette combinations that fall within the range of Section 508 compliant foreground/background color contrast ratios. To ensure that text remains accessible, **use only these permitted color combinations**.
 
+**Format**: `ds-u-color--[COLOR NAME]-[SHADE]`
+
 @hide-markup
```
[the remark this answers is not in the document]
bernardwang: yeah i didn’t change this one much, i’m not sure how handle it. we have specific guidance that we need to use those color combos, so that’s why i don’t know if we can remove combos. i was considering grouping by background color, but then later on there are combos that are grouped by text color...

packages/core/src/utilities/font-family/font-family.scss
bernardwang: yeah i was thinking of leaving that for a separate pr. same with border and border radius. margin and padding is another candidate (the examples r pretty much the same). i also think it should be another PR because maybe our vertical nav should be updated to have subsections, or some other change to expose these newly hidden sections

packages/core/src/utilities/text-decoration/_text-decoration.docs.scss
bernardwang: same thought here: https://github.com/CMSgov/design-system/pull/554#discussion_r347431092

packages/core/src/utilities/font-family/font-family.scss
bernardwang: yeah i was thinking of leaving that for a separate pr. same with border and border radius. margin and padding is another candidate (the examples r pretty much the same). i also think it should be another PR because maybe we should consider updating our vertical nav to have subsections, or some other change to expose these newly hidden sections

bernardwang:
@line47 
It seems like the two biggest feedback is to combine certain pages and show specific pixels values for utilities. I think both are worth doing but considering how large this PR already is, would be better handled in a diff PR (especially because i would like to be able to move on to other work this week before i leave).

Showing pixel values can be easily done for certain pages (vertical ones), but I would like to think over how they can be done on the border page. Similarly, the color and background pages are more difficult. Even if background color is easy, it feels like it should be consistent with the color page, which is another thing to consider.

So I can fix those small changes you commented on, but would you be ok with me filing two new tickets for the other major feedback?

line47:
That sounds great! Thanks @bernardwang  🎉

PR #560, opened 2019-11-25
[FLH-108] Add a note about using <Downshift> props on the <Autocomplete> component.

LouisFettet:
This should be pretty straightforward; while working on [CMS-WDS/flh-consumer#249](https://github.cms.gov/CMS-WDS/flh-consumer/pull/249) I was running into issues with setting an initial value for the `<Autocomplete>` component as well as keeping the input from clearing "on blur" while using the existing props in the documentation. Once I did a bit of digging, I realized I could pass props specific to the `<Downshift>` component via the `<Autocomplete>` component, so I just wanted to add a note about that in case it's useful to others in the future.

### Added
* a brief note about providing `<Downshift>` props to the `<Autocomplete>` component

### Removed
* a reference to the version of `<Downshift>` being used since it would be difficult to keep this updated

### DIFF
--- a/packages/core/src/components/Autocomplete/_Autocomplete.docs.scss
+++ b/packages/core/src/components/Autocomplete/_Autocomplete.docs.scss
@@ -28,7 +28,7 @@ Style guide: components.autocomplete.react
 
 ### Usage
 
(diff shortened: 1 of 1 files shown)


line47:
This looks good to me, thanks for finding @LouisFettet

PR #561, opened 2019-11-26
[NO-TICKET] Allow <Choice> component to specify className for child <FormLabel> component.

LouisFettet:
This is rather minor, but while working on [CMS-WDS/flh-consumer#279](https://github.cms.gov/CMS-WDS/flh-consumer/pull/279) I noticed we had [a style rule](https://github.cms.gov/CMS-WDS/flh-consumer/blob/a2abc8ffe8c9d2922b4eb57219cd9773ffeb105b/src/stylesheets/app.scss#L172-L175) in place to apply a bold `font-weight` on a `<FormLabel>` under a `<Choice>` component, and I figured it would be better to apply this via a `className` prop instead.

Screenshot of how we use a bold label:

### Added
* a new `labelClassName` prop to the `<Choice>` component so that classes can be provided to the child `<FormLabel>` component

### Notes
* An issue I foresee is that the provided `labelClassName` also affects the `hint` and `requirementLabel` content, which may or may not be what we want. If others think this is an issue, we can pass the `labelClassName` prop of `<Choice>` to the `<FormLabel>`'s `labelClassName` prop instead of its `className` prop. Might also be worthwhile to explicitly say in the documentation for the prop that it will may also affect the `hint` and `requirementLabel` content.
* There were more prettier changes here than I expected, but from some quick checks I believe these formatting changes are correct.


bernardwang:
Thanks for putting in this PR @LouisFettet! It seems like this probably isn't a super common case, so I think I'm ok with `labelClassName` on Choice` corresponding to `className` on `FormLabel`. 

Passing props and having conflicting/confusing prop names is a recurring problem on the DS, and if someone also wanted to style the `labelClassName` on `FormLabel`, we might want to opt for a `customFormLabel` or `customChoiceLabel` prop instead. That solution would be more flexible and potentially less confusing, but it's slightly harder to use and there's not a lot of evidence that users need for that level of control yet. @LouisFettet @scitech thoughts?

LouisFettet:
@bernardwang I agree that this is an edge case, and I think that the ideal state is probably the one you outline, where you can provide a `FormLabel` node directly to the `Choice` component (I think this could be done via a `prop` that takes a node, and we default to the current implementation if none is provided?).  

I'm unsure if this would be immediately useful to anyone (besides the situation I outlined in the description), but I'm willing to take a stab at it if y'all think it's a good path forward.  Otherwise, perhaps we leave a comment / TODO somewhere and continue with what's here?

line47:
I think there is value in exploring this type of option but I also think there is some deeper work that needs to happen to use props consistently and in a more understandable way across the design system. 

labelClassName isn't used very consistently in that in some cases it applies the class to the label and in some cases, it applies the class to the legend.

Additionally, I think the styling for form label need to be adapted to give label, hint, and requirement label their own classes that can be used independently. Right now there is some grouping to that element that I think makes alternate styling harder than it should be.

bernardwang:
> you can provide a `FormLabel` node directly to the `Choice` component (I think this could be done via a `prop` that takes a node, and we default to the current implementation if none is provided?).

@LouisFettet Yes, I think this sounds good. Even though this will only be useful to your specific case, I guess its the same story with the other solution as well. If we are going to make a change for a small case we might as well make it as flexible and forward compatible as possible. Let me know if you can update this PR, or if you want, I can handle it too, it shouldnt take too long.

LouisFettet:
@bernardwang sounds good, I can try to knock that change out some time this week.

bernardwang:
Hey @LouisFettet! just taking a second look at this, I was looking at other components with FormLabel to see if they support this and I'm actually thinking its fine to keep it simple and add `labelClassName`. Only `FormLabel` has a prop called `labelClassname` where it is used differently than `TextField`, `Dropdown`, etc.  I can submit a quick PR to do this if you'd like

packages/core/src/components/ChoiceList/Choice.test.jsx
```diff
@@ -94,6 +94,11 @@ describe('Choice', () => {
     expect(input.hasClass('ds-c-choice--inverse')).toBe(false);
   });
 
+  it('applies className to label', () => {
```
bernardwang: Thanks for adding this test!

LouisFettet:
Thanks for pushing this through y'all; I got super sidetracked with other work and never got a chance to circle back here.  My apologies.

Issue #562, opened 2019-12-02
Compiled CSS files no longer included in Release .zip downloads


sawyerh:
**Describe the bug**
I _think_ this used to work. The [documentation details](https://design.cms.gov/startup/sass-and-css/) that a CSS file is included in a release's Zip download, but it currently is not, and it's not immediately clear how someone could get a minified CSS distribution of the design system.

**To Reproduce**
1. Go to the [latest release](https://github.com/CMSgov/design-system/releases/tag/v3.4.1)
2. Download the "Source code" asset
3. Unzip the download and navigate into `packages/core`

**Expected behavior**
Expect a `packages/core/dist/index.css` file to exist after unzipping the download

line47:
I think this bug was introduced here - updating the gitignore should fix it 👍 Thanks @sawyerh

bernardwang:
Tracked here: https://jira.cms.gov/browse/WNMGDS-267

bernardwang:
Should be fixed in the latest doc site deployment!

PR #563, opened 2019-12-02
Add compiled index.css file back to dist folder 

line47:
Add index.css file back to the GitHub repo per the https://design.cms.gov/startup/sass-and-css/ documentation

Fixes https://github.com/CMSgov/design-system/issues/562

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,9 @@
 # Documentation is generated into this directory
 /docs
 
-# Do not commit package build files to github
-/packages/core/dist
+# Do not commit package build files other than the compiled CSS
+/packages/core/dist/**/
+/packages/core/dist/index.js
 /packages/layout/dist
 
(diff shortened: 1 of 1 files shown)


.gitignore
```diff
@@ -1,8 +1,9 @@
 # Documentation is generated into this directory
 /docs
 
-# Do not commit package build files to github
-/packages/core/dist
+# Do not commit package build files other than the compiled CSS
+/packages/core/dist/**/
```
bernardwang: I think an ideal solution would have the Release .zip have the dist files while also preventing dist files from being in GH. Related to [this ticket](https://jira.cms.gov/browse/WNMGDS-267)

bernardwang: This issue seems relatively high priority to fix, but I think it still might be worth looking into release.zip before reverting this .gitignore

line47:
Closing this PR to look into a more permanent solution.

PR #573, opened 2019-12-13
[WNMGDS-325] Fix position sticky in IE11

bernardwang:
### Summary
`<HelpDrawer>` uses `position: sticky` to position the help drawer title, this style is not supported in IE11. 

### How to test
1. Run the branch locally
2. Edit `HelpDrawer.example.jsx` to use a very long title on line 47
3. Install and use [ngrok](https://ngrok.com/) or [localtunnel](https://localtunnel.github.io/www) in order to expose your local branch to a public URL
4. Setup live testing on IE11 on SauceLabs or your own VM to test. 
5. Ensure the long title does not overlap with the help drawer content


line47:
I'm getting an error when trying to run this locally. 
```
[11:01:13] 'docs:build' errored after 1.36 s
[11:01:13] ReferenceError in plugin "run-sequence(build:dev)"
Message:
    window is not defined
Stack:
ReferenceError: window is not defined
    at Object.<anonymous> (/Users/scottweber/Documents/github/design-system/node_modules/stickyfilljs/dist/stickyfill.js:546:4)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/scottweber/Documents/github/design-system/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/scottweber/Documents/github/design-system/packages/core/dist/components/HelpDrawer/HelpDrawer.js:22:21)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/scottweber/Documents/github/design-system/node_modules/babel-register/lib/node.js:152:7)
    at Module.load (internal/modules/cjs/loader.js:599:32)
    at tryModuleLoad (internal/modules/cjs/loader.js:538:12)
    at Function.Module._load (internal/modules/cjs/loader.js:530:3)
    at Module.require (internal/modules/cjs/loader.js:637:17)
    at require (internal/modules/cjs/helpers.js:22:18)
    at Object.<anonymous> (/Users/scottweber/Documents/github/design-system/packages/core/dist/components/index.js:139:19)
    at Module._compile (internal/modules/cjs/loader.js:689:30)
    at Module._extensions..js (internal/modules/cjs/loader.js:700:10)
    at Object.require.extensions.(anonymous function) [as .js] (/Users/scottweber/Documents/github/design-system/node_modules/babel-register/lib/node.js:152:7)
```

packages/core/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -48,12 +48,16 @@ export class HelpDrawer extends React.PureComponent {
             </Button>
           </div>
         </div>
-        <div className="ds-c-help-drawer__body ds-u-md-font-size--small ds-u-lg-font-size--base ds-u-padding--2">
-          {children}
-        </div>
-        <div className="ds-c-help-drawer__footer ds-u-fill--primary-alt-lightest ds-u-md-font-size--small ds-u-lg-font-size--base ds-u-padding--2">
-          <h4 className="ds-text ds-u-margin--0">{footerTitle}</h4>
-          <div className="ds-text ds-u-margin--0">{footerBody}</div>
+        <div className="ds-c-help-drawer__body">
```
bernardwang: @davidakennedy @line47
I referencing [usa spending's glossary](https://www.usaspending.gov/#/) during this PR, and I noticed their markup used `aside`, `role: dialog`, and even a div with `role: navigation` wrapping the close button. 
https://gyazo.com/4caf6f1a806eb67a77d5f4d1ab58d939

Without researching too much, some of these sound like they should apply to us, i.e. I could see our help drawer using `aside`, `header`, and `footer`. I also need to do more screen reader testing for the roles.

What do yall think, should we update our markup to be more semantic as well? And if so, what should we change?

davidakennedy: > Without researching too much, some of these sound like they should apply to us, i.e. I could see our help drawer using aside, header, and footer. I also need to do more screen reader testing for the roles.
> 
> What do yall think, should we update our markup to be more semantic as well? And if so, what should we change?

This is a good question to ask. Here are some of my thoughts:

As currently implemented, I see the help drawer more like a [nonmodal dialog](https://www.nngroup.com/articles/modal-nonmodal-dialog/), maybe even closer to an accordion in functionality, but not appearance.

The spending glossary is a bit different in that they trap focus, like a true modal, even if the visual design is more like a nonmodal. It's also more complex than our help drawer in that when you click buttons in it, it slides to the word definition. Our drawer is usually used in less complex ways.

In most of the ways I've seen the help drawer used, it's less about needing to grab people's attention with a must-do task and more about read this important information. So I don't think the `role="modal"` markup makes sense for us. Unless people are using it in more complex ways. But maybe that means we need a separate component. That role cancels out the normal role of an `<aside>` element, which is complementary. So using an `<aside>` in the spending glossary is kind of a wash. If we used an `<aside>` in the help drawer, we get [this warning in aXe](https://dequeuniversity.com/rules/axe/3.3/landmark-complementary-is-top-level). So I wouldn't do that. I do think `<header>` and `<footer>` would help, but only if the content is log enough (to be syndicated, like an article), and the header and footer elements had other elements in them. Like dates, categories, etc. I probably wouldn't add them now.

I do think we could improve the help drawer button semantics with `aria-pressed` or `aria-expanded and aria-controls`. That would give the control pressed to open the drawer more meaning.

bernardwang: Gotcha, thank you for looking into this!

My question is, is there any benefit to adding an aside, and do you think that accurately describes the help drawer? I think we might be able to fix that aXe error if there’s some benefit to using aside

And as far as header/footer, I believe the header is always pretty simple (title and a button for closing), and footers on App3 include some links and phone numbers for additional support. Also, help drawer content can range from a paragraph, to a lot of text with lists and links. The main reason why we wouldn’t use header/footer in this case is because most use cases aren’t complex enough to warrant this right?

bernardwang: I was thinking about `aria-expanded/aria-controls` and `aria-pressed`.  I wonder if it makes sense to add `aria-expanded/aria-controls` considering you will only see the button if `aria-expanded` is true. In my mind expanded/controls is more suited for a button that is always visible, but toggles an expandable section. I feel similarly with `aria-pressed`, which in my understanding is for simulating a "toggle button" for screen readers, which doesnt seem as applicable in this situation.

davidakennedy: > is there any benefit to adding an aside, and do you think that accurately describes the help drawer?

There's not a strong benefit to adding one from an assistive technology perspective. Screen reader users will be able to navigate/skip to it because it's a landmark, but because it would be nested in another landmark (most likely), screen reader users won't be able to skip over it entirely. It goes against a best practice, but not the end of the world either. This is one of those cases that's hard because we won't always know the markup around the component.

I do think an aside is the right element for this type of content, at least how I've seen the help drawer used in the wild.

> The main reason why we wouldn’t use header/footer in this case is because most use cases aren’t complex enough to warrant this right?

Right. We could maybe make that an option with the ability to add additional markup, if needed? I could see it being needed in some cases for sure.

Re: `aria-expanded/aria-controls` and `aria-pressed`, this is a tough case because the component is somewhere between an accordion and modal. When I was thinking about `aria-expanded/aria-controls` and `aria-pressed`, I was thinking of adding it to the button that triggered opening the help drawer. That way, a person would know that the button is going to expand something when clicked.

bernardwang: Ah those explanations help a lot thank you! It sounds like for now its probably not a big priority to add `aside` and `header/footer` than. 

Re: `aria-expanded/aria-controls` and `aria-pressed`, I see what you mean about putting it on the button that triggers opening, that makes sense. I feel like the fact that the open and close buttons are separate is slightly more similar to a modal, so its definitely tricky. 

This was really good discussion and context, thank you @davidakennedy again! My understanding is that theres no urgent a11y changes that this PR should wait on.

line47:
I'm on it 👍

packages/core/src/components/HelpDrawer/__snapshots__/HelpDrawer.test.jsx.snap
```diff
@@ -27,28 +27,36 @@ exports[`HelpDrawer renders a snapshot 1`] = `
...
-      <div>
+      <div
+        className="ds-c-help-drawer__content ds-u-md-font-size--small ds-u-lg-font-size--base ds-u-padding--2"
+      >
         <p>
-          Some footer content
+          content
         </p>
       </div>
+      <div
```
line47: I think the footer div needs to be moved to after the closing `ds-c-help-drawer__body` div so it lives outside of the body. Right now the footer is floating up the page. This is in Chrome and IE 11. 
Screen Shot 2020-01-14 at 12 24 02 PM

packages/core/src/components/HelpDrawer/HelpDrawer.scss
```diff
@@ -35,16 +34,22 @@
...
+  position: relative;
+  width: 100%;
 }
 
-.ds-c-help-drawer__body {
-  // Stretch the body so that the footer sticks to the
-  // bottom of the screen
-  flex: 1 0 auto;
+.ds-c-help-drawer__content-container {
+  bottom: 0;
```
line47: I don't think the styles for `ds-c-help-drawer__content-container` are needed I removed them and got the same experience after moving the footer.

line47:
When I use a long title the close button gets squashed to the right see the screenshot below.
I think it is due to the flex of the header coupled with the max-width setting on the entire help drawer. 
Screen Shot 2020-01-14 at 12 52 44 PM

packages/core/src/components/HelpDrawer/__snapshots__/HelpDrawer.test.jsx.snap
bernardwang: Hm yeah, the tricky thing is that if the help drawer body is very long, we want the footer to be hidden until you scroll all the way down. [This is the existing behavior at least](https://gyazo.com/473039de8365b93b577b8b23a33a43f9), I'm also open to changing it

bernardwang:
@line47 Ah yeah, the two issues you mentioned are introduced in [this last commit](https://github.com/CMSgov/design-system/pull/573/commits/fe5d27afc7dc00fd3110adbac8a9258fe21d6a07). Previously `ds-c-help-drawer__content-container` applied flexbox and allowed for a `flex-grow` that would always push the footer to the bottom. However, this wasnt working on IE11 some reason so I tried to remove it. 

A simple fix would be to change the existing behavior and always show the footer as you suggested [here](https://github.com/CMSgov/design-system/pull/573#discussion_r366477086), but we should get design to take a look

bernardwang:
@line47 Looks like its working after all! I think the reason why I thought it wasnt working was because the browser was still updating the CSS or I was testing an outdated version.

The issue with the compressed "close" button still exists however, I believe this was an existing issue.

line47:
It looks like using `display: table` on the close button for the help drawer fixes the clipping issue. Not sure if that is the solution we want to go forward with or not. It seems to be an issue with flex and IE11.

packages/core/src/components/HelpDrawer/HelpDrawer.scss
```diff
@@ -35,20 +35,26 @@
...
 
 .ds-c-help-drawer__body {
-  flex-grow: 1; // Stretch so that the container takes up the rest of the screen
-  position: relative;
-  width: 100%;
+  // flex-grow: 1; // Stretch so that the container takes up the rest of the screen
+  // position: relative;
+  // width: 100%;
+    // height: 100vh;
+    overflow-y: scroll;
```
line47: only needing to scroll vertically

packages/core/src/components/HelpDrawer/HelpDrawer.scss
```diff
@@ -35,20 +35,26 @@
...
-  bottom: 0;
+  // bottom: 0;
   display: flex; // Flex layout helps stick the footer to the bottom
   flex-direction: column;
-  left: 0;
-  overflow: scroll;
-  position: absolute;
-  right: 0;
-  top: 0;
+  height: calc(100vh - 60px);
```
line47: don't feel great about this height calculation and am wondering if there is a better way

bernardwang: what is the height calculating? the footer height?

bernardwang:
> It looks like using `display: table` on the close button for the help drawer fixes the clipping issue. Not sure if that is the solution we want to go forward with or not. It seems to be an issue with flex and IE11.

If this works you should include it in this PR!

packages/core/src/components/HelpDrawer/HelpDrawer.scss
bernardwang: I'm getting this: https://gyazo.com/29616b924a6aebd5bdc752748b0401ba

packages/core/src/components/HelpDrawer/HelpDrawer.scss
```diff
@@ -40,13 +40,15 @@
...
     overflow-y: scroll;
+   
 }
 
 .ds-c-help-drawer__content-container {
   // bottom: 0;
   display: flex; // Flex layout helps stick the footer to the bottom
   flex-direction: column;
-  height: calc(100vh - 60px);
+  padding-bottom: 80px;
```
bernardwang: I don't think we can use styles that set a specific value because both footer and body can be arbitrary lengths. We'll have to look into something other than flexbox too because of IE11 :/ 

Screen Shot 2020-01-22 at 11 00 48 AM

packages/core/src/components/HelpDrawer/HelpDrawer.scss
```diff
@@ -43,14 +43,26 @@
...
 }
+.ds-c-help-drawer__content-container {
+  display: flex;
+  flex-direction: column;
+  height: 100%;
+}
 
 .ds-c-help-drawer__content {
+  flex-grow: 1;
+  flex-shrink: 0;
```
bernardwang: Ah was the key to making it work flex-shrink?

bernardwang:
@line47 Amazing work! The `flex-shrink` fix did the trick, if you can give this a ✅ I think we are good to go!

PR #575, opened 2019-12-16
[NO TICKET] Update github link to use uploaded zip

bernardwang:
### Changed
The "download code and design files" button no longer uses the default source code zip file, it uses the following URL:
```
https://github.com/CMSgov/design-system/releases/download/3.4.2/design-system-v3.4.2.zip
```
From now on, each release should have a zip uploaded with build files named in the following format:
`design-system-v${pkg.version}.zip`

### DIFF
--- a/guides/RELEASE-PROCESS.md
+++ b/guides/RELEASE-PROCESS.md
@@ -1,5 +1,23 @@
 # Release process
 
+1. **Create a release draft on GitHub**
+
+   1. [Draft a new release on GitHub](https://github.com/CMSgov/design-system/releases/new)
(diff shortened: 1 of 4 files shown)


line47:
I found a few other places we need to update the download link

- https://github.com/CMSgov/design-system/blob/017476e06b3f002f8e36bb053f81edc5148fc968/packages/docs/src/pages/startup/installation.md
- https://github.com/CMSgov/design-system/blob/master/packages/docs/src/pages/index.md

bernardwang:
@line47 I pushed a commit that changes those links to point to `https://github.com/CMSgov/design-system/releases/latest`. Those are just markdown files so they can't be dynamic links, I'm thinking pointing them to the latest release should be enough.

PR #577, opened 2019-12-16
[WNMGDS-309] Create flex utilities

bernardwang:
### Added
- `.ds-u-flex-direction-[]` utility and documentation

### Changed
- Updated the flex-wrap documentation

### DIFF
--- a/packages/core/src/utilities/flexbox/_flexbox.docs.scss
+++ b/packages/core/src/utilities/flexbox/_flexbox.docs.scss
@@ -4,7 +4,7 @@ Flexbox
 
 Use the [flexbox](https://css-tricks.com/snippets/css/a-guide-to-flexbox/) utility classes to control various flexbox properties. These are especially useful when combined with [flexbox grid](/layout/grid#layout.grid.alignment) classes.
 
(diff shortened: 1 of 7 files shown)


line47:
This looks good 👍 
I know this wasn't in the original ticket but do you think we should add the `Format ` and `Responsive format `

bernardwang:
@line47 Are you talking about adding to [this section](https://gyazo.com/7bffd974322eefff3be75560e5f3df54)? Currently none of those "Responsive" sections show format because they are already included under each specific utility. I.e. in Flexbox there are 4 different utilities/formats, so I'm not sure it's possible to include a single format for a general responsive flexbox section.

line47:
Ah yeah, that makes sense, maybe what I'm thinking is that there isn't any explanation of what type of breakpoint styling is happening there or what to expect. 

The styling used in the example is `ds-u-sm-justify-content` so maybe we say that above the example so folks know?

bernardwang:
Ah I see. Given that thats a change that would have to be made across many pages (none of the responsive examples do that I believe), would you be ok with leaving that for another ticket?

bernardwang:
Giving it some more thought, we also already link to the [breakpoints page](http://localhost:3000/guidelines/responsive/) on every responsive example, and the code is always available for people to see. Since the main goal of the responsive examples is to show the styles changing with different breakpoints, I could also see the argument that its not a huge priority to expose the utility classes.

PR #580, opened 2019-12-17
[WNMGDS-89] Use size prop

bernardwang:
### Changed
Used `size` prop in `ChoiceList`, removed ability to pass in a 'medium' size value which only applied to  `Select`, which is currently deprecated and discouraged from use anyway

### DIFF
--- a/packages/core/src/components/ChoiceList/ChoiceList.jsx
+++ b/packages/core/src/components/ChoiceList/ChoiceList.jsx
@@ -234,6 +234,9 @@ ChoiceList.propTypes = {
    */
   onComponentBlur: PropTypes.func,
   onChange: PropTypes.func,
+  /**
+   * Sets the size of the checkbox or radio button
+   */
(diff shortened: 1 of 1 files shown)


packages/core/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -237,10 +234,7 @@ ChoiceList.propTypes = {
    */
   onComponentBlur: PropTypes.func,
   onChange: PropTypes.func,
-  /**
-   * If the component renders a select, set the max-width of the input either to `'small'` or `'medium'`.
-   */
-  size: PropTypes.oneOf(['small', 'medium']),
```
line47: I would maybe add a description like `Sets the size of the checkbox or radio button to be small`

bernardwang: I added the comment!

PR #582, opened 2019-12-19
[WNMGDS-341] Misc Button fixes

bernardwang:
### Changed
- Allow the usage of `ds-u-button--inverse` with primary/success/danger variations

### Fixed
- Fix border color not styled correctly for inverse buttons
- Fix disabled styling for buttons using the `href` prop
- Use updated buttons on doc site GithubLinks
- Fix doc site GithubLink styling

### DIFF
--- a/packages/core/src/base/body.scss
+++ b/packages/core/src/base/body.scss
@@ -1,3 +1,4 @@
+/* stylelint-disable selector-class-pattern */
 @import '@cmsgov/design-system-support/src/settings/index';
 
 /*
@@ -37,3 +38,4 @@ Style guide: style.base
   background-color: $color-background-inverse;
   color: $color-base-inverse;
 }
+/* stylelint-enable */
--- a/packages/core/src/base/body.scss
+++ b/packages/core/src/base/body.scss
@@ -37,18 +37,5 @@ Style guide: style.base
   @extend %ds-base;
   background-color: $color-background-inverse;
   color: $color-base-inverse;
-  a {
-    color: $color-base-inverse;
-
-    &:visited,
-    &:active,
-    &:focus,
-    &:hover {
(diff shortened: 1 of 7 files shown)


packages/core/src/base/body.scss
```diff
@@ -37,18 +37,5 @@ Style guide: style.base
   @extend %ds-base;
   background-color: $color-background-inverse;
   color: $color-base-inverse;
-  a {
```
bernardwang: Moved link styling to `packages/core/src/base/typography/link.scss` for better organization

packages/core/src/components/Button/Button.jsx
```diff
@@ -58,18 +58,9 @@ export class Button extends React.PureComponent {
 
   classNames() {
     const variationClass = this.props.variation && `ds-c-button--${this.props.variation}`;
-    const disabledClass = this.props.disabled && 'ds-c-button--disabled';
+    const disabledClass = this.props.disabled && this.props.href && 'ds-c-button--disabled';
```
bernardwang: We only want to add the `ds-c-button--disabled` class for Buttons with a `href` prop (aka `<a> elements styled like a button), because normally we just use the `disabled` attribute

packages/core/src/components/Button/Button.scss
```diff
@@ -56,18 +60,19 @@
   padding: $spacer-half $spacer-1;
 }
 
-.ds-c-button--primary,
-.ds-c-button--primary:visited {
-  background-color: $button-primary-bg;
+.ds-c-button--transparent,
```
bernardwang: This whole section is a terrible diff, basically I moved the css for `.ds-c-button--primary`, danger and success to the end of the file so that it would override inverse styles

davidakennedy:
Changes look good, @bernardwang!

PR #587, opened 2020-01-03
[WNMGDS-330] Remove USWDS dep

bernardwang:
### Summary
We only use USWDS in `support` as a basis for our scss variables. Since most of the variables are overridden and the variables in the updated USWDS system have been overhauled, we have diverged enough to simply remove USWDS as a dependancy.

### Changed
- Combined USWDS variables into existing scss files
- Removed uswds from support's package.json and the gulp tasks

### Removed
- Removed USWDS breakpoints code because we already provide our own breakpoints system
- Removed $georgia, $helvetica sass variables
- Removed references to `lighten` and `darken` in the color variable file, as most of the comments were incorrect


packages/support/src/settings/_index.scss
```diff
@@ -1,22 +1,14 @@
-// Vendor overrides
-@import 'override.uswds';
-
-// Vendor variables
-// Defined after overrides so overrides work
-
 // Manually set Bourbon's font stack variables. This allows us to avoid
 // installing Bourbon as a USWDS dependency.
 $georgia: 'Georgia', 'Cambria', 'Times New Roman', 'Times', serif;
```
bernardwang: I don't see these font variables used anywhere, but I left them in just in case

packages/support/src/settings/_variables.layout.scss
```diff
@@ -64,3 +49,15 @@ $grid-form-gutter-width: $spacer-2 !default;
 
 // Z-index
 $z-dialog: 1000 !default;
+
+// Site variables
+$site-margins: 3rem !default;
```
bernardwang: These variables are *mostly* unused in our repo, however I am keeping them because some are used i.e. `$site-max-width`

packages/support/src/settings/_variables.shadow.scss
```diff
@@ -8,3 +8,8 @@ $shadow-base-blur-radius: 4px;
 $shadow-base-color: rgba(0, 0, 0, 0.25);
 $shadow-base: $shadow-base-offset-x $shadow-base-offset-y $shadow-base-blur-radius
   $shadow-base-color;
+
+// Inherited from USWDS 
+$box-shadow: 0 0 2px $shadow-base-color !default;
```
bernardwang: We should standardize these in a future PR

packages/support/src/settings/_index.scss
line47: I don't see them used either and am thinking we should remove them

packages/support/src/settings/_variables.layout.scss
line47: Couldn't we remove the unused variables? What's the reasoning for keeping them? Maybe someone else is using them at another design system level?

packages/support/src/settings/_variables.shadow.scss
line47: Is there a work item for this future work?

packages/support/src/settings/_variables.color.scss
```diff
@@ -36,16 +76,25 @@ $color-muted-inverse: #bac5cf !default;
 // Focus
 // USWDS has a default focus color, but we want one that can be applied to
 // fields that are displayed on top of an inverted background
```
line47: should we remove the 2 commented lines below `// Focus` ?

line47:
A few additional tweaks:

- I think the link on this page https://github.com/CMSgov/design-system/blob/master/packages/docs/src/pages/startup/site-package.md to typography variables needs to be updated as well.

- uswds@1.3.1 is in the packages/support yarn.lock file, I think we want to remove it from there as well.

packages/support/src/settings/_variables.layout.scss
bernardwang: Yes, in case someone who is using the CMS DS is using them. I would be ok with removing them normally, but some of these variables are being used by several of our components

packages/support/src/settings/_variables.shadow.scss
bernardwang: not yet! ideally it could be sorted out after the focus PR

bernardwang: Actually, once this PR is merged in to master we can handle it in https://github.com/CMSgov/design-system/pull/510

bernardwang:
@line47 feedback should be addressed!

PR #590, opened 2020-01-06
No ticket/background color utility page fix

line47:
Noticed that when running locally I was getting an error on the background-color.example.html file.
Prettier was formatting the var block in a way that wasn't allowing it to render 

it was doing this, which was causing issues 
```
<% Object.keys(groups).forEach(fill
  => { -%>
```


bernardwang:
Ah I saw this error, and I included it in DK's focus PR (I should have fixed it in a separate PR like this). When I looked at it I was able to fix it just by going from arrow functions to normal functions declarations like so:
https://github.com/CMSgov/design-system/pull/510/files#diff-0702339ea62e595fc8516fd32178c23cR10 
I think that should fix it without ignoring prettier

line47:
Updated and ready for another pass :)

PR #591, opened 2020-01-06
[WNMGDS-347] Fix ARIA label not being applied to Dropdown.

scitech:
Note: I haven't updated the snapshots or anything, just wanted to open this before I forgot.

### Fixed

This fixes the ARIA label not currently being applied to the `<Dropdown>` component.

image

The component was passing through the camel case `ariaLabel` prop, but ARIA props are special in JSX in that they use a hyphen like a regular html attribute: https://reactjs.org/docs/accessibility.html#wai-aria.


packages/core/src/components/Dropdown/Dropdown.test.jsx
```diff
@@ -40,7 +40,7 @@ function render(customProps = {}, optionsCount = 1, deep = false) {
 
 describe('Dropdown', () => {
   it('renders a select menu', () => {
-    const data = render({ value: '1' });
+    const data = render({ value: '1', ariaLabel: 'test aria label' });
```
1Copenut: @scitech I like this approach to adding the `aria-label` explicitly. Is there a way we can check for a `FormLabel` and throw a warning if the component has both present? Asking because an aria-label will override the label on the page, and that's almost never what we want. Form input labels are ideally written descriptively for all users.

My recommendation is to warn/error on this state if possible, or if not, note this as a warning in the component documentation.

Any further thoughts @davidakennedy ?

line47: Hey @1Copenut, for my own clarification, are you saying that `aria-label` should be used only if there isn't a `formLabel` in use?

1Copenut: Correct. We should not use an aria-label if we already have a `formLabel` associated with the select. The `formLabel` is the preferred means to identify a form input/select, but we can and should use the aria-label in cases like a checked child.

bernardwang: @1Copenut My last commit adds some documentation and a console warning!

packages/core/src/components/Dropdown/Dropdown.jsx
```diff
@@ -19,6 +19,18 @@ export class Dropdown extends React.PureComponent {
           `Cannot use 'options' and 'children' React properties at the same time in the <Dropdown> component. Please use 'children' for custom options and 'options' for general cases`
         );
       }
+      // 'ariaLabel' is provided with a `label` prop that is not an empty string
```
bernardwang: @scitech @1Copenut lmk if these checks work! I tried to add basic logs for when both `ariaLabel` and `label` are used, and when an empty `label` is used and an `ariaLabel` is necessary.

The checks are a little annoying because `label` is Proptypes.node, so it can be a string or html. There's still the chance that someone can put in `label={<div/>}` or something, and then the Dropdown won't have either `ariaLabel` or a correct `label`, but I'm assuming thats not common.

1Copenut: An `axe` check should pick up on improper labels, if they're missing the `htmlFor` attribute. Visual checks would hopefully spot HTML as string output.

PR #592, opened 2020-01-08
[WNMGDS-340] - Create link class

line47:
### Added
- Added `.ds-c-link` and `.ds-c-link--inverse` as class options


packages/core/src/base/typography/link.scss
```diff
@@ -27,7 +27,7 @@
 }
 
 // Change color of links that are descendants of .ds-base--inverse
-.ds-base--inverse a {
+.ds-base--inverse a, .ds-c-link--inverse {
```
line47: The default A tag is still styled by default and also is still styled as a child of `.ds-base--inverse`

packages/core/src/base/typography/_typography.docs.scss
```diff
@@ -18,7 +18,7 @@ Link
 
 Links connect users to a different page or further information.
 
-The only base HTML element the design system applies styling to is the `<a>` element. To prevent this, you can override the `$ds-include-base-html-rulesets` Sass variable and set it to `false`.
+The only base HTML element the design system applies styling to is the `<a>` element. To prevent this, you can override the `$ds-include-base-html-rulesets` Sass variable and set it to `false`. You can also use `.ds-c-link` and `.ds-c-link--inverse` for stlyling links.
```
bernardwang: typo here

bernardwang: Also, what do you think about tweaking the sentence to be:
 "you can also use ... classes for applying the link styling" 

Slightly different because there may be scenarios where you want to apply link styling to non `<a>` elements

line47: Ah I see, I'm trying to think of a scenario where you'd use the link styling which is only text color instead of using a utility class for changing the text color. I would think we would want to encourage that route as opposed to using the link class to style items other than links.

packages/core/src/base/typography/link.scss
```diff
@@ -8,7 +8,7 @@
   // would cause the selector's specificity to be higher than most other component
   // selectors (i.e. ds-c-button), necessitating overly specific selectors anytime
   // a developer wanted to change an anchor's color property.
-  a {
+  a, .ds-c-link {
```
bernardwang: This is inside the `$ds-include-base-html-rulesets` statement, when we actually want `.ds-c-link` to be always defined. One way we could potentially organize this is by using `extend` and sass placeholders

```
%link {
   color: $color-primary;
   &:hover { ... }
   ...
} 

%link-inverse {...}

.ds-c-link {
   @extend %link;
}

.ds-c-link--inverse {
   @extend %link-inverse;
}

@if $ds-include-base-html-rulesets {
   a {
      @extend %link;
   }

   .ds-base--inverse a {
       @extend %link-inverse;
   }
}

File: packages/core/src/base/typography/_typography.docs.scss (hunk 2 above)
Comment (bernardwang) [reply packages/core/src/base/typography/_typography.docs.scss]: Oh right, I think the link styling should also include `text-decoration: underline` as well. There still the issue of when we would want to style a non link element as a link, but personally I think providing that as a small option is useful, especially for writing CSS.

File: packages/core/src/base/typography/link.scss [hunk 4]
```diff
@@ -1,43 +1,61 @@
...
-  // <a> is the only base HTML element in the design system that
-  // has a style declaration. In all other cases, styles are applied using a
-  // namespaced class name. This selector isn't scoped under .ds-base, since that
-  // would cause the selector's specificity to be higher than most other component
-  // selectors (i.e. ds-c-button), necessitating overly specific selectors anytime
-  // a developer wanted to change an anchor's color property.
-  a {
-    color: $color-primary;
+%link {
+  color: $color-primary;
```
bernardwang: I think this should also have a `text-decoration: underline;` and `cursor: pointer;`

bernardwang: same with `%link-inverse`

packages/core/src/base/typography/link.scss
```diff
@@ -1,43 +1,61 @@
...
-// Change color of links that are descendants of .ds-base--inverse
-.ds-base--inverse a {
+%link-inverse {
   color: $color-base-inverse;
 
   &:visited,
   &:active,
   &:focus,
   &:hover {
-    // :visited links can only style color, not opacity. This color styling is equivalent to:
```
bernardwang: I think this comment is worth keeping

packages/core/src/base/typography/link.scss
```diff
@@ -1,43 +1,61 @@
 /* stylelint-disable selector-class-pattern */
 @import '@cmsgov/design-system-support/src/settings/index';
 
-@if $ds-include-base-html-rulesets {
-  // <a> is the only base HTML element in the design system that
-  // has a style declaration. In all other cases, styles are applied using a
-  // namespaced class name. This selector isn't scoped under .ds-base, since that
```
bernardwang:
```
  // namespaced class name. This selector isn't scoped under .ds-base, since that	
  // would cause the selector's specificity to be higher than most other component	
  // selectors (i.e. ds-c-button), necessitating overly specific selectors anytime
```
Interestingly I'm noticing this exact issue with inverse `<a>` elements.
`a` is scoped under `.ds-base--inverse`, which is very specific, making it difficult to change the color of inverse link elements.

bernardwang: This is an existing problem I believe. A good example of this is the links in the header:
Screen Shot 2020-01-09 at 11 21 59 AM
The styles from `c-header--title` are being overridden by `.ds-base--inverse a:visited`. I'm not entirely sure how to fix this, because we definitely want a way to default style links inside an inverse container.

line47:
These are some good updates and especially the link header catch. However, I'm a bit hesitant on creating the `.ds-c-link--darker` class and think that we can achieve the same outcome by using variables as we were previously.

PR #593, opened 2020-01-08
[WNMGDS-343] Update dropdown to allow custom options

bernardwang:
### Added
- `children` prop to `<Dropdown>` to allow for custom options
- additional example for custom options

### DIFF
--- a/packages/core/src/components/Dropdown/Dropdown.jsx
+++ b/packages/core/src/components/Dropdown/Dropdown.jsx
@@ -121,7 +121,7 @@ Dropdown.propTypes = {
    */
   className: PropTypes.string,
   /**
-   * Optional prop for defining custom dropdown options. When using the `children` prop, `options` should be an empty list.
+   * Used to define custom dropdown options. When using the `children` prop, `options` should be an empty list.
    */
   children: PropTypes.node,
   /**
--- a/packages/core/src/components/Dropdown/Dropdown.jsx
+++ b/packages/core/src/components/Dropdown/Dropdown.jsx
@@ -8,10 +8,17 @@ export class Dropdown extends React.PureComponent {
   constructor(props) {
     super(props);
 
-    if (props['fieldRef']) {
-      console.error(
-        `[Deprecated]: Please remove the React property 'fieldRef' for the <Dropdown> component. It is no longer supported and will be removed in a future release, use 'inputRef' instead.`
-      );
+    if (process.env.NODE_ENV === 'development') {
+      if (props.fieldRef) {
+        console.error(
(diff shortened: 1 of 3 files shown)


packages/core/src/components/Dropdown/Dropdown.jsx
```diff
@@ -96,6 +97,7 @@ export class Dropdown extends React.PureComponent {
           {...selectProps}
         >
           {optionElements}
+          {children}
```
bernardwang: This is one of those situations where we want either `options` or `children`, but theres no good way to have that be enforced with Proptypes. Because we almost always want `options` to be defined, I left `options` as a required prop, while `children` is optional.

packages/core/src/components/Dropdown/Dropdown.jsx
```diff
@@ -95,7 +96,8 @@ export class Dropdown extends React.PureComponent {
           /* eslint-enable no-return-assign */
           {...selectProps}
         >
-          {optionElements}
+          {/* Render custom options if provided */
+          children || optionElements}
```
bernardwang: This is one of those situations where we want either options or children, but theres no good way to have that be enforced with Proptypes. Because we almost always want options to be defined, I left options as a required prop, while children is optional.

scitech: You could provide a function to the options proptype, but it gets kind of complicated:
```
options: function(props) {
  if (props.children) return new Error('Cannot provide options and children at the same time');

  if (!props.options) return new Error('options prop must be defined');

  if (!Array.isArray(props.options)) return new Error('options prop must b an array')

  if (!props.options.every(option => option.label && option.value)) return new Error('options array does not match required shape')
}
```

It would be easier if you could use the PropTypes validators within the custom validator function but it looks like you can't. So I think it's probably simplest as you have it

bernardwang: Yeah I'll do this and just add some info in the react prop documentation!

bernardwang:
@scitech I pushed some commits to add some more docs and a warning, lmk if that looks good to you!

packages/core/src/components/Dropdown/Dropdown.jsx
```diff
@@ -111,6 +120,10 @@ Dropdown.propTypes = {
    * Additional classes to be added to the root element.
    */
   className: PropTypes.string,
+  /**
+   * Optional prop for defining custom dropdown options. When using the `children` prop, `options` should be an empty list.
+   */
+  children: PropTypes.node,
```
line47: I'm not seeing the `children` prop in the doc site prop table, should I be?
I see `className` and then `defaultValue `but don't see `children`

bernardwang: Tracked here: https://jira.cms.gov/browse/WNMGDS-349

Issue #595, opened 2020-01-13
Deprecation warnings should not log to console in production


scitech:
**Description**

When upgrading the design system in my project I encountered a lot of error-level console messages due to deprecated components.

I propose that the design system follow the convention used by libraries like React where deprecation warnings are warning-level and are only visible in development. So instead of:
```
console.error('<Select> is deprecated');
```

It could be:
```
if (process.env.NODE_ENV !== 'production') {
  console.warn('<Select> is deprecated');
}
```

Because it's wrapped in a `process.env.NODE_ENV` check, application build tools would be able to strip out the `console.warn` call when building for production. This way design system application developers will be able to see and fix the warnings, the production bundle will be slightly smaller, and applications will have less noise in our production console.

**Steps to reproduce the issue**

1. Use a `<Select>` component

bernardwang:
Ah thanks for the issue, this is a big oversight, we can get it in for the next release.

bernardwang:
Tracked by https://jira.cms.gov/browse/WNMGDS-346

bernardwang:
Should be fixed in v3.5.0

PR #596, opened 2020-01-13
[WNMGDS-346] Use console warn and remove logs from production

bernardwang:
### Changed
- Removes console.error in favor of warn
- Scopes logs to development and test envs only


bernardwang:
Closes https://github.com/CMSgov/design-system/issues/595

bernardwang:
We could add that, but I also think that most people assume a dependency shouldn't be logging out dev related warnings in production. TBH I feel like this is mainly an oversight on our part

bernardwang:
Also, this needs to wait for this PR to merge 
https://github.com/CMSgov/design-system/pull/593

PR #602, opened 2020-01-17
[WNMGDS-353] Rename labelClassName prop in <FormLabel>

bernardwang:
### Added
- `textClassName` prop to replace `labelClassName`

### Deprecated
- `labelClassName` prop from `<FormLabel>`

### Removed
- `labelClassName` prop from `<MonthPicker>`

### DIFF
--- a/packages/core/src/components/MonthPicker/MonthPicker.jsx
+++ b/packages/core/src/components/MonthPicker/MonthPicker.jsx
@@ -131,11 +131,9 @@ export class MonthPicker extends React.PureComponent {
   }
 
   renderLabel() {
-    const classes = classNames('ds-u-font-weight--bold', this.props.labelClassName);
     return (
       <FormLabel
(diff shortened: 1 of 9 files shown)


packages/core/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -131,11 +131,9 @@ export class MonthPicker extends React.PureComponent {
   }
 
   renderLabel() {
-    const classes = classNames('ds-u-font-weight--bold', this.props.labelClassName);
     return (
       <FormLabel
         className="ds-u-visibility--screen-reader"
-        labelClassName={classes}
```
bernardwang: I removed `labelClassName` prop in `MonthPicker` because this `FormLabel` was hidden regardless. I'm not sure if anyone was actually using the prop or if there is a legitimate use case I'm unaware of. 
@line47

PR #604, opened 2020-01-20
[WNMGDS-292] Support react router in <VerticalNav> and <StepList>

bernardwang:
### Summary
Currently we support react-router through a combo of `href` and `onClick` props, however, this method is not documented at all. This PR adds a more direct way to handle this use case through a `component` prop. This prop is used in `<Button>, <StepList> and <VerticalNav>` to allow rendering a custom component including `<Link>`. This is hopefully a more explicit and easily discoverable way to handle `react-router`

### Added
- `component` prop in `<VerticalNav>` and `<StepList>`. This prop can be defined in `VerticalNavItem` objects and `Step` objects, and they can also be defined at root levels to be inherited by sub levels.

### Changed
- Used new prop in react examples
- Updated Button example to be consistent
- Improved step object documentation, use a table


packages/core/src/components/StepList/_StepList.docs.scss
```diff
@@ -46,21 +46,100 @@ Style guide: patterns.step-list.react
...
-- **started** (`bool`) - Whether the step has been started
-- **isNextStep** (`bool`) - Whether this is the next unstarted step
-- **title** (`string`) - Text to display as the step title
-- **description** (`string`) - Additional text to display under (only rendered for top-level steps)
-- **linkText** (`string`) - Alternative text for the link or button for this step. Will override the defaults
-- **onClick** (`func`) - onClick handler for this specific step's link/button
-- **steps** (`Step[]`) - Array of substeps
-
+<table class="ds-c-table">
+  <thead>
```
bernardwang: Because of the way <StepList> is setup, we cannot simply use `<Step>` to show the step object documentation. In order to be consistent with other prop documentation, I converted the bulleted list to a table. Unfortunately, this needs to be updated manually and is not automatically generated from React prop documentation

packages/core/src/components/StepList/StepList.jsx
```diff
@@ -5,7 +5,11 @@ import Step from './Step';
 export const StepList = ({ steps, ...props }) => (
   <ol className="ds-c-step-list ds-u-margin-top--4">
     {steps.map((step, i) => (
-      <Step step={step} key={step.id || i} {...props} />
+      <Step
```
bernardwang: I defined 2 ways to use `component` in `<StepList>` and `VerticalNav`. It can either be set on a granular level within each specific `step` or `item` object, or it can be set on higher levels of `step` or `item`. 

I.e., you can set `component` once on the root `VerticalNav` component in order to make every single nav item to use the passed in component, OR you can set it on a single child if you only want one nav item to have a custom component. 

I'm thinking for the react-router example, it wont be necessary to have such granular control. Should I just keep it simple and have 1 `component` prop on the root element?

packages/core/src/components/VerticalNav/VerticalNavItemLabel.jsx
```diff
@@ -53,10 +55,11 @@ export class VerticalNavItemLabel extends React.PureComponent {
       onClick: this.props.onClick ? this.handleClick : undefined
     };
 
-    if (this.LabelComponent === 'a') {
-      props = Object.assign(props, this.anchorProps());
-    } else if (this.LabelComponent === 'button') {
+    if (this.LabelComponent === 'button') {
       props = Object.assign(props, this.buttonProps());
+    } else if (this.LabelComponent !== 'div') {
```
pwolfert: At first I was confused by why we were singling out `div`s, but now I see that that's the default return value for the `componentType` function. Would it improve readability to do something like this?
```js
const DEFAULT_LABEL_COMPONENT_TYPE = 'div';
...
return DEFAULT_LABEL_COMPONENT_TYPE;
...
else if (this.LabelComponent !== DEFAULT_LABEL_COMPONENT_TYPE)
```
or `FALLBACK_LABEL_COMPONENT_TYPE` or some other name that makes sense?

packages/core/src/components/StepList/_StepList.docs.scss
pwolfert: Looks nice 👍

packages/core/src/components/VerticalNav/VerticalNav.example.jsx
```diff
@@ -1,14 +1,23 @@
+import PropTypes from 'prop-types';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import VerticalNav from './VerticalNav';
 
+// Mock react-router example
+const Link = props => (
+  // <Link to={props.href} {...props}>{props.children}</Link>
+  <span {...props}>{props.children}</span>
```
pwolfert: I feel like by having this example we're almost condoning unsemantic interactive elements. What if we just did something like this:
```js
<a {...props}><span className="really-special-link">{props.children}</span></a>
```

packages/core/src/components/index.js
```diff
@@ -18,6 +18,9 @@ export * from './Review/Review';
 export * from './SkipNav/SkipNav';
 export * from './Spinner/Spinner';
 export * from './StepList/StepList';
+export * from './StepList/Step';
+export * from './StepList/StepLink';
+export * from './StepList/SubStep';
```
pwolfert: Why are we exporting these internal components now?

packages/core/src/components/VerticalNav/VerticalNav.example.jsx
bernardwang: Ah good point, it does look like that. I chose span arbitrarily so that you could differentiate the elements using `component` in the example, but a class will do the trick!

packages/core/src/components/index.js
bernardwang: I saw that `VerticalNav` was also exporting `VerticalNavItem`, `VerticalNavItemLabel` so I assumed it was intended to expose these ones too. Happy to revert this change, I'm just not super clear on whats considered internal.

pwolfert: I don't know what the right choice is here exactly, but my thinking was that if the component doesn't have a purpose outside of its parent and there's no way to build our own parent from the children anyway (like `<StepList><StepItem /></StepList>`, then there's no point in exposing it or documenting it.

pwolfert: And if we expose it, we should document it.

bernardwang: That makes sense to me,

> but my thinking was that if the component doesn't have a purpose outside of its parent and there's no way to build our own parent from the children anyway

That makes sense to me. Looking at both `VerticalNav` it looks like we shouldn't be exposing those either but I can take care of that in a different PR

PR #610, opened 2020-01-23
Add hidden print utility class

hannaliebl:
### Added
Added `ds-u-visibility--hidden-print` class that hides content during printing using `media @print` query.


line47:
@bernardwang can you take a look when you have a moment 👍

PR #612, opened 2020-01-24
[WNMGDS-358] update lodash.template

line47:
Fix lodash.template security issue by updating lodash.template to 4.5.0

- Needed to remove `gulp-util` as there were nested dependencies and[ it was deprecated](https://www.npmjs.com/package/gulp-util). 
- Then needed to replace what gulp-util was being used for with `ansi-colors`, `fancy-log` and `replace-ext`
- Also needed to update `gulp-sass` as it had a dependency on `gulp-util` as well

There is still a `gulp-util` dependency on our version of gulp which is `^3.9.1` that when I try and update to `^4.0.0` i am no longer able to build with `yarn build` and get errors.


tools/gulp/common/log-util.js
```diff
@@ -3,62 +3,12 @@
...
 const pkg = require('../../../package.json');
-const gutil = require('gulp-util');
-const chalk = gutil.colors;
+const log = require('fancy-log');
+const c = require('ansi-colors');
 const notifier = require('node-notifier');
 
 const shellPrefix = '$';
 
-function drawFlag() {
```
line47: Removes this ascii USA flag from the log as we are not longer using USWDS

line47:
`yarn build` and `yarn test` both work and do not report any new issues. :) Not sure if we want to incrementally merge this and then work on the `gulp` issue or try and roll that work in here. 

☝️ @bernardwang

line47:
Our version of gulp which is `^3.9.1` has a child dependency of `gulp-util "^3.0.0"` and `gulp-util` has a `lodash.template` dependency. So we need to update our version of gulp to remove gulp-util to then remove lodash.template

line47:
Closing this PR to focus on upgrading babel/jest and webpack instead

PR #615, opened 2020-01-31
Upgrade node/yarn

bernardwang:
### Summary
We were using different versions of node and yarn in team, this PR bumps yarn to the latest and node up to `11`. I was running into many issues with node version 12 and above, but I think this will work for now.


packages/core/src/components/Dialog/__snapshots__/Dialog.test.jsx.snap
```diff
@@ -149,7 +149,7 @@ exports[`Dialog renders react-aria-modal 1`] = `
           active={true}
           focusTrapOptions={
             Object {
-              "initialFocus": undefined,
+              "escapeDeactivates": true,
```
bernardwang: For whatever reason this snapshot changes with an upgraded node and it was bothering me for a while.

PR #620, opened 2020-02-13
[WNMGDS-356] Update browser support page

line47:
- Bolded the text about which versions of browsers are supported
- Added info about Progressive enhancement and Graceful degradation


packages/docs/src/pages/startup/browser-support.md
```diff
@@ -3,4 +3,16 @@ title: Browser support
 weight: 5
 ---
 
-The CMS design system is designed to support older and newer browsers through [progressive enhancement](https://en.wikipedia.org/wiki/Progressive_enhancement). The current major version of the design system (2.0) follows the [2% rule](https://gds.blog.gov.uk/2012/01/25/support-for-browsers/): we officially support any browser above 2% usage as observed by [analytics.usa.gov](https://analytics.usa.gov/). Currently, this means that the newest versions of Chrome, Firefox, Safari, and Internet Explorer 11 and up are supported.
+The CMS design system follows the [2% rule](https://gds.blog.gov.uk/2012/01/25/support-for-browsers/): we officially support any browser above 2% usage as observed by [analytics.usa.gov](https://analytics.usa.gov/). **Currently, this means that the newest versions of Chrome, Firefox, Safari, Internet Explorer 11, Edge, and Samsung Internet are supported.**
```
bernardwang: Do you think its necessary to include Samsung Internet? I know its above 2%, but its specific to Android so it seems weird to be categorized as a browser.

bernardwang: Also, it looks like we support above 2% and last 2 versions

packages/docs/src/pages/startup/browser-support.md
```diff
@@ -3,4 +3,16 @@ title: Browser support
...
+
+The CMSDS is designed to be compatible with different operating systems and browser versions in order to provide a broad range of support. Progressive enhancement and graceful degredation methodologies are used to accomplish this.
+
+## Progressive enhancement
```
bernardwang: Do we need to provide all this information if the request was mainly about the list of browsers supported? This info doesn't feel very relevant to the page imo (i.e. graceful degradation)

packages/docs/src/pages/startup/browser-support.md
line47: Samsung internet was listed on analytics.usa site https://analytics.usa.gov/health-human-services/ which was why I included it. I think it would be nice in the future to be able to utilize their API to pull in metrics to show the stats on our site. 

I'll update the text to include last 2 versions, thanks!

packages/docs/src/pages/startup/browser-support.md
line47: I'm good with that, I don't feel strongly about it. I was looking at the assets and styleguide sites and thinking about incorporating related info on the page. 

I'll update to remove it

packages/docs/src/pages/startup/browser-support.md
bernardwang: I understand it’s on the website, but I don’t think it makes much sense to include because it’s a mobile browser. Just like if “Safari (in-app)” was above 2% it isn’t  practical to include it, and it’s not like anyone is testing that specific mobile browser version. If you look at [browserlist](https://browserl.ist/?q=%3E2%25), samsung browser also falls under a separate category for mobile browsers.

bernardwang: I also wonder if we should also change our language around analytics.gov, because we actually use browserlist for prefixing. As far as I know, aside from browser list we just manually test, and we certainly don’t test according to analytics.gov.

line47: I think it would be ideal to use https://analytics.usa.gov/health-human-services/ as a lense for our browser support info. Browserlist is great but shows global support whereas analytics.gov gets us closer to our target audience.

bernardwang: Ok, we can leave it then! My concern was that we say we support it, but we don't actually use it in any meaningful way besides manual testing, and it would be more accurate to show browserlist, which is what we actually use. However, I'm guessing this the same thing USDWS does as well and it probably a gov website policy stance more than anything else.

line47: I wonder if there is a way we can use `my stats` to pull analytics into browserlist https://github.com/browserslist/browserslist-ga

bernardwang: it’s possible to dynamically get the list of browsers, but i’m doubtful it will be helpful to display on the page. it’s just a very long list of every browser and their version. if we really want to surface this information, i think we should just link to the website that displays all the prefixed browsers

bernardwang: what do you think about removing samsung internet? it’s the last thing for me in this PR

line47: Yeah I'll remove that and add a link to https://browserl.ist/?q=last+2+versions%2C+%3E2%25

packages/docs/src/pages/startup/browser-support.md
```diff
@@ -3,6 +3,6 @@ title: Browser support
 weight: 5
 ---
 
-The CMS design system follows the [2% rule](https://gds.blog.gov.uk/2012/01/25/support-for-browsers/): we officially support any browser above 2% usage and the last 2 browsers versions for each browser as observed by [analytics.usa.gov](https://analytics.usa.gov/). **Currently, this means that the newest versions of Chrome, Firefox, Safari, Internet Explorer 11, Edge, and Samsung Internet are supported.**
+The CMS design system follows the [2% rule](https://gds.blog.gov.uk/2012/01/25/support-for-browsers/): we officially support any browser above 2% usage and the last 2 browsers versions for each browser as observed by [analytics.usa.gov](https://analytics.usa.gov/). For specific browser details check out [browserl.ist](https://browserl.ist/?q=last+2+versions%2C+%3E2%25). **Currently, the latest 2 versions of Chrome, Firefox, Safari, Internet Explorer 11, and Edge, and Samsung Internet are supported.**
```
bernardwang: Looks like Samsung Internet is still here. And super small nitpick, but what do you think about changing the link from `browserl.ist` to `browserlist`?

line47:
Ready for a final pass @bernardwang

PR #622, opened 2020-02-14
[WNNMGDS-270] Update prop doc rendering

bernardwang:
### Changed
- Show `children` props in documentation
- Hide prop docs with `@HidePropDoc` in the description instead of `_`


packages/core/src/components/Button/Button.jsx
```diff
@@ -136,7 +139,7 @@ Button.propTypes = {
    */
   onClick: PropTypes.func,
   /**
-   * (Deprecated) Access a reference to the `button` or `a` element. Please use `inputRef` instead.
+   * @HidePropDoc [Deprecated] Access a reference to the `button` or `a` element. Please use `inputRef` instead.
```
line47: I don't think `@HidePropDoc` is working as expected. I saw this in the button props section.
Screen Shot 2020-02-17 at 9 48 26 AM

bernardwang:
@line47 feedback should be addressed

PR #623, opened 2020-02-14
[WNMGDS-231] Remove chromedriver temporary fix

bernardwang:
## Fixed
- Reverted temporary fix [here](https://github.com/CMSgov/design-system/pull/471)

### How to test
- Run `yarn test:e2e` locally to confirm the e2e tests are working correctly. You might need to update your Chrome for this to work.
- Make sure the travis tests are working

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,6 @@ before_install:
   - export PATH="$HOME/.yarn/bin:$PATH"
   - google-chrome-stable --version
   - chromedriver --version
-  - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 http://localhost &
 install:
   - yarn
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,10 @@
 dist: trusty
+addons:	
+  chrome: stable
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.21.1
   - export PATH="$HOME/.yarn/bin:$PATH"
-  - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 --no-sandbox --disable-dev-shm-usage http://localhost &
+  - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 http://localhost &
 install:
   - yarn
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ addons:
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.21.1
   - export PATH="$HOME/.yarn/bin:$PATH"
+  - google-chrome-stable --version
+  - chromedriver --version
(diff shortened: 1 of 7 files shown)


bernardwang:
@pwolfert I updated chrome driver and selenium and it seemed to fix the original issue, but i’m not super familiar with travis, so i’m not sure if this will work long term (i.e when the chrome stable version travis is updated). Would you be able to take a quick look and see if everything checks out?

tools/jest/e2e.environment.js
```diff
@@ -11,12 +12,12 @@ class WebDriverEnvironment extends NodeEnvironment {
   }
 
   async buildDriver() {
-    let driver = new Builder().forBrowser(this.browserName);
+    const driver = new Builder().forBrowser(this.browserName);
 
     if (this.chromeOptions) {
```
bernardwang: Setting chromeOptions with updated chromedriver methods

line47:
my `yarn e2e` tests worked fine, I'll need to look into how to run Travis tests locally

.travis.yml
```diff
@@ -1,8 +1,10 @@
 dist: trusty
+addons:	
+  chrome: stable
 before_install:
   - curl -o- -L https://yarnpkg.com/install.sh | bash -s -- --version 1.21.1
   - export PATH="$HOME/.yarn/bin:$PATH"
-  - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 --no-sandbox --disable-dev-shm-usage http://localhost &
+  - google-chrome-stable --headless --disable-gpu --remote-debugging-port=9222 http://localhost &
```
pwolfert: I'm not entirely sure that Selenium webdriver even uses this running instance of chrome. I actually wonder if the tests would run without this line

bernardwang: I can test removing that line! also what do you mean by their chrome plugin? would that let us specify a version of chrome?

pwolfert: `chrome-stable` is their plugin for providing chrome, and it's annoying because you can't specify an exact version. I put a link to it in that old PR that you linked to

bernardwang: Oh, are you referring to [`this addon`](https://github.com/CMSgov/design-system/pull/623/files/c818dc4b26ba7adde0703c2cfec4e0b4e9d0169a#diff-354f30a63fb0907d4ad57269548329e3R3)? I tried removing that and the tests seem to run fine, is that preferable?

pwolfert: Yeah, if that line isn't doing anything, let's get rid of it. It just confuses things.

pwolfert: Sorry, I just meant this line `google-chrome-stable --headless`, not removing the add-on. We need the add-on so that chrome is installed

pwolfert: Wait, sorry, you removed the addon and it still worked? If that's the case, don't take my word for it that it's necessary to keep around!

pwolfert: What a great mystery Travis CI is 🔮

bernardwang: No it doesn't work when I remove it, I just made a mistake haha

PR #626, opened 2020-02-19
[WNMGDS-317] Deprecate support & layout packages

bernardwang:


bernardwang:
@line47 These changes are major breaking changes, and I don't think we can just publish them regularly. I think for now we should merge these changes into a separate branch, and then think about how we want to roll out these major change (i.e. bumping `design-system-core` to 4.0.0 or making a new beta packages?)

line47:
I agree and think that is a great idea 👍

PR #627, opened 2020-02-21
[NO-TICKET] doc site update to usa flag path for header

line47:
I noticed when looking at the Medicare and Healthcare design system doc sites that the USA flag in the top banner is broken due to an incorrect reference.

This PR is to fix the broken USA flag in the USA banner for both medicare and healthcare doc sites 
https://github.cms.gov/pages/MedicareGov/ds-site-package/
https://github.cms.gov/pages/CMS-WDS/ds-healthcare-gov/

The path to the USA flag is currently
https://github.cms.gov/public/images/us_flag_small.png (which 404s)

and after the update, the path works as expected. 
https://github.cms.gov/pages/CMS-WDS/ds-healthcare-gov/public/images/us_flag_small.png

### DIFF
--- a/packages/docs/src/scripts/components/UsaBanner.jsx
+++ b/packages/docs/src/scripts/components/UsaBanner.jsx
@@ -12,7 +12,7 @@ const UsaBanner = function(props) {
         <div className="ds-u-display--flex ds-u-flex-direction--row ds-u-align-items--start ds-u-sm-align-items--center">
           <img
             className="c-usa-banner__header-flag"
-            src="../../public/images/us_flag_small.png"
+            src="./public/images/us_flag_small.png"
             alt="U.S. flag"
           />
           <p className="c-usa-banner__header-text">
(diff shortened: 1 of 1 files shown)


packages/docs/src/scripts/components/UsaBanner.jsx
```diff
@@ -12,7 +12,7 @@ const UsaBanner = function(props) {
         <div className="ds-u-display--flex ds-u-flex-direction--row ds-u-align-items--start ds-u-sm-align-items--center">
           <img
             className="c-usa-banner__header-flag"
-            src="/public/images/us_flag_small.png"
+            src="./public/images/us_flag_small.png"
```
bernardwang: oh wow thats all it took? thats great! i assumed it was some sort of build process issue

line47: Yeah I think so, I was doing some testing in the browser but don't think any of the build would muck up the path

packages/docs/src/scripts/components/UsaBanner.jsx
```diff
@@ -12,7 +12,7 @@ const UsaBanner = function(props) {
...
             alt="U.S. flag"
           />
           <p className="c-usa-banner__header-text">
```
bernardwang: I think you might need to add the dot to relative paths on the images on line 40 and 52 as well

line47:
Updated and ready for a final pass @bernardwang

PR #633, opened 2020-02-25
[WNMGDS-286] Remove deprecated items

bernardwang:
### Removed
- `<Select>`
- "Right to left" choices


bernardwang:
@line47 We could possibly also include in this PR removing other deprecated props like `fieldRef`. These have been marked as officially deprecated for a few months now as well. A third option is to merge these deprecations with the `bernard/new-cms` branch to include it with the v2. What do you think?

line47:
I'm thinking that adding them to the `bernard/new-cms` branch might be best

PR #638, opened 2020-03-02
[WNMGDS-384] Allow custom heading level

bernardwang:
This PR makes `heading` and `headingLevel` props more consistent across the design system. Prior to this PR, only some components used `headingLevel`, and some components used the prop name `title` instead of `heading`. 

### Added
- `headingLevel` prop to `HelpDrawer`, `StepList`, `Alert`, `Review`, 

### Changed
- Renamed `title` prop to `heading` in `StepList`, `HelpDrawer`, `Dialog`
   - Added deprecation console warnings and updated documentation

### DIFF
--- a/packages/core/src/components/HelpDrawer/HelpDrawer.jsx
+++ b/packages/core/src/components/HelpDrawer/HelpDrawer.jsx
@@ -22,6 +22,8 @@ export class HelpDrawer extends React.PureComponent {
       footerBody,
       footerTitle
     } = this.props;
+    const Heading = `h${this.props.headingLevel}` || `h3`;
+
     /* eslint-disable jsx-a11y/no-noninteractive-tabindex, react/no-danger */
     return (
       <div className="ds-c-help-drawer">
@@ -31,13 +33,13 @@ export class HelpDrawer extends React.PureComponent {
            * so things display as expected when the body content overflows
            */}
           <div className="ds-u-fill--gray-lightest ds-u-padding--2 ds-u-display--flex ds-u-align-items--start">
-            <h3
+            <Heading
               ref={el => (this.titleRef = el)}
               tabIndex="0"
               className="ds-u-text--lead ds-u-margin-y--0 ds-u-margin-right--2"
             >
               {title}
-            </h3>
+            </Heading>
             <Button
               aria-label={ariaLabel}
               className="ds-u-margin-left--auto ds-c-help-drawer__close-button"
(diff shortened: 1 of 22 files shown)


packages/core/src/components/StepList/_StepList.docs.scss
```diff
@@ -76,6 +76,25 @@ Step object
         <p>Additional text to display under (only rendered for top-level steps)</p>
       </td>
     </tr>
+    <tr>
```
bernardwang: It would be really nice to not have to manually update `StepList` documentation. In the future we should consider reorganizing `StepList` to be more like `VerticalNav`, which is able to neatly expose `VerticalNav` item object props because of the way `VerticalNavItem` is built. This will be a breaking change, but it would be nice to not only fix the documentation, but add some consistency to how we build these more complex, "nested" components. 

https://jira.cms.gov/browse/WNMGDS-386

packages/core/src/components/StepList/StepList.jsx
```diff
@@ -20,7 +20,9 @@ export const StepList = ({ steps, ...props }) => (
 export const stepShape = {
   id: PropTypes.string,
   href: PropTypes.string.isRequired,
-  title: PropTypes.string.isRequired,
+  title: PropTypes.string.isRequired, // [Deprecated]
+  heading: PropTypes.string.isRequired,
+  headingLevel: PropTypes.oneOf(['1', '2', '3', '4', '5']),
```
line47: `title` should be replaced with `heading` for the following prop descriptions
- steps
- actionsLabelText
- descriptionLabelText
- substepsLabelText

packages/core/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -64,18 +81,31 @@ export class HelpDrawer extends React.PureComponent {
...
+  /**
+   * Helps give more context to screen readers on the button that closes the Help Drawer
+   */
   ariaLabel: PropTypes.string,
   closeButtonText: PropTypes.string,
   children: PropTypes.node.isRequired,
   footerBody: PropTypes.node,
   footerTitle: PropTypes.string,
+  /**
+   * Text for the HelpDrawer title. Required because the `heading` will be focused on mount.
```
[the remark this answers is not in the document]
bernardwang: I chose to leave this one because it seemed less helpful for the description for `heading` to be "text for the `heading`".

packages/core/src/components/StepList/StepList.jsx
bernardwang: I cant remove the `title` in steps if its to be backwards compatible.

PR #639, opened 2020-03-02
[WNMGDS-385] - Update dialog to focus on window 

### LINKED ISSUE #631: Modal Dialog - allow for flexibility of tabindex - a11y issue
**Description**

Would like to add tabindex=-1 to container of the modal dialog component. 

Currently, the modal window is only reading out the heading level one, and the Add generic button. This is probably due to the button being focused when the modal is opened. Users with visual issues may not understand why they're being prompted to add a generic without the entire modal being read aloud.

**Possible Remediation:**
Setting focus on the modal window's container instead of the button should cause the entire modal to be read aloud. Even if some screen readers do not honor this, users will be able to cursor through the modal and hear all the relevant information.

- Add a tabindex="-1" to the container so it can take keyboard focus. 
- Focus should still be trapped inside the modal once users press TAB

`<div
  aria-labelledby="dialog-title"
  class="ds-c-dialog ds-base"
  id="react-aria-modal-dialog"
  role="dialog"
  tabindex="-1">`

**Additional information** 
- related ticket [https://jira.cms.gov/browse/MCT-1825](https://jira.cms.gov/browse/MCT-1825)

**WCAG Guidance:** [Keyboard: Understanding SC 2.1.1](https://www.w3.org/TR/UNDERSTANDING-WCAG20/keyboard-operation-keyboard-operable.html)

image

line47:
- Added `focusDialog` to always set focus on the entire dialog window 
- Added `dialog-content ` id to the `<main>` section of the dialog
- Added `dialog-conent` to titleID `titleId="dialog-title dialog-content"`

closes #631

### DIFF
--- a/packages/core/src/components/Dialog/__snapshots__/Dialog.test.jsx.snap
+++ b/packages/core/src/components/Dialog/__snapshots__/Dialog.test.jsx.snap
@@ -10,9 +10,10 @@ Object {
   "wrapper": <Displaced
     dialogClass="ds-c-dialog ds-base"
     escapeExits={true}
+    focusDialog={true}
     includeDefaultStyles={false}
     mounted={true}
-    titleId="dialog-title"
+    titleId="dialog-title dialog-content"
     underlayClass="ds-c-dialog-wrap"
     underlayClickExits={false}
   >
@@ -33,6 +34,7 @@ Object {
         </Button>
       </header>
       <main
+        id="dialog-content"
         role="main"
       >
         Foo
@@ -62,6 +64,7 @@ exports[`Dialog renders react-aria-modal 1`] = `
   <Displaced
     dialogClass="ds-c-dialog ds-base"
     escapeExits={true}
+    focusDialog={true}
     getApplicationNode={
       [MockFunction] {
         "calls": Array [
@@ -72,7 +75,7 @@ exports[`Dialog renders react-aria-modal 1`] = `
     includeDefaultStyles={false}
     mounted={true}
     onExit={[MockFunction]}
-    titleId="dialog-title"
+    titleId="dialog-title dialog-content"
     underlayClass="ds-c-dialog-wrap"
     underlayClickExits={false}
   >
@@ -84,10 +87,11 @@ exports[`Dialog renders react-aria-modal 1`] = `
               class="ds-c-dialog-wrap"
             >
               <div
-                aria-labelledby="dialog-title"
+                aria-labelledby="dialog-title dialog-content"
                 class="ds-c-dialog ds-base"
                 id="react-aria-modal-dialog"
                 role="dialog"
+                tabindex="-1"
               >
                 <div
                   role="document"
@@ -111,6 +115,7 @@ exports[`Dialog renders react-aria-modal 1`] = `
                     </button>
                   </header>
                   <main
+                    id="dialog-content"
                     role="main"
                   >
                     Bar
@@ -126,6 +131,7 @@ exports[`Dialog renders react-aria-modal 1`] = `
         dialogClass="ds-c-dialog ds-base"
         dialogId="react-aria-modal-dialog"
         escapeExits={true}
+        focusDialog={true}
         focusTrapPaused={false}
         getApplicationNode={
           [MockFunction] {
@@ -138,7 +144,7 @@ exports[`Dialog renders react-aria-modal 1`] = `
         mounted={true}
         onExit={[MockFunction]}
         scrollDisabled={true}
-        titleId="dialog-title"
+        titleId="dialog-title dialog-content"
         underlayClass="ds-c-dialog-wrap"
         underlayClickExits={false}
         underlayColor="rgba(0,0,0,0.5)"
@@ -149,7 +155,7 @@ exports[`Dialog renders react-aria-modal 1`] = `
           active={true}
           focusTrapOptions={
             Object {
-              "initialFocus": undefined,
+              "initialFocus": "#react-aria-modal-dialog",
             }
           }
           paused={false}
@@ -161,12 +167,13 @@ exports[`Dialog renders react-aria-modal 1`] = `
               style={Object {}}
             >
               <div
-                aria-labelledby="dialog-title"
+                aria-labelledby="dialog-title dialog-content"
                 className="ds-c-dialog ds-base"
                 id="react-aria-modal-dialog"
                 key="b"
                 role="dialog"
                 style={Object {}}
+                tabIndex="-1"
               >
                 <div
                   role="document"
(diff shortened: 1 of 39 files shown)


line47:
@davidakennedy Could you take a look here and see if it is working as expected?

packages/core/src/components/Dialog/__snapshots__/Dialog.test.jsx.snap
```diff
@@ -10,9 +10,10 @@ Object {
   "wrapper": <Displaced
     dialogClass="ds-c-dialog ds-base"
     escapeExits={true}
+    focusDialog={true}
     includeDefaultStyles={false}
     mounted={true}
-    titleId="dialog-title"
+    titleId="dialog-title dialog-content"
```
davidakennedy: In my testing, I found that the `dialog-content` ID needed to be on an element without children, otherwise screen readers aren't what or how much to announce. I tested with VoiceOver, and still need to test with JAWS and NVDA, but it seems like a good rule to follow.

So instead of being on the `main` element, it would be on a paragraph inside the `main`, as an example.

That's why I was wondering if it's better to create a separate example with this type of setup?

line47: Oh gotcha - what would that look like for the example provided in #631  where there are 2 paragraphs in the `main` element?

bernardwang: @line47 sorry, but what is the purpose of the `dialog-content` again? is that needed to work with `focusDialog`?

bernardwang: @line47 @davidakennedy If we are using `focusDialog`, would it still be necessary to also change the `titleId` to include both the title and content?

line47: Adding to the `titleId ` was to get screen readers to speak the title and content. But the content is variable so maybe it makes more sense to leave it out and add some guidance on how to add it in if desired. 

> Also, you can pass an additional ID via the titleId prop, that way screen readers will associate two labels with it, and usually announce both.

davidakennedy: > what would that look like for the example provided in #631 where there are 2 paragraphs in the main element?

@line47 In my testing, it would be best to put a wrapper div around the elements in the main container that need to be announced. All three major screen reader/browser combinations announced it that way. When the `dialog-class` is on the `main` element, only JAWS with IE 11 announces it right. Not exactly sure why that is the case.

Should we make a separate example to demo this use case?

line47: Making sure I'm understanding here... so folks would need to customize the dialog component to get additional screen reader announcements based on usage of the `titleId` prop and what works best for their scenario?

davidakennedy: Right. We'd could advise them what to do based on the content of their modal, and what they were after.

bernardwang:
I also noticed something, it seems our existing dialog example is broken somehow. Notice how the focus trap isnt working
https://design.cms.gov/example/components.dialog.size/

line47:
The react version is working as expected https://design.cms.gov/example/components.dialog.react/ this appears to be for the plain HTML versions only

packages/core/src/components/Dialog/__snapshots__/Dialog.test.jsx.snap
bernardwang: Ah I see, its a little odd that we need to use `titleId` to also read the content.  I thought normally if you focus a container it will read out everything inside, but since `react-aria-modal` is doing something with `titleId` it might be overriding the default browser behavior maybe. 

Regardless, can we just wrap the `children` with another div with the `dialog-content` className? Something like this:
```
<main role="main">
        <div id="dialog-content">
          {children}
        </div>
</main>
```
It adds an extra level of nesting but it would make it easier on the user

line47: I was thinking that example would work if there was only one paragraph in the `{children}` section but if there was more than one paragraph of content addition customization would be needed.

line47: So would there be a difference between these 2 code blocks?
 I was thinking these are essentially the same where `id="dialog-content"` is outside of children. and what we Ideally want is to put an id on each child element with a unique id.
```
<main role="main">
        <div id="dialog-content">
          {children}
        </div>
</main>
```
```
<main role="main" id="dialog-content">
          {children}
</main>
```

bernardwang: > In my testing, it would be best to put a wrapper div around the elements in the main container that need to be announced. All three major screen reader/browser combinations announced it that way

I just suggested the additional `div` because of what @davidakennedy said. My understanding is that if theres a wrapper with `dialog-content` it wouldn't matter if its 1 or 2 paragraphs.

line47: Ah gotcha, I missed that nuance in the earlier thread. I definitely think adding the extra div would be beneficial to set folks up for success
```
<main role="main">
        <div id="dialog-content">
          {children}
        </div>
</main>
```

line47: @davidakennedy could you give this another pass?

davidakennedy: @line47 I tested some more and it's still not working well.

For some reason, both VoiceOver and NVDA don't announce the content in the `main` element, as the second element referenced by `aria-labelledby`. To be honest, not sure why that is, probably just quirks between browsers and screen readers. The way I found that it works across the board is to have it not on a `main`, but another element.

Also, if we go this route,  we could either hide the outline in some browsers on the dialog div or have it be more visually appealing.

davidakennedy: @line47 Retesting this after your ping in Slack. This works with all major combinations nicely:
- IE 11/JAWS
- Edgium/JAWS
- Chrome/JAWS
- NVDA/Firefox
- VoiceOver/Safari

Still wondering about this from my last comment:
> Also, if we go this route, we could either hide the outline in some browsers on the dialog div or have it be more visually appealing.

line47:
@bernardwang and @davidakennedy do you all think we should address the focus styling in this PR of the modal? 

See the screenshot of the focus which is now on the entire modal window
Screen Shot 2020-03-05 at 12 48 37 PM


I'm inclined to put a new ticket in for that work.

bernardwang:
I'm fine with a new ticket. We never added a global focus style for `tabindex`ed elements, but I'm ok with adding one for the modal

line47:
OK awesome, thanks!

PR #641, opened 2020-03-05
[WNMGDS-391] Fix MonthPicker legend

bernardwang:
### Changed
- Added keypress event for  link `<Button>`, i.e.  links styled as buttons will have `onClick` triggered on space
- Move label and hint into `legend` for `<MonthPicker>`. This should fix the label being read by the screen reader twice and should make the markup more consistent with other form elements.
- Deprecate `inverse` prop in `<Button>`, it is renamed to `inversed` to be consistent with all other components
- Made prop types consistent for the `component` prop

### Removed
- `heading` and `headingLevel` props have been removed from `<MonthPicker>` as they are no longer necessary.

### How to test
- Insure `<Button>` behavior is the same, and that link `<Button>`s can be clicked with a space while being focused
- Test `<MonthPicker>` and insure it works well on screenreaders


packages/core/src/components/Button/Button.example.jsx
```diff
@@ -22,10 +23,14 @@ ReactDOM.render(
...
-    <Button className="ds-u-margin-right--1" variation="transparent" href="javascript:void(0);">
+    <Button
+      className="ds-u-margin-right--1"
+      href="javascript:void(0);"
+      onClick={() => alert('Link styled as button clicked!')}
+    >
       Button with `href` prop
     </Button>
-    <Button className="ds-u-margin-right--1" component={Link}>
+    <Button className="ds-u-margin-right--1" component={Link} href="/">
```
line47: I'm wondering if we could leave the `<button>` as a `<button>` here instead of switching it to a link. Generally, I think we should use links for navigating and not triggering functionality on the page.

bernardwang: Its the only way to keep the buttons in the legend and keep the existing design
https://adhoc.slack.com/archives/C645GLLNN/p1583190102065800?thread_ts=1582927324.039200&cid=C645GLLNN

line47: I was thinking that since the buttons are not inside the legend that we could do something like what Trevor suggested from that thread: 

> You might try fieldset, legend, div with buttons, div with checkboxes + labels, and use aria-describedby to reference the legend on each checkbox.

bernardwang: I see what you mean, the last commit implements that

davidakennedy:
I've done a good amount of testing on this. I like this structure better, but am getting some inconsistencies around how screen readers are announcing the labels, hints and buttons. Mostly around the labels and hints though.

Some notes on results:

## VoiceOver/Safari

- Navigating through checkboxes with tab key announces label and hint.
- Navigating through the checkboxes with form controls menu announces the hint once, and nothing if no hint is there.
- Navigating through the buttons with the tab key announces the label and hint twice.
- Navigating through the buttons with form controls menu announces the label and hint once.
- Links in the labels/hints are not announced.

## JAWS/IE 11

- Navigating through checkboxes with tab key announces label and hint.
- Navigating through the checkboxes with next form controls shortcut announces the label and hint.
- Navigating through the buttons with the tab key announces the label and hint.
- Navigating through the buttons with the next form controls shortcut announces the label and hint.
- Links in the labels/hints are not announced.

## JAWS/Edgium

- Navigating through checkboxes with tab key announces the hint, but not the label.
- Navigating through the checkboxes with next form controls shortcut announces the hint, but not the label.
- Navigating through the buttons with the tab key announces the label and hint.
- Navigating through the buttons with the next form controls shortcut announces the label and hint.
- Links in the labels/hints are not announced.

## NVDA/Firefox

- Navigating through checkboxes with tab key announces the hint, but not the label.
- Navigating through the checkboxes with next form controls shortcut announces the hint, but not the label.
- Navigating through the buttons with the tab key announces the label and hint. Sometimes, it announces the label twice if text length is shorter.
- Navigating through the buttons with the next form controls shortcut announces the label and hint. Sometimes, it announces the label twice if text length is shorter.
- Links in the labels/hints are not announced.

## VoiceOver/Safari on iOS

- Navigating through checkboxes with touch announces the hint, but not the label.
- Navigating through the checkboxes with rotor controls  announces the hint, but not the label.
- Navigating through the buttons with touch announces the label and hint.
- Navigating through the buttons with rotor controls announces the label and hint.
- Links in the labels/hints are not announced.

## Talk Back/Chrome on Android

- Navigating through checkboxes with touch announces the hint, but not the label.
- Navigating through the checkboxes with the context menu controls  announces the hint, but not the label.
- Navigating through the buttons with touch announces the label and hint.
- Navigating through the buttons with the context menu announces the label and hint.
- Links in the labels/hints are not announced.

I'm wondering if taking away some of the wrapper divs might help? I need to experiment and see.

bernardwang:
@davidakennedy I'm curious to know how this compares to the default ChoiceList behavior for checkboxes. Currently the only thing that differentiates this `<MonthPicker>` implementation from the default is that there's extra buttons, and that we use a `aria-describedby` to reference the legend in each checkbox. 

Considering all the inconsistencies you are seeing are around the label and hint, maybe we should try removing the `aria-describedby`.

bernardwang:
@davidakennedy Actually yeah, thinking on this more, I don't think theres any reason for the checkboxes to need an extra  `aria-describedby` now that we moved the hint back into the `legend`. Maybe I can remove it and we can see if the inconsistencies are still there? (Or if the inconsistencies are an existing problem that we should handle in a different PR)

davidakennedy:
> I'm curious to know how this compares to the default ChoiceList behavior for checkboxes.

Yeah, taking a look to see... I do think we may be able to remove the `aria-describedby`. Going to test a bit more.

bernardwang:
@davidakennedy I updated the PR to remove aria described by, it seems to be working well for me, are you able to take a final pass at this?

davidakennedy:
For a detailed readout of how my testing went, see below. Overall, I think we need to make just one more tweak to this: Removing the `aria-describedby` on the buttons. Right now, they're announced twice in some screen readers. With the buttons in the fieldset, they're announced twice in some cases. So removing the attribute eliminates that.

I think the present solution, while reducing some screen reader affordances, also makes it less verbose in cases. So it's a win for me. Make that change and let's merge it.

## Testing results

**Some notes on results for the current month picker:**

## VoiceOver/Safari
- Navigating through checkboxes with tab key announces label twice and hint once.
- Navigating through the checkboxes with form controls menu announces the hint once, and nothing if no hint is there.
- Navigating through the buttons with the tab key announces the label.
- Navigating through the buttons with form controls menu announces the label.
- Links in the labels/hints are not announced.

## JAWS/IE 11
- Navigating through checkboxes with tab key announces label and hint.
- Navigating through the checkboxes with next form controls shortcut announces the label and hint.
- Navigating through the buttons with the tab key announces the label.
- Navigating through the buttons with the next form controls shortcut announces the label.
- Links in the labels/hints are not announced.

## NVDA/Firefox
- Navigating through checkboxes with tab key announces the hint, but not the label.
- Navigating through the checkboxes with next form controls shortcut announces the hint, and label.
- Navigating through the buttons with the tab key announces the label.
- Navigating through the buttons with the next form controls shortcut announces the label.
- Links in the labels/hints are not announced.

**Some notes on results for the updated month picker PR:**

## VoiceOver/Safari
- Navigating through checkboxes with tab key announces label and hint.
- Navigating through the checkboxes with form controls menu announces nothing (normal behavior, after testing other fieldsets and legends).
- Navigating through the buttons with the tab key announces the label and hint twice.
- Navigating through the buttons with form controls menu announces the label and hint.
- Links in the labels/hints are not announced.

## JAWS/IE 11
- Navigating through checkboxes with tab key announces label and hint.
- Navigating through the checkboxes with next form controls shortcut announces the label and hint.
- Navigating through the buttons with the tab key announces the label and hint.
- Navigating through the buttons with the next form controls shortcut announces the label and hint.
- Links in the labels/hints are not announced.

## NVDA/Firefox
- Navigating through checkboxes with tab key announces nothing (normal behavior, after testing other fieldsets and legends; it only announces when entering or exiting the fieldset).
- Navigating through the checkboxes with next form controls shortcut announces nothing (normal behavior, after testing other fieldsets and legends).
- Navigating through the buttons with the tab key announces the hint and label.
- Navigating through the buttons with the next form controls shortcut announces the label and hint.
- Links in the labels/hints are not announced.

bernardwang:
@davidakennedy Thanks again, I really appreciate your thoroughness! the most recent commit removes that `aria-describedby`

PR #643, opened 2020-03-05
[WNMGDS-376] - Add accessibility page to doc site 

line47:
**Note:** 
- Add a page to the doc site explaining how we approach and test for accessibility.


packages/docs/src/pages/guidelines/accessibility.md
```diff
@@ -0,0 +1,35 @@
+---
+title: Accessibility
+---
+
+CMS strives to make its Design System work for as many people as possible, including people with disabilities.
+
+## What we do to help meet and exceed standards
+
+- build accessibility into our [guiding principles](https://github.com/CMSgov/design-system/blob/master/guides/GUIDING-PRINCIPLES.md).
+- include accessibility as part of our design process when working on new or existing components.
```
bernardwang: maybe this bullet and the one below can be combined? both are talking about new components

packages/docs/src/pages/guidelines/accessibility.md
```diff
@@ -0,0 +1,35 @@
...
+- provide accessibility as a [checkpoint in new components](https://github.com/CMSgov/design-system/blob/master/.github/ISSUE_TEMPLATE/component-validation-checklist.md) created in the design system.
+- add accessibility as a key aspect to much of the [component documentation](https://github.com/CMSgov/design-system/blob/master/guides/WRITING-DOCUMENTATION.md).
+- conduct regular accessibility reviews of new and existing features and documentation for the design system.
+
```
bernardwang: We should probably also add our e2e testing that uses axe for automated a11y testing. Also locally we use that browser bookmarklet for a11y as well

line47: Hey @bernardwang which browser bookmarklet are you referring to?

packages/docs/src/pages/guidelines/accessibility.md
line47: Maybe updating "include accessibility as part of our design process when working on new or existing components." to something like "include accessibility as part of our design process." would still give enough details and also allow us to keep both bullet points?

bernardwang: Maybe... I feel like this bullet doesn't really say anything specific. And  honestly, accessibility should be included in design/product/engineering processes as well. The way USWDS handles this is by linking to [Accessibility for teams](https://accessibility.digital.gov/) which has some details and guidance for different roles.

packages/docs/src/pages/guidelines/accessibility.md
bernardwang: This one: https://khan.github.io/tota11y/
Screen Shot 2020-03-09 at 10 23 12 AM

Although in retrospect, I don't think we should mention it. It was legacy tool (we should consider removing it even), and `axe-webdriver` in e2e testing and `axe` browser extension for web testing are much more comprehensive.
Maybe we could even link to Ad Hoc's a11y tool guidance for more info here? https://adhocteam.us/topics/accessibility/index.html

bernardwang:
Sorry for all the content feedback! I should have posted this on the google doc first!

packages/docs/src/pages/guidelines/accessibility.md
```diff
@@ -0,0 +1,35 @@
...
+title: Accessibility
+---
+
+CMS strives to make its Design System work for as many people as possible, including people with disabilities.
+
+## What we do to help meet and exceed standards
+
+- build accessibility into our [guiding principles](https://github.com/CMSgov/design-system/blob/master/guides/GUIDING-PRINCIPLES.md).
+- include accessibility as part of our design process when working on new or existing components.
+- provide accessibility as a [checkpoint in new components](https://github.com/CMSgov/design-system/blob/master/.github/ISSUE_TEMPLATE/component-validation-checklist.md) created in the design system.
```
bernardwang: Ahh, I'm conflicted on this. Accessibility is included in that issue template, but we never incorporate that into our process. Issue templates are only used by external people on github, and we've never seen anyone file an issue with that template, or with code introducing a new component. The issue template almost feels like false advertising, or a process that is technically true, but in practice meaningless.

packages/docs/src/pages/guidelines/accessibility.md
line47: I like the idea of linking out to the accessibility for teams site - I'll include that!
Maybe we can remove the 2nd bullet point altogether then?

packages/docs/src/pages/guidelines/accessibility.md
line47: I think @davidakennedy had some plans for this accessibility checklist unless I'm mistaken?

packages/docs/src/pages/guidelines/accessibility.md
line47: I'm adding in a line about using "automated accessibility testing using [axe-webdriverjs](https://github.com/dequelabs/axe-webdriverjs)" but don't think we should link out to the Ad Hoc a11y page

packages/docs/src/pages/guidelines/accessibility.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Accessibility
+---
+
+CMS strives to make its Design System work for as many people as possible, including people with disabilities.
+
+## What we do to help meet and exceed standards
+
+- build accessibility into our [guiding principles](https://github.com/CMSgov/design-system/blob/master/guides/GUIDING-PRINCIPLES.md).
```
bernardwang: oops I thought I posted this already, but i guess not

The guiding principles page only mentions accessibility once, and it feels like a really outdated guide. It has very specific info on class naming and principles behind css utilities, but nothing on a more general product principles like accessibility. I wonder if we should remove this bullet for now and work on writing a new guiding principles to include in our doc site or in the GH wiki

packages/docs/src/pages/guidelines/accessibility.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Accessibility
+---
+
+CMS strives to make its Design System work for as many people as possible, including people with disabilities.
+
+## What we do to help meet and exceed standards
```
bernardwang: Could we capitalize the bullet points?

packages/docs/src/pages/guidelines/accessibility.md
```diff
@@ -0,0 +1,36 @@
...
+
+- build accessibility into our [guiding principles](https://github.com/CMSgov/design-system/blob/master/guides/GUIDING-PRINCIPLES.md).
+- provide accessibility as a [checkpoint in new components](https://github.com/CMSgov/design-system/blob/master/.github/ISSUE_TEMPLATE/component-validation-checklist.md) created in the design system.
+- add accessibility as a key aspect to much of the [component documentation](https://github.com/CMSgov/design-system/blob/master/guides/WRITING-DOCUMENTATION.md).
+- conduct regular accessibility reviews of new and existing features and documentation for the design system.
+- Embed [accessibility and inclusive design practices](https://accessibility.digital.gov/) into our workflow.
```
bernardwang: Could we move this to be first or second bullet point?

bernardwang:
@line47 sorry again for all the changes, but maybe it would be better for me to give my feedback on the google doc?

packages/docs/src/pages/guidelines/accessibility.md
line47: I think that works

bernardwang:
@line47 I think the content in [the doc](https://docs.google.com/document/d/1mTizaepejESHUiyOStSqX5W1f5kHviYhp3ruhqL93xQ/edit#) is more finalized, would you be able to update with this PR?

line47:
I think this is ready for another pass @bernardwang and @davidakennedy 👍

packages/docs/src/pages/guidelines/accessibility.md
```diff
@@ -0,0 +1,39 @@
...
+
+More information on Section 508 can be found at:
+
+- [United States Access Board](http://www.access-board.gov/508.htm)
+- [Section508.gov](http://www.section508.gov)
+- [HHS Digital Communications Division](http://www.hhs.gov/web/508)
+
+## Current browser/technology support
+
+The current browser support for the design system [can be viewed here](https://design.cms.gov/startup/browser-support/). We test with the following assistive technology pairings:
```
bernardwang: This should be `{{root}}` instead of `https://design.cms.gov`

line47:
Updated to use `{{root}}`

PR #646, opened 2020-03-09
[WNMGDS-359] Stop exporting VerticalNav and Tab internal components

bernardwang:
### Summary
Currently we export some internal components (i.e. `VerticalNavItem` and `VerticalNavItemLabel`). This PR removes all internal components that shouldn't be used by a user of a DS.


packages/core/src/components/Tabs/_Tabs.docs.scss
```diff
@@ -30,16 +30,6 @@ A `TabPanel` is a presentational component which accepts a tab's content as its
 Style guide: components.tabs.react-tabpanel
 */
 
-/*
```
bernardwang: `<Tab>` is used by `<Tabs>`, but not by the user.

PR #647, opened 2020-03-09
[WNMGDS-382] Update datefield and textfield input type

line47:
### Summary
- Changed number fields to use `type="text"`, `nputmode="numeric"` and `pattern="[0-9]*"` [more details here ](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/)
- Removed css that prevented arrows for `type="number"` inputs. This means the arrows will be visible for people using `<TextField type="number">` and who are not using the new `numeric` prop


bernardwang:
I believe the numeric type needs to be added in the masked textfields as well

line47:
Masked fields are not using the number type, are you saying to add this to the examples?
```
inputMode="numeric"
pattern="[0-9]*"
type="text"
```

bernardwang:
@line47 I was thinking of adding `inputMode` and `type` attributes automatically in `TextField.jsx` when a mask is used (because all our masks are number based). There's a few other changes that would be ideal to include in this PR as well (that I didn't include in the jira ticket, my bad)
- Remove the appearance styling in `TextField.scss` that hides the arrows for numbers
- Remove instances of `min` and `max` in `DateField` (they only apply to `type="number"` textfields)
- Update documentation warning against using `type="number"` for things like phone numbers, zips, etc.

bernardwang:
I'm also wondering if we should modify the `<TextField>` component to handle more of this automatically. Right now users just have to _know_  not to use `text="numeric"` and to use `inputMode="numeric"` and `type="text"`. 

Since 95% of the time we don't want users to be using `type="number"`, should we make this guidance easier to use? Random ideas of mine: 
- Output a console warning when someone uses `type="number"`
- Adding a new `numeric` prop that will automatically add the attributes
- Automatically add the attributes when `type="numeric"` (not a standard HTML input type)
- Make it impossible to use `type="number"` by override the `type` prop to automatically use `inputMode="numeric"` and `type="text"` when `type="number"` is given 

@davidakennedy @line47 thoughts?

packages/core/src/components/TextField/TextField.example.jsx
```diff
@@ -11,7 +11,14 @@ ReactDOM.render(
...
     />
-    <TextField defaultValue="123" label="Number field" name="number_example" type="number" />
+    <TextField
+      defaultValue="123"
+      label="Number field"
+      name="number_example"
+      type="text"
+      inputmode="numeric"
+      pattern="[0-9]*"
+    />
```
pwolfert: I just tested this on mobile Chrome, and it works like a charm.

line47:
A few thoughts from the comments above: 

> I was thinking of adding inputMode and type attributes automatically in TextField.jsx when a mask is used (because all our masks are number based).

Are we assuming that all masks will only be used for numbers? If so we might want to update our documentation to be more explicit. I'm thinking that might be a bit restrictive for folks to add those attributes automatically. 

> Remove the appearance styling in TextField.scss that hides the arrows for numbers

I'm wondering if we would want to keep this styling for those instances where folks are using `type=number`, unless we are saying that folks should not be using `type=number` at all.

> Remove instances of min and max in DateField (they only apply to type="number" textfields)

It looks like on DateField that we are using `yearMin` and `yearMax` props, are you saying we get rid of this functionality from DateField now that we are not using number type?

---
- I like the idea of `Output a console warning when someone uses type="number"`
- I'm also thinking about this `Make it impossible to use type="number" by override the type prop to automatically use inputMode="numeric" and type="text" when type="number" is given`. I'm concerned about this a bit because it means that we are not allowing type=number across the board and I'm wondering if there are some valid reasons for folks to use type=number

packages/core/src/components/TextField/TextField.example.jsx
line47: I like the idea of including additional characters based on the masked example

bernardwang: how did you test pattern? doesnt it only work inside a form and with a submit button? Im wondering if this will affect teams like app3 not using native validation...

pwolfert: I didn't test the pattern, but inputmode=numeric worked for bringing up a numpad on my phone

line47:
@bernardwang feel free to start in on this when you're ready - I'd like to talk through the updates you've made at some point too 👍

davidakennedy:
Overall, I like this change and was actually going to propose it, so ya'll beat me to it. 😄 

Some thoughts:

> I'm wondering if we would want to keep this styling for those instances where folks are using type=number, unless we are saying that folks should not be using type=number at all.

@line47 My take here would be that we strike a balance. We don't use type=number anywhere, and advise against it. However, we don't restrict people from using it, and we keep the styles in case they need them.

> It looks like on DateField that we are using yearMin and yearMax props, are you saying we get rid of this functionality from DateField now that we are not using number type?

I think we could get around this by having some validation via JS or maybe the `pattern` attribute?

bernardwang:
Changes I made in the most recent updates
- Added a `numeric` prop to `<TextField>` to handle `inputMode` and `type` automatically. 
- Added a console warning for fields using `type="numeric"`
- Refactored `<TextField>` to move mask logic into `<Mask>`

bernardwang:
> My take here would be that we strike a balance. We don't use type=number anywhere, and advise against it. However, we don't restrict people from using it, and we keep the styles in case they need them.

@davidakennedy We ended up with a solution similar to what you said, we don't use it anywhere, we advise against using it, and we warn developers who are using it. 

The one difference is that I decided to remove the styling that hides the arrows. I think if a team really does decide to use `type="number"`, then we should provide the default HTML and browser experience. The styling to remove the arrows was an imperfect solution to begin with, as it slightly improved the user experience, but is a bit misleading because you can still interact with the input with keyboard arrows.

packages/core/src/components/TextField/TextField.jsx
```diff
@@ -234,8 +223,16 @@ TextField.propTypes = {
    */
   multiline: PropTypes.bool,
   name: PropTypes.string.isRequired,
+  /**
+   * Sets `inputMode`, `type`, and `pattern` to improve accessiblity and consistency for number fields. Use this prop instead of `type="number"', see [here](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/) for more information.
```
line47: type="number" doesn't have a proper backtick

line47:
Looks good to me, one small formatting tweak in the prop explanation

packages/core/src/components/TextField/TextField.jsx
bernardwang: @line47 should be addressed in latest commit

PR #650, opened 2020-03-10
[WNMGDS-389] Add migration info to doc site

bernardwang:
### Summary
- Renames `core` folder to `cmsds`
- Updates documentation and package.json to reflect new npm package `@cmsgov/design-system`
   - The `docs` folder doesn't use the new package yet, as its not published yet.
- Adds migration information to the doc site


packages/cmsds/package.json
```diff
@@ -1,10 +1,11 @@
 {
-  "name": "@cmsgov/design-system-core",
-  "version": "3.6.1",
+  "name": "@cmsgov/design-systeme",
```
pwolfert: Very French 😆

packages/cmsds/README.md
```diff
@@ -49,10 +50,14 @@ The design system follows a variation of [ITCSS](http://thomasbyttebier.be/blog/
 ├── fonts
 ├── images
 └── src                 Non-compiled Sass and JSX
-    ├── base            Base HTML styles
```
pwolfert: 👍

pwolfert: This has moved down to `styles` now, right? I guess it says `etc...`, so it's not necessary to list it out explicitly 👍

packages/docs/package.json
```diff
@@ -6,6 +6,7 @@
   "repository": "CMSgov/design-system",
   "dependencies": {
     "@cmsgov/design-system-core": "^3.6.1",
+    "@cmsgov/design-system-layout": "^3.6.1",
```
pwolfert: Is this something temporary that will hopefully not be needed after we have new build scripts? Why is this needed?

packages/docs/src/pages/index.md
```diff
@@ -9,6 +9,13 @@ It is currently being applied to [HealthCare.gov](https://www.healthcare.gov/).
 
 ## Getting started
 
+<div class="ds-c-alert">
+  <div class="ds-c-alert__body">
+    <h2 class="ds-c-alert__heading">CMS Design System v4 has been released</h2>
+    <p class="ds-c-alert__text">For more information on v4 and how to upgrade, see our <a href="http://localhost:3000/startup/migrating-v4/">migration guide</a>.</p>
```
pwolfert: `localhost:3000` is hardcoded into this link

packages/docs/src/pages/startup/migrating-v4.md
```diff
@@ -0,0 +1,65 @@
...
+CMS Design System v4 introduces several breaking changes, and this migration guide will outline several steps to upgrade to the latest version.
+
+## NPM packages
+
+The `@cmsgov/design-system-core`, `@cmsgov/design-system-support`, `@cmsgov/design-system-layout` npm packages have been deprecated and replaced with a new consolidated package [`@cmsgov/design-system`](https://www.npmjs.com/package/@cmsgov/design-system). This is now the only dependency you need to use the design system and you can simply replace the old packages with `@cmsgov/design-system` in your `package.json`.
+
+The new file structure below illustrates where the old `support` and `layout` packages can now be found.
+
+```
+cmsds@4.0.0
```
pwolfert: Should this include `dist` at the top level?

packages/docs/src/pages/startup/migrating-v4.md
```diff
@@ -0,0 +1,65 @@
...
+## Fonts and Images
+
+Fonts and images are now stored in `@cmsgov/design-system/dist/fonts` and `@cmsgov/design-system/dist/images`. Previously they were stored in `@cmsgov/design-system/fonts` and `@cmsgov/design-system/images`.
+
+If you’re using a tool like webpack to collect the images and fonts during your app’s build process, you will need to update the appropriate path variables. A typical Sass import might have looked like this:
+
+```
+$font-path: "~@cmsgov/design-system-core/fonts";
+$image-path: "~@cmsgov/design-system-core/images";
+@import "~@cmsgov/ds-healthcare-gov/src/index.scss";
```
pwolfert: My bad, this and the one below should reference the design system package instead of the hc.gov site package.

packages/docs/src/pages/startup/sass-and-css.md
```diff
@@ -8,13 +8,13 @@ weight: 1
 The easiest way to add the design system's styles to your site is by referencing its minified CSS.
 
 1. Download the zip file from the latest CMS design system release and open that file.
-1. Copy the `packages/core/dist/index.css` file into a relevant place in your code base — likely a directory where you keep third-party libraries. In the example below, our directory is `css/vendor`.
+1. Copy the `packages/cmsds/dist/index.css` file into a relevant place in your code base — likely a directory where you keep third-party libraries. In the example below, our directory is `css/vendor`.
```
pwolfert: Should we rename the package folder again to just `design-system`? At some point I'm thinking this could be confusing.

packages/docs/src/pages/startup/migrating-v4.md
```diff
@@ -0,0 +1,65 @@
...
+
+In v4, it will need to be changed to
+
+```
+@import "~@cmsgov/ds-healthcare-gov/dist/index.scss";
+```
+
+## Fonts and Images
+
+Fonts and images are now stored in `@cmsgov/design-system/dist/fonts` and `@cmsgov/design-system/dist/images`. Previously they were stored in `@cmsgov/design-system/fonts` and `@cmsgov/design-system/images`.
```
pwolfert: Now that I see this here, I'm realizing this won't be true unless we update the existing build scripts to copy the fonts and images from src to dist and copy the scss files from src to dist. That shouldn't actually be that hard of a change, and I could do it separately from my build script PR

packages/cmsds/README.md
bernardwang: The contents of the `base` folder were moved into styles (`typography`, `icons`, `base`). I actually removed it `base` folder because I felt like it didnt do a great job of describing the type of  styles inside the folder, and also because theres a second `base` folder for the `ds-base` class. Thinking on it again, maybe it is worth having another layer of organization in `styles`, but I'm not sure what to call it.

bernardwang: Heres what it looks like now: 
```
styles
        ├── settings          // variables, mixins, scss that doesnt output css basically
        ├── typography   // Namespaced element selector for typography (ds-h1, ds-text)
        ├── base               // Namespaced element selector for body (ds-base)
       └── etc...  
```
@line47 @pwolfert What do yall think? What should the folder be called that contains `typography` and `base`? Fyi `icon` will eventually be moved to components, it doesnt belong here really

packages/docs/src/pages/startup/sass-and-css.md
bernardwang: hm thats a good point, i like that its consistent with the package name. I'd like to get @line47 @davidakennedy thoughts too

packages/docs/package.json
bernardwang: The docs package still needs to import the old packages because the new package isnt published yet, so we should be able to remove it after we publish `@cmsgov/design-system`. I added it back because the docs were broken in the branch to start with haha

packages/docs/src/pages/startup/migrating-v4.md
bernardwang: I think it depends on when we are deciding to update the doc site with this migration guide. If we are waiting until after the scripts are done, i dont think we need the `dist` because we arent including it in the release anymore right?

bernardwang: Oh I realized the documentation on this page references `dist` a lot, I think I misunderstood. I thought that once we move the SCSS src and assets inside the `dist` folder, we can just release the `dist` folder as the package?

bernardwang: This would have the other added benefit of users not having to update their fonts and image path variables, because although its changed on our end (moved to `dist`), if we just distribute `dist` then its the same for the users

packages/cmsds/README.md
pwolfert: I think it's fine how you have it now. It was a good move to put it in the styles folder because it's only related to styles, and the folders just keep the styles folder clean

pwolfert: Actually, here's another idea: What if the `utilities` folder moves inside styles too since it's only about utility CSS classes? And what if `base`, `icons`, and `typography` moves into utilities because they are actually utility classes?

pwolfert: One issue, however, is that the docs site is organized to have them called out separately from the rest of the utilities. Do they need to be? Could we have another landing page for the new Design section that touches on those elements and links to their pages in the utilities section?

packages/docs/src/pages/startup/sass-and-css.md
bernardwang: ok, now that we clarified some things in the CMS slack, even if we distribute the `dist` folder as originally planned, doesnt that mean the user doesnt have to navigate through the `packages` folder? The zip on the website probably shouldnt have that right?

pwolfert: Yeah, I think it is indeed a good idea to pare that zip on github down to something like
```
zip.zip
  ├── design-assets
  ├── dist
      ├── index.js
      ├── index.scss
      ├── etc...
  ├── examples
```
We don't have to call it `dist` here, but I can't think of another name

pwolfert: Or even just the contents of the `dist` folder inside the root of the zip, but I don't know if we would want to leave the other stuff out because of designers.

packages/docs/src/pages/index.md
```diff
@@ -9,6 +9,13 @@ It is currently being applied to [HealthCare.gov](https://www.healthcare.gov/).
 
 ## Getting started
 
+<div class="ds-c-alert">
+  <div class="ds-c-alert__body">
+    <h2 class="ds-c-alert__heading">CMS Design System v4 has been released</h2>
+    <p class="ds-c-alert__text">For more information on v4 and how to upgrade, see our <a href="https://design.cms.gov/startup/migrating-v4/">migration guide</a>.</p>
```
pwolfert: Would it actually work to use `{{root}}` here like we do later in this file? Or does it not work in HTML sections?

packages/docs/src/pages/startup/migrating-v4.md
pwolfert: Update: We had a slack discussion about this and concluded that the package would still need to have a `dist` folder in it but that it just wouldn't include a `src` folder anymore.

packages/docs/src/pages/index.md
bernardwang: Ah good point, i think it might work let me try

packages/docs/src/pages/startup/migrating-v4.md
pwolfert: This needs to be changed from `ds-healthcare-gov` to `design-system-core`, and the code block below it needs to change to `design-system`

pwolfert: Whoops, you did fix this 👍

README.md
```diff
@@ -57,8 +57,6 @@ These scripts can all be run from the root level of the repo:
   - Lints Sass using stylelint
 - `yarn test:e2e`
   - Runs end to end tests
-- `yarn test:e2e packages/core/Autocomplete`
-  - Runs a single end to end test, this example runs Autocomplete
```
line47: could this be updated to `yarn test:e2e packages/cmsds/Autocomplete` ?

packages/cmsds/README.md
```diff
@@ -49,10 +50,14 @@ The design system follows a variation of [ITCSS](http://thomasbyttebier.be/blog/
...
-    ├── components      Sass and React components
+    ├── components      React component source and test files
```
line47: Do we want to mention scss for the `components` section as well?

packages/cmsds/README.md
bernardwang: Ah I see. I think the biggest distinction between `utilities` and `base`/`typography` is that `utilities` are applying specific CSS properties, while `base`/`typography` are generally for HTML element related styles like `h1`, `p`, `body` etc (`icons` will be moved to components because it will become a React component). This difference is the reason why there's a difference in naming convention, `ds-h1` vs  `ds-u-font-size--title`. 

I was thinking about how there are `scss` files in all 3 folders, so instead of organizing according to js vs scss, it might be better to organize it by the website hierarchy. I was considering how `component` and `utilities` are the two largest pieces of documentation content, so it might be better to keep them at the root.  Similarly, `typography`, `base` are under the same top level website category of ["Styles"](https://gyazo.com/97896682de061524168520a6d83523bd) (or "Design" in later versions)

I'm not really sure whats the best situation here... I wish it would be an easy typical css/js organization, but I also think modeling it off of the website structure can make it easier to find relevant files for new maintainers

line47: I'm thinking of it like  `base`, `icons` and `typography` are high-level items that impact setting up a child DS and are used in components like alerts, buttons, etc... while the utilities are more for modification or extending components of the design system. To me, that's the distinction as to why they are grouped the way they are. 🤔 
I'd also say this is why spacing and grid are separated as well.

packages/docs/src/pages/startup/migrating-v4.md
```diff
@@ -0,0 +1,70 @@
...
+   ├── package.json
+   └── dist
+       ├── index.css        CSS
+       ├── index.scss
+       ├── index.js
+       ├── fonts
+       ├── images
+       ├── components
+       │   ├── Button
+       │   └── etc…
```
pwolfert: This one is an ellipsis character, whereas the one below is three `.`. Same with the one in `utilities`

packages/docs/src/pages/index.md
line47: {{root}} might only work in the context of Markdown

README.md
bernardwang: i tried it, and i dont think it works currently.

packages/cmsds/README.md
bernardwang: Oh right, I moved `grid` to utilities but I could see it going along with `spacing`, `typography` and `base` too...

packages/docs/src/pages/index.md
bernardwang: I see other examples of it in example html files, but I haven't seen another one in a markdown file. Technically this is HTML inside a markdown file so it should work... I just cant test because the doc building is a bit messed up right now :/

pwolfert: Right now that link will at least be valid in production. We can circle back to it later

bernardwang: nevermind! i got it, it works!

PR #651, opened 2020-03-13
[WNMGDS-394] Add design-system-scripts package with build script

pwolfert:
### Added
- `@cmsgov/design-system-scripts` package (`packages/design-system-scripts`) that contains a single script to start

### Changed
- Moved `fonts` and `images` directories in `packages/design-system` from the root into the `src` folder, and the build script will copy them into the `dist` folder.
  - If the build script is running on a child design system package, it will also copy the `fonts` and `images` folders from the core `@cmsgov/design-system` package to the child design system's `dist` folder so the assets are distributed directly from that package.

### Deprecated
- Old scripts in `tools` are not in working order since the `fonts`/`images` changes

#### Building the design system
From the root of the design system repo, the build script can be run on the core design system package with the following steps:
1. In the `packages/design-system-scripts` directory, run `npm link`. Note that `yarn link` will not work for what we're doing here.
2. In the root of the repo directory, run `npm link @cmsgov/design-system-scripts`
3. Run the build script with `yarn cmsds build packages/design-system`

#### Building a child design system
To preview what how it works for a child design system:
2. In the `packages/design-system-scripts` directory, run `npm link`
3. Download [child-design-system-example.zip](https://github.com/CMSgov/design-system/files/4350481/child-design-system-example.zip)
4. Extract it so that it's a sibling of the `design-system` repo on your filesystem and `cd` to it. Alternatively, extract it wherever you like but modify the `package.json` so that the relative paths to the packages are accurate. The relative paths are temporary until we actually publish the new packages.
5. Run `yarn install`. Note that you will have had to built the design system package following the instructions in the previous section.
6. Run `npm link @cmsgov/design-system-scripts`
7. Run `yarn cmsds build ./`
8. Open up index.html in a browser and make sure the compiled styles work.
9. Inspect the `dist` directory and make sure everything's there.


pwolfert:
I was sure I set the base to the `bernard/new-cmsds` branch like three times, but I guess it didn't take.

pwolfert:
I forgot that this assumes the core package has images and fonts in its `dist` folder, which it doesn't if it pulls from npm. I'm adding another step to get around that.

pwolfert:
And now I just discovered the problem with saving the `.css` and `.scss` files to `dist`. The `.scss` files now have ambiguous import statements. 😞

packages/design-system-scripts/gulp/build.js
```diff
@@ -0,0 +1,137 @@
+/**
+ * @file The build tasks handle compiling and optimizing both the
+ *  package assets and the documentation site. Essentially makes
+ *  everything production-ready.
+ */
+const babel = require('gulp-babel');
+const count = require('gulp-count');
+const del = require('del');
+const gulp = require('gulp');
+const getPackageName = require('./common/getPackageName');
```
bernardwang: I _think_ we can get rid of this because our directories will be named after their npm packages right?

bernardwang: although maybe we cant assume that for child DS, nevermind

packages/design-system-scripts/gulp/build.js
```diff
@@ -0,0 +1,137 @@
...
+const streamPromise = require('./common/streamPromise');
+const { compileSass } = require('./sass');
+const { printStats } = require('./stats');
+const { log, logTask, logIntroduction } = require('./common/logUtil');
+
+const CORE_PACKAGE_NAME = '@cmsgov/design-system-core';
```
bernardwang: Is this temporary? Like after we publish`@cmsgov/design-system@4.0.0` this variable will change?

packages/design-system-scripts/gulp/build.js
```diff
@@ -0,0 +1,137 @@
...
+        })
+      )
+      .pipe(gulp.dest(`${dir}/dist`))
+  );
+}
+
+/**
+ * Builds the source package
+ */
+async function buildSrc(sourcePackageDir) {
```
bernardwang: This organization is amazing btw, much more readable and it makes alot of sense!

packages/design-system-scripts/gulp/build.js
```diff
@@ -0,0 +1,137 @@
...
+  await cleanDist(sourcePackageDir);
+  await copyAll(sourcePackageDir);
+  await compileJs(sourcePackageDir);
+  await compileSass(sourcePackageDir);
+  logTask('✅ ', 'Build succeeded');
+  log('');
```
bernardwang: Whats this `log` for?

packages/design-system-scripts/package.json
```diff
@@ -0,0 +1,87 @@
+{
+  "name": "@cmsgov/design-system-scripts",
+  "version": "3.6.0",
```
bernardwang: should this be changed to 4.0.0

pwolfert: I figured we'd fix all the package versions with lerna when we go to publish or when we merge it into master

packages/design-system-scripts/gulp/build.js
pwolfert: Yeah, and that was my most recent source of error 😂😭

bernardwang:
I'm getting stylelint errors with this rule `scss/at-import-partial-extension-blacklist': ['scss'],`. I think we can remove it from `stylelint-config-design-system`

bernardwang:
> * In the `packages/design-system-scripts` directory, run `npm link`. Note that `yarn link` will not work for what we're doing here.
> * In the root of the repo directory, run `npm link @cmsgov/design-system-scripts`
> * Run the build script with `yarn cmsds build packages/design-system`

Just to clarify, this is a temporary step right? Do we need to do this linking because `@cmsgov/design-system-scripts` isnt actually published yet?

Other question, is `yarn cmsds build packages/design-system` equivalent to `npx cmsds build ...` or something?

packages/design-system-scripts/gulp/sass.js
```diff
@@ -0,0 +1,86 @@
...
+const getDocsDistPath = require('./common/getDocsDistPath');
+const changed = require('gulp-changed');
+const count = require('gulp-count');
+const cssnano = require('cssnano');
+const gulp = require('gulp');
+const gulpIf = require('gulp-if');
+const path = require('path');
+const postcss = require('gulp-postcss');
+const postcssImport = require('postcss-import');
+const postcssInliner = require('postcss-image-inliner');
```
bernardwang: Do we need this in the future? I thought we decided to stop inlining images?

pwolfert: Yep, good catch

pwolfert:
> Just to clarify, this is a temporary step right? Do we need to do this linking because @cmsgov/design-system-scripts isnt actually published yet?

Exactly

> is yarn cmsds build packages/design-system equivalent to npx cmsds build ... or something?

Yep, `npx cmsds` would also execute the script

line47:
When running `yarn build` I'm getting this error 

```
Message:
    No file found for images/information.svg
Details:
    postcssNode: background-image:url(images/information.svg)
    fileName: /Users/scottweber/Documents/github/design-system/packages/design-system/src/index.css
    domainEmitter: [object Object]
    domainThrown: false
```

packages/design-system-scripts/gulp/sass.js
bernardwang: Yesterday I was trying to find some info online about inlining vs minification. Do you have any context on that, aside from simplifying our build scripts and assets? I was thinking one other advantage might be that the CSS bundle will be slightly smaller (images can be loaded after).

bernardwang:
@line47 I believe `yarn start` will be broken in this PR because we moved the images. I believe in this PR   most of our old scripts will be broken. This PR adds new scripts for compiling design system assets, and the new doc site generation script is not implemented yet.

bernardwang:
@pwolfert Just to confirm, for the child design system testing, the scripts currently cant override and combine JS from two packages yet right? In my testing only the scss is being compiled

pwolfert:
@bernardwang, the JavaScript imports do work, but the compiled JS doesn't bundle all the dependencies into one file like the Sass does, because we can rely on npm's dependency tree to make sure the right code is imported. If the child design system imports the design system JS and exports their own version of a button component, you'll only see a tiny bit of code in the compiled output, and that's expected.

pwolfert:
@line47, to confirm, what Bernard said about the scripts being broken because images moving is correct.

line47:
I was able to build the design system and build the child design system successfully 🎉

bernardwang:
@pwolfert Im seeing for the child DS, no JS compiled at all :/
https://gyazo.com/1d4c405a0b7af9ea10143f74a2cbe5c2

pwolfert:
Okay, updated the example zip with a change to the package.json and updated the instructions to have you `yarn install` in the child ds first.

line47:
Here is what's in my dist directory of the child design system after following the instructions

Screen Shot 2020-03-18 at 2 19 42 PM

- components had buttonoverride.js and index.js
- fonts had the bitter and openSans fonts
- images had images

PR #658, opened 2020-03-18
[WNMGDS-380] add note about links in legends in IE11

line47:


davidakennedy:
Maybe we can be more specific? Something like:

Avoid links in legends when possible. Links inside legends are not read by  some screen readers, most notably Internet Explorer 11, paired with JAWS. If you need a pattern like this, put the link outside of the legend element, but keep all other useful hint text in the legend.

Some background discussions on this issue:

[Move hint error outside legend](https://github.com/CMSgov/design-system/pull/546)
[Links embedded in a fieldset > legend not reading correctly with JAWS + IE11](https://github.com/CMSgov/design-system/issues/529)

line47:
Closing this PR because the folder structure changed - please see https://github.com/CMSgov/design-system/pull/660

PR #659, opened 2020-03-19
[WNMGDS-395] Build-docs and dev-server scripts

pwolfert:
### TODO
- [x] `cmsds build-docs` builds src and then builds the docs package into html pages
- [x] `cmsds build-docs` runs webpack statically after building src and docs
- [x] `cmsds start` runs webpack dev server and watches for changes to build the source package or the docs
- [x] Fix babel errors when using the `build-docs` script in child design systems.
- [x] Update or remove the unit tests for the script modules

### Summary
- Added the `build-scripts` command to the `design-system-scripts` package.
- Refactored doc generation scripts to handle child design systems
- Changed the `yarn build` script alias in package.json to execute the new build script
- Temporarily removed test and lint scripts

#### Core CMSDS:
1. Run `yarn install`
2. Run `yarn build` to build `design-system` and `design-system-docs`
3. Run `yarn start` to build, serve, and watch files for the doc site
4. Test the site is reloaded correctly in the following scenarios:
     a. React component in `design-system` is updated
     b. React component prop documentation in `design-system` is updated
     c. HTML and React component examples in `design-system` are updated
     d. Documentation content `design-system` is updated (i.e. `.docs.scss` file)
     e. CSS in `design-system` is updated
     f. React component in `design-system-docs` is updated
     g. CSS in `design-system-docs` is updated

#### Child DS
1. Download and extract 
[child-design-system-example .zip](https://github.com/CMSgov/design-system/files/4471973/child-design-system-example.zip)

2. Make sure `child-design-system-example` is in the same directory containing the core CMSDS repo, [aka adjacent to the `design-system` repo folder](https://gyazo.com/44e05b866805335b6f9a060ff327d08b).
2. Run `yarn install` (This will also `npm link` the core ds packages)
3. Go through the same tests listed above for the core CMSDS
4. Additionally test these overrides:
    a. Overriding a core React component and prop docs, ensuring it is outputted in the `dist` (i.e. `src/Button/Button.jsx`)
    b. Overriding a core React documentation (i.e. `src/Button/_Button.docs.scss`)
    c. Overriding core DS css (i.e. `src/Button/Button.scss`)
    d. Adding new DS components, documentation and css (i.e. `src/NewBadge`)
    e. Overriding doc site css (i.e. `docs/example.scss` or `docs/index.scss`)


packages/design-system-scripts/gulp/docs/docs.js
```diff
@@ -0,0 +1,103 @@
...
+    const sourcePackageDirs =
+      pkg.name === CORE_PACKAGE_NAME
+        ? [sourcePackageDir]
+        : [sourcePackageDir, `node_modules/${CORE_PACKAGE_NAME}`];
+
+    // Consolidate the chain of docs directories into one temporary working directory,
+    // then perform all our operations on that working directory as if it were
+    // our docs package. This allows us to have child design systems with their
+    // own docs that inherit all the docs from the core design system.
+    const tempDir = await copyDocsToTempDir(docsPackageDirs);
```
pwolfert: This is an important part. Instead of building everything directly from the docs package, we copy the parent docs package and child design system docs package into a working directory first.

packages/design-system-scripts/gulp/docs/generatePages/generateReactExample.js
```diff
@@ -0,0 +1,91 @@
...
+ * @param {String} rootPath - Root docs site path
+ * @return {Promise}
+ */
+function generateReactExample(page, docsPath, rootPath) {
+  return new Promise((resolve, reject) => {
+    if (rootPath) {
+      rootPath = `${rootPath}/`;
+    }
+
+    const examplePath = path.resolve(__dirname, '../../../packages', page.reactExamplePath);
```
pwolfert: This is an example of a line that will need to change. This is from the old script, when this script was in this relative location to the packages directory. It's not necessary to use `__dirname` at all anymore. The new version of this line might be as simple as:
```js
const examplePath = path.resolve(page.reactExamplePath);
```

packages/design-system-scripts/gulp/docs/generatePages/savePage.js
```diff
@@ -0,0 +1,98 @@
...
+}
+
+/**
+ * @param {String} uri
+ * @param {String} docsPath
+ * @returns {Object}
+ */
+function docsFilePath(uri, docsPath) {
+  if (uri === 'public') throw Error('Filename can\'t be "public"');
+  const dir = path.resolve(docsPath, uri);
```
pwolfert: This is an example of a `path.resolve` call that was changed. The old version used to look like this:
```js
const dir = path.resolve(__dirname, '../../../', docsPath, uri);
```

packages/design-system-scripts/gulp/docs/docs.js
bernardwang: I think in our existing doc script we just process both directories and then merge the pages using `uniquePages`. Is there a particular benefit for a temp directory?

pwolfert: Ah, I didn't pay attention to the `uniquePages` part. This may be unnecessary then if we modify `uniquePages` to have the new desired behavior that maintains the override order

pwolfert: Maintaining the original src paths might end up being the better option for link generation

packages/design-system-scripts/gulp/docs/runWebpackStatically.js
```diff
@@ -0,0 +1,46 @@
+/**
```
pwolfert: My notes on this:

[Referring to the old scripts,] I just realized how webpack works differently between the static build and the _start_ command. When you run `yarn start`, the underlying gulp watch task starts the `server` task, and that runs its own webpack instance that is separate from the one in the `webpack` gulp task in `gulp/webpack.js`. This was confusing! So I'm going to try to make that less confusing by naming better. I'm going to call them `docs/runWebpackStatically.js` and `docs/runWebpackServer.js`.

packages/design-system-scripts/gulp/docs/generatePages/generateDocPage.js
```diff
@@ -0,0 +1,90 @@
...
+      // In development mode we let the client handle all of the React rendering,
+      // since if we were generating the HTML pages in our build process, Gulp would
+      // need restarted each time a React file changes, which is super annoying.
+      return '';
+    }
+
+    // On the client-side the "root" and "githubUrlBase" variables are defined
+    // via Webpack, but we also need to define them here for "server-side" rendering
+    process.env.root = rootPath;
+    process.env.githubUrlBase = githubUrl;
```
pwolfert: Notes:
> Documentation pages are fed data through `window.pages` and `window.routes`. I want to make the docs package public and remove its dependencies to relative files outside its package folder, so I need to pass in necessary data at build time through global variables in the same fashion. For example, I want to pass in the github url root for the repository for which we're building a docs site. 

This line is used when we statically render the html files. The corresponding definition of this for the webpack dev server (`yarn start`) is in `packages/docs/webpack.config.js` using the define plugin.

packages/design-system-scripts/gulp/docs/runWebpackServer.js
```diff
@@ -0,0 +1,46 @@
...
+ */
+
+const argv = require('yargs').argv;
+const path = require('path');
+const webpack = require('webpack');
+const webpackDevMiddleware = require('webpack-dev-middleware');
+const webpackHotMiddleware = require('webpack-hot-middleware');
+const webpackStatsConfig = require('../webpack-stats');
+
+module.exports = (gulp, shared) => {
```
pwolfert: Note that this is a raw copy of the old script and hasn't been updated to move way from defining tasks with gulp. I mostly wanted the file names committed as a template of where I was going to go next

packages/docs/webpack.config.js
```diff
@@ -10,7 +10,7 @@ const webpack = require('webpack');
  * @param {Boolean} hotReload - Enable Webpack's hot module replacement
  * @return {Object} Webpack config
  */
-function createConfig(docsPath, rootPath = '', srcPaths, hotReload = true) {
+function createConfig(docsPath, srcPaths, githubUrl, rootPath = '', hotReload = true) {
```
pwolfert: This function will likely need further changes, and I've also been debating with myself whether this should live here in the docs package or in the scripts package.

packages/design-system-scripts/cli.js
```diff
@@ -20,24 +26,60 @@ yargs
...
+        .option('rootPath', {
+          default: '',
+          description:
+            'The path of the docs site relative to the domain root. For example, if your docs site is hosted at www.domain.com/design/ your rootPath would be `design/`'
+        })
+        .option('githubUrl', {
+          default: '',
+          description:
+            'The base path for your GitHub repository URLs. This is used to render links to releases, issues, etc.'
+        });
```
pwolfert: I defined these for this command only because I figured neither of them would need to be customized for the dev server. Root path would always be `''` for hosting at the root of `localhost:3000`, and the githubUrl really doesn't matter.

packages/design-system-scripts/gulp/docs/generatePages/savePage.js
bernardwang: Could you remind me why we dont need `__dirname` or `../../../` again? Is it because `docsPath` is now just `packages/design-system-docs/dist`?

packages/design-system-scripts/gulp/docs/runWebpackStatically.js
bernardwang: thank you thats a helpful explanation!

packages/design-system-scripts/gulp/docs/generatePages/generateDocPage.js
bernardwang: What do you think about eventually moving the webpack config into the scripts package? Is that possible?

packages/docs/webpack.config.js
bernardwang: what are some of the pros of keeping it in the docs package?

packages/design-system-scripts/cli.js
bernardwang: Ah, were these customizable in the original scripts? and when would we define `githubUrl`?

package.json
```diff
@@ -21,84 +20,19 @@
...
-    "@babel/preset-env": "^7.8.4",
-    "@babel/preset-react": "^7.8.3",
-    "@babel/register": "^7.8.4",
+    "@cmsgov/design-system": "file:./packages/design-system",
     "@cmsgov/design-system-core": "latest",
+    "@cmsgov/design-system-docs": "file:./packages/docs",
+    "@cmsgov/design-system-scripts": "file:./packages/design-system-scripts",
     "@cmsgov/eslint-config-design-system": "file:./packages/eslint-config-design-system",
     "@cmsgov/stylelint-config-design-system": "file:./packages/stylelint-config-design-system",
-    "autoprefixer": "9.6.0",
```
bernardwang: removing all these deps... very nice

packages/design-system-scripts/gulp/docs/generatePages/savePage.js
pwolfert: It's because where we'll be executing the script (the pwd) vs where the script lives isn't always the same now. In the CMSDS repo, we'll be running it from the root, but the script will live in packages/design-system-scripts. In child design systems, it could be anywhere really. `path.resolve(__dirname)` would give you the directory where that specific module is (aka `/Users/whatever/design-system/packages/design-system-scripts/gulp/docs/generatePages/savePage.js`. We don't really care where that specific module is because the only path that matters is where the file we're referencing is relative to where the script is being executed. If you were to execute the script in your home directory,  `path.resolve('.')` will always return your home directory no matter which module is calling it, which is more helpful.

packages/design-system-scripts/gulp/docs/generatePages/generateDocPage.js
pwolfert: I was debating whether or not to move it into the scripts package. It is so tightly coupled to the build scripts that I think it might be a more appropriate place to have it there. But then I've also thought that the react components for the docs site are also tightly coupled to the build script, so if I were to move the webpack config, I may as well move the react components too and just keep the docs package as the documentation itself. I couldn't come to a solid decision on my own, so I went with having fewer changes for you to review.

packages/docs/webpack.config.js
pwolfert: Answered this in a comment above except that it wasn't pros so much as a weak sippery slope argument against it. I think it's totally reasonable to move it into scripts.

packages/design-system-scripts/cli.js
pwolfert: This was not customizable in the original script: https://github.com/CMSgov/design-system/blob/master/packages/docs/src/scripts/shared/githubUrl.js#L2

packages/design-system-scripts/gulp/docs/generatePages/generateDocPage.js
pwolfert: I could see the whole `docs/src/scripts` folder moving out of the docs package if we wanted that package to just be documentation and the scripts package deals with how to compile it all

bernardwang: Ah I see what you mean, so you didnt want to move the `webpack.config` without also moving `docs/src/scripts`? Down the line I was also thinking of doing moving that into scripts so that child DS's could get updates to the doc site even if they only have documentation content. This way the experience of writing documentation would be something along the lines of [Docz's build in components](https://www.docz.site/docs/built-in-components).

packages/design-system-scripts/gulp/build.js
```diff
@@ -37,7 +39,17 @@ function copySass(dir) {
 }
 
 function copyAll(dir) {
-  return Promise.all([copyAssets(dir), copyJson(dir), copySass(dir)]);
+  // Check to see if this is a core package. If it's not, copy assets from core npm package
+  const copyAssetsTask = async (dir) => {
```
bernardwang: @pwolfert I moved the copying from core npm package logic here, in order to generalize/simplify `copyAssets`. This way after I remove the temp dir, I can use it [here](https://github.com/CMSgov/design-system/pull/659/commits/af42b932eb28752c0f22ffa2e6d7155e51a31723#diff-4cbc3eb22061dce437a86ae65943d8d6R53) in the build-docs script. I'm still wrapping my head around some of this, so lmk if this looks about right to you

pwolfert: Yeah, that works too. I had only intended using this for the source package and not the docs package since the docs package only has images and not fonts. This anonymous function below, however, will _either_ copy from core _or_ the `dir`. I believe you want it to `await` both of the `copyAssets` calls.

packages/design-system-scripts/gulp/docs/docs.js
```diff
@@ -85,6 +86,7 @@ module.exports = {
     await cleanDist(docsPackageDir);
     await extractReactDocs(sourcePackageDirs, options.rootPath);
     await generatePages(sourcePackageDirs, docsPackageDir, reactDataPath, options);
+    await runWebpackStatically(sourcePackageDir, docsPackageDir, options.rootPath);
```
bernardwang: @pwolfert Was this what you had in mind for running webpack statically? I also have to call `compileDocsSass` after this right?

https://github.com/CMSgov/design-system/pull/659/commits/9a28dabce8364f050b64afea61167ac737a312b2

packages/design-system-scripts/gulp/build.js
bernardwang: Should be fixed here: https://github.com/CMSgov/design-system/pull/659/commits/6c6747329cd09fc51bfea78f3108865df4ec5faf

packages/design-system-scripts/gulp/docs/docs.js
pwolfert: Yeah, that looks great! Yes about compiling sass; the original did it in this order: https://github.com/CMSgov/design-system/blob/master/tools/gulp/build.js#L95

packages/design-system-scripts/cli.js
```diff
@@ -20,24 +26,60 @@ yargs
     },
     handler: async argv => {
       const { build } = require('./gulp/build');
-      await build(argv.sourcePackageDir);
+      await build(argv.sourcePackageDir, { ...argv });
+    }
+  })
+  .command({
+    command: 'build-docs <sourcePackageDir> <docsPackageDirs..>',
```
bernardwang: @pwolfert In what case would there be multiple `docsPackageDirs`? I thought these were the two use cases:
`build-docs packages/design-system packages/design-system-docs`
and
`build-docs ./src ./docs` (in the case of mgov ds)

pwolfert: The second use case is actually
```
yarn cmsds build-docs . node_modules/@cmsgov/design-system-docs docs
```

pwolfert: That's in order to inherit all the documentation from the core documentation package

pwolfert: An alternative would be to read the package.json in the specified docs directory and to do the same thing we're doing for the assets in the source package, where if it's not the core one, we include the core one.

pwolfert: https://github.com/CMSgov/design-system/pull/659/files#diff-b902b49a887f2e3a4fd421bf1e450047R46

bernardwang: yeah i think maybe we could do that!

packages/design-system-scripts/cli.js
```diff
@@ -56,8 +64,13 @@ yargs
     },
     handler: async argv => {
       logIntroduction();
-      const { startDocsServer } = require('./gulp/server');
-      await startDocsServer(argv.sourcePackageDir, argv.docsPackageDir, { ...argv });
+      const { buildSrc } = require('./gulp/build');
```
bernardwang: I reorganized these main task to be more granular so that we have better control of which tasks are run and in which order. I.e. if we break out `printStats` from `build` we can run it last in the case of `build-docs`. This was also helpful for `watch` task, which needed to run specific parts of each task depending on the file type it is watching

packages/design-system-scripts/gulp/docs/webpack/createWebpackConfig.js
```diff
@@ -2,10 +2,10 @@
 const fs = require('fs');
 const path = require('path');
 const webpack = require('webpack');
-const getDocsDistPath = require('../common/getDocsDistPath');
+const getDocsDistPath = require('../../common/getDocsDistPath');
 
-function createWebpackConfig(sourcePackageDir, docsPackageDir, options) {
-  const distPath = getDocsDistPath(docsPackageDir, options.rootPath, 'public/scripts/');
```
bernardwang: Oops this was supposed to be in the next commit, this is organized a bit badly

packages/design-system-scripts/gulp/docs/webpack/index.js
```diff
@@ -7,19 +7,19 @@
  */
 const path = require('path');
 const util = require('util');
-const webpack = util.promisify(require('webpack'));
```
bernardwang: @pwolfert I sunk a lot of time into this bug where webpack would work fine statically, but it would throw this undefined error for webpack static. I eventually found out that removing the promisify for the webpack server did the trick, but I don't really understand why because I was using `await` for both anyway?

packages/design-system-scripts/cli.js
pwolfert: 💯

pwolfert: I figured I'd have to refactor these as I went along but just wanted a working basic build script at first

packages/design-system-scripts/gulp/docs/webpack/index.js
pwolfert: Hmm, I'm not sure why that would be. I'm sorry that it caused you so much grief though

bernardwang:
@line47 @pwolfert This PR is in a semi reviewable state. If you follow the `How to test` instructions the build-docs script should be working for the core CMSDS. I think it might make more sense to have another PR for making it work for child DS, as this is already a really big and long PR. 

Tmr morning I will clean up the tests and this should be reviewable

package.json
```diff
@@ -4,101 +4,30 @@
...
-    "build": "NODE_ENV=production gulp build --env=production",
-    "build:theme": "yarn build --theme",
+    "build": "NODE_ENV=production yarn cmsds build-docs packages/design-system packages/design-system-docs --env=production",
     "bump": "lerna publish --skip-git --skip-npm",
     "precommit": "lint-staged",
-    "postinstall": "lerna bootstrap",
+    "postinstall": "lerna bootstrap && npm link ./packages/design-system-scripts",
     "posttest": "yarn lint",
-    "start": "NODE_ENV=development gulp watch --env=development",
+    "start": "NODE_ENV=development yarn cmsds start packages/design-system packages/design-system-docs --env=development",
```
pwolfert, line 12: I think the `--env=development` can be removed now, right?

packages/design-system-scripts/gulp/watch.js
```diff
@@ -0,0 +1,89 @@
...
+const { copyAll } = require('./build');
+const {
+  extractReactDocs,
+  generatePages,
+  copySourcePackageAssets,
+  copyDocsPackageAssets
+} = require('./docs');
+const { runWebpackServer } = require('./docs/webpack');
+
+async function watchSourcePackage(sourcePackageDir, docsPackageDir, options, browserSync) {
```
pwolfert, line 19: I love the organization in this file

bernardwang:
@line47 @pwolfert I updated the description with details on how to test. This PR should now pass tests and should support child design systems as well.

pwolfert:
I downloaded your latest version of the child design system and ran `yarn install`, but there was a parsing error when trying to build:

```
ERROR in /Users/patrickwolfert/Projects/child-design-system-example/node_modules/@cmsgov/design-system-docs/src/index.jsx 9:15
Module parse failed: Unexpected token (9:15)
You may need an appropriate loader to handle this file type, currently no loaders are configured to process this file. See https://webpack.js.org/concepts#loaders
| 
| function render() {
>   const Page = <Docs page={window.page} routes={window.routes} />;
| 
|   if (process.env.NODE_ENV === 'production') {
 @ multi react-hot-loader/patch webpack-hot-middleware/client /Users/patrickwolfert/Projects/child-design-system-example/node_modules/@cmsgov/design-system-docs/src/index.jsx index[2]
ℹ ｢wdm｣: Failed to compile.
```

line47:
☝️ I got the same error as Patrick did above

line47:
It looks like the docs.css is not working on the child ds like it does the core doc site.
.docs__sidebar is not applied in the same way and the left nav is wider for the chid ds doc site.

changing the order of this index.scss fixes it but I don't think that's what we want to do
```@import '@cmsgov/design-system-docs/src/index.scss';
@import '../../src/index.scss';```

**Comment by bernardwang** [comment]
@line47 Could you elaborate on that? Do you mean overriding the default documentation isnt working for the child DS?

**Comment by bernardwang** [comment]
@pwolfert Sorry for the last minute updates, but you need to pull https://github.com/CMSgov/design-system/pull/659/commits/dc852102f3243c48f92f3a14a0bfcf57ebdae463 to fix that error. I also just pushed up some other fixes I found and uploaded an updated child DS example

**Comment by pwolfert** [comment]
I love the override messages:

Screen Shot 2020-04-13 at 5 21 40 PM

I tried going to one of the component pages but got a bunch of errors. I've also tried spinning up a simple http server that serves the `docs/dist` folder, and I got similar JavaScript errors on component pages.

**Comment by line47** [comment]
@bernardwang this all for running a child ds.

- I expected the left nav to be the same width as it is on design.cms.gov but the class `docs__sidebar` is being overridden. I think it might have to do with the ordering of CSS. 

**Here is a screenshot of the wide nav**
Screen Shot 2020-04-14 at 9 49 28 AM


I also noticed that the css is not being minimized (should it be?) http://localhost:3000/index.css

**Comment by bernardwang** [comment]
@pwolfert I think I know which error you are talking about, it should be fixed in the latest commit!

@line47 The CSS wont be minimized unless we pass in `NODE_ENV = production`, which by default only happens on `yarn build`. I'm not able to reproduce that wide nav in the child DS some reason... It looks like this for me on the default child design system example.
Screen Shot 2020-04-14 at 4 12 10 PM

**Comment by pwolfert** [comment]
Child design system build and start scripts work! LGTM

**Comment by line47** [comment]
This is fantastic and working as expected! I can make all the modifications called out in the lists above for both the core and child DS.

The left nav is no longer an issue for me, not sure what was up there.

**One small issue I found:**
When I made changes to `docs/src/example.scss` there were no changes. 

I do get the terminal message however the changes are not visible 
``` 
2 Sass files processed in ./docs
[Browsersync] Reloading Browsers...
```

I was adding this to the `example.scss` file 
```
.ds-display {
    font-style: italic !important;
}
```

**Comment by bernardwang** [comment]
@line47 I tested it on the core and child DS and adding this seemed to work for me.
```
.ds-display {
    font-style: italic !important;
}
```
Screen Shot 2020-04-15 at 3 55 32 PM
Since I can't reproduce it I'm going to go ahead and merge! I can get any follow ups in a different PR

PR #664, opened 2020-04-02
remove archive folder and old Sketch file

line47:
### Removed
Removed old Sketch file in the and the design-assets archive folder


line47:
Here is the conversation around it - https://github.com/CMSgov/design-system/pull/439

PR #669, opened 2020-04-13
[WNMGDS-423] - Add focus styles to Sketch UI kit

line47:
### Added focus styles for 
- Links
- Buttons
- Checkboxes & Radios
- Text inputs & Selects
- Tabs
- Navigation 
Screen Shot 2020-04-13 at 12 00 29 PM
Screen Shot 2020-04-13 at 12 11 21 PM
Screen Shot 2020-04-13 at 12 11 28 PM
Screen Shot 2020-04-13 at 12 11 42 PM
Screen Shot 2020-04-13 at 12 11 48 PM
Screen Shot 2020-04-13 at 12 16 32 PM
Screen Shot 2020-04-13 at 12 19 12 PM
Screen Shot 2020-04-13 at 12 19 21 PM
Screen Shot 2020-04-13 at 12 19 25 PM


bernardwang:
This might be part of a larger sketch clean up, but a lot of the components are poorly named, either with labels that dont match the design, or with generic "Copy" inside the label. These aren't symbols, so I guess its less important, but ideally they would be labeled well. 
Screen Shot 2020-04-14 at 10 28 16 AM
Screen Shot 2020-04-14 at 10 29 53 AM

line47:
Yeah I'm thinking this is a larger effort for cleanup. I cleaned up some naming on the symbols I was touching. I still need to put in some follow on tasks for specific updates.

melwoodard:
Chiming in late! Most looks good, but a couple small enhancements - when I change a radio button or checkbox symbol from default to focus state, it bumps down and to the right 1px to accommodate the border. It'd be great to update the non-focus states of these components to include 1px of padding on the top and left sides of the symbol so that all states are aligned. 

checkbox-focus

radio-focus

Similar 1px shift happens with the label in a tab

tabs-focus


Should I log a separate jira ticket @line47?

line47:
Thanks @melwoodard please file a ticket in JIRA or if you feel inspired put in a Pr for the fix 😄

melwoodard:
Filed! :smile: https://jira.cms.gov/browse/WNMGDS-450

PR #671, opened 2020-04-14
[WNMGDS-354] Collapse React prop documentation table on small screens

nichia:
The React props documentation table for some components (i.e. Autocomplete, Date field, Vertical Nav) can get wider than the page. This PR explores the options to compact the table and add horizontal scrolling. 

### Changed
Update React props documentation table to:
- include horizontal scrolling when the table is wider than the page width
- be responsive on small screens (width <= 544px) and reformat the table to a row layout for each prop with it's corresponding title

### How to test
- Select a component and ensure that horizontal scrolling is enabled when the screen width is narrower than the table width
- Continue reducing the screen width, on small screen width, check that the table is reformatted to display each React prop with a card layout
- Accessibility test - check that screen reader can recognize the reformatted table on small screen width

### DIFF
--- a/packages/docs/src/scripts/components/ReactPropDoc.jsx
+++ b/packages/docs/src/scripts/components/ReactPropDoc.jsx
@@ -71,15 +71,15 @@ class ReactPropDoc extends React.PureComponent {
   render() {
     return (
       <tr>
-        <td>
+        <td data-title="Name">
           <code className="ds-u-font-weight--bold">{this.props.name}</code>
           {this.isRequired()}
         </td>
-        <td>
+        <td data-title="Type">
           <code>{this.type()}</code>
         </td>
-        <td>{this.defaultValue()}</td>
-        <td>{this.description()}</td>
+        <td data-title="Default">{this.defaultValue()}</td>
+        <td data-title="Description">{this.description()}</td>
       </tr>
     );
   }
--- a/packages/docs/src/scripts/components/ReactPropDoc.jsx
+++ b/packages/docs/src/scripts/components/ReactPropDoc.jsx
@@ -69,6 +69,7 @@ class ReactPropDoc extends React.PureComponent {
   }
 
   render() {
+    // Specify ARIA roles attribute for table to make responsive HTML table accessible to screen readers
     return (
       <tr role="row">
         <td role="cell" headers="columnname">
--- a/packages/docs/src/scripts/components/ReactPropDoc.jsx
+++ b/packages/docs/src/scripts/components/ReactPropDoc.jsx
@@ -13,7 +13,7 @@ class ReactPropDoc extends React.PureComponent {
 
   description() {
     if (this.props.description) {
-      return <div dangerouslySetInnerHTML={{ __html: this.props.description }} />;
+      return <span dangerouslySetInnerHTML={{ __html: this.props.description }} />;
     }
   }
 
@@ -70,16 +70,36 @@ class ReactPropDoc extends React.PureComponent {
 
   render() {
     return (
-      <tr>
-        <td data-title="Name">
+      <tr role="row">
+        <td role="cell" headers="columnname">
+          <span className="docs_table__column-header ds-u-font-weight--bold" aria-hidden="true">
+            Name
+            <br />
+          </span>
           <code className="ds-u-font-weight--bold">{this.props.name}</code>
           {this.isRequired()}
         </td>
-        <td data-title="Type">
+        <td role="cell" headers="columntype">
+          <span className="docs_table__column-header ds-u-font-weight--bold" aria-hidden="true">
+            Type
+            <br />
+          </span>
           <code>{this.type()}</code>
         </td>
-        <td data-title="Default">{this.defaultValue()}</td>
-        <td data-title="Description">{this.description()}</td>
+        <td role="cell" headers="columndefault">
+          <span className="docs_table__column-header ds-u-font-weight--bold" aria-hidden="true">
+            Default
+            <br />
+          </span>
+          {this.defaultValue()}
+        </td>
+        <td role="cell" headers="columndescription">
+          <span className="docs_table__column-header ds-u-font-weight--bold" aria-hidden="true">
+            Description
+            <br />
+          </span>
+          {this.description()}
+        </td>
       </tr>
     );
   }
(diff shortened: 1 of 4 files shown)


line47:
This looks good @nichia I really like the reformatting of the table at smaller screens! 🎉 
I have a few suggestions for the styling at larger screens as well as smaller screens I'd like to suggest.

- Consider adding the class `ds-c-table--borderless` on the table to give it less visual noise. This gives a simpler look to the table which I think would make it easier to scan. see the screenshot below.
Screen Shot 2020-04-15 at 9 13 20 AM


- Consider altering the styling of the table at small screens to stack the content instead of keeping 2 cells left to right. 
    - To achieve this we can eliminate the absolute positioning and fixed percentages and let the content flow naturally and the containers will adapt to fit. 
    - I would also suggest making the internal lines less bold at smaller screen sizes to take the visual noise down.
     - Highlighting the header would be a great way to break up the scrolling and let a user know where a new prop begins. 

Screenshot of alternate styling that stacks the content by TD 
Screen Shot 2020-04-15 at 9 20 57 AM

packages/docs/src/styles/components/_ReactPropDocs.scss
```diff
@@ -0,0 +1,63 @@
...
+
+.docs_table {
+  @media (max-width: $width-sm) {
+    table,
+    thead,
+    tbody,
+    th,
+    td,
+    tr {
+      display: block;
```
davidakennedy, line 15: Thanks @nichia for jumping on a call this morning to talk about testing with screen readers, and for being proactive about accessibility. 🎉 

This solution looks elegant, but unfortunately breaks the semantics of tables for screen readers. Using `display: block` on tables elements means that screen readers don't know it's a table any longer. So users can't get the extra context needed when browsing it.

I shared these links with @nichia, which outline why and how to approach responsive tables in an accessible way. It's more code, but possible.

See:
What display properties do to tables: https://adrianroselli.com/2018/02/tables-css-display-properties-and-aria.html
A responsive, accessible table: https://adrianroselli.com/2017/11/a-responsive-accessible-table.html

I'd recommend taking one of the approaches outlined in the post. I'd be happy to test again too!

packages/docs/src/styles/components/_ReactPropDocs.scss
```diff
@@ -0,0 +1,50 @@
...
+      /* Hide table headers (.ds-u-visibility--screen-reader) */
+      border: 0;
+      clip: rect(0, 0, 0, 0);
+      height: 1px;
+      overflow: hidden;
+      padding: 0;
+      position: absolute;
+      width: 1px;
+      // Workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1241631
+      word-wrap: normal;
```
line47, line 19: I think the only styles you need for `thead` to hide would be 
```
height: 0;
overflow: hidden;
position: absolute;
```

and the others could be removed here

packages/docs/src/styles/components/_ReactPropDocs.scss
```diff
@@ -0,0 +1,50 @@
...
+      word-break: break-word;
+    }
+
+    td:last-child {
+      border-bottom: 0;
+    }
+
+    td:first-child {
+      background-color: $color-gray-lightest;
+      border-top: 1px solid $color-gray;
```
line47, line 41: I think you can get rid of `border-top` for `td:first-child`

packages/docs/src/styles/components/_ReactPropDocs.scss
```diff
@@ -0,0 +1,50 @@
...
+    tr {
+      border: 1px solid $color-gray;
+      display: block;
+      margin: 0 0 0.5rem;
+    }
+
+    td {
+      border-top: 1px solid $color-gray-lightest;
+      display: block;
+      margin-top: -1px;
```
line47, line 31: I think you can remove the negative margin on `margin-top` and add a line above `border-top` for `border: none;` This gets rid of the negative margin needed and implements the borders a bit cleaner.

nichia: Thanks, that does look cleaner.

packages/docs/src/styles/components/_ReactPropDocs.scss
nichia: Presently, it's coded like the CMS DS Hide table headers (.ds-u-visibility--screen-reader) for accessibility. The suggested change works visually, will check screen-reader now.

davidakennedy:
@nichia Nice work on this! I tested with major screen reader and browser combinations and it's working as expected. Tested in:

- Safari + VoiceOver
- Firefox + NVDA
- IE 11 + JAWS
- Chrome + JAWS

The only small bit of feedback I'd add is that we might want to comment in the React code somewhere why we're adding roles to table elements. On the surface, it looks like an anti-pattern, but we know why we did it. But a passerby or our future selves might forget. 😄 

Great work on this! 🎉

nichia:
Good point - will add comment in the code regarding the Aria attributes.

packages/docs/src/scripts/components/ReactPropDoc.jsx
```diff
@@ -69,17 +69,38 @@ class ReactPropDoc extends React.PureComponent {
...
+        <td role="cell" headers="columnname">
+          <span className="docs_table__column-header ds-u-font-weight--bold" aria-hidden="true">
+            Name
+            <br />
+          </span>
           <code className="ds-u-font-weight--bold">{this.props.name}</code>
           {this.isRequired()}
         </td>
-        <td>
+        <td role="cell" headers="columntype">
```
bernardwang, line 83: @nichia @davidakennedy Sorry if I'm missing some context here, but whats the point of these hidden `span`s again? Is this unhidden when the table is collapsed?

packages/docs/src/scripts/components/ReactPropDoc.jsx
```diff
@@ -69,17 +69,38 @@ class ReactPropDoc extends React.PureComponent {
...
+          <span className="docs_table__column-header ds-u-font-weight--bold" aria-hidden="true">
```
bernardwang, line 84: @nichia @davidakennedy Sorry if I'm missing some context here, but whats the point of these hidden spans again? Is this unhidden when the table is collapsed?

PR #672, opened 2020-04-15
[WNMGGDS-396] Create linting scripts

bernardwang:
### Summary
- Moved stylelint gulp task to new packages
- Added eslint gulp task to new packages (replaces cli command that was previously used)
- Added a `--fix` option
- Removed old eslint and stylelint config packages

### How to test
- Make some linter errors in JS & CSS files in the source and docs directories
   - Easy CSS linter error is to remove new lines after each class definition
   - Easy JS linter error is to change a `const` to a `let`
- Run `yarn lint` and ensure all errors are reported
- Run `NODE_ENV=test yarn lint` and ensure the command errors out
- Run `NODE_ENV=test yarn lint --fix` and ensure the errors are automatically fixed and the command doesn't error

Child DS
- Download [child-design-system-example.zip](https://github.com/CMSgov/design-system/files/4519040/child-design-system-example.zip)
) and extract to same directory as the `design-system` repo
- Run through the same tests as above ^^ (there will be more errors because I didnt bother fixing them)


packages/design-system-scripts/gulp/lint.js
```diff
@@ -0,0 +1,94 @@
...
+      }))
+      .pipe(gulpIf(isFixed, gulp.dest(src)))
+      .pipe(gulpIf(process.env.NODE_ENV === 'test', eslint.failAfterError()))
+  );
+}
+
+module.exports = {
+  async lintDirectories(directories, fix) {
+    logTask('🔎 ', `Linting "src" directory in: ${directories.join(', ')}`);
+    // TODO: Fix failAfterError with streams
```
bernardwang, line 88: @pwolfert Our previous scripts configured linting to `failAfterError` when the env = test. I'm guessing this is mainly so that Travis CI will error when linting fails. 

When I test `failAfterError` on this branch, the linting task just stops for the directory its currently working on, and the task just completes early. I think this might have to do with the way we are promisifying streams. I was referring to [this](https://nodejs.org/api/stream.html#stream_stream_finished_stream_options_callback) but I havent been able to find a solution yet.

pwolfert: I just played around with this for a while, and I couldn't get it to throw an error either, and then I reread that documentation you linked to like the fourth time and realized that `error: false` is what we want here
```js
function streamPromise(stream) {
  return finished(stream, { readable: false, writeable: false, error: false });
}
```
At first I was thinking the default _true_ value is what we wanted.

pwolfert: Changing that line in `packages/design-system-scripts/gulp/common/streamPromise.js` made it throw the error and fail for me

bernardwang: Ahh thank you!

packages/design-system-scripts/package.json
```diff
@@ -32,9 +32,13 @@
     "enzyme": "3.10.0",
     "enzyme-adapter-react-16": "1.14.0",
     "enzyme-to-json": "3.3.5",
-    "eslint": "^4.18.1",
+    "eslint": "^6.8.0",
```
bernardwang, line 35: @pwolfert Should `eslint`, `stylelint`, `prettier` be peer deps? I believe they are also needed in the root package.json's `husky` precommit hooks. I originally tried moving these to peer deps and installing them to the root package.json, but [it didn't work easily](https://github.com/CMSgov/design-system/pull/672/commits/ef73463922e4acc6c0f445f18615a16eaebd6d65), maybe because the `design-systems-scripts` package is not located with the `node_modules` at the root.

pwolfert: I think they should probably be normal `dependencies` because the user (child design system developer) expects `cmsds lint` to just work. In the case of the core DS, we might need it for other things, but we don't force the child design system to use any linter if they don't want to (and we don't require them to delint before committing). That's my opinion at least

bernardwang: gotcha that makes sense! so if we dont want to include `husky` precommit hooks, do you think i should add `prettier` to the linter too? Right now our eslint and stylelint are configured to work with `prettier`

line47:
**For the core**
When i run `NODE_ENV=test yarn lint --fix` the items in the `design-system-docs` get fixed but the items in `design-system` do not.

For example, I removed a semicolon from `Alert.scss` and when running the command above the source of the scss file got replaced with 
```
[{"source":"/Users/scottweber/Documents/github/design-system/packages/design-system/src/components/Alert/Alert.scss","deprecations":[],"invalidOptionWarnings":[],"parseErrors":[],"errored":true,"warnings":[{"line":16,"column":10,"rule":"CssSyntaxError","severity":"error","text":"Missed semicolon (CssSyntaxError)"}]}]
```

I also modified `Alert.jsx` by changing a `const` to a `let` and it was not fixed.

Here is what the terminal read

```
NODE_ENV=test yarn lint --fix
yarn run v1.21.1
$ yarn cmsds lint packages/design-system packages/design-system-docs --fix
$ /Users/scottweber/Documents/github/design-system/node_modules/.bin/cmsds lint packages/design-system packages/design-system-docs --fix
[10:19:01] 🔎  Linting "src" directory in: packages/design-system, packages/design-system-docs
30 Sass files linted in packages/design-system-docs/src
34 JS files linted in packages/design-system-docs/src
118 Sass files linted in packages/design-system/src
[10:19:04] 

packages/design-system/src/components/Alert/Alert.scss
 16:10  ✖  Missed semicolon   CssSyntaxError




events.js:170
      throw er; // Unhandled 'error' event
      ^
Error: Failed with 1 error
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

line47:
**Running for child ds** 

When I run `yarn lint` it works for SCSS but I don't see any JS files getting linted. 
I also get the following errors:
```
/Users/scottweber/Documents/github/design-system/packages/design-system-scripts/node_modules/eslint/lib/cli-engine/cascading-config-array-factory.js:335
            throw error;
            ^

Error: Failed to load plugin 'jsx-a11y' declared in 'package.json » ./node_modules/@cmsgov/design-system-scripts/eslint.config.js': Cannot find module 'eslint-plugin-jsx-a11y'
Require stack:
- /Users/scottweber/Documents/github/child-design-system-example /__placeholder__.js
Referenced from: /Users/scottweber/Documents/github/design-system/packages/design-system-scripts/eslint.config.js
    at Function.Module._resolveFilename (internal/modules/cjs/loader.js:668:15)
    at Function.resolve (internal/modules/cjs/helpers.js:19:19)
    at Object.resolve (/Users/scottweber/Documents/github/design-system/packages/design-system-scripts/node_modules/eslint/lib/shared/relative-module-resolver.js:44:50)
    at ConfigArrayFactory._loadPlugin (/Users/scottweber/Documents/github/design-system/packages/design-system-scripts/node_modules/eslint/lib/cli-engine/config-array-factory.js:959:39)
    at ConfigArrayFactory._loadExtendedPluginConfig (/Users/scottweber/Documents/github/design-system/packages/design-system-scripts/node_modules/eslint/lib/cli-engine/config-array-factory.js:779:29)
    at ConfigArrayFactory._loadExtends (/Users/scottweber/Documents/github/design-system/packages/design-system-scripts/node_modules/eslint/lib/cli-engine/config-array-factory.js:725:29)
    at ConfigArrayFactory._normalizeObjectConfigDataBody (/Users/scottweber/Documents/github/design-system/packages/design-system-scripts/node_modules/eslint/lib/cli-engine/config-array-factory.js:660:25)
    at _normalizeObjectConfigDataBody.next (<anonymous>)
    at ConfigArrayFactory._normalizeObjectConfigData (/Users/scottweber/Documents/github/design-system/packages/design-system-scripts/node_modules/eslint/lib/cli-engine/config-array-factory.js:596:20)
    at _normalizeObjectConfigData.next (<anonymous>)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**Running `NODE_ENV=test yarn lint`**
I get the following
```
yarn run v1.21.1
warning package.json: No license field
$ yarn cmsds lint ./ ./docs
warning package.json: No license field
$ '/Users/scottweber/Documents/github/child-design-system-example /node_modules/.bin/cmsds' lint ./ ./docs
[10:36:45] 🔎  Linting "src" directory in: ./, ./docs
5 Sass files linted in docs/src
5 Sass files linted in src
[10:36:45] 

docs/src/example.scss
 3:29  ✖  Unexpected missing end-of-source newline   no-missing-end-of-source-newline

docs/src/pages/styles/_color-palette.docs.scss
 36:1  ✖  Expected no more than 1 empty line   max-empty-lines
 49:1  ✖  Expected no more than 1 empty line   max-empty-lines
 65:1  ✖  Expected no more than 1 empty line   max-empty-lines
 66:1  ✖  Expected no more than 1 empty line   max-empty-lines




events.js:170
      throw er; // Unhandled 'error' event
      ^
Error: Failed with 5 errors
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

**Running `NODE_ENV=test yarn lint --fix`**
Fixes up the scss files but doesn't fix the JS files

packages/design-system-scripts/package.json
bernardwang: I decided to include prettier in the lint scripts.

On a related topic though, @pwolfert how do you think we should handle `react` dependencies in the packages? `react` is a peer dep in `design-system`, but a normal dep in `design-system-docs` and `design-system-scripts`.

pwolfert: I think it should still be a peer dep for dependents (users) since we want them to be in control of which version of react they use, but for the child design systems, we it's probably okay to prescribe the version by making it a real dependency...or we could also leave it up to child design systems to decide, and then it's always a peer dependency and we make react a real dependency of the root of our design system.

bernardwang:
@line47 adding the `NODE_ENV=test` argument will cause the command to stop when the linter encounters an error. If you run the command without it, it should complete and fix everything

bernardwang:
@line47 I did some more testing and I realized that removing semicolon actually causes the SASS compiler to error. I don't know why this would cause the issue where the file is replaced with `[{"source":"/User...`, but the linting and fixing seems to work with other linting errors when the SASS compiles. You can see this by removing the new lines in between class definitions for example

packages/design-system-scripts/package.json
bernardwang: im ok with leaving it to the child DS to decide, and just make react and react-dom a peer dep across all the packages

bernardwang:
@pwolfert @line47 I pushed some commits, this PR should be ready for a final review.
The biggest changes in those commits: 
- Added `prettier` to the linting script
- Removed [hardcoded logic](https://github.com/CMSgov/design-system/pull/672/commits/bdc8e8bba1489c25b3cc36c915b91b1460d4a9c9#diff-0366342d617c082f1fbb8d39d5c13a25L58) that would limit linting inside a `src` directory
- Added a `ignorePatterns` option to replace that logic ^^ and provide a way to ignore files or directories

The main reasoning behind these changes was because I wanted to be able to lint the `design-system-scripts` directory as well, which doesn't have a `src` directory. These changes are intended to give more options over the linting script, as you can now lint any directory, and you can filter out directories/files with the `ignorePatterns` option

line47:
When I run `yarn lint` for the child design system example im getting an error.
```
Error: Failed to load plugin 'jsx-a11y' declared in 'package.json » ./node_modules/@cmsgov/design-system-scripts/eslint.config.js': Cannot find module 'eslint-plugin-jsx-a11y'
```

bernardwang:
@line47 that issue should be fixed in the latest commit!

bernardwang:
@line47 The error is from the `eslint` dependencies not being found, @pwolfert and I discussed it and we think the error is from `npm link`, meaning it should go away if we publish the `design-system-scripts` package and use it normally. 

I think we can confirm this after we publish the first beta version of `design-system-scripts`, and at the very least this issue only affects JS linting, so it's impact is relatively limited.

packages/design-system-scripts/package.json
bernardwang: Because of this PR: https://github.com/CMSgov/design-system/pull/721,
I think it makes sense to make eslint, stylelint and prettier peer deps

PR #673, opened 2020-04-16
[WNMGDS-432] Update sketch file to organize color palette

line47:
### Added
- Added Layer styles organized in the same way they are found on design.cms.gov 
- Added Color palette symbol for more consistent documentation within Sketch

### Changed
- Updated palette to look similar to design.cms.gov swatches
- Used new layer style background colors on elements like alerts and buttons and other backgrounds.

### Removed
- Removed symbols for each color and instead used Sketch layer styles

### Fixed
- There were 3 or 4 colors that were incorrect and had a slightly different hex value 

### Screenshots of layer style organization
Screen Shot 2020-04-16 at 4 06 53 PM
Screen Shot 2020-04-16 at 4 07 00 PM

### Screenshot of color swatch symbols for consistent updates
Screen Shot 2020-04-16 at 4 07 15 PM

### Screenshot of layer styles being used in alerts
Screen Shot 2020-04-16 at 4 07 43 PM

### Screenshot of color palette page
Screen Shot 2020-04-16 at 4 20 33 PM


line47:
@melwoodard that's a great point!  
I left those "others" in the layer styles as they don't pertain to the color palette and need further exploration. I plan to put some follow on work for general layer styles improvements.

PR #677, opened 2020-04-21
[WNMGDS-321] Update class references to use .classname format

line47:
Updated all places where a classname is used to use the `.CLASSNAME` format that inludes a `.`


packages/design-system-docs/src/scripts/components/HtmlExample.jsx
```diff
@@ -33,7 +33,7 @@ class HtmlExample extends React.PureComponent {
     return (
       <div className="c-markup__header">
         <h4 className="ds-u-font-size--h5 ds-u-margin-bottom--0">
-          Modifier: <code>{this.name()}</code>
+          Modifier: <code>.{this.name()}</code>
```
line47, line 36: Updated the HTML example so the `.` is always included.

line47:
Thanks for the review @nichia I've updated the PR 👍

nichia:
Have checked the PR and it's good to go.

packages/design-system-docs/src/example.js
```diff
@@ -31,14 +31,14 @@ function setModalExamples() {
   const dialogHeading = document.getElementById('dialog-heading');
   const dialogWrap = document.getElementById('dialog-wrap');
 
-  document.querySelectorAll('.dialog-open-btn').forEach((btn) => {
+  document.querySelectorAll('.dialog-open-btn').forEach(btn => {
```
bernardwang, line 34: @line47 A lot of these style changes should not have been included in this PR, as they go against our new linting rules. I'm not sure exactly why they are being included here (maybe your local dev env needs to be updated or reconfigured), but if you run `yarn lint` these changes should be undone.

line47: Hmmm yeah that is odd, I'll take a look. Thanks!

line47: Looks like these were introduced when I merged master into this branch 
https://github.com/CMSgov/design-system/pull/677/commits/61ebb45f4a799301ffe934ae0d4bdfd5266090b1

PR #682, opened 2020-04-23
[WNMGDS-354] React prop documentation table collapses at small screens

nichia:
This pull request replaces [[WNMGDS-354] Collapse React prop documentation table on small screens #671](https://github.com/CMSgov/design-system/pull/671). 

It pulls from the most recent master repo (CMS Design System V4) to avoid the branch conflicts of the original PR which was based off an older version of master repo.


packages/design-system-docs/src/styles/components/_ReactPropDocs.scss
```diff
@@ -0,0 +1,54 @@
+// Responsive docs table
```
line47, line 1: I'm thinking we should probably use display and responsive utility classes in the HTML, Then I don't think you'd need to use the custom SCSS because you could use utility classes. 

- https://design.cms.gov/utilities/display-visibility/#utilities.display-visibility.visibility
- https://design.cms.gov/utilities/display-visibility/#utilities.display-visibility.responsive

bernardwang:
@nichia @line47 Reposting my feedback from slack here:

Personally, I’d like to keep the borders in the prop table. I think it would be nice to keep it consistent with the collapsed version which uses borders. I also think if we do choose to use the borderless style, we should update it to match USWDS.
The USWDS docs say that a borderless tables should be used for table information to feel like a part or extension of text. To me, the grey background in the table header defeats that purpose of that and “implies” a border
Screen Shot 2020-04-22 at 10 22 23 AM

Another option would be to introduce a new table styling similar to [here](https://designsystem.digital.gov/design-tokens/#using-design-tokens), where theres an outer border but no lines dividing each column.
Screen Shot 2020-04-22 at 10 09 32 AM

packages/design-system-docs/src/styles/components/_ReactPropDocs.scss
bernardwang: There isn't a responsive utility for `ds-u-visibility--screen-reader`, so we would need custom SCSS anyway I believe. Maybe we could use the class via an extend though: 
`@extend .ds-u-visibility--screen-reader`

nichia: The `@extend .ds-u-visibility--screen-reader` would be ideal for the table header on small screens but extending selector only works within the same media context. A work around can be to explicitly create a mixin for the ds-u-visibility--screen-reader in the utility class and then we can import the mixin in the media query. For this ticket, i thought it best not to change the utility class.

bernardwang: ah thats right, you mentioned you cant extend in a media. i think what you have here looks good!

packages/design-system-docs/src/scripts/components/ReactPropDoc.jsx
```diff
@@ -69,17 +69,38 @@ class ReactPropDoc extends React.PureComponent {
   }
 
   render() {
+    // Specify ARIA roles attribute for table to ensure responsive HTML table is accessible to screen readers
     return (
-      <tr>
-        <td>
+      <tr role="row">
+        <td role="cell" headers="columnname">
+          <span className="docs_table__column-header ds-u-font-weight--bold" aria-hidden="true">
```
bernardwang, line 76: Down the road it would be nice to have this `aria-hidden` be dynamic, depending on if the text is hidden or not. This is good for now though!

bernardwang:
I did have one last thought on design though, on certain tables the lack of column dividers leaves a lot of white space. If we revisit the doc site table design in the future I would also be ok with reverting the full screen table design to the default for this reason. Again, this doesnt impact this PR at all, just putting it out for future consideration!
Screen Shot 2020-04-24 at 2 35 58 PM

PR #683, opened 2020-04-24
[NO-TICKET] Dropdown docs page was not finding dropdown example

line47:
The dropdown docs is looking for capital D Dropdown. 
https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/Dropdown/_Dropdown.docs.scss#L8

This update should fix that error


bernardwang:
I think the naming convention we have been using is lowercase for html examples. Maybe we should be updating the Dropdown doc file instead

line47:
Ah yeah - you are right! I was thrown off as all the other file names start with uppercase 🤔

PR #686, opened 2020-04-27
[WNMGDS-351] Remove determine props 'type' logic from ChoiceList component

nichia:
The ChoiceList component has logic that render `checkbox` fields or `radio` buttons for the user depending on certain props and conditions. This isn't very transparent as this logic is done on the program level. The change requires the explicit parameter of `checkbox` or `radio`. 

It defines component type as a required field, hence an explicit component type is required when using the ChoiceList component.  

### Changed
`<ChoiceList>`
- Change props `type` to a Required field
- Remove the logic that determines props `type` value
- Deprecate props `multiple` field and display deprecated message
- Modify example file to reflect the changes
- Modify test file to reflect the changes

`<Choice>`
- Change props `type` to a Required field
- Remove default props `type` value
- Modify `<MonthPicker>` to call `<Choice>` with argument `type="checkbox"`
- Modify example file to reflect the changes
- Modify test file to reflect the changes

### Deprecated
`multiple` props is deprecated, explicitly specify the `type` props instead.

### How to test
`<ChoiceList>`
- Go to Components "Checkbox & Radio" page - check react `<ChoiceList>` example frame functions correctly
- Check developer tools console for warning and deprecated messages -  to test this, modify ChoiceList.example.jsx to replace line 31 type="checkbox" with 'multiple'.:
```Warning: Failed prop type: The prop `type` is marked as required in `ChoiceList`, but its value is `undefined`.```
```[Deprecated]: Please remove the 'multiple' prop in <ChoiceList>, use type="checkbox" instead. This prop is deprecated and will be removed in a future release.```
- Check for warning message when a single 'radio' button is rendered. To test, modify ChoiceList.example.jsx to remove line 10 so that this test contains one `choices` element and type="radio". Check developer tools console for warning messages:
```[Warning]: Use type="checkbox" for components with only one choice. A single radio button is disallowed because it prevents users from deselecting the field.```

`<Choice>`
- Go to Components "Checkbox & Radio" page - check react `<Choice>` example frame functions correctly
- Go to Components "Month Picker" page - check the example functions correctly

### CODE REVIEW

**Comment by line47** [comment]
I was able to verify `Warning: Failed prop type: The prop `type` is marked as required in `ChoiceList`, but its value is `undefined`.
    in ChoiceList` by using the `multiple` prop

However, I was not able to see this console warning 
```[Deprecated]: Please remove the 'multiple' prop in <ChoiceList>, use 'type' instead. This prop is obsolete and will be removed in a future release.```

It looks like there are some conflicts to resolve as well

File: packages/design-system-docs/src/scripts/components/__tests__/ReactPropDoc.test.js [hunk 1]
```diff
@@ -40,13 +45,23 @@ describe('ReactPropDoc', () => {
   it('should render type', () => {
     const data = shallowRender();
 
-    expect(data.wrapper.find('td').at(1).text()).toBe(data.props.type.name);
+    expect(
```
bernardwang, line 48: Where these changes from the linter? What happens if you run `yarn lint`?

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -7,6 +7,15 @@ import classNames from 'classnames';
 export class ChoiceList extends React.PureComponent {
   constructor(props) {
     super(props);
+
+    if (process.env.NODE_ENV !== 'production') {
+      if (props.multiple) {
+        console.warn(
+          `[Deprecated]: Please remove the 'multiple' prop in <ChoiceList>, use 'type' instead. This prop is obsolete and will be removed in a future release.`
```
bernardwang, line 14: could you update this to say 
```
use `type="checkbox"` instead
``` 
Also please replace "obsolete" with "deprecated"

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -139,8 +147,7 @@ ChoiceList.propTypes = {
    */
   labelClassName: PropTypes.string,
   /**
-   * Allows the user to select multiple choices. Setting this to `true` results
-   * in a list of checkbox fields to be rendered.
+   * @hide-prop [Deprecated] This prop is obsolete after changing `type` to a required prop
```
bernardwang, line 150: Same here, please use deprecated language

packages/design-system/src/components/ChoiceList/_ChoiceList.docs.scss
```diff
@@ -15,7 +15,7 @@ Style guide: components.choice
 
 A `ChoiceList` component can be used to render a radio button group, or checkbox group.
 
-By default the component determines the type of field for you, taking into account accessibility and usability best practices. So, you can pass in an array of `choices` and let it determine what type of field would be best for the user, or alternatively you can manually pass in the `type` prop.
+Pass in the `type` prop to specify the type of field to display for the user.
```
bernardwang, line 18: Maybe this sentence can be combined with the one above, so something like: 

`A `ChoiceList` component can be used to render a radio button group, or checkbox group. Use the `type` prop to specify the type of field to display.`

packages/design-system-docs/src/scripts/components/__tests__/ReactPropDoc.test.js
nichia: Scott - thanks for testing. I've removed the `multiple` props from the example file. To test this, modify `ChoiceList.example.jsx` to replace line 31 `type="checkbox"` with `multiple`.

Bernard - these are my changes, I'll need to go and resolve the conflicts. I've changed `toBe` with `toMatch`.

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
nichia: Noted, have changed the wording to deprecated.

packages/design-system/src/components/ChoiceList/_ChoiceList.docs.scss
nichia: Thanks, that does read better.

bernardwang:
I noticed the `type()` function is still present, could you remove this as well?
https://github.com/CMSgov/design-system/pull/686/files#diff-80ef1a0df24d4c812a5dd9612347dd62L41

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -7,6 +7,15 @@ import classNames from 'classnames';
 export class ChoiceList extends React.PureComponent {
   constructor(props) {
     super(props);
+
+    if (process.env.NODE_ENV !== 'production') {
+      if (props.multiple) {
```
bernardwang, line 12: Could you also add a console warning to handle the case for `this.props.choices.length === 1`? The message should be something like: "Use type=checkbox for <ChoiceList> components with only one choice. A single radio button is disallowed because it prevents users from deselecting the field."

nichia:
I originally removed this function, but after further discussion with Scott on deprecating props, i added this function back. Reason for keeping it in the meantime is to act as safety net for existing apps using the DS. Initial phase is to log messages and still support existing usage. Next phase, say in a couple of months time, we will disable this `multiple` field usage and remove the type() logic.
Example - apps that presently use the `multiple` prop and do not specify the `type` prop of checkbox or radio, a warning message `type` required field and deprecated `multiple` messages will be logged on devl tools, and their app should still function normally. The next phase would be just removing this type() logic and to only refer `type` prop in rendering the fields.

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
nichia: Thanks, that is a good point. Will add that.

bernardwang:
@nichia @line47 Ah I see. Personally think it would be acceptable to just remove the logic. It's a pretty simple `prop` search and replace, and we are already doing a major bump for the next release, this could just be included in our release notes as a breaking change. Plus, if we wanted to avoid breaking changes altogether we would technically also avoid making `type` required, as that will cause any `ChoiceList` without `type` to error on compile. Removing the `type()` logic is just as much of a breaking change as is making `type` required.

bernardwang:
I just noticed, `<Choice>` also has a `type` prop. I think we should make this consistent with `<ChoiceList>` and make the `type` required, and remove the default.

nichia:
Thanks for pointing that out. Yes, I noticed the <Choice> has prop `type` with default of `checkbox`, and that <MonthPicker> called it and did not set any prop `type`. Should have check back with you then about it.  I'll now update <Choice> to change prop `type` to a required field, to stay consistent with <ChoiceList>. Will update <MonthPicker> and any other associated tests/examples to reflect this change.

PR #687, opened 2020-04-28
[WNMGDS-428]Add default button inverse and transparent button inverse example in docs

nichia:
Add Default and Transparent buttons for inverse example to simplify the page similar to the USWDS.

### Changed
- Remove primary, success and danger buttons from inverse example
- Add Default and Transparent examples with Inverse setting 
- Move Size variant buttons to end of example


packages/design-system/src/components/Button/button.example.html
```diff
@@ -75,6 +75,53 @@ <h6 class="preview__label">Transparent button</h6>
 <button type="button" class="ds-c-button ds-c-button--transparent ds-c-button--focus">Focus</button>
 <button type="button" class="ds-c-button ds-c-button--transparent" disabled>Disabled</button>
 
+<h6 class="preview__label">Default button inverse</h6>
```
bernardwang, line 78: It looks like the ordering was changed here, the inverse buttons should go last

packages/design-system/src/components/Button/button.example.html
```diff
@@ -75,6 +75,53 @@ <h6 class="preview__label">Transparent button</h6>
...
+  </button>
+  <button type="button" class="ds-c-button ds-c-button--focus ds-c-button--inverse">
+    Focus
+  </button>
+  <button type="button" class="ds-c-button ds-c-button--inverse" disabled>
+    Disabled
+  </button>
+</div>
+
+<h6 class="preview__label">Transparent button inverse</h6>
```
bernardwang, line 97: Could we combine these inverse buttons to the use same `example--inverse` container? I know it gets a bit tricky because there are 2 sets of inverse buttons with 2 labels, but this is the only component example with multiple inverse sections.

Maybe we could have 1 inverse container and 1 label thats just "Inverse buttons"? I know its not as descriptive but I think it's acceptable as the label would be repetitive anyway. @line47 thoughts?
Screen Shot 2020-04-29 at 10 49 18 AM

PR #691, opened 2020-04-29
Docs: Improve guidance for form validation

### LINKED ISSUE #690: Docs: Improve guidance for form validation
**Description**

I'd like to see more guidance around how validation can be handled. In the design system itself, we leave this to the teams working with it as to how to handle things. I'd like to add a few key guidelines to help teams make better decisions.

Some areas I'd like to add points about:
- Display errors without relying on color alone.
- How to communicate those errors to assistive technology.

Some good prior art out there:
https://design.va.gov/patterns/forms#error-messages
https://design-system.service.gov.uk/components/error-message/

davidakennedy:
## Added
Added some bullet points to the form validation docs around:
* Display errors without relying on color alone.
* How to communicate those errors to assistive technology.

Closes #690


line47:
@davidakennedy OK if I resolve the conflict on this?

davidakennedy:
@line47 Go for it!

packages/design-system-docs/src/pages/components/FormLabel/_FormLabel.docs.scss
```diff
@@ -59,6 +59,10 @@ Style guide: components.form-label.react
 
 - Place inline validation messages within the field's `<label>` element.
 - Visually align inline validation messages with the input fields, so people using screen magnifiers can read them quickly.
+- Either place an icon next to the error, labeling the icon as error with ARIA or off-screen text, or use text in front of the message itself. Like: "Error: The email address is not a valid email address." The use of the icon or text not only clearly labels the following text as an error, but allows users to know that the message is an error without relying on color alone.
```
line47, line 62: I'm thinking we might want to add a follow on issue to update our examples that show errors to show them in a better way. For example, we don't sow good error examples here https://design.cms.gov/components/choice/

I'm also wondering if we should update the forms page with these same bullets?
https://design.cms.gov/guidelines/forms/

davidakennedy: > I'm thinking we might want to add a follow on issue to update our examples that show errors to show them in a better way.

Yeah, I had that next on my list. I can create an issue.

> I'm also wondering if we should update the forms page with these same bullets or link to the validation section of the forms page instead of listing the content in 2 places?

That's a good point. I saw those, and went back and forth on the best place. I put it here because I thought it was better to have the guidance at the component level. But that may break down when we have multiple components needing the same guidance. I think it's better not to duplicate. It may be better to just put it on the forms page, and/or if the content builds up over time, have a separate page for error messages.

bernardwang: Hm, this is tricky because some of these bullets seem specific to the FormLabel (The error message should be wrapped in an `role="alert"`), while others are more general. I think overall this guidance could live in the Forms guidelines, but it would be nice to eventually have a [Validation component similar to USWDS](https://designsystem.digital.gov/components/form-controls/#validation).

bernardwang: USWDS also has a [form template section](https://designsystem.digital.gov/components/form-templates/) which is really nice. I'm not sure if we can do this, but it might be nice to create a more complex form example and to put in it the Patterns section?

line47: I like the idea of having a more complex form example in the patterns

Issue #565, opened 2019-12-06
Add ability to pass autocomplete props into date field's individual text fields


sawyerh:
I'd love to use the `<DateField>` React component and also set `autoComplete` prop on each of the individual text fields (month, day, year). The main use case would be for birthdates, which would use the `bday-day`, `bday-month`, and `bday-year` [autocomplete values](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete#Values).

bernardwang:
Tracked here: https://jira.cms.gov/browse/WNMGDS-350

bernardwang:
What were you thinking of for the `autoComplete` prop? An object that sets the 3 text fields? Or a single boolean to control all text fields?
```
autoComplete = {
   day: true,       // sets `bday-day` autocomplete value
   month: true,  // sets `bday-month` autocomplete value
   year: false,    // sets `bday-year` autocomplete value
}
autoComplete = true;   // sets all autocomplete values
```

sawyerh:
@bernardwang Between those two options, a single boolean prop feels better. I'm also wondering whether something like `dayAutocomplete`, `monthAutocomplete`, etc would be the most flexible, but don't feel too strongly since birthdate is currently the only autocomplete'able date (besides credit cards, but that's just month and year).

line47:
Tracked here https://jira.cms.gov/browse/WNMGDS-350

PR #693, opened 2020-05-06
[WNMGDS-350] Add autoComplete prop to Date field component

nichia:
Add autoComplete prop to Date field component to enable autofill of birthday day, month and year.

### Changed
- New `autoComplete` prop to set the `autocomplete` attribute on the date fields to `bday-day`, `bday-month` and `bday-year`.

### How to test
Go to Components "Date field" page and check React props table for `autoComplete` prop

[https://github.com/CMSgov/design-system/issues/565]
[https://design-system.service.gov.uk/components/date-input/]

### DIFF
--- a/packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
+++ b/packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
@@ -44,6 +44,20 @@ ReactDOM.render(
       yearInvalid
     />
     <ControlledDateField />
+    <fieldset className="ds-u-margin-top--3 ds-u-padding-left--3 ds-u-padding-bottom--3 ds-u-border--2">
+      <legend className="ds-u-font-size--h3">Autocomplete Example</legend>
+      <DateField
+        label={
+          <span>
+            DateField using <code>autoComplete</code> for a date of birth
+          </span>
+        }
+        autoComplete
+        monthName="dob-month"
+        dayName="dob-day"
+        yearName="dob-year"
+      />
+    </fieldset>
   </Fragment>,
   document.getElementById('js-example')
 );
--- a/packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
+++ b/packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
@@ -44,20 +44,6 @@ ReactDOM.render(
       yearInvalid
     />
     <ControlledDateField />
-    <fieldset className="ds-u-margin-top--3 ds-u-padding-left--3 ds-u-padding-bottom--3 ds-u-border--2">
-      <legend className="ds-u-font-size--h3">Autocomplete Example</legend>
-      <DateField
-        label={
-          <span>
-            DateField using <code>autoComplete</code> for a date of birth
-          </span>
-        }
-        autoComplete
-        monthName="dob-month"
-        dayName="dob-day"
-        yearName="dob-year"
-      />
-    </fieldset>
   </Fragment>,
   document.getElementById('js-example')
 );
--- a/packages/design-system/src/scripts/components/DateField/DateField.jsx
+++ b/packages/design-system/src/scripts/components/DateField/DateField.jsx
@@ -170,7 +170,7 @@ DateField.defaultProps = {
 
 DateField.propTypes = {
   /**
-   * Controls autocomplete attributes `bday-day`, `bday-month` and `bday-year` 
+   * Adds `autocomplete` attributes `bday-day`, `bday-month` and `bday-year` to the corresponding `<DateField>` inputs
    */
   autoComplete: PropTypes.bool,
   /**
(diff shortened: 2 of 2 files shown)


packages/design-system/src/scripts/components/DateField/DateField.jsx
```diff
@@ -166,6 +169,10 @@ DateField.defaultProps = {
 };
 
 DateField.propTypes = {
+  /**
+   * Controls autocomplete attributes `bday-day`, `bday-month` and `bday-year`
```
bernardwang, line 173: Maybe we could reword this? Something like:
```
Adds `autocomplete` attributes `bday-day`, `bday-month` and `bday-year` to the corresponding `<DateField>` inputs
```
```

File: packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx [hunk 2]
```diff
@@ -44,6 +44,20 @@ ReactDOM.render(
       yearInvalid
     />
     <ControlledDateField />
+    <fieldset className="ds-u-margin-top--3 ds-u-padding-left--3 ds-u-padding-bottom--3 ds-u-border--2">
```
bernardwang, line 47: I almost wonder if we could just not include a new example, as its currently hard to test anyway. Maybe we could add a new example in the future when we get confirmation from UKgov on how to test?

nichia: That'a a good point. We'll wait for confirmation form UKgov on how to acceptably test this feature before adding an example.

packages/design-system/src/scripts/components/DateField/DateField.jsx
```diff
@@ -114,6 +114,7 @@ export class DateField extends React.PureComponent {
             name={this.props.monthName}
             value={this.props.monthValue}
             aria-describedby={labelId}
+            autoComplete={this.props.autoComplete && 'bday-month'}
```
line47, line 117: I'm thinking using something more generic for the prop names would serve more use cases in the future.
Something like `dayAutocomplete`, `monthAuthocomplete`, `yearAutocomplete`.

nichia: Date field autocomplete is currently available for birthdate and credit card month & year only and the change as of right now only applies for birth day, month and year. 

To be more generic and caters for autocomplete of birth date, credit card date and future cases, your suggestion makes sense. So if props are defined for `dayAutocomplete`, `monthAutocomplete` and `yearAutocomplete`, then when using `Date field`, these props can be set to autocomplete for any token depending on use cases. 

Example: 
- `yearAutocomplete='bday-year' ` to autocomplete for birth year
 - `yearAutocomplete='cc-exp-year'` to autocomplete for credit card year.

packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
bernardwang: Just curious, how did you end up reaching out to UKgov?

packages/design-system/src/scripts/components/DateField/DateField.jsx
bernardwang: yeah, I touched on this with Sawyer in the [original issue](https://github.com/CMSgov/design-system/issues/565#issuecomment-574250780). I think since the only autocomplete case that we can currently anticipate is for birthdays (credit cards don't need the full month/date/year), it's easier on the user to just have a single `autoComplete` prop. I can't think of a case where a user would want to have autocomplete only on the year for example, so I dont think we need to provide that level of granularity.

packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
nichia: I emailed using this link https://design-system.service.gov.uk/get-in-touch/ 
and have received a reply which I'm following up on.

packages/design-system/src/scripts/components/DateField/DateField.jsx
nichia: Yes, we'll have simplicity with just one autcomplete prop

packages/design-system/src/scripts/components/DateField/DateField.jsx
```diff
@@ -166,6 +169,10 @@ DateField.defaultProps = {
 };
 
 DateField.propTypes = {
+  /**
+   * Adds `autocomplete` attributes `bday-day`, `bday-month` and `bday-year` to the corresponding `<DateField>` inputs
+   */
+  autoComplete: PropTypes.bool,
```
bernardwang, line 175: @nichia I was looking over this and I realized we probably should have used the name `birthdayAutocomplete` instead of `autocomplete`....

PR #697, opened 2020-05-13
[WNMGDS-466] Update Datefield html example

nichia:
Update Date field html example to include the autocomplete attributes for birthday.


bernardwang:
LGTM

PR #704, opened 2020-05-22
[WNMGDS-483] Fix currency mask for Firefox

### LINKED ISSUE #702: TextField with 'currency' mask shows error when data contains commas
**Describe the bug**
We're using the text field and when the data has commas the border goes red and the hover message is "Please match the requested format".

**To Reproduce**
1. Go to a text field that has a "currency" mask set.
2. enter a large number that contains commas
3. See the red border.
4. hover over the red border and see "Please match the requested format"

**Expected behavior**
Text fields with currency mask should accept values with commas

**Screenshots**

Screen Shot 2020-05-21 at 12 56 27 PM (2)

 - OS: MacOS Catalina 10.15.4
 - Browser Firefox
 - Version 76.0.1

line47:
### Changed
- Changed the regex for the currency mark pattern to be `^[0-9,]+(.[0-9]{1,2})?$`

### Removed
- Removed pattern attribute from examples as it is defined and added from the Mask.jsx 

### Fixed
- This fixes #702 

**How to test**
- Test the regex here - you can test here https://regex101.com/r/TEwmdB/1
- Run the site locally and open up firefox and navigate to the masked field page
    - Enter numbers in the currency field and ensure that no browser error is triggered

### DIFF
--- a/packages/design-system-docs/src/pages/components/TextField/Mask.example.jsx
+++ b/packages/design-system-docs/src/pages/components/TextField/Mask.example.jsx
@@ -12,7 +12,7 @@ class ControlledCurrencyField extends React.PureComponent {
     super(props);
 
     this.state = {
-      currencyValue: '2,500',
+      currencyValue: '2500',
     };
   }
 
--- a/packages/design-system-docs/src/pages/components/TextField/Mask.example.jsx
+++ b/packages/design-system-docs/src/pages/components/TextField/Mask.example.jsx
@@ -12,7 +12,7 @@ class ControlledCurrencyField extends React.PureComponent {
     super(props);
 
     this.state = {
-      currencyValue: '2500',
+      currencyValue: '2,500',
     };
   }
 
@@ -56,11 +56,10 @@ const Example = () => {
         label="Currency"
         mask="currency"
         inputMode="numeric"
-        pattern="[0-9]*"
         type="text"
         name="currency_example"
         onBlur={(evt) => handleBlur(evt, 'currency')}
-        defaultValue="2500"
+        defaultValue="2,500"
       />
 
       <TextField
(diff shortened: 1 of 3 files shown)


line47:
I need to update snapshots for Travis to be happy however I don't think updating snapshots is working. `yarn update-snapshots` gives the error 
```
$ NODE_ENV=test bash scripts/jest.sh -u
bash: scripts/jest.sh: No such file or directory
```

nichia:
I think we can also just add the `,` to currency maskPattern `currency: '[0-9.,-]*'`

line47:
Updated the pattern and the snapshot. Thanks @nichia

packages/design-system-docs/src/pages/components/TextField/Mask.example.jsx
```diff
@@ -76,7 +75,6 @@ const Example = () => {
         label="Social security number (SSN)"
         mask="ssn"
         inputMode="numeric"
```
bernardwang, line 77: Why are these patterns being removed for ssn and zip? They do look like they should be updated to match the values in `Mask.jsx` though

packages/design-system-docs/src/pages/components/TextField/Mask.example.jsx
```diff
@@ -12,7 +12,7 @@ class ControlledCurrencyField extends React.PureComponent {
     super(props);
 
     this.state = {
-      currencyValue: '2500',
+      currencyValue: '2,500',
```
bernardwang, line 15: The state should still contain the unmasked value which has no comma

packages/design-system-docs/src/pages/components/TextField/Mask.example.jsx
line47: I found in my testing that the pattern attribute gets added automatically and doesn't need to be part of the example. Setting the `TextField` mask to="ssn' automatically sets the pattern to `[0-9-*]*` Mask.jsx

bernardwang: Ah that makes sense thank you.

PR #705, opened 2020-05-26
[WNMGDS-444] Fix dev console warning msg for VerticalNav

nichia:
This PR fixed the VerticalNav warning message shown on developer console. 
I was not able to reproduce the other issues with failed tests on running `yarn test` and `yarn test:e2e`.
 
### Changed
- Changed items PropTypes to check for object type instead of `shape` of object type

### How to test
- Run the design system site locally and check the developer console to ensure that no warning message is displayed for VerticalNav

### DIFF
--- a/packages/design-system/src/components/VerticalNav/VerticalNav.jsx
+++ b/packages/design-system/src/components/VerticalNav/VerticalNav.jsx
@@ -73,7 +73,7 @@ VerticalNav.propTypes = {
   /**
    * An array of `VerticalNavItem` data objects
    */
-  items: PropTypes.arrayOf(PropTypes.shape(PropTypes.object)).isRequired,
+  items: PropTypes.arrayOf(PropTypes.object).isRequired,
   /**
    * Indicates this list is nested within another nav item.
    */


line47:
I no longer see the VerticalNav error when running locally 🎉

packages/design-system/src/components/VerticalNav/VerticalNav.jsx
```diff
@@ -73,7 +73,7 @@ VerticalNav.propTypes = {
   /**
    * An array of `VerticalNavItem` data objects
    */
-  items: PropTypes.arrayOf(PropTypes.shape(PropTypes.object)).isRequired,
```
bernardwang, line 76: Could you try and use this: 
`items: PropTypes.arrayOf(PropTypes.shape(VerticalNavItem.propTypes)),`
similar to `VerticalNavItem` proptype

nichia: That change will work on the site too - however, running`yarn test` with this will result in a VerticalNavItem error:

```
 FAIL  packages/design-system/src/components/VerticalNav/VerticalNavItem.test.jsx
  ● Test suite failed to run

    TypeError: Cannot read property 'propTypes' of undefined

      74 |    * An array of `VerticalNavItem` data objects
      75 |    */
    > 76 |   items: PropTypes.arrayOf(PropTypes.shape(VerticalNavItem.propTypes)),
```

bernardwang: ok this works for me then!

PR #706, opened 2020-05-26
[WNMGDS-65] Add table component

nichia:
### Added
Create an accessible table component featuring:
- Horizontal scrolling when the width of the table exceeds the screen width (for keyboard support make scrollable element focusable)
- Responsive table by collapsing to a vertically stacked layout on small viewport
- Striped table by highlighting alternating rows to help increase readability
- Default column width

#### Help wanted: Please review the table components and leave comments

Usage (see core DS react props table & table components for examples): 
```
<Table {attributes}>
    <TableCaption {attributes}>
          {caption contents}
    </TableCaption>
    <TableHead>
          <TableRow>
            <TableHeaderCell {attributes} />
			~
            <TableHeaderCell {attributes} />
          </TableRow>
    </TableHead>
    <TableBody>
      	<TableRow>
            <TableDataCell {attributes} />
                        ~
            <TableDataCell {attributes} />
         </TableRow>
  </TableBody>
</Table>
```
### Building of table components inspired by:
- [Medicare.gov Plan compare table](https://www.medicare.gov/plan-compare/#/plan-details/2020-S4802-083-0?year=2020&lang=en#costs-by-drug-tier)
- [Australian government design system](https://designsystem.gov.au/components/table) 
- [Inclusive components design by Heydon Pickering](https://inclusive-components.design/data-tables/) 
- [Multi-header data table by Trevor Pierce](https://codepen.io/tpierce_402/pen/YzzqwmZ)
- [How to build accessible & responsive HTML table by Dominic Mercier](https://medium.com/@d.mercier84/how-to-build-accessible-responsive-html-table-a930386485c4)
- [A responsive accessible table by Adrian Roselli](https://adrianroselli.com/2017/11/a-responsive-accessible-table.html#ResponsiveScrollingKeyboard)


packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -0,0 +1,95 @@
...
+    <TableRow>
+      {simpleHeaders.map((header, index) => {
+        return (
+          <TableHeader
+            key={header.key}
+            title={header.title}
+            type={header.type}
+            width={header.width}
+            scope="col"
+            id={tableId + '_col_' + index}
```
bernardwang, line 54: I believe `uniqueId` should be called for each table header, something like this:
```
id={uniqueId('tbl_col_')}
```
Same goes for the uses of `tableId` below

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
 Style guide: components.table.complex
 */
 
+/*
+`<Table>`
+
+`This is a beta version of the Table components. The functionality and usage is subject to change in response to the feedback from design system users.`
```
bernardwang, line 26: I don't think its necessary to say its beta here, instead use the `@status` modifier. This is hidden pretty deep in our documentation and there arent any examples so dont worry.

Adding `@status Beta` should do the trick here.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
...
+
+@react-props Table.jsx
+
+Style guide: components.table.react
+*/
+
+/*
+`<TableBody>`
+
+A `TableBody` react component.
```
bernardwang, line 39: I don't think these default descriptions are necessary

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
 Style guide: components.table.complex
 */
 
+/*
+`<Table>`
+
+`This is a beta version of the Table components. The functionality and usage is subject to change in response to the feedback from design system users.`
+A `Table` component renders data table using semantic markup and ARIA. The table components allows for customization as needed by the users.
```
bernardwang, line 27: I can help you on this, but we probably want to give more guidance on how our react table works, as well as the different responsive considerations and options.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
bernardwang: Maybe instead of the template "A [react component name] react component", we could more specific description about what HTML element it corresponds to and how it should be used? Either way I imagine this should be brief.

packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx
```diff
@@ -19,35 +27,22 @@ class ReactPropDocs extends React.PureComponent {
...
-                Description
-              </th>
-            </tr>
-          </thead>
-          <tbody>{this.rows()}</tbody>
-        </table>
-      </div>,
+      <Table key="propDocsTable" stacked="sm" scrollTable>
+        <TableCaption className="ds-u-padding--1 ds-u-visibility--screen-reader">
+          React Properties Documentation
```
bernardwang, line 34: I don't think a caption here is necessary considering we have the `h3` saying props.

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,140 @@
...
+  scrollTable,
+  scrollableCaption,
+  children,
+  ...attributeOptions
+}) => {
+  const container = useRef(null);
+  const captionID = useRef('caption-' + Math.random().toString(36).substr(2, 9));
+  const [tabIndex, setTabIndex] = useState(null);
+
+  useEffect(() => {
```
bernardwang, line 31: We should definitely leave comments explaining this logic and the reasoning for it (i.e. to show a message when scrollbar appears).

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,140 @@
...
+    if (scrollTable) {
+      const { scrollWidth, clientWidth } = container.current;
+      const scrollable = scrollWidth > clientWidth;
+      setTabIndex(scrollable ? '0' : null);
```
bernardwang, line 35: Is this still necessary? I'm not entire sure what this function does but I thought we don't need to set tab index anymore

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const renderScrollableCaption = () => {
+    const isScrollable = tabIndex === '0';
+    return scrollTable && isScrollable ? scrollableCaption : null;
+  };
+
+  return (
+    <>
+      {renderScrollableCaption()}
+      <div ref={container} tabIndex={tabIndex} {...attributeScrollTable}>
```
bernardwang, line 91: Is this wrapper `div` always necessary, even if `scrollTable` is false?

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,140 @@
...
+  className,
+  stacked,
+  striped,
+  scrollTable,
+  scrollableCaption,
+  children,
+  ...attributeOptions
+}) => {
+  const container = useRef(null);
+  const captionID = useRef('caption-' + Math.random().toString(36).substr(2, 9));
```
bernardwang, line 28: What is this logic for? Is this for generating a random id?

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,140 @@
...
+
+  const isTableCaptionComponents = (child) => {
+    return child && child.type === TableCaption;
+  };
+
+  const renderChildren = (captionId) => {
+    return React.Children.map(children, (child) => {
+      if (scrollTable && isTableCaptionComponents(child)) {
+        // Extend props on tables before rendering.
+        return React.cloneElement(child, {
```
bernardwang, line 74: Do you mind elaborating on this logic? Why do we need to apply the `captionId` to all the table children?

packages/design-system/src/components/Table/TableBody.jsx
```diff
@@ -0,0 +1,30 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import classNames from 'classnames';
+
+export const TableBody = ({ children, className, ...attributeOptions }) => {
```
bernardwang, line 5: Is `attributeOptions` just any additional props? Usually I've seen this named `other` or named the component like `tableBodyProps`

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,140 @@
...
+   */
+  stacked: PropTypes.oneOf(['sm', 'md', 'lg']),
+  /**
+   * A striped variation of the table.
+   */
+  striped: PropTypes.bool,
+  /**
+   * Horizontal scroll table.
+   */
+  scrollTable: PropTypes.bool,
```
bernardwang, line 133: Maybe we can rename `scrollTable` to `scrollable`?

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,140 @@
...
+  /**
+   * Additional scrollable text/node to display
```
bernardwang, line 135: Consider adding more detail:

```
Additional text or content to display when the horizontal scrollbar is visible. This prop will only be used when the `scrollable` prop is also set.
```

packages/design-system/src/components/Table/Table.jsx
bernardwang: More detail would be helpful here too:
```
Applies a horizontal scrollbar when the `Table` contents exceed the container width. It's recommended to use the `scrollableCaption` prop with `scrollable`.

File: packages/design-system/src/components/Table/TableCell.jsx [hunk 14]
```diff
@@ -0,0 +1,84 @@
...
+   */
+  children: PropTypes.node,
+  /**
+   * Additional classes to be added to the table cell element.
+   */
+  className: PropTypes.string,
+  /**
+   * The table cell data.
+   */
+  data: PropTypes.node.isRequired,
```
bernardwang, line 65: What's the difference between `data` and `children`?

packages/design-system/src/components/Table/TableCell.jsx
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * The `headers` attribute contains a list of `id` attributes of the associated data. If there is more than one id, they are separated by spaces.
+   */
+  headers: PropTypes.string,
```
bernardwang, line 69: Is this prop used in this component?

packages/design-system/src/components/Table/TableCell.jsx
```diff
@@ -0,0 +1,84 @@
...
+  /**
+   * Additional classes to be added to the stacked Title element.
+   */
+  stackedClassName: PropTypes.string,
+  /**
+   * The stacked row title for responsive table accessiblity.
```
bernardwang, line 75: To clarify, this should basically be required if the parent `Table` has `stacked= true` right?

packages/design-system/src/components/Table/TableHeader.jsx
```diff
@@ -0,0 +1,97 @@
...
+      {...attributeOptions}
+    >
+      {renderStackedTitle()}
+      {title}
+    </th>
+  );
+};
+
+TableHeader.defaultProps = {
+  className: '',
```
bernardwang, line 56: I don't think any of these default `className` or `stackedClassName` props are necessary.

packages/design-system/src/components/Table/TableBody.jsx
nichia: `attributeOptions` refers to any other attributes that get added to TableBody which have be picked up by the destructuring assignment from the prop objects.

packages/design-system/src/components/Table/TableCell.jsx
```diff
@@ -0,0 +1,84 @@
...
+  type,
+  ...attributeOptions
+}) => {
+  const classes = classNames(
+    'ds-c-table__cell',
+    className,
+    type === 'numeric' ? 'ds-c-table__cell--numeric' : null
+  );
+
+  const stackedClasses = classNames(
```
bernardwang, line 20: What do you think about moving this definition into `renderStackedTitle`, as thats the only place these are used?
Something like this:
```
const renderStackedTitle = () => {
   if (stackedTitle && stackedTitle.length > 0) {
       const stackedClasses = classNames(...);
       return (
           <span aria-hidden="true" className={stackedClasses}>...</span>
       )
   }
}

File: packages/design-system/src/components/Table/Table.jsx (hunk 10 above)
Comment (nichia) [reply packages/design-system/src/components/Table/Table.jsx:74]: Need to apply `captionId` to Table Caption only so that the tabIndex can be tied back to Caption for accessibility

Comment (nichia) [reply packages/design-system/src/components/Table/Table.jsx:74]: It only looks for TableCaption :
```
const isTableCaptionComponents = (child) => {
    return child && child.type === TableCaption;
  };
```

File: packages/design-system/src/components/Table/Table.jsx (hunk 9 above)
Comment (nichia) [reply packages/design-system/src/components/Table/Table.jsx:28]: Yes, we need a unique number generated for the captionID.

File: packages/design-system-docs/src/pages/components/Table/_Table.docs.scss (hunk 2 above)
Comment (line47) [reply packages/design-system-docs/src/pages/components/Table/_Table.docs.scss:26]: it looks like we have the following status available: `Draft` and `Ready` and the `@status Draft` needs to be added to the top of the page in order to render unless we update the `PageBlock` script to allow for @status to be used in page sections.

File: packages/design-system/src/components/Table/Table.jsx (hunk 8 above)
Comment (nichia) [reply packages/design-system/src/components/Table/Table.jsx:91]: Good pick up, it's not necessary when `scrollTable` is false

File: packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx (hunk 5 above)
Comment (line47) [reply packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx:34]: I'd like to see us using the `caption` instead of the `h3` as it keeps the component tighter together and more relevant to the content. An `h3` could lose the hierarchy on the page or need to be changed while a caption is directly tied to the table. 

I'd be curious as to what @davidakennedy thinks here

File: packages/design-system/src/components/Table/Table.jsx (hunk 7 above)
Comment (nichia) [reply packages/design-system/src/components/Table/Table.jsx:35]: Not exactly sure.
From https://inclusive-components.design/data-tables/ - it talks about horizontal scroll needs tabIndex for keyboard support. Then to give screen reader context of the tabIndex focus, one way is to tie this `TabIndex=='0' `to the Caption via `captionID`.
@davidakennedy - would you be able to help confirm which way to go for this?

File: packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx (hunk 5 above)
Comment (nichia) [reply packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx:34]: If we set `scrollTable` true, this is needed to tie back to `tabIndex`

File: packages/design-system-docs/src/pages/components/Table/Table.example.jsx (hunk 1 above)
Comment (nichia) [reply packages/design-system-docs/src/pages/components/Table/Table.example.jsx:54]: This `id` need to correspond to `header` - so need to just use one tableId - but just noticed that `uniqueId` may not work for us here for cases when we have multiple tables on the page - so will need to change to probably a random math unique id generation used for captionID

File: packages/design-system/src/components/Table/TableHeader.jsx [hunk 19]
```diff
@@ -0,0 +1,97 @@
...
+  /**
+   * The title of table header/column.
+   */
+  title: PropTypes.PropTypes.node.isRequired,
+  /**
+   * Type of the header, can be either text or numeric for left or right alignment respectively.
+   */
+  type: PropTypes.oneOf(['text', 'numeric']),
+  /**
+   * The width of the header.
```
bernardwang, line 92: I think we should elaborate more in this comment here and mention these are percentages of the width container.
```
Sets the width of `TableHeader` to a percentage of the `Table` width
```

packages/design-system/src/components/Table/TableCell.jsx
nichia: For complex table with multi-headers table or irregular header columns, users need to assign an `id` to each column heading and tie that `id` to the respective cell via the Cell `headers`
see here for examples: https://codepen.io/tpierce_402/pen/YzzqwmZ

packages/design-system/src/components/Table/TableCell.jsx
nichia: This is any additional Stacked table class that need to be add when we display the stacked Table Cells

nichia: Probably need more detail description

packages/design-system/src/components/Table/TableCell.jsx
nichia: Good point, i have not used the children for the cell, instead have put the cell value in data. Will explore using children only and maybe keep data for customised rendering?

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,140 @@
...
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * Additional classes to be added to the root table element.
+   */
+  className: PropTypes.string,
+  /**
+   * Responsive design breakpoint prefix to apply stack cells style at different viewpoint sizes.
+   */
+  stacked: PropTypes.oneOf(['sm', 'md', 'lg']),
```
line47, line 125: I'm wondering if we change this to `breakpoint` or `responsiveBreakpoint` ?

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -0,0 +1,95 @@
+import {
+  Table,
+  TableBody,
+  TableCaption,
+  TableCell,
+  TableHead,
+  TableHeader,
```
bernardwang, line 7: I noticed that `TableHeader` and `TableHead` are similar in name. I know they correspond to `thead` and `th`, which are just similar to begin with, but I wonder if we can clarify the naming?

Maybe `th` and `td` could correspond to `TableHeaderCell` and `TableDataCell`? This way we can differentiate cells from the table caption, header, and body

packages/design-system/src/components/Table/Table.jsx
bernardwang: or maybe even `responsiveTable`? or

packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx
davidakennedy: Good thoughts! Tables don't always have to have a caption. So it's okay if it doesn't have one, and you relied on the h3. It all depends on the content around the table, and how complex the table is in terms of content.

To get the best of both worlds, you could put a h3 inside the caption. That way, the heading structure is there and you're not repeating info. That would be a good way to go here.

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,151 @@
...
+
+  // `tabIndex` state is needed to make table container focusable and to display scrollable message when table width exceeds viewport.
+  // Set attribute `tabIndex = 0` to make table container focusable to enable keyboard support of using the arrow keys.
+  // Also, provide context for screen reader users as they are able to focus on the region.
+  // Do this by using table's <caption> to label the scrollable region using aira-labelleby
+  const attributescrollable = scrollable && {
+    className: 'ds-c-table__wrapper',
+    role: 'group',
+    'aria-labelledby': captionID.current,
+    'aria-live': 'polite',
```
davidakennedy, line 74: In talking to @1Copenut, this is needed to let screen readers know that the area is scrollable, because they may not be able to see the page change. So it's good to have it, but it doesn't work because the alert is outside of the table. So it doesn't pick up the addition.
- We could move the alert in the table caption. May not be ideal in all instances of the component because not all tables need a caption. But maybe if there wasn't one, this was added in there?
- Or we could put the `aria-live` on the wrapper div and put the notice just inside that `div`, before the table. That may be cleaner code wise.

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,151 @@
...
+      </div>
+    </>
+  );
+};
+
+Table.defaultProps = {
+  scrollableActiveAlert: (
+    <div className="ds-u-margin-y--1">
+      <Alert>
+        <p className="ds-c-alert__text" aria-hidden="true">
```
davidakennedy, line 116: Connected to the above feedback, if screen readers need to hear the alert on aria-live, we'll need to remove the `aria-hidden`.

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,151 @@
...
+    className
+  );
+
+  // `tabIndex` state is needed to make table container focusable and to display scrollable message when table width exceeds viewport.
+  // Set attribute `tabIndex = 0` to make table container focusable to enable keyboard support of using the arrow keys.
+  // Also, provide context for screen reader users as they are able to focus on the region.
+  // Do this by using table's <caption> to label the scrollable region using aira-labelleby
+  const attributescrollable = scrollable && {
+    className: 'ds-c-table__wrapper',
+    role: 'group',
```
davidakennedy, line 72: I did some research on the group role vs. region role. Some good information here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_group_role

I like keeping it as group because that means it doesn't surface programmatically to screen readers in navigation within the screen reader. Which is okay here because I don't think this info is that vital.

packages/design-system/src/components/Table/Table.jsx
nichia: Added alert to caption.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -3,6 +3,8 @@ Table
 
 Tables show tabular data in columns and rows.
 
+@status Beta
```
line47, line 6: Let's change this to `@status Draft` https://design.cms.gov/guidelines/component-maturity/

packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx
bernardwang: I'm good with an `h3` inside the caption as long as the styles don't change. And @nichia, what do you mean its needed to tie back to `tabIndex`? I thought `tabIndex` is for making the table focusable, and we decided in this iteration not to worry about adding additional focus functionality

packages/design-system/src/components/Table/TableBody.jsx
bernardwang: Right I understand that, I'm asking what values does `attributeOptions` hold right now? `other` is the common name for destructured variables that are usually empty (except when users pass on additional props), and a naming pattern like `tableBodyProps` would be specific to a destructured variable that holds a group of props. I think this naming could be more helpful.

packages/design-system/src/components/Table/TableCell.jsx
bernardwang: I think it makes sense to only use `children`, as any value inside `children` will just render inside `TableCell` which I think is what we want. What do you mean by customized rendering?

packages/design-system/src/components/Table/TableCell.jsx
```diff
@@ -0,0 +1,84 @@
...
+      { isValidStackedTitle } && (
+        <span aria-hidden="true" className={stackedClasses}>
+          {stackedTitle}
+        </span>
+      )
+    );
+  };
+
+  return (
+    <td className={classes} role="cell" {...attributeOptions}>
```
bernardwang: Is `type` ever passed onto the `td`?

packages/design-system/src/components/Table/TableCell.jsx
bernardwang: Right I understand that, I was asking if this `prop` was being used. I see now that it's inside `attributeOptions`. I think in this case I would suggest explicitly define `headers` instead of using `attributeOptions`.

packages/design-system/src/components/Table/TableDataCell.jsx
```diff
@@ -0,0 +1,74 @@
...
+   */
+  className: PropTypes.string,
+  /**
+   * The `headers` attribute contains a list of `id` attributes of the associated data. If there is more than one id, they are separated by spaces.
+   */
+  headers: PropTypes.string,
+  /**
+   * Additional classes to be added to the stacked Title element.
+   */
+  stackedClassName: PropTypes.string,
```
line47, line 63: I noticed that the stackedClassName does not get applied to the spans at the row inside a `th`
Should they be applied to the span in the `th` here? 

In the example, I set the class name `stackerCell`

Screen Shot 2020-06-15 at 11 01 18 AM

packages/design-system/src/components/Table/Table.jsx
bernardwang: Actually on second thought, responsive might not be the best description considering horizontal scroll is also a way to make the table responsive. I think including `stack` in the naming does make sense. Maybe `stackBreakpoint`?

packages/design-system/src/components/Table/Table.jsx
bernardwang: I like the idea of putting the alert in the caption dynamically.

packages/design-system/src/components/Table/Table.jsx
bernardwang: You should be able to use `uniqueId` if you want a unique id for the caption

packages/design-system/src/components/Table/Table.jsx
bernardwang: I would also consider making `scrollableCaption` required, this way `scrollable` tables will always have a caption. It already has a default value and users can always override it.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +22,79 @@ Markup: table-complex.example.html
...
+
+@react-props Table.jsx
+
+Style guide: components.table.react
+*/
+
+/*
+`<TableBody>`
+
+A `TableBody` component renders the table body element <tbody> and contains a set of table rows defining the table data cell.
```
line47, line 41: wrap <tbody> with single backtick quotes so it shows up like this 
```
`<tbody>`
```

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +22,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableBody.jsx
+
+Style guide: components.table.react-tablebody
+*/
+
+/*
+`<TableCaption>`
+
+A `TablecCaption` component specifies the caption of a table.
```
line47, line 51: change description to 
> A `TablecCaption` component renders table caption element `<caption>`.

and add backticks around the 
```
`caption`
```

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +22,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableCaption.jsx
+
+Style guide: components.table.react-tablecaption
+*/
+
+/*
+`<TableDataCell>`
+
+A `TableDataCell` component renders the table data cell element <td>.
```
line47, line 61: add single backtick quotes around the td
```
`<td>`
```

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +22,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableDataCell.jsx
+
+Style guide: components.table.react-tabledatacell
+*/
+
+/*
+`<TableHead>`
+
+A `TableHead` component renders the table head element <th> and contains a set of rows defining the table header columns.
```
line47, line 71: table head element should be `<thead>`
```
`<thead>`
```

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
bernardwang: `uniqueId` should always generate a unique id, regardless of if there are multiple tables. We already discussed this in our chat, but instead of using random ids here, we should just explicitly define clear readable values because this is an example.

packages/design-system/src/components/Table/Table.jsx
bernardwang: Small nitpick, but you don't need to explain what a hook is or how it works, just the part about how this hook listens to the window resize event to dynamically handle horizontal scrollable tables.

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,145 @@
...
+  scrollable,
+  scrollableAlert,
+  children,
+  ...attributeOptions
+}) => {
+  const container = useRef(null);
+  // The captionID is calculated and stored as init value of a ref.
+  // This ensures that the ID remains constant for all renders.
+  const captionID = useRef('caption-' + Math.random().toString(36).substr(2, 9));
+  const [tabIndex, setTabIndex] = useState(null);
```
bernardwang, line 31: Instead of having `tabIndex` inside the state, what do you think about using a boolean named `scrollable` or `isTableScrollable`. This is more descriptive than tab index, and you can use it directly instead of including logic like this: `const isActive = tabIndex === '0';`

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +22,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableHead.jsx
+
+Style guide: components.table.react-tablehead
+*/
+
+/*
+`<TableHeaderCell>`
+
+A `TableHeaderCell` component renders the table header <th> element.
```
line47, line 81: add single backtick quotes around the `th` to get it to render 
```
`<th>`
```

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +22,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableHeaderCell.jsx
+
+Style guide: components.table.react-tableheadercell
+*/
+
+/*
+`<TableRow>`
+
+A `TableRow` component renders the table row <tr> element and contains a row of table cells.
```
line47, line 91: add single backtick quotes around the `<tr>` element
```
`<tr>`
```

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,145 @@
...
+    // Set `tabIndex` to `0` if the table width is wider than the viewport and Table is set for horizontal scrolling.
+    if (scrollable) {
+      const { scrollWidth, clientWidth } = container.current;
+      const scrollableActive = scrollWidth > clientWidth;
+      setTabIndex(scrollableActive ? '0' : null);
+
+      const debouncedHandleResize = debounce(function handleResize() {
+        const { scrollWidth, clientWidth } = container.current;
+        const scrollableActive = scrollWidth > clientWidth;
+        setTabIndex(scrollableActive ? '0' : null);
```
bernardwang, line 46: I think this state change could also be a bit more smart, we could add an if statement to only set the state when needed

packages/design-system/src/components/Table/TableCaption.jsx
```diff
@@ -0,0 +1,31 @@
...
+   */
+  children: PropTypes.node,
+  /**
+   * Additional classes to be added to the caption element.
+   */
+  className: PropTypes.string,
+  /**
+   * Additional text or content to display when the horizontal scrollbar is visible to give the user notice of the scroll behavior.
+   */
+  scrollableAlert: PropTypes.node,
```
bernardwang, line 28: Based off [this discussion](https://github.com/CMSgov/design-system/pull/706#discussion_r440244955)

I think this should be required, this way scrollable tables will always have a caption. It already has a default value and users can always override it.

packages/design-system/src/components/Table/TableCell.jsx
nichia: I was thinking adding an option via prop to pass a render function for customizing the children data. But decided not to as the children value should contain renderable data already.

packages/design-system/src/components/Table/TableCell.jsx
nichia: Not yet, it defaults to 'text', will add numeric column to pass type='numeric'

packages/design-system/src/components/Table/Table.jsx
nichia: Good point.

packages/design-system/src/components/Table/Table.jsx
nichia: This many not work for screens with more than one table created by different components as `uniqueId` generates sequential number unique to each component, so we'll end up having duplicates in multi-table scenario.

packages/design-system/src/components/Table/TableDataCell.jsx
nichia: It applied stackedClassName to the span for me. Can you try again.

packages/design-system/src/components/Table/Table.jsx
nichia: Not sure I understand correctly, it presently checks for if `scrollable` is set, then perform the listen for window resize event.  Is this what you're suggesting we do?

packages/design-system/src/components/Table/TableDataCell.jsx
line47: It worked for me, thanks!

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,125 @@
...
+
+      // Create window resize event listener that calls the debouncedHandleResize function
+      window.addEventListener('resize', debouncedHandleResize);
+
+      // Clean up the establishd event listeners
+      return () => {
+        window.removeEventListener('resize', debouncedHandleResize);
+      };
+    }
+  });
```
bernardwang, line 53: Relaying feedback from chat:

https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
theres no need to create and destroy the window listeners each time if we pass in an extra parameter so the react hook can tell when its actually necessary to update the window listeners. I would probably use the scrollable prop as the parameter to check

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,125 @@
...
+
+  // Listens to the window resize event to dynamically handle horizontal scrollable tables.
+  useEffect(() => {
+    // Set isTableScrollable `true` if the table width is wider than the viewport and Table is set for horizontal scrolling.
+    if (scrollable) {
+      const { scrollWidth, clientWidth } = container.current;
+      const isScrollActive = scrollWidth > clientWidth;
+      setTableScrollable(isScrollActive);
+
+      const debouncedHandleResize = debounce(function handleResize() {
```
bernardwang, line 39: This can be defined outside of the effect I think

packages/design-system/src/components/Table/Table.jsx
bernardwang: Oh, I always thought that `uniqueId` would just handle that case. Have you tested this? Just even though they seem sequential they should ensure no duplicates happen, otherwise whats the point of the function?

packages/design-system/src/components/Table/TableCell.jsx
bernardwang: Yeah, I think `data` was left over from when we would pass in objects containing table data. Now that we are doing the approach of `children`, i dont think its necessary

packages/design-system/src/components/Table/Table.jsx
nichia: Yes, `uniqueId` creates unique value within a component with features like prefixing the unique value. So another component using `uniqueId` will end up with the same sequential value. `uniqueId`  can be substituted by a helper function that accepts prefixes to generates unique sequential value.

davidakennedy:
I did some more testing on this, and just wanted to share the results:

- Safari + VoiceOver: Table navigation works well on large and small screens. However, I don't hear the keyboard navigation notice after I resize the screen. Need to dig deeper here.
- Edge and JAWS: Table navigation works well on large and small screens. I hear the keyboard navigation notice after I resize the screen.
- IE11 and JAWS: Table navigation works well on large and small screens. I don't hear the keyboard navigation notice after I resize the screen. I don't see this as a blocker, especially as IE11 is an older browser.
- Chrome and JAWS: Table navigation works well on large and small screens. I hear the keyboard navigation notice after I resize the screen.
- Firefox and NVDA: Table navigation works well on large and small screens. I hear the keyboard navigation notice after I resize the screen.

So overall, it's working really well, @nichia! 🎉  I'm going to dig into the Safari issue to see what I find.

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -44,6 +40,141 @@
...
+    @include ds-c-table--stacked;
+  }
+}
+
+/* Hide individual cell heading reserved for stacked responsive table */
+.ds-c-table--stacked__col-header {
+  display: none;
+}
+
+.ds-c-table--striped tbody {
```
bernardwang, line 107: I believe this is a new style that we are supporting, so we should bring this up to the top similar to `.ds-c-table--borderless`. We also need to add a new HTML example for this style

packages/design-system/src/components/Table/TableCaption.jsx
```diff
@@ -0,0 +1,51 @@
...
+    <caption className={classes} {...tableCaptionProps}>
+      {children}
+      {scrollActive && scrollCaption}
+    </caption>
+  );
+};
+
+TableCaption.defaultProps = {
+  scrollCaption: (
+    <Alert className="ds-u-margin-y--1 ds-u-font-size--small ds-u-font-weight--normal">
```
davidakennedy, line 25: Regarding one of my previous comments: https://github.com/CMSgov/design-system/pull/706#issuecomment-646247189

Can we change the role on this alert from region to status? In my testing, that makes the keyboard message announce in IE11 and Safari.

nichia: @davidakennedy I've changed role to 'status'. From reading up on role status, it mentions that this status is similar to aria-live="polite" & aria-relevant='additions', so I've removed those attributes too.

packages/design-system/src/components/Table/TableDataCell.jsx
```diff
@@ -0,0 +1,74 @@
...
+  /**
+   * The stacked row title for responsive table accessiblity.
+   */
+  stackedTitle: PropTypes.string,
+  /**
+   * Type of the data, can be either text or numeric for left or right alignment respectively.
```
line47, line 69: Consider changing this prop from type to something like `align` or `alignment` with values of left or right. This approach is a more generic and directly applies to how the prop is used. 

Type is currently used on `TableDataCell` and `TableHeaderCell` 

Upon further consideration in our docs we recommend users to 
> Right-align numerical data in tables. You can use a text alignment utility class for this. This makes it easier to scan and compare values.

Should we remove this type prop all together and let users define classes to accomplish the same thing?

CC @bernardwang

nichia: Good point @line47 - alignment of data can be done via utility classes. I'm good with removing the type prop.

packages/design-system/src/components/Table/Table.jsx
bernardwang: Have you tested only using `uniqueId` instead of using that function? We use `uniqueId` across the application, and if it only was unique within a single instance of a component than alot of our components should have broken.

packages/design-system/src/components/Table/TableBody.jsx
bernardwang: @nichia I dont think this question was answered, what values is in `attributeOptions`. I ask because depending on that, I think the variable could be renamed to `other` or something more specific

bernardwang: https://github.com/CMSgov/design-system/pull/706#discussion_r440242715
That is a good example of why it might be better to explicitly define certain variables and make the destructured variable naming very clear.

packages/design-system/src/components/Table/TableCaption.jsx
bernardwang: @nichia This is still relevant I believe

packages/design-system/src/components/Table/TableDataCell.jsx
bernardwang: I would also prefer to remove

bernardwang: Yeah I think we can remove the prop and make sure to include an example with right aligned numerical data

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -0,0 +1,110 @@
...
+            );
+          }
+        })}
+      </TableRow>
+    );
+  });
+};
+
+ReactDOM.render(
+  <Table stackBreakpoint="sm" scrollable striped>
```
bernardwang, line 104: Instead of calling `renderHeader()` or `renderRows()`, I would show that code inline so people can see the table structure better

packages/design-system/src/components/Table/TableDataCell.jsx
bernardwang: Ah i see the default example does that already

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -0,0 +1,110 @@
...
+      <TableRow key={uniqueId('row_')}>
+        {simpleHeaders.map((header, columnIndex) => {
+          if (columnIndex === 0) {
+            return (
+              <TableHeaderCell
+                key={header.key}
+                scope="row"
+                id={'row' + rowIndex}
+                stackedTitle={header.title}
+                stackedClassName="stackerCell"
```
bernardwang, line 79: Is this className actually being used? In general, I think examples don't have to make sure of every prop, just the ones relevant to the use case

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -0,0 +1,110 @@
...
+  TableHead,
+  TableHeaderCell,
+  TableRow,
+} from '@cmsgov/design-system';
+import React from 'react';
+import ReactDOM from 'react-dom';
+import uniqueId from 'lodash.uniqueid';
+
+const simpleHeaders = [
+  { title: 'Branch', key: 'branch', width: '10' },
```
bernardwang, line 15: I would also personally not define `width` for all of these, maybe 1 only. Generally speaking I think we want users to not use width unless they really need to right? The default table spacing should handle it most of the time

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -17,9 +19,84 @@ Table with two headers and a caption
 
 Markup: table-complex.example.html
 
+.ds-c-table--striped - Striped table
```
bernardwang, line 22: I see you added the stripe class to the existing complex example. I think we should update our Table examples to be consistent with other pages, i.e. [the button page](http://localhost:3000/components/button/)

I've never told you any of this context or these design decisions, but basically this page uses our legacy design for examples, where we use the `modifier` tag to automatically generate documentation for variants. You can see that [here](https://github.com/CMSgov/design-system/pull/706/files/818b6597b1862b4d9e7ae812b44475216a1ea71f#diff-c408290eec1f1b2fbb29243804f68058L10) and in `table.example.html`, which uses EJS templating to generate documentation. 

In the past, we revamped most of our pages to use the new pattern, where we try to have only 1 larger example file, and label them instead of just showing the class. 

You can see an example of this by comparing this [legacy style](https://design.cms.gov/style/base/) with the [newer style](https://design.cms.gov/utilities/background-color/)

bernardwang: Basically, I think it makes sense to update `table.example.html` to contain the default, borderless, and striped variants inside a single, large table. Then we can keep the complex table example the same as a separate section, as its technically not a variant, but just a different more complex use case. Another example of this is the [button with icons](https://design.cms.gov/components/button/#components.button.icons), they have their own separate example file because they dont actually have a separate variant classname, and they are just a different use case of the default styling.

nichia: Good suggestion.

davidakennedy:
Hey @nichia, I tested the latest based on your changes here: https://github.com/CMSgov/design-system/pull/706#discussion_r444149799.

Everything works as I'd expect.

I did notice that with Safari and JAWS, if the browser window changes side, causing the status alert to trigger, it announces the status, and then moves onto the whole table. I'm not sure if that's because of a screen reader setting or a standard thing for `role="status"`. It doesn't happen with NVDA. Regardless, I think it's good to go. We can always iterate on this part if we need to later.

packages/design-system/src/components/Table/Table.jsx
nichia: Thanks, retested uniqueId and it works.

packages/design-system/src/components/Table/TableBody.jsx
nichia: Updated to have more explicit usage of the variables.

packages/design-system/src/styles/components/_Table.scss
nichia: Html example updated for the new style.

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
nichia: Move renderHeader() processing code but kept renderRows() as it iterates through the data array.

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
nichia: Thanks for picking up, it's been removed.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
nichia: Updated `table.example.html` to contain the through variants.

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
bernardwang: Both `renderHeader` and `renderRows` can be defined inline in the `return` statement. Putting it inline rather than defining separately makes it easier to understand the markup

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
bernardwang, line 6: I think we should actually be using Draft here, you can see this in `PageHeader`. I'm also going to update this status stuff in a different ticket: https://jira.cms.gov/browse/WNMGDS-532

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
nichia: Okay, the header and rows are linline in the `return` statement.

bernardwang:
@nichia @davidakennedy Because we put the horizontal scroll notice in the caption, it's also affected when scroll horizontally. We probably need to rethink the [aria labels on the caption](https://github.com/CMSgov/design-system/pull/706#pullrequestreview-433689912) and also move the ["Props" heading out of the caption](https://github.com/CMSgov/design-system/pull/706#discussion_r438474809)

bernardwang:
For the react table example, can we pair down the content to match [USWDS](https://designsystem.digital.gov/components/table/). I think in general we want to avoid super text heavy examples

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
 Style guide: components.table.complex
 */
 
+/*
+`<Table>`
+
+`Table` and sub-table components render data tables using semantic markup and ARIA attributes.
```
bernardwang, line 26: Maybe we can tighten up this description, similar to `Tabs`

`Table` is a container component that contains `TableCaption`, `TableHead` and `TableBody` as children, as well as `TableRow`, `TableHeaderCell` and `TableDataCell` for the table content. While the usage of these components follows ordinary HTML table semantics, this React implementation includes additional responsive features including horizontal scrolling and vertically stacked rows.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
...
+
+@react-example Table.example.jsx
+
+@react-props Table.jsx
+
+Style guide: components.table.react
+*/
+
+/*
+`<TableBody>`
```
bernardwang, line 37: I think the order of these components should follow the markup, i.e. TableCaption, TableHead, TableBody, TableRow, TableHeaderCell and TableDataCell

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
...
+
+A `TableBody` component renders the table body element `<tbody>` and contains a set of table rows defining the table data cell.
```
bernardwang, line 39: `TableBody` renders a `<tbody>` element and contains table rows defining the table data.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableBody.jsx
+
+Style guide: components.table.react-tablebody
+*/
+
+/*
+`<TableCaption>`
+
+A `TablecCaption` component specifies the `<caption>` of a table.
```
bernardwang, line 49: `TableCaption` renders the `<caption>` element.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableDataCell.jsx
+
+Style guide: components.table.react-tabledatacell
+*/
+
+/*
+`<TableHead>`
+
+A `TableHead` component renders the table head element `<thead>` and contains a set of rows defining the table header columns.
```
bernardwang, line 69: `TableHead` renders the `<thead>` element and contains table rows defining the table columns.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableHeaderCell.jsx
+
+Style guide: components.table.react-tableheadercell
+*/
+
+/*
+`<TableRow>`
+
+A `TableRow` component renders the table row `<tr>` element and contains a row of table cells.
```
bernardwang, line 89: `TableRow` renders a `<t>` element.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -20,6 +20,79 @@ Markup: table-complex.example.html
...
+
+@react-props TableHead.jsx
+
+Style guide: components.table.react-tablehead
+*/
+
+/*
+`<TableHeaderCell>`
+
+A `TableHeaderCell` component renders the table header `<th>` element.
```
bernardwang, line 79: `TableHeaderCell` renders a `<th>` element

packages/design-system/src/components/Table/TableRow.jsx
```diff
@@ -0,0 +1,26 @@
...
+  return (
+    <tr className={classes} role="row" {...others}>
+      {children}
+    </tr>
+  );
+};
+
+TableRow.propTypes = {
+  /**
+   * The table row contents, usually `TableHeader` and `TableCell`.
```
bernardwang, line 17: This should be updated to `TableDataCell` and `TableHeaderCell`

packages/design-system/src/components/Table/TableHeaderCell.jsx
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Additional classes to be added to the stacked Title element.
+   */
+  stackedClassName: PropTypes.string,
+  /**
+   * The stacked row title for responsive table accessiblity.
+   */
+  stackedTitle: PropTypes.string,
+  /**
+   * Sets the width of `TableHeader` to a percentage of the `Table` width.
```
bernardwang, line 75: Sets the width of `TableHeaderCell` to a percentage of the `Table` width.

bernardwang: Actually, I think we should consider removing this prop. Users can manually set the width of their `TableHeaderCell` with ordinary CSS

packages/design-system/src/components/Table/TableHeaderCell.jsx
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Scope of the header, can be 'row' or 'col'.
+   */
+  scope: PropTypes.oneOf(['row', 'col']),
+  /**
+   * Additional classes to be added to the stacked Title element.
+   */
+  stackedClassName: PropTypes.string,
+  /**
+   * The stacked row title for responsive table accessiblity.
```
bernardwang, line 71: Could we make this description more helpful? It doesnt really explain when this prop is used, or what value should be provided.

packages/design-system/src/components/Table/TableHeaderCell.jsx
```diff
@@ -0,0 +1,80 @@
...
+  /**
+   * Additional classes to be added to the row element.
+   */
+  className: PropTypes.string,
+  /**
+   * The `id` attribute is used as a header for other cells to create associations between headers and cells.
+   */
+  id: PropTypes.string,
+  /**
+   * Scope of the header, can be 'row' or 'col'.
```
bernardwang, line 63: I think this can simply be `Scope of the header cell`

packages/design-system/src/components/Table/TableHeaderCell.jsx
bernardwang: also is this required? if it is required in certain scenarios that we cant validate through proptypes, we want to validate manually and have console warnings for developers

packages/design-system/src/components/Table/TableHeaderCell.jsx
```diff
@@ -0,0 +1,80 @@
...
+   */
+  children: PropTypes.node,
+  /**
+   * Additional classes to be added to the row element.
+   */
+  className: PropTypes.string,
+  /**
+   * The `id` attribute is used as a header for other cells to create associations between headers and cells.
+   */
+  id: PropTypes.string,
```
bernardwang, line 61: Is this required?

packages/design-system/src/components/Table/TableHeaderCell.jsx
```diff
@@ -0,0 +1,80 @@
...
+  );
+};
+
+TableHeaderCell.defaultProps = {
+  scope: 'col',
+};
+
+TableHeaderCell.propTypes = {
+  /**
+   * The table header cell contents.
```
bernardwang, line 51: I think we can remove this `children` description

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * Apply vertically stacked row style at different viewpoint sizes.
+   */
+  stackBreakpoint: PropTypes.oneOf(['sm', 'md', 'lg']),
+  /**
+   * A striped variation of the table.
+   */
+  striped: PropTypes.bool,
+  /**
+   * Apply horizontal scrollbar when the `Table` contents exceed the container width.
```
bernardwang, line 113: "Applies a horizontal scrollbar and scrollable notice when the `Table`'s contents exceed the container width."

bernardwang: If we choose to move `scrollableCaption` out of the caption, we should put it here and consider renaming it to `scrollableNotice` or `scrollableAlert` or something

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -0,0 +1,118 @@
...
+  /**
+   * The table contents, usually `TableCaption`, `TableHead` and `TableBody`.
+   */
+  children: PropTypes.node.isRequired,
+  /**
+   * Additional classes to be added to the root table element.
+   */
+  className: PropTypes.string,
+  /**
+   * Apply vertically stacked row style at different viewpoint sizes.
```
bernardwang, line 105: "Applies responsive styles to vertically stack rows at the specified breakpoints."

packages/design-system/src/components/Table/TableHeaderCell.jsx
nichia: Yes, added more descriptions to the prop

packages/design-system/src/components/__tests__/index.test.js
```diff
@@ -35,6 +35,7 @@ function getDirectories(paths) {
 }
 
 const ignoredComponents = [
+  'TableContext',
```
bernardwang, line 38: nice job!

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -9,93 +9,47 @@ import {
...
-              key={header.key}
-              width={header.width}
-              scope="col"
-              id={'column' + header.key}
-            >
-              {header.title}
-            </TableHeaderCell>
-          );
-        })}
+        <TableHeaderCell id={'column' + 1} scope="col">
```
bernardwang, line 18: This is much more readable thank you!

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -102,17 +126,24 @@ Table.propTypes = {
    */
   className: PropTypes.string,
   /**
-   * Apply vertically stacked row style at different viewpoint sizes.
+   * Applies responsive styles to vertically stacked rows at different viewpoint sizes.
+   * Required props to be used in conjunction with this includes `headers`, `id` and `stackedTitle`
```
bernardwang, line 130: Slight wording suggestions. Also I know I originally suggested this wording haha
"Applies responsive, stacked table style at the specified breakpoint. When `stackBreakpoint` is set, `headers`, `id` and `stackedTitle` props are required in `TableHeaderCell` and `TableDataCell`.

bernardwang: Also is that how it works? Are all 3 props required for both header cell and data cell? we should be clear which props are required where.

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -102,17 +126,24 @@ Table.propTypes = {
...
    * A striped variation of the table.
    */
   striped: PropTypes.bool,
   /**
-   * Apply horizontal scrollbar when the `Table` contents exceed the container width.
+   * Applies a horizontal scrollbar and scrollable notice when the `Table`'s contents exceed the container width.
    */
   scrollable: PropTypes.bool,
+  /**
+   * Additional text or content to display when the horizontal scrollbar is visible to give the user notice of the scroll behavior.
```
bernardwang, line 143: This is great!

packages/design-system/src/components/Table/TableCaption.jsx
```diff
@@ -31,14 +21,9 @@ TableCaption.propTypes = {
    */
   className: PropTypes.string,
   /**
-   * @hide-prop This gets passed through from the parent Table when the horizontal scroll is active.
-   */
-  _scrollActive: PropTypes.bool,
-  /**
-   * Additional text or content to display when the horizontal scrollbar is visible to give the user notice of the scroll behavior.
-   * This prop will only be used when the `Table` `scrollable` prop is set and the table width is wider than the viewport.
+   * @hide-prop This gets passed from the parent <Table> component when the horizontal scroll is active.
```
bernardwang, line 24: nice use of @hide-prop

packages/design-system/src/components/Table/TableDataCell.jsx
```diff
@@ -10,6 +11,14 @@ export const TableDataCell = ({
   stackedTitle,
   ...others
 }) => {
+  const tableStackableContext = useContext(TableContext);
+  if (process.env.NODE_ENV !== 'production') {
+    if (tableStackableContext && !headers) {
+      console.warn(
+        `Please provide a 'headers' prop for stackable table, it is a required prop for screen readers to create association between the heading id and data cells.`
```
bernardwang, line 18: "The `headers` prop in `TableDataCell` is required for stackable tables. This prop is needed to associate the headings with data cells in the responsive stacked view."

packages/design-system/src/components/Table/TableDataCell.jsx
```diff
@@ -31,24 +40,24 @@ export const TableDataCell = ({
...
   /**
    * Additional classes to be added to the table cell element.
    */
   className: PropTypes.string,
   /**
-   * The `headers` attribute contains a list of `id` attributes of the associated data. If there is more than one id, they are separated by spaces.
+   * Table Data cells must have a `headers` attribute for stackable table.
+   * The `headers` attribute must link to a Table Header cell’s `id`.
+   * If there is more than one `id`, they are separated by spaces.
+   * For screen readers to create association between the header and data cells.
```
bernardwang, line 52: A little more compact:

`TableDataCell` must define a `headers` prop for stackable tables. The `headers` prop is needed to associate header and data cells for screen readers. 

`headers` consist of a list of space-separated ids that each correspond to a `TableHeaderCell` element. [Read more on the `headers` attribute.](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td#Attributes)

packages/design-system/src/components/Table/TableHeaderCell.jsx
```diff
@@ -9,14 +10,18 @@ export const TableHeaderCell = ({
...
-  const classes = classNames(
-    'ds-c-table__header',
-    className,
-    width ? 'ds-c-table__header--width-' + width : null
-  );
+  const tableStackableContext = useContext(TableContext);
+  if (process.env.NODE_ENV !== 'production') {
+    if (tableStackableContext && !id) {
+      console.warn(
+        `Please provide an 'id' prop for stackable table, it is a required prop for screen readers to create association between the heading id and data cells.`
```
bernardwang, line 19: Similar feedback as above, we want to have relatively consistent messaging, you can also see examples of this in `HelpDrawer.jsx`

davidakennedy:
@nichia @bernardwang We had most of this conversation in the CMS Slack, but wanted to share it here too.

Currently, moving the notice outside of the caption means that:

- screen reader users are hearing the contents of the table announced on width change, but not the scroll message. Before, they heard both.
- If screen reader users navigate to the table with common screen reader shortcuts, they won't hear the scroll message because it's not in the caption. Typically, captions are announced when you navigate to a table in a screen reader. It helps give extra context if it's there.

I understand not wanting to have the scroll message scroll as well.

I believe it's more resilient to have it inside the caption because we get the above for free. But if we wanted to move it outside of the caption, we could:

- Wrap it in a div with the proper roles, with the table also inside that.
- We’d need to add an `id` to the notice, and associate that with `aria-describedby` to the caption. That way, people would hear the screen notice if they landed on the table by skipping to it.

So that could work, but would require some testing.

bernardwang:
@nichia @davidakennedy For the purpose of moving this PR forward, I'm ok with reverting the recent caption notice changes and just handle that in a followup, considering it might take some time to get feedback on that

nichia:
@bernardwang @davidakennedy I've reverted the scrollable notice to appear on Table Caption. Ready for a re-test.
Note: To do - add followup work to move scrollable notice outside of Table Caption and update Alert component to accept role of 'status'.

bernardwang:
@nichia I retested and I cant seem to get the scroll caption to show up.

packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx
```diff
@@ -28,11 +28,8 @@ class ReactPropDocs extends React.PureComponent {
 
   render() {
     return [
-      <h3 key="propDocsHeader">Props</h3>,
       <Table key="propDocsTable" stackBreakpoint="sm" scrollable>
-        <TableCaption className="ds-u-visibility--screen-reader">
-          React Properties Documentation
-        </TableCaption>
+        <TableCaption className="ds-u-padding-y--2 ds-u-font-size--h3">Props</TableCaption>
```
bernardwang, line 32: Instead of using `ds-u-font-size--h3`, can we just put the h3 inside the caption

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -70,8 +61,10 @@ export const Table = ({
   // Do this by using table's <caption> to label the scrollable region using aira-labelleby
```
bernardwang, line 61: small typo here

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -128,22 +110,17 @@ Table.propTypes = {
...
   /**
-   * Applies a horizontal scrollbar and scrollable notice when the `Table`'s contents exceed the container width.
+   * Applies a horizontal scrollbar and scrollable notice on `TableCaption` when the `Table`'s contents exceed the container width.
    */
   scrollable: PropTypes.bool,
-  /**
-   * Additional text or content to display when the horizontal scrollbar is visible to give the user notice of the scroll behavior.
-   * This prop will only be used when the `Table` `scrollable` prop is set and the table width is wider than the viewport.
-   */
-  scrollableNotice: PropTypes.node,
```
bernardwang: I still like the `scrollableNotice` to be inside the parent `Table` component, this way its right next to the `scrollable` prop. Is it possible to pass the `scrollableNotice` to the `TableCaption` children? Maybe through the `TableContext`?

packages/design-system-docs/src/scripts/components/ReactPropDocs.jsx
bernardwang: Actually, can we just move the `h3` out of the caption altogether? Putting it in the caption means it will scroll with the table. Even if we accept that the caption notice scrolls with the table, I don't think its necessary for the heading to as well

packages/design-system/src/components/Table/TableDataCell.jsx
```diff
@@ -0,0 +1,65 @@
...
+  children,
+  className,
+  headers,
+  stackedClassName,
+  stackedTitle,
+  ...others
+}) => {
+  const tableStackableContext = useContext(TableContext);
+  if (process.env.NODE_ENV !== 'production') {
+    if (tableStackableContext && !headers) {
```
bernardwang, line 16: I think this is also missing a warning for a missing `stackedTitle`

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -0,0 +1,56 @@
...
+  TableDataCell,
+  TableHead,
+  TableHeaderCell,
+  TableRow,
+} from '@cmsgov/design-system';
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+ReactDOM.render(
+  <Table stackBreakpoint="sm" scrollable>
```
bernardwang, line 14: The stackable part is working great, but for some reason I cant get the example to show the scrollable notice. Currently the text in this example never causes the table to overflow its container and start the horizontal scrolling. This maybe can be handled in a followup, but maybe we need to update the content and styling so that theres an easy to test the scrollable behavior.

nichia: Added another column `Links` to enable scroll behavior.

packages/design-system/src/components/Table/Table.jsx
nichia: Keeping it consistent including it with the extending TableCaption props

davidakennedy:
@bernardwang @nichia I tested the table again and didn't find any issues.

davidakennedy:
@bernardwang I approved, and I think that should fix it.

PR #707, opened 2020-05-27
[WNMGDS-455] Fix webpack tree shaking

bernardwang:
### Summary
- This PR adds ES modules to our `design-system` package to support Webpack 4's tree shaking
- The "Using React components" page has been rewritten and updated
- The `react-app` example has been updated and can be used to test tree shaking.

### How to test

1. Test the docs page
    - Check the file sizes of the doc page bundle and each example page.
    - Run `yarn build` to build the new ESM dist.
    - Check to see if files are significantly smaller.
2. Test with the `react-app` example
    - Change line 6 in the `package.json` in `react-app` to be: `"@cmsgov/design-system": "file:../../packages/design-system",`
    - Run `rm -rf node_modules` to delete node modules
    - Run `npm install` in the `react-app` example. 
    - Run `npm run build` to output the bundle.
    - Change line 4 of `react-app/src/scripts/index.js` to `import { Alert } from '@cmsgov/design-system/dist/index.js';` to simulate the original behavior, compare the bundle and ensure it is much larger
3. Check tree shaking is working in the child design systems [child-design-system-example.zip](https://github.com/CMSgov/design-system/files/4713248/child-design-system-example.zip)as well.


packages/design-system-scripts/gulp/build.js
```diff
@@ -52,6 +53,45 @@ async function copyAll(dir) {
...
 }
 
+/**
+ * Similar to compileJS but babel is configured for esmodules
+ */
+async function compileEsmJs(dir) {
+  const src = path.join(dir, 'src', 'components');
+
+  // Create esm entry file
+  await fs.writeFileSync(path.join(dir, 'dist', 'index.es.js'), `export * from './esnext';`);
```
bernardwang, line 63: I created 2 entry points (1 for ESM, 1 for CommonJS) by moving all the `exports` into an `index.js` in the `components` folder, so that I could easily generate a new `index.es.js` entry point without worrying about changing all the import paths. 

AKA, I didn't have to worry about doing this change:
```
export * from './components/Alert/Alert';

export * from './esnext/Alert/Alert'
```

packages/design-system-docs/src/pages/startup/components.md
```diff
@@ -0,0 +1,88 @@
...
+    onClick={() => alert('Button clicked!')}>
+    Button react component
+  </Button>
+  document.querySelector('#app'),
+);
+```
+
+See [React component documentation]({{root}}/components/button/#components.button.react) for more react examples, guidance, and detailed prop documentation.
+
+<h2>Using CSS components</h2>
```
line47, line 58: I'm wondering if this section could be titled `using plain HTML components` instead?

packages/design-system-docs/src/pages/startup/components.md
```diff
@@ -0,0 +1,88 @@
...
+
+If your project doesn't use React, we provide a CSS-only version of our components. This includes styles for every component in the library, but you’ll be responsible for writing the correct markup and updating classes and DOM attributes in response to user events.
```
line47, line 60: consider updating the paragraph to 

> If your project doesn't use React, we provide plain HTML versions of most of our components. This includes styles for every component in the library, but you’ll be responsible for writing the correct markup and updating classes and DOM attributes in response to user events.

packages/design-system-docs/src/pages/startup/components.md
```diff
@@ -0,0 +1,88 @@
...
+    <p class="ds-c-alert__text">
+      Alert warning body text
+    </p>
+  </div>
+</div>
+```
+
+See [component documentation]({{root}}/components/alert/) for more HTML/CSS examples and guidance.
+
+<h2 id="examples" class="ds-h2 ds-u-color--primary-darker">Examples</h2>
```
line47, line 77: I think if we just use <h2>Examples</h2> here that Markdown with automatically create in ID

packages/design-system-docs/src/pages/startup/components.md
```diff
@@ -0,0 +1,88 @@
+---
+title: Using components
+weight: 3
+---
+
+<h2>Using React components</h2>
```
line47, line 6: Use markdown where we can in this doc, for example, change heading to use the ### syntax

line47:
Bundles are giving the size of `bundle.js (271 KiB)` using both
- import { Alert } from '@cmsgov/design-system';
- import { Alert } from '@cmsgov/design-system/dist/index.js';

Maybe I'm missing something but I'm not seeing a difference in size.

packages/design-system-docs/src/pages/startup/components.md
bernardwang: I believe `CSS components` is a more common terminology for this for type of HTML/CSS implementation. At the very least, [polaris](https://github.com/Shopify/polaris-react#using-the-css-components) uses this terminology, and I modeled this page of their docs

bernardwang:
I need to test some more things with child design systems using this, so actually please hold off on reviewing!

packages/design-system-docs/src/pages/startup/components.md
bernardwang: This was existing code. It seems like "Examples" and "Need help or ran into an issue?" all seem to have that additional class. Maybe it's best to leave it here and then address it on all the pages later on?

bernardwang:
@line47 https://github.com/CMSgov/design-system/pull/707#issuecomment-637017235
It seems to work for me, are you making sure to run `yarn build` before hand? Also to clarify, you are editing the `index.js` in `react-app` right?

pwolfert:
I got the same result that Scott got. My bundle sizes are the same. Do I need to do any npm-linking or anything like that? Do I need to explicitly tell it somewhere to import the ES Modules version?

packages/design-system-docs/src/pages/startup/components.md
line47: Sounds good to me

line47:
I'm running `yarn build` at the root - should I be running it inside `react-app` ?

I am editing the `index.js` in `react-app > src`

bernardwang:
@line47 @pwolfert Ah oops, thats it I forgot to include instructions to actually use the new package dist (its using 4.0.0-beta.1 in the code). I updated the instructions on how to test to change the dep to use `file:`. NPM linking shouldnt necessary as long as you `npm install` in `react-app` after running `yarn build` at least once before.

pwolfert:
Woot! Woot! React-app example build stats:
- From master: `271 KB`
- This branch without ES modules: `385 KB`
- This branch with ES modules: `145 KB`

pwolfert:
It looks like when building the child design system example, the `dist/index.es.js` references `./esnext`, but there is no index file there because the original components folder didn't have one. In the child design system example, `src/index.js` is used to import components directly, but the generated `index.es.js` assumes we're just importing the components directory.

bernardwang:
@pwolfert Good catch! I can make that change to the child DS example. Do you think it's fine to document/enforce that structure as a requirement for child design systems? When I was writing the build scripts for ESM I wanted to keep the ESM entry point at the root, and this was the best I could think of, but it does feel a bit fragile

bernardwang:
Yeah that could work... Another idea I just had was to avoid the generating altogether and just not have the entry points be at the root. For NPM users it wouldnt matter as much because of the `main` and `module` properties in the package.json. That being said, it certainly is nice to have everything in the root...

bernardwang:
Yeah thinking about it some more, we could just give up on having entry points in the root and opt for something closer to USWDS's `dist` folder structure, where the entry point is in its corresponding file type folder. 
```
@cmsgov/design-system/
├── dist/
│   ├── css/
│   ├── components/
│   ├── esnext/
│   ├── fonts/
│   ├── images/
│   └── scss/

```

bernardwang:
@pwolfert @line47, I'm anticipating a lot of changes for this change (including documentation changes, CSS/SCSS build changes, etc). So I'm thinking of putting up a follow up PR instead, and this PR can be ready to review in the meantime

pwolfert:
@bernardwang, are you saying that package.json will have entries that import the component and esnext folders directly? That seems fine to me since we're not really doing anything in that root index.js. There's only ever one folder with JavaScript now.

line47:
**Test the docs page**
- Index.js is 211K
- example.js is 121K

**Test with the react-app example**
- bundle.js  145 KiB  before updating index.js
- bundle.js  395 KiB after updating index.js

I'm unsure how to test the tree shaking and will circle back for that

bernardwang:
@line47 The child DS example wont work because of the issue Patrick pointed out. I think its probably easier to fix it in the followup PR

packages/design-system-docs/src/pages/startup/components.md
```diff
@@ -0,0 +1,88 @@
...
+    </p>
+  </div>
+</div>
+
+```jsx
+import { Button } from '@cmsgov/design-system';
+import { TextField, ChoiceList } from '@cmsgov/design-system';
+```
+
+The `@cmsgov/design-system` package is configured to automatically work with Webpack 4's tree shaking. We provide a separate entry point (`index.es.js`) and component directory (`esnext`) for ES modules to support these optimizations.
```
pwolfert, line 30: Should we say that the resolving of this `index.es.js` module specifically is what's automatic? I read this as the tree-shaking in webpack working automatically, but it sounds to me like I have to import that `index.es.js` file directly.

pwolfert: What about something like this? I just added on to that last sentence to explain.
> The `@cmsgov/design-system` package is configured to automatically work with Webpack 4's tree shaking. We provide a separate entry point (`index.es.js`) and component directory (`esnext`) for ES modules and configure our package so that Webpack will use this entry point when _ES Modules_ are preferred.

pwolfert: Maybe I'm going in circles. I just don't want them thinking that if they import `@cmsgov/design-system/dist/components` that it will somehow magically change that path into the ES modules one

nichia:
Testing: 
- Step 1 - Yay! resulted in significant .css & .js file size reduction 
- Step 2 - did not show any file size changes
- Step 3 - to be tested in another pr

packages/design-system-docs/src/pages/startup/components.md
bernardwang: thats definitely more clear, I'll make that change

pwolfert: Great! Other than this, it all LGTM 👍

bernardwang: I went with something a little more concise. I kept the last sentence, and removed the word "automatic" to make it seem less magic

> The `@cmsgov/design-system` package provides an ES modules version of our code (located in `dist/esnext`) to support Webpack 4's tree shaking optimizations. Our package is configured to let Webpack know to resolve imports from `"@cmsgov/design-system"` to the ES modules version when preferred.

bernardwang:
Merging and will get anything else in followup PR!

packages/design-system-docs/src/pages/startup/components.md
pwolfert: Yeah, that's better!

bernardwang: Oops, I thought i pushed this change up before merging but I didnt, I'll get it in the follow up

PR #711, opened 2020-06-01
[WNMGDS-456] Remove aria-labels from StepList list items

nichia:
### Changed
- Removed `aria-label` attribute for list items with no label text

### How to test
- Add the below lines of code to StepList.example.jsx file, just below line 18 `<StepList` 
```
descriptionLabelText=""
substepsLabelText=""
actionsLabelText=""
```
- Run the design system site locally and inspect the html elements to ensure that aria-label is not shown for the step's description, step's actions and step's substeps.

### DIFF
--- a/packages/design-system/src/components/StepList/Step.jsx
+++ b/packages/design-system/src/components/StepList/Step.jsx
@@ -19,6 +19,11 @@ export const Step = ({ step, ...props }) => {
     }
   }
 
+  const getAriaLabel = (text) => {
+    const isValidTemplate = text && text.length > 0;
+    const label = isValidTemplate ? text.replace('%{step}', step.heading || step.title) : undefined;
+    return { 'aria-label': label };
+  };
   const Heading = `h${step.headingLevel || '2'}`;
   const start = step.isNextStep;
   const resume = step.started && !step.completed;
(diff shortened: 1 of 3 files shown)


bernardwang:
@nichia DK and I made some comments on https://jira.cms.gov/browse/WNMGPCII-409 that might change this PR entirely (sorry about verifying this ticket incorrectly).

Regardless of that, could you also include an update to `StepList.example.jsx` in this PR?
Basically I dont think these props are necessary as they are already defaulted
```
      completedText="Completed"
      editText="Edit"
      resumeText="Resume"
      startText="Start"
```

line47:
Sounds like more work is to be done before reviewing is that right @nichia ?

nichia:
Yes, we'll be changing the scope for it a little. Will let you know when ready for testing again @line47

nichia:
@bernardwang - glad that DK and you picked up on this change. I'll remove the props you listed (as they are already the default values) from the examples when making the changes.

packages/design-system/src/components/StepList/Step.jsx
```diff
@@ -30,9 +30,24 @@ export const Step = ({ step, ...props }) => {
     'ds-c-step__content--with-content': step.description || step.steps,
   });
   const { actionsLabelText, substepsLabelText, descriptionLabelText } = props;
-  const actionsLabel = actionsLabelText.replace('%{step}', step.heading || step.title);
-  const substepsLabel = substepsLabelText.replace('%{step}', step.heading || step.title);
-  const descriptionLabel = descriptionLabelText.replace('%{step}', step.heading || step.title);
+  const actionsLabel =
+    actionsLabelText === ''
```
bernardwang, line 34: What do you think about abstracting this out into a helper function? something like `getAriaLabel`

bernardwang:
```
const getAriaLabel = (text) => {
   const isValidTemplate = text && text.length > 0;
   const label = isValidTemplate ? text.replace('%{step}', step.heading || step.title) : undefined;
   return { 'aria-label': label };
} 
```

packages/design-system/src/components/StepList/StepList.jsx
```diff
@@ -74,16 +74,19 @@ StepList.propTypes = {
   /**
    * A template string for the aria-label describing a step's actions where
    * the substring `%{step}` is replaced with that step's `heading`.
+   * To remove `aria-label` from a step's actions,  set `actionsLabelText=""`
```
bernardwang, line 77: I don't know if this is necessary, most of the time we don't want users to remove the aria-labels, so I think its enough to provide an option without necessarily calling it out that overtly. The fact that it's not marked as required might be enough

packages/design-system/src/components/StepList/Step.jsx
nichia: That's great! Looks totally cleaning!

packages/design-system/src/components/StepList/StepList.jsx
nichia: Yes, that works too.

packages/design-system/src/components/StepList/Step.jsx
nichia: I thought committing the the change with a function helper would give me Eslint error, but it didn't. Which is great, but I don't understand why I get Eslint error when I add helper methods in other react function components?

PR #712, opened 2020-06-02
[WNMGDS-459] Add red border to Dropdown with error message

nichia:
### Changed
- Add red border to dropdown with error message
- Add example of dropdown with error message

### How to test
- Run the design system site locally and check that dropdown with error message is displayed with red border

### DIFF
--- a/packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
+++ b/packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
@@ -40,7 +40,7 @@ ReactDOM.render(
       options={dropdownOptions}
       defaultValue={'1'}
       errorMessage="Error message example"
-      label="Dropdown example"
+      label="Error dropdown example "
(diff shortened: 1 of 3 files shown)


packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
```diff
@@ -36,6 +36,13 @@ ReactDOM.render(
       label="Medium dropdown example"
       name="medium_dropdown_field"
     />
+    <Dropdown
+      options={dropdownOptions}
+      defaultValue={'1'}
+      errorMessage="Error message example"
+      label="Dropdown example"
```
line47, line 43: This looks good, one change that would be helpful would be to update the label to be `Error dropdown example` to be consistent with the naming convention

nichia: Done. Thanks for picking that up.

PR #713, opened 2020-06-03
[WNMGDS-457] Change HelpDrawerToggle to use inline instead of inline-block

### LINKED ISSUE #689: HelpDrawerToggle uses inline-block rather than inline when inline prop set to true
**Describe the bug**

The `HelpDrawerToggle` DS component has a prop named ‘inline’ that is supposed to set the container span to “display inline or block”, but it adds `display: inline-block`; when `<HelpDrawerToggle inline=true>`. This can cause display issues when the `HelpDrawerToggle` is a string of text in a paragraph element, particularly on small/mobile screen sizes.

**To Reproduce**

Create a `<HelpDrawerToggle inline=true>` with a sentence with of text as the link and place it in a paragraph with other text. See first screenshot.

We are able to override the default style to force `display: inline !important;`, but it seems like `< HelpDrawerToggle inline=true>` should use inline rather than inline-block.

**Expected behavior**

We expect that `<HelpDrawerToggle inline=true>` would apply the `display: inline;` style to the element, not `display: inline-block;`.

**Screenshots**

Using Chrome DevTools at mobile screen sizes

**What we get:**
localhelp test healthcare gov_(iPhone 5_SE)

**What we expected:**
localhelp test healthcare gov_(iPhone 5_SE) (1)

 - OS: macOS
 - Browser: Chrome
 - Version: 81
Using the responsive design tools in Chrome DevTools to size to screen widths <768px 

**Smartphone (please complete the following information):**
 - Device: N/A
 - OS:  N/A
 - Browser  N/A
 - Version  N/A

**Additional context**

nichia:
### Changed
- Change to use inline style when `inline` prop is set for `<HelpDrawerToggle>` component

### How to test
- Update HelpDrawer.example.jsx to replace this section `<HelpDrawerToggle ... </HelpDrawerToggle>` with below code
`        <p>
          Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt
          ut labore et dolore magna aliqua. <span> </span>
          <HelpDrawerToggle
            inline
            helpDrawerOpen={this.state.showHelp}
            showDrawer={() => this.handleDrawerOpen()}
          >
          Toggle the help drawer Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua
          </HelpDrawerToggle>
          .
        </p>`
- Run the design system site locally and check that the drawer text link stays on the same line as the paragraph.

### Fixed
- Fixes #689

### DIFF
--- a/packages/design-system-docs/src/pages/utilities/display-visibility/_display-visibility.docs.scss
+++ b/packages/design-system-docs/src/pages/utilities/display-visibility/_display-visibility.docs.scss
@@ -11,9 +11,9 @@ Display
 
 > The display CSS property specifies the type of rendering box used for an element. In HTML, default display property values are taken from behaviors described in the HTML specifications or from the browser/user default stylesheet. &mdash; [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
 
-**Format**: `.ds-u-display--[inline|block|inline-block|flex|none]`
+**Format**: `.ds-u-display--[block|flex|inline|inline-block|none]`
 
-**Responsive format**: `.ds-u-[breakpoint]-display--[inline|block|inline-block|flex|none]`
+**Responsive format**: `.ds-u-[breakpoint]-display--[block|flex|inline|inline-block|none]`
 
 Markup: display.example.html
 
--- a/packages/design-system-docs/src/pages/utilities/display-visibility/_display-visibility.docs.scss
+++ b/packages/design-system-docs/src/pages/utilities/display-visibility/_display-visibility.docs.scss
@@ -11,9 +11,9 @@ Display
 
 > The display CSS property specifies the type of rendering box used for an element. In HTML, default display property values are taken from behaviors described in the HTML specifications or from the browser/user default stylesheet. &mdash; [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
 
-**Format**: `.ds-u-display--[block|inline-block|flex|none]`
+**Format**: `.ds-u-display--[inline|block|inline-block|flex|none]`
 
-**Responsive format**: `.ds-u-[breakpoint]-display--[block|inline-block|flex|none]`
+**Responsive format**: `.ds-u-[breakpoint]-display--[inline|block|inline-block|flex|none]`
 
 Markup: display.example.html
 
--- a/packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
+++ b/packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
@@ -12,7 +12,7 @@ export class HelpDrawerToggle extends React.PureComponent {
   }
 
   render() {
-    const blockInlineClass = `ds-u-display--${this.props.inline ? 'inline-block' : 'block'}`;
+    const blockInlineClass = `ds-u-display--${this.props.inline ? 'inline' : 'block'}`;
     /* eslint-disable jsx-a11y/anchor-is-valid */
     return (
       // Use a <span> since a <div> may be invalid depending where this link is nested
--- a/packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
+++ b/packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
@@ -32,15 +32,27 @@ export class HelpDrawerToggle extends React.PureComponent {
 
 /* eslint-disable react/no-unused-prop-types */
 HelpDrawerToggle.propTypes = {
-  /** Whether or not the Help Drawer controlled by this toggle is open or closed. This value is used to re-focus the toggle that opened the drawer when the drawer closes. */
+  /**
+   * Whether or not the Help Drawer controlled by this toggle is open or closed.
+   * This value is used to re-focus the toggle that opened the drawer when the drawer closes.
+   */
   helpDrawerOpen: PropTypes.bool.isRequired,
+  /**
+   * The HelpDrawerToggle content
+   */
   children: PropTypes.node.isRequired,
-  /** Additional classes for the toggle button anchor element */
+  /**
+   * Additional classes for the toggle button anchor element.
+   */
   className: PropTypes.string,
-  /** Adds display inline or block style to the HelpDrawerToggle */
+  /**
+   * Adds `display: inline` to the HelpDrawerToggle.
+   */
   inline: PropTypes.bool,
-  /** This function is called with an id that the toggle generates. It can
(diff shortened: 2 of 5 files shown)


nichia:
Hey @dcloud, I've updated the code to apply `display: inline;` instead of `display: inline-block;` however, I'm not able to replicate the issue, so i cannot really test that the change has fixed the line break. 

The example you included in the issue #689 is clear and helpful. Can you also show the html code for this example?

dcloud:
From the [FLH Results page](https://localhelp.healthcare.gov/#/results?lat=36.028416&lng=-78.92828159999999&state=NC) (which currently has a style override to get `inline !important`):

```
<div class="ds-u-margin-bottom--2 flh-c-help-drawer-toggle"><strong>Agents/brokers and assisters </strong>are trained and certified by the Marketplace to provide application help. <span class="ds-u-display--inline-block"><a href="javascript:void(0);" class="qa-help-drawer-toggle">Learn about the differences between agents/brokers and assisters</a></span>.</div>
```

nichia:
Thanks @dcloud - after changing to `display: inline`, the helpdrawer text link now stays on the same line as the paragraph. However, it has removed the space separator ([see attached](https://drive.google.com/drive/u/0/folders/0AAq79C_GQhE4Uk9PVA)). Did you come across the same issue and had to find another way to add a space before the helpdrawer text link?

whatbird:
@nichia unless i'm misunderstanding, the missing space is just due to good ol' HTML whitespace rendering.

it may not be the responsibility of the HelpDrawer to apply the surrounding whitespace, and a preceding `{" "}` or additional CSS should solve the spacing issue.

nichia:
Thanks @dcloud @whatbird, looks like I'm good to go with it now.

packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
```diff
@@ -12,7 +12,7 @@ export class HelpDrawerToggle extends React.PureComponent {
   }
 
   render() {
-    const blockInlineClass = `ds-u-display--${this.props.inline ? 'inline-block' : 'block'}`;
```
bernardwang, line 15: We should rename this class to be `inlineClass` as well

packages/design-system/src/styles/utilities/_display-visibility.scss
```diff
@@ -9,6 +9,10 @@
   display: inline-block !important;
 }
 
+.ds-u-display--inline {
```
bernardwang, line 12: Usually we don't want to add an additional utility for this type of ticket, however in this case it seems to make sense. Because we are adding, we should also update the documentation and examples for the display/visibility utility page

nichia: Good pick up. Done.

packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
```diff
@@ -12,11 +12,11 @@ export class HelpDrawerToggle extends React.PureComponent {
...
+    const inlineClass = `ds-u-display--${this.props.inline ? 'inline' : 'block'}`;
     /* eslint-disable jsx-a11y/anchor-is-valid */
     return (
       // Use a <span> since a <div> may be invalid depending where this link is nested
-      <span className={blockInlineClass}>
+      <span className={inlineClass}>
         <a
           href="javascript:void(0);"
           className={this.props.className}
```
line47, line 22: I'm thinking it would be good to update the `inline` prop description to be something like: `Applies an inline class to the toggle` 

The current description is a bit confusing

packages/design-system-docs/src/pages/utilities/display-visibility/_display-visibility.docs.scss
```diff
@@ -11,9 +11,9 @@ Display
 
 > The display CSS property specifies the type of rendering box used for an element. In HTML, default display property values are taken from behaviors described in the HTML specifications or from the browser/user default stylesheet. &mdash; [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/display)
 
-**Format**: `.ds-u-display--[block|inline-block|flex|none]`
+**Format**: `.ds-u-display--[inline|block|inline-block|flex|none]`
 
-**Responsive format**: `.ds-u-[breakpoint]-display--[block|inline-block|flex|none]`
```
bernardwang, line 16: Could you move the inline after block? So the ordering is `block`, `inline`, `inline-block` etc

packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
nichia: Done

packages/design-system-docs/src/pages/utilities/display-visibility/_display-visibility.docs.scss
bernardwang: I think `inline` should follow `block` so you can actually see the inline elements behaving together. I think its ok if it looks wonkier when you resize the screen, as thats expected with the inline element. 
Screen Shot 2020-06-04 at 2 16 52 PM

bernardwang: `block` also needs to go first in my opinon. If you look at [USWDS](https://designsystem.digital.gov/utilities/display/) they order it with block first. In fact, I think we should would move flex to second as well.

packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
```diff
@@ -37,7 +37,7 @@ HelpDrawerToggle.propTypes = {
   children: PropTypes.node.isRequired,
   /** Additional classes for the toggle button anchor element */
   className: PropTypes.string,
-  /** Add display inline or block to parent span */
+  /** Adds display inline or block style to the HelpDrawerToggle */
```
bernardwang, line 40: Personally I prefer:
> Adds `display: inline` to the `HelpDrawerToggle`

It doesnt actually add `block` style as thats the default.

packages/design-system-docs/src/pages/utilities/display-visibility/_display-visibility.docs.scss
nichia: Okay, let's take a look at the change to this order - block|flex|inline|inline-block.

PR #714, opened 2020-06-03
[WNMGDS-254] Move the aria-live attributes to checkedChildren element

### LINKED ISSUE #484: Allow ariaRegion in Choice components to be toggled via prop
## Background
Over on the [CMS HITECH APD project](https://github.com/18F/cms-hitech-apd/), we ran into [a rather hilarious issue](https://github.com/18F/cms-hitech-apd/pull/1731) where, immediately upon page load, our app announces all non-empty ARIA regions to screen readers. (I believe it has something to do with our wrapper architecture not being seen as a new page load, but as a set of new components mounted into an existing page.) 

We managed to fix this for our own components, but the CMS Design System's Choice component with conditional children wraps ARIA regions around those children by default, causing [this bug.](https://github.com/18F/cms-hitech-apd/issues/1743) 

## Our temporary workaround
We fixed this in our project by [writing a wrapper](https://github.com/18F/cms-hitech-apd/pull/1760) around the Choice component that brute-force removes the ARIA region (content still is announced to a screen reader, at least in Safari with VoiceOver).

## Proposed change
The wrapper solution is clunky and introduces unnecessary overhead. It would be really nice if we could just pass something like an `ariaRegion={false}` prop to the component to toggle it.

nichia:
### Changed
- Move the aria-live attributes to checkedChildren element
- Reference https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions

### How to test
- Run the design system site locally and inspect the html elements to ensure that aria-live attributes is shown for checked children element only.

### Fixed
Fixes #484


nichia:
@davidakennedy I've moved the aria-live attributes to the checkChildren element but I have not done the screen reader testing yet. I did do a visual check that those elements are updated using chrome dev tools. Let me know whether this change is working - if not, we can explore other ways to get aria-live working

bernardwang:
@nichia it looks like some extra commits from the other PR were added to this branch https://github.com/CMSgov/design-system/pull/714/commits. Could you remove them?

nichia:
Okay, done.

nichia:
Hold off from testing - need to make more changes

nichia:
Okay, ready to be tested!

bernardwang:
@nichia I was looking this over again and I'm actually not sure if this solution will fix the original problem, sorry for overlooking this. I think we should ask the original ticket filer for more details on how to reproduce the original error.

nichia:
Sure. Right now I don't know how to replicate the issue too, but from reading the aria-live mdn which I've noted above, seems to indicate we should place aria-live to checkedChildren elements as that's the region we want to expose when a checked button leads to content change. Again, i don't know how to test it yet, but will follow up.

bernardwang:
@nichia I did some more testing and I am able to reproduce actually. If you go to the [checked children example](https://design.cms.gov/example/components.choice.react/) in VO in safari and reload the page you can see it sometimes will read the "Checkbox B - with children" immediately.  If I'm not mistaken, it will read that even if the `Choice` isnt checked on load.

In this branch, think it won't announce onload when the `Choice` isnt checked, and I'm seeing it less announce on load much less frequently when the `Choice` is checked. It definitely looks like an improvement, but @davidakennedy can help confirm. 

@nichia let me know if you are able to reproduce too.

nichia:
With the VO testing that I did on Safari and Chrome, the present Choice components seems to get the same VO notification as the version with changes made for this PR. Also, I'm not able to replicate the issue of repeated aria-live notification on load. @davidakennedy maybe you can replicate the issue or confirm whether we want to keep the change made for this PR as the VO sounds the same as the present version?

davidakennedy:
Good exploration, @nichia! I did some testing on this:

* I wasn't able to replicate the immediate announcement of the aria-live regions in Safari and VoiceOver. @bernardwang Would love to know if you did anything special there?
* The PR version works well in Safari + VoiceOver, JAWS + Chrome, JAWS + IE11. It has issues in NVDA + Firefox, where it doesn't announce the children when they pop down.
* I'd love to be able to reliably reproduce the issue originally reported so we can figure out the exact issue.

I'm going to do some more research on this to see if I can find more clues to the cause, and possible workarounds.

nichia:
@davidakennedy does the aria-live region on current version works for in NVDA + Firefox to announce the children when they pop down? Curious to find out whether moving aria-live region to checkedChildren caused this issue or it's a known issue with NVDA and there might be some work-around for this?

davidakennedy:
>  does the aria-live region on current version works in NVDA + Firefox to announce the children when they pop down?

@nichia Yes, it works as expected currently. This is what I wanted to research a bit. To see if this is a quirk with this screen reader and if there might be another workaround.

bernardwang:
@nichia @davidakennedy I retested this morning and I think I made a mistake last time. I think what happened was I had my focus on the checked children ChoiceList, but then when I refreshed the focus was reapplied to that same element. I mistakenly thought it was the aria-live

nichia:
@davidakennedy - i understand NVDA + firefox runs on windows, and I have a mac. Is there a way I can test this PR using NVDA (or other variations) on my laptop?

davidakennedy:
> Is there a way I can test this PR using NVDA (or other variations) on my laptop?
@nichia It's possible, but takes more setup. You need to have VM software, plus some other stuff. See: https://dequeuniversity.com/mac/windows-screen-readers

line47:
Hey @nichia you might find this document on setting up screen readers helpful 
https://github.com/adhocteam/a11y-snippets/blob/master/worksheets/how-to-set-up-screen-readers.md

davidakennedy:
Hey @nichia, I did some more research and testing on this today. Here's what I found:

Adding the `aria-live` attribute dynamically to a `div`, especially if it's not on the page can/will be problematic. On this page from MDN says:

> Dynamically adding an element with an aria-live attribute or specialized role to the document also won't result in any announcement by assistive technologies (as, at that point, the browser/assistive technologies are not aware of the live region yet, so cannot monitor it for changes).

So I think the fact that some of the browsers/screen readers work this way is a happy accident. 😄 It's probably best if we don't go this route, and try to come up with another solution.

We're still not sure why the way this component is built impacted the original issue in #484. I wonder if we can dig deeper there, or if there other things to try.

nichia:
Thanks @davidakennedy for the research. I agree with the approach to first replicate the original issue  then explore other options to handle that.

bernardwang:
Since we havent had a response in [this thread](https://github.com/CMSgov/design-system/issues/484) and we are unable to reproduce, I'm going to close this

PR #718, opened 2020-06-15
[NO-TICKET] Status badge update

line47:
### Added
- The ability to set status badge for individual page blocks (for example setting status of `Draft` on a react component and not the entire page) 

### Changed
- Changed the `Draft` definition to `Implement with caution` instead of `Do not implement`
- Changed the react component sections to make the word React part of the header, so the components in page now read like **React `<Alert>` Component** 

### Removed
- Removed `Work in progress` badge option from docs as we don't use that or have it documented.

### Header comparison screenshots
Left is current and right is the new updated version
Screen Shot 2020-06-15 at 9 17 26 AM

### Status badge applied at the page block level screenshot

Screen Shot 2020-06-15 at 9 17 38 AM

### DIFF
--- a/guides/WRITING-DOCUMENTATION.md
+++ b/guides/WRITING-DOCUMENTATION.md
@@ -68,7 +68,7 @@ Supported flags:
   - `[NAME]` is the filename of the react example to be displayed
 
 - **`@responsive`** Renders breakpoint toggles for the markup example.
-- **`@status [NAME]`** Displays a status badge. Supported values: `Draft`, `Work in progress`, `Ready`, `Deprecated`.
+- **`@status [NAME]`** Displays a status badge. Supported values: `Draft`, `Ready` and `Deprecated`.
 - **`@uswds [URL]`** Marks the component as a US Web Design System component. Enter the URL so the documentation can link to the corresponding USWDS page.
 
 ### Modifiers
--- a/packages/design-system-docs/src/scripts/components/PageBlock.jsx
+++ b/packages/design-system-docs/src/scripts/components/PageBlock.jsx
@@ -7,6 +7,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 import ReactContent from './ReactContent';
 import Source from './Source';
+import classNames from 'classnames';
 
 class PageBlock extends React.PureComponent {
   markupExamples() {
@@ -42,6 +43,24 @@ class PageBlock extends React.PureComponent {
     );
   }
 
+  statusBadge() {
+    if (this.props.status) {
+      const classes = classNames(
+        'ds-c-badge ds-u-margin-right--1 ds-u-text-transform--capitalize',
+        {
+          'ds-c-badge--success': this.props.status === 'Ready',
+          'ds-c-badge--alert': this.props.status === 'Draft',
+        }
+      );
+
+      return (
+        <div>
+          <span className={classes}>{this.props.status}</span>
+        </div>
+      );
+    }
+  }
+
   description() {
     if (this.props.description) {
       return (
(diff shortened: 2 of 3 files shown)


guides/WRITING-DOCUMENTATION.md
```diff
@@ -68,7 +68,7 @@ Supported flags:
   - `[NAME]` is the filename of the react example to be displayed
 
 - **`@responsive`** Renders breakpoint toggles for the markup example.
-- **`@status [NAME]`** Displays a status badge. Supported values: `Draft`, `Work in progress`, `Ready`, `Deprecated`.
+- **`@status [NAME]`** Displays a status badge. Supported values: `Draft`, `Ready` and `Deprecated`.
```
bernardwang, line 71: I think we should edit these, we've never used `Ready`.

packages/design-system-docs/src/scripts/components/PageBlock.jsx
```diff
@@ -63,33 +91,30 @@ class PageBlock extends React.PureComponent {
     // text below the title + code snippet block. It's hacky, but works.
     if (this.props.hideHeader || this.props.header.match(/---/)) return;
 
-    const source = this.props.reactComponentPath && (
-      <Source key="reactSource" reactComponentPath={this.props.reactComponentPath} />
-    );
+    const subheader = this.props.reactComponentPath && <span key="subheader">React&nbsp;</span>;
```
bernardwang, line 94: Personally I'm not a big fan of adding the "React component" wording. I think it will be a bit clunky and repetitive looking when there are multiple, like in `Tab` or `ChoiceList`. I would prefer to bring back the tabs we used to have to separate into html, react, and guidance sections. Of course this might not be a big priority right now, and I think ideally we have a designer look at our design in general, so I would prefer to keep the UI here same if possible

packages/design-system-docs/src/scripts/components/PageBlock.jsx
```diff
@@ -42,6 +43,24 @@ class PageBlock extends React.PureComponent {
     );
   }
 
+  statusBadge() {
+    if (this.props.status) {
+      const classes = classNames(
+        'ds-c-badge ds-u-margin-right--1 ds-u-text-transform--capitalize',
+        {
+          'ds-c-badge--success': this.props.status === 'Ready',
```
bernardwang, line 51: I'm wondering if we should update these colors too. I don't think we've ever used ready, and we should add a color for Deprecated which is just the default grey right now. Maybe "Draft" can be `ds-c-badge--warn` and "Deprecated" can be `ds-c-badge--alert`.
This should be updated in `statusBadge()` line 12 in `PageHeader.jsx` as well. I would even add a comment to both instances of `statusBadge()` explaining how they are both copies of the same code so in the future we wont forget to change both.

guides/WRITING-DOCUMENTATION.md
bernardwang: This would also be a good opportunity to update the comment on line 87 in `processKssSection`

bernardwang:
Currently adding a `@status` modifier to the page header will also result in a badge being added to the first `PageBlock`. I'm not sure if theres an easy way to remove that badge besides adding logic in `PageBlock` to only show the status badges when there's React. This leaves us in a pretty confusing state where theres `@status` modifiers for the header, for react page blocks, but not for other HTML page blocks.

I feel like this is another case where the ideal solution would be to get a designer's input and rethink this page, and in the meantime it might be simpler to just keep the original behavior. I'm definitely still open to this change, and I want to hear others thoughts on this.

Screen Shot 2020-06-15 at 10 40 26 AM

guides/WRITING-DOCUMENTATION.md
line47: I'm not following this section? 

```HTML
<button class="ds-c-button {{modifier}}">Button label</button>
```

line47:
Closing for now - possibly revisiting in the future

PR #721, opened 2020-06-15
[WNMGDS-511] Fix lint scripts for Child DS

bernardwang:
### About
eslint wasnt working for child DSs, because [eslint plugins need to be installed relative to the project](https://eslint.org/docs/developer-guide/shareable-configs#publishing-a-shareable-config) using `@cmsgov/design-system-scripts`. Ordinarily this would be done by defining every eslint plugin as a peer dep. However in my testing I found that eslint dependencies just work when you install them via an eslint config package (i.e. `eslint-config-nava` or the original `@cmsgov/eslint-config-design-system`).

Because I didn't want to add 5-7 eslint plugins as peer deps, and I was also concerned about the child DS's ability to customize their linting, I decided to bring back the eslint and stylelint config packages.

The implication is that linting in child design systems will work more similarly to how they currently work in site packages. Instead of providing a linting script that handles all the configs, contains all the needed dependencies, and just works, the new linting script will be pretty minimal, and just use the prettier, stylelint & eslint configs and versions defined in the project. We will provide a set of recommended base configs in the form of `@cmsgov/eslint-config-design-system` & `@cmsgov/stylelint-config-design-system` (we don't provide a `prettier` config because the only option we have is using single quotes). 

TLDR: the updated linting script is just a nicer, shorthand way to run prettier, stylelint, and eslint in one go. We provide recommended eslint and stylelint configs packages, but its up the child DS to use them and define their configs.

### Summary
- Restore `@cmsgov/eslint-config-design-system`
- Restore `@cmsgov/stylelint-config-design-system`
- Update `@cmsgov/design-system-scripts`
   - move `eslint`, `stylelint`, `prettier` to peer deps
   - Simplify lint scripts by relying on default config file functionality, rather than handling it all inside `@cmsgov/design-system-scripts`

Here's what a child DS will look like after this PR:
Screen Shot 2020-06-17 at 3 59 07 PM

### How to test
- Run `yarn lint` on the main repo to ensure its working the same
- Run `yarn lint` on [child-design-system-example.zip](https://github.com/CMSgov/design-system/files/4787062/child-design-system-example.zip) and ensure it is working
- Compare to running `yarn lint` with the master, and you should get [this error](https://github.com/CMSgov/design-system/pull/672#issuecomment-618422317)


line47:
linting is working for both the core and child design system for me.

However, I'm getting something odd for the child design system.
The first time I ran the lint I got errors 
```
6 Sass files linted in ./
[09:49:40] 

docs/src/example.scss
 1:9  ✖  Expected single quotes   string-quotes
 3:9  ✖  Expected single quotes   string-quotes

docs/src/index.scss
 1:9  ✖  Expected single quotes   string-quotes
 3:9  ✖  Expected single quotes   string-quotes

src/styles/index.scss
 1:9  ✖  Expected single quotes   string-quotes
 3:9  ✖  Expected single quotes   string-quotes

src/styles/components/_index.scss
 1:9  ✖  Expected single quotes   string-quotes
 2:9  ✖  Expected single quotes   string-quotes



28 Files formatted with Prettier in ./
10 JS files linted in ./
✨  Done in 11.66s.
```

But single quotes are being uses in the scss files so those errors were not correct.
```
@import '@cmsgov/design-system-docs/src/example.scss';
@import '../../src/styles/index.scss';
```


When I ran a second time everything passed.

```
6 Sass files linted in ./
28 Files formatted with Prettier in ./
10 JS files linted in ./
✨  Done in 11.30s.
```

bernardwang:
@line47 Ah thats because Prettier formatted it afterwards.

nichia:
Here is the `yarn lint` errors listed on child ds: 
```
/Users/tchutnichia/Code/Support/child-design-system-example-lint/node_modules/@cmsgov/design-system-scripts/node_modules/eslint/lib/cli-engine/cascading-config-array-factory.js:335
            throw error;
            ^

Error: Failed to load config "@cmsgov/eslint-config-design-system" to extend from.
Referenced from: /Users/tchutnichia/Code/Support/child-design-system-example-lint/.eslintrc.js
```

bernardwang:
@nichia did you make sure the main `design-system` repo had the right branch when you tested the child ds? You should also try removing `node_modules` from the child DS and reinstalling. Let me know if you arent able to get it to work

nichia:
Tried again, and `yarn lint` on Core and Child DS tested successful, it did not give error messages mentioned above. `yarn lint` on Child DS tested correctly in picking up the problems of Multiple exports for name `Button`. Great work!

bernardwang:
@pwolfert i took a closer look at what create-react-app does, they use eslint-loader with webpack to handle this, but i think we might be able to get something closer to that in the future, and be able to remove the peer deps while also allowing for custom linking config. I’m going to merge this one as is, but in the future i’ll take another look put up a PR. Either way we are going to need those config packages, i think it’s just a matter of having design-system-scripts import them properly

PR #724, opened 2020-06-18
[WNMGDS-503] Add ability to disable/enable robust gold focus styles

line47:
### Added
- Boolean variable `$ds-include-focus-styles` to set to activate the robust gold focus styles.
- Added USWDS focus style variables to support original focus implementation 
- By default the robust gold focus styles are disabled 

### Changed
- Changed the focus styles to use if/else statements to activate and deactivate the robust gold focus styles.

**See the PR where the robust gold focus styles were added.** https://github.com/CMSgov/design-system/pull/510

### How to test
- Run the site locally 
- Add the `$ds-include-focus-styles` variable in a stylesheet like `packages/design-system/src/styles/index.scss` and set to `true` 
- You should see the robust gold focus styles (see the button example screenshot) 

Screen Shot 2020-06-18 at 4 36 44 PM


davidakennedy:
@line47 I remember having a conversation related to this awhile back. Are we providing an option to disable the focus styles because teams are having trouble implementing it or they're conflicting in some way? Or something else?

Just curious if there's something we can improve?

line47:
@davidakennedy the reasoning behind this update was that teams using the design system were less likely to update to the version where the robust gold styles were introduced. It was a stylistic departure that they were not interested in adopting.

line47:
You are right @nichia the default and focused buttons don't have much visual difference. 

However, I think any style updates to the focus styles would happen in a follow on ticket.

line47:
I was also wondering where we could document this enable/disable feature of the focus styles somewhere on the site?

It seems we could benefit from having a page like this 
https://designsystem.digital.gov/documentation/settings/

Maybe we could use this page ?
https://design.cms.gov/guidelines/themes/

thoughts @bernardwang @nichia ?

nichia:
- https://design.cms.gov/guidelines/themes/ would be a good place to document this feature.
- What do you think of either adding notes about `focus` style not  enabled or removing the `focus` buttons from the examples if the robust gold focus styles is disabled as the visual styles would be the same as `default` buttons?
- Another thought - if the focus styles is present in the core ds releases and child ds, would it be better to keep the focus styles enabled as a default? That way this change will not affect existing adopters and the option to disable is available for systems that choose not to adopt this feature.

line47:
> What do you think of either adding notes about focus style not enabled or removing the focus buttons from the examples if the robust gold focus styles is disabled as the visual styles would be the same as default buttons?

I think leaving the focus buttons in place is OK as it shows that there is not much visual difference. In addition, we will be circling back around to revisit the implementation in the future. 

> Another thought - if the focus styles is present in the core ds releases and child ds, would it be better to keep the focus styles enabled as a default? That way this change will not affect existing adopters and the option to disable is available for systems that choose not to adopt this feature.

I don't know any teams that are using release 3.6.0 or greater (when the focus styles were introduced) - I checked both the healthcare and medicare repos 

- https://github.cms.gov/CMS-WDS/ds-healthcare-gov/blob/master/package.json
- https://github.cms.gov/MedicareGov/ds-site-package/blob/master/package.json

bernardwang:
@line47 most teams arent consuming update via the site packages, they just update their version of core. site packages are not a good way of telling what individual teams are using unfortunately.

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -22,33 +22,35 @@
...
-  }
-
-  // Removes inner outline on buttons in Firefox.
-  &::-moz-focus-inner {
-    border: 0;
+    // Removes inner outline on buttons in Firefox.
+    &::-moz-focus-inner {
+      border: 0;
+    }
   }
```
bernardwang, line 40: I think there should be an else here
[original](https://github.com/CMSgov/design-system/pull/510/files#diff-08998df4528f2cb9cfdc571c93ad3ce6L22)

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -135,27 +139,28 @@
...
+  @if $ds-include-focus-styles {
+    &:focus,
+    &.ds-c-button--focus {
+      background-color: $focus-color;
+      border-color: transparent;
+      box-shadow: $focus-shadow-link-inverse;
+      color: $color-base;
+      opacity: 1;
+    }
   }
```
bernardwang, line 151: We need an else statement here for [this](https://github.com/CMSgov/design-system/pull/510/files#diff-08998df4528f2cb9cfdc571c93ad3ce6L100)

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -135,27 +139,28 @@
     border-color: rgba($border-color-inverse, 0.8);
```
bernardwang, line 139: I'm not actually sure why `ds-c-button--active` is included in this statement, when we have [this](https://github.com/CMSgov/design-system/pull/724/files#diff-ebbcedeff6d7cdd3db8f187464ffad26R153) a few lines down. It might be a bug, and we can probably remove those active selectors

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -1,26 +1,21 @@
 @import '../settings/index.scss';
-
```
bernardwang, line 2: Were all these new line deletes from your code editor?

packages/design-system/src/styles/components/_Dialog.scss
```diff
@@ -43,14 +42,17 @@ $dialog-spacer: $spacer-4;
...
-
-  &:focus {
-    background-image: url('#{$image-path}/close.svg');
+  @if $ds-include-focus-styles {
+    &:focus {
+      background-image: url('#{$image-path}/close.svg');
+    }
+  }
+  @else {
+    background-image: url('#{$image-path}/close--primary-darker.svg');
```
bernardwang, line 55: Wow I didn't know we even had a separate svg just for this. Ideally this wouldn't be background-image, so that we can actually style the svg instead of including a separate image

packages/design-system/src/styles/settings/variables/_color.scss
```diff
@@ -84,6 +84,11 @@ $color-base-inverse: $color-white !default;
 $color-muted-inverse: #bac5cf !default;
 
 // Focus
+// Default focus styles
+$color-focus: #3e94cf !default;
```
bernardwang, line 88: I think we should still probably remove these

packages/design-system/src/styles/settings/variables/_shadow.scss
```diff
@@ -8,3 +8,8 @@ $shadow-base-blur-radius: 4px;
 $shadow-base-color: rgba(0, 0, 0, 0.25);
 $shadow-base: $shadow-base-offset-x $shadow-base-offset-y $shadow-base-blur-radius
   $shadow-base-color;
+
+// Inherited from USWDS
+$uswds-box-shadow: 0 0 2px $shadow-base-color !default;
+$uswds-focus-shadow: 0 0 3px $color-focus, 0 0 7px $color-focus !default;
```
bernardwang, line 14: I would also prefer to remove these. The few places that use these variables should just use the hardcoded value, and ideally using the `$focus-color` variable

packages/design-system/src/styles/settings/variables/_build.scss
```diff
@@ -8,3 +8,5 @@ $ds-include-base-html-rulesets: true !default;
 
 // Set $asset-pipeline to true if you're using the Rails Asset Pipeline
 $asset-pipeline: false !default;
+
+$ds-include-focus-styles: false !default;
```
bernardwang, line 12: We should probably add a comment here to explain this variable:

```
// Set $ds-include-focus-styles to enable the beta CMSDS focus styles. These styles can be further customized via the `$focus-color` variable and `focus-text` mixin
```

packages/design-system/src/styles/settings/variables/_build.scss
```diff
@@ -8,3 +8,5 @@ $ds-include-base-html-rulesets: true !default;
 
 // Set $asset-pipeline to true if you're using the Rails Asset Pipeline
 $asset-pipeline: false !default;
```
bernardwang, line 10: Not related to focus styles, but we can probably remove this $asset-pipeline variable. I dont think we were really supporting it to begin with, it might have been leftover from uswds

bernardwang:
Not related to this PR, but I think the `tab.example.html` is broken, the image paths should be corrected to: `{{root}}/images/symbols`

packages/design-system/src/styles/components/_Button.scss
line47: For which portion of code? 
it looked to me like this block of code was there before and after the focus styles

```
 &:active,
  &.ds-c-button--active {
    border-color: $color-primary-darkest;
    color: $color-primary-darkest;
  }
```

packages/design-system/src/styles/components/_Button.scss
line47: Ah good catch, I'm thinking we remove this block from below 
```
&:active,
  &.ds-c-button--active {
    border-color: rgba($border-color-inverse, 0.6);
    color: rgba($color-base-inverse, 0.6);
  }
```

packages/design-system/src/styles/components/_Button.scss
line47: It looked to me like this block of code should remain regardless of if the focus styles are enabled or not.
```
  &:active,
  &.ds-c-button--active {
    border-color: rgba($border-color-inverse, 0.6);
    color: rgba($color-base-inverse, 0.6);
  }
```

bernardwang:
> @davidakennedy the reasoning behind this update was that teams using the design system were less likely to update to the version where the robust gold styles were introduced. It was a stylistic departure that they were not interested in adopting.

@davidakennedy Some more context here, we heard from MCT that they aren't able to adopt any focus style change (not this one, or even USWDS's focus styles) because they are in the process of visual alignment with another team not using the DS. Prior to this PR there wasn't a good way to turn off these global styles (this was an oversight on our part I think), so our priority is to not block MCT or other teams from adopting future versions of the DS.

I think the lesson here is that because focus styles changes are global, highly visible, and usually aren't selectively implemented like components, they are difficult to be adopted by individual teams. I think we may have jumped the gun originally, as focus styles probably need to be implemented with top down guidance and prioritization from CMS in order for them to be adopted across an ecosystem. 

Unfortunately there's other problems with this, as not all teams are using the design system as a dependency so even if CMS did want this to happen it would probably be difficult. In the meantime, I think our approach is to disable the new focus styles by default, iterate on making the styles more customizable (i.e. adopting simpler implementation like USWDS, using focus mixins that can be overriden, etc), and then later push this forward with more support from CMS.

packages/design-system/src/styles/components/_Choice.scss
line47: This was from my local code editor when running some linting locally.

packages/design-system/src/styles/components/_Dialog.scss
line47: Yeah, it's not consistent with how we add images to the button component https://design.cms.gov/components/button/

I'll add a new issue to update the modal button to use an SVG instead of a background image

packages/design-system/src/styles/settings/variables/_color.scss
line47: I was thinking we should keep them because they are used when the gold focus styles are disabled.

bernardwang:
> * Another thought - if the focus styles is present in the core ds releases and child ds, would it be better to keep the focus styles enabled as a default? That way this change will not affect existing adopters and the option to disable is available for systems that choose not to adopt this feature.

I think we could call out the new setting variable in the documentation for teams that want to continue to adopt the beta focus styles (assuming we have the variable off by default).

packages/design-system/src/styles/components/_Dialog.scss
bernardwang: Ah, I already filed one too: https://jira.cms.gov/browse/WNMGDS-519
Maybe they can be combined. 
I did want to add, the reason why the button examples are different is because the icons are optional, and user that want an icon will add a svg manually. In this case, the close button is by default and the user shouldn't really need to customize. From a user perspective, its slightly easier to have it implemented with a `background-image`. However, on the other side for certain icons that might need to be customized, or for performance sake (not having to include multiple svg files), it could be advantageous to use an svg element. So i think it will have to be a case by case basis, and we will probably also have to consider accessibility impact, like if we move the svg to markup, do we need to include `alt-text`?

bernardwang:
> Not related to this PR, but I think the `tab.example.html` is broken, the image paths should be corrected to: `{{root}}/images/symbols`

@line47  Actually, you can just ignore this. This is happening in a bunch of files and I can put up a quick PR to fix it.

nichia:
LGTM

packages/design-system-docs/src/pages/startup/migrating-v2.md
```diff
@@ -64,6 +64,8 @@ In CMSDS v2, it will be changed to
 
 Imports to `@cmsgov/design-system-layout` or `@cmsgov/design-system-support` are now included in the main `scss` entry point. If you would like to only import specific styles, see the `@cmsgov/design-system` [`README.md`](https://github.com/CMSgov/design-system/blob/master/packages/design-system/README.md#file-structure) for more info on our SASS architecture.
```
bernardwang, line 65: Could we add a section on focus styles? Maybe explaining how it was previously released in v3 but it was moved under a flag because it is still under beta development. I feel like focus styles are pretty big so it makes sense to add some sort of explanation

line47: Yeah, I'll add some info explaining 👍

packages/design-system/src/styles/components/_Choice.scss
bernardwang: hm... thats weird, because i dont think the linter should have those rules, or at least I'm not getting those changes when i lint. are you running `yarn lint`?

line47: I think that might have been my local linter and not `yarn lint` :-/

bernardwang:
Screen Shot 2020-06-25 at 2 16 15 PM
Hmm not entirely sure why this is the case, but the focus style makes the button smaller, I dont think thats how it was before and it should be consistent with other focus styles which are outside the element

packages/design-system/src/styles/settings/variables/_build.scss
bernardwang: @line47 small clean up here

packages/design-system/src/styles/components/_Choice.scss
bernardwang: do you mind adding them back? it makes the spacing inconsistent in a lot of areas and it makes the PR harder to merge in general

packages/design-system/src/styles/components/_Button.scss
bernardwang: I don't think thats the best approach, as we want the `active` style to be different than hover. I think we want to do the opposite and remove line 136 and 138

packages/design-system/src/styles/settings/variables/_color.scss
bernardwang: Yeah you are right, and this actually made me realize something, in the few places that use `$color-focus`, we dont use the default browser focus styles. Was this how it was originally, or did previously everywhere used `$color-focus` and our own custom focus shadow? @davidakennedy do you happen to remember?

bernardwang: A comparison: 
Screen Shot 2020-06-25 at 2 25 44 PM
Screen Shot 2020-06-25 at 2 25 38 PM

packages/design-system-docs/src/pages/startup/theming.md
```diff
@@ -38,7 +38,31 @@ Sass variables are documented on the relevant documentation pages, and are defin
 - [Colors]({{root}}/styles/color/)
 - [Spacing]({{root}}/styles/spacing/)
 - [Type sizes]({{root}}/styles/typography/)
-- You can also [browse all Sass variable files on GitHub](https://github.com/CMSgov/design-system/tree/master/packages/design-system/src/styles/settings/variables).
+
+You can also [browse all Sass variable files on GitHub](https://github.com/CMSgov/design-system/tree/master/packages/design-system/src/styles/settings/variables).
+
+#### Focus settings
+
+<table class="ds-c-table ds-c-table--borderless docs_table">
```
bernardwang, line 46: I really like using the table for CSS variables and other data, but I think this is the first time we use it for build settings. I think it only makes sense if there are multiple entries in the table, or if we include the other build settings too like `$ds-include-base-html-rulesets`, `$font-path`, and `$image-path`. However, this would require rethinking how our docs are organized and it would be better suited for a `Build settings` section.

I think for now it would be easier for us to replace this table with an explanation, and a list of all the focus variables and mixins.

bernardwang: For example, we randomly list the focus variables [here](https://github.com/CMSgov/design-system/pull/510/files#diff-1a9408d59d6706d79301b86f714aa077L54). I think we should move that from `Textfield.docs.scss` to here

packages/design-system-docs/src/pages/styles/color-palette/color-focus.example.html
```diff
@@ -0,0 +1,12 @@
+<% var colors = [ 'focus-color', 'focus-color-inverse', 'focus-border-inverse' ]; -%>
+<div class="">
+  <% colors.forEach(color => { -%>
+  <article class="ds-u-margin-bottom--1">
+    <div
+      class="c-swatch__preview c-swatch__preview--condensed ds-u-radius--circle ds-u-fill--<%= color %>"
+    ></div>
+    <code>$color-<%= color %></code>
```
bernardwang, line 8: The naming here should be updated to just `<%= color %>`

packages/design-system-docs/src/pages/startup/migrating-v2.md
```diff
@@ -64,6 +64,12 @@ In CMSDS v2, it will be changed to
...
 
+See the [theming page]({{root}}/startup/theming/) for more specific information on overriding the default Sass variables or CSS declarations.
+
+### Focus styles
```
bernardwang, line 69: I noticed 2 things:
- I think Focus styles should be a level 2 heading that happens after "Folder structure"
- I think "Fonts and Images" should be a level 3 heading under "Folder structure"

packages/design-system-docs/src/pages/startup/migrating-v2.md
```diff
@@ -64,6 +64,12 @@ In CMSDS v2, it will be changed to
...
+
+We released focus styles in [version 3.6.0](https://github.com/CMSgov/design-system/releases/tag/3.6.0) and with this V2 update we are disabling these gold focus styles by default and placing them in `beta` development as we continue to iterate on them. If you want to continue to use the gold focus styles please see the guidance on the [theming page]({{root}}/startup/theming#focus-settings)on how to enable them.
```
bernardwang, line 71: What do you think about this to give more context?

## Focus styles

CMSDS focus styles will now be considered a `beta` feature, and will be turned off by default. Focus styles can be reenabled by setting the `$ds-include-focus-styles` build variable to true, but these styles will likely change in the future as we iterate on our design. See the [theming page](https://design.cms.gov/startup/theming/#focus-settings) for more information on how to customize or enable focus styles.

### Why the change?
When we first released focus styles in [version 3.6.0](https://github.com/CMSgov/design-system/releases/tag/3.6.0), our goal was to provide a recommended focus style implementation that would incorporate accessibility best practices, much of which was informed by [UKgov's research](https://design-system.service.gov.uk/get-started/focus-states/) and [USWDS's focus state implementation](https://designsystem.digital.gov/). While we spent 4 months developing, iterating, and gathering feedback before the release, we made several mistakes in the process. Due to communication and process shortcomings, our team wasn't able to fully consider how our design would impact existing focus styles, or whether or not our design was too specific for certain products. Our team assumed that the SCSS variables we provided were sufficient to adapt the focus styles for different use cases and we neglected to add an easy way to opt out of the new feature. 

In this release, we worked to correct these mistakes by putting the new styles under a flag and turning them off by default. In the future, we plan on iterating on our default styles to be more generalized and more in line with existing focus styles on CMS products. We will also be adding  additional mixins, variables, and documentation to make it easier for teams to. customize focus styles.

packages/design-system/src/styles/settings/variables/_color.scss
line47: I tried to revert it back to how it was previously - which involved going through DKs PR https://github.com/CMSgov/design-system/pull/510/files

packages/design-system/src/styles/components/_Button.scss
bernardwang: else {
  &:focus,
  &.ds-c-button--focus {
    border-color: $color-primary-darker;
    color: $color-primary-darker;
  }
}

packages/design-system/src/styles/components/_Button.scss
bernardwang:
```
else {
   &:focus,
   &.ds-c-button--focus {
       border-color: rgba($border-color-inverse, 0.8);
       color: rgba($color-base-inverse, 0.8);
   }
}
```

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -103,14 +107,14 @@
   &.ds-c-button--active {
     border-color: transparent;
   }
-
```
bernardwang, line 106: Super small, but this code block should have new lines before and after

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -22,33 +22,35 @@
     border-color: $color-primary-darker;
     color: $color-primary-darker;
   }
+  @if $ds-include-focus-styles {
```
bernardwang, line 25: Same with this code block, should have new lines before and after

packages/design-system/src/styles/base/_link.scss
```diff
@@ -1,4 +1,5 @@
 /* stylelint-disable selector-class-pattern */
+
```
bernardwang, line 2: This can be deleted

packages/design-system-docs/src/pages/startup/theming.md
```diff
@@ -38,7 +38,29 @@ Sass variables are documented on the relevant documentation pages, and are defin
...
+#### Focus style settings
+
+The focus styles are disabled by default. However if you'd ike to use them
+set the `$ds-include-focus-styles` variable to `true`.
+
+**Focus style variables**
+
+- `$focus-color` - Sets background highlight color.
+- `$focus-shadow` - Sets box-shadow on text fields.
+- `focus-shadow-link` - Sets box-shadow on buttons and links.
```
bernardwang, line 53: Missing a $ here

packages/design-system-docs/src/pages/startup/migrating-v2.md
```diff
@@ -46,48 +46,60 @@ The `dist` folder structure has been updated for the `@cmsgov/design-system` npm
     └── scss/             Uncompiled CSS
 ```
 
-### Importing SCSS/CSS
+### Fonts and Images
```
bernardwang, line 49: ## Folder structure
### Importing SCSS/CSS
### Importing Javascript
### Fonts and Images
## Focus styles
## Versioning

packages/design-system-docs/src/pages/startup/theming.md
```diff
@@ -38,7 +38,29 @@ Sass variables are documented on the relevant documentation pages, and are defin
...
+- `$focus-color-inverse`- Sets background highlight color on dark backgrounds.
+- `$focus-shadow-inverse`- Sets box-shadow on text fields on dark backgrounds.
+- `$focus-shadow-link-inverse` - Sets box-shadow on buttons and links on dark backgrounds.
+
+**Mixins**
+
+- `focus-text` - Mixin for links and buttons.
+- `focus-text-inverse` - Mixin for links and buttons on dark backgrounds.
+
+Show the CMSDS focus styles. These styles can be further customized via the [Focus color variables](http://localhost:3000/styles/color/#styles.color.focus) and `focus-text` mixin
```
bernardwang, line 63: I think this is left over from a previous commit and can be removed.

bernardwang: Maybe we could replace this paragraph with this:
```
See the [v2 migration guide](https://design.cms.gov/startup/migrating-v2/#focus-styles) for more background on our focus styles.
```

packages/design-system/src/styles/base/_link.scss
line47: Are you referring to `/* stylelint-disable selector-class-pattern */` here?

bernardwang: Oh i'm talking about the new line, just pointing out the last bits of new line clean up

packages/design-system/src/styles/settings/variables/_color.scss
davidakennedy: Yeah, @line47 is right, @bernardwang. That was one of the biggest changes in my original PR, it applied things everywhere vs. selectively.

PR #725, opened 2020-06-22
[WNMGDS-507] Fix cannot read property 'checked' of null

### LINKED ISSUE #717: [WNMGDS-507]: Choicelist: Cannot read property 'checked' of null
[WNMGDS-507](https://jira.cms.gov/browse/WNMGDS-507)

**Describe the bug**

About once a minute, the MCT frontend records a `Cannot read property 'checked' of null` error in the Choicelist component, at this line: https://github.com/CMSgov/design-system/blob/6b0e84b7f66c8d5d3bbe2272c037e2c938404dcf/packages/design-system/src/components/ChoiceList/Choice.jsx#L50

presumably `this.input` is null when the component is receiving an `unchecked` event, for some reason that I don't know. I see the error happening on both Chrome and Edge in our new relic stats; I've never reproduced it myself.

**To Reproduce**

see below

**Expected behavior**

The choicelist component doesn't throw errors

**Screenshots**
JavaScript Errors 2020-06-12 10-25-20

 - OS: unknown
 - Browser chrome, edge

nichia:
### Changed
- Fix the issue of `Cannot read property 'checked' of null` after a checked parent uncontrolled radio button (with children radio buttons) gets deselected. After that, any subsequent selecting of the children radio buttons will result in the error msg. 

### How to test
- Run the design system site locally and check that Choice component examples works. Specifically check that  `Radio buttons with multi-level children` works as intended and inspect dev tools to ensure that it does not throw errors.

### Fixed
Fixes #717


packages/design-system/src/components/ChoiceList/Choice.jsx
```diff
@@ -25,13 +26,20 @@ export class Choice extends React.PureComponent {
       // Event emitters are only relevant for uncontrolled radio buttons
       if (this.props.type === 'radio') {
         this.uncheckEventName = `${this.props.name}-uncheck`;
-        dsChoiceEmitter.on(this.uncheckEventName, this.handleUncheck.bind(this));
+        dsChoiceEmitter.on(this.uncheckEventName, this.handleUncheck);
```
bernardwang, line 29: Ahhhh! was it binding to an old reference of `this`?

nichia: That took a long time to figure out that unbind from emitter `off` did not work because of `this.handleUncheck.bind(this)`. Once this got moved out, then the addition of unbind of emitter in componentWillUnmount finally works.

nichia: Yes, the `checked` of null issue was due to the emitter containing the old dom references - we need to remove this reference each time the parent radio button is changed which triggers componentWillUnmount

packages/design-system-docs/src/pages/components/ChoiceList/Choice.example.jsx
```diff
@@ -16,13 +16,92 @@ const dropdownOptions = [
...
     options={dropdownOptions}
-    defaultValue={'B'}
+    defaultValue="B"
     label="Dropdown example"
     labelClassName="ds-c-label ds-u-margin-top--0"
     name="dropdown_choices_field"
   />
 );
 
+class ControlledChoiceButton extends React.PureComponent {
```
bernardwang, line 26: I appreciate that you included this controlled example, especially for testing this makes things a lot easier, but I think after all the PR feedback is in we should remove this from the example

packages/design-system/src/components/ChoiceList/Choice.jsx
bernardwang: this looks like a good opportunity to use react hooks, what do you think?

packages/design-system-docs/src/pages/components/ChoiceList/Choice.example.jsx
nichia: Sounds good, after testing, we can revert the Choice.example.jsx back to the original

packages/design-system/src/components/ChoiceList/Choice.jsx
```diff
@@ -13,9 +13,10 @@ export class Choice extends React.PureComponent {
     super(props);
     this.input = null;
     this.handleChange = this.handleChange.bind(this);
+    this.handleUncheck = this.handleUncheck.bind(this);
     this.id = this.props.id || uniqueId(`${this.props.type}_${this.props.name}_`);
 
-    if (typeof this.props.checked === 'undefined') {
+    if (typeof this.props.checked === 'undefined' || this.props.defaultChecked) {
```
bernardwang, line 19: Sorry, this code wont actually work. This will set the component to controlled even if a user passes in `defaultChecked = true`. You should just remove this change I believe

line47:
I was able to test and did not get any issues in the console, great work @nichia 🎉 

I also tested after removing the code @bernardwang [pointed out here](https://github.com/CMSgov/design-system/pull/725/files#discussion_r443739484) and everything still worked as expected.

packages/design-system/src/components/ChoiceList/Choice.jsx
nichia: If Choice was a functional component, then yes, it would be good to use react hooks now. As it is presently a class component, i think it's better to just add the componentWillUnmount lifecycle to clean up the emitter event instead of rewriting as a functional component.

packages/design-system/src/components/ChoiceList/Choice.jsx
```diff
@@ -59,11 +67,11 @@ export class Choice extends React.PureComponent {
 
     if (!this.isControlled) {
       this.setState({ checked: evt.target.checked });
-    }
```
bernardwang, line 62: So the event emitter should only fire for uncontrolled Choices right?

bernardwang: Wasn't the MCT example controlled?

packages/design-system/src/components/ChoiceList/Choice.jsx
bernardwang: Ah right, we cant rewrite this component because it uses refs anyway. I think in the future though in general if we are working on a component and we *can* easily modernize it, we should try to. Like maybe this doesn't have to be done in this PR, but I'm thinking on quick modernization could be to remove instances of `.bind` by adding [arrow functions](https://babeljs.io/docs/en/babel-plugin-proposal-class-properties)

packages/design-system/src/components/ChoiceList/Choice.jsx
```diff
@@ -59,11 +67,11 @@ export class Choice extends React.PureComponent {
...
 
-    if (this.uncheckEventName && evt.target.checked) {
-      // Emit the uncheck event so other radio options update their state
-      dsChoiceEmitter.emitEvent(this.uncheckEventName, [this.id]);
+      if (this.uncheckEventName && evt.target.checked) {
```
bernardwang, line 71: I wonder if this event emitter logic could be more clear. It seems like `this.uncheckEventName` is always the same, and can pretty much be replaced by `${this.props.name}-uncheck`. If we did that, we can remove `this.uncheckEventName` from this if statement and replace it with `this.props.type === 'radio'`. This might make the code more clear that all this event emitter logic only happens for uncontrolled radio buttons

bernardwang: We also don't have to get rid of `this.uncheckEventName`, i'd just prefer not to use it in this if statement

packages/design-system/src/components/ChoiceList/Choice.jsx
nichia: MCT had controlled radio button on the top level. The issue was for when MCT uses the uncontrolled radio button further done the chain.

packages/design-system/src/components/ChoiceList/Choice.jsx
nichia: Great suggestion.

packages/design-system/src/components/ChoiceList/Choice.jsx
bernardwang: and when you reproduced it, you were using uncontrolled right?

nichia: Yes. I tested will controlled & uncontrolled, radio & check buttons. The issue only occurs for uncontrolled radio buttons because of the use of emitter and not cleaning out the emitter on the unmount of the component. Hence, the component is referencing this leftover ref pointing to null value.

PR #728, opened 2020-06-30
Bernard/fix e2e tests

line47:
- Adding e2e test fix to hotfix branch


line47:
only want a small subsection of these changes closing to try another route

PR #731, opened 2020-07-01
[WNMGDS-526] Optimize scripts

bernardwang:
### Summary
After adding support for typescript I found our build scripts to be incredibly slow. This PR is to optimize some of these scripts and reorganize the code to be easier to understand. 

### How to test
First ensure a clean build
- run `git clean -fdx` on the main design system branch
- run `yarn build` on the main design system branch
- run `yarn start` on the child design system 

Test all the build scripts on the core CMSDS and the WIP m.gov child DS
[mgov.zip](https://github.com/CMSgov/design-system/files/4886099/mgov.zip)
In particular, test `yarn start` with the mgov child DS
- Updating React props / Typescript types
- Updating component code will compile code and site
- Updating styles will update site
- Updating example files will update site


packages/design-system-scripts/cli.js
```diff
@@ -164,6 +170,11 @@ yargs
           type: 'array',
           default: ['**/node_modules/**', '**/dist/**'],
         })
+        .option('failAfterError', {
```
bernardwang, line 173: Previously we would have our linting fail on error if the env is set to `TEST`. Now it is just a command line option.

packages/design-system-scripts/gulp/common/constants.js
```diff
@@ -4,13 +4,17 @@ const CORE_SOURCE_PACKAGE = '@cmsgov/design-system';
 const CORE_DOCS_PACKAGE = '@cmsgov/design-system-docs';
 
 const REACT_DATA_DIR = `tmp/data`;
-const REACT_DATA_FILENAME = 'react-doc.json';
-const REACT_DATA_PATH = path.resolve(REACT_DATA_DIR, REACT_DATA_FILENAME);
+const REACT_PROP_DATA_FILENAME = 'react-prop.json';
```
bernardwang, line 7: Split up the react data into 2 files, one for example code and one for prop data. This allows us to watch the two file types separately, and make `yarn start` slightly more efficient

packages/design-system-scripts/gulp/docs/generatePages/generatePages.js
```diff
@@ -1,16 +1,33 @@
-const addReactDocs = require('./addReactDocs');
+// Run babel transforms on src files so we can run JSX scripts in Gulp tasks
+require('@babel/register')({
+  only: [/(@cmsgov\/design-system-docs|design-system\/packages\/([a-z-_]+)\/src|generateDocPage)/],
+  presets: [
```
bernardwang, line 4: Presets and plugins are identical to the core CMSDS default babel.config.js, but we hardcode it so that customized babel configs in child design systems wont affect our doc site compiling

packages/design-system-scripts/gulp/docs/generatePages/generatePages.js
```diff
@@ -73,27 +90,47 @@ function generatedPagesCount(resultGroups) {
   return count;
 }
 
+// Use `changedPath` to only process pages detected from gulp watch
```
bernardwang, line 93: We use the path of the changed file to selectively filter out files that didnt change, reducing the amount of pages that get regenerated during `yarn start`

line47:
**For the core** 
everything worked as expected when running scripts.
There were a few times where this was sent back when running tests

```
A worker process has failed to exit gracefully and has been force exited. 
This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.
```

**For child ds**

Start
Works but saw quite a few parseComponent errors
- parseComponent Unable to generate react component source for /Users/scottweber/Documents/github/mgov/src/components/Card/Card.tsx

- Editing Simple footer doesn’t rebuild and gives parseComponent errors 


Build 
Works but saw quite a few parseComponent errors
- parseComponent Unable to generate react component source for /Users/scottweber/Documents/github/mgov/src/components/Card/Card.tsx

Test
- All tests failed

E2e
-  Command "test:e2e" not found.
Update-snapshots
- Command "update-snapshots" not found.

Lint 
- Works!

bernardwang:
@line47 @nichia I've updated the mgov zip and the testing instructions, it should be working now

bernardwang:
`yarn test:e2e` only works for the core because there arent any e2e tests to run. 


> A worker process has failed to exit gracefully and has been force exited. 
> This is likely caused by tests leaking due to improper teardown. Try running with --runInBand --detectOpenHandles to find leaks.

This is an existing message, its not introduced in this PR

line47:
When i run `yarn start` on the child ds i get the following a bunch of times 
```
babel.transform(code, { filename: 'file.ts', presets: [/* your preset */] });

See https://babeljs.io/docs/en/options#filename for more information.
[15:00:27] react-docgen /Users/scottweber/Documents/github/mgov/node_modules/@cmsgov/design-system/src/components/TextField/TextField.jsx
[15:00:27] react-docgen Error: [BABEL] unknown: Preset /* your preset */ requires a filename to be set when babel is called directly,
```

I also get this a bunch of times for various components 
```
[15:00:28] addReactDocs Invalid react component for node_modules/@cmsgov/design-system-docs/src/pages/components/Alert/_Alert.docs.scss
```

bernardwang:
@line47  are you making sure to run `yarn build` after running `git clean -fdx` on the main design system?

bernardwang:
And fyi, I did notice the `parseComponent Unable to generate react component source` error. It has to do with the provided `githubUrl` matching the name of the folder. In this case, the name of the folder is incorrect (mgov). I'll update the error message to be more informative but I might have to think about a better solution in the future

PR #733, opened 2020-07-08
[WNMGDS-327] Change radio and checkbox checkmark to be created with SCSS instead of image

### LINKED ISSUE #568: Checkbox and Radio: Checkmarks show as clipped in Windows High Contrast Mode
**Describe the bug**
If you're using Windows High Contrast Mode, the checkmarks in the checkbox and radio button component appear cut off or clipped.

**To Reproduce**
1. Enable High Contrast Mode on a Windows machine. Settings > Ease of Access > High Contrast
2. View: https://design.cms.gov/components/choice/
3. See the clipped checkmarks.

**Expected behavior**
The checkmarks should appear in their whole shape like they do normally.

**Screenshots**
Checkmarks in high contrast mode

 - OS: Windows 10
 - Browser: Edge
 - Version:  44.18362.387.0

line47:
This fixes #568

- Updated the checkmark to be created with SCSS, similar to how Gov.uk does it. https://design-system.service.gov.uk/components/checkboxes/
- Removed the white-checkmark.svg as it was only used i the choice component 
- Updated the SCSS and organized it a bit better

**To test** 

- Open up the doc site and visit the choice page
- See that the checkmark is now visible on checked or selected items for both radio buttons and checkboxes

The screenshot from my test (checks are visible)

Screen Shot 2020-07-08 at 3 11 20 PM


davidakennedy:
Looks good in IE 11 and Edge, pre-Chromium. I do get these off breaks in some of the circles and squares around the radio buttons in checkboxes in new Edge.

High contrast mode on Windows 10 showing radio buttons and checkboxes with a break in the borders around them

I don't know if that's an oddity with my machine, or an actual issue. I turned the mode off and on twice, and still got it.

line47:
@davidakennedy does that break show if you view the live design system choice component? 
https://design.cms.gov/components/choice/

davidakennedy:
It does, @line47. So maybe not an issue with this specific PR.

line47:
@davidakennedy yeah I'd rather track as a separate issue to get this one closed out

davidakennedy:
@line47 Makes sense. Want me to open a new issue?

line47:
Sure that'd be great @davidakennedy, thanks!

bernardwang:
Not a huge deal, but the new CSS is slightly different than the font awesome icon. I think we should either replace the font awesome icon in the Icon page, or take a stab at making it more close, ideally both haha.

Old
Screen Shot 2020-07-14 at 12 47 36 PM
New
Screen Shot 2020-07-14 at 12 47 30 PM

bernardwang:
I think there are also [SVG to CSS converters](https://websemantics.uk/tools/svg-to-background-image-conversion/) out there, although I havent tried it yet. I would also comment in the SCSS file where the CSS arrow code is coming from, and why we use it instead of SVG

line47:
That's a good point @bernardwang and I hadn't noticed that the corners on the current checkbox were rounded. 
I feel like the straight version is cleaner and more easily customizable. 

Do the check images have to be the same as the checkboxes and radios?
I don't think they necessarily need to but this would be good to discuss.  

I looked into the SVG to CSS converter and it seemed to do the job but the code it provided would probably be hard to modify. 
```
  background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='20' height='20' viewBox='0 0 216 146'%3E%3Cpath fill='%23fff' d='M168.86 37.966l-11.08-11.08c-1.52-1.52-3.367-2.28-5.54-2.28-2.172 0-4.02.76-5.54 2.28L93.254 80.414 69.3 56.38c-1.52-1.522-3.367-2.282-5.54-2.282-2.172 0-4.02.76-5.54 2.28L47.14 67.46c-1.52 1.522-2.28 3.37-2.28 5.542 0 2.172.76 4.02 2.28 5.54l29.493 29.493 11.08 11.08c1.52 1.52 3.368 2.28 5.54 2.28 2.173 0 4.02-.76 5.54-2.28l11.082-11.08L168.86 49.05c1.52-1.52 2.283-3.37 2.283-5.54 0-2.174-.76-4.02-2.28-5.54z'/%3E%3C/svg%3E");
```

bernardwang:
@line47 It looks like thats a background image which is where we started off with, I didn't look closely enough into the converter I linked. 

Regardless, I don't think any teams are modifying the checks, if they really did they can always override the CSS. If anything, it was easier to override the styles in the past because it used a background image vs a lot CSS. The main issue for me is less about customization, and more about consistency. It feels weird to document our icon as the font awesome check, and then have out implementation use something else, even if it's only slightly different.

I would prefer to first try to update the CSS to make the check slightly rounded to match the font-awesome icon, and if that doesn't work I'm ok with the current icon as long as we document it well in the code and in maybe even on the site

line47:
This is ready for another pass @bernardwang - I used the SVG to image converter to create the CSS for the checkmark to make sure it is identical.

I believe this should work in high contrast mode as well. However, I'm having trouble with my local windows VM. @davidakennedy can you take another look at this?

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -129,12 +114,13 @@ $ds-c-inset-border-width: $spacer-half;
...
 
 // Display a checkmark
+// Using CSS to display the checkmark because background images don't how in high contrast mode on windows.
 .ds-c-choice:checked + label {
-  &::after {
-    opacity: 1;
-  }
   &::before {
     background-color: $color-primary;
+    background-image: url("data:image/svg+xml;charset=utf8,%3Csvg xmlns='http://www.w3.org/2000/svg' width='20' height='20' viewBox='0 0 216 146'%3E%3Cpath fill='%23fff' d='M168.86 37.966l-11.08-11.08c-1.52-1.52-3.367-2.28-5.54-2.28-2.172 0-4.02.76-5.54 2.28L93.254 80.414 69.3 56.38c-1.52-1.522-3.367-2.282-5.54-2.282-2.172 0-4.02.76-5.54 2.28L47.14 67.46c-1.52 1.522-2.28 3.37-2.28 5.542 0 2.172.76 4.02 2.28 5.54l29.493 29.493 11.08 11.08c1.52 1.52 3.368 2.28 5.54 2.28 2.173 0 4.02-.76 5.54-2.28l11.082-11.08L168.86 49.05c1.52-1.52 2.283-3.37 2.283-5.54 0-2.174-.76-4.02-2.28-5.54z'/%3E%3C/svg%3E");
```
bernardwang, line 121: Isnt this still a `background-image`? I thought that thats what caused the original issue.

line47: Yeah you're right I was thinking it might work differently but it doesn't. I think I'm going to go back to the full SCSS check without the rounded corners and update the check icon to match

line47:
When I was looking into updating the check icon to match - I discovered that the icons are not created consistently. When i tried to update the success and success-fill icons they were larger than the original and also didn't have the same spacing around the icon. 

TLDR
I think all the icons here should be revisited and created in the same way. 
I think this should be done in a separate issue.

line47:
to do: 

- [ ] Add info on the icons page explaining how the checkmark for radios and checkboxes is Scss

line47:
@bernardwang added some description to the icons page 👍

PR #736, opened 2020-07-09
[WNMGDS-536] Add more doc content templates

bernardwang:
### Summary
- Added `npmPackage` option
- Added support for `{{npm}}`, `{{github}}`, and `{{name}}` templates

### How to test
- Run the doc site and ensure all the content using the new templates is correct
- Try changing various template values to see the doc site update

### DIFF
--- a/packages/design-system-docs/src/pages/guidelines/accessibility.md
+++ b/packages/design-system-docs/src/pages/guidelines/accessibility.md
@@ -2,7 +2,7 @@
 title: Accessibility
 ---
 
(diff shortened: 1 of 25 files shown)


line47:
@nichia the values are defined in `packages/design-system-scripts/cli.js`

line47:
I was wondering the same thing - 

> There are many other files that {{name}}, {{npm}} and {{github}} can be used to replace their respective strings. These would be good to have done in a separate PR.

PR #740, opened 2020-07-13
[WNMGDS-152] Child design system example

line47:
### Added
- Adding README to the example explaining how to set up a child design system 
- Added source code to get started

### How to test
- go to the examples > child design system
- run `yarn install`
- run `yarn start` 
- ensure the site is running locally and any changes you make to the following rebuild
    - component changes
    - component style changes
    - documentation changes
    - example code changes


bernardwang:
@line47 I took a look at the branch, could we update the custom font to be more realistic? Right now it just applies Anton to everything. So for example picking 2 new fonts, 1 for header and 1 for the body. And just in general styling it so it looks more like something a team would actually want to use. 

I think it would also be useful to add some examples of custom icons or something

line47:
@bernardwang I've updated the fonts, however, I'm having issues with the fonts in `base` building and am unsure why - can we take a look together? 

https://github.com/CMSgov/design-system/blob/WNMGDS-152-child-ds-example/examples/child-design-system/src/styles/index.scss#L14

examples/child-design-system/src/styles/base/_override.body.scss
```diff
@@ -0,0 +1,16 @@
+%ds-base {
```
bernardwang, line 1: nice this is pretty important, most people don’t know about base

examples/child-design-system/src/styles/base/_override.body.scss
```diff
@@ -0,0 +1,16 @@
...
+  font-family: $font-sans;
+  font-size: $base-font-size;
+  line-height: $base-line-height;
+}
+
+.ds-base {
+  @extend %ds-base;
+  color: $color-primary-alt;
```
line47: I couldn't get any of the scss files in `base` dir to do anything

bernardwang: yeah, that's what i was seeing too with my testing. Moving the imports to core around changes things up, but then it causes other overrides to fail. I think in general our override process is super fragile we need to spend some time debugging it. I removed the overrides that didn't seem to work

line47: I agree 💯  we need to look into it deeper. I'm going to add something to the backlog for that

bernardwang:
@line47 Could you please add a `How to test` section?

examples/child-design-system/README.md
```diff
@@ -0,0 +1,103 @@
...
+There are many independent product teams across different contracts and companies that work under the same CMS sites (i.e. Medicare.gov, Healthcare.gov).
+
+These product teams need a shared resource specific to those sites. It’s not possible or desirable for the CMSDS core to cover those site specific case.
+
+This is the reasoning behind creating the CMS design system family which consists of the CMSDS core and child design systems.
+
+## Getting started
+
+1. Download this child design system example folder
+1. Run `yarn install` to install dependencies
```
nichia, line 12: Number the points

line47: Markdown will automatically at the number needed, so you can use `1.` for all the list items. See the rendered version here - https://github.com/CMSgov/design-system/pull/740/files?short_path=8bb03a6#diff-8bb03a618d7ed005710786e52c92b146

line47:
Thanks @nichia I added the missing font!

bernardwang:
Crap! i made a ton of local changes to the read me and i forgot to push up... i think i may have lost those

line47:
Oh no! 😖

bernardwang:
@line47 Nevermind! I got them back with `git reflog`. I made some pretty large changes to the README

examples/child-design-system/README.md
```diff
@@ -10,94 +10,97 @@ This is the reasoning behind creating the CMS design system family which consist
...
 
-We suggest using GitHub pages `gh-pages` to publish the doc site build files.
+If you are writing a content only page then you should be creating a Markdown (`.md`) file. However if you are are writing documentation for a component, pattern or utility you should follow the [KSS format outlined for writing component documentation](https://github.com/CMSgov/design-system/blob/master/guides/WRITING-DOCUMENTATION.md).
 
-### Modifying the design
+We have also provided example documentation for an overriden `Button` component, a new `Card` component, and a new `border-style` utility for reference.
+
+### Publishing your child design system doc site
 
-- (Color variables)[https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/styles/settings/variables/_color.scss]
```
bernardwang: I felt like this information was already covered on the theming page, and didnt necessarily flow with the rest of the content on the page. However, I think it would make sense to maybe make a new section about customizing the DS that links the theming info, as well as how to customize the components

examples/child-design-system/src/styles/base/_override.typography.scss
```diff
@@ -6,5 +6,5 @@
 .ds-h4,
 .ds-h5,
 .ds-h6 {
```
bernardwang, line 8: I swapped out oswald for lato because it was more useable and realistic as a body font. I also added these typography overrides so we actually used both new fonts

line47: Oh ha! I was going to use Oswald for the headings, thanks for adding those!

examples/child-design-system/README.md
line47: Gotcha, yeah that makes sense. I def don't want us to have multiple versions of documentation.

line47:
This looks good to me! I think it would still be good to add a custom image example

PR #741, opened 2020-07-14
[NO_TICKET] Misc fixes changes

bernardwang:
### Summary
- Add `skipBuild` option for `cmsds build-docs`
- Fix @src alias bug in child ds
- Add gh-pages for publishing
- Upgrade npm version

### How to test
- Use new npm version by running `nvm use` and reinstalling, rerunning all scripts


packages/design-system-scripts/gulp/docs/webpack/createReactExampleWebpackConfig.js
```diff
@@ -37,7 +37,7 @@ module.exports = (sourceDir, reactExampleEntry, typescript) => {
     resolve: {
       modules: ['node_modules'],
       alias: {
-        '@src': path.resolve(sourceDir),
+        '@src': path.resolve(sourceDir, 'src'),
```
pwolfert, line 40: What is `@src` used for?

bernardwang: Now that the example files are moved to a separate directory, the react component imports in those example files will also have to be updated. The core uses `lerna` so it can directly import components from `@cmsgov/design-system`, but the child design systems don't, so they would have to update the imports to use really long relative paths. The `alias` is just a [helpful shorthand](https://github.cms.gov/MedicareGov/ds-site-package/pull/74/commits/0ab7c89e308bce88199a4a33194d5a3200723669) to import from the `sourceDirectory`

Originally I had imports just resolve from the `sourceDirectory` instead of using an `alias`, but I thought it might be better to be explicit. I also need to document stuff like this pretty well for the child DS haha

pwolfert: And 👍  on documenting it.

PR #743, opened 2020-07-15
[WNMGDS-545] Fix rootPath option

bernardwang:
### Summary
The `rootPath` option is needed in order to update HTML build files to have the correct paths when the doc site is published. For example, if the doc site is published at `domain.com/design/`, the `rootPath` would be set to `design`, and the `index.html` will link to `href="/design/index.css` instead of `href="/index.css"`.

The way `rootPath` used to work was the HTML paths would be updated **and** a [duplicate set of build files](https://github.cms.gov/CMS-WDS/ds-healthcare-gov/tree/master/docs) would be generated inside directories that match the root path. The duplicate build files are used to serve the site locally (i.e. `localhost:3000/pages/CMS-WDS/ds-healthcare-gov`) while the build files at the root (i.e. files under `docs/`) are the files that are actually used on the Github Pages site.

In this PR, when `rootPath` is set in the cmsds config file **and** you are building for production, the resulting build files will use the `rootPath` in their paths. The `rootPath` is ignored in local development so that the files served by `browsersync` will work correctly. You can also ignore `rootPath` when running `yarn build` by providing the `ignoreRootPath` option, this is used when you want to build in production but also want to serve the site locally, like when running e2e tests.

### How to test
- Run `yarn start` and quickly ensure everything is normal
- Go to `cli.js` and comment out line 89 under `yarn start`, and re run the command
- You then need to move all the files inside `docs/dist` to `docs/dist/design-system`, and then go to `localhost:3000/design-system/`. This simulates publishing to a domain with `design-system` as the `rootPath`.
- Test the entire site, especially all links to other parts of the site
- Undo all those changes and run `yarn test:e2e`


packages/design-system-scripts/gulp/common/getDocsDistPath.js
```diff
@@ -1,9 +0,0 @@
-const path = require('path');
-
-/**
- * The documentation can be setup to be published in a subdirectory. This is
- * a simple helper method for outputting the proper build path.
- */
-module.exports = (docsPath, rootPath = '', basename = '') => {
```
bernardwang, line 7: Because going forward we are just always building to `dist`, this isn't necessary any more

packages/design-system-docs/src/scripts/components/Nav.jsx
```diff
@@ -30,11 +31,29 @@ function isParentOfSelectedChild(items, selectedId) {
   return false;
 }
 
+function updateItemsWithRootPath(items) {
```
bernardwang, line 34: The reason why we didnt have to do this in the old scripts is because the `rootPath` used to be accounted for earlier on, it was taken into account in earlier on, [like here](https://github.com/CMSgov/design-system/pull/743/files#diff-2b20f8962ae43cec333c2999d49f8e02L49). This PR arguably makes the doc site react components more complex, because you need to handle `rootPath`, but IMO it also makes the doc generation scripts simpler as well. Instead of `rootPath` affecting the location of build files and some other internal data like `referenceURI`, we only use `rootPath` when generating markup, whether that's in the doc site react components, or in `generateHtmlExample.js` for example

bernardwang: Also technically the doc site react components had to handle the `rootPath` with images as well anyway like [here](https://github.com/CMSgov/design-system/pull/743/files#diff-ffdd7c2f4e66cdd30639d9e5d788ff20L52), so 🤷

packages/design-system-docs/src/scripts/components/UsaBanner.jsx
```diff
@@ -12,7 +13,7 @@ const UsaBanner = function (props) {
         <div className="ds-u-display--flex ds-u-flex-direction--row ds-u-align-items--start ds-u-sm-align-items--center">
           <img
             className="c-usa-banner__header-flag"
-            src="/images/us_flag_small.png"
+            src={path.join('/', process.env.rootPath, "/images/us_flag_small.png")}
```
bernardwang, line 16: I might make a small helper for this sometime

bernardwang:
This method isnt *great* in a lot of ways. It's weird for the build files to have different paths when running in development vs in production. On the other side, its also weird to put the build files inside extraneous folders in the `dist` folder (i.e. `docs/dist/pages/CMS-WDS/ds-healthcare-gov`), when you really only need to publish the files at `docs/dist` to github pages. 

To me the main advantage of this implementation is that it's slightly simpler, and it ensures the build files in `dist` don't include any extraneous folders or files

packages/design-system-scripts/gulp/docs/generatePages/generateDocPage.js
```diff
@@ -31,14 +31,14 @@ function generateDocPage(routes, page, docsPath, options) {
     return ReactDOMServer.renderToString(React.createElement(Docs, { page, routes: [] }, null));
   };
 
-  if (options.rootPath) options.rootPath = `${options.rootPath}/`;
+  const rootPath = (options.rootPath) ? `${options.rootPath}/` : null;
```
pwolfert, line 34: If we make this an empty string instead of null, do we need all those `rootPath || ''` later on?

packages/design-system-scripts/gulp/docs/generatePages/generateDocPage.js
```diff
@@ -31,14 +31,14 @@ function generateDocPage(routes, page, docsPath, options) {
...
 
-  const head = `${seo(page, options.rootPath)}
+  const head = `${seo(page, rootPath)}
```
pwolfert, line 36: It looks like the `seo` function no longer uses the `rootPath` parameter.

packages/design-system-scripts/gulp/docs/generatePages/generateHtmlExample.js
```diff
@@ -33,8 +33,8 @@ function processMarkup(markup, modifier) {
  * @param {String} rootPath - Root docs site path
  * @return {Promise}
  */
-function generateHtmlExample(page, modifier, docsPath, { rootPath }) {
-  if (rootPath) rootPath = `${rootPath}/`;
+function generateHtmlExample(page, modifier, docsPath, options) {
+  const rootPath = (options.rootPath) ? `${options.rootPath}/` : null;
```
pwolfert, line 37: Same thing here as in generateDocsPage: It doesn't look like `rootPath` being null is used anywhere else in this function but to allow it to fall through to `|| ''`; could we just use an empty string here instead of null and get rid of the `|| ''` below?

packages/design-system-scripts/gulp/docs/generatePages/generatePages.js
```diff
@@ -179,7 +178,7 @@ async function generateExamplePages(pageSection, docsPath, sourceDir, options, c
 module.exports = async function generatePages(sourceDir, docsDir, options, changedPath) {
   logTask('📝 ', 'Generating documentation pages');
 
-  const docsPath = getDocsDistPath(docsDir, options.rootPath);
+  const docsPath = path.join(docsDir, 'dist');
```
pwolfert, line 181: 👨‍🍳 👌

packages/design-system-scripts/jest/e2e.global-setup.js
```diff
@@ -8,7 +8,8 @@ const APP_PORT = 3001;
 
 function buildApp() {
   log(chalk.green('\nBuilding docs site in production mode...\n'));
-  childProcess.execSync('yarn build --skipLatest', {
+  // Build files in production while ignoring rootPath
+  childProcess.execSync('yarn build --skipLatest --ignoreRootPath', {
```
pwolfert, line 12: 👍

pwolfert:
Most things are working, but I noticed these internal links don't have the root path: https://github.com/CMSgov/design-system/blob/WNMGDS-545/fix-root-path/packages/design-system-docs/src/pages/components/Autocomplete/_Autocomplete.docs.scss#L31

bernardwang:
@pwolfert I addressed your feedback, and I also fixed a lot of internal links (all ones located in `*.docs.scss` files)

packages/design-system-scripts/gulp/common/replaceTemplateTags.js
```diff
@@ -2,28 +2,31 @@
  * Replace template tags with string values
  */
 function replaceTemplateTags(str, options) {
+  // KSS processes the markdown in documentation content before we can replace the tags
+  // So unfortunately we need to match for the encoded version of the template tags
+  // When KSS is removed we can update the regex here
   if (options.rootPath) {
-    str = str.replace(/{{root}}/g, `/${options.rootPath}`);
+    str = str.replace(/{{root}}|%7B%7Broot%7D%7D/g, `/${options.rootPath}`);
```
bernardwang, line 9: This isnt great and mainly has to do with how KSS works. Eventually we can get rid of this, and maybe even one day move to an implementation where we dont need to use `{{root}}` in the docs at all, and we can handle it for them.

pwolfert: Would it make sense to pull these regex literal out into a constants? Or maybe we can put the branching in ternaries so the messy encoded strings aren't duplicated, like:
```js
return str
  .replace(/{{root}}|%7B%7Broot%7D%7D/g, options.rootPath ? `/${options.rootPath}` : '')
  .replace(/{{npm}}|%7B%7Bnpm%7D%7D/g, options.npmPackage ? `/${options.npmPackage}` : '')
  .replace(/{{github}}|%7B%7Bgithub%7D%7D/g, options.githubUrl ? `/${options.githubUrl}` : '')
  .replace(/{{name}}|%7B%7Bname%7D%7D/g, options.name ? `/${options.name}` : '')
```
or
```js
const tagRegex = tag => new RegExp(`${tag}|${encodeURI(tag)}`, 'g');
return str
  .replace(tagRegex('{{root}}'), options.rootPath ? `/${options.rootPath}` : '')
  .replace(tagRegex('{{npm}}'), options.npmPackage ? `/${options.npmPackage}` : '')
  .replace(tagRegex('{{github}}'), options.githubUrl ? `/${options.githubUrl}` : '')
  .replace(tagRegex('{{name}}'), options.name ? `/${options.name}` : '')
```
Or even make a singular `replaceTemplateTag` function that takes the string, the tag string, and the option

bernardwang: I opted for the second option, much cleaner! I didn't make a singular `replaceTemplateTag` because  `rootPath` replaces slightly differently than that rest, it prepends a `/`. I could make the default `rootPath` `/` instead of empty string and make this nicer, but I didn't want to have to update that everywhere else for this PR.

PR #750, opened 2020-07-23
[NO_TICKET] Misc beta changes

bernardwang:
### Summary
- Fix error with lint script when prettier, eslint, or stylelint isnt installed when using any of the `ignore` options
- Remove `lead` styling applied to the first `p` elements on the doc sites to simplify styling for child doc sites. 
- Update button inverse example to have 1 inverse background.
- Move fonts out of settings to ensure its only included in the build css once.


packages/design-system-docs/src/pages/components/Button/button.example.html
```diff
@@ -130,3 +83,54 @@ <h6 class="preview__label">Size variant buttons</h6>
 <button class="ds-c-button ds-c-button--big">
   Big
 </button>
+
+<h6 class="preview__label">Inverse buttons</h6>
```
bernardwang, line 87: @line47 In this PR I combined the 2 inverse examples, does that sound ok to you?

packages/design-system-docs/src/styles/components/_details.scss
```diff
@@ -31,10 +31,6 @@
     @extend .ds-text;
   }
 
-  > p:not([class]):first-child {
```
bernardwang, line 34: Screen Shot 2020-07-28 at 11 53 08 AM
This is what it looks like now. Previously the first `p` element would be larger because it would be styled as a `lead`

PR #753, opened 2020-07-27
[NO_TICKET] Update html examples

nichia:
### Summary
Update the html examples (article, form and responsive) to reference css file with the cmsds v2 folder structure.

### DIFF
--- a/packages/design-system/src/styles/settings/variables/_build.scss
+++ b/packages/design-system/src/styles/settings/variables/_build.scss
@@ -1,6 +1,6 @@
 // Paths
-$font-path: './fonts' !default;
-$image-path: './images' !default;
+$font-path: '../fonts' !default;
+$image-path: '../images' !default;
 
 // Override these variables to change what CSS rulesets get outputted when
 // the Sass is transpiled to CSS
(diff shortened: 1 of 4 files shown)


packages/design-system/src/styles/settings/variables/_build.scss
```diff
@@ -1,6 +1,6 @@
 // Paths
-$font-path: './fonts' !default;
-$image-path: './images' !default;
+$font-path: '../fonts' !default;
+$image-path: '../images' !default;
```
line47, line 3: Can you talk about this change here? I reverted back to the single `./` and didn't see a difference. What is this change fixing?

nichia: The `../` fixes those error msgs on Website when viewing the dev tool's console 
Screen Shot 2020-07-27 at 10 15 18 AM

nichia: as the `src/font` & `src/image` folders are on the same level as `src/styles`

nichia: I'll check again, looks like maybe it did not fix that.

nichia: Checked - yes, it fixed the error msgs (I forgot to yarn build again after changing the variables).

nichia:
Transferred to pr WNMGDS-495 Update CMSDS examples.

PR #755, opened 2020-07-28
Add product/application info to migration page

line47:
Added some information on how to update an application to use the new npm dependencies


packages/design-system-docs/src/pages/startup/migrating-v2.md
```diff
@@ -7,6 +7,8 @@ The CMS Design System v2 release introduces several breaking changes, and this m
 
 ## NPM packages
 
+### Child design systems
```
bernardwang, line 10: Why is this header added here? The info below is general and not specific to child design systems

packages/design-system-docs/src/pages/startup/migrating-v2.md
```diff
@@ -29,6 +31,34 @@ New usage:
   }
 ```
 
+### Applications using a child design system
```
bernardwang, line 34: Could we mention the old name too? Some thing like this:
`### Child design system packages (formerly site packages)`

packages/design-system-docs/src/pages/startup/migrating-v2.md
```diff
@@ -29,6 +31,34 @@ New usage:
...
+
+The `core`, `support`, `layout` npm packages are no longer needed for applications using a child design system. The only dependency needed is the child design system package in your `package.json`.
```
bernardwang, line 36: It would also be nice to link to the new child ds page: `{{root}}/startup/child-design-systems/`

packages/design-system-docs/src/pages/startup/migrating-v2.md
line47: I was thinking that it is only needed for child design systems. But i suppose we could have people only using the core, I'll remove this header

packages/design-system-docs/src/pages/startup/migrating-v2.md
```diff
@@ -29,6 +29,34 @@ New usage:
   }
 ```
 
+### Child design system packages (formerly site packages)
+
+The `core`, `support`, `layout` npm packages are no longer needed for applications using a [child design system]({{root}}/startup/child-design-systems/). The only dependency needed is the child design system package in your `package.json`.
```
bernardwang, line 34: Usage of child design system NPM packages have also been consolidated and simplified. Applications that use child design system packages only need one design system dependency going forward. The child design system packages will contain core styles, components, and assets in addition to child design system specific customizations.

PR #761, opened 2020-07-30
Test babel register

bernardwang:
### Summary
CBJ isnt able to finish generating doc site because of the babel/register not working. This PR updates the regex to include CBJ's directory

CBJ error
Screen Shot 2020-07-30 at 11 18 02 AM
Notice the directory is `s3/packages/design-system-docs` rather than `design-system/packages/`

### How to test
- Ensure doc site generates in core and child ds


line47:
both the doc site and child-design-system example work for me 👍

PR #762, opened 2020-07-31
[WNMGDS-495] - Update CMSDS examples

nichia:
### Summary
- Update the HTML/CSS examples article, form and responsive to reference css file with the cmsds v2 folder structure
- Update React App example
- Update Webpack example
- Add create-react-app example


examples/html-css/README.md
```diff
@@ -0,0 +1,9 @@
+## Example: Pure HTMS/CSS
+
+This shows the usage of CMS design system components with a pure HTML/CSS set up. Simply include the CSS file in the HTML document.
+
+### Usage
+
+```html
+<link rel="stylesheet" href="packages/design-system/dist/css/index.css" />
```
line47, line 8: For the form HTML/CSS example, the images are not showing for the error message or the checkboxes or radios

line47:
I'd like to see the HTML/CSS examples combined into a single example instead of three separate index files.

examples/html-css/README.md
line47: This is also happening with the alert on the article example

nichia: Have you `yarn install` & `yarn build` the branch?

line47: That did the trick - We might want to mention the need to run `yarn install` and `yarn build` to get things to work

examples/webpack-demo/README.md
```diff
@@ -35,5 +35,3 @@ This project uses `yarn` to manage Node packages. If you're already set up', ski
 ### Go forth and edit!
 
 You can add `html` files inside the `pages` directory to build bigger sites, add more CSS or SCSS files, images, fonts, what have you. This is the very basic scaffolding one might need to build bigger. If you get stuck or need help, feel free to file a Github issue: https://github.com/CMSgov/design-system/issues.
-
```
line47, line 38: I think in general this README should be simplified to be like our others with the basic info like `create-react-app`.

line47:
Nice work! Everything is working as expected. 

I left a few comments about 
- making the HTML/CSS example a single index file
- simplifying the readme for the webpack demo to make it less verbose


I'm also curious about the need for these examples and would like to try and find out if they are being used. 

I think the HTML/CSS examples could be removed because we have this page 
https://design.cms.gov/startup/sass-and-css/

I'm also wondering about the webpack example and how much usage it gets.

line47:
Looks good! I'd still like to see the 3 HTML examples on one HTML page for simplicity. The sections could be slimmed down and you could break the page into sections with H1 tags

nichia:
@line47 Check this change - moved html-css examples to a single index file. Does this work?

examples/create-react-app/.env
```diff
@@ -0,0 +1,3 @@
+SASS_PATH=node_modules:src
+SKIP_PREFLIGHT_CHECK=true
+EXTEND_ESLINT=true
```
bernardwang, line 3: Looks like this should have a new line here

examples/create-react-app/.env
```diff
@@ -0,0 +1,3 @@
+SASS_PATH=node_modules:src
+SKIP_PREFLIGHT_CHECK=true
```
bernardwang, line 2: just curious, whats this for?

examples/create-react-app/README.md
```diff
@@ -0,0 +1,8 @@
+## Example: create-react-app
+
+This shows the usage of CMS design system components in a [create-react-app](https://create-react-app.dev/docs/getting-started/) project.
+
+## Getting started
+
+1. Install packages: `yarn install`
+1. Run the application: `yarn start`
```
bernardwang, line 8: We should probably have some more information here. Specifically a step for installing the CMSDS as a dependancy, using it in our CSS, setting up font/image paths, etc

examples/html-css/index.html
```diff
@@ -3,10 +3,15 @@
   <head>
     <meta charset="utf-8" />
     <title>Design System</title>
-    <link rel="stylesheet" href="../../packages/core/dist/index.css" />
+    <link rel="stylesheet" href="../../packages/design-system/dist/css/index.css" />
```
bernardwang, line 6: What happened to using unpkg?
https://unpkg.com/

examples/html-css/index.html
```diff
@@ -3,10 +3,15 @@
...
     <link rel="stylesheet" href="style.css" />
   </head>
   <body class="ds-base">
+    <header class="ds-u-padding--3 ds-u-sm-padding--3 ds-u-display--block ds-u-fill--gray-light">
```
bernardwang, line 10: I thought we decided on three different HTML examples inside `html-css`? So instead of defining a new HTML file just literally copying the 3 html examples into a folder called `html-css`

examples/create-react-app/.env
nichia: To fix the dependency issue because the stand along create-react-app is nested within the design-system folder - refer to link for more info https://www.reddit.com/r/reactjs/comments/abot9o/how_to_fix/ (Also noted that pet-shared-compoents has this variable set as well)

examples/html-css/index.html
nichia: Yes, we did, and yesterday I tried merging the examples and it looks pretty good. Check it out and see what you think. I have then updated the docs that references this example link.

examples/html-css/index.html
nichia: Usage of https://unpkg.com/ didn't work, it cannot find the font & images folder. I'm thinking that could be because current npm release has the variables for font/images at `./` folder. After the release of this PR which changes the variables to look for fonts/images at `../` folder, I'll re-test the use of https://unpkg.com/

packages/design-system/src/styles/settings/variables/_build.scss
```diff
@@ -1,6 +1,6 @@
 // Paths
-$font-path: './fonts' !default;
-$image-path: './images' !default;
+$font-path: '../fonts' !default;
```
bernardwang, line 2: Why is this change being made?

nichia: Notice how if you view github pages, the flag image is not found. This fixes the issue of not locating the fonts & images when we view the `.html` file

nichia: By setting the default to reference `../` folder, accesses to the fonts & images files works.

examples/html-css/README.md
```diff
@@ -0,0 +1,17 @@
+## Example: Pure HTMS/CSS
```
bernardwang, line 1: HTMS/CSS typo

examples/html-css/README.md
```diff
@@ -0,0 +1,17 @@
...
+
+This shows the usage of CMS design system components on a pure HTML/CSS site. Simply include the CSS file in the HTML document.
```
bernardwang, line 3: Maybe we can include the description of the 3 old examples here, form, article, responsive

examples/html-css/index.html
nichia: Have updated to use https://unpkg.com/

examples/html-css/index.html
bernardwang: This styling seems super heavy handed
Screen Shot 2020-08-06 at 12 15 13 PM

Can we remove the gray fill and label each example like so?
Screen Shot 2020-08-06 at 12 15 41 PM

nichia: That looks much better!

examples/create-react-app/.env
bernardwang: could we include this info in a comment?

examples/create-react-app/.env
bernardwang, line 3: Same with this, we can leave comments here explaining this allows you to override create react app default linting

examples/create-react-app/.env
```diff
@@ -0,0 +1,3 @@
+SASS_PATH=node_modules:src
```
bernardwang, line 1: can we have a comment here explaining how this is used in the font/image paths?

examples/create-react-app/README.md
```diff
@@ -0,0 +1,10 @@
+## Example: create-react-app
+
+This shows the usage of CMS design system components in a [create-react-app](https://create-react-app.dev/docs/getting-started/) project.
+
+Refer to [install with NPM]({{root}}/startup/installation/) for instruction on installing the design system as a dependency package. This example includes a basic `.env` and `index.scss` to reference the dist folder.
```
bernardwang, line 5: Could we expand on this? What does "reference the dist folder" mean? Maybe we can say the `.env` is for a basic create-react-app configuration with the design system.

examples/html-css/README.md
bernardwang: I dont think this needs to be mentioned because the path in the HTML/CSS use case will be different for every app. We have it setup here directing to our packages, but thats not relevant to most users, nor would users be probably testing it via our github repo

bernardwang: if anything, i would prefer this path to be replaced with something more realistic like:
```
href="vendor/cmsgov-design-system-2.0.0/index.css
```
and then maybe add a note that the path should go to whereever the CMSDS css file is stored in their project

examples/html-css/README.md
```diff
@@ -0,0 +1,22 @@
...
+
+```html
+<link rel="stylesheet" href="packages/design-system/dist/css/index.css" />
+```
+
+## Getting started
+
+Install and build the core design system to ensure html-css example can reference the minified CSS.
+
+1. Install packages: `npm install` or `yarn install`
```
bernardwang, line 20: Same info here, most HTML/CSS apps wont be using the CMSDS via npm, they will be downloading and manually using the CSS file

examples/html-css/index.html
```diff
@@ -3,10 +3,18 @@
   <head>
     <meta charset="utf-8" />
     <title>Design System</title>
-    <link rel="stylesheet" href="../../packages/core/dist/index.css" />
+    <link
+      rel="stylesheet"
+      href="https://unpkg.com/@cmsgov/design-system@2.0.0/dist/css/index.css"
```
bernardwang, line 8: Can we leave a comment here explaining that in a real project this will lead to your local CSS file, while in this case we are using a CDN as an example

examples/webpack-demo/src/style.scss
```diff
@@ -1,6 +1,8 @@
...
+$asset-path: '/examples/webpack-demo/dist/';
+
+// Paths to images and fonts relative from the SCSS entry point
 $font-path: '#{$asset-path}fonts';
 $image-path: '#{$asset-path}images';
 
-@import '~@cmsgov/design-system-layout/src/index';
-@import '~@cmsgov/design-system-core/src/index';
+// Core CMSDS styles
+@import '~@cmsgov/design-system/dist/scss/index';
```
bernardwang, line 8: nice!

package.json
```diff
@@ -56,5 +56,8 @@
   "browserslist": [
     "> 2%",
     "Last 2 versions"
-  ]
+  ],
+  "dependencies": {
+    "node-sass": "^4.14.1"
```
bernardwang, line 61: What is this from? I dont think we want to be adding dependencies

bernardwang: I believe its already covered in the cmsds scripts package

examples/create-react-app/src/components/App.js
```diff
@@ -0,0 +1,100 @@
...
+        <center>
+          <Badge variation="info" size="big">
+            <img className="c-usa-banner__header-flag" src={usflag} alt="U.S. flag" />
+            &nbsp;CMS Design system
+          </Badge>
+        </center>
+      </header>
+
+      <Tabs>
+        <TabPanel id="summary" tab="Summary">
```
bernardwang, line 23: This is really great Ni! I like how you put in content here. I wonder if we can simplify this though, so the content doesnt detract from the example. Mayb we can just remove the other Tabs besides summary?

nichia: I thought the Tabs component would be fun to use. What do you think if I just keep it to 2 tabs - summary and uswds tabs only?

package.json
nichia: Good pickup, it shouldn't be there. It's just needed for the stand along create-react-app folder.

examples/create-react-app/src/components/App.js
bernardwang: Ok that’s fine, i just think we should be intentional with how we create examples. Generally i’m not a huge fan of arbitrary or verbose example content, as it doesn’t really provide the user with anything. in this example, usage of react component is the most important thing, and that doesn’t have to be achieved with tabs

bernardwang: the child design system example is a good example of this. originally we had random style changes, but as soon as we tried to the example realistic, we ran into many problems, and we found the need to provide a better example of organization and overrides

examples/create-react-app/README.md
```diff
@@ -2,7 +2,7 @@
 
 This shows the usage of CMS design system components in a [create-react-app](https://create-react-app.dev/docs/getting-started/) project.
 
-Refer to [install with NPM](https://design.cms.gov/startup/installation/) for instructions on how to install the design system as a dependency package. The `.env` is for a basic create-react-app confirguration with the design system. The [`/src/styles/index.scss`](https://design.cms.gov/startup//sass-and-css/) includes the design system's styles and paths for the image and font directories.
+Refer to [install with NPM](https://design.cms.gov/startup/installation/) for instructions on how to install the design system as a dependency package. The `.env` is for a basic `create-react-app` confirguration with the design system. The [`/src/styles/index.scss`](https://design.cms.gov/startup/sass-and-css/) includes the design system's styles and paths for the image and font directories.
```
bernardwang, line 5: Configuration is spelled wrong, the link to the sass and css page is incorrect. i would search for {{root}} to see an example of how the url should look. also, since you are specifically linking to the section for fonts and images, please use # to directly link to that section of the page

nichia: Thanks, fixed typo and `#` link to fonts/images. As for {{root}}, i didn't use it because the `.md` is viewed on github and not on a web page, hence i found {{root}} is not working, so had to refer to the full url path.

bernardwang: Ah right, {root}} is right

examples/create-react-app/.env
```diff
@@ -0,0 +1,12 @@
+# To use imports relative a path you specify, and from node_modules without
+# adding the `~` prefix
```
bernardwang, line 2:
```suggestion
# Add the `SASS_PATH` variable to allow Create React App to import from `node_modules` without adding the `~` prefix
```

examples/create-react-app/.env
```diff
@@ -0,0 +1,12 @@
...
+# Refer to https://create-react-app.dev/docs/adding-a-sass-stylesheet for more info.
+SASS_PATH=node_modules:src
+
+# Fixes the issue of the stand along creat-react-app referencing
+# the parent folder's node_modules dependencies
```
bernardwang, line 7:
```suggestion
# Fixes the issue of the stand along creat-react-app referencing
# the parent folder's node_modules dependencies
```

nichia: @bernardwang - this looks the same?

PR #765, opened 2020-08-04
Release 2.0.0

bernardwang:
### Summary
- Bumped `design-system`, `design-system-docs`, `design-system-scripts` to 2.0.0
- Bumped `eslint-config-design-system` and `stylelint-config-design-system` to 2.0.0
- Updated example package.jsons
- Updated v2 banner copy

### Fixed
- Fixed an e2e test regarding copy

### DIFF
--- a/design-assets/cms-design-system-ui-kit.xml
+++ b/design-assets/cms-design-system-ui-kit.xml
@@ -7,7 +7,7 @@
       <url></url>
       <title></title>
     </image>
-    <generator>Sketch</generator>
+    <generator>Sketch </generator>
     <item>
       <title>CMS Design System UI kit - v2.0.0</title>
       <pubDate>Tue, 4 Aug 2020 12:00:00</pubDate>
--- a/design-assets/cms-design-system-ui-kit.xml
+++ b/design-assets/cms-design-system-ui-kit.xml
@@ -9,9 +9,9 @@
     </image>
     <generator>Sketch</generator>
     <item>
-      <title>CMS Design System UI kit - v0.7</title>
-      <pubDate>Mon, 18 May 2020 15:11:00</pubDate>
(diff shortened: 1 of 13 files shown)


packages/design-system-docs/package.json
```diff
@@ -9,7 +9,7 @@
   "repository": "CMSgov/design-system",
   "license": "SEE LICENSE IN LICENSE.md",
   "dependencies": {
-    "@cmsgov/design-system": "2.0.0-beta.5",
+    "@cmsgov/design-system": "^2.0.0",
```
line47, line 12: Wondering why the `^` was used here and not with other dependencies?

design-assets/cms-design-system-ui-kit.xml
```diff
@@ -9,9 +9,9 @@
     </image>
     <generator>Sketch</generator>
     <item>
-      <title>CMS Design System UI kit - v0.7</title>
-      <pubDate>Mon, 18 May 2020 15:11:00</pubDate>
-      <enclosure url="https://github.com/CMSgov/design-system/raw/master/design-assets/CMS-Design-System-UI-kit.sketch?raw=true" type="application/octet-stream" sparkle:version="0.7"/>
+      <title>CMS Design System UI kit - v2.0.0</title>
```
bernardwang, line 12: @line47 does this look right for updating sketch? also is the library updated enough for v2?

bernardwang: If there's updates to make we can hold off on updating sketch. I.e. what should we do with the focus styles in sketch?

line47: That looks right! Thanks, @bernardwang I think we are good to update Sketch

packages/design-system-docs/package.json
bernardwang: I think we don't want an exact version, this way when we update `design-system` we dont need to also update this dep. Not 100% sure though, we can adjust in the future if needed

PR #766, opened 2020-08-04
[NO_TICKET] Update default font image path

bernardwang:
### Summary
- Update the default font and image path to `../fonts` rather than `./fonts`

### How to test
- Ensure the core builds
- Ensure both child design systems build


line47:
✅  The core design system builds 
✅  The medicare child design system builds 
✅  The healthcare child design system builds

PR #768, opened 2020-08-06
[WNMGDS-576] Update styles for transparent inverse buttons

### LINKED ISSUE #767: Transparent button has background color when inversed
**Describe the bug**

The [rule that makes the transparent button transparent](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/styles/components/_Button.scss#L99) actually gets overridden by [this rule which has more specificity](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/styles/components/_Button.scss#L207-L215). From the notes it looks like the team is already aware of it, so I guess I don't know if it's intentional or a bug.

**To Reproduce**
1. View the inverse button example
2. Change the background color of the panel from the default dark blue inverse color to a slightly different color
3. See that the background color of the transparent button is different

**Expected behavior**
I expect the transparent button's background to be transparent. I know this is outside the intended use case, but intuitively I expect it to be transparent.

**Screenshots**
Screen Shot 2020-08-05 at 2 01 07 PM

**Additional context**
We're using it for a dev tool that is a black utility belt that exists on lower environments to help testers. After upgrading CMSDS, the transparent buttons are now blue.
Screen Shot 2020-08-05 at 2 16 27 PM

line47:
This PR adds transparent backgrounds to inverse transparent buttons

- The transparent inverse buttons (aka inverse links) had a background color on them of `#112e51` and now they have a transparent background
- Transparent inverse buttons that were disabled had a background color of `#081627` and now they also have a transparent background 

### Screenshot before update
the transparent inverse buttons should look like links and they don't 
Screen Shot 2020-08-06 at 9 52 50 AM

### Screenshot after update
The transparent inverse buttons now look like links 
Screen Shot 2020-08-06 at 9 52 42 AM

Fixes #767 

### How to test

1. View the inverse button example
1. Change the background color of the panel from the default dark blue inverse color to a slightly different color
1. Make sure that the transparent inverse buttons look good and don't show a background  or border color


packages/design-system/src/styles/components/_Button.scss
```diff
@@ -195,14 +195,17 @@
...
+    border-color: transparent;
   }
 }
 
 .ds-c-button--inverse.ds-c-button--transparent,
 .ds-c-button--inverse.ds-c-button--transparent:visited {
   @extend %transparent-inverse-button;
 }
+.ds-c-button--inverse.ds-c-button--transparent:disabled {
+  background-color: transparent;
```
bernardwang, line 207: This is the one tricky part, transparent buttons generally should be styled like links, but links dont have a disabled styling. I think disabled transparent buttons should look like other buttons, aka, it should have a background color.

line47: I'm not sure I follow - Are you saying you think disabled transparent buttons (links) should have a background color?

bernardwang: Yes, I think the disabled transparent buttons should have a background color, I think it should look the same as the other disabled buttons.

Reasoning for this is that the transparent buttons are basically "buttons styled like links". The current disabled transparent button styling doesn't make it clear that the button is disabled, and there's no such thing as a disabled link.

pwolfert: That makes sense to me. I have no problem with this given our use case. The disabled button would be that grey color, which wouldn't clash with our background.

line47: Is this what you'd expect? for the disabled inverse transparent button (AKA disabled link) 
Screen Shot 2020-08-11 at 2 13 00 PM

pwolfert: Oh, I didn't realize that it was a dark blue. I don't really care. I don't have any disabled buttons in my UtilityBelt, and if I did, I'd care more about them being clearly disabled than about the color being neutral.

bernardwang: @line47 that looks almost done, I think once the underline is removed its good to go.

bernardwang:
@line47 This might be a bit out of scope of this PR, but could we also make sure the non inverse disabled style is consistent too?
Screen Shot 2020-08-13 at 4 44 43 PM

bernardwang:
Regarding the fact that we have a different disabled style on dark background, I think in the future we should simplify it so there are only 2 disabled styles, outline and solid. USWDS has the same solid disabled style on dark backgrounds as well:
Screen Shot 2020-08-13 at 4 46 36 PM

PR #774, opened 2020-08-11
[WNMGDS-582] Improve css for child ds

bernardwang:
### Summary
While updating the mgov child design system, we found several core styles that made it very difficult to override styles.

- Adds `!default` to more variables, including `$color-background-inverse`
- Removes the `link` import in `Alert` and `UsaBanner` preventing those link styles from being applied multiple times and potentially overriding child design system styles

### How to test
- Ensure no visual changes to core
- Test this local branch on the mgov child design system, ensure that the banner is now teal and that links are working correctly.


packages/design-system-docs/src/styles/components/_UsaBanner.scss
```diff
@@ -35,7 +35,9 @@
 }
 
 .c-usa-banner__button {
-  @extend %link-darker;
+  /* stylelint-disable scss/at-extend-no-missing-placeholder */
+  @extend .ds-c-link;
```
bernardwang, line 39: This is the one visual change, the USA button link is now regular link blue, matching USWDS

line47: Nice fix here!

line47:
The core design system doc site looks good!

The Mgov child design system doc site looks good!

I did find one issue in the footer that we should fix and that is the classes on the privacy settings link in the footer.
It has a white background on the button

Screen Shot 2020-08-12 at 10 09 11 AM



The classes applied currently are
```
ds-c-button ds-c-button--small ds-c-button--transparent-inverse ds-u-padding--0 titlelink
```

they should be updated to this to fix the issue 
```
ds-c-button ds-c-button--small ds-c-button--transparent ds-c-button--inverse ds-u-padding--0 titlelink
```

nichia:
LGTM.

PR #775, opened 2020-08-11
[WNMGDS-581] Reorganize doc site scss

bernardwang:
### Summary
Child design system doc sites need to import styles from `design-system-docs/src/index.scss` and `design-system-docs/src/example.scss`. This make it easier to do so while using the child design system styles. By separating out 2 extra scss files, we can provide scss entry files **without** imports to the core CMSDS.

### How to test
- Ensure no doc site visual changes
- Go to any of the child design systems, replace `docs/src/index.scss` and `docs/src/example.scss` with new imports to `design-system-docs/src/styles/index.scss` and `design-system-docs/src/styles/example.scss`. It should look like:
```
// Override default assets path
$font-path: './fonts';
$image-path: './images';

@import '../../src/styles/index';

@import '@cmsgov/design-system-docs/src/styles/example';
// or @import '@cmsgov/design-system-docs/src/styles/index';
```


line47:
This looks good!

I think the paths to the @cmsgov scss files need to remove the `styles` portion of the path for the child design systems to be like this:
- `@import '@cmsgov/design-system-docs/src/index';`
- `@import '@cmsgov/design-system-docs/src/example';`

When I tried to use 
`@import '@cmsgov/design-system-docs/src/styles/example';` i got an error that the file could not be found 

```
Error: File to import not found or unreadable: @cmsgov/design-system-docs/src/styles/example.
        on line 8 of docs/src/example.scss
>> @import '@cmsgov/design-system-docs/src/styles/example';
```

bernardwang:
@line47 
We specifically want to import the new scss files located under styles, we want to avoid importing the original files at the src root. When you get that error are you sure that the local version of the doc site is installed in the child ds?

line47:
🤦 I missed updating the scripts in my earlier testing. It's working as expected, thanks!

nichia:
When I replaced docs/src/index.scss and docs/src/example.scss with the content listed under `How to test`, the child ds layout is incorrect. But when i keep the child ds  index.scss & example.scss and just delete the lines referencing the deleted file, child ds looks good.

nichia:
I updated the child ds scripts incorrectly - re-tested and it works great!

PR #776, opened 2020-08-12
[NO-TICKET] Backstop updates for V2

line47:
Updating the backstop reference files for V2


bernardwang:
@line47 If im not mistaken, I think backstop isnt listed as a dependancy in our root package.json. Could we add that in?

line47:
Need to re-run reference files after table component is added

nichia:
Thanks @line47 - can you add below snapshot of table component example after line 943
`    {
      "label": "CMSDS - Components - Table - React",
      "referenceUrl": "https://design.cms.gov/example/components.table.react/",
      "url": "http://localhost:3000/example/components.table.react/",
      "removeSelectors": ["#frame"]
    },`

line47:
I've updated this PR again - ready for review @nichia and @bernardwang 👍

backstop.json
```diff
@@ -37,25 +37,19 @@
       "removeSelectors": ["#frame"]
```
bernardwang, line 37: small copy update above on line 28, can we update to “Installation” instead of install with npm to match the site

package.json
```diff
@@ -56,5 +56,8 @@
   "browserslist": [
     "> 2%",
     "Last 2 versions"
-  ]
+  ],
+  "dependencies": {
```
bernardwang, line 60: this should be a dev dependancy

PR #777, opened 2020-08-12
[WNMGDS-588] Update to use lifecycle methods

nichia:
### Summary
Table components uses React hooks and context API introduced with React v16.3 (production ready 16.8) which is not consumable by hc.gov child ds currently on older version of React. This pr updates the Table components to use React lifecycle methods to ensure support of child ds/products running on older versions of React.

### Testing
No visual change. Table components continue to work.


packages/design-system-docs/src/styles/components/_ReactPropDocs.scss
```diff
@@ -0,0 +1,65 @@
+// Responsive docs table
+// TODO: Deprecate after the release of React Table Component. This include removing from m.gov child ds.
+
+.docs_table--container {
```
bernardwang, line 4: Why is this being added back? I thought it wasnt needed anymore if we are using the React table

nichia: I put this back temporarily as child ds will fail if it's removed. This will then be removed after child ds .scss has been updated to refer to core for the styles.

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -1,8 +1,7 @@
-import React, { useEffect, useRef, useState } from 'react';
 import Alert from '../Alert/Alert';
 import PropTypes from 'prop-types';
+import React from 'react';
```
bernardwang, line 3: One thing that might be nice to include is a comment linking to this PR, this way we can revert these changes when we eventually get everyone onto react 16.8

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -17,108 +16,118 @@ function debounce(fn, ms) {
...
+  }
 
-  const isTableCaptionComponent = (child) => {
-    return child && child.type === TableCaption;
-  };
+  handleResize() {
+    const { scrollWidth, clientWidth } = this.container;
+    const scrollActive = scrollWidth > clientWidth;
+    this.setState({
+      tabIndex: scrollActive ? '0' : null,
```
bernardwang, line 68: @nichia It looks like this doesnt incorporate later feedback in the original table PR, i.e. using `isTableScrollable` instead of `tabIndex`. If the changes in this PR are just made by reverting to old commits, can you make sure to incorporate all the later updates as well?

packages/design-system-docs/src/styles/components/_ReactPropDocs.scss
bernardwang: sorry im not sure if i understand. why would the child ds fail? isnt the child ds using the react table for the react props?

packages/design-system/src/components/Table/TableDataCell.jsx
```diff
@@ -11,19 +10,8 @@ export const TableDataCell = ({
   stackedTitle,
   ...others
 }) => {
-  const tableStackableContext = useContext(TableContext);
-  if (process.env.NODE_ENV !== 'production') {
```
bernardwang, line 15: Hm its a bit sad that we lose these warnings. Is it possible to use `React.cloneElement` in the root table component in order to pass an prop to the children that replaces `tableStackableContext`?

bernardwang: Actually, I'm ok with this being handled in the other table follow up PR, as we need this PR to be merged soon, before the next release

packages/design-system/src/components/Table/Table.jsx
nichia: Ah, will add back

packages/design-system-docs/src/styles/components/_ReactPropDocs.scss
nichia: The child ds docs/src/index.scss file make references to core ds for the "ReactPropDocs". I believe we can remove "ReactPropDocs" after we release the pr for chld ds where it references core ds for "index.scss" instead of the individual reusable files.

bernardwang: It doesn’t affect the child ds to remove this file, as it will be removed in the next version. the child ds is using version 2.0.0 which has the file

PR #782, opened 2020-08-18
[WNMGDS-591] Fix link color

bernardwang:
### How to test
- Make sure the link color is primary instead of primary-dark


nichia:
The fix looks good. 

Notices two other areas that requires similar fix: Alert.scss and Dialog.scss

packages/design-system/src/styles/components/_Alert.scss
```diff
@@ -24,7 +24,9 @@
       text-decoration: none;
     }
   } @else {
-    color: $color-primary-darkest;
```
line47, line 27: Needed to add :focus here

packages/design-system/src/styles/components/_Dialog.scss
```diff
@@ -54,7 +54,9 @@ $dialog-spacer: $spacer-4;
       background-image: url('#{$image-path}/close.svg');
     }
   } @else {
-    background-image: url('#{$image-path}/close--primary-darker.svg');
```
line47, line 57: Also needed to add :focus here, thanks for the catch @nichia

bernardwang: i’m a bit confused why we even use the focus flag here, what was this originally? shouldn’t the the style here be to apply the darker close button on focus all the time

line47: - [Before the focus styles were implemented `close--primary-darker.svg` was used for focus](https://github.com/CMSgov/design-system/blame/89bf81730c3d60154045930ff7fc3679470baade/packages/core/src/components/Dialog/Dialog.scss#L50)
- [When the focus styles were first implemented `close.svg` was used for focus](https://github.com/CMSgov/design-system/blame/0f4386139d23724a999dfe60b6e34f3514c67965/packages/design-system/src/styles/components/_Dialog.scss#L53) because the image was black and fit with the rest of the black and gold focus styling.

line47: So I'm thinking we keep this `if/else` because of the color differences in the close buttons

bernardwang: gotcha thanks for clarifying!

PR #783, opened 2020-08-19
Add spans to html examples with form labels

bernardwang:
### Summary
Form label html examples dont match the react FormLabel component. This is because the react component adds an extra span. This isnt necessary for most HTML examples, but can be useful for styling (i.e. MCT uses it to style the form label as uppercase). This PR updates all the HTML examples to include the extra span so that the Mgov Child DS examples are consistent

How it looked before: 
Screen Shot 2020-08-19 at 12 10 41 PM
How we want it to look
Screen Shot 2020-08-19 at 12 10 44 PM

### How to test
- Ensure no visual changes in the core
- Ensure the Mgov child ds site has the same FormLabel styling.


packages/design-system-docs/src/pages/components/Dropdown/dropdown.example.html
```diff
@@ -10,9 +10,11 @@
...
   <label class="ds-c-label ds-u-margin-top--0" for="options-inverse">
-    Dropdown label
-    <span class="ds-c-field__hint ds-c-field__hint--inverse">
-      Helpful hint text
+    <span
+      >Dropdown label
+      <span class="ds-c-field__hint ds-c-field__hint--inverse">
+        Helpful hint text
+      </span>
     </span>
```
bernardwang, line 18:
```suggestion
    <span>Dropdown label</span>
    <span class="ds-c-field__hint ds-c-field__hint--inverse">
      Helpful hint text
    </span>
```

line47:
`<span>` tags are also inside `<label>` elements in forms in React examples but are not found in HTML examples. 

The react component for choice renders `<label><span class="">Choice 1</span></label>` while the HTML example renders `<label>Choice </label>`

packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
```diff
@@ -11,7 +11,7 @@
       checked
     />
     <label for="checkbox-1">
-      Checkbox 1
+      <span>Checkbox 1</span>
```
line47, line 14: Added span tags around the examples in this doc to be consistent with the React versions

line47:
Removed checkbox-radio-size.example.html because it was unused

nichia:
MonthPicker react example enclosed `hint` prop in <span>. This resulted in nested <span> as the react component will enclose `hint` props in a span. 
Can update to use fragment `<>` instead of `<span>` for this MonthPicker example.
`      hint={
        <>
          This is a long-form explanation of what we mean by employer coverage, and at the end of
          this sentence is a link to{' '}
          <a href="#more-info" className="ds-u-color--inherit">
            find out more about employer coverage
          </a>
          .
        </>
      }`

PR #784, opened 2020-08-19
[WNMGTAXTOOLS-73] Export getMonthNames function from MonthPicker module

pwolfert:
### Fixed
- Export `getMonthNames` function from `MonthPicker` module, which one of our products relied on

### How to test
1. Check out 4b38710 locally, which has some temporary debugging code in it
2. `yarn install && yarn start`
3. Go to the `MonthPicker` page of the docs site
4. Open up the developer console and see the debug output from that commit


pwolfert:
@bernardwang, thanks! Yeah, it used to be that [`*` was exported from that module](https://github.com/CMSgov/design-system/blob/6389b2d446ddcf700f86e68a7036ab49f8caa3f9/packages/core/src/components/index.js), so that detail was hidden.

PR #785, opened 2020-08-19
[WNMGDS-592] Add IE11 polyfill

bernardwang:
### How to test
Run site locally, run it on IE11 with sauce labs or virtualbox. I used ngrok to surface my localhost to a public url


bernardwang:
@line47 sorry i know we were supposed to pair, but i ended up just putting up the PR very quickly so we can get it in asap for the release

PR #786, opened 2020-08-20
[WNMGLEARN-289][WNMGLEARN-317] Refactor currency mask.

LouisFettet:
### Fixed
- [WNMGLEARN-289](https://jira.cms.gov/browse/WNMGLEARN-289): an issue where the currency input loses precision at large values
- [WNMGLEARN-317](https://jira.cms.gov/browse/WNMGLEARN-317): an issue where inputting multiple decimal points would mask the input as `NaN`

### Background
Previously, the currency mask was converting the user input into a number and back into a string to format it as a dollar amount.  However, we weren't getting a lot of mileage out of this transform (in fact it was causing issues for large numbers), and all of the formatting we want to do can be handled without this transform.

### Summary of changes
This PR removes a lot of the helper functions that were only being used for the currency mask, and replaces them with a single `toCurrency` function, which takes a string input (with input length at least one) and returns a string with the value formatted as a dollar amount.  I also added more tests and refactored the tests to be a bit more DRY.

**Note**: The currency mask on this branch should behave exactly as the one on `master` does (besides the bug fixes), with only one exception: previously inputting extra digits after the hundredths place would cause some rounding to occur (which didn't quite work properly, try inputting `0.999`), and now any additional digits are just trimmed.  Overall the approach here was to preserve user input as much as possible.

While I have pretty high confidence that everything here works as expected, I think there's probably a lot of readability improvements that could be made.  Suggestions are welcome!


packages/design-system/src/components/TextField/Mask.jsx
```diff
@@ -69,33 +47,38 @@ function toDigitsAndAsterisks(value) {
...
+      fractional = fractional.concat('0');
+    } else if (fractional.length > 2) {
+      // Cut all characters after the second decimal place.
+      fractional = fractional.slice(0, 2);
+    }
+    // Clear the fractional if there's no cents. Add the decimal back here.
+    fractional = fractional === '00' ? '' : `.${fractional}`;
   }
+  // Add commas for readability or return zero.
+  whole = whole === '' ? '0' : whole.replace(/\B(?=(?:\d{3})+(?!\d))/g, ',');
```
LouisFettet, line 80: This is definitely the most complicated / illegible part.  To be brief, we're "replacing" the zero-length space between groups of 3 digits with a comma.  

A variation of this regex can be found all over the place; the only modification I made was the addition of the non-capturing group (`?:`) which just makes the match return cleaner (doesn't actually affect the replacement).

Here's a demo of it in action alongside an explanation of the tokens that may be helpful: https://regex101.com/r/JPocnt/2

packages/design-system/src/components/TextField/Mask.test.jsx
```diff
@@ -141,40 +141,51 @@ describe('Mask', function () {
...
   describe('Currency', () => {
-    it('accepts already masked value', () => {
-      const data = render({ mask: 'currency' }, { value: '1,234.50' });
+    const testCurrency = (value, expected) => {
+      const data = render({ mask: 'currency' }, { value });
       const input = data.wrapper.find('input');
-
-      expect(input.prop('value')).toBe('1,234.50');
+      expect(input.prop('value')).toBe(expected);
+    };
```
LouisFettet, line 148: Ideally we would just be calling `toCurrency` directly instead of rendering the component each time, but I figured we shouldn't be exporting functions that aren't going to be used by the consumers of the DS.

bernardwang: I think its actually ok to export `toCurrency` because we only export specific functions: https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/index.js

packages/design-system/src/components/TextField/Mask.jsx
pwolfert: What if you actually include this link in the comment? I think it makes great documentation

packages/design-system/src/components/TextField/Mask.jsx
```diff
@@ -69,33 +47,38 @@ function toDigitsAndAsterisks(value) {
...
-  // evaluate as an integer if there are no digits beyond the decimal, then we
-  // can change it.
-  const hasDecimal = parts[1] !== undefined;
-  if (hasDecimal) {
-    const a = toDigits(parts[0]);
-    const b = toDigits(parts[1]);
-    return sign * parseFloat(`${a}.${b}`);
-  } else {
-    return sign * parseInt(toDigits(parts[0]));
+function toCurrency(value) {
```
pwolfert, line 54: Even with all the transform steps, you've made this function really easy to follow.

packages/design-system/src/components/TextField/Mask.jsx
bernardwang: Could we maybe include this explanation in a comment here?

packages/design-system/src/components/TextField/Mask.jsx
```diff
@@ -123,19 +106,13 @@ function isValueMaskable(value, mask) {
 export function maskValue(value = '', mask) {
   if (isValueMaskable(value, mask)) {
     if (mask === 'currency') {
-      // Format number with commas. If the number includes a decimal,
-      // ensure it includes two decimal points
-      const number = toNumber(value);
```
bernardwang, line 128: Thank you for removing these helpers, it really simplifies things!

packages/design-system/src/components/TextField/Mask.test.jsx
pwolfert: I actually like that it's testing the component as a whole. Maybe we can have both?

packages/design-system/src/components/TextField/Mask.jsx
bernardwang: agreed! this is a hugeee improvement

bernardwang:
While testing this PR I found some issues with the other non currency masked fields, I made a ticket to track them here: https://jira.cms.gov/browse/WNMGDS-597

[One of them](
https://gyazo.com/86bc2d61635fbd6e3f5a672aec0ff095) was if a non currency masked field didnt have enough numeric digits to match the `maskDeliminatedRegex`, `deliminateRegexGroups()` would just return the original value. I think this could be improved by returning a processed value that removes non numerics.

For example, currently a zip code with `1-234` will result in `1-234` when the field is masked on blur. However, `1-2345` will result in `12345`

LouisFettet:
@bernardwang @pwolfert thanks so much for the reviews. I made some follow-up changes in d0311b8, namely:
- added a couple extra comments
- moved the transform being done on the whole number portion a little higher (has no effect, just makes more sense to make these edits from left to right)
- changed the tests a bit; now we're only testing the component to make sure it runs the mask when appropriate.  Otherwise, we can skip the component testing and just test the mask.  Let me know if this is okay by y'all!

packages/design-system/src/components/TextField/Mask.jsx
```diff
@@ -69,33 +47,40 @@ function toDigitsAndAsterisks(value) {
...
+      // Pad with a zero for two decimal places.
+      fractional = fractional.concat('0');
+    } else if (fractional.length > 2) {
+      // Cut all characters after the second decimal place.
+      fractional = fractional.slice(0, 2);
+    }
+    // Clear the fractional if there's no cents. Add the decimal back here.
+    fractional = fractional === '00' ? '' : `.${fractional}`;
   }
+  return `${sign}${whole}${fractional}`;
```
whatbird, line 83: can a bit of this be done with `toLocaleString`?

PR #787, opened 2020-08-20
[WNMGDS-596] Fix for checked checkbox background variable

line47:
Checked checkboxes and radios were not using the proper variable. 
This Pr updates to the correct variable.

### How to test
- run the doc site locally and visit the choice page and ensure the proper color is being used `#0071bc`


packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -135,8 +135,8 @@ $ds-c-inset-border-width: $spacer-half;
     opacity: 1;
   }
   &::before {
-    background-color: $color-primary;
-    border-color: $color-primary-darker;
+    background-color: $choice-checked-background-color;
+    border-color: $choice-checked-background-color;
```
bernardwang, line 139: Shouldnt this be `$choice-border-color`?

line47: previously `$choice-checked-background-color` was used https://github.com/CMSgov/design-system/blame/bb15cccb521dc283ae53577906ac83f0a83b0159/packages/design-system/src/styles/components/_Choice.scss#L104

`$choice-border-color` is `$color-base` and is used for when checkboxes are not checked

bernardwang:
@line47 Could we maybe expand this PR, and write a ticket for it? I'm thinking it would be nice to fix the inverse disabled styling as well, which adds an additional box shadow instead of just using `$choice-border-color-inverse`. 


After we merge the PR that reverts the checkbox change, the resulting requirements of this ticket could be:
- Unrevert the checkbox change
- Use `$choice-border-width` consistently in the disabled style
- Improve disabled choice styling by consistently using the same variables, and potentially adding new variables to use.

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -73,7 +73,7 @@ $ds-c-inset-border-width: $spacer-half;
 
       &::before {
         background-color: $color-gray-lighter;
-        border: 1px solid $color-gray-light;
+        border: $choice-border-width solid $color-gray-light;
```
line47, line 76: I'm unsure about the `border-width` change here mainly because we are changing the look of the disabled check/radio from 1px to 2px without consulting other teams on this change/update and am wondering how that would impact or surprise them downstream.

line47:
@bernardwang I like the idea of expanding this PR, I'll put a Jira ticket in and mention this in progress PR.

line47:
Tracked here https://jira.cms.gov/browse/WNMGDS-596

bernardwang:
@line47 I was thinking about the original issue with the background image checkmark not showing when printing, I wonder if it's possible to potentially have both the original image and CSS? Maybe theres some sort of styling possible where the original background image is on top of the CSS checkmark?

bernardwang:
@line47  @nichia I'm going to replace this branch with another one with a clean commit history!

PR #794, opened 2020-08-25
[WNMGDS-527] Add disabled prop to DateField

line47:
### Added
- Added `disabled` prop to disable the entire `DateField` component.
- Added test for `disabled` `DateField`

### Changed
- Updated React example name to be more explicit 

## How to test

1. Run the doc site locally and navigate to the Date field page
1. Update the React `DateField` example by adding the `disabled` prop 
1. See that the day, month, and year fields are disabled

### DIFF
--- a/packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
+++ b/packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
@@ -1,5 +1,5 @@
-import React, { Fragment } from 'react';
 import { DateField } from '@cmsgov/design-system';
+import React from 'react';
 import ReactDOM from 'react-dom';
 
 class ControlledDateField extends React.PureComponent {
@@ -22,7 +22,7 @@ class ControlledDateField extends React.PureComponent {
               DateField using default <code>dateFormatter</code>
             </span>
           }
-          hint={'Try to enter a date with invalid number of digits'}
+          hint="Try to enter a date with invalid number of digits"
           monthValue={this.state.month}
           dayValue={this.state.day}
           yearValue={this.state.year}
@@ -34,16 +34,17 @@ class ControlledDateField extends React.PureComponent {
 }
 
 ReactDOM.render(
-  <Fragment>
+  <div>
     <DateField
-      label="DateField example"
+      label="DateField example with invalid year"
       errorMessage="Please enter a year in the past"
       monthDefaultValue="10"
       dayDefaultValue="31"
-      yearDefaultValue="2020"
+      yearDefaultValue="2050"
       yearInvalid
     />
+
     <ControlledDateField />
-  </Fragment>,
+  </div>,
   document.getElementById('js-example')
 );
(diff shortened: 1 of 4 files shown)


packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
```diff
@@ -34,16 +34,17 @@ class ControlledDateField extends React.PureComponent {
 }
 
 ReactDOM.render(
-  <Fragment>
+  <div>
```
bernardwang, line 37: Why was this changed to a div? Couldnt we keep it as a Fragment, or as the Fragment short syntax (i.e. `<>`)?

packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
```diff
@@ -34,16 +34,17 @@ class ControlledDateField extends React.PureComponent {
...
     <DateField
-      label="DateField example"
+      label="DateField example with invalid year"
       errorMessage="Please enter a year in the past"
       monthDefaultValue="10"
       dayDefaultValue="31"
-      yearDefaultValue="2020"
+      yearDefaultValue="2050"
```
bernardwang, line 43: This is a bit funny haha

packages/design-system/src/components/DateField/DateField.jsx
```diff
@@ -186,6 +189,10 @@ DateField.propTypes = {
    * By default `dateFormatter` will be set to the `defaultDateFormatter` function, which prevents days/months more than 2 digits & years more than 4 digits.
    */
   dateFormatter: PropTypes.func,
+  /**
+   * Disables the entire field.
```
bernardwang, line 193:
```suggestion
   * Disables all three input fields
```

packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
line47: 🤦

packages/design-system-docs/src/pages/components/DateField/DateField.example.jsx
line47: When I try to commit is get an eslint error, on Fragment. 
```
error  'Fragment' is defined but never used  no-unused-vars
```

bernardwang: The error message is saying you defined `Fragment` but didnt use it, so all you need to do is remove the unused `Fragment` import

line47: When i remove the Fragment import I but keep the `<Fragment>` I get this error. This is why I changed to use the `<div>`
```
 error  'Fragment' is not defined  react/jsx-no-undef
```

bernardwang: You want to use the Fragment shorthand instead of `Fragment`
```
<>
   <DateField> 
</>
```
you can find other examples of this if you search for `<>`

PR #795, opened 2020-08-26
[WNMGDS-487] Add aria-label to FormLabel label and ChoiceList legend

### LINKED ISSUE #708: Add the ability to add aria-label to ChoiceList
**Describe the issue:**
We (MCT) would like the ability to add an `aria-label` to the ChoiceList component. One of our radio input groups requires an `aria-label` to clarify for screen reader users the text displayed in the `legend`. 

https://design.cms.gov/components/choice/

line47:
### Added
- Added `ariaLabel` prop `<FormLabel>` `<label>` element

Closes #708

## How to test `<FormLabel>`
1. Run the doc site locally 
1. add the `aria-label` prop to the `<FormLabel>` example 
1. See that `aria-label` is rendered on the `<label>` element of the `<FormLabel>`  example

## How to test aria-label text on legend
1. Run the doc site locally 
1. Add the following code to the <ChoiceList> example to render the `arial-label` on the legend for the choice 
```JSX
<fieldset>
       <FormLabel component="legend" aria-label="testing Aria">
           Legend text   
       </FormLabel>
       <Choice defaultChecked hint="Checkbox A hint" name="checkbox_choice" type="checkbox" value="a">
           Checkbox A
       </Choice>
       <Choice name="checkbox_choice" type="checkbox" value="b">
           Checkbox B
       </Choice>
</fieldset>
```
1. See that `aria-label` is rendered on the `<legend>` element of the `<ChoiceList>` example


packages/design-system/src/components/DateField/__snapshots__/DateField.test.jsx.snap
```diff
@@ -6,6 +6,8 @@ exports[`DateField has custom yearMax and yearMin 1`] = `
 >
   <legend
     className="ds-c-label"
+    component="legend"
```
bernardwang, line 9: @line47 the addition of `component` and `hint` in the snapshots means that they are being passed to the html element inside `labelProps`, which we dont want.

packages/design-system/src/components/FormLabel/FormLabel.jsx
```diff
@@ -50,14 +50,24 @@ export class FormLabel extends React.PureComponent {
...
+      fieldId,
+      id,
+      children,
+      textClassName,
+      className,
+      inversed,
+      errorMessage,
+      requirementLabel,
+      ...labelProps
+    } = this.props;
```
bernardwang, line 63: we want to define all of our props here including `component` and hint` to ensure only extra props get passed to `labelProps`

line47: Thank you!

packages/design-system/src/components/FormLabel/__snapshots__/FormLabel.test.jsx.snap
```diff
@@ -186,6 +186,7 @@ exports[`FormLabel renders requirementLabel string 1`] = `
 exports[`FormLabel supports additional classNames 1`] = `
```
bernardwang, line 186: could we update this test case to be "supports additional attributes", and pass in something other than `labelClassName`?

packages/design-system/src/types/FormLabel/FormLabel.d.ts
```diff
@@ -51,6 +51,9 @@ export interface FormLabelProps {
   textClassName?: string;
 }
 
-export default class FormLabel extends React.Component<FormLabelProps, any> {
+export default class FormLabel extends React.Component<
+  FormLabelProps,
```
bernardwang, line 55: We need to extend the React.HTMLProps here, other wise passing in an attribute not defined in `FormLabelProps` will cause an error, i.e. `aria-label`

bernardwang:
@line47 is this ready to review? the definition file doesnt look updated yet

line47:
@bernardwang Pushing the def file updates here in a few 👍

line47:
@bernardwang ready for review now

packages/design-system/src/components/FormLabel/FormLabel.test.jsx
```diff
@@ -101,8 +101,8 @@ describe('FormLabel', () => {
     expect(wrapper).toMatchSnapshot();
   });
 
-  it('supports additional classNames', () => {
-    const props = { className: 'ds-u-foo', labelClassName: 'ds-u-bar' };
+  it('supports additional attributes', () => {
+    const props = { className: 'ds-u-foo', ariaLabel: 'testing aria' };
     const wrapper = shallow(<FormLabel {...props}>{labelText}</FormLabel>);
```
nichia, line 107: Maybe add the check for the ariaLabel here?
`expect(wrapper.prop('ariaLabel')).toBe('testing aria');`

line47: Thanks, @nichia I just updated the test 👍

PR #796, opened 2020-08-26
[WNMGDS-558] Followup table layout changes

nichia:
### Changed
- Change the stacked table layout to align with [uswds visual](https://designsystem.digital.gov/design-tokens/color/theme-tokens/)
- Combine TableHeaderCell & TableDataCell to form TableCell and add component props to identify 'th' or 'td'. 
  - Add alignment prop to TableCell to align content as `center`, `left` or `right`. 
  - Refer to [material-ui ds for an example](https://material-ui.com/components/tables/#structure).
- Update responsive table example to separate into horizontal scroll table and stackable table. 
- Replace html complex table example with complex/irregular headers table
- Add typescript definition files

### Stacked table layout:
Screen Shot 2020-09-02 at 8 58 13 AM
Screen Shot 2020-09-02 at 8 58 04 AM


nichia:
Follow-up changes for Table is ready for testing (note: typescript definition file still has some issues and is still being worked on).
@davidakennedy - the stacked table layout has been changed and is now similar to [uswds](https://designsystem.digital.gov/design-tokens/typesetting/overview/). In my testing using voice over, it seems to read similar to the uswds stacked format, however, I have noticed that it reads differently to the earlier version of our stacked table layout. I'm keen to hear your feedback.

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -2,68 +2,100 @@ import {
...
-        </TableDataCell>
-        <TableDataCell scope="row" stackedTitle="Year" headers={'column' + 4}>
-          1791
-        </TableDataCell>
-      </TableRow>
-    </TableBody>
-  </Table>,
+  <>
+    <Table scrollable>
+      <TableCaption>Horizontal scroll table</TableCaption>
```
bernardwang, line 15:
```suggestion
      <TableCaption>Responsive scrollable table</TableCaption>
```

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -2,68 +2,100 @@ import {
...
+        <TableRow>
+          <TableCell id="column_1">Document title</TableCell>
+          <TableCell id="column_2">Description</TableCell>
+          <TableCell id="column_3">Links</TableCell>
+          <TableCell id="column_4">Year</TableCell>
+        </TableRow>
+      </TableHead>
+      <TableBody>
+        <TableRow>
+          <TableCell stackedTitle="Document title" headers="column_1">
```
bernardwang, line 26: These stacked titles arent necessary for the scrollable example

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -2,68 +2,100 @@ import {
...
+        </TableRow>
+        <TableRow>
+          <TableCell stackedTitle="Document title" headers="column_1">
+            Bill of Rights
+          </TableCell>
+          <TableCell stackedTitle="Description" headers="column_2">
+            The first ten amendments of the U.S. Constitution guaranteeing rights and freedoms.
+          </TableCell>
+          <TableCell stackedTitle="Links" headers="column_3">
+            <em>
```
bernardwang, line 52: Could we remove these `em` elements just to simplify the example?

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -2,68 +2,100 @@ import {
...
-          </em>
-        </TableDataCell>
-        <TableDataCell scope="row" stackedTitle="Year" headers={'column' + 4}>
-          1791
-        </TableDataCell>
-      </TableRow>
-    </TableBody>
-  </Table>,
+  <>
+    <Table scrollable>
```
bernardwang, line 14: Could we use a custom `scrollableAlert` here that describes what the prop is for and why we need it?

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -2,68 +2,100 @@ import {
...
+            </em>
+          </TableCell>
+          <TableCell stackedTitle="Year" headers="column_4">
+            1791
+          </TableCell>
+        </TableRow>
+      </TableBody>
+    </Table>
+
+    <hr className="ds-u-margin-top--2" />
```
bernardwang, line 63: Could we remove this `hr` and just put the margin on the table element itself?

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -24,8 +24,8 @@ Style guide: components.table.complex
 `<Table>`
 
 `Table` is a container component that contains `TableCaption`, `TableHead` and `TableBody` as children,
-as well as `TableRow`, `TableHeaderCell` and `TableDataCell` for the table content. While the usage of
-these components follows ordinary HTML table semantics, this React implementation includes additional
+as well as `TableRow` and `TableCell` for the table content. While the usage ofthese components
+follows ordinary HTML table semantics, this React implementation includes additional
```
bernardwang, line 28:
```suggestion
as well as `TableRow` and `TableCell` for the table content. These components mostly follow ordinary HTML table semantics, but also include some additional 
```

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -76,23 +76,13 @@ Style guide: components.table.react-tablerow
 */
 
 /*
-`<TableHeaderCell>`
+`<TableCell>`
 
-`TableHeaderCell` renders a `<th>` element.
+`TableCell` renders a `<th>` element when the parent component is a header or otherwise a `<td>` element.
```
bernardwang, line 81:
```suggestion
`TableCell` dynamically renders a `<th>` or `<td>` element based on the parent component or user specified `component` prop. By default `TableCell` will automatically render a `<th>` element if the parent component is `TableHead`, otherwise it will render a `<td>` element.
```

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -75,12 +75,19 @@ export class Table extends React.PureComponent {
...
   renderChildren() {
     return React.Children.map(this.props.children, (child) => {
-      // Extend props on TableCaption before rendering.
-      if (this.props.scrollable && isTableCaption(child)) {
+      if (isTableCaption(child)) {
+        // Extend props on TableCaption before rendering.
+        if (this.props.scrollable) {
+          return React.cloneElement(child, {
+            _id: this.captionID,
+            _scrollActive: this.state.isTableScrollable,
```
bernardwang, line 83: Could we rename `this.state.isTableScrollable` to be `scrollActive` to avoid confusion?

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -172,6 +171,19 @@ Table.propTypes = {
...
+   * A stackable variation of the table.
+   * When `stackable` is set, `id` or `headers` prop is required in `TableCell`
+   */
+  stackable: PropTypes.bool,
+  /**
+   * Applies responsive styles to vertically stacked rows at different viewpoint sizes.
+   */
+  stackableBreakpoint: PropTypes.oneOf(['sm', 'md', 'lg']),
+  /**
+   * A striped variation of the table.
```
bernardwang, line 184:
```suggestion
   * Applies the striped variation of the table.
```

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -172,6 +171,19 @@ Table.propTypes = {
    * This prop will only be used when the `Table` `scrollable` prop is set and the table width is wider than the viewport.
    */
   scrollableNotice: PropTypes.node,
+  /**
+   * A stackable variation of the table.
```
bernardwang, line 175:
```suggestion
   * Enables responsive stacking styles at the specified `stackableBreakpoint`.
```

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -172,6 +171,19 @@ Table.propTypes = {
...
+   * When `stackable` is set, `id` or `headers` prop is required in `TableCell`
+   */
+  stackable: PropTypes.bool,
+  /**
+   * Applies responsive styles to vertically stacked rows at different viewpoint sizes.
```
bernardwang, line 180:
```suggestion
   * The viewpoint size at which responsive vertically stacked row styles are applied. Only used when the `stackable` prop is set to `true`.
```

packages/design-system/src/types/Table/Table.d.ts
```diff
@@ -30,8 +20,23 @@ export interface TableProps {
...
+   */
+  stackable?: boolean;
+  /**
+   * Applies responsive styles to vertically stacked rows at different viewpoint sizes.
+   */
+  stackableBreakpoint?: TableStackableBreakpoint;
+  /**
+   * A striped variation of the table.
+   */
+  striped?: boolean;
```
bernardwang, line 35: if we add a `striped` prop we should do the same for `borderless`

packages/design-system/src/types/Table/Table.d.ts
```diff
@@ -30,8 +20,23 @@ export interface TableProps {
...
   scrollableNotice?: React.ReactNode;
+  /**
+   * A stackable variation of the table.
+   * When `stackable` is set, `id` or `headers` prop is required in `Table`
+   */
+  stackable?: boolean;
+  /**
+   * Applies responsive styles to vertically stacked rows at different viewpoint sizes.
+   */
+  stackableBreakpoint?: TableStackableBreakpoint;
```
bernardwang, line 31: This is something we can iterate on, but it might be nice later on to add a `stackableHeaderWidth` prop or something, for adjusting the percentage in case 40% doesnt work

packages/design-system/src/types/Table/TableCaption.d.ts
```diff
@@ -23,6 +23,6 @@ export interface TableCaptionProps {
   _scrollableNotice?: React.ReactNode;
 }
 
-declare const TableCaption: React.FC<TableCaptionProps>;
+declare const TableCaption: React.FC<React.HTMLProps<TableCaptionProps>>;
```
bernardwang, line 26: This doesnt look right, you are passing in our own defined props into `React.HTMLProps`

bernardwang:
In order for the stacked design to be consistent with our default table styles, can we update the stacked borders to look more like this?
Screen Shot 2020-08-31 at 2 35 02 PM

bernardwang:
This is a requirement from the last PR review that I forgot about, but I think we should update the borderless design to remove the background color, this would be consistent with USWDS.



Similarly, the borderless stacked design could look something like this:
Screen Shot 2020-08-31 at 2 39 03 PM

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -55,37 +61,60 @@
...
+  tr,
+  caption {
+    display: block;
+  }
+
+  td,
+  th {
     border: 0;
+    display: grid;
+    grid-template-columns: 42% auto;
```
bernardwang, line 73: Why is this 42 instead of 40?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -55,37 +61,60 @@
...
+    position: relative;
+    z-index: 100;
```
bernardwang, line 75: what is this used for? is this to position it on top of the :before pseudo element?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -55,37 +61,60 @@
...
+      width: 40%;
+    }
+
+    > * {
+      grid-column: 2;
+    }
   }
 
   tr {
+    background-color: color('white');
```
bernardwang, line 92: Is this supposed to be a variable? this isnt compiling into CSS right for me

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -55,37 +61,60 @@
...
     border: 1px solid $color-gray;
+    border: 1px solid;
```
bernardwang, line 94: extra border here

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -55,37 +61,60 @@
...
+    padding-bottom: 2px;
+    position: relative;
+
+    &::after {
+      background-color: $color-gray-lightest;
+      bottom: 0;
+      content: '';
+      display: block;
+      left: 0;
+      max-width: 38%;
```
bernardwang, line 106: same here, any reason why its 38 instead of 40?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -55,37 +61,60 @@
...
+      position: absolute;
+      top: 0;
+      width: 38%;
+      z-index: 0;
```
bernardwang, line 110: Is the zIndex required?

packages/design-system/src/components/Table/Table.jsx
line47: These suggestions are great @bernardwang - I think we should change `viewpoint` to `viewport`

packages/design-system/src/types/Table/Table.d.ts
```diff
@@ -30,8 +20,23 @@ export interface TableProps {
    * This prop will only be used when the `Table` `scrollable` prop is set and the table width is wider than the viewport.
    */
   scrollableNotice?: React.ReactNode;
+  /**
+   * A stackable variation of the table.
+   * When `stackable` is set, `id` or `headers` prop is required in `Table`
+   */
+  stackable?: boolean;
+  /**
+   * Applies responsive styles to vertically stacked rows at different viewpoint sizes.
```
line47, line 29:
```suggestion
   * Applies responsive styles to vertically stacked rows at different viewport sizes.
```

packages/design-system/src/styles/components/_Table.scss
line47: I think setting to 42% starts the second column to the right a bit to mimic the left cell padding. 
Screen Shot 2020-08-31 at 4 21 08 PM 

If it is set to 40% the text is too close to the grey background and is not spaced evenly.

packages/design-system/src/styles/components/_Table.scss
line47: This is to position the `:before` element which is the table heading, on top of the `:after` element which is the grey background.

packages/design-system/src/styles/components/_Table.scss
line47: I think this could be changed to 40%

packages/design-system/src/styles/components/_Table.scss
line47: I don't think it's needed

packages/design-system/src/types/Table/Table.d.ts
line47, line 35: I think it would be better and more flexible to give info on classes that can be applied to the table instead of building in the `striped` prop?

nichia: The zebra `striped` prop is applicable for both the border and borderless (via className='ds-c-table--borderless') tables. I don't think we need another setting for borderless then.

nichia: Ah yes, for stacked table, the layout is bordered, we haven't checked the borderless class to style it accordingly.

nichia: Right, so we can apply to the Table component `className="className="ds-c-table--striped"` instead of usiing the `striped` prop

packages/design-system/src/styles/components/_Table.scss
bernardwang: hm, is there a better way of styling that cell padding? I could see this being confusing or hard to customize for folks

packages/design-system/src/styles/components/_Table.scss
bernardwang: could we style the after element to have -9999 z index instead?

packages/design-system/src/styles/components/_Table.scss
bernardwang: ok great, lets remove it

packages/design-system/src/types/Table/Table.d.ts
bernardwang: Most react components with variations provide a prop, rather than relying on the user to use the correct class. They still can always apply classes, but i think if theres a variation we should provide a prop for it, i.e. `striped` and `borderless`.

That being said, I think its also a possibility to provide a `variation` prop with `striped` and `borderless` as options, but i dont know if its a possibility for both striped and borderless variations together

nichia: Fixed it to have the borderless layout for stacked table for className='ds-c-table--borderless'

packages/design-system/src/types/Table/Table.d.ts
nichia: I've changed it to max-width of 120px basing off the uswds template and it would be good to add `stackableHeaderWidth` prop for adjusting the header size. Will note down.

packages/design-system/src/types/Table/TableCaption.d.ts
nichia: Thanks for picking that up. Fixed it

packages/design-system/src/styles/components/_Table.scss
nichia: Yes, updated to `grid-template-columns: 120px 1fr`

packages/design-system/src/styles/components/_Table.scss
nichia: Updated to have  -9999 z index for the `:after` pseudo element

packages/design-system/src/styles/components/_Table.scss
nichia: Removed, it wasn't necessary

packages/design-system/src/types/Table/Table.d.ts
nichia: I see where we could have another prop for `borderless` variant. Will note down on list to add later.

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -2,68 +2,87 @@ import {
...
+          <TableCell>
+            <a href="# ">https://billofrightsinstitute.org/founding-documents/bill-of-rights/</a>
+          </TableCell>
+          <TableCell>1791</TableCell>
+        </TableRow>
+      </TableBody>
+    </Table>
+
+    <Table className="ds-u-margin-top--2" stackable stackableBreakpoint="md">
+      <TableCaption>Stackable table</TableCaption>
```
line47, line 52:
```suggestion
      <TableCaption>Responsive stackable table</TableCaption>
```

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -24,8 +24,7 @@ Style guide: components.table.complex
 `<Table>`
 
 `Table` is a container component that contains `TableCaption`, `TableHead` and `TableBody` as children,
-as well as `TableRow`, `TableHeaderCell` and `TableDataCell` for the table content. While the usage of
-these components follows ordinary HTML table semantics, this React implementation includes additional
+as well as `TableRow` and `TableCell` for the table content. These components mostly follow ordinary HTML table semantics, but also include some additional 
 responsive features including horizontal scrolling and vertically stacked rows.
 
 @react-example Table.example.jsx
```
line47, line 30: Add responsive preview similar to what is found on other responsive pages https://design.cms.gov/utilities/font-size/

```suggestion
@react-example Table.example.jsx
@responsive
```

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -13,7 +13,7 @@ Style guide: components.table
 */
 
 /*
-Table with two headers and a caption
+Table with multi headers and a caption
```
line47, line 16: Should we show a basic non-responsive table under the React section as well?

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -172,6 +171,19 @@ Table.propTypes = {
    * This prop will only be used when the `Table` `scrollable` prop is set and the table width is wider than the viewport.
    */
   scrollableNotice: PropTypes.node,
+  /**
+   * Enables responsive styles to vertically stack rows at the specified `stackableBreakpoint`.
+   * When `stackable` is set, `id` or `headers` prop is required in `TableCell`
+   */
+  stackable: PropTypes.bool,
+  /**
+   * The viewpoint size at which responsive vertically stacked row styles are applied. Only used when the `stackable` prop is set to `true`.
```
line47, line 180:
```suggestion
   * The viewport size at which responsive vertically stacked row styles are applied. Only used when the `stackable` prop is set to `true`.
```

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
bernardwang: Hm... I just wanted to point out we currently only use this in Patterns and on responsive utilities. We currently dont have any official standards on when to use this but this is probably fine...

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
bernardwang: Personally I think it would make the page really long, and the basic usage is very similar to the existing example

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -55,40 +61,80 @@
...
+        background-color: transparent;
+        border-right: 0;
+      }
+    }
+
+    td:last-child,
+    th:last-child {
+      border-bottom: 2px solid;
     }
   }
```
line47, line 139: I think we should update the class name for the responsive classes to match the format of our other responsive classes. like .`ds-c-[breakpoint]-table--stacked`, `.ds-c-sm-table--stacked`

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
line47: Yeah, I was thinking it would be easier for folks to know when the responsive classes kick in as the breakpoints are called out in the tabs 
Screen Shot 2020-09-02 at 1 41 44 PM

packages/design-system/src/styles/components/_Table.scss
bernardwang: good call

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -69,18 +69,25 @@ export class Table extends React.PureComponent {
     const { scrollWidth, clientWidth } = this.container;
     const isScrollActive = scrollWidth > clientWidth;
```
bernardwang, line 70: Small styling preference, but one way this could be simplified is:
```
const scrollActive = scrollWidth > clientWidth;
this.setState({ scrollActive });
```

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -115,7 +123,7 @@ export class Table extends React.PureComponent {
       'aria-labelledby': this.captionID,
       'aria-live': 'polite',
       'aria-relevant': 'additions',
-      tabIndex: this.state.isTableScrollable ? '0' : null,
+      tabIndex: this.state.scrollActive ? '0' : null,
```
bernardwang, line 126: @davidakennedy usually wouldnt we set the `tabIndex` to `-1` for enabling focus on a non focusable element?

packages/design-system/src/types/Table/Table.d.ts
bernardwang: I believe this still needs to be done

packages/design-system/src/components/Table/Table.jsx
nichia: Looks like `tabIndex` of `0` is for keyboard focus - https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/tabindex

packages/design-system/src/components/Table/TableCell.test.jsx
```diff
@@ -0,0 +1,147 @@
...
+  }
+
+  return {
+    props: props,
+    wrapper: mount(<Table {...props}>{children}</Table>),
+  };
+}
+
+describe('Table', function () {
+  it('renders a table thead <th> element', () => {
```
bernardwang, line 56: Could we also add some tests for using the `component` and `scope` props, which overrides the default values based off the parent element

packages/design-system/src/components/Table/TableCaption.test.jsx
```diff
@@ -29,5 +29,7 @@ describe('Table', function () {
     const table = wrapper.find('caption');
     expect(table).toHaveLength(1);
     expect(table.hasClass('ds-c-table__caption')).toBe(true);
+
+    expect(wrapper).toMatchSnapshot();
```
bernardwang, line 33: Could we add a test for when TableCaption renders the scrollableNotice

packages/design-system/src/components/Table/TableRow.jsx
```diff
@@ -1,26 +1,40 @@
...
     </tr>
   );
 };
 
 TableRow.propTypes = {
   /**
-   * The table row contents, usually `TableDataCell` and `TableHeaderCell`.
+   * The table row contents, usually `TableCell`.
    */
   children: PropTypes.node.isRequired,
```
bernardwang, line 29: I dont think this needs to be required

packages/design-system/src/components/Table/TableCaption.jsx
```diff
@@ -8,11 +8,11 @@ export const TableCaption = ({
...
-  ...others
+  ...tableCaptionProps
 }) => {
   const classes = classNames('ds-c-table__caption', className);
   return (
-    <caption className={classes} id={_id} {...others}>
+    <caption className={classes} id={_id} {...tableCaptionProps}>
       {children}
       {_scrollActive && _scrollableNotice}
     </caption>
```
bernardwang, line 18: I don't think className needs to be inside the react prop docs: https://github.com/CMSgov/design-system/pull/796/files#diff-265b21c2ce32ee008f705e6c9081ce0fR30, as we pretty much want the user to treat `TableCaption`, `TableBody` and `TableHead` the same as the native HTML.

Could you add the @hide-prop tag to it?

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -76,23 +75,13 @@ Style guide: components.table.react-tablerow
 */
```
bernardwang, line 75: I actually wonder if we can remove the `@react-props` for `TableCaption`, `TableHead`, and `TableBody`, as they are all basically supposed to be used like HTML.
https://github.com/CMSgov/design-system/pull/796/files#diff-c408290eec1f1b2fbb29243804f68058L63

The only important info in the react prop docs for those components right now is what each component usually contains as its children, and we could easily move that to the component description. 

Before:
Screen Shot 2020-09-02 at 2 05 49 PM
After:
Screen Shot 2020-09-02 at 2 06 46 PM

nichia: That does keep the docs leaner to capture the info in description only.

nichia: There's an issue with our doc generation script if I remove the props, the `view source file` will not sure up. But if i just @hide-prop all the props, the doc site will just display the header without content. 
To avoid having an incomplete doc site, I'll keep this present layout and add this item to the list of changes we can look into.
Screen Shot 2020-09-02 at 4 45 21 PM

packages/design-system/src/components/Table/TableCaption.jsx
nichia: Will address this later

packages/design-system/src/components/Table/TableCell.test.jsx
nichia: Yes, already in the test (line 100 to 122) but I'll modified the text to say `renders a table tbody <th> row header element which overwrites default header row component to <th>` and `sets a table tbody <th> role="rowheader" which overwrites default role value "cell"`

packages/design-system/src/components/Table/TableCaption.test.jsx
nichia: Not sure how to set viewport size to activate scrollableNotice. Will work on it

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -58,7 +59,7 @@ Style guide: components.table.react-tablehead
 /*
 `<TableBody>`
 
-`TableBody` renders the `<tbody>` element and contains table rows defining the table data.
+`TableBody` renders the `<tbody>` element and will typically contain `TableRow` elements to define table data.
 
 @react-props TableBody.jsx
```
line47, line 64: Remove `@react-props TableBody.jsx` and other `@react-prop` sections for `<TableCaption>, <TableHead>, <TableBody>, <TableRow>`

```suggestion

```

nichia: If I remove that, then the link for `View source file` will disappear (see attached image further up).

line47 (approved):

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
bernardwang: @nichia Maybe we could write another ticket for this work?

packages/design-system/src/components/Table/TableCell.test.jsx
bernardwang: Yeah i think this test file would be good to revisit, I'm having some trouble understanding them, as I dont see the `component` and `scope` props being explicitly used or passed in

PR #797, opened 2020-08-26
[WNMGWINSHOP-607] TypeScript example project and definition fixes

pwolfert:
In order to be able to verify my changes to the TypeScript definitions, I added an example TS project using create-react-app. I had been trying to verify the changes in our Enroll prodcuts, but there were too many layers. I'll also use this example to verify my TS definitions for the `ds-healthcare-gov` child design system.

### Added
- A `create-react-app` example in TypeScript

### Fixed
- TypeScript definitions now export `getMonthNames` and `unmaskValue` functions
- Fixed TypeScript definition for `Badge` component that previously caused a compiler error to be thrown when using the `size` prop


pwolfert:
If you check out 966d44a, it uses `unmaskValue`, so you can run `yarn tsc --noEmit` in the example folder to verify that there are no errors, or just run the example with `yarn start`.

packages/design-system/src/types/Badge/Badge.d.ts
```diff
@@ -26,6 +26,8 @@ export interface BadgeProps {
   variation?: BadgeVariation;
 }
 
-declare const Badge: React.FC<React.HTMLProps<HTMLSpanElement> & BadgeProps>;
+type OmitProps = 'size' | 'label';
```
bernardwang, line 29: thank you for this fix!

packages/design-system/src/types/MonthPicker/MonthPicker.d.ts
```diff
@@ -67,14 +67,16 @@ export interface MonthPickerProps {
    * For internationalization purposes, the text for the "Select all"
    * button must be passed in as a prop.
    */
-  selectAllText: string;
+  selectAllText?: string;
```
bernardwang, line 70: Could we update the PropTypes in `MonthPicker.jsx` too? Its currently listed as required when it should be optional because theres already a default value

PR #799, opened 2020-09-01
[WNMGDS-593] Document disabled choices

bernardwang:
### Summary
- Use `disabled` and `inversed` props better in `Choice.jsx`
- Update `Choice.example.jsx`
- Updated and consolidated `ChoiceList.example.jsx` and `checkbox-radio.example.html`

- Deprecated `children` prop in favor of `label` in `Choice.jsx`
- Updated definition files for all components with complex Proptypes (`VerticalNav`, `StepList`,`ChoiceList`)

### How to test
- Review all updated examples, ensure inverse and disabled styles are applied consistently and correctly
- Test updated definition files


packages/design-system/src/components/ChoiceList/Choice.jsx
```diff
@@ -111,6 +112,7 @@ export class Choice extends React.PureComponent {
           className={inputClasses}
           id={this.id}
           onChange={this.handleChange}
+          disabled={disabled}
```
bernardwang, line 115: explicitly pass in `disabled` prop

bernardwang:
> For the `choice` example:
> * the `radio` buttons cannot be unchecked after it's been selected
> * it's hard to see the separation for the checkbox, radio examples. It would be good to include some form of grouping via fieldset

Thats the expected behavior for radio buttons
I decided to remove the fieldsets because thats covered in the `ChoiceList` example. People who are using the `Choice` component will be building their own custom stuff anyway

packages/design-system-docs/src/pages/components/ChoiceList/Choice.example.jsx
```diff
@@ -1,155 +1,66 @@
...
-      </Choice>
-    </fieldset>
-
-    <hr className="ds-u-margin-top--2" />
-
-    <fieldset className="ds-c-fieldset">
-      <legend className="ds-c-label">
-        <span>Radio buttons</span>
-      </legend>
-      <Choice defaultChecked name="radio_choice" type="radio" value="a">
```
line47, line 76: Consider adding a shared `name=` to the radio choices so the user can toggle them on and off to see how they work. Otherwise, you can't uncheck a radio button once you check it.

nichia: I'm thinking that it may be good to include `fieldsets` in the `choice` examples to highlight that if we use `choiceList`, the fieldsets/legends are taken cared off by the react `choiceList` component. On the other hand, if we use the `choice` component, we need to manage all these html element separately.

nichia: But, you're right, people will write custom code anyway, so it's also good to go without the `fieldsets` info.

packages/design-system-docs/src/pages/components/ChoiceList/Choice.example.jsx
```diff
@@ -1,155 +1,62 @@
...
+    <Choice
+      name="checkbox_choice_small"
+      size="small"
+      type="checkbox"
+      label="Small Checkbox E"
+      value="e"
+    />
+    <Choice name="radio_choice" type="radio" label="Radio A" value="a" />
+    <Choice
+      name="radio_choice_checked"
```
line47, line 44: Use the same `name=` for radio buttons here so they can be unchecked 

```suggestion
      name="radio_choices"
```

bernardwang:
@line47 @nichia I addressed both of your feedback, but I forgot to include definition file updates as well. Could you all re-review that part?

line47 (approved):
LGTM!

nichia (approved):
LGTM

packages/design-system-docs/src/pages/components/ChoiceList/Choice.example.jsx
bernardwang: I ended up adding fieldsets back @nichia

PR #800, opened 2020-09-02
[NO-TICKET] Remove Initialize parameter from tsc definition

nichia:
### Changed
Update MonthPicker tsc file definition to remove initializing the `short` parameter


bernardwang (approved):
Thanks for the quick fix @nichia

PR #805, opened 2020-09-03
[WNMGDS-611] Generically copy src files

bernardwang:
### Summary
- Files that arent located in `components`, `styles`, `fonts`, and `images` in `src` will automatically be copied to `dist`

### How to test
- Add files and folders under `src`, make sure they are in dist after running `yarn start`
- Add files that wont be processed by the build scripts (i.e. `css` files in `components`, `js` files in `styles`), make sure they arent in dist after running `yarn start`


nichia (approved):
Nice coverage of cases for files/folders other then the standard covered by build scripts

line47 (approved):

PR #809, opened 2020-09-04
Release 2.0.3

bernardwang:


nichia (approved):

line47:
Should we bump the Sketch XML file to 2.0.3 as well?

bernardwang:
> Should we bump the Sketch XML file to 2.0.3 as well?

I'm not sure, it doesnt seem like we were keeping the Sketch version and the NPM version synced, so I didnt update, but if we want the version number to match the release notes maybe we should update it to 2.0.2

bernardwang:
@line47 like if we wanted to keep the sketch version and code version synced we shouldnt have updated the version in the PR. 

Maybe going foward we could only bump the sketch version on releases with sketch changes, and then when we do, we make sure the version matches the correct release notes? This way we dont bump the sketch version every 2 weeks when theres no changes and confuse folks, but at the same time we still have release notes? 

Orrrrr.... we could maintain our own set of release notes in a markdown file haha. I really dont know the best solution here

line47 (approved):

PR #810, opened 2020-09-04
[WNMGDS-529] Error styling on radio button and checkbox options

line47:
### Added
- Added the class `.ds-c-fieldset--error`  to the `<fieldset>` tag when the `errorMessage` prop is used on `<ChoiceList>`.
    -  This class `.ds-c-fieldset--error` is then used to style the checkbox and radio border color for error states making the borders red.

Screen Shot 2020-09-04 at 4 32 47 PM

### How to test
1. run the doc site locally
1. Add an `errorMessage` to  a `<ChoiceList>` example
1. See that the choice radios or checkboxes are also red when an error is present
1. For disabled check or radio options see that the border with for inverse and non-inverse items is the same


nichia (commented):
The change looks good.Do you think there would be cases with error message and the 'children' of choiceList/choice need be outlined in red too?

line47:
Currently, we don't offer an `errorMessage` prop for `<Choice>` and I'm unsure how teams are using errors for the individual `<Choice>` component. So I'm not sure we have enough info to add that in this Pr.

nichia:
I tried adding the class `ds-c-fieldset--error` to the custom radio inputs example on Choice.example.jsx and the boxes are outlined in red (nice) and the child boxes do remain as per regular black outline.

bernardwang:
@line47 since there are visual changes to choice in this PR, what do you think about also including a fix for this issue we discussed in CMS Slack?
https://cmsgov.slack.com/archives/CT169N26L/p1599070354008200

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -29,6 +29,18 @@ $ds-c-inset-border-width: $spacer-half;
...
 
+// Style radio and checkboxes when errored
+.ds-c-fieldset--error {
+  .ds-c-choice + label::before,
+  .ds-c-choice:checked + label::before {
+    border-color: $color-error;
+  }
+  .ds-c-choice--inverse + label::before,
+  .ds-c-choice--inverse:checked + label::before {
+    border-color: $color-error-light;
```
bernardwang, line 40: This is great, I think this is what we should be doing, but I wanted to point out this isnt consistent with most of our form elements, for both HTML and React.

I.e. the html TextField example doesnt make the input border red.
Screen Shot 2020-09-08 at 11 35 13 AM

The React TextField applies an error border, but it doesnt apply a lighter error border with inverse.
Screen Shot 2020-09-08 at 11 39 40 AM

The way we implement inverse in these scenarios is also important, like in this case we rely on the parent element having `.ds-c-choice--inverse`, but in other components we directly have `ds-c-field--error` classes. We also have to consider how our implementation results in complexity for child design systems, like i dont think our child design systems are overriding or even aware of this `error-light` pattern.

bernardwang: basically ^^ this was a long winded way of saying i think we really need to think about how we deal with error and inverse styles across our inputs in general.

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -68,8 +68,17 @@ export class ChoiceList extends React.PureComponent {
...
+      'ds-c-fieldset',
+      {
+        'ds-c-fieldset--error': this.props.errorMessage,
+      },
+      this.props.className
+    );
+
     return (
-      <fieldset className={classNames('ds-c-fieldset', this.props.className)}>
+      // <fieldset className={classNames('ds-c-fieldset', this.props.className)}>
```
bernardwang, line 80: this can be removed

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -111,7 +123,7 @@ $ds-c-inset-border-width: $spacer-half;
 
     &::before {
       background-color: $color-gray-lighter;
-      border: 1px solid $color-gray-light;
+      border-color: $color-gray-light;
```
bernardwang, line 126: thanks!

packages/design-system/src/components/ChoiceList/Choice.test.jsx
```diff
@@ -113,7 +113,6 @@ describe('Choice', () => {
     const input = wrapper.find('input');
 
     expect(input.hasClass('ds-c-choice')).toBe(true);
-    expect(input.hasClass('ds-c-choice--right')).toBe(false);
```
line47, line 116: Removed this because we no longer support the right to left alignment of choice components and I think that should have been removed previously.

https://github.com/CMSgov/design-system/commit/680304ee58f908d1791842e7bd5e5d92076fae88
https://github.com/CMSgov/design-system/commit/f49673b5b718c9fc1395d1cb3f5ee3f6d441b173

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -133,20 +144,6 @@ $ds-c-inset-border-width: $spacer-half;
...
-.ds-c-choice--right + label {
-  padding-left: 0;
-  padding-right: $choice-size + $spacer-1;
-
-  &::before {
-    left: auto;
-    right: 0;
-  }
-}
-
```
line47, line 149: Removing instances of `.ds-c-choice--right` because we no longer support the right to left alignment of choice components and I think that should have been removed previously.

See these commits:
https://github.com/CMSgov/design-system/commit/680304ee58f908d1791842e7bd5e5d92076fae88
https://github.com/CMSgov/design-system/commit/f49673b5b718c9fc1395d1cb3f5ee3f6d441b173

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -68,8 +68,16 @@ export class ChoiceList extends React.PureComponent {
...
+      'ds-c-fieldset',
+      {
+        'ds-c-fieldset--error': this.props.errorMessage,
+      },
+      this.props.className
+    );
+
     return (
-      <fieldset className={classNames('ds-c-fieldset', this.props.className)}>
+      <fieldset className={fieldsetClasses}>
```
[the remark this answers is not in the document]
bernardwang: I also wonder for simplicity sake if we should just remove the `error-light` inverse style. Its not consistently implemented now anyway, and maybe we should just remove it from certain examples

line47:
Thanks for the help on this @nichia ! 
It's ready for another review and probably some refactoring 👍

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -30,13 +30,19 @@ export class ChoiceList extends React.PureComponent {
    * Creates a list of Choice components
    */
   choices() {
+    console.log(this.props.choices);
```
bernardwang, line 33: this should be removed

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -30,13 +30,19 @@ export class ChoiceList extends React.PureComponent {
...
       choiceProps.name = this.props.name;
       choiceProps.onBlur = (this.props.onBlur || this.props.onComponentBlur) && this.handleBlur;
       choiceProps.onChange = this.props.onChange;
       choiceProps.size = this.props.size;
       choiceProps.type = this.props.type;
+      choiceProps.className = choiceProps.className
+        ? choiceProps.className.concat(' ', this.props.errorMessage ? 'ds-c-choice--error' : null)
+        : this.props.errorMessage
+        ? 'ds-c-choice--error'
+        : null;
```
bernardwang, line 45: You want to be using the `classNames` dependency to handle this logic. Generally its not a good idea to use multiple tertiary statements like this, as it makes it hard to read.

I believe this section can be simply accomplished with:
```
choiceProps.className = classNames(choiceProps.className, {
   'ds-c-choice--error': this.props.errorMessage,
})
```

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -123,7 +135,6 @@ $ds-c-inset-border-width: $spacer-half;
 
     &::before {
       background-color: rgba($color-muted-inverse, 0.15);
-      box-shadow: 0 0 0 1px $color-muted-inverse;
```
bernardwang, line 126: What does this change do?

bernardwang:
@line47 Can we confirm with a designer if this visual change makes sense? This feels a bit odd to me.
Screen Shot 2020-09-09 at 3 08 52 PM

bernardwang:
The HTML example needs to be updated too 
Screen Shot 2020-09-09 at 3 08 45 PM

bernardwang:
Small thing thats not related to this PR, but the HTML and React Choicelist examples should be identical. I noticed the html inverse example used radio instead of checkboxes, could you also include a change in this PR to make them both the same?

packages/design-system/src/styles/components/_Choice.scss
bernardwang: Oh I see, its for the inverse styling.

bernardwang:
In general I'm thinking about how this will be at least a minor bump, as it's introducing a pretty big visual change to the ChoiceList component. Teams that are using the ChoiceList component will be seeing all these red outlines all of a sudden, and they might need a way to override that.

I'm wonder if maybe we should limit the scope of this, or make it easy to override or turn off.

packages/design-system/src/styles/components/_Choice.scss
line47: When box-shadow was there for inverse choices only it gave the appearance of a wider border. Take a look a the live documentation site https://design.cms.gov/components/choice/ 

I opted to remove this so the borders can be applied consistently to inverse and non-inverse disabled examples.

line47:
@bernardwang Can you say more what feels odd about this?

image

bernardwang:
@line47 I wasn't sure about the color contrast between the blue and the red border.

line47:
There is still some discussion around the best way to approach this design challenge. Multiple design colleagues feel this approach to highlighting the checkbox or radio by turning the border-color red could lead to confusion instead of clarity.  

We going to discuss this at the next design system family touchpoint meeting to try and figure a path forward

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -37,6 +37,9 @@ export class ChoiceList extends React.PureComponent {
       choiceProps.onChange = this.props.onChange;
       choiceProps.size = this.props.size;
       choiceProps.type = this.props.type;
+      choiceProps.className = classNames(choiceProps.className, {
```
bernardwang, line 40: I think this should be applied to the `inputClassName` not `className`. This is consistent with how `ds-c-field--error` works as well, and should simplify the CSS

line47: Hmmm, with choice items we hide the input element and are styling the before element based on the label `.ds-c-choice:checked + label::before`?

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -27,53 +67,22 @@ $ds-c-inset-border-width: $spacer-half;
       margin-left: $spacer-1;
     }
   }
+  // Style radio and checkboxes when errored
+  // &.ds-c-choice--error + label::before {
+  //   border-color: $color-error;
+  // }
```
line47, line 73: commented out the error styling so child design systems could use if they wanted to

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -10,6 +10,46 @@ $ds-c-inset-border-width: $spacer-half;
   opacity: 0;
   position: absolute;
 
+  + label {
```
line47, line 13: Tried to also start organizing this SCSS file into `.ds-c-choice` and `.ds-c-choice--inverse` sections

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
bernardwang: Right, that part doesnt change. What I'm saying is that we should be applying the `ds-c-choice--error` class to the `input`, similar to the other modifiers. 
Screen Shot 2020-09-23 at 2 07 50 PM

packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
```diff
@@ -2,58 +2,68 @@
...
-      id="checkbox-3"
-      type="checkbox"
-      name="checkbox-choices"
-      value="checkbox-3"
-      disabled
-    />
-    <label for="checkbox-3">
-      <span>Disabled choice 3</span>
-    </label>
+    <div>
```
bernardwang, line 5: Did we add these `divs` just to be consistent with how our `Choice` component renders with a div?

packages/design-system/src/styles/components/_Choice.scss
bernardwang: Hmm, this whole stylesheet is pretty confusing in general... I was thinking of separating the blocks into the 3 major variants, `ds-c-choice`, `ds-c-choice--inverse` and `ds-c-choice--error`. `ds-c-choice__checkedChild` would be another block altogether. Would you mind if i gave a shot at organizing it later?

line47: That's fine by me, thanks for asking!

packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
line47: OK, I can remove the `<div>` s

packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
```diff
@@ -41,7 +41,7 @@
       Example error message
     </span>
     <input
-      class="ds-c-choice"
+      class="ds-c-choice cds-c-choice--error"
```
bernardwang, line 44:
```suggestion
      class="ds-c-choice ds-c-choice--error"
```

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -37,6 +37,9 @@ export class ChoiceList extends React.PureComponent {
       choiceProps.onChange = this.props.onChange;
       choiceProps.size = this.props.size;
       choiceProps.type = this.props.type;
+      choiceProps.inputClassName = classNames(choiceProps.inputClassName, {
+        'ds-c-choice--error': this.props.errorMessage,
```
bernardwang, line 41: I think we probably have to use `classNames` here too, in case the user specifies an `inputClassName` for the Choice as well. I think this is just one of the problems with not having a separate prop for error :/

packages/design-system/src/styles/components/_Choice.scss
bernardwang: I take back what I said, what you laid out here is really good. I made some minor changes but kept most of what you organized here

bernardwang (approved):
Thanks for all the work on this @line47! I had one comment about the error message copy but everything else looks good to me!

packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
```diff
@@ -83,21 +88,20 @@
     <fieldset class="ds-c-fieldset">
       <legend class="ds-c-label"><span>Inverse example</span></legend>
       <span class="ds-c-field__hint ds-c-field__hint--inverse">Helpful hint text</span>
-      <span class="ds-c-field__hint ds-u-color--error-light">Example error message</span>
```
bernardwang, line 86: I noticed this change, so far we have been using `Example error message` across all the field components. If we make the change here I would prefer to make it everywhere. Otherwise, I would prefer to keep things consistent and leave this out

line47: Ah gotcha, I'll revert this then

bernardwang:
Sorry, I actually accidentally deleted the checked children styling! I'll add that back in right now

nichia (commented):
The red border for error state without a checked choice/button looks great. 

For error state where there’s a checked choice/button, will it work having red border on unchecked items alongside the regular border for checked items? 
Screen Shot 2020-09-25 at 8 22 57 AM
For these cases, maybe we can explore either having red border for all checkbox/radio buttons, or just keeping all items without the red border?

bernardwang:
@line47 It turns out the approach of only adding the error styles on HC.gov wont work, as we also need the other disabled/checked/focus styles to have precedence over the error styles. Since that would require us to copy the entire `Choice.scss` file over, I'm just going to opt for including the error style in the core under a flag, as originally discussed

PR #813, opened 2020-09-10
[WNMGDS-622] Update inverse examples

bernardwang:
### Summary
- This PR updates HTML and react examples for most form components
- This PR adds inverse examples to each component, ensuring the error state is displayed

### How to test
Walk through every updated component and ensure the examples are correct with the appropriate markup


packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
```diff
@@ -67,6 +68,18 @@ ReactDOM.render(
...
       </optgroup>
     </Dropdown>
+    <div className="example--wrapper example--inverse">
+      <Dropdown
+        labelClassName="ds-u-margin-top--0"
+        options={dropdownOptions}
+        defaultValue="1"
+        errorMessage="Example error message"
+        hint="Helpful hint text"
+        label="Inverse dropdown"
```
line47, line 78: On other inverse versions, we use the language `Inverse example` - We might want to be consistent here for the label and use `Inverse example`

line47: I'm wondering if we should use this type of format for all inverse examples `Inverse COMPONENT-NAME example`

line47:
Overall I think these changes look good! 

I would like to see the naming of labels for components be more consistent.

For example, on some component pages, we use variations for label names sometimes they use the component name with the word example, and sometimes they are more realistic examples you might find in an application.
- Checkbox example
- Phone number (for form label) 
- Date of birth (for Datefield) 
- Inverse example
- Inverse buttons
- Inverse field


What I'm thinking would be to follow this convention
- COMPONENT-NAME example 
- inverse COMPONENT-NAME example 


For example:
- Checkbox example
- Text field example
- Inverse text field example

What do you all think? @nichia @bernardwang

packages/design-system-docs/src/pages/components/FormLabel/form-label.example.html
```diff
@@ -1,9 +1,22 @@
...
+    <span class="ds-c-field__hint">
+      Optional. We'll use this number as a backup if we need to contact you about your application.
+    </span>
+    <span class="ds-c-field__hint ds-u-color--error" role="alert">
+      Please enter a valid phone number.
+    </span>
+  </label>
+  <div class="example--wrapper example--inverse">
+    <label class="ds-c-label" for="inverse-field-id">
+      <span>Inverse Example</span>
```
line47, line 13: Make `e` in example lower case
```suggestion
      <span>Inverse example</span>
```

nichia (commented):
Not related to this pr, but the `yarn test:e2e` failed on line 13 of choice.e2e.test.js - maybe we can replace line 13 with below line to fix it?
`    const el = await getElementByClassName('ds-c-fieldset');
`

packages/design-system-docs/src/pages/components/DateField/datefield.example.html
```diff
@@ -33,6 +37,7 @@
         name="birthdate[day]"
         autocomplete="bday-day"
         aria-describedby="dob-datefield"
+        value="31"
       />
     </div>
     <span class="ds-c-datefield__separator">/</span>
```
nichia: Add `ds-c-field--error` to line 50 class to keep it consistent with the inverse example

nichia:
Regarding https://github.com/CMSgov/design-system/pull/813#issuecomment-691235470 - I like the use of standard name convention for those examples which do not have a story/theme attached, similar to the `<choice> <choiceList> <dropdown> examples`. For those examples with a story context, it might look odd if we apply the convention (like `Phone number` replace with `Form label example` might look odd unless we change the wording of the subsequent lines)?

packages/design-system-docs/src/pages/components/DateField/datefield.example.html
```diff
@@ -108,7 +47,7 @@
           ><span>Year</span></label
         >
         <input
-          class="ds-c-field ds-c-field--error ds-c-field--year"
+          class="ds-c-field ds-c-field--year"
```
nichia, line 50:
```suggestion
          class="ds-c-field ds-c-field--error ds-c-field--year"
```
keep year field consistent with the inverse example

bernardwang:
> For example, on some component pages, we use variations for label names sometimes they use the component name with the word example, and sometimes they are more realistic examples you might find in an application.

I think this is definitely something we should discuss and potentially revisit for another PR. The reason why things are inconsistent is that there is a need to show variations and typical use cases. Ideally our examples could do both, but its often pretty tricky when we are trying to be as concise and helpful as possible.

packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
bernardwang: Yeah I noticed that as well, the reason why I use "Inverse dropdown" here is because the other dropdown examples use "dropdown" in the labels. The components that use "Inverse example" dont mention the component name in it. Maybe it would be better to avoid using dropdown in the label name?

nichia (approved):

PR #815, opened 2020-09-12
Omit props for choices that we override with values from ChoiceList

pwolfert:
For the ChoiceList `choices` prop, omit the props that we override with values from ChoiceList. Without this fix, ChoiceList choice objects would all require a `name` and `type`. I'm omitting all of the props here that are overridden by https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/ChoiceList/ChoiceList.jsx#L34-L43 because they don't get passed down to the Choice component.

I'm putting this PR up so we have something to talk about, but maybe part of it should be some kind of internal tests to protect against these kinds of regressions.

### Fixed
- TypeScript: Fix ChoiceList props' type definition requiring that members of `choices` array have `name` and `type` defined. These come from the parent `ChoiceList` component.

### DIFF
--- a/packages/design-system/src/types/ChoiceList/ChoiceList.d.ts
+++ b/packages/design-system/src/types/ChoiceList/ChoiceList.d.ts
@@ -6,16 +6,15 @@ export type ChoiceListSize = 'small';
 export type ChoiceListType = 'checkbox' | 'radio';
 
 // Omit props that we override with values from the ChoiceList
-export type ChoiceProps = Omit<
-  ChoiceComponentProps,
-  'inversed',
-  'name',
-  'onBlur',
-  'onChange',
-  'size',
-  'type',
-  'inputRef'
->;
+type OverridenChoiceProp =
+  | 'inversed'
+  | 'name'
+  | 'onBlur'
+  | 'onChange'
+  | 'size'
+  | 'type'
+  | 'inputRef';
+export type ChoiceProps = Omit<ChoiceComponentProps, OverridenChoiceProp>;
 
 export interface ChoiceListProps {
   /**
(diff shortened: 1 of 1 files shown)


nichia:
Thanks for catching the issue and working on this PR @pwolfert. Tested the change and it works for hc.gov child ds but I'm getting below error for m.gov child ds. Not sure what the missing argument is yet.
`error TS2314: Generic type 'Omit' requires 2 type argument(s).
TypeScript: 1 semantic error`

packages/design-system/src/types/ChoiceList/ChoiceList.d.ts
```diff
@@ -1,10 +1,22 @@
...
+export type ChoiceProps = Omit<
+  ChoiceComponentProps,
+  'inversed',
+  'name',
+  'onBlur',
+  'onChange',
+  'size',
+  'type',
+  'inputRef'
+>;
```
nichia, line 18:
```suggestion
// Omit props that we override with values from the ChoiceList
type OmitChoiceProps = 'inversed' | 'name' | 'onBlur' | 'onChange' | 'size' | 'type' | 'inputRef';
export type ChoiceProps = Omit<
  ChoiceComponentProps, OmitChoiceProps
>;
```

pwolfert: I also didn't realize that the other type defs were using a different syntax

bernardwang (approved):
I tested this in the CRA TS example and Choice/ChoiceList are working as expected. Both child DS run fine as well. Because this is a bug, I think it makes sense to keep this PR as is and make a new ticket for implementing a better TS testing process to avoid these regressions in the future

nichia:
@bernardwang - not sure why when I tested with m.gov child ds, I'm getting the below error.
 `TS2314: Generic type 'Omit' requires 2 type argument(s). TypeScript: 1 semantic error TypeScript: emit succeeded (with errors) (node:19181) UnhandledPromiseRejectionWarning: Error: TypeScript: Compilation failed`

M.gov child ds only works for me if i change the tsc omit code to as listed below. Wonder if it's due to my local laptop setup again.:
```
type OmitChoiceProps = 'inversed' | 'name' | 'onBlur' | 'onChange' | 'size' | 'type' | 'inputRef';
export type ChoiceProps = Omit<
  ChoiceComponentProps, OmitChoiceProps
>
```

PR #821, opened 2020-09-17
[WNMGDS-524] Fix eslint logging and run linter

bernardwang:
### Summary
- This PR ensures that eslint is run and reported for all files in the provided directory. Previously only some errors would show.
- It also fixes the logging to accurately display how many JS files were linted
- It runs eslint and fixes many linter errors

### How to test
- Run `yarn lint` on different directories
- Create linter errors to see if they are reported
- Use the `--fix` flag to see if automatically fixable linter errors are fixed
- Ensure the documentation around the CMSDS scripts are updated

### DIFF
--- a/examples/create-react-app/src/components/App.test.js
+++ b/examples/create-react-app/src/components/App.test.js
@@ -1,6 +1,6 @@
+import App from './App';
 import React from 'react';
 import { render } from '@testing-library/react';
-import App from './App';
 
 test('renders learn react link', () => {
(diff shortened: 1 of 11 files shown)


nichia (approved):
That's great work fixing the eslint! 
Tested the eslint logging and with fix option - works great.
Just some eslint error for `react/no-danger` and `react/jsx-no-target-blank` that requires manual disabling in the code before merging.

line47 (approved):
LGTM

PR #827, opened 2020-09-24
[WNMGDS-596] Improve choice variables and unrevert the checkmark change

bernardwang:
### Summary
- Adds several new choice theme variables and uses them in `Choice.scss`
- Unreverts the checkmark change #789


packages/design-system/src/styles/settings/variables/_forms.scss
```diff
@@ -9,10 +9,19 @@ $input-border-radius: $border-radius !default;
 $input-border-color: $color-base !default;
 $input-border-color-inverse: $color-black !default;
 
-$choice-checked-background-color: $color-primary !default;
+// Choice theme variables
+$choice-background-color: $color-background !default;
```
bernardwang, line 13: There is a variable for `background-color` and `border-color` for every choice state (i.e. default, checked, disabled).  In addition,  there are default and disabled variables for inverse choices.

I also added `$choice-border-radius` which will be useful in the Mgov DS

packages/design-system-docs/src/pages/components/ChoiceList/_ChoiceList.docs.scss
```diff
@@ -93,10 +93,14 @@ Style guide: components.choice.react
 
 The following Sass variables can be overridden to theme choice fields:
 
+- `$choice-size`
```
line47, line 96: I think it would be good to put these into a table with descriptions of what each variable does - USWDS follows a similar approach https://designsystem.digital.gov/components/alert/#component-settings 
Screen Shot 2020-09-24 at 3 25 14 PM

bernardwang: yeah, we should make a ticket for that for all the components and utilities I think. It would also be nice to redesign the guidance section in the future and organize it better :/

line47: I'm going to add a ticket for it now 👍

line47 (approved):

nichia (commented):
The change looks good on core and hc.gov child ds.
Noticed that the checkmarks is not centered on the medicare.gov child ds:
Screen Shot 2020-09-28 at 10 54 26 AM

line47:
@nichia I ran into that as well and I think this is because of the CSS checkbox. The M.gov design system might need to be updated to account for this unless we can figure a way to center the check no matter the size of the box containing it

bernardwang:
@nichia @line47 yep, thats exactly it. The CSS checkbox is absolutely positioned, unlike the background image which is automatically centered. We just need to update the mgov DS after we release

nichia (approved):

bernardwang:
https://github.com/CMSgov/design-system/pull/827#commitcomment-42734405

@line47 Part of this is from my laziness, but I wonder if it's necessary to add this. Right now it's pretty much internal to HC.gov only, and theres the strong possibility the choice error highlight  will become part of the core soon anyway. I'm a bit wary of having to maintain documentation for all these conditional styles (i.e. focus styles, which are documented but not visible?), and I wonder if it would be simpler to not surface it outside of our code.

line47:
@bernardwang so you're thinking about it in more of an internal-only thing to do some testing with rather than part of the core that will persist?

bernardwang:
@line47 Yeah I'm guessing the flag will likely be temporary right? We are starting out with just HC.gov, but I bet if we bring it to Allyssa again, Mgov will likely adopt this as well. That's why I'd favor just not documenting the internal flag on the doc site, and then just documenting the change in the release notes when we eventually make the change to the core as well.

PR #828, opened 2020-09-25
Fix Tooltips with React versions < 16.6

bernardwang:
### Summary
- Downgrade `react-transition-group` to `^2.0.0` to avoid the requirement of React 16.6 (they use `React.createContext`)
- Add `core-js` as a dep to the scripts package

### How to test
- Test the tooltips component in the HC gov child ds


nichia (approved):
LGTM for core & child ds.

PR #829, opened 2020-09-25
[WNMGDS-613] Additional table props

nichia:
### Changed
- Update borderless table style to not have header background color
- Update tests to include props and caption scrollable notice 
- Update display of data title on a stacked table to word wrap
- Update striped table to have a different header background color

### How to test
- Run `yarn test` and check that it completes successfully
- Check that borderless table style do not have header background color
- Update `stackedTitle` of <Table.example.jsx> example to have various length and check that the stacked table looks correct
- Update examples to test striped table for bordered & borderless table


line47 (approved):
LGTM

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -40,13 +41,9 @@
...
 }
 
-.ds-c-table__header--label span {
-  display: block;
-  text-align: center;
-  text-transform: uppercase;
-}
-
 @mixin ds-c-table--stacked {
+  $stacked-title-max-width: 160px;
```
line47, line 45: I think we might want to change this variable name to be more specific - what about `$table-stacked-title-max-width`?

line47 (dismissed):
Have a small change to consider here

packages/design-system/src/components/Table/TableBody.test.jsx
```diff
@@ -6,35 +6,36 @@ import TableRow from './TableRow';
...
 
   if (!children) {
-    children = <TableBody {...defaultTableBodyProps}>{defaultTableBodyChildren}</TableBody>;
+    children = <TableBody {...props}>{defaultTableBodyChildren}</TableBody>;
   }
-
   return {
     props: props,
-    wrapper: mount(<Table {...props}>{children}</Table>),
+    wrapper: mount(<Table>{children}</Table>),
```
bernardwang, line 23: dont you still want to apply the `props` to the wrapper?

bernardwang, line 23: Why are you rendering a <Table> in this test file? Generally if you are testing `TableBody` you will just be testing that component in isolation

packages/design-system/src/components/Table/TableCaption.test.jsx
```diff
@@ -4,15 +4,13 @@ import TableCaption from './TableCaption';
...
+function render(customProps = {}, captionProps = {}, children) {
   const props = Object.assign({}, customProps);
+  const captProps = Object.assign({}, captionProps);
 
   if (!children) {
-    children = <TableCaption {...defaultCaptionProps}>{defaultCaptionChildren}</TableCaption>;
+    children = <TableCaption {...captProps}>{defaultCaptionChildren}</TableCaption>;
   }
 
   return {
```
bernardwang, line 16: Same feedback here, I dont think the <Table> is necessary

packages/design-system/src/components/Table/TableCell.test.jsx
```diff
@@ -7,30 +7,24 @@ import TableRow from './TableRow';
 import { mount } from 'enzyme';
 
 const defaultTableHeadChildren = (
```
bernardwang, line 9: Same here, unit tests should be in isolation. If you want to ensure that `TableCell` is rendering correctly based off its parent element, you should be testing `_isTableHeadChild` in `TableCell.test.jsx`, and also testing that that prop is being set and passed onto the children correctly in `TableBody.test.jsx` and `TableHead.test.jsx`

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -1,13 +1,18 @@
 @import '../settings/index.scss';
 
+$table-header-background-color: $color-gray-lightest !default;
+$table-stripe-background-color: $color-gray-lightest !default;
+$table-stripe-header-background-color: $color-primary-alt-lightest !default;
```
bernardwang, line 5: We shouldnt use `$color-primary-alt-lightest` here I dont believe, unless we have designers consensus on it. Most designers said that the current gray background is ok because it is also bold, and there are other stylings to explore like increasing the border width. Since there is no consensus, I would just keep it as is.

bernardwang: I understand that in that CMS post someone said that existing tables had that style, but those tables weren't even using the CMSDS table styles. And even if they were, we shouldn't jump to change our styles just because of that one case.

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -1,13 +1,18 @@
...
+$table-stacked-title-max-width: 160px;
```
bernardwang, line 6: Is this referring to the max width of the header when the table is stacked? could we rename this and add a comment explaining what it is?
`table-stacked-header-max-width`

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -1,13 +1,18 @@
 @import '../settings/index.scss';
 
+$table-header-background-color: $color-gray-lightest !default;
+$table-stripe-background-color: $color-gray-lightest !default;
```
bernardwang, line 4: Can we rename these striped variables to be `table-striped-...` rather than `stripe`? Just to be consistent with what we are calling this variation

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -71,7 +86,7 @@
     border: 0;
     border-top: 1px solid;
     display: grid;
-    grid-template-columns: 120px 1fr;
+    grid-template-columns: minmax(0, $table-stacked-title-max-width) 1fr;
```
bernardwang, line 89: Oh whats this `minmax` function? I've never seen it before

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -80,8 +95,9 @@
       // stylelint-enable selector-no-qualifying-type
       flex: none;
       font-weight: bold;
-      max-width: 120px;
-      width: 50%;
+      max-width: $table-stacked-title-max-width;
+      width: 80%;
```
bernardwang, line 99: Why 80% here?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -40,10 +46,19 @@
...
+  }
+
+  tbody {
+    tr:nth-child(even) {
+      background-color: $table-stripe-background-color;
+    }
+  }
 }
 
 @mixin ds-c-table--stacked {
```
bernardwang, line 64: Why is this a mixin instead of a normal class?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -40,10 +46,19 @@
   }
 }
 
-.ds-c-table__header--label span {
-  display: block;
-  text-align: center;
-  text-transform: uppercase;
+@mixin ds-c-table--striped-style {
```
bernardwang, line 49: Same here, why are you using a mixin? I see it being applied twice in this style sheet but I dont think thats necessary. Generally we dont add mixins unless they are intended to be used across files or by the user. they are also supposed to be defined in our mixins folder. That's why I dont think we should be defining mixins without a specific purpose

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -1,13 +1,18 @@
 @import '../settings/index.scss';
 
+$table-header-background-color: $color-gray-lightest !default;
```
bernardwang, line 3: These variables should be documented in the Guidance section of the Table docs. See [here](https://github.com/CMSgov/design-system/blob/master/packages/design-system-docs/src/pages/components/ChoiceList/_ChoiceList.docs.scss#L92) for an example

packages/design-system/src/styles/components/_Table.scss
nichia: Using minmax to fix the inconsistent word-wrap issue for grid 
https://developer.mozilla.org/en-US/docs/Web/CSS/minmax 
https://css-tricks.com/css-grid-in-ie-debunking-common-ie-grid-misconceptions/

packages/design-system/src/styles/components/_Table.scss
nichia: what do you think if i change it to extend as called in 2 places?

packages/design-system/src/styles/components/_Table.scss
nichia: what do you think if i change it to extend for %ds-c-table--striped-style as it's called in 2 places?

packages/design-system/src/styles/components/_Table.scss
nichia: To keep some spacing before the vertical line of the data title

packages/design-system/src/styles/components/_Table.scss
bernardwang: Ah I forgot that makes sense. If thats the case could we name it something less official? like `striped-table`

packages/design-system/src/styles/components/_Table.scss
bernardwang: Ah right, I forgot. Same feedback as above then, no need for the ds class naming convention, just call it `table-stacked` or `table-stacked-style`

packages/design-system/src/styles/components/_Table.scss
bernardwang: oh cool! nice find for the fix. Could you add a comment briefly summarizing the issue and why you need it? Including the css-tricks link would be helpful too

bernardwang: Is this the most relevant part?
https://css-tricks.com/css-grid-in-ie-debunking-common-ie-grid-misconceptions/#fit-content-is-not-ie-friendly-but
we are able to get `fit-content` like behavior with minmax?

packages/design-system/src/styles/components/_Table.scss
bernardwang: I dont quite understand, do you have a screenshot of what that means? If there's a mix-width already, wouldnt we just want the width to be 100%? or even just remove it? I dont see any difference when i remove the width: 80%

nichia: Screen Shot 2020-09-30 at 1 58 34 PM

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -1,67 +1,12 @@
 @import '../settings/index.scss';
 
 $table-header-background-color: $color-gray-lightest !default;
-$table-stripe-background-color: $color-gray-lightest !default;
-$table-stripe-header-background-color: $color-primary-alt-lightest !default;
-$table-stacked-title-max-width: 160px;
+$table-striped-background-color: $color-gray-lightest !default;
+$table-striped-header-background-color: $color-gray-lightest !default;
+// Width of the header column for a stacked table
+$table-stacked-header-max-width: 160px;
```
bernardwang, line 7: Missing a 
```suggestion
$table-stacked-header-max-width: 160px !default;
```

packages/design-system/src/styles/components/_Table.scss
bernardwang: What about using `calc`? something like this:
`calc(100% - #{$spacer-2})`

And then also include a comment explaining why we need this here

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -153,26 +94,82 @@ $table-stacked-title-max-width: 160px;
...
+
+  th,
+  td {
+    border-left: 0;
+    border-right: 0;
+  }
+}
+
+.ds-c-table--stacked {
+  @include table-stacked;
```
bernardwang, line 139: nice thanks!

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -80,8 +40,9 @@
...
       flex: none;
       font-weight: bold;
-      max-width: 120px;
-      width: 50%;
+      max-width: $table-stacked-header-max-width;
+      width: 80%;
+      word-wrap: break-word;
     }
 
     > * {
```
bernardwang, line 48: We can handle this in a followup PR, but in general this scss file is pretty hard to read. I think it would be really helpful to add more comments to label each selector, and describe some of the code here

packages/design-system/src/styles/components/_Table.scss
bernardwang: @line47 I think this has been addressed, and we went with `$table-stacked-header-max-width` to be consistent with the header terminology

bernardwang:
Another thing we should address in a future PR, but I just noticed we are directly styling the html elements. We should be styling with classnames, this is what we do for all other components as well, and what [UKGov does](https://design-system.service.gov.uk/components/table/). In general, we want to avoid styling html elements directly unless we are intentional about it, like with `<a>`. Even in those cases we provide a flag to disable this

bernardwang (approved):
This LGTM! We can handle unit tests and other changes in a follow up

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -171,6 +171,10 @@ $table-stacked-header-max-width: 160px !default;
     tr:nth-child(even) {
       background-color: $table-striped-background-color;
     }
+
+    tr:nth-child(odd)::before {
+      background-color: transparent;
```
nichia, line 176: Great fix here

Issue #817, opened 2020-09-16
Cannot override Review component classes


radavis:
**Describe the bug**

```
<Review className="ds-u-border-bottom--0 ds-u-padding-y--0" />
```

generates the following HTML

```
<div class="ds-c-review ds-u-border-bottom--2 ds-u-padding-y--2 ds-u-justify-content--between ds-u-display--flex ds-u-border-bottom--0 ds-u-padding-y--0">...</div>
```

which achieves the desired result of eliminating the bottom border, but **does not** eliminate the y-axis padding.

**To Reproduce**
1. See above

**Expected behavior**
I would expect to be able to override the utility classes on this component to achieve the desired styling. 

**Screenshots**

![Screen Shot 2020-09-16 at 10 52 27 AM (3)](https://user-images.githubusercontent.com/1299034/93354271-cfed0580-f80a-11ea-82c6-d3d7c340d6b2.png)

**Desktop (please complete the following information):**
 - OS: macOS 10.15.6
 - Browser: Chrome
 - Version: 85

bernardwang:
@radavis Thanks for submitting this! It does look like the `<Review>` component is incorrectly applying utility classes instead of including it in the component styles. This issue is tracked [here](https://jira.cms.gov/browse/WNMGDS-632)

bernardwang:
This issue has been fixed in the latest v2.2.0 release. The review component doesnt generate utility classes by default anymore (the styles are now included inside `ds-c-review`), so your utility classes should work.

PR #836, opened 2020-10-06
[WNMGDS-632] Apply Review classes

nichia:
This fixes the issue of Review component classes cannot be overridden because the component applies utility styles which uses `!important`.

### Changed
- Update Review component to apply it's component styling instead of using utility classes.

### Added 
- `ds-c-review` and `ds-c-review__heading` classes 

### How to test
- Update Review example to [override classes](https://github.com/CMSgov/design-system/issues/817) and check that the styles applied correctly

### DIFF
--- a/packages/design-system-docs/src/pages/components/Review/review.example.html
+++ b/packages/design-system-docs/src/pages/components/Review/review.example.html
@@ -1,6 +1,6 @@
 <div class="ds-c-review">
   <div class="ds-u-margin-right--2">
-    <h3 class="ds-c-review__heading ds-text">
+    <h3 class="ds-c-review__heading">
       Name
     </h3>
     <div>Jane Doe</div>
@@ -11,7 +11,7 @@
 </div>
 <div class="ds-c-review">
   <div class="ds-u-margin-right--2">
-    <h3 class="ds-c-review__heading ds-text">
+    <h3 class="ds-c-review__heading">
       Date of Birth
     </h3>
     <div>December 3, 1981</div>
@@ -22,7 +22,7 @@
 </div>
 <div class="ds-c-review">
   <div class="ds-u-margin-right--2">
-    <h3 class="ds-c-review__heading ds-text">
+    <h3 class="ds-c-review__heading">
       States Lived
     </h3>
     <div>
--- a/packages/design-system-docs/src/pages/components/Review/review.example.html
+++ b/packages/design-system-docs/src/pages/components/Review/review.example.html
@@ -1,8 +1,6 @@
-<div
-  class="ds-u-border-bottom--2 ds-u-padding-y--2 ds-u-justify-content--between ds-u-display--flex"
->
+<div class="ds-c-review">
   <div class="ds-u-margin-right--2">
-    <h3 class="ds-text ds-u-margin-bottom--0 ds-u-font-weight--bold ds-u-display--inline-block">
+    <h3 class="ds-c-review__heading ds-text">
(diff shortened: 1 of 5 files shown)


line47 (approved):
This looks good to me

packages/design-system/src/components/Review/Review.jsx
```diff
@@ -25,10 +21,7 @@ export class Review extends React.PureComponent {
       onEditClick,
       editContent,
     } = this.props;
-    const classes = classNames(
-      'ds-c-review ds-u-border-bottom--2 ds-u-padding-y--2 ds-u-justify-content--between ds-u-display--flex',
-      className && className
-    );
+    const classes = classNames('ds-c-review', className && className);
```
bernardwang, line 24: I'm actually not sure if the check on `className` is necessary. `classNames()` might actually handle that automatically, @nichia can you test and make sure?

bernardwang:
We also need to update the `Review` html examples

packages/design-system-docs/src/pages/components/Review/review.example.html
```diff
@@ -1,8 +1,6 @@
-<div
-  class="ds-u-border-bottom--2 ds-u-padding-y--2 ds-u-justify-content--between ds-u-display--flex"
->
+<div class="ds-c-review">
   <div class="ds-u-margin-right--2">
-    <h3 class="ds-text ds-u-margin-bottom--0 ds-u-font-weight--bold ds-u-display--inline-block">
+    <h3 class="ds-c-review__heading ds-text">
```
bernardwang, line 3: What is this additional `ds-text` for?

nichia: Thanks for picking this up. Have updated.

packages/design-system/src/styles/components/_Review.scss
```diff
@@ -0,0 +1,16 @@
...
+  border-bottom: 2px solid $border-color;
+  display: flex;
+  justify-content: space-between;
+  padding-bottom: 16px;
+  padding-top: 16px;
+}
+
+.ds-c-review__heading {
+  display: inline-block;
+  font-weight: $font-bold;
```
line47, line 13: I think this font size needs to be added. Take a look at this local branch and compare to production and you will see the production header is smaller than locally. 

```suggestion
  font-weight: $font-bold;
  font-size: $base-font-size;
```

line47 (approved):
After the font-size is added this looks good to me

packages/design-system/src/styles/components/_Review.scss
bernardwang: nice catch!

packages/design-system/src/styles/components/_Review.scss
```diff
@@ -0,0 +1,17 @@
...
+  font-size: $base-font-size;
+  margin-bottom: 0;
+  margin-top: 0;
```
bernardwang, line 16: does this margin-top need to be here too? I dont see that originally in the utility classes

bernardwang: oh i see is this from the `ds-text`?

bernardwang (approved):
LGTM!

packages/design-system/src/styles/components/_Review.scss
nichia: It's from `ds-u-margin-bottom--0`

PR #837, opened 2020-10-07
[WNMGDS-568] Add styles for field error message class

nichia:
Add class `ds-c-field__error-message` and use in-place of `.ds-c-field__hint .ds-u-color--error` for handling the display of error message by the `FormLabel` component that currently has this error message handling.

### Changed
- Add classes `ds-c-field__error-message` and ` ds-c-field__error-message--inverse`.
- Use these classes for display of field error message instead of the current usage of `ds-c-field__hint .ds-u-color--error`

### How to test
- Check dev console log to make sure the new classes is applied for the following examples `Date field`, `Checkbox & Radio`  and `Form Label & Legend`

### DIFF
--- a/packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
+++ b/packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
@@ -37,7 +37,7 @@
   </fieldset>
   <fieldset class="ds-c-fieldset">
     <legend class="ds-c-label"><span>Radio example</span></legend>
-    <span class="ds-c-field__error" role="alert">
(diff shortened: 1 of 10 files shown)


line47:
I wonder if we should add something like `Apply the .ds-c-field__error-message class to error text.` here on the form label documentation page? 
https://github.com/CMSgov/design-system/blob/WNMGDS-568/add-field-error-class/packages/design-system-docs/src/pages/components/FormLabel/_FormLabel.docs.scss#L61

bernardwang (approved):
This looks good to me! Great work @nichia !

line47 (approved):

PR #839, opened 2020-10-07
[WNMGDS-546] - Add guidance for placeholder dropdowns

line47:
### Added
- Added examples of dropdowns with placeholders
- Added guidance for when to use a placeholder vs a selected default

### To Review 
- run the doc site locally 
- visit the dropdown page and see the updated guidance and examples with placeholders

### DIFF
--- a/packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
+++ b/packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
@@ -13,7 +13,7 @@ const dropdownOptions = [
   { label: 'Option 8', value: '8' },
 ];
 const dropdownOptions2 = [
-  { label: '--Please select an option--', value: '' },
+  { label: '- Select an option -', value: '' },
   { label: 'Option 1', value: '1' },
   { label: 'Option 2', value: '2' },
   { label: 'Option 3', value: '3' },
--- a/packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
+++ b/packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
@@ -12,12 +12,23 @@ const dropdownOptions = [
   { label: 'Option 7', value: '7' },
   { label: 'Option 8', value: '8' },
 ];
+const dropdownOptions2 = [
+  { label: '--Please select an option--', value: '' },
+  { label: 'Option 1', value: '1' },
+  { label: 'Option 2', value: '2' },
+  { label: 'Option 3', value: '3' },
+  { label: 'Option 4', value: '4' },
+  { label: 'Option 5', value: '5' },
+  { label: 'Option 6', value: '6' },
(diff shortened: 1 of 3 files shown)


packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
```diff
@@ -12,12 +12,23 @@ const dropdownOptions = [
   { label: 'Option 7', value: '7' },
   { label: 'Option 8', value: '8' },
 ];
+const dropdownOptions2 = [
+  { label: '--Please select an option--', value: '' },
```
bernardwang, line 16: Tiny thing, but maybe the default could be:
```suggestion
  { label: '- Select an option -', value: '' },
```

packages/design-system-docs/src/pages/components/Dropdown/_Dropdown.docs.scss
```diff
@@ -42,7 +42,7 @@ Style guide: components.dropdown.react
 
 - **Make sure to test.** Test dropdowns thoroughly with members of your target audience. Several usability experts suggest they should be the “UI of last resort.” Many users find them confusing and difficult to use.
 - **Avoid dependent options.** Avoid making options in one dropdown menu change based on the input to another. Users often don’t understand how selecting an item in one impacts another.
-- **Use a good default.** When most users will (or should) pick a particular option, make it the default: ``<option selected="selected">Default</option>``
+- **Use a good default.** When most users will (or should) pick a particular option, make it the slected default. **If a good default is not an option**, use a descriptive placeholder like `--Please select a state--` as the selcted default.
```
bernardwang, line 45:
```suggestion
- **Use a good default.** When most users will (or should) pick a particular option, make it the selected default. **If a good default is not an option**, use a descriptive placeholder like `- Select a state -` as the selected default.
```

bernardwang (approved):
Small suggestions but this looks good to me! Nice work

line47:
Thanks @bernardwang - I applied your suggestions 👍

nichia (approved):
LGTM

packages/design-system-docs/src/pages/components/Dropdown/Dropdown.example.jsx
```diff
@@ -12,12 +12,23 @@ const dropdownOptions = [
   { label: 'Option 7', value: '7' },
   { label: 'Option 8', value: '8' },
 ];
+const dropdownOptions2 = [
```
bernardwang, line 15: Actually, I wonder if we can make every single dropdown use the same options. Reason being is that this extra `dropdownOptions2` is going to show up in the code snippet for this example, and it makes the example code unnecessarily longer.

bernardwang: Plus I dont think its a bad idea in general for all our examples to use that placeholder option

PR #841, opened 2020-10-08
[WNMGDS-648] Apply cursor: not-allowed to disabled label and box

nichia:
### Changed
- Apply cursor: not-allowed to disabled label and box

- Limit the clickable area to box/button and label text

Before change (clickable area - green and blue background):
Screen Shot 2020-10-13 at 11 42 09 AM

Screen Shot 2020-10-13 at 10 59 37 AM

After change (clickable area - green and blue background):
Screen Shot 2020-10-13 at 11 41 33 AM

Screen Shot 2020-10-13 at 11 00 30 AM


bernardwang:
We need to be applying the cursor not allowed to the `<span>` inside the `label`, not to the entire label. Otherwise the cursor will change when the label has a huge width. We also need to handle this for disabled choices with hint text. 

Screen Shot 2020-10-08 at 10 30 35 AM

bernardwang:
Technically, I think theres an existing bug where `pointer: cursor` is applied to the entire label width instead of the span and button only. Could we also address that in this PR as well? Its fixed the same way as I described in my previous comment, just make sure to apply the `pointer` style to `span` and `::before` elements inside the `label`, rather than the entire label
Screen Shot 2020-10-08 at 11 16 39 AM

nichia:
Ah, thanks for picking that up. Will make the change for this and cursor pointer.

nichia:
@bernardwang I have updated so the the `cursor: pointer` only applies to label elements (right now, we only use span), before and after pseudo-element , and `cursor: not-allowed` applies for disabled label elements (right now we only have span) and the before and after pseudo-elements.

line47:
I'm seeing an issue where the hint text area is much wider then the actual hint text thus the not-allowed curror is applied anywhere in the yellow region of the screenshot. 

It looks like adding what @bernardwang mentioned above would work to add the following styles to `.ds-c-label`
```CSS
 flex-direction: column;
 justify-content: center;
 align-items: flex-start;
``` 

Screen Shot 2020-10-08 at 2 18 30 PM

nichia:
Thanks, I'll try that out. 
Another question, I noticed that I can check/uncheck the box by click anywhere on the light blue area. This PR change will result in the pointer showing on the label text and box area only, instead of the green, blue and beige area. Would that be an issue?

Screen Shot 2020-10-08 at 2 33 13 PM

line47:
I would see what you mention above as an improvement @nichia 
Clicking way out to the right of the actual label text and have it select seems like it could introduce confusion for users

nichia:
Ah, but changing the `cursor: pointer` to apply for the label text does not sync with the actual area for checking/unchecking the box. The current code has the `cursor:pointer` appears for all the area that we can check/uncheck the button (which is the label tag element as covered by the green, blue area). Not sure if we should change to have cursor:pointer to limit to label text area.

bernardwang:
> Ah, but changing the `cursor: pointer` to apply for the label text does not sync with the actual area for checking/unchecking the box. The current code has the `cursor:pointer` appears for all the area that we can check/uncheck the button (which is the label tag element as covered by the green, blue area).

I think what both @line47 and I are suggesting is that that's an issue. We want the pointer and click events to only work on the choice text and button. I believe the super wide label width is a bug. If we want, we can confirm this with folks by posting in the CMS slack

line47 (approved):
LGTM!

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -13,23 +13,22 @@ $ds-c-inset-border-width: $spacer-half;
...
+    align-items: flex-start;
+    cursor: pointer;
     display: flex;
+    flex-direction: column;
     flex-wrap: wrap;
     font-weight: $font-normal;
+    justify-content: center;
     margin: $spacer-1 0;
-    max-width: $measure-base;
+    max-width: fit-content;
```
bernardwang, line 24: We want to be doing this, we use `$measure-base` here for a reason. See this utility and theme variable:
http://localhost:3000/utilities/measure/

bernardwang: Prod:
Screen Shot 2020-10-13 at 3 12 39 PM

This PR:
Screen Shot 2020-10-13 at 3 13 54 PM

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -13,23 +13,22 @@ $ds-c-inset-border-width: $spacer-half;
 
   // Choice label
   + label {
-    align-items: center;
+    align-items: flex-start;
+    cursor: pointer;
```
bernardwang, line 17: As per [this feedback here](https://github.com/CMSgov/design-system/pull/841#issuecomment-705668831), I think we should be applying the cursor to the `span` and the `:before` elements inside a `label`, not to the label itself.

bernardwang: Instead of trying to make the label width smaller, we should make the children of the label not expand to the full width of the parent label.

bernardwang (changes requested):

packages/design-system/src/styles/components/_Choice.scss
nichia: Good pick up on the `$measure-base` utility. The issue in using this is that the `clickable area` is the width of the `$measure-base`, but the cursor pointer is only showing for the label text. 

I've changed the cursor pointer/not-allowed to apply only to the label text. 

Do you think we should still apply the cursor pointer/not-allowed only to the label text or is more accurate to have the cursor pointer apply for the `$measure-base` width?

line47:
Looks good to me, good catch @bernardwang 👍

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -13,11 +13,12 @@ $ds-c-inset-border-width: $spacer-half;
...
-    align-items: center;
-    cursor: pointer;
+    align-items: flex-start;
     display: flex;
+    flex-direction: column;
     flex-wrap: wrap;
     font-weight: $font-normal;
+    justify-content: center;
     margin: $spacer-1 0;
     max-width: $measure-base;
```
bernardwang, line 23:
```suggestion
    max-width: $measure-base;
    width: fit-content;
```

nichia: Thanks - it's ready for a rerun @bernardwang @line47

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -13,11 +13,13 @@ $ds-c-inset-border-width: $spacer-half;
 
   // Choice label
   + label {
-    align-items: center;
+    align-items: flex-start;
```
bernardwang, line 16: Is this necessary? I removed `align-items: flex-start;` and `flex-wrap: wrap;` and everything worked fine still

bernardwang (approved):
I had one comment about removing some unnecessary flexbox styles, but aside from that this looks great. Awesome work @nichia!

packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
```diff
@@ -86,7 +86,7 @@
   </fieldset>
   <div class="example--wrapper example--inverse">
     <fieldset class="ds-c-fieldset">
-      <legend class="ds-c-label"><span>Inverse example</span></legend>
+      <legend class="ds-c-label ds-c-label--inverse"><span>Inverse example</span></legend>
```
bernardwang, line 89: Is this because the CSS is off in IE11? None of the other inverse labels have this class, I'd prefer to add the inverse label to all of them, or none of them. For this PR, can we avoid making this change, and then we can handle it in the future?

bernardwang:
@nichia I tested again and it seems to work well in Chrome. However in IE11 the labels seem to go full width, similar to the existing behavior. This isnt ideal, but since this is what we already have, i think its fine. 

Unfortunately though, `fit-content` doesnt work in IE11.
https://caniuse.com/?search=fit-content
This breaks the `max-width` we set. it looks like we might need to work on this a bit more before merging
Screen Shot 2020-10-14 at 4 29 40 PM

packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
nichia: This is only for the html example, without `ds-c-label--inverse`, label text is black instead of white on ie11. I noticed that the components correctly applies these classes `"ds-c-label ds-c-label--inverse"` for inverse label.

nichia:
Interesting, on IE11 the formatting looks fine for me. However, the fit-content doesn't seem to work as the cursor pointer/not-allowed is being applied to the max-width instead of the label size. Let's continue to test out some more
Screen Shot 2020-10-14 at 5 43 35 PM

nichia:
Changing to `display: table` ends up fitting the width to label text and works for ie11 too

line47:
In my testing using the following works for latest versions of Chrome, Edge, and Firefox) 
``` CSS
  display: flex;
  flex-wrap: wrap;
  width: fit-content;
```

But does not work for IE11 because IE 11 does not understand fit-content

line47 (approved):

bernardwang (approved):
@nichia I must have had something wrong last time, as the max width does seem to be working in IE11. Nice work! 
Screen Shot 2020-10-15 at 1 57 52 PM

packages/design-system-docs/src/pages/components/ChoiceList/checkbox-radio.example.html
bernardwang: Right, thats what I assumed. I think for this PR we should remove this change due to the reasons I mentioned above. In this case I think its better to not fix it than to only fix one label. You are welcome to create a followup PR that fixes this for all labels, but dont include it here.

nichia: Great, will remove this fix and add a followup PR for fixing the html examples with inverse label.

PR #848, opened 2020-10-13
[NO_TICKET] Remove USWDS SCSS

bernardwang:
### Summary
- Until we have the tokens RFC theres no reason to include it in our deps. We arent using USWDS in master anyway so we shouldnt expose it via our SCSS


line47:
I agree - does it make sense to keep a USWDS branch around?

nichia:
I think having a branch would be useful to do exploratory testing during the RFC

bernardwang:
@line47 @nichia This PR is removing USWDS from master, it doesnt impact the other branches we have for USWDS. The one I have been working off of is `bernard/adopt-usdws-tokens`, you can see I merge into it in this PR https://github.com/CMSgov/design-system/pull/840

line47 (approved):

Issue #838, opened 2020-10-07
Allow masked fields to use aria-labelledby instead of aria-label


tbolt:
Currently, the masked currency field requires a label and optionally an `aria-label`. If you want to use `aria-labelledby` you can do so by passing in a custom parameter but this results in having both `aria-label` and `aria-labelledby`, which isn't ideal. (Though, using safari with voice over, _only_ the `aria-labelledby` was being announced, which is the intended behavior we are after.)

The scenario where we want to use `aria-labelledby` in lieu of `aria-label` is in tables with masked inputs-
<img width="745" alt="screen shot" src="https://user-images.githubusercontent.com/2279260/95338656-2a650900-0881-11eb-876c-aeb477bae941.png">

The experience is much better with `aria-labelledby` as it doesn't repeat the same "Enter amount in dollars" for every single input.

Additional resources:
https://www.w3.org/TR/WCAG20-TECHS/ARIA9.html#ARIA9-ex2
https://www.w3.org/WAI/WCAG20/Techniques/working-examples/ARIA9/ex2.html

davidakennedy:
Thanks for filing this issue, @tbolt 

I agree that changing this up is the way to go. Doing some research, it looks like this was added to give screen reader users more context around currency in the fields. See: https://github.com/CMSgov/design-system/pull/249#pullrequestreview-101265881

I do think removing the `aria-label` here is best, since it overrides everything else, meaning screen reader users won't get the context of the original label. But we may want to keep the context of letting users know this is a dollar amount field. I'll do some research, and share here.

bernardwang:
@davidakennedy I agree, I think we should remove the behavior of automatically adding an `aria-label` for currency masks. All that does currently enforce an `aria-label` with "Enter amount in dollars.", and I feel like we could maybe just move that to the guidance section. 

This behavior also introduces another weird issue where any `aria-label` provided to a currency mask is automatically overridden by `${this.props.label}. Enter amount in dollars.`

davidakennedy:
@bernardwang I agree. I took a closer look at this last week, and I feel like whether or not you need the extra "Enter amount in dollars." depends on how detailed the label is. You might need that. You might not.

bernardwang:
Issue is tracked here: https://jira.cms.gov/browse/WNMGDS-659

PR #851, opened 2020-10-15
[WNMGDS-659] Remove default currency ariaLabel

bernardwang:
### Summary
This PR fixes https://github.com/CMSgov/design-system/issues/838
A default `ariaLabel` was added to currency masks originally to give screen reader users more contexts, but this prevents users from defining their own `ariaLabel` for currency fields.

Instead of requiring a hardcoded `ariaLabel`, we want the `ariaLabel` as a normal prop.

### How to test
- Ensure `ariaLabel` isnt defaulted for currency mask


nichia (approved):
LGTM

line47 (approved):

PR #852, opened 2020-10-16
[WNMGDS-665] Add polyfills required for ie11 support

nichia:
### Summary
- Update Month Picker Component to add `.includes` property polyfill for IE11 support. 
- Update ChoiceList Component to use `indexOf` instead of `.includes` for IE11 support. 
- Update currency Masked field to use `indexOf` instead of `startsWith`
- Move IE11 support for Tooltips Popper to the Tooltips component instead of the polyfills in `createWebpackConfig`.  This way the applications using the CMSDS don't have do any extra polyfill config

### How to test
- Run the design system doc site locally
- Ensure Month Picker component works correctly on IE11
- Ensure ChoiceList `onComponentBlur` function works on IE11 and other browsers (currently, there're other warning messages showing up on IE11 which we'll look at separately)
- Ensure Masked field for currency works
- Ensure Tooltips component works


bernardwang:
I'm actually thinking maybe its easier to just not use `includes`. We could simply define a function like so:

```
// Custom String includes function to avoid IE11 polyfill
function includes(str, searchStr) {
   return str.indexOf(str) !== -1;
}
```

This way teams don't need to add another polyfill;

bernardwang:
@nichia I think its fine to just keep it inside MonthPicker. I dont believe it's used elsewhere right now

bernardwang:
@nichia Just reposting my feedback here for visibility, I think instead of including the [polyfills in webpack](https://github.com/CMSgov/design-system/pull/852/files#diff-7d0e43683dd0947e14cb601aa556b1d2fd3799899836c7cc9a92f41794ebfbfbR9), we should be able to include them inside the components themselves.

This would mean importing specific core-js polyfills inside the components that need them, i.e. MonthPicker and Tooltips

nichia:
Have updated to include the MonthPicker, ChoiceList and Tooltip components to import specific core-js polyfills and removed these polyfills from `webpackconfig`

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -1,3 +1,4 @@
+import 'core-js/stable/array/includes';
```
bernardwang, line 1: Can we add this comment too:
// Polyfills required for IE11 compatibility

packages/design-system/src/components/MonthPicker/MonthPicker.jsx
```diff
@@ -1,3 +1,4 @@
+import 'core-js/stable/array/includes';
```
bernardwang, line 1: Same here

bernardwang:
@nichia I didnt realize this earlier, but there is only 1 instance of `includes` in ChoiceList. Can we actually swap that one out for indexOf? Monthpicker has multiple so the polyfill makes more sense

bernardwang (approved):
I had one request to remove the `includes` from ChoiceList, but everything else looks great. Thanks for finding and fixing this @nichia !

line47 (approved):

PR #853, opened 2020-10-19
[WNMGDS-650] Table followup

nichia:
### Summary
- Add grid autoprefixer to get stackable table working on IE11
- Tidy table styles, add comments
- Update tests

### How to test
- Ensure responsive and stackable table works on various browsers


packages/design-system/src/styles/components/_Table.scss
```diff
@@ -30,25 +30,31 @@ $table-stacked-header-max-width: 160px !default;
   th {
     border: 0;
     border-top: 1px solid;
+    /* stylelint-disable-next-line value-no-vendor-prefix */
+    display: -ms-grid;
```
bernardwang, line 34: Hm, I'm wondering why our `browserlist` and `autoprefixer` setup doesnt do this automatically?

nichia: I'm not sure yet. But even with the prefix, still not working correctly on ie 11 yet. Still trying to figure out why.

bernardwang: IE11 only has partial support of grid https://caniuse.com/css-grid

bernardwang: I found this resource too: https://dev.to/eblairmckee/css-grid-in-ie11-it-s-possible-and-not-as-hard-as-you-think-okp

line47 (approved):

line47:
Tested in IE11 and it's working great!

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -30,26 +30,32 @@ $table-stacked-header-max-width: 160px !default;
...
+    /* autoprefixer grid: on */
     display: grid;
     // CSS grid doesn't support word-wrap. Use minmax as a workaround
     // https://css-tricks.com/css-grid-in-ie-debunking-common-ie-grid-misconceptions/#fit-content-is-not-ie-friendly-but
     grid-template-columns: minmax(0, $table-stacked-header-max-width) 1fr;
     position: relative;
 
-    // stylelint-disable selector-no-qualifying-type
+    /* stylelint-disable-next-line selector-no-qualifying-type */
     &[data-title]::before {
```
bernardwang, line 42: It would help to leave a comment explaining what this psuedo element is and what its for

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -30,26 +30,32 @@ $table-stacked-header-max-width: 160px !default;
...
+      /* autoprefixer grid: on */
+      grid-column: 1;
       max-width: $table-stacked-header-max-width;
-      // allow for spacing on right side of data-title column
       width: calc(100% - #{$spacer-2});
-      word-wrap: break-word;
     }
 
-    > * {
+    > span {
```
bernardwang, line 52: I know you already covered this comment in the JSX file, but it would be helpful to add a comment here to explain that TableCell content is wrapped in a span for IE11 grid support

packages/design-system/src/styles/components/_Table.scss
bernardwang: [Here's](https://github.com/CMSgov/design-system/pull/853/files#diff-2b709213c3173751c76a6658e5db4facf82984bc78f95a2598faf72564b4ebcfL63) another example of an area that could benefit from a comment

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -30,26 +30,32 @@ $table-stacked-header-max-width: 160px !default;
   th {
```
bernardwang, line 30: Can we move [these styles](https://github.com/CMSgov/design-system/pull/853/files#diff-2b709213c3173751c76a6658e5db4facf82984bc78f95a2598faf72564b4ebcfL77-L78) under this block? This way all `TableCell` styles are in the same block.

bernardwang:
This can be addressed in a followup, but I noticed `ds-c-table__caption` is the one exception where we use a className to style the table. To keep things consistent with the existing layout, we should probably either add classNames for all of them while continuing support existing behavior, or we should remove `ds-c-table__caption`

bernardwang (approved):
Great work on this @nichia! The IE11 grid fix is really a huge accomplishment! Kudos!

bernardwang:
One other comment, I noticed the scrollable alert in IE11 wasnt showing the icon correctly some reason. Did anyone else see this?

packages/design-system/src/styles/components/_Table.scss
nichia: Which styles to move to the `th` block on line 30? The link for `these styles` doesn't seem to work.

nichia:
The scrollable alert icon on IE11 looks the same as on other browsers for me.

line47:
This is what I saw in IE 11 
Screen Shot 2020-10-28 at 2 55 07 PM

I think what is causing the issues are the utility classes added to the alert `ds-u-margin-y--1 ds-u-font-size--small ds-u-font-weight--normal` 

I'm wondering if we add some custom class here or styles to the alert that appear in the caption instead of adding the utility classes.

nichia:
Mm, thanks for the pick up. Somehow, the icon looks the same on other browsers when I tested. I'll follow up separately to identify the cases for incorrect Alert icon display.

PR #855, opened 2020-10-20
[WNMGDS-645] [WNMGDS-488] - Add date picker and update font bug

line47:
### Added
- Added a component for the date picker (previously it wasn't it's own Sketch symbol) 

### Fixed
- Fixed the body font in the Sketch file to be `Body Text as Open Sans, 16`


savannahmillion (approved):
I can see the date field here in the Sketch file, so I believe this PR solves the problem! I have some notes:

- I'd be careful to call this the date field rather than the date picker so as not to confuse it with the month picker component (it's labeled date field in the project file itself, so no change here needed imo!) 
- It looks like this component includes error message text but doesn't appear to include the error message state for the field (3pc solid #e31c3d outline) – it's possible that's a state that I'm just not seeing (I haven't used Sketch in a while) but if it doesn't, I wonder if that should be added? I'd defer to users of this file for this decision
- I'm using an obsolete version of Sketch (63.1) & this appears bug-free to me, but it's possible folks with a more updated Sketch version may see differences that I can't

PR #856, opened 2020-10-20
[WNMGDS-667] Update inverse html examples

nichia:
### Changed
The checkbox and date field inverse html example is displaying the legend text in black instead of white for IE11.
Update to include the inverse styling so that CMSDS doc site displays legend text 
correctly on IE11 for:
- Checkbox example
- Date field example
eg:
Screen Shot 2020-10-20 at 9 55 14 AM
fixed: 
Screen Shot 2020-10-20 at 10 01 36 AM


bernardwang:
Do we know why this is happening in IE11? The `.ds-base--inverse` styles applied to `example--inverse` *should* be applying a white color. It would be helpful to identify the cause of the issue before moving to patch it

nichia:
It looks like `legend` tag on IE11 is not inheriting the class styling from the parents eg, it's not inheriting `example--inverse`, so setting the `.ds-c-label–inverse` fixes the `Choice List` and `Date field` examples. Not sure why as I'm not finding any write-ups about this happening on IE11 when I searched the web. 
Our other examples that uses `label` tag for `Dropdown` and `Label form` works fine on IE11 with inheriting the `example--inverse` class styles from it's parents.

line47:
I think this was an oversight on the inverse example code - `.ds-c-label--inverse` has been part of the FormLabel component for a while and I think not having that class lack specificity that IE 11 needs. 

I could not find any documentation on why this is happening though.

line47 (approved):

bernardwang:
@nichia @line47 I agree that the `.ds-c-label--inverse` class has been a part of FormLabel, but I think the inverse examples have been relying on the `ds-base--inverse` styling, which for some reason isnt working for legends. I havent been able to figure out why IE11 was not applying that correctly, which is an issue on its own. 

We could just add the inverse class but it doesnt feel right to fix the underlying issue, which is `ds-base--inverse` isnt working for legends. Also, even if we were to add the inverse class, we should add it consistently across all inverse forms.

nichia:
Yes, it's weird because the `label` for the same date field html example (eg. Day, Month, Year) do inherit the styling, but the `legend` doesn't

line47:
So it looks like it has something to do with specificity. 

If I add the following to the `.example--inverse` styling it works in IE11 - we may want to explore updating `.ds-base--inverse` to include the following 

```CSS
 fieldset legend { 
      color: $color-base-inverse;
  }
```

nichia:
Thanks @line47 , that's a good find. Interesting that ie11 has different specificity for legend tag

bernardwang:
> So it looks like it has something to do with specificity.
> 
> If I add the following to the `.example--inverse` styling it works in IE11 - we may want to explore updating `.ds-base--inverse` to include the following
> 
> ```css
>  fieldset legend { 
>       color: $color-base-inverse;
>   }
> ```

I'm not sure, what do you think is overriding the styles in `example--inverse`? `ds-c-label` and `ds-c-fieldset` dont have any styles for `color`. The code you added doesnt address any specificity issues, it just adds new styles to html elements, which I dont think we want to be doing.

packages/design-system-docs/src/pages/components/DateField/datefield.example.html
```diff
@@ -97,7 +97,7 @@
         <div class="ds-l-col--auto">
           <label class="ds-c-label ds-c-datefield__label" for="year"><span>Year</span></label>
           <input
-            class="ds-c-field ds-c-field--error ds-c-field--year"
+            class="ds-c-field ds-c-field--error ds-c-field--inverse ds-c-field--year"
```
bernardwang, line 100: Is this still necessary?

packages/design-system/src/styles/components/_FormLabel.scss
```diff
@@ -1,6 +1,8 @@
 @import '../settings/index.scss';
 
 .ds-c-label {
+  // Correct color inheritance from `legend` elements in IE
```
bernardwang, line 4: Can you link normalize.css for extra context to where we are getting this from?

bernardwang, line 4:
```suggestion
  // Correct color inheritance for `legend` elements in IE11
```

packages/design-system-docs/src/pages/components/DateField/datefield.example.html
nichia: This is for this change (https://github.com/CMSgov/design-system/pull/847) - i think we missed updating this html example

bernardwang (approved):
Still one more feedback about adding a comment linking to normalize.css, but aside from that this looks great

PR #860, opened 2020-10-22
[WNMGDS-676] Update help drawer toggle to be a button

bernardwang:
### Summary
Previously the help drawer toggle was a link, even though it is used like a button to activate the help drawer. This PR updates the toggle to be a transparent button, and adds a `ds-c-help-drawer__toggle` class

### How to test
- Ensure no visual regressions
- Test the toggle works well on screenreaders and IE11


nichia:
Update LGTM. 
Just left to update snapshot test to reflect removal of `inline` property as a default. Also, just a note that backstop test shows a very slight visual change on using button.

line47:
@bernardwang It looks like the plain HTML example needs to be updates as well for the help drawer toggle

line47 (approved):

nichia (approved):

bernardwang:
I'm waiting to merge this after @davidakennedy can take a look. My screenreader testing seemed fine, but I'm not sure if there are any other unintended consequences

davidakennedy (approved):
From an accessibility perspective, this is good to go. I tested in VoiceOver, NVDA, JAWS and Narrator.

@bernardwang One other thought I had reviewing and testing: By having `display: block;` on the button, it means that you can't put these in the middle of sentences. Maybe I'm missing a way to toggle that, but it seems like that's a use case.

line47:
There is an `inline` prop on the `<HelpDrawerToggle>` to switch to inline instead of block 👍

PR #863, opened 2020-10-30
[NO-TICKET] Update Helpdrawer e2e test

nichia:
### Change
Update the Helpdrawer e2e test to click on button instead of link


bernardwang (approved):
Great catch @nichia

PR #866, opened 2020-11-02
[WNMGDS-621] Add dense table style

nichia:
### Added
Add table-padding variables and dense table style for use:
- ds-c-table--dense
example:
Screen Shot 2020-11-03 at 1 03 56 PM

### Changed
Change scrollable alert font size to normal


packages/design-system-docs/src/pages/components/Table/table.example.html
```diff
@@ -2,7 +2,7 @@
 <% [ 
   { title: 'Default table', className: 'ds-c-table'}, 
   { title: 'Borderless table', className: 'ds-c-table--borderless'}, 
-  { title: 'Striped table', className: 'ds-c-table--striped'}, 
+  { title: 'Striped and densed table', className: 'ds-c-table--striped ds-c-table--dense'},
```
line47, line 5: Should we remove the modifier code here and replace with full HTML for all examples since we are trying to get away from that?

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -144,7 +144,7 @@ export class Table extends React.PureComponent {
 Table.defaultProps = {
   scrollableNotice: (
     <Alert
-      className="ds-u-margin-y--1 ds-u-font-size--small ds-u-font-weight--normal"
+      className="ds-u-margin-y--1 ds-u-font-size--normal ds-u-font-weight--normal"
```
line47, line 147: the class `ds-u-font-size--normal` can be removed from the code as it doesn't exist

```suggestion
      className="ds-u-margin-y--1 ds-u-font-weight--normal"
```

packages/design-system/src/components/Table/__snapshots__/Table.test.jsx.snap
```diff
@@ -63,7 +63,7 @@ exports[`Table applies scroll table 1`] = `
       _scrollActive={false}
       _scrollableNotice={
         <Alert
-          className="ds-u-margin-y--1 ds-u-font-size--small ds-u-font-weight--normal"
+          className="ds-u-margin-y--1 ds-u-font-size--normal ds-u-font-weight--normal"
```
line47, line 66: Update snapshots after removing `ds-u-font-size--normal`

line47 (changes requested):
This looks good there area a few things to adjust on class names and modifier example HTML

packages/design-system-docs/src/pages/components/Table/table.example.html
bernardwang: This isnt exactly modifier code, this is just the templating language we use, its called EJS.

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -5,6 +5,85 @@ $table-striped-background-color: $color-gray-lightest !default;
...
+  td {
+    border-left: 0;
+    border-right: 0;
+  }
+}
+
+.ds-c-table--dense {
+  th,
+  td {
+    padding: $table-padding-dense $spacer-2;
```
bernardwang, line 55: I would put the entire padding in the variable:
i.e.
```
padding: $table-padding-dense;
```

packages/design-system-docs/src/pages/components/Table/table.example.html
bernardwang, line 5: I would recommend splitting these examples up, so 1 dense table and striped table

packages/design-system/src/components/Table/Table.jsx
nichia: Thanks for picking that up.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -117,6 +117,8 @@ The following Sass variables can be overridden to theme table fields:
 - `$table-striped-background-color`
 - `$table-striped-header-background-color`
 - `$table-stacked-header-max-width`
+- `$table-padding-default`
```
bernardwang, line 120: I think this can just be `$table-padding`

bernardwang: and then the other one will be `$table-dense-padding`

packages/design-system-docs/src/pages/components/Table/table.example.html
```diff
@@ -3,6 +3,7 @@
   { title: 'Default table', className: 'ds-c-table'}, 
   { title: 'Borderless table', className: 'ds-c-table--borderless'}, 
   { title: 'Striped table', className: 'ds-c-table--striped'}, 
+  { title: 'Densed table', className: 'ds-c-table--dense'},
```
bernardwang, line 6: This should just be called "Dense table"

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -143,10 +143,7 @@ export class Table extends React.PureComponent {
 
 Table.defaultProps = {
```
bernardwang, line 144: We should be adding a `dense` prop to the table just like `striped`

bernardwang: Also make sure to update the Table definition files, I think those are probably outdated and need to be tested

packages/design-system/src/components/Table/Table.test.jsx
```diff
@@ -47,44 +47,41 @@ describe('Table', function () {
     const table = wrapper.find('table');
 
     expect(table.hasClass('ds-c-table--borderless')).toBe(true);
+  });
 
-    expect(wrapper).toMatchSnapshot();
+  it('applies dense table classes', () => {
```
bernardwang, line 52: Thanks for adding this!

packages/design-system/src/components/Table/Table.test.jsx
```diff
@@ -127,8 +122,6 @@ describe('Table', function () {
       const tableCaption = wrapper.find('TableCaption');
 
       expect(tableCaption.prop('_scrollableNotice')).toBe('foo scrollable notice');
-
```
bernardwang, line 130: I think we should keep this

packages/design-system/src/components/Table/Table.test.jsx
```diff
@@ -99,8 +96,6 @@ describe('Table', function () {
       expect(divWrapper.prop('aria-live')).toBe('polite');
       expect(divWrapper.prop('aria-relevant')).toBe('additions');
       expect(divWrapper.prop('tabindex')).toBeUndefined();
-
-      expect(wrapper).toMatchSnapshot();
```
bernardwang, line 103: Same here

packages/design-system/src/components/Table/Table.test.jsx
bernardwang: Was there a particular reason why you deleted all these snapshots? I think some are probably not necessary but it doesnt hurt to keep them. Other tests definitely should have them, because they test for markup changes.

nichia: Ah, I was thinking checking for the props/classes would be sufficient to cover the test, but yes, the snapshot can additionally catch the markup changes. Will add back.

packages/design-system-docs/src/pages/components/Table/_Table.docs.scss
```diff
@@ -117,6 +117,8 @@ The following Sass variables can be overridden to theme table fields:
 - `$table-striped-background-color`
 - `$table-striped-header-background-color`
 - `$table-stacked-header-max-width`
+- `$table-padding`
+- `$table-dense-padding
```
bernardwang, line 121: Missing a closing ` here

packages/design-system/src/components/Table/Table.test.jsx
bernardwang: I think you can remove the snapshots for simple class changes, but for any tests that involve more markup changes or multiple attribute changes, we should add the snaps

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -5,6 +5,85 @@ $table-striped-background-color: $color-gray-lightest !default;
...
+  }
+}
+
+.ds-c-table__wrapper {
+  overflow-x: auto;
+}
+
+.ds-c-table__caption {
+  font-weight: $font-bold;
+  margin-bottom: $spacer-1;
```
bernardwang, line 84: If there's already a `margin-bottom` here, can we just remove the margin utility from the default scrollable notice?

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -143,10 +145,7 @@ export class Table extends React.PureComponent {
...
   scrollableNotice: (
-    <Alert
-      className="ds-u-margin-y--1 ds-u-font-size--small ds-u-font-weight--normal"
-      role="status"
-    >
+    <Alert className="ds-u-margin-y--1 ds-u-font-weight--normal" role="status">
```
bernardwang, line 148: I kind of feel like we can remove these classnames altogether, and just add `font-weight: $font-normal` to `ds-c-alert__text`.

bernardwang: not really a huge deal, but its nice to avoid some utilities

packages/design-system/src/styles/components/_Table.scss
nichia: The spacing is less, but workable. What do you think?

Current:
Screen Shot 2020-11-04 at 4 04 44 PM

Removed margin-bottom from scrollable notice: Screen Shot 2020-11-04 at 4 05 10 PM

packages/design-system/src/components/Table/Table.jsx
nichia: I'd rather avoid adding `fond-weight: $font-normal` to `ds-c-alert__text` just in case it's a breaking change for products using this class which currently inherits the font-weight.

nichia:
The usa banner looks good. Can we add a toggle to `Here's how you know` for showing/hiding the content?
Screen Shot 2020-11-04 at 4 37 12 PM

packages/design-system/src/components/Table/Table.jsx
bernardwang: Ok, I'm ok with that too. I don't think it will be a breaking change as the default alert font weight is normal, but I agree its fine to play it safe

packages/design-system/src/components/Table/__snapshots__/Table.test.jsx.snap
```diff
@@ -45,80 +15,12 @@ exports[`Table applies responsive table 1`] = `
 </div>
 `;
 
-exports[`Table applies scroll table 1`] = `
```
bernardwang, line 48: @nichia Where is this snapshot from? It look important to keep

bernardwang (approved):
I had 1 last small comment on tests but this looks really great. The code has been cleaned up alot, great work Ni!

packages/design-system/src/components/Table/__snapshots__/Table.test.jsx.snap
nichia: Yes, that snapshot has been added back.

bernardwang:
@line47  this is ready to review again

line47:
Looking at the table PR I noticed the stripped stackable table looks like this - I wasn’t sure if that is the look we are going for where each section becomes stripped?

Screen Shot 2020-11-06 at 1 40 43 PM

I also realize this is out of scope for the PR - You all think a follow up issue should be created?

nichia:
Okay, leave it ATM until we get other feedback.

line47 (approved):

PR #867, opened 2020-11-02
Fix substep definition file

bernardwang:

### DIFF
--- a/packages/design-system/src/types/StepList/Step.d.ts
+++ b/packages/design-system/src/types/StepList/Step.d.ts
@@ -1,28 +1,8 @@
 import * as React from 'react';
-
-// Step Object is used in <Step> and <SubStep>
-export interface StepObject {
-  id?: string;
-  href: string;
-  title?: string;
-  heading: string;
-  headingLevel?: '1' | '2' | '3' | '4' | '5';
-  description?: string;
-  linkText?: string;
-  completed?: boolean;
-  started?: boolean;
-  isNextStep?: boolean;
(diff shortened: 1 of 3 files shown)


nichia (commented):
Getting this error message on the react-create-app-typescript that references the core steplist.
```Failed to compile
/Users/tchutnichia/Code/design-system/examples/create-react-app-typescript/src/components/App.tsx
TypeScript error in /Users/tchutnichia/Code/design-system/examples/create-react-app-typescript/src/components/App.tsx(21,11):
Type '{ id: string; heading: string; href: string; started: boolean; completed: boolean; }' is not assignable to type 'StepProps'.
  Object literal may only specify known properties, and 'id' does not exist in type 'StepProps'.  TS2322

    19 |       steps={[
    20 |         {
  > 21 |           id: 'taxYear',
       |           ^
    22 |           heading: 'Choose a tax year',
    23 |           href: '#step-1',
    24 |           started: true,
This error occurred during the build time and cannot be dismissed.```

However, if i update the package & app to reference the updated hcgove-design-system steplist, it works.

bernardwang:
@nichia I'm unable to reproduce that error on my local branch. This is the same code as the HC.gov DS PR, so it should work the same. Can you try again with a fresh install and build? We need to make sure both the core DS and the CRA typescript example are reset and rebuilt

nichia (approved):
Tried updating with `yarn build` and it's working!

line47 (approved):

PR #869, opened 2020-11-03
[WNMGDS-480] Add usa banner component

line47:
### Added
- Adding new USA Banner component
- Adding SVG icons for USA Flag, HTTPs Icon, Lock Icon and Dot Gov Icon, which are also used in the banner
- Added documentation page for USA Banner as well as guidance 
- Added HTML and React examples 

### How To test
- Then visit the USA Banner component page and ensure that the content expands and collapses properly and the styling is correct.


packages/design-system-docs/src/pages/components/UsaBanner/_UsaBanner.docs.scss
```diff
@@ -0,0 +1,45 @@
...
+/*
+---
+
+### When to use
+
+- To identify as an official government site. Most government sites should use the banner.
+
+### When to consider alternatives
+
+- If you don't use a .gov/.mil domain and HTTPS. The banner text identifies .gov/.mil domains and HTTPS as indicators that a website is an official government website. Use the banner only if your site uses both the proper TLD and HTTPS.
```
bernardwang, line 32: cms wont have `.mil` i believe

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -0,0 +1,47 @@
+<div class="ds-c-usa-banner ds-u-padding-x--3">
```
bernardwang, line 1: We generally dont want to use utilities with the banner classes so its easier for folks to override

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -0,0 +1,47 @@
...
+  <header class="ds-c-usa-banner__header">
+    <div
+      class="ds-u-display--flex ds-u-flex-direction--row ds-u-align-items--start ds-u-sm-align-items--center"
```
bernardwang, line 4: Can this be combined with `ds-c-usa-banner__header`?

packages/design-system/src/components/UsaBanner/UsaBanner.e2e.test.js
```diff
@@ -0,0 +1,20 @@
...
+import assertNoAxeViolations from '../helpers/e2e/assertNoAxeViolations';
+import { getElementByClassName } from '../helpers/e2e';
+
+const rootURL = `${ROOT_URL}/example/components.usabanner/`;
+
+describe('Usa banner component', () => {
+  it('Usa banner should render', async () => {
+    await driver.get(rootURL);
+
+    const el = await getElementByClassName('c-usa-banner');
```
bernardwang, line 13: `c-usa-banner` needs to be updated

bernardwang:
oops sorry, i didnt see this is work in progress

packages/design-system-docs/src/pages/components/UsaBanner/_UsaBanner.docs.scss
```diff
@@ -0,0 +1,45 @@
+/*
+Usa Banner
```
bernardwang, line 2: Can we use upper case, so USA Banner

packages/design-system-docs/src/pages/components/UsaBanner/_UsaBanner.docs.scss
```diff
@@ -0,0 +1,45 @@
...
+
+### When to consider alternatives
+
+- If you don't use a .gov domain and HTTPS. The banner text identifies .gov domains and HTTPS as indicators that a website is an official government website. Use the banner only if your site uses both the proper TLD and HTTPS.
+- Any time it would be misleading. The banner should be used to reduce confusion. Avoid using the banner on any site meant only for testing or otherwise not meant to be identified as an official government website.
+
+### Usage
+
+- Use the provided text without customization. The banner is most effective as an identifier and a learning tool when its message is consistent across government websites. With only a few exceptions (see Implementation guidance), sites should use the TLD-appropriate text provided, unaltered. Use the Spanish version of the banner for Spanish-language websites.
+- Use the version appropriate to your website’s TLD. If your project uses a .mil top-level domain, use the .mil banner text.
```
bernardwang, line 38: We can remove this

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -0,0 +1,47 @@
...
+      <p class="ds-c-usa-banner__header-text">
+        <span>An official website of the United States government</span
+        ><button class="ds-c-usa-banner__button" aria-expanded="false" aria-controls="gov-banner">
+          Here’s how you know
+        </button>
+      </p>
+    </div>
+  </header>
+  <div class="ds-c-usa-banner__content ds-u-padding-y--3" id="gov-banner" hidden="">
+    <div
```
bernardwang, line 16: Same here, ideally these utilities can be removed

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -0,0 +1,81 @@
...
+            </p>
+          </div>
+        </div>
+      </div>
+    </div>
+  );
+};
+
+UsaBanner.propTypes = {
+  isBannerOpen: PropTypes.bool.isRequired,
```
bernardwang, line 77: Can change this so that the component handles the banner toggling? I could imagine this working more similarly to the Tooltip component rather than like the HelpDrawer component, where the opening/closing is handled by the component itself

bernardwang: This is especially because if you look at the USWDS banner, on mobile the interaction for toggling the banner changes from clicking the link, to clicking the entire banner.

bernardwang: Maybe if we find that teams need to control the banner more, we could add `bannerOpen` and `toggleBanner` as optional props?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -0,0 +1,62 @@
...
+            className="ds-c-usa-banner__button"
+            aria-expanded={props.isBannerOpen}
+            aria-controls="gov-banner"
+          >
+            Here’s how you know
+          </button>
+        </p>
+      </header>
+      <div className="ds-c-usa-banner__content" id="gov-banner" hidden={!props.isBannerOpen}>
+        <div className="ds-u-display--flex ds-u-flex-direction--column ds-u-sm-flex-direction--row ds-u-flex-wrap--nowrap">
```
bernardwang, line 30: I think these utilities can still be handled by `ds-c-usa-banner__content`

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -0,0 +1,62 @@
...
+
+const UsaBanner = function (props) {
+  const bannerHeader = props.isBannerOpen
+    ? 'ds-c-usa-banner__header ds-c-usa-banner__header--expanded'
+    : 'ds-c-usa-banner__header';
+
+  return (
+    <div className="ds-c-usa-banner">
+      <header className={bannerHeader}>
+        <UsaFlagIcon />
```
bernardwang, line 16: This icon is great! how did you end up converting to SVG?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -0,0 +1,62 @@
...
+            aria-controls="gov-banner"
+          >
+            Here’s how you know
+          </button>
+        </p>
+      </header>
+      <div className="ds-c-usa-banner__content" id="gov-banner" hidden={!props.isBannerOpen}>
+        <div className="ds-u-display--flex ds-u-flex-direction--column ds-u-sm-flex-direction--row ds-u-flex-wrap--nowrap">
+          <div className="ds-c-usa-banner__guidance">
+            <DotGovIcon />
```
bernardwang, line 32: Same here, how did you get this SVG?

packages/design-system/src/components/index.js
```diff
@@ -30,4 +30,9 @@ export { default as TabPanel } from './Tabs/TabPanel';
 export { default as TextField, unmaskValue } from './TextField/TextField';
 export { default as Tooltip } from './Tooltip/Tooltip';
 export { default as TooltipIcon } from './Tooltip/TooltipIcon';
+export { default as UsaBanner } from './UsaBanner/UsaBanner';
+export { default as UsaFlagIcon } from './UsaBanner/UsaFlagIcon';
```
bernardwang, line 34: I don't think we need to export these icons, as they are only used inside the banner component. If they were generalized icons that we wanted folks to use I think it make sense to export them

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
bernardwang: Oh i see, we already had this SVG in the images

bernardwang: Instead of redefining our .svg files as react components, I think we should consider using this: 
https://github.com/airbnb/babel-plugin-inline-react-svg

Once we install that dependency to our scripts package, and we update our babel.config.js, we should be able to directly import our svg icon files and use them as React components

bernardwang: This could also be a very promising solution for moving away from using icons as background images

bernardwang: It would also be really nice to update our svg files to be cleaner and formatted well, so they are easy to read and use.

packages/design-system/src/images/icon-dot-gov.svg
```diff
@@ -0,0 +1 @@
+<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 54 54"><defs><style>.cls-1{fill:#2378c3;}.cls-2{fill:none;stroke:#005ea2;stroke-miterlimit:10;}</style></defs><title>dot gov icon</title><path class="cls-1" d="M36.5,20.91v1.36H35.15a0.71,0.71,0,0,1-.73.68H18.23a0.71,0.71,0,0,1-.73-0.68H16.14V20.91l10.18-4.07Zm0,13.57v1.36H16.14V34.48a0.71,0.71,0,0,1,.73-0.68h18.9A0.71,0.71,0,0,1,36.5,34.48ZM21.57,23.62v8.14h1.36V23.62h2.71v8.14H27V23.62h2.71v8.14h1.36V23.62h2.71v8.14h0.63a0.71,0.71,0,0,1,.73.68v0.68H17.5V32.45a0.71,0.71,0,0,1,.73-0.68h0.63V23.62h2.71Z"/><circle class="cls-2" cx="27" cy="27.12" r="26"/></svg>
```
bernardwang, line 1: Could we rename these icons to relate to the usa banner?
i.e. `usa-banner-dot-gov.svg`

bernardwang: and vice versa for the other icons, 
`usa-banner-flag.svg`, 
`usa-banner-https.svg`, 
`usa-banner-lock.svg`.

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -1,40 +1,48 @@
-.c-usa-banner {
+.ds-c-usa-banner {
   background-color: #f0f0f0;
```
bernardwang, line 2: Can we create theme variables for background-color and color?

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -60,37 +68,44 @@
...
+    padding: 0 $spacer-1;
     width: 50%;
   }
 }
 
-.c-usa-banner__icon {
+.ds-c-usa-banner__icon {
+  flex-shrink: 0;
+  margin-right: 0.5rem;
   width: 2.5rem;
```
bernardwang, line 85: I think we can remove this `width` here

packages/design-system-docs/src/scripts/components/UsaBanner.jsx
```diff
@@ -4,23 +4,23 @@ import path from 'path';
 
 const UsaBanner = function (props) {
   const bannerHeader = props.isBannerOpen
```
bernardwang, line 6: This file should be removed altogether, the doc site should just be using the core USABanner moving forward

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -60,37 +68,44 @@
   }
```
bernardwang, line 68: I think we need to add a `padding-right: 2px` for `ds-c-usa-banner__button::after` as well to prevent the icon from cutting off

bernardwang:
@line47 For the SVG react feedback, I put up this PR using that babel plugin I mentioned:
https://github.com/CMSgov/design-system/pull/871

You can cherry-pick [this commit](https://github.com/CMSgov/design-system/pull/871/commits/77b83b666f03ad16954c48c831301e7160b76c12) from that branch to get that functionality in this branch, and move this work forward

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
line47: I think we need to keep that internal div because the `ds-c-usa-banner__content` is using hidden to show and hide the HTML element and If we set that div to use `display:flex` that will take precedent over the hidden. 
See the note here: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden

packages/design-system/src/styles/components/_UsaBanner.scss
line47: If I don't provide a with the images don't show

line47:
@bernardwang should I remove the .jsx Icons in favor os using the plugin to reference the actual SVGs?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
bernardwang: ok that makes sense. maybe we can add a new class instead? we can call it  content wrapper?

bernardwang:
> @bernardwang should I remove the .jsx Icons in favor os using the plugin to reference the actual SVGs?

yes that’s what i would recommend, i did a quick test and it seems to work, but let me know how it goes

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
line47: I got this working but I'm sure it could be improved - thanks for letting me take on this work!

nichia (commented):
Not sure why, but when i click on `Here’s how you know` for both the html and react component, it did not toggle to show the text. I can see the text in the code snippet though so it should have expanded to show the text.

nichia:
Just saw that the react component does work to toggle text once I open a `New tab`

nichia:
Great work, the usaBanner looks great on CMSDS doc site!

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -0,0 +1,43 @@
+<div class="ds-c-usa-banner">
+  <header class="ds-c-usa-banner__header">
+    <img class="ds-c-usa-banner__header-flag" src="/images/us_flag_small.png" alt="U.S. flag" />
```
bernardwang, line 3: This is using the png in the doc site instead of the svg from the core. I think all the images in the HTML example should be updated to use the core svg files

packages/design-system/src/components/UsaBanner/UsaBanner.test.js
```diff
@@ -0,0 +1,8 @@
+import React from 'react';
+import UsaBanner from './UsaBanner';
+import renderer from 'react-test-renderer';
+
+it('renders correctly', () => {
```
bernardwang, line 5: We should also have a test for when the banner is opened

packages/design-system/src/images/usa-banner-flag.svg
```diff
@@ -0,0 +1,9 @@
+<svg width="16" height="11" class="ds-c-usa-banner__header-flag" xmlns="http://www.w3.org/2000/svg">
```
bernardwang, line 1: I dont think we should be applying these classes directly to the source svg file. You should be able to pass in the class when you use the SVG in React

You can look at [react-svg-loader](https://github.com/boopathi/react-svg-loader/tree/master/packages/react-svg-loader) as an example, it works similarly to our svg babel plugin

bernardwang: Another small thing, could you use [svgo](https://github.com/svg/svgo/) to clean up these svg files? svgo is a svg minimizer that will remove unnecessary attributes, and it also has a `pretty` option to format them as well. I believe you can just run it via command line

packages/design-system/src/styles/settings/variables/_color.scss
```diff
@@ -109,3 +109,7 @@ $color-background-dialog: $color-white !default;
 $border-color: $color-gray-lighter !default;
 $border-color-dark: $color-gray-dark !default;
 $border-color-inverse: $color-white !default;
+
+// USA Banner
+$usa-banner-background-color: #f0f0f0;
```
bernardwang, line 114: I think this should be a theme variable, so it should be inside the USA banner scss file.

packages/design-system-docs/src/scripts/components/Header.jsx
```diff
@@ -1,6 +1,6 @@
 import GitHubLinks from './GitHubLinks';
 import React from 'react';
-import UsaBanner from './UsaBanner';
+import { UsaBanner } from '@cmsgov/design-system';
```
nichia, line 3: We can remove the props being passed to the component
        <UsaBanner
          isBannerOpen={this.state.isBannerOpen}
          onToggleBanner={this.handleToggleBanner}
        />

nichia: and remove the state and handToggleBanner from here.

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -0,0 +1,60 @@
+import React, { useState } from 'react';
+import DotGovIcon from '../../images/usa-banner-dot-gov.svg';
+import HttpsIcon from '../../images/usa-banner-https.svg';
+import LockIcon from '../../images/usa-banner-lock.svg';
+import UsaFlagIcon from '../../images/usa-banner-flag.svg';
+
+const UsaBanner = function () {
+  const [isActive, setActive] = useState('false');
```
nichia, line 8: useState doesn't works great but hcgov-design-system doesn't support it yet as it's on older version of react. Can revert to how header.jsx is currently keeping the state using class constructor.

packages/design-system/src/images/usa-banner-flag.svg
line47: I tried to use svgo and it messed up the SVGs even when I'm only using the `--pretty` command

bernardwang: Hm how did it mess them up?

line47: The no longer rendered - I can add a screenshot

bernardwang:
I wonder if instead of merging this branch into master we should merge it into a temporary branch called `usa-banner` or something. This way followup work would be merged into `usa-banner` until we are ready to release the beta.

Off the top of my head, there are at least 2 followups needed, 1 for spanish content and 1 for improving mobile/responsive banner

line47:
@bernardwang I updated this to merge into `feature-usa-banner` branch instead of `master`

packages/design-system/src/images/usa-banner-flag.svg
line47: This is after running `svgo -f packages/design-system/src/images --pretty` which worked and updated all the SVGs

However, it broke the rendering

Screen Shot 2020-11-09 at 3 23 39 PM

Here is the before and after `--pretty` updates which removed the `fill="#000000"` and some dashes in the path element

### Before
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 52 64">
    <path fill="#000000" fill-rule="evenodd" d="M26 0c10.493 0 19 8.507 19 19v9h3a4 4 0 0 1 4 4v28a4 4 0 0 1-4 4H4a4 4 0 0 1-4-4V32a4 4 0 0 1 4-4h3v-9C7 8.507 15.507 0 26 0zm0 8c-5.979 0-10.843 4.77-10.996 10.712L15 19v9h22v-9c0-6.075-4.925-11-11-11z"/>
</svg>
```
### After
```
<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 52 64">
    <path fill-rule="evenodd" d="M26 0c10.493 0 19 8.507 19 19v9h3a4 4 0 014 4v28a4 4 0 01-4 4H4a4 4 0 01-4-4V32a4 4 0 014-4h3v-9C7 8.507 15.507 0 26 0zm0 8c-5.979 0-10.843 4.77-10.996 10.712L15 19v9h22v-9c0-6.075-4.925-11-11-11z"/>
</svg>
```

packages/design-system/src/components/UsaBanner/UsaBanner.e2e.test.js
```diff
@@ -0,0 +1,20 @@
+/* global driver */
+import { ROOT_URL } from '../helpers/e2e/constants';
+
+import assertNoAxeViolations from '../helpers/e2e/assertNoAxeViolations';
+import { getElementByClassName } from '../helpers/e2e';
+
+const rootURL = `${ROOT_URL}/example/components.usabanner/`;
```
nichia, line 7:
```suggestion
const rootURL = `${ROOT_URL}/example/components.usa-banner.react/`;
```

nichia:
Not sure if it's intentional, when I reduce the viewport, the flag on the banner is cut off a little on the right side, making it more square then rectangle

nichia:
Just tried view the expanded banner on IE11 and the alignment is different. Can you see if that happens on your side too.

line47:
Here is my view in IE 11 which looks OK

Screen Shot 2020-11-10 at 12 13 53 PM

Screen Shot 2020-11-10 at 12 14 32 PM

line47:
@bernardwang when you referred to `improving mobile/responsive banner` were you talking about the differences between the open close of our USA banner at small screens and the one seen on V2 of USWDS where they have an `X` for the close button? 
Screen Shot 2020-11-10 at 1 06 46 PM

nichia (approved):
The usa banner works great!

bernardwang:
> @bernardwang when you referred to `improving mobile/responsive banner` were you talking about the differences between the open close of our USA banner at small screens and the one seen on V2 of USWDS where they have an `X` for the close button?
> Screen Shot 2020-11-10 at 1 06 46 PM

Yep thats exactly it, i think that can be a new ticket

bernardwang (approved):
This looks good to me considering spanish and mobile will be handled in the future. great work scott!

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -0,0 +1,103 @@
+<div class="ds-c-usa-banner">
+  <header class="ds-c-usa-banner__header">
+    <svg
+      width="16"
+      height="11"
+      class="ds-c-usa-banner__header-flag"
+      xmlns="http://www.w3.org/2000/svg"
```
bernardwang, line 7: One thing you can do similar to other html examples using svg is reference the svg file directly using `<use href> </use>`. See the button icons html example

bernardwang:
@line47 Also, I know I asked you to handle the banner state within the component in [this comment](https://github.com/CMSgov/design-system/pull/869#discussion_r517522517), I did want to point out that I suggested this with the assumption that users never need to control the banner themselves, and at this point I don't know if thats the case. 

It's a bit tricky because have components that are on both opposites of the spectrum, i.e. HelpDrawer doesnt manage state, but Tooltip does. My guess is that this component might be something in between where most of the time it doesnt need to manage state, but sometimes there might be edge cases. 

All this to say is that how we handle state is subject to change as we iterate on this and get more feedback

PR #873, opened 2020-11-10
Omit Autocomplete-only props from AutocompleteProps type def

pwolfert:
In our TypeScript definitions, I recently changed `AutocompleteProps` in order to include `DownshiftProps`, a type imported from the downshift library. Our `Autocomplete` component [passes all props down to a `Downshift` component](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/Autocomplete/Autocomplete.jsx#L129) except for ones called out in [this destructuring assignment](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/Autocomplete/Autocomplete.jsx#L123). Therefore in order for the TypeScript compiler in my application to be fine with me passing a Downshift prop (`initialInputValue `) to `Autocomplete`, I had to [make `AutocompleteProps` extend `DownshiftProps`](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/types/Autocomplete/Autocomplete.d.ts#L9). I introduced this change in https://github.com/CMSgov/design-system/pull/870.

But it turns out that it was throwing errors in other projects, because we don't actually pass _all_ props from `Autocomplete` down to the `Downshift` component, so `AutocompleteProps` had props that were incompatible with `DownshiftProps`. This fixes that conflict.

### Fixed
- Fixed TypeScript error `'AutocompleteProps' incorrectly extends interface 'DownshiftProps<any>'`


nichia (approved):
Thanks for the PR.
Tested core ds, create-react-app-typescript, Hcgov ds, Mgov ds, Taxtools, coverage-tools-frontend and they look to be working as expected.

bernardwang (approved):
I'm running coverage-tools-frontend with this branch no problem, thanks for putting this up @pwolfert

pwolfert:
No problem 👍 . Any idea why the CI tests are still pending? Feel free to merge it without me

PR #875, opened 2020-11-12
[WNMGDS-697] Update class namespace to utilities

nichia:
The existing grid classes is grouped under the old `ds-l` naming convention. Now that `design-system-layout` is deprecated, we can rename the classes to fit the utilities naming convention.

### Changed
- Update grid classes with namespace of `ds-u` and move to `utilities` folder
- Update all references to new namespace

### Deprecated
- Deprecate grid classes with namespace of `ds-l`


nichia:
As this is a breaking change, I have kept existing grid.scss file with namespace of `ds-l` for now. We can remove this file from `base` folder when a deprecating process is firmed up.

nichia:
I flipped the order of grid classes tickets. This PR is to be worked on after #876

PR #878, opened 2020-11-13
[WNMGDS-693] Allow hiding doc site pages

bernardwang:
### Summary
- Adds `hideSection` flag to hide KSS sections or pages

### How to test
- Add `@hide-section` to a KSS page and ensure the entire page is removed from navigation, and no html files related to that page are generated (including example pages)
- Add `@hide-section` to a KSS subsection and ensure that section is removed
- Add `hideSection` to a markdown file and ensure that section is removed
- Do all the above steps on a Child DS as well


guides/WRITING-DOCUMENTATION.md
```diff
@@ -68,14 +68,15 @@ The description should describe what pattern does or should be used for in plain
...
+| **`@react-props [NAME]`**   | `@react-props Button.jsx`                                   | Displays the React prop documentation for a component. `[NAME]` is the filename of the react component where the Proptypes are defined. |
+| **`@react-example [NAME]`** | `@react-example Button.example.jsx`                         | Displays the example file using this React component. `[NAME]` is the filename of the react example to be displayed.                    |
+| **`@responsive`**           | `@responsive`                                               | Renders breakpoint toggles for the markup example.                                                                                      |
+| **`@status [NAME]`**        | `@status Draft`                                             | Displays a status badge. `[NAME]` must be one of the following values: `Draft`, `Ready`, `Deprecated`.                                  |
+| **`@uswds [URL]`**          | `@uswds https://designsystem.digital.gov/components/alerts` | `[URL]` is the link to the corresponding US Web Design System component.                                                                |
```
nichia, line 79: Under  Writing general documentation section, add this new `hide-section: true` optional property. Also noticed `hide-markup` property is missing from this section too.

nichia (approved):
Just noted extra info to add to doc.
Other then that, the change works great on core and child ds. Great work.

guides/WRITING-DOCUMENTATION.md
```diff
@@ -68,14 +68,15 @@ The description should describe what pattern does or should be used for in plain
...
-| **`@status [NAME]`**        | `@status Draft`                                              | Displays a status badge. `[NAME]` must be one of the following values: `Draft`, `Work in progress`, `Ready`, `Deprecated`.              |
-| **`@uswds [URL]`**          | `@uswds https://designsystem.digital.gov/components/button/` | `[URL]` is the link to the corresponding US Web Design System component.                                                                |
+| flag name                   | example                                                     | flag description                                                                                                                        |
+| --------------------------- | ----------------------------------------------------------- | --------------------------------------------------------------------------------------------------------------------------------------- |
+| **`@hide-section`**         | `@hide-section`                                             | Hides an entire documentation section. If used on a page's top level section, it will hide the entire page                              |
```
line47, line 73: Add note to call out how to hide the entire page. 

```suggestion
| **`@hide-section`**         | `@hide-section`                                             | Hides an entire documentation section. **Note:** If used on a page's top level section, it will hide the entire page                              |
```

line47:
### For CMSDS
When I add the `hide-section` to the first section to hide the page the pages hides but I can still get to the examples if i have the URL for example with the Alert component the page is hidden but I can still get to and view: 

- http://localhost:3000/example/components.alert/
- http://localhost:3000/example/components.alert.variations/


I was only able to hide an entire Markdown page and a not section - but I think the front matter is the only way to apply the `hideSection` in the markdown pages. 

### For Mgov DS
- working as expected in the same ways as above

guides/WRITING-DOCUMENTATION.md
bernardwang: Yeah, so our KSS and Markdown flags are actually the same, and so these two tables arent technically correct. I would prefer not to update the Markdown flag table in this PR, as theres several flags that arent updated across both tables. In the future we will be removing KSS anyway, so we will eventually have a single table of frontmatter values

line47 (approved):
Nice - happy to see this change

bernardwang:
@scott I looked into this and right now all example files are also generated regardless of the flag. I could fix this by rewriting the `generateExamplePages` function, but its a bit involved so i might just leave a comment for now. Does that work?

Issue #877, opened 2020-11-13
Autocomplete component not working in my system


thetif:
**Describe the bug**
I am trying to use the Autocomplete component in my application, but it is not rendering correctly. My repo is https://github.com/CMSgov/eAPD. I tried replacing the whole app with something just rendering the Autocomplete component. Initially it worked, but then when I tried to incorporate it into the app again it didn't and I couldn't get it to work on it's own again.

**Expected behavior**
It should behave like the example on the guide, but instead it is unresponsive.

**Screenshots**
I took screen shots of the rendered component from the guide and the rendered version in my code.

<img width="629" alt="Screen Shot 2020-11-09 at 4 00 39 PM" src="https://user-images.githubusercontent.com/62778/99091385-cf1fe800-259d-11eb-9be1-b428b911685b.png">

![Screen Shot 2020-11-09 at 3 59 08 PM](https://user-images.githubusercontent.com/62778/99091421-dba44080-259d-11eb-8aae-06f0348c1539.png)

**Desktop (please complete the following information):**
 - OS: iOS
 - Browser: chrome and firefox

**Additional context**
We are using node version 10, react version 16.13.1, and design-system version 2.1.1.

nichia:
Tracked here: https://jira.cms.gov/browse/WNMGDS-700

PR #881, opened 2020-11-19
[WNMGDS-700] Autocomplete component not rendering

nichia:
Autocomplete component is not rendering on some applications like eADP.

Found that the application does not support `child.type` with the resulting Autocomplete data not displaying. 
 
### Changed
-  Add a fallback conditional `child.type.displayName and child.type.name` checks for Autocomplete component
- Table component has similar `child.type` logic as the Autocomplete component so it does not work on eADP either. Update Table component with similar changes.

### Fixed
This fixes https://github.com/CMSgov/design-system/issues/877

### How to test
- Check that `Autocomplete` and scrollable feature of `Table` component are working
- On eADP, the `Autocomplete` component and `Table` scrollable feature do not work correctly
- Clone the [eADP repo](https://github.com/CMSgov/eAPD) to local system and setup a test environment. Follow the instruction on [setup readme.md](https://github.com/CMSgov/eAPD/blob/master/README.md) or [setting up manually](https://github.com/CMSgov/eAPD/wiki/Development-Environment#manually)
- Select one of the eADP component (eg. Root.js) for testing:
  - import the `Autocomplete` components (eg. copy the autocomplete examples from CMSDS). Check that `Autocomplete` works correctly
  - import the `Table` components (eg. copy the table examples from CMSDS). Check that `Table` works correctly


bernardwang:
@nichia do you know when `child.type` is not supported? Is it a particular React version? I ask because ideally we can test this fix without using eAPD's repo

nichia:
I initially thought it was to do with React version, but I was mistaken as I don't have this issue when I tested using the React version that eADP has on a simple Create-react-app. For Tab component, there's a similar usage of child.type with a comment that says `Preact` has issue with it.

packages/design-system/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -29,7 +29,10 @@ import uniqueId from 'lodash.uniqueid';
  * @return {Boolean} Is this a TextField component?
  */
 function isTextField(child) {
-  return child != null && child.type === TextField;
+  // Some applications does not support child.type
+  return (
+    child && (child.type === TextField || (child.props && child.props.label && child.props.name))
```
bernardwang, line 34: Have you tried looking online? After some initial research it looks like there are issues with HOC changing the name or display name
https://stackoverflow.com/questions/28218814/how-to-access-the-class-name-of-a-react-object-in-this-props-children

bernardwang: https://reactjs.org/docs/higher-order-components.html#convention-wrap-the-display-name-for-easy-debugging

bernardwang: https://stackoverflow.com/questions/43913054/reactjs-react-children-foreach-can-i-get-the-child-component-name

This one seems the most promising, there's a comment that references `babel-plugin-add-react-displayname`

nichia: Added `babel-plugin-add-react-displayname` so we can just reference 'child.type.displayName'

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -83,7 +83,7 @@ export class Table extends React.PureComponent {
             _scrollableNotice: this.props.scrollableNotice,
           });
         }
-      } else if (this.props.stackable) {
+      } else if (this.props.stackable && child.props) {
```
bernardwang, line 86: Why are we adding `child.props` here?

packages/design-system/src/components/Autocomplete/Autocomplete.jsx
```diff
@@ -29,7 +28,10 @@ import uniqueId from 'lodash.uniqueid';
  * @return {Boolean} Is this a TextField component?
  */
 function isTextField(child) {
-  return child != null && child.type === TextField;
+  return (
+    // Some application does not support child.type
+    child && child.type && child.type.displayName === 'TextField'
```
bernardwang, line 33: We want to also use `type.name` as a backup, and I recommend using lodash to simplify this logic
```
import { get, uniqueId } from 'lodash';
return get(child, 'type.name') || get(child, 'type.displayName')
```

packages/design-system/src/components/Table/Table.jsx
nichia: We want to avoid extending to text/strings and only extend if child is an object.

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -43,7 +45,7 @@ export class Table extends React.PureComponent {
       if (
         props.scrollable &&
         Array.isArray(props.children) &&
-        !props.children.some((child) => child.type === TableCaption)
+        !props.children.some((child) => child.type.displayName === 'TableCaption')
```
bernardwang, line 48: Can we reuse `isTableCaption` here?

nichia: Yes, using the default prop `isTableCaption` will reduce the bundle size.

line47 (approved):
This is working for me in the eAPD app, however probably need another technical pass

.eslintrc.js
```diff
@@ -1,5 +1,6 @@
 module.exports = {
   extends: '@cmsgov/eslint-config-design-system',
+  parser: 'babel-eslint',
```
bernardwang, line 3: @nichia what is this for?

packages/design-system/src/components/TextField/TextField.jsx
```diff
@@ -8,6 +8,9 @@ import uniqueId from 'lodash.uniqueid';
 export { unmaskValue } from './Mask';
 
 export class TextField extends React.PureComponent {
+  // Set component name to make child.type.displayName available to other components (eg. Autocomplete)
+  static displayName = 'TextField';
```
bernardwang, line 12: why is this one static?

.eslintrc.js
nichia: The [eslint parsing error for `static` syntax](https://stackoverflow.com/questions/36001552/eslint-parsing-error-unexpected-token) and added `babel-eslint` to fix that.

Also `static` requires  `@babel/plugin-proposal-class-properties`

packages/design-system/src/components/TextField/TextField.jsx
nichia: Using `static` to set the `displayName` property for the class.
Ah, just realized I should try declaring class static property after the class definition, like what I did for the functional component. Then wont need the `@babel/plugin-proposal-class-properties` and `babel-eslint`

nichia: Will give that a try

nichia: Have updated to declare class property `displayName` after the class definition and removed the `babel` dependencies

bernardwang (approved):
This LGTM! I prefer for the eAPD engineer to confirm the fix though before merging

line47:
Can we add the eAPD engineer to the list for reviews?

PR #888, opened 2020-11-27
[WNMGDS-698] Add Spanish translation prop to US banner component

nichia:
### Added
- Added `locale` prop to display USA Banner in Spanish

### How To test
- Then visit the USA Banner component page and ensure the correct text is printed when the Spanish translation prop (`locale = "es"`) is set


nichia:
Follow similar method as how the Month Picker passes a `locale` prop for month name localization. However, translation for USA Banner is different as we already have the exact text for the 'Spanish' version, hence I've set the `localeSpanish` prop as a boolean flag instead.

line47:
Instead of merging to master we should instead merge to the[ feature-usa-banner](https://github.com/CMSgov/design-system/tree/feature-usa-banner) branch

packages/design-system-docs/src/pages/components/UsaBanner/_UsaBanner.docs.scss
```diff
@@ -0,0 +1,43 @@
...
+Markup: usabanner.example.html
+
+Style guide: components.usa-banner
+*/
+
+/*
+`<UsaBanner>`
+
+@react-example UsaBanner.example.jsx
+
```
line47, line 19:
```suggestion
@react-props UsaBanner.jsx
```

line47:
I'm not seeing the SVG images in the plain HTML/CSS example  
Screen Shot 2020-11-30 at 11 30 33 AM

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -16,6 +17,60 @@ class UsaBanner extends React.PureComponent {
   }
 
   render() {
+    const tld = `.gov`;
```
bernardwang, line 20: Translation content should be inside a .json file similar to in `hcgoc-design-system`. See this as an example: https://github.com/CMSgov/hcgov-design-system/tree/master/src/locale

bernardwang: I would also consider using a translation framework like [react-i18next](https://react.i18next.com/guides/quick-start). While this is the only component that needs internationalization in the DS, a framework would standardize how we do translations for the future.

line47: I'm Wondering if we should hold off on using a translation framework until we have a more established need for it beyond this usa banner?

nichia: Add translation framework react-i18next to this branch [WNMGDS-698/add-USA-banner-spanish-prop-i18n](https://github.com/CMSgov/design-system/tree/WNMGDS-698/add-USA-banner-spanish-prop-i18n). While this works, it make be better to not implement this yet as the latest version of react-i18next uses hooks/context and we're not looking to update yet. For the test branch, I'm used the older version of react-i18next.

nichia: @bernardwang @line47 - we only have Usa Banner that needs translation, I prefer to keep it simple for now by using the .json files. We can add the framework when we have more components that requires translation or when we've upgraded to later React version.

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -6,7 +6,16 @@
       class="ds-c-usa-banner__header-flag"
       xmlns="http://www.w3.org/2000/svg"
     >
-      <use xlink:href="{{root}}/dist/images/usa-banner-flag.svg"></use>
+      <g fill="none" fill-rule="evenodd">
```
bernardwang, line 9: Is there a reason why we cant link to the banner icon?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -16,6 +19,8 @@ class UsaBanner extends React.PureComponent {
   }
 
   render() {
+    const Translation = this.props.locale === 'es' ? SpanishTranslations : EnglishTranslations;
```
bernardwang, line 22: What do you think about this instead:

```
const t = this.props.locale === 'es' ? SpanishTranslations.usaBanner : EnglishTranslations.usaBanner;
```
```

File: packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html (hunk 3 above)
Comment (nichia) [reply packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html:9]: The icons won't show up - https://github.com/CMSgov/design-system/pull/888#issuecomment-735895196

File: packages/design-system/src/components/UsaBanner/UsaBanner.jsx [hunk 5]
```diff
@@ -41,21 +46,27 @@ class UsaBanner extends React.PureComponent {
             <div className="ds-c-usa-banner__guidance">
               <DotGovIcon className="ds-c-usa-banner__icon" />
               <p className="ds-c-usa-banner__media-body">
-                <strong>Official websites use .gov</strong>
-                <br />A <strong>.gov</strong> website belongs to an official government organization
-                in the United States.
+                <strong>{Translation.usaBanner.domainHeaderText}</strong>
```
bernardwang, line 49: This is one of the nice things about i18next
https://react.i18next.com/guides/the-drawbacks-of-other-i18n-solutions#can-you-translate-combined-jsx-nodes-in-one-sentence

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
bernardwang: @nichia Have you tried looking into why the icons arent showing up?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
bernardwang: I'd prefer to use `react-i18next` as it contains a feature to prevent splitting translations (as referenced [here](https://github.com/CMSgov/design-system/pull/888#discussion_r537647678)). It's also a far more standardized approach to translation that I feel we should eventually adopt anyway.

 However, in the interest of finishing this PR, @nichia we can go with this solution. In the meantime, we should ask engineers what they prefer from the UsaBanner and do some more research on how internationalized components could work in an application.

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
nichia: I missed the cause, will check into it

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
nichia: Will add another ticket to add translation framework.

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
nichia: Haven't found the cause/fix yet. Adding another ticket for this.

nichia:
@bernardwang @line47 - ready for another review

line47 (approved):
LGTM! Nice work!

packages/design-system/src/images/usa-banner-https.svg
```diff
@@ -1,4 +1,4 @@
-<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 54 54">
-    <path class="cls-1" fill="#719f2a" d="M34.72,34.84a1.29,1.29,0,0,1-1.29,1.29H20.57a1.29,1.29,0,0,1-1.29-1.29V27.12a1.29,1.29,0,0,1,1.29-1.29H21V23.26a6,6,0,0,1,12,0v2.57h0.43a1.29,1.29,0,0,1,1.29,1.29v7.72Zm-4.29-9V23.26a3.43,3.43,0,0,0-6.86,0v2.57h6.86Z"/>
-    <circle fill="none" stroke="#538200" stroke-miterlimit="10" class="cls-2" cx="27" cy="27.12" r="26"/>
+<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 54 54">
+    <path id="svg" fill="#719f2a" d="M34.72,34.84a1.29,1.29,0,0,1-1.29,1.29H20.57a1.29,1.29,0,0,1-1.29-1.29V27.12a1.29,1.29,0,0,1,1.29-1.29H21V23.26a6,6,0,0,1,12,0v2.57h0.43a1.29,1.29,0,0,1,1.29,1.29v7.72Zm-4.29-9V23.26a3.43,3.43,0,0,0-6.86,0v2.57h6.86Z"/>
```
bernardwang, line 2: Theres a way to combine these two SVG elements into a single one by wrapping the two in a `<g>` element. This way we only need one id

bernardwang:
Can you update the HTML examples to match the React one too. Also, can you make the example banners open on default?

nichia:
@bernardwang @line47 - ready for another check

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -1,3 +1,34 @@
+<!-- prettier-ignore -->
+<% [
```
bernardwang, line 2: Why do we need to put the json inside the HTML example? It would be simpler to just use the text in the example

bernardwang: @nichia

nichia: I went the path of using ejs to render the same html once for English and then Spanish to make the html example shorter. Do you think it'll be clearer to just write the full same html code twice with it's respective language?

nichia: It does look clearer. Will update

PR #889, opened 2020-11-30
[No Ticket] - Fix HelpDrawerToggle React Warning

erinrep:
### Fixed
There is a React warning happening for the HelpDrawerToggle:
```
Warning: React does not recognize the `helpDrawerOpen` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `helpdraweropen` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
          in button (created by Button)
          in Button (created by HelpDrawerToggle)
          in HelpDrawerToggle (at HelpDrawerWrapper.tsx:69)
```
The `helpDrawerOpen` prop was being applied to the Button component which it shouldn't be, so I added it to line 17 to pull it out from the others variable before others is applied to the Button

### DIFF
--- a/packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
+++ b/packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
@@ -14,7 +14,7 @@ export class HelpDrawerToggle extends React.PureComponent {
   }
 
   render() {
-    const { className, children, inline, showDrawer, ...others } = this.props;
+    const { className, children, inline, showDrawer, helpDrawerOpen, ...others } = this.props;
     const classes = classNames(
       'ds-c-help-drawer__toggle',
       inline && 'ds-u-display--inline',
--- a/packages/design-system/src/components/HelpDrawer/__snapshots__/HelpDrawerToggle.test.jsx.snap
+++ b/packages/design-system/src/components/HelpDrawer/__snapshots__/HelpDrawerToggle.test.jsx.snap
@@ -4,7 +4,6 @@ exports[`HelpDrawerToggle renders a button 1`] = `
 <Button
   className="ds-c-help-drawer__toggle"
   component="button"
-  helpDrawerOpen={false}
   inputRef={[Function]}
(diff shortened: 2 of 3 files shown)


nichia (approved):
LGTM. Thanks

erinrep:
@nichia what's your process for merging after an approval? Should I wait until all reviewers have approved or go ahead and merge?

line47 (approved):

nichia:
@erinrep We typically wait for 2 reviews. But for a small change, we do merge after one reviewer has approved. This PR looks ready for merging.

packages/design-system/src/components/HelpDrawer/HelpDrawerToggle.jsx
```diff
@@ -14,7 +14,7 @@ export class HelpDrawerToggle extends React.PureComponent {
   }
 
   render() {
-    const { className, children, inline, showDrawer, ...others } = this.props;
```
bernardwang, line 17: This is something we forgot to do in the past, but we should also update `HelpDrawerToggle.d.ts` to use the React.HTMLProps for the button element considering we allow props to be passed down.

Something like this: 
```
export default class HelpDrawerToggle extends React.Component<
React.HTMLProps<HTMLButtonElement> & HelpDrawerToggleProps,
  any
> {
  render(): JSX.Element;
}
```

bernardwang: @nichia or I can also handle this if you are busy @erinrep, just let us know!

erinrep: @bernardwang that would be great, I've got a lot on my plate today. Thanks!

bernardwang:
@line47 @nichia This is ready to merge after testing the new typescript changes

line47 (approved):
Looks good! Thanks @bernardwang

nichia (approved):
Thanks for picking that up. LGTM

PR #894, opened 2020-12-07
[NO_TICKET] Add react-i18next to USA banner

nichia:
### Added
- Add React-i18next to USA banner for Internalization 

### Closed
After further exploration, we decided to hold off from adding react-i18next to CMSDS. Banner is the only component where we would handle `en/es` locales, and so it does not make sense to add that dependency and potentially add a bundle size cost for downstream apps that may not use the same library or the same version of the library


packages/design-system-docs/src/pages/components/UsaBanner/UsaBanner.example.jsx
```diff
@@ -2,4 +2,12 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 import { UsaBanner } from '@cmsgov/design-system';
 
-ReactDOM.render(<UsaBanner />, document.getElementById('js-example'));
+ReactDOM.render(
+  <div>
+    <h6 className="preview__label">.gov Domains</h6>
```
bernardwang, line 7: Maybe these labels can just be generalized to be 
`English banner` and `Spanish banner`

packages/design-system/src/components/UsaBanner/I18nUsaBanner.jsx
```diff
@@ -0,0 +1,33 @@
...
+ * A container component responsible for passing an instance
+ * of i18next to all child components using react-i18next's
+ * `translate` HOC. Note that we use I18nextProvider in order
+ * to avoid conflicts with other apps using react-i18next.
+ * See https://github.com/i18next/react-i18next/issues/382 for
+ * more context on why we need to do it this way.
+ */
+const I18nUsaBanner = function (props) {
+  return (
+    <I18nextProvider i18n={i18n} initialLanguage={props.initialLanguage}>
```
bernardwang, line 18: Instead of making a separate wrapper file, can we just use `I18nextProvider` inside `UsaBanner`?

nichia: That's a good suggestion, will look into it.

nichia:
After further exploration, we decided to hold off from adding react-i18next to CMSDS. Banner is the only component where we would handle `en/es` locales, and so it does not make sense to add that dependency and potentially add a bundle size cost for downstream apps that may not use the same library or the same version of the library

PR #897, opened 2020-12-08
[WNMGDS-712]: Change Alert component default heading to h2

### LINKED ISSUE #880: Change default heading level in Alert component from H3 to H2
**Describe the bug**
The default heading in the Alert component is a H3. While technically, there's nothing wrong with this, I see in a lot of accessibility reviews where the heading level isn't changed as it should be to account for a better hierarchy. This results in situations where the hierarchy goes from H1 to H3. In most situations, a H2 would suffice as the default.

**To Reproduce**
1. Go to `https://design.cms.gov/example/components.alert.variations/`
2. View the heading in the Alert component

**Expected behavior**
A better default heading value that results in less heading hierarchy issues.

davidakennedy:
## Changed
- Change Alert component default heading to h2

Fixes #880

## How to test
- Go to the alerts page and view the source to see that an `<h2>` is used for the alert header when a specific header is not specified


nichia:
Looks good. Can we change the html.examples to have component heading to `h2` as well.

line47:
I went ahead and make those updates @nichia

nichia (approved):

bernardwang:
I responded in the original GH issue here https://github.com/CMSgov/design-system/issues/880#issuecomment-741891792

bernardwang:
Since this would be a minor change for existing Alert components, I think we should hold off on this until we are able to [confirm the impact](https://github.com/CMSgov/design-system/issues/880#issuecomment-743373226)

line47:
I'm thinking we might be able to confirm this with our CMS POs maybe?

line47:
Wondering if we could create a breaking changes branch to merge this and other work into to release at a later date?
@bernardwang @nichia ?

line47 (approved):

nichia (approved):
Just need to add `6` to the [alert typescript def file](https://github.com/CMSgov/design-system/blob/46cbd9eabc73491c90519d0bc25ce122ff220261/packages/design-system/src/types/Alert/Alert.d.ts#L3) : `export type AlertHeadingLevel = '1' | '2' | '3' | '4' | '5' | '6';`
Other then that, it's all good.

PR #899, opened 2020-12-09
[WNMGDS-707] Create shared form container component

bernardwang:
### Background
This PR came out of the [HCgov Error Enhancement doc](https://docs.google.com/document/d/1LzwKkEEyL7hz415d3wJDL830PEOVOat7UAWq9Lygjoc/edit?ts=5fbbdef9#), which describes making our form components more modular and consistent to make it easier to change the inline error placement. While the error placement work is paused until further testing in 2021, these changes can be worked on immediately.

### Summary
- This PR adds a shared `FieldContainer` component that is used across all form components. 
   - `FieldContainer` handles functionality for creating unique ids for label and input, for rendering the `FormLabel` (and consequently the inline error), and logic used in some shared `props` like `focusTrigger` and `inputRef`.
   - It formalizes a set of standard form component props including `className`, `labelClassName`, `fieldClassName`, `id`, `labelId`, `name`, etc. While these props were already in use across form components, `FieldContainer` is now the standard interface in which these props will be passed into. 
   - `FieldContainer` accepts "field input" components as `children`. It will pass props to the field inputs via [render props](https://reactjs.org/docs/render-props.html).
- Form components have been refactored to use `FieldContainer`, and new "field input" components have been defined.
   - `Dropdown` uses `Select` 
   - `TextField` uses `TextInput`
   - `DateField` uses `DateInput`
   - `ChoiceList` already used `Choice` components, so no new component had to be defined.
   - `MonthPicker` also uses `Choice`

### How to test
- Test existing functionality for each form component
- Test Typescript in CRA typescript example or another application


packages/design-system-docs/src/pages/components/TextField/TextField.example.jsx
```diff
@@ -15,7 +15,7 @@ ReactDOM.render(
       label="Password field"
       name="password_example"
       type="password"
-      value="Example value"
+      defaultValue="Example value"
```
bernardwang, line 18: The old password TextField example doesnt let you edit the text, this fixes that

packages/design-system-docs/src/pages/components/DateField/datefield.example.html
```diff
@@ -101,7 +101,7 @@
             type="text"
             inputmode="numeric"
             pattern="[0-9]*"
-            id="year"
+            id="year-2"
```
bernardwang, line 104: Fixes a console warning about duplicate `id`s

packages/design-system/src/components/FieldContainer/FieldContainer.jsx
```diff
@@ -0,0 +1,165 @@
...
+
+    const ComponentType = this.props.component;
+    const classes =
+      ComponentType === 'fieldset'
+        ? classNames(this.props.className, 'ds-c-fieldset')
+        : this.props.className;
+
+    // Props that have been transformed or renamed by <FieldContainer>
+    const transformedFieldProps = {
+      className: this.props.fieldClassName,
```
bernardwang, line 66: Because our existing form components `props` along to different components, our we have prefixed props. I.e. `TextField` accepts `className`, `labelClassName` and `fieldClassName`, which are all applied to the root `div` element, the `label` element, and the `input` element respectively. When creating a `TextInput` element, it would be ideal to use a `className` prop. However, this introduces an issue where `FieldContainer` has to map `fieldClassName` to `className` when it is passed to the field input.

bernardwang: The drawback is that you would have to understand how `FieldContainer` renames these props. Plus, most field input components like `TextInput` and `Select` will be private anyway, so maybe the consistency of the prop naming convention isnt as important

packages/design-system/src/components/TextField/TextField.jsx
```diff
@@ -22,110 +20,22 @@ export class TextField extends React.PureComponent {
...
-          requirementLabel={requirementLabel}
-          inversed={inversed}
-        >
-          {label}
-        </FormLabel>
-        {this.renderFieldAndMask(field, mask)}
-      </div>
+      <FieldContainer {...containerProps} component="div" labelComponent="label">
+        {({ className, errorMessage, id, inversed, name, setRef }) => (
+          <TextInput {...{ ...inputProps, className, errorMessage, id, inversed, name, setRef }} />
```
bernardwang, line 36: @pwolfert Another alternative is for `FieldContainer` to only pass along transformed props like `id`, `labelId` and `setRef`. The other shared props still have to be handled in `TextField`, but this approach could be more explicit.

```
const containerProps = pick(this.props, containerPropList);
const inputProps = omit(this.props, omit(containerPropList, ['name', 'inversed', 'errorMessage']));

return (
    <FieldContainer {...containerProps}>
         {({  id, setRef }) => (
              <TextInput {...{ ...inputProps, id, setRef }} />
         )}
      </FieldContainer>
);
```

bernardwang: I'm actually starting to like this approach a bit ^^
Especially if we remove the `fieldClassName` mapping logic from `FieldContainer`, this approach makes `FieldContainer` a bit simpler

pwolfert: Is `containerPropList` equivalent to `Object.keys(FieldContainer.propTypes)`?

pwolfert: Oh, one thing this pattern is not taking care of, though, is the `fieldClassNames` prop and any other ones you were transforming before

pwolfert: I realize now that when I proposed
```js
<TextInput 
  {...pick(this.props, Object.keys(TextInput.propTypes))}
```
that doesn't account for the extra arbitrary props you can pass to the input like aria labels

pwolfert: Even though the following with `errorMessage`, `inversed`, and `name` is more verbose than having it in an array, it's only a few special cases, and they get grouped visually with `className={this.props.fieldClassName}` which is also a special case
```js
const inputOnlyProps = omit(this.props, Object.keys(FieldContainer.propTypes));

return (
  <FieldContainer 
    {...this.props}
    className={containerClassName}
    component="div"
    labelComponent="label"
  >
    {({ id, setRef }) => (
      <TextInput 
        {...inputOnlyProps}
        {...{id, setRef}}
        className={this.props.fieldClassName}
        errorMessage={this.props.errorMessage}
        inversed={this.props.inversed}
        name={this.props.name}
      />
    )}
  </FieldContainer>
);
```
It may not be so bad to have special cases if it's really clear what they are, and I feel like it's pretty easy to read and see them when they're inlined in the JSX. You're welcome to disagree, of course. The difference I see here is that there is only one calculated `inputOnlyProps` at the top, and the rest is from explicit JSX prop assignments below.

bernardwang: Yeah I agree, its only a few special cases. I also realized I can remove the `name` prop from `FieldContainer`, it was only used in generating ids anyway. 

I also realized, passing in  `this.props` into `FieldContainer` might not work with Typescript. I think I'll need to add a `pick(this.props, Object.keys(FieldContainer.propTypes));` to prevent undefined props errors. Still, since we can just use `FieldContainer.propTypes`, I feel this isn't so bad.

pwolfert: Oh right, I was thinking arbitrary undefined props were fine since we did it on other components, but those "arbitrary" props are actually being inherited from `JSX.Element`, so they aren't actually arbitrary.

bernardwang: Ah I forgot, this wont work either, because `fieldClassName` is still being passed to `TextInput` via `inputOnlyProps`.
```
    const containerProps = pick(this.props, Object.keys(FieldContainer.propTypes));
    const inputOnlyProps = omit(this.props, Object.keys(FieldContainer.propTypes));
    ...
   <TextInput
         {...inputOnlyProps}
         {...{ id, setRef }}
         className={this.props.fieldClassName}
```

I could do this:
```
const inputOnlyProps = omit(this.props, Object.keys(FieldContainer.propTypes).concat(['fieldClassName]);
```
But I'm thinking the extra complexity from prop renaming probably isnt worth it, especially since they are private components anyway. I'm going to just rename them to always be called `fieldClassName`

pwolfert: I like the idea of just changing it to `fieldClassName` 👍

nichia:
`DateField` console log is showing React warning for the `fieldClassName` prop. Does it need to be remove from the list of de-structured props  ```react-dom.development.js:506 Warning: React does not recognize the `fieldClassName` prop on a DOM element.```

bernardwang:
@nichia That should be fixed a few commits ago, and I dont see it locally. Let me know if its still an issue for you

packages/design-system/src/components/FieldContainer/FieldContainer.jsx
```diff
@@ -128,6 +124,10 @@ FieldContainer.propTypes = {
    * Text showing the requirement ("Required", "Optional", etc.). See [Required and Optional Fields]({{root}}/guidelines/forms/#required-and-optional-fields).
    */
   requirementLabel: PropTypes.node,
+  /**
+   * A function that returns a field input element to accept render props
+   */
+  render: PropTypes.func.isRequired,
```
bernardwang, line 130: I opted for using the `render` prop over `children` because <Dropdown> is a form component that also needs to use the `children` prop. By renaming to `render`, we can avoid handling [this exception in `Dropdown`](https://github.com/CMSgov/design-system/pull/899/commits/21e168ff8f350fd1964eec01e5190f3d2f14531e#diff-d2fde628dd9ee937d45834fa7105cdc0d7a00c0951aacd4b1165d4d01ea3b0ffL31)

packages/design-system/src/components/FieldContainer/FieldContainer.jsx
```diff
@@ -54,10 +54,6 @@ export class FieldContainer extends React.PureComponent {
       setRef: this.setFieldRef,
     };
 
-    // Render children with render props or normally
-    // We allow normal rendering to simplify unit tests in ChoiceField
-    const render = typeof children === 'function' ? children(fieldInputProps) : children;
```
bernardwang: Originally I allowed both normal ReactNode and functions for the `children` to avoid breaking tests in `ChoiceList` and `MonthPicker`, but only allowing function is a better solution overall

nichia (commented):
To not negate the advantage of using `React.PureComponent`, the [Caveats of react `render props`](https://reactjs.org/docs/render-props.html#caveats) mentions a work around which is to define the `render` prop as an instance method. I'm not sure how much of a performance improvement we'll have in keeping the React.PureComponent vs React.Component though.

For example, in the case of `TextField.jsx` - it may be good to render `TextInput` prop as an instance method if we'll gain performance improvement for the component.

```js
export class TextField extends React.PureComponent {
  constructor(props) {
    super(props);
    this.containerProps = pick(this.props, Object.keys(FieldContainer.propTypes));
    this.inputOnlyProps = omit(this.props, Object.keys(FieldContainer.propTypes));
    this.renderTheInputText = this.renderTheInputText.bind(this);

      if (props.type === 'number') {
        console.warn(
          `Please use the 'numeric' prop instead of 'type="number"' unless your user research suggests otherwise.`
        );
      }
    }
  }

  // Defined as an instance method, `this.renderTheInputText` always
  // refers to *same* function when we use it in render
  renderTheInputText({id, setRef} ) {
    return <TextInput
          {...this.inputOnlyProps}
          {...{ id, setRef }}
          errorMessage={this.props.errorMessage}
          inversed={this.props.inversed}
        />;
  }

  render() {

    // Add clearfix class
    const containerClassName = classNames(
      'ds-u-clearfix', // fixes issue where the label's margin is collapsed
      this.props.className
    );

    return (
      <FieldContainer
        {...this.containerProps}
        className={containerClassName}
        component="div"
        labelComponent="label"
        render={this.renderTheInputText}
      />
    );
  }
}
```

bernardwang:
@nichia with the latest commit the caveat should be handled for all the form elements except MonthPicker. For that component moving the render into an instance method broke the functionality of the component. We can iterate on that one and try to optimize it in the future

nichia (commented):
Taxtools `checkbox` and `dropdown` does not recognize the checking/selection. Can't figure out why it doesn't work now that render prop (renderFieldInput) to changed to instance method (it worked with the initial render prop).
Also, on core and child ds, the autocomplete doesn't recognize the selection either.

bernardwang:
@nichia Hm, I reverted the change and just updated `FieldContainer` to extend from `React.Component` for now.

nichia (approved):
Test on the form container on core and child ds and taxtools, windowshop and coverage-tools-frontend and all works as expected. 
Great work!

PR #904, opened 2020-12-16
[WNMGDS-718] Add responsive styling to USA banner component

nichia:
### Changed
- Add responsive feature to USA banner component on small viewport.
Screen Shot 2020-12-17 at 9 26 20 AM

### How To test
- Then visit the USA Banner component page and ensure the USA banner displays correctly on resizing of the viewport

### DIFF
--- a/packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
+++ b/packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
@@ -15,13 +15,13 @@
       <p class="ds-c-usa-banner__header-text">
         <span>An official website of the United States government</span>
         <span class="ds-c-usa-banner__header-action" aria-hidden="true">Here’s how you know</span>
-        <button class="ds-c-usa-banner__button" aria-expanded="false" aria-controls="gov-banner">
+        <button class="ds-c-usa-banner__button" aria-expanded="true" aria-controls="gov-banner">
           <span class="ds-c-usa-banner__button-text">Here’s how you know</span>
         </button>
       </p>
     </div>
   </header>
-  <div class="ds-c-usa-banner__content" id="gov-banner" hidden="">
+  <div class="ds-c-usa-banner__content" id="gov-banner">
     <div
       class="ds-u-display--flex ds-u-flex-direction--column ds-u-sm-flex-direction--row ds-u-flex-wrap--nowrap"
     >
@@ -84,13 +84,13 @@
         <span class="ds-c-usa-banner__header-action" aria-hidden="true"
           >Así es como usted puede verificarlo</span
         >
-        <button class="ds-c-usa-banner__button" aria-expanded="false" aria-controls="gov-banner">
+        <button class="ds-c-usa-banner__button" aria-expanded="true" aria-controls="gov-banner">
           <span class="ds-c-usa-banner__button-text">Así es como usted puede verificarlo</span>
         </button>
       </p>
     </div>
   </header>
-  <div class="ds-c-usa-banner__content" id="gov-banner" hidden="">
+  <div class="ds-c-usa-banner__content" id="gov-banner">
     <div
       class="ds-u-display--flex ds-u-flex-direction--column ds-u-sm-flex-direction--row ds-u-flex-wrap--nowrap"
     >
--- a/packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
+++ b/packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
@@ -15,7 +15,7 @@
       <p class="ds-c-usa-banner__header-text">
         <span>An official website of the United States government</span>
         <span class="ds-c-usa-banner__header-action" aria-hidden="true">Here’s how you know</span>
-        <button class="ds-c-usa-banner__button" aria-expanded="true" aria-controls="gov-banner">
+        <button class="ds-c-usa-banner__button" aria-expanded="false" aria-controls="gov-banner">
           <span class="ds-c-usa-banner__button-text">Here’s how you know</span>
         </button>
       </p>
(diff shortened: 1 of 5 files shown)


packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -68,11 +63,136 @@ $usa-banner-text-color: #1b1b1b;
...
+.ds-c-usa-banner__button {
+  background-color: transparent;
+  border: 0;
+  border-radius: 0;
+  bottom: 0;
+  box-shadow: none;
+  color: #005ea2;
+  display: block;
+  font-size: 0.8rem;
+  -moz-osx-font-smoothing: auto;
```
bernardwang, line 98: What is this from? Is this copied from the USWDS CSS? If so, it would be better to reference the SCSS

bernardwang: This SCSS contains alot of random values like `#005ea2`. We should be using variables for commonly used variables so they are easier to understand and modify.

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -23,24 +23,31 @@ export class UsaBanner extends React.PureComponent {
...
+                aria-expanded={this.state.isBannerOpen}
+                aria-controls="gov-banner"
+              >
+                <span className="ds-c-usa-banner__button-text">{t.bannerActionText}</span>
+              </button>
+            </p>
+          </div>
         </header>
         <div className="ds-c-usa-banner__content" id="gov-banner" hidden={!this.state.isBannerOpen}>
           <div className="ds-u-display--flex ds-u-flex-direction--column ds-u-sm-flex-direction--row ds-u-flex-wrap--nowrap">
```
bernardwang, line 53: Is it easier to use a grid here? I know this is the existing code but I'm guessing it might be more succinct

packages/design-system/src/styles/components/_UsaBanner.scss
nichia: Thanks for for picking this up. Will refactor to use variables to make it more understandable and customizable.

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
nichia: That's an idea, will check it out

nichia: Have tidied the scss. I'm making another branch to refactor the banner to grid-layout

bernardwang: @nichia why do you need another branch to do this? wont it be a simple replacement of some classes?

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -1,13 +1,15 @@
 $usa-banner-background-color: #f0f0f0;
 $usa-banner-text-color: #1b1b1b;
+$usa-banner-close-image-background-color: #e6e6e6;
+$usa-banner-header-action-color: #005ea2;
 
 .ds-c-usa-banner {
   background-color: $usa-banner-background-color;
   color: $usa-banner-text-color;
   font-family: 'Source Sans Pro Web', 'Helvetica Neue', Helvetica, Roboto, Arial, sans-serif;
```
bernardwang, line 9: This isnt a font family we currently provide right? Again, this code is coming from compiled CSS. We should be defining and using SCSS variables and working with the CMSDS CSS architecture

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -17,10 +19,16 @@ $usa-banner-text-color: #1b1b1b;
   font-size: 0.79853rem;
   font-weight: 400;
```
bernardwang, line 20: Same here, instead of directly copying compiled values we should be trying to build our own classes using the USWDS as a reference. These styles are from the USWDS typescale and type system, they shouldnt be just copied and used like this.

bernardwang: This is also the only place in the DS that uses rem. We eventually want to use rem anyway, but if they are hardcoded here as compiled values, and if we arent using CMSDS variables, then theres no way to enforce consistency between this component and other styles.

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
line47: I was thinking we would want to keep the markup as close to if not exactly like the USWDS Banner so we can reference their styles and get any future updates in an easier way?

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -68,11 +65,128 @@ $usa-banner-text-color: #1b1b1b;
...
+      height: 3rem;
+      position: absolute;
+      right: 0;
+      top: 0;
+      width: 3rem;
+    }
+    // stylelint-enable scss/media-feature-value-dollar-variable
+
+    &[aria-expanded='true']::after {
+      background: url('#{$image-path}/close-alt.svg') no-repeat center/1rem 1rem;
```
bernardwang, line 140: Is this a typo here? @nichia

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -1,20 +1,25 @@
...
+        <svg
+          width="16"
+          height="11"
+          class="ds-c-usa-banner__header-flag"
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <use href="{{root}}/images/usa-banner-flag.svg#svg"></use>
+        </svg>
+      </p>
+      <p class="ds-c-usa-banner__header-text">
```
bernardwang, line 15: Can we make the HTML example open by default?

bernardwang (approved):
I had a comment about a typo, but this is working well in Chrome and Safari. It LGTM considering we will address refactoring the SCSS in a later ticket

nichia:
Thanks, that was using the `background` property instead of the individuals background-image, repeat etc properties. Have fixed it.

packages/design-system/src/styles/components/_UsaBanner.scss
nichia: Fixed to use background-image etc

PR #913, opened 2021-01-08
[WNMGDS-734] Fix core-js imports in Tooltip

bernardwang:
### Summary 
- Move core-js dep into design-system package

### How to test
- Test this PR in [windowshop](https://github.cms.gov/CMS-WDS/windowshop/compare/master...pwolfert/cmsds-2.2.1-error) and ensure this error does not occur: 
Untitled

### DIFF
--- a/packages/design-system-scripts/package.json
+++ b/packages/design-system-scripts/package.json
@@ -29,7 +29,6 @@
     "chromedriver": "87.0.0",
     "cli-table2": "^0.2.0",
     "colors": "1.3.3",
-    "core-js": "^3.6.5",
     "cssnano": "4.1.10",
     "cssstats": "3.3.0",
     "del": "^3.0.0",
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -20,6 +20,7 @@
   "dependencies": {
     "@popperjs/core": "^2.4.4",
     "classnames": "^2.2.5",
+    "core-js": "^3.6.5",
     "downshift": "^3.2.10",
     "ev-emitter": "^1.1.1",
     "focus-trap-react": "^6.0.0",


nichia (approved):
LGTM.
btw - local testing of `windowshop` also works before the fix, the issue only surfaces if we reference the hcgov ds from npm.

bernardwang:
> LGTM.
> btw - local testing of `windowshop` also works before the fix, the issue only surfaces if we reference the hcgov ds from npm.

Hm, this is a bit concerning. Maybe we should try to reproduce locally before merging

bernardwang:
@nichia The reason why it passes locally is because we are installing the local branch with `file:`, which just copies over the `node_modules`. This is different than how its installed from the NPM registry

nichia:
Going to try testing with `yarn link` to see if that can duplicate the issue.

bernardwang:
@nichia i believe yarn link will work similarly unfortunately, we need to use a better testing method in the future. for now we can probably just merge as it’s a simple change

nichia:
Yes, wasn't able to duplicate when using `yarn link` for testing either. 

However, when trying to duplicate the issue, I noticed that the current HC.gov DS pattern for `Header` and `Footer` is not displaying when testing locally. Yet it works on [HC.gov DS Github pages](https://cmsgov.github.io/hcgov-design-system/patterns/header/).
Could all this be related?

bernardwang:
> Yes, wasn't able to duplicate when using `yarn link` for testing either.
> 
> However, when trying to duplicate the issue, I noticed that the current HC.gov DS pattern for `Header` and `Footer` is not displaying when testing locally. Yet it works on [HC.gov DS Github pages](https://cmsgov.github.io/hcgov-design-system/patterns/header/).
> Could all this be related?

Did you look further and see if there were any errors? It doesnt sound like it would be related but it could be something else

bernardwang:
@nichia Could you file a ticket for the Header/Footer issue? I took a quick look and it doesnt look related at all to this, but it is a weird bug we should fix

PR #914, opened 2021-01-13
[WNMGDS-564] Allow bottom placement error messages

bernardwang:
### Summary 
- Add the `errorPlacement` prop to allow bottom placed error messages
- Handle `errorId` in <FieldContainer>, pass it down to `TextField`, `Dropdown`,  and `DateField` in order to link the input to the error message with `aria-describedby`

Note: This PR doesn't yet include the addition of `aria-invalid` to the error messages

### How to test
- Test the [doc site](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-564/allow-bottom-placement-error-messages/components/form-label/) in different browsers and with different screenreaders 
- Test switching error placement on a single component level with the `errorPlacement` prop
- Test switching error placement on an entire component by updating a component's default `errorPlacement` prop value
- Run this PR in a product applications to ensure no changes to existing top placement, and no obvious errors for bottom placement
- Run this PR in an Typescript application to check the typing is correct


packages/design-system/src/components/DateField/DateInput.jsx
```diff
@@ -51,71 +51,51 @@ export class DateInput extends React.PureComponent {
...
       className: 'ds-l-col--auto',
       labelClassName: 'ds-c-datefield__label',
+      disabled: this.props.disabled,
       inversed: this.props.inversed,
       onBlur: (this.props.onBlur || this.props.onComponentBlur) && this.handleBlur,
       onChange: this.props.onChange && this.handleChange,
       numeric: true,
     };
 
+    // Add additional id to `aria-describedBy` with bottom placed errors
```
bernardwang, line 65: I generalized rendering of the day, month and year fields

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
```diff
@@ -60,16 +69,22 @@ interface FieldContainerProps {
...
+  static defaultProps = {
+    errorPlacement: 'top'
+  }
+  
   constructor(props: FieldContainerProps) {
     super(props);
 
     this.id = props.id || uniqueId('field_');
-    this.labelId = props.labelId || uniqueId('field_label_');
+    this.labelId = props.labelId || `${this.id}-label`;
```
bernardwang, line 80: I updated the generated `ids` to be more readable, i.e. `field_1`, `field_1-label`, `field_1-error` instead of `field_1`, `field_label_2`, `field_error_3`.

packages/design-system/src/components/FormLabel/FormLabel.jsx
```diff
@@ -51,6 +36,18 @@ export class FormLabel extends React.PureComponent {
     );
   }
 
+  errorMessage() {
+    if (this.props.errorMessage) {
+      // Include fallback for errorId for usage outside of FieldContainer
+      const errorId = this.props.errorId || `${this.props.fieldId}-error`;
```
bernardwang, line 42: Normally `errorId` is passed into `FormLabel` via `FormContainer`, but some apps are using `FormLabel` independently.

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
```diff
@@ -12,7 +13,15 @@ interface FieldContainerProps {
    * The HTML element used to render the container
    */
   component: 'div' | 'fieldset';
+  /**
+   * A unique ID to be used for the error message. If one isn't provided, a unique ID will be generated.
+   */
+  errorId?: string,
```
bernardwang, line 19: The errorId is handled the same as all the other ids, you can provide one, and if not, a unique one will be generated

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
```diff
@@ -97,23 +113,38 @@ export class FieldContainer extends React.Component<FieldContainerProps> {
...
+
     return (
-      <ComponentType className={classes}>
+      <ComponentType className={classes} >
         <FormLabel
           className={labelClassName}
           component={labelComponent}
-          errorMessage={errorMessage}
-          fieldId={this.id}
+          errorMessage={errorPlacement === 'top' ? errorMessage : null}
```
bernardwang, line 143: Avoid using the `errorMessage` prop in `FormLabel` when using the bottom placement

packages/design-system/src/components/Dropdown/Dropdown.jsx
```diff
@@ -33,10 +33,14 @@ export class Dropdown extends React.PureComponent {
         {...containerProps}
         component="div"
         labelComponent="label"
-        render={({ id, setRef }) => (
+        render={({ id, errorId, setRef }) => (
           <Select
             {...inputOnlyProps}
             {...{ id, setRef }}
+            // Link input to bottom placed error message
+            aria-describedBy={
```
nichia, line 41:
```suggestion
            aria-describedby={
```

packages/design-system/src/components/TextField/TextField.jsx
```diff
@@ -36,10 +36,14 @@ export class TextField extends React.PureComponent {
...
         component="div"
         labelComponent="label"
-        render={({ id, setRef }) => (
+        render={({ id, errorId, setRef }) => (
           <TextInput
             {...inputOnlyProps}
-            {...{ id, setRef }}
+            {...{id, setRef}}
+            // Link input to bottom placed error message
+            aria-describedBy={
```
nichia, line 44:
```suggestion
            aria-describedby={
```

packages/design-system/src/components/Dropdown/Dropdown.jsx
nichia, line 41:
```suggestion
            aria-describedby={
```

packages/design-system/src/components/DateField/DateInput.jsx
pwolfert: Clever reuse of `classNames` 👍

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
```diff
@@ -82,11 +97,12 @@ export class FieldContainer extends React.Component<FieldContainerProps> {
     }
   }
 
-  render() {
+  render(): React.ReactNode {
```
pwolfert, line 100: Is this return type necessary? I've never had to annotate this before

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
```diff
@@ -97,23 +113,38 @@ export class FieldContainer extends React.Component<FieldContainerProps> {
...
     // Field input props handled by <FieldContainer>
-    const fieldInputProps = {
+    const fieldInputProps = { 
       id: this.id,
       labelId: this.labelId,
+      errorId: this.errorId,
       setRef: this.setFieldRef,
     };
 
+    const renderBottomError = () => {
```
pwolfert, line 130: Since this is not used in a conditional block below, does it need to be a function? Could it just be a value?

pwolfert (approved):
Most of this LGTM 👍; I just have a couple questions below?

nichia (commented):
Nice handling of the error props and function. Just noted some suggested changes.

packages/design-system/src/components/DateField/DateInput.jsx
nichia: Nice refactoring.

line47:
Bottom error placement on the autocomplete breaks the design and separates the autocomplete box from the input box.
I think we could explore updating the markup to place the autocomplete options box inside the same div the input lives in. 

Screen Shot 2021-01-25 at 11 42 04 AM

line47:
I tested this across browsers in different healthcare applications as well as in the typescript CRA and everything looked good with the exception of the autocomplete issue found above. 

I'm also wondering where we can document this global message placement change?
I'm thinking it should be part of the cmsds.config file to keep these type of global flags in the same file for ease of use.

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
bernardwang: I was getting an typescript warning on it, maybe my typescript config is wonky. I was getting other weird typescript errors too that i need to look into

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
bernardwang: yep good catch!

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
bernardwang: When the errorMessage is placed on the bottom we handle it inside `FieldContainer` [here](https://github.com/CMSgov/design-system/pull/914/files/7c88fbfe9f3ab2fdd10910cd6dac08d9734b5881#diff-68aab052f9429968f245cd29037c4ead7ebaf1192a95cea8aec887b86e5e970cR156). When its top placement, we pass errorMessage into `FormLabel` to be rendered

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
```diff
@@ -131,6 +163,7 @@ export const FieldContainerPropKeys = [
   'className',
   'component',
   'errorMessage',
+  'errorPlacement',
```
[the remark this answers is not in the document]
bernardwang: good catch!

packages/design-system/src/components/Dropdown/Dropdown.jsx
bernardwang: Did you end up committing this change? @nichia did it work ok?

nichia: Not sure why, but when I tried the using `aria-describedby`, it doesn't show up on the html. So I did not commit this change and we'll just see a warning message of `aria-describedBy`

bernardwang:
@line47  I looked into this, and basically theres no easy good solution as far as I can tell. If possible, I'd prefer to handle this in a different PR. 
https://github.com/CMSgov/design-system/pull/914/files#issuecomment-766953535

I came up with a solution that applies `position: absolute` to the bottom placed error 
autocomple1
autocomplete3

However there are some issues with this, i.e. absolutely positioned error message wont take up space, and also if the error message is very long it will conflict with the search clear button. 
I could alternatively update the positioning of the autocomplete list since it is already absolutely positioned, but we would still need to handle the search clear button. Basically this requires a bit more thought and can be handled in a follow up

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
```diff
@@ -80,6 +80,10 @@ export class ChoiceList extends React.PureComponent {
   }
 }
 
+ChoiceList.defaultProps = {
```
bernardwang, line 83: Originally I had hoped to have the `errorPlacement` default occur inside `FieldContainer`, but there are several areas where the higher level components need to access this prop to properly handle the bottom placed error. I.e. Dropdown, TextField, and Autocomplete (eventually)

bernardwang: This means the flag will have to be used in several areas instead of one, but I'm not sure if theres another solution here.

line47: So does this mean that the setting of the defaults will need to happen in the components themselves instead of the higher level `FieldContainer` ?

nichia:
- Dropdown.jsx is missing `import classNames from 'classnames'`;

nichia:
Testing on HCgov DS, windownshop and taxtools runs as expected. However, noticed that: 
- `yarn test` is showing errors with: `'React' was used before it was defined  no-use-before-define`: not sure if there's a [mismatch of typescript version?](https://stackoverflow.com/questions/63818415/react-was-used-before-it-was-defined)
- and `yarn test:e2e` has some errors

nichia:
Other then the errors with yarn test & test:e2e, it also works great on examples Typescript and React applications, and on Mgov DS too, but I'm not able to test on coverage-tools-frontend yet as there's a non-related issues causing the test to fail.

packages/design-system/src/components/ChoiceList/ChoiceList.jsx
bernardwang: yep pretty much. Once we make the flag it will be the same either way for whoever is setting the default but it means there are multiple components using the flag rather than just 1

davidakennedy (changes requested):
@bernardwang This is looking pretty good! I want to do some more testing, but wanted to share this feedback so far:

* Are we going to try and tackle associating error messages better with components that have fieldsets and legends? Like I experimented with here: https://davidakennedy.github.io/error-msg-prototypes/#option12 Or is that something we should do in a separate PR? I wouldn't want this change to go out with something in place to better help on fieldsets and legends.
* Date field labels aren't associated properly. Looks like this is missing, even though it's in the current DS. Labels are missing the `for="id"` part.
* Dropdown has error in the console: `ReferenceError: Can't find variable: classNames`.

Will keep testing!

nichia:
Also just did another round of testing and noticed similar points that DK mentioned above like the label missing the "for=id" and "aria-describedby". I've captured these in the below:
Dropdown component:
Screen Shot 2021-01-28 at 2 38 57 PM

Date field component:
Screen Shot 2021-01-28 at 2 36 35 PM

ChoiceList component:
Screen Shot 2021-01-28 at 2 32 18 PM

Textfield component:
Screen Shot 2021-01-28 at 2 24 55 PM

bernardwang:
@davidakennedy

> * Are we going to try and tackle associating error messages better with components that have fieldsets and legends?

Yes, I just forgot to handle this for ChoiceList. I handled it in DateField with `aria-describedby`, but I'll should just handle every `fieldset` component to use the hidden error message

> Labels are missing the for="id" part.

Ah thanks for the catch, this was a bug that was fixed in the latest commits

nichia (approved):
LGTM.
Great work Bernard.

davidakennedy (approved):
Changes look good. Nice work, @bernardwang!

Let's address adding `aria-invalid` on fieldsets in a follow up, if we decide it makes sense.

PR #916, opened 2021-01-15
[WMNGDS-722] Update Banner scss

nichia:
### Changed
- Refactor banner scss
- Update svg files U.S Flag and Locked padlock to have aria-labelledby for accessibility
- Update banner examples and test snap-shot 

### How to test
- Run the doc site locally with `yarn start` to test the Banner component (just noticed that we're not able to test on the demo site for this as the image file are not being picked up correctly on the demo site)
- Ensure banner works on various browsers and is accessible by screen readers


packages/design-system-docs/src/pages/components/UsaBanner/UsaBanner-es.example.jsx
```diff
@@ -0,0 +1,11 @@
+import React from 'react';
+import ReactDOM from 'react-dom';
+import { UsaBanner } from '@cmsgov/design-system';
+
+ReactDOM.render(
+  <div>
+    <h6 className="preview__label">Spanish banner</h6>
```
bernardwang, line 7: Why is this a separate example file?

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -2,25 +2,28 @@ $usa-banner-background-color: #f0f0f0;
...
+  font-size: 1rem;
   padding: 0;
 }
 
 .ds-c-usa-banner__header {
   align-items: flex-start;
   display: flex;
   flex-direction: row;
-  font-size: 0.79853rem;
+  font-size: 0.8rem;
```
bernardwang, line 20: 13px or `$small-font-size`

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -2,25 +2,28 @@ $usa-banner-background-color: #f0f0f0;
...
   font-weight: 400;
```
bernardwang, line 21: We could replace it with `$font-normal`

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -2,25 +2,28 @@ $usa-banner-background-color: #f0f0f0;
...
-  line-height: 1.12707;
+  line-height: inherit;
```
bernardwang, line 22: Is this line necessary? Does the style change if you just delete it?

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -75,30 +79,15 @@ $usa-banner-header-action-color: #005ea2;
...
 .ds-c-usa-banner__button {
   background-color: transparent;
   border: 0;
-  border-radius: 0;
   bottom: 0;
   box-shadow: none;
   display: block;
-  font-size: 0.8rem;
-  height: auto;
+  font-size: inherit;
```
bernardwang, line 88: Same with this, can we remove this style?

packages/design-system/src/styles/components/_UsaBanner.scss
nichia: Needed this, otherwise style will change.

packages/design-system/src/styles/components/_UsaBanner.scss
nichia: Same for this.

packages/design-system-docs/src/pages/components/UsaBanner/UsaBanner-es.example.jsx
nichia: Merged back with original example file.

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -24,6 +22,60 @@ export class UsaBanner extends React.PureComponent {
...
+        {/* eslint-disable-next-line no-template-curly-in-string */}
+        <desc id={`banner-lock-description-${this.id}`}>A locked padlock</desc>
+        <path
+          fill="#000000"
+          fillRule="evenodd"
+          d="M26 0c10.493 0 19 8.507 19 19v9h3a4 4 0 0 1 4 4v28a4 4 0 0 1-4 4H4a4 4 0 0 1-4-4V32a4 4 0 0 1 4-4h3v-9C7 8.507 15.507 0 26 0zm0 8c-5.979 0-10.843 4.77-10.996 10.712L15 19v9h22v-9c0-6.075-4.925-11-11-11z"
+        />
+      </svg>
+    );
+
```
nichia, line 78: Svg files for `U.S flag and lock` are placed here to fix the accessibility issue of browsers not reading the `role="img"` flag even after updating the script to pass `role, aria-labelledby, title and desc` svgmin options to the `dist/imag/` folder. 
Also noted that uswds banner have the `lock svg` file function in the code instead of on the external image file as well - is it to workaround the accessibility issue of screen reader not reading correctly?

packages/design-system-docs/src/pages/components/UsaBanner/usabanner.example.html
```diff
@@ -1,32 +1,40 @@
...
+        width="16"
+        height="11"
+        xmlns="http://www.w3.org/2000/svg"
+        class="ds-c-usa-banner__header-flag"
+        aria-labelledby="banner-flag-title-1 banner-flag-description-1"
+        focusable="false"
+      >
+        <title id="banner-flag-title-1">Flag</title>
+        <desc id="banner-flag-description-1">U.S. flag</desc>
+        <use href="{{root}}/images/usa-banner-flag.svg#svg"></use>
```
nichia, line 16: Noticed that on IE 11, the svg `<use... />` command is picking up the image content on this banner html example. Have confirmed that it is the same for `button with Icon` component where it works on other browsers except for IE11.

bernardwang: nice job testing, we can file a separate ticket for that issue

packages/design-system-scripts/gulp/common/copyFontsImages.js
```diff
@@ -15,10 +15,21 @@ function minimizeSvg(srcGlob, dest) {
...
+            {
               removeDesc: false,
+            },
+            {
               removeHiddenElems: false,
             },
+            {
+              removeUnknownsAndDefaults: {
+                keepRoleAttr: true,
+              },
```
bernardwang, line 31: @nichia did this work?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -1,15 +1,15 @@
-import DotGovIcon from '../../images/usa-banner-dot-gov.svg';
 import EnglishTranslations from '../../locale/en.json';
-import HttpsIcon from '../../images/usa-banner-https.svg';
-import LockIcon from '../../images/usa-banner-lock.svg';
+import IconDotGov from '../../images/usa-banner-dot-gov.svg';
```
bernardwang, line 2: why was this renamed?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
bernardwang: I would prefer to leave the SVG usage as it originally was, even if the title, description and role are not working. we can always fix that later by filing a ticket to fix svg min

bernardwang: the problem here is that our SVG in react system is to import the SVG file directly, and use it inline. This promotes code reuse and consistently with our icons, and is better than writing the SVG inline manually. also, regardless of svgmin, you should still be able to use role, just apply it to the imported SVG as another prop

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
nichia: Want to group the svg icons together.

nichia: Would it be better to keep it's original naming?

packages/design-system-scripts/gulp/common/copyFontsImages.js
nichia: Yes it worked. I checked the svg file created under `dist/image` to confirm that the `role="image"` option is present

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
nichia: Ah, I didn't think of applying the role to the imported svg as another prop. Will give that a try to see if that fixes the issue of screen reader not recognizing the `U.S. flag` and `Lock` SVG file as an image like when the svg inline is added manually.

nichia: Updated to use the inline SVG and pass in `role` prop

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
line47: That makes sense to be prefixing the imports with "Icon" to keep them together

packages/design-system/src/images/usa-banner-lock.svg
```diff
@@ -1,3 +1,5 @@
-<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 52 64">
+<svg role="img" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 52 64" aria-labelledby="banner-lock-title banner-lock-description" focusable="false">
+    <title id="banner-lock-title">Lock</title>
+    <desc id="banner-lock-description">A locked padlock</desc>
```
line47, line 3: I don't think we need to use `aria-labelledby` for the lock or dot-gov icon because they are purely decorative and are there to accompany the text

line47 (changes requested):
One small change and then looks good to me!

packages/design-system/src/images/usa-banner-lock.svg
nichia: I added it to stay consistent with USWDS which has `aria-labelleby` for `U.S flag` on the header and the `Lock` image on this content `Secure .gov websites use HTTPS` for screen readers

nichia: And the `dot-gov` and other `lock` image is set with aria-hidden.

line47: I see what you mean Ni! Thanks! 
I think I agree with needing aria attributes for those images but I'm not going to block the merge. I'll bring this up with the USWDS team.

line47 (approved):

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
bernardwang: you dont need these variables anymore, you should just be using the icon components directly in the JSX

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -85,6 +94,11 @@ UsaBanner.defaultProps = {
 
 UsaBanner.propTypes = {
   /**
+   * A unique `id` to be used on the usa banner.
```
bernardwang, line 97: `A unique ID to be applied to the banner content. A unique ID will be generated if one isn't provided.`

packages/design-system/src/images/usa-banner-flag.svg
```diff
@@ -1,4 +1,6 @@
-<svg width="16" height="11" xmlns="http://www.w3.org/2000/svg">
+<svg role="img" width="16" height="11" xmlns="http://www.w3.org/2000/svg" aria-labelledby="banner-flag-title banner-flag-description" focusable="false">
```
bernardwang, line 1: Isnt `aria-labelledby` already handled elsewhere? The SVG file should be a generic SVG file, attributes that depend on ids and classes should be applied when they are actually used in React

packages/design-system/src/images/usa-banner-lock.svg
bernardwang: @nichia [Same feedback as here](https://github.com/CMSgov/design-system/pull/916/files#r560486331). I'm ok with keeping the aria attributes but they should be handled in React not in the default SVG file

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -2,45 +2,46 @@ $usa-banner-background-color: #f0f0f0;
...
-  font-size: 0.79853rem;
-  font-weight: 400;
-  line-height: 1.12707;
-  min-height: 2rem;
-  padding-bottom: 0.5rem;
-  padding-top: 0.5rem;
+  font-size: 12px;
+  font-weight: $font-normal;
+  min-height: $spacer-4;
+  padding-bottom: $spacer-1;
```
bernardwang, line 22: Can we combine these padding styles into a single line

packages/design-system/src/images/usa-banner-flag.svg
bernardwang: Update: sorry, I didn't realize the aria attributes are using `id`s that are also defined in the same SVG file. If that's the case then its fine

bernardwang:
@nichia I know this isnt related to the SCSS, but it might be worth adding a `className` prop in this PR, in case folks need to custom style the banner component

bernardwang (approved):
Great work @nichia LGTM!

PR #917, opened 2021-01-25
[WNMGDS-745] Fix HCgov Header/Footer displaying empty screen

nichia:
## Summary
Fix HCgov Header/Footer component page not displaying correctly on local testing of HCgov design system document site (note: odd that it does work on github pages though)

This issue was introduced in [PR Autocomplete component not rendering](https://github.com/CMSgov/design-system/pull/881)

### How to test
- Setup HCgov child design system in the package.json to reference this branch with `file:../design-system/package/design-system/`
- Run `git clean -fdx` and then `yarn install` to Install the local DS branch
- Test that the Header and Footer component pages are displaying correctly

### DIFF
--- a/packages/design-system/src/components/Autocomplete/Autocomplete.jsx
+++ b/packages/design-system/src/components/Autocomplete/Autocomplete.jsx
@@ -14,7 +14,6 @@
  * an unacceptable regression of the user experience.
  */
 
-import { get, uniqueId } from 'lodash';
 import Button from '../Button/Button';
 import Downshift from 'downshift';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


bernardwang:
oh this is super odd , how did you find out this that fixes the hcgov error? was this changed recently?

nichia:
Had to tediously test the changes to find that the fix is to separate the import for `lodash` functions

bernardwang:
> Had to tediously test the changes to find that the fix is to separate the import for `lodash` functions

Oh wow, could you link the PR or commit that did this change?

bernardwang:
@nichia It looks like we should avoid using [lodash per method packages](https://lodash.com/per-method-packages) as well, so I think it should be `lodash/uniqueid` instead of `lodash.uniqueid`

nichia:
There's 9 components that use `import uniqueId from 'lodash.uniqueid';` - think it's okay to update all in this pr?.

bernardwang:
@nichia I made this ticket as a follow up to lodash imports for the future
https://jira.cms.gov/browse/WNMGDS-762

bernardwang (approved):
This is working for me, great work finding this @nichia! If only we could figure out why this is causing the error

PR #919, opened 2021-01-27
[WNMGDS-731] Update responsive stacked table

nichia:
## Summary
Update the stacked table component to more closely align with the recently released [USWDS responsive stacked table design](https://designsystem.digital.gov/components/table/#responsive-stacked-table).

## Updated
- Layout of stacked table

## Added
- Stacked table with headers variant via `stackableStyles="header"` props

#### Stacked bordered table - new layout
Screen Shot 2021-01-27 at 2 10 14 PM

#### Stacked table - old layout
Screen Shot 2021-01-27 at 2 09 56 PM

### How to test
- Ensure stacked table works on various browsers and is accessible by screen readers
- http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-731/Update-responsive-stacked-table/


nichia:
@davidakennedy - not sure how to confirm if screen reader accessibility is right for this PR. I was trying to test VoiceOver on this PR compare to uswds stacked table - but not sure why I can't get VoiceOver to work on [uswds stacked table](https://designsystem.digital.gov/components/table/#responsive-stacked-table)

packages/design-system/src/components/Table/Table.jsx
```diff
@@ -191,6 +193,10 @@ Table.propTypes = {
    * The viewport size at which responsive vertically stacked row styles are applied. Only used when the `stackable` prop is set to `true`. [Read more on breakpoints](/guidelines/responsive/)
    */
   stackableBreakpoint: PropTypes.oneOf(['sm', 'md', 'lg']),
+  /**
+   * Stacks the table cells on narrow screens and visually promotes the first cell of every row into a “header” for that group.
```
bernardwang, line 197: What is this prop for? We didnt have it originally

nichia: Added the StackableStyles props - specify 'header` if we want the stacked table to display with the first row as a header, with bold font and background color.

nichia: Also just added the `stacked borderless table with headers` to the React table example

bernardwang: Hm I see, its a new stacked table design then. I think for this ticket it's better to avoid adding that prop and new style to avoid expanding the scope of this work. Teams that want that style can always manually add it, or request us for it, so maybe we can tackle this in the future

bernardwang: Another reason why not to include it is because our `props` arent currently setup well to handle it. We probably dont want to call the prop `stackableStyles`, and it doesnt make sense to make it `Proptype.oneOf(['header'])` if there is only one option. We need to think more intentionally on how we want folks to be using these variants and props.

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -105,101 +104,106 @@ $table-dense-padding: $spacer-half $spacer-2 !default;
...
+    @include table-stacked-header-styles;
   }
 }
 
 .ds-c-table--stacked {
   @include table-stacked;
 }
 
-.ds-c-sm-table--stacked {
+.ds-c-table--stacked-sm {
```
bernardwang, line 194: Why was this naming pattern changed?

nichia: Trying to rename this so it stays closer with the BEM naming convention, i'm thinking it's better to do this while the table is still in draft status.

bernardwang: I can see the argument to name it this way, as the breakpoints could be seen as a modifier, but the current convention for classes with breakpoints in the DS is for the breakpoint to follow the prefix. If you look up any of our responsive utilities, you can see an example of this.

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -85,7 +84,7 @@ $table-dense-padding: $spacer-half $spacer-2 !default;
   text-align: left;
 }
 
-@mixin table-stacked {
+@mixin table-stacked-styles {
```
bernardwang, line 87: I think this doesnt have to be renamed

bernardwang, line 87: What about calling this `stacked-table-styles`

packages/design-system/src/styles/components/_Table.scss
nichia: Thanks for picking this up, i overlooked our ds convention. The styles name should follow it. Will revert back the name.

bernardwang:
@nichia This can be handled quickly in a different PR, but could we rename `dense` to `compact` to match USWDS?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -105,102 +104,86 @@ $table-dense-padding: $spacer-half $spacer-2 !default;
...
-      grid-column: 1;
-      max-width: $table-stacked-header-max-width;
-      width: calc(100% - #{$spacer-2});
-    }
-
-    // TableCell content is wrapped in a span for IE11 grid support
-    > span {
-      // Fixes IE11 display bug https://css-tricks.com/css-grid-in-ie-debunking-common-ie-grid-misconceptions/
-      display: block;
+  thead {
```
bernardwang, line 107: Why is this necessary? Isn't the styling for `thead tr` already handle visually hiding the `thead`?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -105,102 +104,86 @@ $table-dense-padding: $spacer-half $spacer-2 !default;
...
-      border-right: 1px solid;
-      bottom: 0;
-      content: '';
-      display: block;
-      left: 0;
-      max-width: $table-stacked-header-max-width;
-      position: absolute;
-      top: 0;
-      width: 100%;
+    th:first-child,
```
bernardwang, line 125: What is this styling for? Doesnt the first element in a `tr` have a top border?

bernardwang: Doesn't this styling override this one? https://github.com/CMSgov/design-system/pull/919/files#diff-2b709213c3173751c76a6658e5db4facf82984bc78f95a2598faf72564b4ebcfR137

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -105,102 +104,86 @@ $table-dense-padding: $spacer-half $spacer-2 !default;
...
+      th {
+        background-color: inherit;
+      }
     }
-  }
 
-  td:first-child,
-  th:first-child {
-    border-top: 0;
+    &:first-child {
```
bernardwang, line 137: What is this styling for? Why is it that there is only top border styling applied to the first children of the first row? Shouldnt every row have a top border style?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -105,102 +104,86 @@ $table-dense-padding: $spacer-half $spacer-2 !default;
...
-      width: 100%;
+    th:first-child,
+    td:first-child {
+      border-top-width: 0;
+    }
+
+    &:nth-child(odd) {
+      td,
+      th {
+        background-color: inherit;
```
bernardwang, line 133: What is this for?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -105,102 +104,86 @@ $table-dense-padding: $spacer-half $spacer-2 !default;
...
+    }
   }
 
-  &.ds-c-table--borderless {
-    tr {
-      border: 0;
+  // stylelint-disable selector-no-qualifying-type
+  th[data-title],
+  td[data-title] {
+    padding-bottom: 12px;
```
bernardwang, line 150: Is there a ready why the table cells have a different padding? Normally we are using the `$table-padding` variable, why should this change for stacked?

nichia:
Thanks for making the SCSS updates and simplification, the changes look good.

packages/design-system/src/styles/components/_Table.scss
nichia: Thanks for this pick up, I've removed the striped styles from stacked table initially, but have now updated to apply striped styles on stacked table as well. 
Also, for our styles, the gray stripe is applied on even rows whereas uswds applied stripes on odd rows. Do you think we should change to 'odd' rows as well?

bernardwang: I saw that the USWDS stacked table design doesnt use the striped styles, so I'm actually included to do the same and removed striped styles from stacked tables

bernardwang: and as per our convo, we should leave the striped update to a separate ticket and PR

packages/design-system/src/styles/components/_Table.scss
bernardwang: @nichia

packages/design-system/src/styles/components/_Table.scss
bernardwang: @nichia

packages/design-system/src/styles/components/_Table.scss
nichia: Thanks for this pickup, we should remove this 12px setting and just be using the $table-padding variable

packages/design-system/src/styles/components/_Table.scss
nichia: Yes, the thicker border styles is needed at the top and bottom

packages/design-system/src/styles/components/_Table.scss
bernardwang: Oh ok, do you know where the 12px originally came from?

nichia: 12px is the bottom padding that uswd has for stacked table

packages/design-system/src/styles/components/_Table.scss
nichia: Thanks, this is not necessary, `thead tr` already handle visually hiding.

packages/design-system/src/styles/components/_Table.scss
nichia: If we remove this, stacked table border-top on <tr> row 2 and higher will be thicker. Have added comment on why we have this line.

line47 (approved):
This looks good to me!

davidakennedy (changes requested):
I tested these with our major browser/screen reader combos and the results were mixed.

Overall, I wonder if there's a way we can avoid relying on CSS generated content. Screen readers handle that differently and it makes for an annoying experience in some combinations.

*Safari/VoiceOver*: Announces the CSS generated content, making information repeated. Also, the table navigation in VoiceOver appears to not work correctly.

*Edge/Jaws*: Mostly works as expected.

*Chrome/Jaws*: Mostly works as expected.

*Firefox/NVDA*: The table navigation in NVDA appears to not work correctly.

I think we can probably classify some repeated information as annoyance rather than showstopper. But we should try to fix the table navigation issues. I'm going to do some more research on this stuff.

bernardwang:
Thanks for reviewing @davidakennedy! I also ran into the repetition and table navigation issue. I think the `psuedo` element is the main culprit here, because it interferes with the arrow commands that are used in navigation. However, I think this will have to be handled in the future, as currently there isnt an easy way to apply something like `aria-hidden` to the `psuedo` elements, and I cant really think of another CSS only option. We could explore something that requires a markup change in the future, but considering the existing Table solution in prod also uses `psuedo` elements and has the same issues you are describing, we shouldn't let that block this from merging.

davidakennedy (approved):
@nichia and I chatted through the issues here and came up with a strategy:

- We'll merge this as is, since we're still in draft phase.
- We'll continue to research the main blocking issue, which is NVDA table navigation not working.
- We'll also research why Safari/VoiceOver doesn't announce table headers, and see if we can make it less aggressive about announcing the CSS generated content.

We'll keep working on this in the next sprint.

PR #922, opened 2021-01-29
[WNMGDS-686] Use aria-invalid attribute in form components

bernardwang:
## Summary
- Handle `aria-invalid` in TextField and Dropdown when `errorMessage` prop is provided
- Handle `aria-invalid` in FieldContainer when `fieldset` component type is used and `errorMessage` prop is provided
- Refactor and move aria attribute handling to TextInput and Select
- Add tests and snapshots

## How to test
- http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-686/add-aria-invalid

### DIFF
--- a/packages/design-system/src/components/DateField/DateInput.jsx
+++ b/packages/design-system/src/components/DateField/DateInput.jsx
@@ -117,10 +117,6 @@ DateInput.propTypes = {
    * Disables all three input fields.
    */
   disabled: PropTypes.bool,
-  /**	
-   * The ID of the error message applied to this field.	
-   */	
-  errorId: PropTypes.string,
   /**
    * Applies the "inverse" UI theme
    */
--- a/packages/design-system/src/components/DateField/DateInput.jsx
+++ b/packages/design-system/src/components/DateField/DateInput.jsx
@@ -78,6 +78,7 @@ export class DateInput extends React.PureComponent {
         }}
         autoComplete={this.props.autoComplete && `bday-${type}`}
         aria-describedby={this.props.labelId}
+        aria-invalid={this.props[`${type}Invalid`]}
       />
     );
   }
--- a/packages/design-system/src/components/DateField/DateInput.jsx
+++ b/packages/design-system/src/components/DateField/DateInput.jsx
@@ -117,9 +117,9 @@ DateInput.propTypes = {
    * Disables all three input fields.
    */
   disabled: PropTypes.bool,
-  /**
-   * The ID of the error message applied to this field.
-   */
+  /**	
+   * The ID of the error message applied to this field.	
+   */	
   errorId: PropTypes.string,
   /**
    * Applies the "inverse" UI theme
--- a/packages/design-system/src/components/FieldContainer/FieldContainer.test.tsx
+++ b/packages/design-system/src/components/FieldContainer/FieldContainer.test.tsx
@@ -6,6 +6,7 @@ const defaultProps = {
   label: 'Foo',
   component: 'div',
   labelComponent: 'label',
+  errorPlacement: 'top',
   /* eslint-disable */
   render: ({ id, labelId, setRef }) => {
     return <input id={id} aria-describedby={labelId} ref={setRef} />;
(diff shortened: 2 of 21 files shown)


packages/design-system/src/components/DateField/DateInput.jsx
```diff
@@ -78,6 +78,7 @@ export class DateInput extends React.PureComponent {
         }}
         autoComplete={this.props.autoComplete && `bday-${type}`}
         aria-describedby={this.props.labelId}
+        aria-invalid={this.props[`${type}Invalid`]}
```
bernardwang, line 81: Because of this change, `DateField` components will use `aria-invalid` in 2 different ways. The attribute will be on the `fieldset` that contains the 3 fields, and additionally on any field that has an error state. @davidakennedy 

So the resulting markup will be something like this:
```
<fieldset aria-invalid=true>
   <legend>...</legend>
   <div>
      <input/>
      <input/>
      <input aria-invalid=true/>
   </div>
</fieldset>
```

Let me know what you think of this, if its overkill or not. We could scale it back to only using the `aria-invalid` on the `legend` if thats less confusing. We could also only use `aria-invalid` on the inputs for the `DateField` component, but right now the way it work is that `aria-invalid` is automatically added to errored components using `fieldset` and `legend`, so that would be a bit tricker to setup.

nichia:
For top error placement of TextField, will aria-invalid="true" and aria-describedby still apply?  Screen Shot 2021-02-01 at 4 26 07 PM

bernardwang:
> For top error placement of TextField, will aria-invalid="true" and aria-describedby still apply? Screen Shot 2021-02-01 at 4 26 07 PM

yes, we still need to show `aria-invalid=true`. We dont need `aria-describedby` which is why there isnt a value there. However, I can update the code so we dont include the attribute if its blank

nichia:
The usage of aria-invalid attribute in form components works as expected. 
Below is what I have observed as the usages of `aria-invalid`, `aria-describedby` and `visually hidden error message`:
Screen Shot 2021-02-02 at 9 46 02 AM

packages/design-system/src/components/DateField/DateInput.jsx
line47: I think ideally having it on the input only would be good but given the context I understand why it's at the fieldset level.

packages/design-system/src/components/FieldContainer/FieldContainer.test.tsx
```diff
@@ -25,24 +26,21 @@ function render(customProps = {}, deep = false) {
...
-        labelComponent: 'legend',
-      },
-      true
-    );
+    const data = render({
+      component: 'fieldset',
+      labelComponent: 'legend',
+    });
 
-    expect(data.wrapper.find(data.props.component).length).toBe(1);
```
line47, line 44: I got an error on this test when running `yarn test` 
```
 Expected: "Foo"
    Received: ["Foo", null]

      48 |     const data = render();
      49 | 
    > 50 |     expect(data.wrapper.find('FormLabel').prop('children')).toBe(data.props.label);
         |                                                             ^
      51 |   });
      52 | 
      53 |   it('passes hint to FormLabel', () => {
```

nichia:
`import classNames from 'classnames';` is not used, it can be deleted from `Dropdown.jsx`

bernardwang:
@nichia @line47 
This error happens when `react-docgen` is trying to parse `FieldContainer.tsx` and `InlineError.tsx` files for react prop documentation. The HC.gov DS doesnt contain a `tsconfig.json`, so it errors at this point. However, both FieldContainer and InlineError are private components that arent documented, so its not a big deal. We can also add a `tsconfig.json` to the HC.gov DS when we release this to get rid of the error

```
react-docgen Error: Cannot load custom tsconfig.json from provided path: ./tsconfig.json
```

bernardwang:
@line47 @nichia 
The reason why child DS doc sites aren't showing the bottom error change is because of the way we are testing, and our dependencies. The child DS's are using the `design-system-docs` dependency to get those examples files, and those examples files import from the `design-system` package. Because the package.json for `design-system-docs` specifies `"@cmsgov/design-system": "^2.2.1"`, the example files are importing components from 2.2.1, not your local branch. One fix is to manually replace the files inside `hcgov-design-system/node_modules/@cmsgov/design-system-docs/node_modules/@cmsgov/design-system` with the correct local files 


Untitled

nichia:
Would you know why we did not have this with the last round of error placement testing?

packages/design-system/src/components/DateField/DateInput.jsx
bernardwang: > We could scale it back to only using the aria-invalid on the legend if thats less confusing

Oops I had a typo, I meant to say we could scale the `aria-invalid` back to only on the `fieldset`, the `legend` doesnt have anything.

bernardwang: @davidakennedy 
Just to clarify, you are ok with having 2 instances of `aria-invalid` on the DateField component, 1 instance is on the `fieldset` and the other is on errored `input` fields

bernardwang:
> Would you know why we did not have this with the last round of error placement testing?

@nichia Honestly not sure, it seems like this problem should have happened last round too

nichia (approved):
Great work, the change looks good on core ds and child ds!

bernardwang:
> Would you know why we did not have this with the last round of error placement testing?
> I thought by these files references on child ds `package.json` would be telling the child ds to reference core ds aria-invalid branch?:
> ` "@cmsgov/design-system-docs": "file:../design-system/packages/design-system-docs", "@cmsgov/design-system-scripts": "file:../design-system/packages/design-system-scripts", "@cmsgov/design-system": "file:../design-system/packages/design-system",`

@nichia It is, but `"file:../design-system/packages/design-system-scripts"` is referencing `"@cmsgov/design-system": "^2.2.1"`. When we run everything locally, we use `lerna` to link all these parts together, so we are using the local packages instead of whats inside the package.json. However, when we test with `file:../` we are only copying files over, and those files specify the version inside the NPM registry, not our local packages.

bernardwang:
In the interest of closing out this ticket, I'll be merging this PR. This is merging into `feature/form-container` not master, so there will be opportunities to address feedback

PR #927, opened 2021-02-04
[WNMGDS-781] Rename dense table

nichia:
### Changed
- Rename `dense` to `compact` to match USWDS table

## How to test
- Check the Table component example to ensure that `dense` table is now referenced as `compact` table
- http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-781/rename-dense-table

### DIFF
--- a/packages/design-system/src/components/Table/TableCell.jsx
+++ b/packages/design-system/src/components/Table/TableCell.jsx
@@ -72,9 +72,11 @@ export const TableCell = ({
       {...tableCellProps}
     >
       {/* Add column heading for stacked table */}
-      <span aria-hidden="true" className="ds-c-table--stacked__col-header">
-        {stackedTitle}
-      </span>
+      {_stackable && (
+        <span aria-hidden="true" className="ds-c-table--stacked__col-header">
+          {stackedTitle}
+        </span>
+      )}
       {children}
     </Component>
   );
--- a/packages/design-system/src/components/Table/TableCell.jsx
+++ b/packages/design-system/src/components/Table/TableCell.jsx
@@ -69,14 +69,9 @@ export const TableCell = ({
       scope={defaultScope}
       headers={headers}
       id={id}
+      data-title={stackedTitle}
       {...tableCellProps}
     >
-      {/* Add column heading for stacked table */}
-      {_stackable && (
-        <span aria-hidden="true" className="ds-c-table--stacked__col-header">
-          {stackedTitle}
-        </span>
-      )}
       {children}
     </Component>
   );
--- a/packages/design-system/src/components/Table/TableCell.jsx
+++ b/packages/design-system/src/components/Table/TableCell.jsx
@@ -69,9 +69,12 @@ export const TableCell = ({
       scope={defaultScope}
       headers={headers}
       id={id}
-      data-title={stackedTitle}
       {...tableCellProps}
     >
+      {/* Add column heading for stacked table */}
+      <span aria-hidden="true" className="ds-c-table--stacked__col-header">
+        {stackedTitle}
+      </span>
       {children}
     </Component>
   );
--- a/packages/design-system/src/styles/components/_Table.scss
+++ b/packages/design-system/src/styles/components/_Table.scss
@@ -48,7 +48,8 @@ $table-compact-padding: $spacer-half $spacer-2 !default;
   }
 }
 
-.ds-c-table--compact {
+.ds-c-table--compact,
+.ds-c-table--dense {
   th,
   td {
     padding: $table-compact-padding;
@@ -84,11 +85,6 @@ $table-compact-padding: $spacer-half $spacer-2 !default;
   text-align: left;
 }
 
-// Hide column heading reserved for stacked table
-.ds-c-table--stacked__col-header {
-  display: none;
-}
-
 @mixin stacked-table-styles {
   thead tr {
     // Hide column headings when responsive table stacked vertically on narrow viewports
@@ -120,7 +116,7 @@ $table-compact-padding: $spacer-half $spacer-2 !default;
     // Every row has a bottom border
     border-bottom: $spacer-half solid $table-border-color;
     border-top: 0;
-    display: block;
+    display: table-row;
 
     // Remove the initial th/td border-top to fix the thicker top border on <tr> 2 and more
     th:first-child,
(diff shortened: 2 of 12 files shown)


packages/design-system/src/styles/components/_Table.scss
```diff
@@ -132,18 +137,12 @@ $table-dense-padding: $spacer-half $spacer-2 !default;
...
-      content: attr(data-title);
-      display: block;
-      font-weight: bold;
-      padding-bottom: $spacer-half;
-    }
+  // Add column heading for stacked table
+  .ds-c-table--stacked__col-header {
+    display: block;
+    font-weight: bold;
+    padding-bottom: $spacer-half;
```
nichia, line 144: @davidakennedy - the column header text instead of pseudo element fixes the issue of repeated information.

packages/design-system/src/components/Table/TableCell.jsx
```diff
@@ -69,9 +69,14 @@ export const TableCell = ({
       scope={defaultScope}
       headers={headers}
       id={id}
-      data-title={stackedTitle}
       {...tableCellProps}
     >
+      {/* Add column heading for stacked table */}
+      {_stackable && (
```
bernardwang, line 75: The problem with this is that this requires the HTML only examples to mainly add this code, which is quite alot of boilerplate

packages/design-system/src/styles/components/_Table.scss
bernardwang: Same feedback [here](https://github.com/CMSgov/design-system/pull/927/files#r570382699). Lets hold off on addressing that issue in this ticket, we should file a new ticket for that and properly assess the different options before picking a solution

packages/design-system/src/types/Table/Table.d.ts
```diff
@@ -16,9 +16,9 @@ export interface TableProps {
    */
   className?: string;
   /**
-   * Applies the dense variation of the table.
+   * Applies the compact variation of the table.
    */
-  dense?: boolean,
+  compact?: boolean,
```
bernardwang, line 21: Similar to other props that are renamed, can we add a warning for folks using the `dense` prop? Search for `console.warn` for examples

packages/design-system/src/styles/components/_Table.scss
nichia: Ok, have reverted back to using pseudo element. I've changed the <tr> display to `table-row`. Would that help address the `Firefox/NVDA` navigation issue?

nichia: @davidakennedy = Here is an example of [responsive stacked table on hc.gov](https://www.healthcare.gov/third-party-privacy-policies/). The thead is `display: none` when stacked whereas we've just hidden the header display but still made it visible to screen reader. Do you think we should just make it `display: none`?

davidakennedy:
I did a lot of testing on this today. @nichia and I have been talking about the two main issues I brought up in the last PR:

**Pseudo-elements getting announced**: Ultimately, I don't think this is a huge deal. But we could avoid it if we  did this:

```html
<td class="ds-u-text-align--left" role="cell" headers="column_1" data-title="Document title">
  <span aria-hidden="true">Declaration of Sentiments</span>
</td>
```

And putting the pseudo-element on the span. But after further testing, I don't think this is a must do. I'll leave it up to you all as to whether to implement it.

**Firefox/NVDA table navigation not working**: This is when you enter a table element, you should be able to press `CTL + ALT + arrow keys`. On small screen view, this breaks in Firefox. It does work with NDVA on Chrome and Edge. It seems to be related to the use of `display: block` on table elements. See [this bug for NVDA](https://github.com/nvaccess/nvda/issues/9879). This also occurs in the current live version of the table. I missed this when we first tested.

I tested [this example](https://codepen.io/aardrian/pen/LQddwO) from [this Adrian Roselli blog post](https://adrianroselli.com/2018/02/tables-css-display-properties-and-aria.html), which our code is based on, and that works as expected in Firefox/NVDA. So I'm curious what's different? @nichia will take a look.

It would be great if we can fix this. At least it's only affecting Firefox.

bernardwang:
@davidakennedy @nichia I looked into the NVDA bug and the response table example from Adrian Roselli. 

From what I understand from the NVDA bug, setting the `<thead>` to `position: absolute` is what's causing NVDA to mess up, [according to this comment](https://github.com/nvaccess/nvda/issues/9879#issuecomment-576053393) it is the same as using `display: block`, which breaks table semantics. 

Interestingly, Adrian Roselli's table doesnt use `thead` at all, just a regular `<tr role="row">` containing `<th role="columnheader">`. When his table stacks at a small screensize, `display: none` is applied to the row that contains his column headers, so I believe the column is just not announced on a small screen size (unless you press the "add fake cell headers at narrow width" button). 

In our table, we originally used the `ds-u-visibility--screen-reader` utility to visually hide the `thead` element in order to preserve the native behavior of the column being announced. After reading these articles, I think we should try using `display: none` instead. This might fix issues with NVDA, and it would also fix the issue of the column header being repeated twice.

bernardwang (approved):
The `compact` renaming looks good to me! 

Approving, but the issues DK pointed out can be addressed under a different ticket and PR

davidakennedy:
@nichia @bernardwang That's a good analysis. Bernard! Leaving this for the next iteration.

I tested this some more, and one key difference I noticed today after more testing, and what I think is leading to the NVDA table navigation not working:

- One small screen sizes, NVDA announces the table as having one column and one row. Which isn't true, but makes sense as to why NVDA won't work with the up, right or left arrow keys.
- If you navigate on Adrian Roselli's example, it announces the table correctly as having five columns, six rows. So that may be the issue.
- I do think going with `display: none` would be better for the double announcements.

bernardwang:
>One small screen sizes, NVDA announces the table as having one column and one row. Which isn't true, but makes sense as to why NVDA won't work with the up, right or left arrow keys.

Oh hm, I wasnt aware that this was the case... that seems like a different issue altogether and I'm not sure what could be causing this.

davidakennedy:
Something @bernardwang said made me look at the markup again. @nichia I think we need to put `role="rowgroup"` on any `thead`, `tbody` or `tfoot` elements. Because we're building a custom table after breaking the semantics, I think Firefox and NVDA see those as random `div`s inside a table. Untested, but I think that may help us here. See: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Table_Role

line47 (approved):

PR #928, opened 2021-02-04
[WNMGDS-791] Add errorMessageClassname prop

bernardwang:
### Summary 
- Adds `errorMessageClassName` prop to all form components

### How to test
- Ensure props are being passed around correctly
- Use the new prop in a Typescript app

### DIFF
--- a/packages/design-system/src/components/FormLabel/FormLabel.jsx
+++ b/packages/design-system/src/components/FormLabel/FormLabel.jsx
@@ -47,7 +47,11 @@ export class FormLabel extends React.PureComponent {
       }
 
       return (
-        <InlineError id={errorId} inversed={this.props.inversed}>
+        <InlineError
+          id={errorId}
+          inversed={this.props.inversed}
+          className={this.props.errorMessageClassName}
+        >
           {this.props.errorMessage}
         </InlineError>
       );
@@ -65,6 +69,7 @@ export class FormLabel extends React.PureComponent {
       className,
       inversed,
       errorMessage,
+      errorMessageClassName,
       errorId,
       requirementLabel,
       ...labelProps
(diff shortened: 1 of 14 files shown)


packages/design-system/src/components/FormLabel/FormLabel.jsx
```diff
@@ -98,6 +103,10 @@ FormLabel.propTypes = {
   component: PropTypes.oneOf(['label', 'legend']),
   /** Enable the error state by providing an error message. */
   errorMessage: PropTypes.node,
+  /**
+   * Additional classes to be added to the error message
+   */
+  errorMessageClassName: PropTypes.string,
```
nichia, line 109: Add this prop to typescript definition file `FormLabel.d.ts`

nichia:
The `errorMessageClassName` prop works as expected. Just left adding this prop to the typescript definition files.

nichia (approved):
LGTM

packages/design-system/src/types/MonthPicker/MonthPicker.d.ts
```diff
@@ -28,6 +30,14 @@ export interface MonthPickerProps {
...
   label: React.ReactNode;
   errorMessage?: React.ReactNode;
+  /**
+   * Additional classes to be added to the error message
+   */
+  errorMessageClassName?: string;
+  /**
+   * Location of the error message relative to the field input
+   */
+  errorPlacement?: MonthPickerErrorPlacement;
```
bernardwang, line 40: For future reference @nichia  you dont need to define a new type here, you can just use the inline `"top" | "bottom"` value.  Its only setup this way because we used a script to auto generate these typings initially

PR #931, opened 2021-02-08
Allow bottom placed errors in form components

bernardwang:
## Summary
- Adds support for writing core components in Typescript
- Adds  `FormContainer` and `InlineError` components
- Adds `errorPlacement` and `errorMessageClassName` prop
- Update form components to handle bottom placed errors using `errorPlacement` and `FormContainer`
- Updates form components to use `aria-invalid`
- Update `<Autocomplete>` component to handle bottom placed error

## How to test
- Run and build the DS doc site and package
   - http://design-system-demo.s3-website-us-east-1.amazonaws.com/feature/form-container 
- Test the CMSDS on CRA and both child design systems
- Test form component's markup and accessibility in different browsers and assistive tech
- Doubl check the typescript types and proptypes


bernardwang:
@nichia @line47 I think this PR is ready to be reviewed and merged into master. It only adds the ability to change the error placement (which should be backwards compatible), and it doesn't include the feature flags. We'd wanna be pretty careful with testing it, but we also wanna get it into master before our next release, as its not good practice to let big feature branches hang out for a while

nichia (approved):
Great work Bernard. Have tested on Core DS, CRA examples and child DSs & associated apps and they all work as expected.

PR #934, opened 2021-02-09
[WNMGDS-751] Create react feature flags

bernardwang:
## Summary
- Adds `flags.js` to `src/components` with new `ERROR_PLACEMENT_DEFAULT` flag
- Use new flags in form components, remove `Proptype` `defaultProps`
- Fix `react-docgen` error in child design systems without typescript option enabled

## How to test
- Setup and build the HC.gov child DS using the `bernard/feature-flag-test` branch, make sure to install from your local CMSDS branch
- Test usage of the HC.gov child DS form components in the CRA example and in a product app

## TODO:
- The doc site is still unfortunately not rendering the correct error placement, even when using the flag. Still unsure why.


packages/design-system-scripts/gulp/docs/extractReactData/parseReactFile.js
```diff
@@ -107,10 +107,16 @@ function getRelativePath(file, options) {
  */
 function parseComponent(file, options) {
   let docs;
+
   if (file.extname === '.tsx') {
-    // Use `react-docgen-typescript` for `tsx` files
-    docs = tsDocgen.withCustomConfig('./tsconfig.json').parse(file.path);
-    processDocgenTemplates(docs[0], options);
+    // Avoid processing .tsx files unless typescript config is enabled
```
bernardwang, line 112: This fixes an unrelated bug. I could maybe move this to a different PR as well

packages/design-system/src/components/FieldContainer/FieldContainer.tsx
```diff
@@ -105,7 +106,7 @@ export class FieldContainer extends React.Component<FieldContainerProps> {
       className,
       component,
       errorMessage,
-      errorPlacement,
+      errorPlacement = errorPlacementDefault(),
```
bernardwang, line 109: The flag is used by form components in `FieldContainer`. Dropdown and TextField are the exceptions, they also use the flag because they are also needed inside the form input components (i.e. Select and TextInput)

nichia (approved):
The flags seems to work as expected on core, CRA and product apps. 
The feature flags design is great, the code looks good, strange that somehow the flag cannot get updated on hcgov ds for the docs site.

PR #935, opened 2021-02-10
[WNMGDS-810] Update stacked table for screen reader

nichia:
## Summary
- Add role="rowgroup" for `<thead>` and `<tbody>`
- Remove display:none for stacked table `<thead>`

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/ni/table-screen-reader


nichia:
@davidakennedy - created this test PR to continue testing of possible solution for the NVDA/Firefox navigation issue.
Let me know if this change work or not.

davidakennedy:
@nichia Still getting the same results. Leave this though, and let me do some more testing. In testing this one, and the previous PR, I noticed NVDA announces the table as having 1 column and two rows, which isn't correct. I think this is part of our problem. In testing the [Adrian Roselli example](https://codepen.io/aardrian/pen/LQddwO), NVDA doesn't do that.

It doesn't look like we do anything with the roles and table structure on small screens besides modifying with CSS, right?

nichia:
Yes, small screens is just updated with CSS

davidakennedy:
@nichia I did a lot of further testing on this today. Overall, I think we should move forward with what's here because it puts the component in line with Adrian Roselli's example and adding the `role="rowgroup"` doesn't harm anything, but doesn't seem to improve anything either.

Firefox and NVDA seem to be buggy with how it currently handles responsive tables, at least compared to other browser/screen reader combos. I'm okay moving forward with this because even though table navigation is wonky, you can still move through the content with the arrow keys normally. Perhaps we should make a note of this somewhere as a known issue?

I tried a number of things to see if I could improve the table navigation in Firefox/NVDA. This included:

- Removing `thead` and using only `tr`.
- Removing the `scope="col"` on the columns. Roselli's example didn't have them, so I wondered if that caused an issue.
- I also testing the above with `display: none` and visually hiding the `thead`.
- I downloaded an updated version of NVDA and tested everything.
- I also retested Roselli's example and got the same results, with it actually reporting 0 columns and 0 rows. So I do think it maybe relates to [how tables are handled](https://github.com/nvaccess/nvda/issues/9879), like I mentioned previously.

nichia:
Thanks @davidakennedy for further testing. Good suggestion on capturing this information, I'll add your comments to the code file so we have a note of the issue and what's been tried.

bernardwang:
@nichia I see commits from this PR https://github.com/CMSgov/design-system/pull/927 in the commit history here. Can you please clean up the commit history so it only contains the relevant commits?

nichia:
Hi @bernardwang - i rechecked the log on the final commit of PR #927 (commit 69a61bd3) and it doesn't include commits from this test pr. Let me know if I may have overlooked the issue.

nichia:
@bernardwang - ready for reviews - have cleaned up the commits.

packages/design-system-docs/src/pages/components/Table/Table.example.jsx
```diff
@@ -47,7 +47,10 @@ ReactDOM.render(
         </TableRow>
       </TableBody>
     </Table>
-
+    {/* Firefox and NVDA seem to be buggy with how it currently handles responsive stacked tables, at least compared to other browser/screen reader combos.
```
bernardwang, line 50: Thanks for adding this comment in!

packages/design-system/src/components/Table/TableBody.jsx
```diff
@@ -15,7 +15,12 @@ export const TableBody = ({ children, _stackable, ...tableBodyProps }) => {
     });
   };
 
-  return <tbody {...tableBodyProps}>{_stackable ? renderChildren() : children}</tbody>;
+  /* eslint-disable jsx-a11y/no-redundant-roles */
```
bernardwang, line 18: Could you add a comment for reenabling this eslint rule after this block of code? Generally if you use `eslint-disable` you want to also use `eslint-enable`

bernardwang:
@nichia I noticed you changed the base branch to `WNMGDS-781/rename-dense-table`, this isnt what we want to do to clean up the commit history. This PR is supposed to merge into `master` so changing the base branch just means your work isnt going to be incorporated into our releases and work. That branch was already merged into master, we generally dont reuse branches unless its a larger feature branch that isnt ready to merge into master yet

packages/design-system/src/components/Table/TableBody.jsx
```diff
@@ -15,7 +15,13 @@ export const TableBody = ({ children, _stackable, ...tableBodyProps }) => {
...
+  return (
+    <tbody role="rowgroup" {...tableBodyProps}>
```
line47, line 20: Should we remove `rowgroup`?  
@davidakennedy mentioned it doesn't add any value so is there another reason to keep it?
> and adding the role="rowgroup" doesn't harm anything, but doesn't seem to improve anything either.

davidakennedy: My thought would be to leave it since screen readers/assistive tech may do something with it in the future. That said, I don't have strong feelings, so okay if you all want to remove them.

bernardwang (approved):
this looks good to me! nice to remove the duplicate header being announced

PR #941, opened 2021-02-18
[NO-TICKET] Direct import of lodash

nichia:
## Summary
HCgov DS docs site displays blank screen for Header and Footer when we don't use the direct import of lodash.

See [this PR](https://github.com/CMSgov/design-system/pull/917) for similar fix.

### How to test
- Setup HCgov child design system in the package.json to reference this branch with `file:../design-system/package/design-system/`
- Run `git clean -fdx` and then `yarn install` to Install the local DS branch
- Test that the Header and Footer component pages are displaying correctly


nichia:
Test snapshot differences occurs as a result of this change. @bernardwang would you take a look at it and see if that's an issue or we can just update the snapshots?

bernardwang (approved):
@nichia Nice catch, this looks good after you update the snaps

PR #946, opened 2021-02-22
[WNMGDS-753] Update styles for focus buttons

line47:
## Summary
- Set the focus styles to use the same text and background color as the hover styles for each button variation when the focus styles are not implemented 

### Changed
- Changed variable values to use the correct colors for focus buttons

## How to test
- Run the doc site locally and see that the focus styles are visible when you tab through the buttons
- Set the focus style flag to true to make sure when focus styles are set they work as expected


packages/design-system/src/styles/components/_Button.scss
```diff
@@ -300,10 +304,12 @@ $button-color-variants: (
       &.ds-c-button--active:focus {
         background-color: nth($color, 3);
         box-shadow: none;
+        color: $color-white;
       }
     } @else {
       &:focus,
       &.ds-c-button--focus {
+        background-color: nth($color, 2);
```
line47, line 312: Adding background color for focus state to match hover state

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -300,10 +304,12 @@ $button-color-variants: (
       &.ds-c-button--active:focus {
         background-color: nth($color, 3);
         box-shadow: none;
+        color: $color-white;
```
line47, line 307: Add text color for focus state to match hover state

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -182,21 +181,26 @@
...
   &:active,
   &.ds-c-button--disabled,
-  &.ds-c-button--hover,
   &.ds-c-button--active {
     border-color: transparent;
+    background-color: transparent;
+  }
+  &:hover,
+  &.ds-c-button--hover {
+    color: rgba($color-base-inverse, 0.8);
```
line47, line 194: Needed unique hover styles for transparent inverse buttons

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -182,21 +181,26 @@
...
+  }
+  // Default inverse focus styles are overridden above
+  @if $ds-include-focus-styles {
```
line47, line 197: Moved focus styles below to make sure order was correct

davidakennedy (approved):
Hi @line47, this looks good. I tested in a number of browsers on Mac and Windows, and didn't notice any issues with the default styles or the beta focus styles. Everything looks as it should.

I did notice one small inconsistency with the transparent buttons on inverse backgrounds with the beta styles enabled. See the GIF. In the example, you can view focus state on an active button for a transparent button on an inverse background. It's the last one in the example in the GIF. Likely not an issue with this PR and something minor, but wanted to flag it for consistency sake. I didn't remember if it was left like that for a reason.

Screen cast cycling through buttons with the keyboard.

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -27,7 +27,7 @@
   &.ds-c-button--focus {
     @if $ds-include-focus-styles {
       background-color: $focus-color;
-      border-color: $color-base;
+      border-color: $color-primary-darker;
```
bernardwang, line 30: Isnt this updating the styling for when the global focus style flag is on? I dont think we need to be updating that for this PR right?

line47: Yep that's right - changing it back. Thanks for the catch

packages/design-system/src/styles/components/_Button.scss
bernardwang: I'm having trouble understanding why this change is necessary, or what it's accomplishing. Dont the transparent inverse buttons already have hover and disabled states?

bernardwang: do you have a screenshot to compare the changes here?

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -182,21 +182,37 @@
...
+  &.ds-c-button--hover {
+    color: rgba($color-base-inverse, 0.8);
+  }
+
   // Default inverse focus styles are overridden above
   @if $ds-include-focus-styles {
     &:focus,
     &.ds-c-button--focus {
       border-color: transparent;
+      background-color: $focus-color;
```
bernardwang, line 201: Same here, isnt this updating the styles for the beta focus styles? Why is this being changed

bernardwang:
I added a commit that rearranges the order of the focus and active styles. Previously you weren able to achieve an active style when it was also focused

packages/design-system/src/styles/components/_Button.scss
bernardwang: Oh is this to address DK's commnt here? https://github.com/CMSgov/design-system/pull/946#pullrequestreview-596579401

packages/design-system/src/styles/components/_Button.scss
bernardwang: oh ok I think I understand, is this addition necessary because we updated the ordering for active to come after focus for the transparent inverse button?

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -182,21 +182,38 @@
 %transparent-inverse-button {
   border-color: transparent;
   background-color: transparent;
+
+  &:disabled,
```
bernardwang, line 186: I think disabled styles should be last, going after `active`

packages/design-system/src/styles/components/_Button.scss
line47: I was switching between the focus styles on and off in the PR to make sure i did not cause any regressions and noticed an issue with the transparent inverse button so decided to put it in this PR. 

**I can back those out of this PR and put them into a different one if you like** 

Screenshot is what the issue was 
Screen Shot 2021-02-24 at 1 34 20 PM

packages/design-system/src/styles/components/_Button.scss
bernardwang: Ok, I see that this is probably feedback to DK's comment https://github.com/CMSgov/design-system/pull/946#pullrequestreview-596579401.

However, it would still help if you could walk me through why we need to add extra disabled and hover styles here

line47: After going back though this They aren't needed becasue they are already defined above, thanks!

packages/design-system/src/styles/components/_Button.scss
line47: Moved them last

bernardwang:
@line47 are you good with closing this PR and updating the ticket with the context behind why we cant easily make this change, and our plan to fix it after taking on focus styles work?

nichia (approved):
Similar to DK, the buttons looks good and not sure about the inconsistent `Active` focus highlight for the 2nd inverse example.

line47:
Yeah - I'll close it out. Working on some focus and button alignment in a Mural to share with Medicare and Healthcare

line47:
Opened this PR back up and believe I have the Mgov styles worked out as well https://github.com/CMSgov/mgov-design-system/pull/16

packages/design-system/src/styles/components/_Button.scss
bernardwang: no this is a great catch thank you

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -182,22 +182,25 @@
...
     }
   }
 
-  &:disabled,
-  &:hover,
-  &:active,
-  &.ds-c-button--disabled,
-  &.ds-c-button--hover,
-  &.ds-c-button--active {
-    border-color: transparent;
```
bernardwang, line 199: Why was this deleted? And it looks like there's a leftover comment here

bernardwang (approved):
I've tested the changes in this PR and in Mgov, they seem good to me! Great work on the MDS overrides @line47! I also had a few comments left in this PR but for the most part it looks ready to go!

packages/design-system/src/styles/components/_Button.scss
line47: This was removed because the transparent border is already declared up on line 183 for the transparent inverse buttons so it was not needed here.

PR #949, opened 2021-02-24
[WNMGDS-575] Beta tealium integration for Helpdrawer

nichia:
## Summary
Implement beta tealium/google analytics integration for the Helpdrawer component

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-575/beta-tealium-for-helpdrawer/
Not sure why the demo link doesn't work(network tab not picking up `collect` tx), I'm looking into it. For now, clone the branch to local system for testing.
- Go to `Help Drawer` component, the React example and click `Toggle the helpdrawer` to open help drawer.
- Inspect DevTools network tab and filter for `collect` to view the analytics data
- `Close` help drawer and inspect DevTools network tab to view the analytics data again


packages/design-system-scripts/package.json
```diff
@@ -15,6 +15,7 @@
     "@babel/core": "^7.8.4",
     "@babel/plugin-proposal-class-properties": "^7.12.1",
     "@babel/plugin-transform-object-assign": "^7.10.4",
+    "@babel/plugin-transform-runtime": "^7.13.6",
```
bernardwang, line 18: Why is this dependency added or needed?

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -24,6 +25,53 @@ export class HelpDrawer extends React.PureComponent {
...
+
+  componentWillUnmount() {
+    const eventAction = 'onComponentWillUnmount';
+
+    /* Send analytics event for helpdrawer close */
+    this.sendAnalyticsEvent(eventAction);
+  }
+
+  sendAnalyticsEvent(eventAction) {
+    const eventProps = this.props.analytics && this.props.analytics[eventAction];
```
bernardwang, line 43: you can use `lodash.get` here instead

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -24,6 +25,53 @@ export class HelpDrawer extends React.PureComponent {
...
+    const sendAnalytics =
```
bernardwang, line 44: can this be simplified?
 `const sendAnalytics = !(eventProps === false)`

bernardwang: even more concise would be:
```
const analyticsDiabled = get(this.props.analytics, eventAction) === false

if (window.utag && !analyticsDisabled)
```

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -24,6 +25,53 @@ export class HelpDrawer extends React.PureComponent {
...
+      typeof eventProps !== 'boolean' || (typeof eventProps === 'boolean' && eventProps !== false);
+
+    if (window.utag && sendAnalytics) {
+      sendTealiumEvent(this.getAnalyticsPayload(eventAction));
+    }
+  }
+
+  getAnalyticsPayload(eventAction) {
```
bernardwang, line 52: I dont think there needs to be another function here. Since you already defined a `sendAnalyticsEvent` helper function, i think it's fine to include this code there. This is a more concise example:
 
```
const analyticsOverrides= get(this.props.analytics, eventAction);
const analyticsDisabled = analyticsObject === false;

if (window.utag && !analyticsDisabled) {
    const payload = merge(this.getDefaultAnalytics(eventAction), analyticsOverrides));
    sendTealiumEvent(payload);
}
```

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -88,9 +136,23 @@ HelpDrawer.defaultProps = {
...
+  ga_eventAction: PropTypes.string,
+  ga_eventLabel: PropTypes.string,
+  ga_eventValue: PropTypes.string,
+});
+
 // TODO: closeButtonText, title/heading should be a string, but it is being used as a node in MCT,
 // until we provide a better solution for customization, we type it as a node.
 HelpDrawer.propTypes = {
+  analytics: PropTypes.shape({
+    onComponentDidMount: PropTypes.oneOfType([PropTypes.bool, AnalyticsEventShape]),
```
bernardwang, line 151: nice job here!

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -88,9 +136,23 @@ HelpDrawer.defaultProps = {
...
+    onComponentWillUnmount: PropTypes.oneOfType([PropTypes.bool, AnalyticsEventShape]),
+  }),
+
+  // - anlalytics onClose: false (boolean or object)  - turn off/on
```
bernardwang, line 155: leftover comment

packages/design-system-scripts/package.json
nichia: The use of `async` for `export async function sendTealiumEvent` is giving referenceError of `regeneratorRuntime is not defined`, and this plugin fixes that. Is there another package that can address this issue? Or alternative is that we don't use async, just regular send and re-try like hrtools.

bernardwang: You shouldn't need an additional plugin for this. If you search for `async` in our codebase you can see it's already being used in a bunch of different places like in e2e tests. If you want to use `async` you just need to include `import "regenerator-runtime/runtime";` (i.e.  in `e2e.global-setup.js` which is run before each e2e test). You can see [this stackoverflow](https://stackoverflow.com/a/61517521) for more info

bernardwang: That being said, I dont know if we want to be introducing `regenerator-runtime/runtime` as a dependency to our library. Since we already use `core-js`, it might be better to use promises instead of `async` and just include a `core-js` polyfill for promise support. You can see Tooltips as an example of this

packages/design-system-scripts/gulp/docs/generatePages/generateReactExample.js
```diff
@@ -35,6 +35,7 @@ function generateReactExample(page, docsPath, sourceDir, { typescript, rootPath
       const head = `
         <title>Example: ${page.reference}</title>
         <link rel="stylesheet" href="/${path.join(rootPath, 'example.css')}" />
+        ${core ? analytics() : ''}
```
bernardwang, line 38: This is purely for testing right?

nichia: I thinking we monitor how ga tracks this for now. If users click `New tab` to open the iframe, I think it'd be good to track the page usage.

bernardwang: Hm, I'm not sure if thats necessary. They are just code examples, we dont need to track users using them.

bernardwang: Isnt it possible to manually test the tealium events by simply pasting the tealium script on the HelpDrawer code example page?

nichia: Ok, will check that out.

nichia:
@isawpalmetto @LouisFettet - thanks for providing the base for this PR's code. I've added you as reviewer to help look for gaps for improvement.

line47:
I was able to test this locally and did see the collection in the network dev tools. 
Are there specifics I should be looking for within the collection information to know that this is working as expected?

nichia:
Thanks @line47 checking for the network tab for the tx is about it for now.

LouisFettet (commented):
Everything here looks good at a glance; I have a couple questions about the larger strategy here, so forgive my ignorance:
1. Is the idea that we would expose the `sendTealiumEvent` function to the products using the DS? I think this would be great to cut down on very repetitive boilerplate we use across our apps.
2. Do we always want the `ga_eventType` for help drawer events to be `'cmsds'`? I noticed that it looks like a developer can override the `ga_eventCategory`, `ga_eventAction`, `ga_eventLabel`, and `ga_eventValue` using the `analytics` prop, but it doesn't look like the `ga_eventType` is part of the proptypes.

nichia:
Thanks for the review:
1. Yes, exposing `sendTealiumEvent` function would be good to reduce repetitive boilerplate. I can update to do that.
2. Good pick up, i'll update the proptypes. We don't have the fixed data definitions for GA4 yet, but `ga-eventType` and any other data that is customizable should be part of the proptypes.

packages/design-system-scripts/gulp/docs/generatePages/generateHtmlExample.js
```diff
@@ -57,4 +59,19 @@ function generateHtmlExample(page, modifier, docsPath, options) {
   );
 }
 
+/**
+ * Blast Analytics code to be included in the <head>.
+ * This loads additional tracking scripts, like Google Analytics.
+ * @return {String}
+ */
+function analytics() {
```
bernardwang, line 67: Small suggestion, since this is identical to the code in `generateDocPage`, can we just import this function from that file?

packages/design-system-scripts/gulp/docs/generatePages/generateReactExample.js
```diff
@@ -61,4 +68,19 @@ async function generateReactExample(page, docsPath, sourceDir, docsDir, { typesc
   });
 }
 
+/**
+ * Blast Analytics code to be included in the <head>.
+ * This loads additional tracking scripts, like Google Analytics.
+ * @return {String}
+ */
+function analytics() {
```
bernardwang, line 76: same feedback as here: https://github.com/CMSgov/design-system/pull/949/files#r597075906

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -24,6 +26,45 @@ export class HelpDrawer extends React.PureComponent {
 
   componentDidMount() {
     if (this.headingRef) this.headingRef.focus();
+
+    const eventAction = 'onComponentDidMount';
```
bernardwang, line 30: This could be condensed to:
`this.sendAnalyticsEvent('onComponentDidMount')`
This is just my personal preference, so I don't feel strongly on it at all

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -24,6 +26,45 @@ export class HelpDrawer extends React.PureComponent {
...
+
+  componentWillUnmount() {
+    const eventAction = 'onComponentWillUnmount';
+
+    /* Send analytics event for helpdrawer close */
+    this.sendAnalyticsEvent(eventAction);
+  }
+
+  sendAnalyticsEvent(eventAction) {
+    const analyticsOverrides = get(this.props.analytics, eventAction);
```
bernardwang, line 44: Instead of exporting `sendTealiumEvent`, I wonder if we could export this entire `sendAnalyticsEvent` function as a helper. This way the boilerplate in components could be reduced even more. I.e.

```
sendAnalyticsEvent(overrides, defaultPayload) {
  const analyticsDisabled = overrides === false;
  
  if (window.utag && !analyticsDisabled) {
      const payload = merge(payload, analyticsOverrides);
      sendTealiumEvent(payload);
    }
}
```

And then usage would just be:
```
componentWillUnmount() {
    const eventAction = 'onComponentWillUnmount';

    /* Send analytics event for helpdrawer close */
    this.sendAnalyticsEvent(get(this.props.analytics, eventAction), get(DefaultAnalytics, eventAction));
  }
  ```

File: packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx [hunk 12]
```diff
@@ -24,6 +26,45 @@ export class HelpDrawer extends React.PureComponent {
...
+    const analyticsDisabled = analyticsOverrides === false;
+
+    if (window.utag && !analyticsDisabled) {
+      const payload = merge(this.getDefaultAnalytics(eventAction), analyticsOverrides);
+      sendTealiumEvent(payload);
+    }
+  }
+
+  getDefaultAnalytics(eventAction) {
```
bernardwang, line 53: This doesnt have to be a component method, it can also just be an object. I recommend defining this at the top of the file like so:
```
const defaultAnalytics = {
   ....
}
```

This way usage can just be `get(defaultAnalytics, eventAction)`

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -88,9 +129,32 @@ HelpDrawer.defaultProps = {
   closeButtonText: 'Close',
   headingLevel: '3',
 };
+
+/**
+ * Defines the shape of an analytics event for tracking that is an object with key-value pairs
+ */
+const AnalyticsEventShape = PropTypes.shape({
```
bernardwang, line 136: this makes sense here for now, but eventually this should be exported and shared across different components using Analytics, and ideally this would be typed with Typescript

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -88,9 +129,32 @@ HelpDrawer.defaultProps = {
...
+  ga_eventAction: PropTypes.string,
+  ga_eventLabel: PropTypes.string,
+  ga_eventValue: PropTypes.string,
+});
+
 // TODO: closeButtonText, title/heading should be a string, but it is being used as a node in MCT,
 // until we provide a better solution for customization, we type it as a node.
 HelpDrawer.propTypes = {
+  /**
+   * Analytics events tracking is enable by default.
```
bernardwang, line 147:
```suggestion
   * Analytics events tracking is enabled by default.
```

packages/design-system/src/components/analytics/Analytcs.test.ts
```diff
@@ -0,0 +1,103 @@
+import { sendTealiumEvent } from './Analytics';
```
bernardwang, line 1: I think this file is misspelled

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
bernardwang: Oh nice, I see you already have a Typescript definition for this already defined in Analytics.ts

packages/design-system/src/components/analytics/Analytics.ts
```diff
@@ -0,0 +1,94 @@
...
+  additional_props?: Record<string, unknown>;
+}
+
+function sendEvent(event: EventType, props: AnalyticsPayload, retry = 0): void {
+  if (window.utag && window.utag[event]) {
+    try {
+      window.utag[event](props);
+    } catch (e) {
+      // eslint-disable-next-line no-console
+      console.log('Error sending event to Tealium', e);
```
bernardwang, line 56: We want to make sure this doesn't get logged out in production.

nichia: Updated to return string instead of console.log

line47 (approved):
This is working for me when i test locally!

PR #952, opened 2021-02-25
[WNMGDS-819] Allow doc site to use the feature flag

bernardwang:
## Summary
- The doc site was previously not working with the feature flag because the example files (imported from `@cmsgov/design-system-docs`) were importing from `@cmsgov/design-system`, rather than the local source directory in the case of the child DS doc site. 
- To solve this I added a new `@design-system` webpack alias to be used in the example files, so when the example files are used in a child DS doc site, they import from the child DS source (which is using the error placement feature flag).
- The old `@src` alias was also removed 

### How to test 
- Setup and build the HC.gov child DS using the `bernard/feature-flag-test` branch, make sure to install from your local CMSDS branch
- Test usage of the HC.gov child DS form components in the CRA example and in a product app
- Test the child design system example

### DIFF
--- a/examples/child-design-system/docs/src/pages/components/Button/Button.example.jsx
+++ b/examples/child-design-system/docs/src/pages/components/Button/Button.example.jsx
@@ -4,9 +4,9 @@
 // https://webpack.js.org/configuration/resolve/#resolvealias
 //
 // Because the react example files are located separately from the design system source directory (`sourceDir` in `cmsds.config.js`),
-// We provide the `@src` alias to allow for easy imports from `sourceDir`
+// We provide the `@design-system` alias to allow for easy imports from `sourceDir`
 // This is also possible in typescript child design systems via the `paths` compiler option in the `tsconfig.json`
-import Button from '@src/components/Button/Button';
+import { Button } from '@design-system';
 import PropTypes from 'prop-types';
 import React from 'react';
 import ReactDOM from 'react-dom';
--- a/examples/child-design-system/docs/src/pages/components/Card/Card.example.jsx
+++ b/examples/child-design-system/docs/src/pages/components/Card/Card.example.jsx
@@ -2,9 +2,9 @@
 // https://webpack.js.org/configuration/resolve/#resolvealias
 //
 // Because the react example files are located separately from the design system source directory (`sourceDir` in `cmsds.config.js`),
-// We provide the `@src` alias to allow for easy imports from `sourceDir`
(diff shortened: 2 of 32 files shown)


nichia (approved):
Great work Bernard, the feature flag works core, child ds and the product that references the child ds.
Just curious for reason in maintaining the `@cmsds/design-system` instead of `@design-system` for the docs generation scripts `packages/design-system-docs/src/scripts/`?

bernardwang:
> Just curious for reason in maintaining the `@cmsds/design-system` instead of `@design-system` for the docs generation scripts `packages/design-system-docs/src/scripts/`?

That's a good point, we could totally do that. I didn't include at first because we arent using many components from `@cmsds/design-system`, but thinking about it again, if we replaced it with `@design-system` we maybe could remove the `@cmsds/design-system` dependency

bernardwang:
@nichia Actually, we still need the `@cmsds/design-system` dependency for the doc site styles, so I'm just going to leave that out of this PR

line47:
I'm not sure how i set the feature flag in the HCgov child design system. Does that happen in the cmsds.config.file?

bernardwang:
@line47 its set inside `src/index.js`. You can see it in the `bernard/feature-flag-test` branch on the hcgov ds

line47:
I noticed the autocomplete for a HCgov child design system doesn't appear to be getting the same styles as the core and the The clear search button is lower and the autocomplete box shows below the error instead of below the text input box 

**Core screenshot**
Screen Shot 2021-02-26 at 3 16 50 PM
**HC gov screenshots** 
Screen Shot 2021-02-26 at 3 18 58 PM
Screen Shot 2021-02-26 at 3 21 26 PM

bernardwang:
@line47 are these screenshots both from your local doc site running? Can you confirm that the HCgov DS is using the correct version of the core?

line47:
Additionally I found with the autocomplete we will need to adjust this style because when the icon is present with the error message clear search button is pushed down. 
```
.ds-c-autocomplete__error-message--clear-btn {
    width: calc(100% - 100px);
}
```

If we change increase the width to this `-110px` for the error message and clear button both sit on the same line. 
I'm thinking it would be best to update the core styles. 
```
.ds-c-autocomplete__error-message--clear-btn {
    width: calc(100% - 110px);
}
```

In the future it might be good to revisit the structure of this component and use some sort of flex/grid option for laying it out.

line47 (approved):
Approve but also noting that we might need to make some style adjustment at the core or hcgov level

PR #961, opened 2021-03-08
[NO_TICKET] Reimplement feature flags with env vars

bernardwang:
## Summary
The old feature flag implementation which was based off a flag variable wasn't working when building for production. This PR replaces that implementation with a `flags` property inside `cmsds.config.js`, which is copied over to `process.env` and used inside the components on build time. This means the HCgov design systems doesnt have to use the `setErrorPlacement` flag helper anymore, and can just set the `flags` property inside `cmsds.config.js`

## How to test
- Build the CMSDS DS with `flags.ERROR_PLACEMENT_DEFAULT` set to `top`, `bottom`, and undefined. Make sure the error placements work as expected, they should default to `top` if the flag is not defined, and use the value if otherwise provided.
- Test the flags when building for production. This can be achieved by running the following scripts:
```
yarn build --ignoreRootPath
npx http-server ./packages/design-system-docs/dist
```
- Test the flags on HCgov DS and product applications

### DIFF
--- a/babel.config.js
+++ b/babel.config.js
@@ -19,6 +19,12 @@ module.exports = function (api) {
     '@babel/plugin-proposal-class-properties',
     '@babel/plugin-transform-object-assign',
     'inline-react-svg',
+    [
+      'transform-inline-environment-variables',
+      {
+        include: ['CMSDS_FLAGS'],
+      },
+    ],
   ];
 
   return {
--- a/cmsds.config.js
+++ b/cmsds.config.js
@@ -18,8 +18,9 @@ module.exports = {
   githubUrl: 'https://github.com/CMSgov/design-system',
   // The name of your design system NPM package. This replaces the {{npm}} template in documentation content.'
   npmPackage: '@cmsgov/design-system',
+
   // React feature flag variables
   flags: {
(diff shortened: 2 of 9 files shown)


line47:
When i try and run the Healthcare.gov design system I get this error

```
WARNING in ./dist/esnext/index.esm.js 14:0-24
"export 'setErrorPlacementDefault' was not found in '@cmsgov/design-system'
```


I added this to cmsds.config.js
```
 flags: {
    ERROR_PLACEMENT_DEFAULT: 'bottom',
  },
```

nichia:
Core ds works well with the env flags. On Hcgov ds, it doesn't seem to reference the env flags set as it's using the same error placement value as what's set on core ds.

@line47 - deleting the lines with `setErrorPlacementDefault` from hcgov ds `src/components/index.js` would resolve the error. I also added the same lines you had for hcgov ds `cmsds.config.js`.

bernardwang:
@nichia @line47 what both of you described is expected, the HCgov DS has to be updated with this PR to remove `setErrorPlacementDefault` in favor of the `cmsds.config.js`

PR #964, opened 2021-03-09
[WNMGDS-41] Create draft accordion component

line47:
## Summary
- This PR adds the accordion component to the design system
- There are 2 style variations `no border` which is the default and `border`
- Accordions can be set to expanded by default on page load

Demo: http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-41-Accordion/components/accordion/
## How to test
- Run this branch locally and test the different props for the `<Accordion>` and `<AccordionItem>`


packages/design-system/src/components/Accordion/Accordion.tsx
```diff
@@ -0,0 +1,24 @@
...
+}
+
+const Accordion: FunctionComponent<{
+  items: AccordionItemData[];
+  className?: string;
+  bordered?: boolean;
+}> = ({ items, className, bordered = false }) => {
+  return (
+    <div className={`ds-c-accordion ${className} ${bordered ? 'ds-c-accordion--bordered' : ''}`}>
+      {items.map((i) => (
```
bernardwang, line 17: Instead of using this pattern of `items`, I think we should move to a model similar to `<Table>`, where the user passes in `AccordianSummary` and `AccordianDetails`, similar to how its implemented here in material ui
https://material-ui.com/components/accordion/

bernardwang: While its a little more markup for users, they also can have more control over the accordian markup and styles, i.e. 

The downside is that the markup above works well for accordions that work individually of each other (i.e. multi-select), but not so much for single select. It would require the user to manage state, which may or may not be a bad thing

nichia:
Have made some updates to include Accordion accessibility by applying WAI-ARIA as per [w3.org recommendation](https://www.w3.org/TR/2019/NOTE-wai-aria-practices-1.1-20190814/#accordion)

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+   */
+  headingClassName?: string;
+  /**
+   * Boolean to expand the accordion.
+   */
+  expanded?: boolean;
+  /**
+   * Text for the accordion item heading.
+   */
+  heading: React.ReactNode | string;
```
bernardwang, line 22: We want to include a `headingLevel` prop here so folks can chose to use something other than an `h2`. See Alert or HelpDrawer for an example of that

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,67 @@
+import React, { FunctionComponent, useState } from 'react';
+import classNames from 'classnames';
+import uniqueId from 'lodash.uniqueid';
+
+interface AccordionItemProps {
```
bernardwang, line 5: We want to make sure our typing for props are exported

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+const AccordionItem: FunctionComponent<AccordionItemProps> = ({
+  children,
+  contentClassName,
+  headingClassName,
+  expanded,
+  heading,
+  id,
+}) => {
+  // Set the state for opening and closing an accordion item
+  const [open, setOpen] = useState(!!expanded);
```
bernardwang, line 38: I dont think we can use React Hooks just yet, there are certain apps that are on older versions of react

bernardwang: @line47 @nichia 
This can be handled in a follow up, but I think we may need to think more on how we handle states in this component. The way it's currently setup only allows for the "multiselect" option, when ideally we can support the single select by default. We probably also want to allow users to manually control individual accordion items, or at least get access to the click events via an event handler (similar to `onOpen` and `onClose` in Tooltips).

There are a few different options here:
1. Make `AccordionItem` strictly a controlled component by adding required `open` and `onClick` props. Then, make the `Accordion` component handle the state for each `AccordionItem` by default, and allow users to select the Accordion behavior, either "single select", "multi select", or "uncontrolled". This allows us to support single/multi select out of the box, while also providing an uncontrolled option if users need custom behavior
2. We continue to handle state inside `AccordionItem` by default, but we add optional `open` and `onClick` props. This way teams that need single select can manually achieve this behavior by overriding the default behavior (similar to `value` and `defaultValue` in form components). The downside to this is that teams need to do more work to achieve a single select behavior, which is the default for USWDS.

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  const contentClasses = classNames('ds-c-accordion__content', contentClassName);
+  const headingClasses = classNames('ds-c-accordion__heading', headingClassName);
+  const contentId = id || uniqueId('accordionItem_');
+  const buttonId = `${contentId}-button`;
+
+  return (
+    <>
+      <h2 className={headingClasses}>
+        <button
+          className="ds-c-accordion__button"
```
bernardwang, line 51: I could see product teams needing to override the accordion button style, could we add a `buttonClassName` or something?

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  expanded,
+  heading,
+  id,
+}) => {
+  // Set the state for opening and closing an accordion item
+  const [open, setOpen] = useState(!!expanded);
+  const onClick = () => {
+    setOpen(!open);
+  };
+  const contentClasses = classNames('ds-c-accordion__content', contentClassName);
```
bernardwang, line 42: Since `children` is being passed through to the accordion content, can we just name this `className` instead of `contentClassName`?

packages/design-system/src/components/Accordion/AccordionItem.tsx
bernardwang: Ah, I realized there's already an `expanded` prop. However, that is only used on initializing the component, so we still need to handle the issue i mentioned above

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,67 @@
...
+  heading: React.ReactNode | string;
+  /**
+   *  If not provided, a unique id will be automatically generated and used.
+   */
+  id?: string;
+}
+
+const AccordionItem: FunctionComponent<AccordionItemProps> = ({
+  children,
+  contentClassName,
```
nichia, line 31: It's strange that github actions shows these `check failures` and yet the PR build shows `All checks have passed`

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,88 @@
...
+  const contentClasses = classNames('ds-c-accordion__content', contentClassName);
+  const buttonClasses = classNames('ds-c-accordion__button', buttonClassName);
+  const contentId = id || uniqueId('accordionItem_');
+  const buttonId = `${contentId}-button`;
+  const HeadingTag = `h${headingLevel}` || 'h2';
+  console.log(HeadingTag);
+  if (heading) {
+    return (
+      <>
+        <HeadingTag className="ds-c-accordion__heading">
```
line47, line 59: I'm getting an error here `Type '{ children: Element; className: string; }' is not assignable to type 'IntrinsicAttributes'.` that seems to be coming from `const HeadingTag = `h${headingLevel}` || 'h2';` and I'm not sure how to fix that ATM

packages/design-system/src/components/Accordion/AccordionItem.tsx
line47: I think we want to have a unique className for the individual content area div as well as the div container that wraps all the accordions.

davidakennedy:
Hey all! This works really well overall. I only found one major issue, not accessibility related. And another issue that may need more exploration, plus some questions.

I did our usual set of testing on this:

- Ran Axe, Wave and Arc Toolkit accessibility checkers on it.
- Tested with a keyboard.
- Tested for color contrast and color blindness.
- Completed Zoom testing in the browser.
- Ran it through JAWS/Edge, JAWS/Chrome, NVDA/Firefox and VoiceOver/Safari. I need to still run through JAWS/IE 11 (see below).

## Issues that need addressed

- In IE 11, I get an error. It looks like something in the build process needs to be configured to support IE11: `Object doesn't support property or method: forEach`.

## Issues that may need further testing/iteration

- Something [pointed out on the Gov.uk accordion](https://design-system.service.gov.uk/components/accordion/): "The plus and minus icon is on the right side of the component, which means users of screen magnifiers might not see it." This could be an issue for our accordion as well. Maybe we explore that further?

## General questions and comments

- Keyboard support doesn't include up/down arrows to move to next/previous accordion. Or Home/End keys. See: https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html This may not be useful for all instances of this component since maybe there's only one on a page. But maybe we give users a way to enable this as needed?
- Is there any CSS that could make adding focus styles to this more difficult? I don't think so after a quick look, but asking in case.
- Noting that many accessible accordion examples include a `role="region"` on the content container. It can be used by screen readers to navigate to the content more easily. I don't think we need to do that here. Headings should provide the same functionality.
- I wonder if it would be helpful to include functionality to add a "Expand all/Collapse all" buttons, especially if many accordions are used on a page. For example, like a FAQ page. Perhaps that's something for the future.
- Should we use `visibility: hidden` in CSS instead of the `hidden` HTML attribute to toggle the visibility? This opens the door to light animation in the future. I wasn't sure if locking one in now would mean it would be harder to change later.

Other accordions I looked at while reviewing this one:

- https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html
- https://design-system.service.gov.uk/components/accordion/
- https://designsystem.digital.gov/components/accordion/
- https://inclusive-components.design/collapsible-sections/
- https://github.com/scottaohara/a11y_accordions

Let me know if there are questions on this, or if I can help with anything else.

nichia:
I do like the "Expand all/Collapse all" buttons feature for this multi-select accordion that DK mentioned.
Thanks for picking up the IE 11 error issue, I've added a jira ticket for it. It looks like an issue for cmsds document site when we're viewing any of the components on IE 11 with DevTools on.

line47:
Hey @nichia maybe you can help me get Keyboard support for up/down arrows working and then we can merge this 👍

nichia:
Sounds good, will do.

nichia:
@line47 Since the `Down Arrow, Up Arrow, Home, End` are [(Optional) keyboard interaction](https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction) - it might be a good place merge/close off this PR and add a separate jira ticket / PR for the Accordion to include additional keyboard interaction and `collapse/expand all` feature ?

line47:
This looks good to me - I think we can merge into the feature branch @nichia @kimieann can you all review?

nichia:
They up/down key works great! 
- I tried testing with nested Accordion component and noticed that the up/down key scrolls through each accordion item, including the nested accordion items that are collapsed. Can we update scrolling to be on one level at a time?
- Can we add test cases for the up/down key triggers

line47:
I didn't even think about people having nested accordions. I don't think that is a practice we want to recommend?

nichia:
I haven't seen nested accordion other then on the [USWDS accordion example](https://designsystem.digital.gov/components/accordion/) where the Accordion is nested.

nichia:
I wonder if there are other applications that use nested accordions with up/down key? For these, what's the pattern for using the up/down key to tab through the accordion items? Tab through items on the same level, or tab through all items including the nested items?

line47:
I've not seen many use cases of nested accordions and don't think it's a good practice from a UX perspective. 
So I don't think we need to support it unless there is a real use case by one of the product teams.

I'll work on updating the tests

davidakennedy:
The keyboard shortcuts work great! 🎉

nichia:
Would be really nice to have up/down key be able to correctly tab throwing the Accordion with nested-Accordion (should apps inadvertently add that design). But understand if we can't have that. I'm fine with just having guidance about not using nested Accordion pattern and maybe note somewhere about this issue in the code?

nichia:
Other then that, great work with the Accordion and the up/down key feature!

packages/design-system-docs/src/pages/components/Accordion/_Accordion.docs.scss
```diff
@@ -0,0 +1,73 @@
...
+
+### When to consider alternatives
+
+Accordions hide content from users and not everyone will notice them or understand how they work.
+Accordions increase cognitive load and interaction cost, as users have to make decisions about what headers to click on.
+
+- Don't use an accordion if users need to see most or all of the information on a page.
+- Don't use an accordion if there is not enough content to warrant condensing.
+
+### Usage
```
kimieann, line 62: Based on the conversation here in the comments, do we need to state here that we do not recommend nesting accordions?

kimieann (commented):
When looking at the [test site](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-41-Accordion/components/accordion/), I am have the following issues.

The first two accordions on the page do not expand at all.

For the second two, because of the size of the iframe, when testing it on the page instead of doing new tab, it makes it appear that the first accordion replaces the second. You have to close the first then to open the second, and you can't see the content of the second.
image

line47:
@kimieann 
>The first two accordions on the page do not expand at all.

We have multiple plain HTML examples on the site and we do not include separate javascript with them. I think this is weird too. I was thinking about maybe we should put a disclaimer that if you use the plain HTML you'll need to write your own JS for the component. This is also the case for Tabs for example - https://design.cms.gov/components/tabs/

>For the second two, because of the size of the iframe, when testing it on the page instead of doing new tab, it makes it appear that the first accordion replaces the second. You have to close the first then to open the second, and you can't see the content of the second.

This is definitely an issue - I would very much like to get rid of the iframe altogether and present the React examples on the page somehow. Maybe this is something you and Ni can look into when working on the doc site updates?

kimieann:
> We have multiple plain HTML examples on the site and we do not include separate javascript with them. I think this is weird too. I was thinking about maybe we should put a disclaimer that if you use the plain HTML you'll need to write your own JS for the component. This is also the case for Tabs for example - https://design.cms.gov/components/tabs/

Oh, I see. Maybe you could do something similar to the tabs component then, where on the HTML markup, the first accordion is expanded and has `Note: This is just an example of the HTML markup. See the React example for a functioning example.`

> This is definitely an issue - I would very much like to get rid of the iframe altogether and present the React examples on the page somehow. Maybe this is something you and Ni can look into when working on the doc site updates?

I'm good with that. I was hoping there was a way to explictly just set the iframe height. For a cheap way around it, you could have one item in the accordion to be default open on load. I think that would force the iframe height to render larger.

packages/design-system-docs/src/pages/components/Accordion/_Accordion.docs.scss
line47: Working with CMS folks to finalize accordion documentation here - https://confluence.cms.gov/display/HCDSG/Accordion+guidance

line47:
That's a good idea! thanks!

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+    this.setState({ isOpen: !this.state.isOpen });
+  }
+
+  render(): React.ReactNode {
+    const {
+      buttonClassName,
+      children,
+      contentClassName,
+      heading,
+      headingLevel = '2',
```
kimieann: The docs site shows a headingLevel as required, but I do not get any error if I do not include it. It just defaults to 2. Also the doc site doesn't show anything in the default column.

If we are going to default to 2, shouldn't we put that in the default column and remove the required tag?

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -0,0 +1,93 @@
...
+  /**
+   * Class to be applied to the header `<button>` of an accordion item.
+   */
+  buttonClassName?: string;
+
+  children?: React.ReactNode;
+  /**
+   * Class to be applied to the heading `<h2>` tag of an accordion item.
+   */
+  contentClassName?: string;
```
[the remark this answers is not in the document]
kimieann: Ah, I see. I think this description is what is wrong. What you pass in as contentClassName does not get applied to the h2, it gets applied to the content div. Can you update the comment?

packages/design-system/src/components/Accordion/AccordionItem.tsx
kimieann: One other thing, no matter what headLevel I pass in, I see that `<h1>` `<h2>` etc gets put in the code, but visually there is no difference. Is this how it should be?

packages/design-system/src/components/Accordion/AccordionItem.tsx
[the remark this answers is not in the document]
line47: This has been updated

packages/design-system/src/components/Accordion/AccordionItem.tsx
line47: > The docs site shows a headingLevel as required, but I do not get any error if I do not include it. It just defaults to 2. Also the doc site doesn't show anything in the default column.

I don't see headingLevel as required here http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-41-Accordion/components/accordion/ but you are right it's not showing as a default of 2 I think this is an issue with our prop autogenerated docs @nichia do you know anything about that?


>One other thing, no matter what headLevel I pass in, I see that <h1> <h2> etc gets put in the code, but visually there is no difference. Is this how it should be?

That is working as expected and it's more for semantics

line47: Thanks!

kimieann (approved):
This draft component looks good to me now!

PR #965, opened 2021-03-09
[NO_TICKET] Allow feature flags to be used in the doc site

bernardwang:
## Summary
The old feature flag implementation which was based off a flag variable wasn't working when building for production. However, the feature flag does work when it is set inside an application, rather than inside the HCgov DS. This PR updates our doc site generation scripts to allow extra JS to be added to react example pages, in order to allow the HCgov DS to set the feature flag.

## How to test
- Test the flags on HCgov DS `bernard/update-cmsds-2.4.2` branch
- Test the flags on product applications, which now need to manually use the error message feature flags.


line47 (approved):
Tested with the core, HC gov design system and Windowshop application and it worked as expected, nice work!

PR #968, opened 2021-03-10
[NO_TICKET] Fix flag typing

bernardwang:
## Summary
- Include type defs for feature flag
- Rely on generating type defs for tsx files

## How to test
- Run the CMSDS in CRA typescript, try using the feature flag functions and other components to ensure the definitions are working as expected


nichia (approved):
The flag works great with CRA typescript works!

PR #971, opened 2021-03-11
[WNMGDS-816] Improve tooltip trigger customization

bernardwang:
## Summary
- Swap usage of tooltip trigger and content in <Tooltip>
  - Replace `children` prop with `title`
  - Replace `triggerContent` prop with `children` 
- Remove `trigger` prefix in props like `triggerClassName`
- Ensure undefined props are passed through to the trigger component
- Convert Tooltip.jsx to Tooltip.tsx

## How to test
- http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-816/improve-tooltip-trigger-customization


packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
```diff
@@ -5,90 +5,97 @@ import ReactDOM from 'react-dom';
...
   <>
     <div className="ds-u-display--flex ds-u-align-items--center ds-u-margin-y--2">
-      <p className="ds-u-margin--0">Tooltip using an icon trigger</p>
+      <p className="ds-u-margin--0">Tooltip with icon trigger</p>
       <Tooltip
         ariaLabel="Label describing the subject of the tooltip"
-        triggerContent={<TooltipIcon />}
-        triggerClassName="ds-c-tooltip__trigger-icon"
-        triggerActiveClassName="ds-c-tooltip-icon--active"
+        className="ds-c-tooltip__trigger-icon"
```
bernardwang, line 11: @LouisFettet Just so you are aware, I did end up deciding to swap how the user specifies the tooltip trigger and content. Now the `children` corresponds to the `trigger`, and `title` corresponds to the tooltip content.

This allows developers to have more flexibility with customizing the trigger content. It also means additional props will be able to automatically passed through to the trigger element.

packages/design-system/src/components/Tooltip/Tooltip.jsx
```diff
@@ -266,25 +268,34 @@ export class Tooltip extends React.Component {
...
   ariaLabel: PropTypes.string,
   /**
-   * Content inside the tooltip body or popover. If this contains interactive elements set the `interactive` prop.
+   * Tooltip trigger content
    */
   children: PropTypes.node.isRequired,
   /**
-   * Classes applied to the tooltip body
+   * When provided, this will render the passed in component for the tooltip trigger. Typically this will be a `button`, `a`,
+ or rarely an `input` element.
```
LouisFettet, line 294: This ends up wrapping the entire tooltip trigger, right? It might be good to spell that out a little more in this description.

LouisFettet (commented):
This looks good from a quick review; I didn't have time to give it a full test run in the screener since it would take a bit of refactoring, but I don't foresee any issues.

Left a few comments, but the only other thing to note is that we should update the `_Tooltip.docs.scss` since it mentions some of the old props.

packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
bernardwang: Ah I see, I was trying to follow the same naming convention as native HTML tooltips, which use the `title` attribute. Its a bit tricky, because the native HTML tooltip is really quite bad, but other design systems like [USWDS](https://designsystem.digital.gov/components/tooltip/) also seem to follow that naming convention

packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
```diff
@@ -5,90 +5,97 @@ import ReactDOM from 'react-dom';
...
-      <p className="ds-u-margin--0 ds-u-margin-right--1">Tooltip using a</p>
-      <Tooltip triggerContent="text trigger" triggerClassName="ds-c-tooltip__trigger-link">
-        Tooltip trigger is styled with dashed underline
-      </Tooltip>
+      <p className="ds-u-margin--0">
+        {'Tooltip with '}
+        <Tooltip
+          className="ds-c-tooltip__trigger-link"
+          component="a"
+          title="Tooltip trigger uses <a> for the trigger, styled with dashed underline"
```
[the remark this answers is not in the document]
bernardwang: Ah thats a good point, so the default styling for links across HCgov uses `border-bottom` instead of `text-decoration`? 

I think we'd want to keep it consistent between tooltips and links, and since the CMSDS link styling uses `text-decoration`, I will probably keep it the same for `ds-c-tooltip__trigger-link`. That being said, if HCgov uses `border-bottom` consistently, we could always override `ds-c-tooltip__trigger-link` and default link styles in the HCgov DS

nichia:
The new tooltip example on doc site looks really good! The changes works great.

As an aside - I just noticed that the `interactive content` of `More info` can be activated with mouse `click` or `enter` key. The `space` bar key cannot activate it. Is this expected?

packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
```diff
@@ -5,90 +5,97 @@ import ReactDOM from 'react-dom';
...
+        </Tooltip>
+      </div>
+      <div className="">
+        <p className="ds-u-margin--0">
+          {'Inverse tooltip with '}
+          <Tooltip
+            className="ds-c-tooltip__trigger-link"
+            component="a"
+            inversed
+            title="Tooltip trigger uses <a> for the trigger, styled with dashed underline"
```
kimieann, line 93: Not sure if I'm looking in the right place since I'm new to the DS PRs, but the inversed inline trigger doesn't show dashed underline for me, even though the tooltip content says it's supposed to be: 
image

packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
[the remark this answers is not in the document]
kimieann: It's strange how much thicker and clunkier dashed is for text-decoration vs. what Louis shows above with border-bottom-style. It almost looks closer to 'dotted' than 'dashed'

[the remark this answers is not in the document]
bernardwang: yeah thats true. I guess if no one is going to use the `text-decoration` style i'd be fine with going with `border-bottom-style`, but it would be great to check with a designer on this

packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
bernardwang: @kimieann @LouisFettet I filed [this ticket](https://jira.cms.gov/browse/WNMGDS-850) to capture the two styling issues that have been brought up

packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
LouisFettet: Gotcha, that makes more sense in context. I'll leave this up to your best judgment then!

packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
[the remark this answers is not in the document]
LouisFettet: > Ah thats a good point, so the default styling for links across HCgov uses `border-bottom` instead of `text-decoration`?

Correct, but I'm not tied to the `border` solution here; I actually think using `text-decoration` (perhaps with `text-underline-offset`) is much more semantically correct.  HCGov's styling is really out-of-date and shouldn't be used as the model in 99% of circumstances.

I mainly wanted to pass this on since Kristen brought up that the dashed underline using `text-decoration` made the text more difficult to read, which I think is a fair criticism.

Anyway, +1 to getting a designer involved here; ideally I think all of our link underlines should have similar placement regardless of how they're being used.

packages/design-system-docs/src/pages/components/Tooltip/Tooltip.example.jsx
line47: I think it would be worthwhile to try and gain alignment between Medicare and Healthcare on tooltip link styling so we can implement consistently across CMS

kimieann (approved):
Since a new ticket was filed for the one issue I had, and the spacebar issue has been resolved, I don't see any other issues with this PR.

nichia (approved):
LGTM

PR #975, opened 2021-03-17
[WNMGDS-823] Update build release files

nichia:
## Summary
The zipped files generated by 'npm pack' for release does not include typescript source files. 
This PR adds typescript source files to the zipped files for publishing to npm.

## How to test
- Run `npm pack ./packages/design-system/` command on terminal - this will generate a `.tgz` file
- Unzip the `.tgz` file and check that it contains all the files from `dist` folder and all the component files under `src/components` folder 
- Specifically check that it should include the typescript files for FormControl and InlineError

### DIFF
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -14,8 +14,8 @@
   "types": "dist/types/index.d.ts",
   "files": [
     "dist",
-    "src/components/**/*.?sx",
-    "!src/components/**/*.test.?sx"
+    "src/components/**/*.{jsx,tsx}",
+    "!src/**/*{.test,.spec,.d}.{jsx,tsx}"
   ],
   "dependencies": {
     "@popperjs/core": "^2.4.4",
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -14,8 +14,9 @@
   "types": "dist/types/index.d.ts",
   "files": [
     "dist",
-    "src/components/**/*.jsx",
-    "!src/components/**/*.test.jsx"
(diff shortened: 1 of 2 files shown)


packages/design-system/package.json
```diff
@@ -14,8 +14,8 @@
   "types": "dist/types/index.d.ts",
   "files": [
     "dist",
-    "src/components/**/*.jsx",
-    "!src/components/**/*.test.jsx"
+    "src/components/**/*.?sx",
```
bernardwang, line 17: This is just a file pattern, so we can be specific here like this:
 `"src/components/**/*.{jsx,tsx}",`

packages/design-system/package.json
```diff
@@ -14,8 +14,8 @@
...
+    "!src/components/**/*.test.?sx"
```
bernardwang, line 18: We could probably improve this line as well, this is taken from some code in our scripts package. we generally just want to only include src component files so that we can extract prop docs
```
    "!src/**/*{.test,.spec,.d}.{jsx,tsx}",
```

packages/design-system/package.json
nichia: Good point.

bernardwang (approved):
LGTM!

prerelease.sh
```diff
@@ -16,7 +16,7 @@ yarn install
 
 echo "${GREEN}Building files and running tests...${NC}"
 yarn build
-yarn test
+yarn test:unit
```
bernardwang, line 19: Thank you for updating this!

PR #976, opened 2021-03-17
[NO-TICKET] Set prop tables on the doc site to be compact

line47:
This PR adds the compact table styling to the React property tables 

DEMO URL 
http://design-system-demo.s3-website-us-east-1.amazonaws.com/NO-TICKET-compact-table-for-props


nichia (approved):
Thanks, the props table looks much better now!

PR #977, opened 2021-03-18
[WNMGDS-853] Generate proptypes automatically

bernardwang:
## Summary
- This PR adds `babel-plugin-typescript-to-proptypes` as a dep to the CMSDS scripts and to our babel.config to automatically generate PropTypes for components written in Typescript. 

## How to test
- Reproduce the error by running CRA example, and using the `InlineError` component. Provide some incorrect props to confirm PropTypes are not working
- Run and build this branch, and install it to the CRA example. Ensure PropTypes are working


bernardwang:
Just a reminder, this plugin should be added to the `babel.config` of both the HCDS and MDS as well, so they also support PropTypes fully

nichia (commented):
Thanks @bernardwang for the work. 
Looks like it works from looking at `<TextField>` component, we now get warning message for the making errorPlacement a required field on `FormControl`. After changing the field to be optional, the warning message will be resolved.


```
checkPropTypes.js:20 Warning: Failed prop type: The prop `errorPlacement` is marked as required in `FormControl`, but its value is `undefined`.
    in FormControl (created by TextField)
    in TextField
    in div
```

kimieann (approved):
This is working for me. Without this branch pulled in, there are no PropType warnings when incorrect types are provided. With this branch, I get: 
image

line47:
I'm able to get the CRA app working however I get an issue with trying to use this with CRA typescript example 

it fails to compile for me
Module not found: Can't resolve '@cmsgov/design-system/dist/types/InlineError/InlineError'
Even though I can find that file in the node_modules

packages/design-system/src/components/InlineError/InlineError.tsx
```diff
@@ -13,7 +13,7 @@ interface InlineErrorProps {
   inversed?: boolean;
 }
 
-function InlineError({ children, className, id, inversed }: InlineErrorProps): React.ReactElement {
+export function InlineError({ children, className, id, inversed }: InlineErrorProps): React.ReactElement {
```
kimieann, line 16: @line47 Can you try now with this change? 

For me, with the create-react-app-typescript, it wasn't pulling in the InlineError export with `export * from './InlineError/InlineError';` in `index.d.ts` - I either had to change that line to `export { default as InlineError}` or put this export above.

line47 (approved):
LGTM!

nichia (approved):
@kimieann - similarly with my testing, i also noticed that this pr still accepts 'non existent props', which is different from typescript which would throw an error of `Property 'nonExistentProp' does not exist on type 'IntrinsicAttributes & InlineErrorProps'.  TS2322.

I'm also not able to find a way to set this checking, but want to note that `babel-plugin-typescript-to-proptypes` is not as strictly typed as typescript for this case.

Other then that, the PR LGTM!

PR #978, opened 2021-03-19
[WNMGDS-854] Fix IE 11 'forEach' error

nichia:
## Summary
This PR fixes the issue of CMSDS document site throwing `forEach` error when viewing the website with DevTools on IE 11.

*To Reproduce*
1. Go to https://design.cms.gov/
2. Click on any component
3. See error

## How to test
- Go to the demo doc site: http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-854/fix-ie11-foreach-error/
- View with DevTools and check that the document site works on IE11 and other browsers


line47 (approved):
Tested with BrowserStack and the issue appears resolved! Nice work Ni!

packages/design-system-docs/src/example.js
```diff
@@ -31,14 +31,16 @@ function setModalExamples() {
   const dialogHeading = document.getElementById('dialog-heading');
   const dialogWrap = document.getElementById('dialog-wrap');
 
-  document.querySelectorAll('.dialog-open-btn').forEach((btn) => {
+  // Convert NodeList into an array with the slice.call() method because
+  // querySelectorAll() returns a NodeList and on IE 11, foreach() only works on Array objects
+  Array.prototype.slice.call(document.querySelectorAll('.dialog-open-btn')).forEach((btn) => {
```
kimieann, line 36: You might be able to use `Array.from` here just for simplicity, but otherwise this works!

kimieann (approved):

packages/design-system-docs/src/example.js
nichia: `Array.from` does look much better, but looks like it's not supported on IE unless additional polyfill is added. Let's stay with `Array.prototype.slice.call` for now

PR #979, opened 2021-03-22
[WNMGDS-826] - Improve tooltipby fixing invalid HTML of unallowed elements in Span

kimieann:
## Summary
[WNMGDS-826](https://jira.cms.gov/browse/WNMGDS-826)
A common Tooltip use case is inline sentences as a glossary term. However, tooltips should also only be attached to buttons, links, and inputs. The only way to achieve an inline style is to attach a tooltip to a link, inside of a block of text like a p element.

Because we want the tooltip content to follow the tooltip trigger, this means tooltip content with interactive content or headings will break HTML validation, as these elements cannot be located inside a p element or span.

**(From ticket): This PR will address this problem in 2 ways**
1. ~Handle `a` element tooltip triggers better by adding `role="button" tabindex="0"` to anchor elements with an `href`.~ This was already done in #971 here: https://github.com/CMSgov/design-system/pull/971/files#diff-1544310ec6e3788354a57d3ee54286c99bf4489ee0dee278fa7b6f8cd0e9dab6R136

2. Use react portal to move the tooltip content to the end of the body element, rather than adjacent to t

### Added
Added the use of React Portal to move the Tooltip Content to the end of the body element, rather than inside the `<span>`

### Changed
There were some existing `<span>` tags inside of the rendering of the Tooltip Content that I needed to change to divs, otherwise we would have still had the problem with interactive content inside of a span.

## How to test
[http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-826/improve-tooltip](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-826/improve-tooltip)

- Validate screenreader still announces content correctly
- Validate content still displays as expected


packages/design-system/src/components/Tooltip/Tooltip.jsx
```diff
@@ -198,9 +199,10 @@ export class Tooltip extends React.Component {
       : {
           onBlur: () => this.handleBlur(),
         };
+    const bodyElement = document.querySelector('body');
 
     const tooltipContent = () => (
-      <span
+      <div
```
kimieann, line 205: I'm not sure if this was actually necessary, but it seems if the purpose of the PR is to prevent interactive content passed in from {title} from being within a span, this also should not be a span.

line47: I feel like a `<div>` is a better choice here anyways as it's a container that doesn't need to be used inline

LouisFettet: Agreed, `<div>` is better here. I was the one that originally suggested using a `<span>` so that it wouldn't cause a break in inline content, but if we're using a portal then that solves that issue.

nichia:
When viewing `Tooltip dialog for interactive content` on Safari and Firfox, the `tab` key cannot tab-through the links in the tooltip content. Not sure if that happens for others?

davidakennedy:
@nichia Make sure you have this setting switched on in the OS settings: https://www.a11yproject.com/posts/2017-12-29-macos-browser-keyboard-navigation/

nichia (approved):
Great work, works well for me!

davidakennedy:
Hey all! I'm catching up on this, as I missed the previous work in #971 even though Bernard pinged me. Sorry about that!

I think I understand why we need to move this direction: To accommodate for inline tooltip triggers. But I'd like to avoid moving the content container away from the trigger, if at all possible. Here's why:

- When the content container is before the end of the `<body>` it means that content can never be inspected by a screen reader user. They only hear it when they hit the trigger and it's announced via `aria-describedby`. But if that content is super long, they have no way to get back to it, other than hitting the trigger a bunch, which isn't great. And we currently have a lot of super long tooltips in production, at least on healthcare.gov.
- Having the content follow the trigger means a screen reader can user can get to that content while the trigger is open (whether interactive content is in the content container or not.).

In [this comment from Bernard](https://github.com/CMSgov/design-system/pull/947#discussion_r586768815), I wasn't sure what exactly was breaking HTML validation? Perhaps there's another way to tackle this? I'm happy to prototype and test.

Am I missing other important context here?

kimieann:
@davidakennedy Hmmm... okay let me see if I can try to summarize the problems we are having and then see if what I did in the latest commit might fix everything.

1 - @LouisFettet recommended we use `<span>` as the container around the trigger link & trigger, so that inline triggers will show like this: 
image

and not this:
image

2 - But when we changed to a `<span>` because it wraps around the trigger link AND the trigger content AND the trigger content could possibly have interactive content (`<p>` or `<a>` tags), this means those would be children of a `<span>` tag, which is not considered valid HTML (you can try here: https://validator.w3.org/nu/#textarea)

3 - Bernard's solution was to move the trigger content to the end of the `<body>` tag, which would then solve the invalid HTML problem, but now causes a problem with screenreaders according to DK.

So [this commit](https://github.com/CMSgov/design-system/pull/979/commits/322eda835d29ba6615146ef737f7c220e0736bff) is my newest attempt at solving the problem, which:
1 - Replaces the `<span>` back with a `<div>` and removes the portal for the trigger content to move to the end of the body. This solves the screenreader problem and the invalid HTML problem.
2 - Styles the `<div>` as an `inline-block`, which should make it display as if it were a span. This should solve the inline problem.

Are there any reasons why we didn't do this originally, or issues that this route could cause? @davidakennedy @LouisFettet

line47:
`Inline-block` seems to work fine however Im wondering about changing that style to `inline` as that is the default value for a both a link and span?

davidakennedy:
@kimieann Thank you for the detailed explanation! I'm following now. I retested this, and the reworked commit works great.

nichia (approved):
Still works great and nice re-work.

davidakennedy:
I wanted to note one more thing that we should open a separate issue for, as it's unrelated to this.

When I tab onto the "Tooltip dialog for interactive content" tooltip, the content in the content container still gets announced even though the container has a `aria-hiddden="true"` on it and it's not displayed. It could be some kind of conflict with `aria-describedby`. I wanted to note it. I've put it on my list to investigate some more.

kimieann:
> When I tab onto the "Tooltip dialog for interactive content" tooltip, the content in the content container still gets announced even though the container has a `aria-hiddden="true"` on it and it's not displayed. It could be some kind of conflict with `aria-describedby`. I wanted to note it. I've put it on my list to investigate some more.

@davidakennedy I created WNMGDS-876 for this

kimieann:
> `Inline-block` seems to work fine however Im wondering about changing that style to `inline` as that is the default value for a both a link and span?

@line47 Changed in https://github.com/CMSgov/design-system/pull/979/commits/3237cd7d68f228f1d38be31be82e199d8d2fa86b

line47:
I'm still having issues where the content inside the `Tooltip dialog for interactive content` button is getting read before i interact with the button. This is happening with the latest code when running locally.

kimieann:
> I'm still having issues where the content inside the `Tooltip dialog for interactive content` button is getting read before i interact with the button. This is happening with the latest code when running locally.

@line47 I created WNMGDS-876 for that issue that David found since it wasn't related to this ticket.

line47 (approved):
LGTM! Awesome work!

PR #980, opened 2021-03-24
[WNMGDS-764] Remove line on last review item

kimieann:
## Summary
[WNMGDS-764](https://jira.cms.gov/browse/WNMGDS-764)
According to the ticket above, `The Review component includes a grey divider after the last review item, which is not necessary`

[Example below](https://design.cms.gov/patterns/review/), notice that after the last review item, there is an unnecessary gray line:
image

When viewing this change at [http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-764/remove-review-line](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-764/remove-review-line), the last line of a review item is removed:
image

### Changed
- Changed css so that it looks for `last-of-type` with class `ds-c-review` and removes the border on that item

**Some caveats**
- I had to change the example page to make the single component make more sense. It needed to be in it's own div, so that it wouldn't get a line underneath it.
Before:
image

After:
image

- This won't always work. It will work everytime a user structures their Review components like this:
```
<div>
    <Review />
    <Review />
    <Review />
</div>
```
BUT, if they do something like this, because the last div is really the last of type in that container, it's not going to remove the border. You can test this by playing around with the example page.
```
<div>
    <Review />
    <Review />
    <Review />
    <div id="someUnrelatedDiv"></div>
</div>
```

## How to test
View [existing](https://design.cms.gov/patterns/review/) vs. [this branch](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-764/remove-review-line)

### DIFF
--- a/packages/design-system-docs/src/pages/components/Review/Review.example.jsx
+++ b/packages/design-system-docs/src/pages/components/Review/Review.example.jsx
@@ -5,47 +5,51 @@ import { Review } from '@design-system';
 const editButtonHref = 'javascript:void(0);';
 ReactDOM.render(
   <div>
-    <Review heading="A single Review component" editHref={editButtonHref}>
-      This is an example of a single React Review component.
-    </Review>
-    <Review
-      className="ds-u-padding-top--6"
-      heading="Multiple Review components"
-      editHref={editButtonHref}
-    >
-      Multiple Review components can be combined together one after another.
-    </Review>
-    <Review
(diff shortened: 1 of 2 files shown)


line47 (approved):
LGTM

nichia (approved):
LGTM. Thanks for providing such detail notes on the changes, makes testing so much easier.

PR #982, opened 2021-03-24
[WNMGDS-850] Fix tooltip trigger link styles

kimieann:
## Summary
[WNGMDS-850](https://jira.cms.gov/browse/WNMGDS-850)

There are two trigger link changes in this PR for trigger link styles.

(1) Based on Bernard's conversation [here](https://github.com/CMSgov/design-system/pull/971#discussion_r594705783) with @LouisFettet , the previous dashed underline made the trigger difficult to read.

**This is the example Louis showed of what a trigger looks like on hc.gov:**
image

**And this is what it looked like in the original Tooltip PR:**
image

This PR introduces a compromise between changing to `border-bottom` and the issue of the thick dashed underline being hard to read. Instead of using `dashed`, I discussed with Bernard using `dotted`, because it actually looks more like the dashed underline in Louis's example above. 

**Below is dotted:**
image

Also, I made use of the text-underline-offset to make the text easier to read. (Note: I don't think IE supports this).

(2) Based on my original review of Bernard's PR (read here), the inverse trigger link was not showing dashed, and was showing a solid underline. This PR attempts to fix this by excluding `.ds-c-tooltip__trigger-link` from the general CSS rule for .ds-base--inverse and explicitly setting the text decoration in the Tooltip SCSS file.

**Before:**
image

**After:**
image

## How to test
View [http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-850/fix-tooltip-trigger-styles](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-850/fix-tooltip-trigger-styles) in multiple browsers. Ensure the dotted underline for the trigger is acceptable, and that the text is easier to read.

Also make sure the inverse trigger is no longer a solid underline.

### DIFF
--- a/packages/design-system/src/styles/components/_Tooltip.scss
+++ b/packages/design-system/src/styles/components/_Tooltip.scss
@@ -1,4 +1,5 @@
 @import '../settings/index.scss';
+@import '../base/link.scss';
 
 $tooltip-icon-size: 16px !default;
 $tooltip-arrow-size: 8px !default;
@@ -21,8 +22,16 @@ $tooltip-background-color-inverse: $color-background !default;
 %trigger-underline-styles {
   text-decoration: underline;
   text-decoration-style: dotted;
-  text-underline-offset: 2px;
+  text-underline-position: under;
 }
+/* stylelint-disable media-feature-value-dollar-variable */
+// Chrome needs this offset becasue it handles text-underline-position differently and without this the underline runs into descending letters like g,j,q,p,y.
+@media screen and (-webkit-min-device-pixel-ratio: 0) {
+  %trigger-underline-styles {
+    text-underline-offset: 2px;
+  }
+}
+/* stylelint-enable media-feature-value-dollar-variable */
 
 // stylelint-disable selector-no-qualifying-type
 // Tooltip trigger style
--- a/packages/design-system/src/styles/components/_Tooltip.scss
+++ b/packages/design-system/src/styles/components/_Tooltip.scss
@@ -3,12 +3,12 @@
 
 $tooltip-icon-size: 16px !default;
 $tooltip-arrow-size: 8px !default;
-$tooltip-plus-border-size: 24px !default;
+$tooltip-icon-container-size: 24px !default;
 
 $tooltip-border-radius: 4px !default;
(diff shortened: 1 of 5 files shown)


line47:
In my testing I observed the following:

Styling looks good - the dotted line and the `text-underline-offset` are applied
- Safari (latest)
- Firefox (latest) 
- Chrome (latest)
- Edge (latest) 

Styling doesn't get applied 
- IE 11 (doesn't apply the dotted style or the `text-underline-offset`)

This works for me and I think it's ok that IE11 is not supporting this at the moment

line47 (approved):
This looks good to me! Nice work!

packages/design-system/src/styles/components/_Tooltip.scss
```diff
@@ -18,6 +19,20 @@ $tooltip-background-color-inverse: $color-background !default;
   margin: 0;
 }
 
+%trigger-underline-styles {
+  text-decoration: underline;
+  text-decoration-style: dotted;
+  text-underline-position: under;
+}
+/* stylelint-disable media-feature-value-dollar-variable */
```
line47, line 27: Added offset for Chrome only as it interprets `text-underline-position` differently and needs the offset to make sure the line is under any text that descends.

packages/design-system/src/styles/components/_Tooltip.scss
```diff
@@ -18,6 +19,20 @@ $tooltip-background-color-inverse: $color-background !default;
   margin: 0;
 }
 
+%trigger-underline-styles {
+  text-decoration: underline;
+  text-decoration-style: dotted;
+  text-underline-position: under;
```
line47, line 25: Added `text-underline-position` to unsure the text link it below descending letters. This works more consistently than offset. 

the `under` value Forces the line to be set below the alphabetic baseline, at a position where it won’t cross any descenders. This is useful for ensuring legibility with chemical and mathematical formulas, which make a large use of subscripts. https://developer.mozilla.org/en-US/docs/Web/CSS/text-underline-position#values

nichia (approved):
Looks great.

PR #985, opened 2021-03-30
[WNMGDS-602] Change close button prop name in Dialog component

kimieann:
**Note:** 
- Add the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) to the PR title like this `[WNMGDS-10] - title of pr here` to link to the related issue in Jira.
- You can automatically [close related GitHub issues by using keywords](https://help.github.com/en/articles/closing-issues-using-keywords).
- A demo doc site is automatically generated for this branch at `http://design-system-demo.s3-website-us-east-1.amazonaws.com/{branch_name}`. For example, the demo for `WNMGDS-10/feature` will be located at `http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-10/feature`.
- If your changes involve code please update the snapshots by running `yarn update-snapshots`.

## Summary
Both Dialog and HelpDrawer have props for the close button, but they are named different. We should make `closeButtonText` and `closeText` naming and usage consistent.

Bernard also had this note in the ticket `We also need to consider usage in MCT where an icon is used instead of text.`

**But since we are just renaming, and not changing the type (which is ReactNode), I assume that if it supported an icon before, it still will. Is there something I'm missin ghere?**

### Changed
Changed `closeText` to `closeButtonText` in Dialog, to be consistent with the HelpDrawer prop name.

### Deprecated
`closeText` is being deprecated. Use `closeButtonText` instead

## How to test
Demo site: http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-602/change-closeButton-prop-name

- Make sure on Demo site everything still looks correct, and closeText is gone and closeButtonText shows
- Run branch locally
- In the `Dialog.example.jsx` file, send in `closeText` as a prop. You should see a warning like below in the console, but it should still work.
image

- Send in `closeButtonText` and it should work without warning.
- Send in nothing and it should default to Close without a warning.

### DIFF
--- a/packages/design-system-docs/src/pages/components/Dialog/Dialog.example.jsx
+++ b/packages/design-system-docs/src/pages/components/Dialog/Dialog.example.jsx
@@ -31,7 +31,6 @@ class Example extends React.PureComponent {
             onExit={() => this.hideModal()}
             getApplicationNode={() => document.getElementById('App')}
             heading="Dialog heading"
-            closeText="CloseyClosed"
             actions={[
               <button className="ds-c-button ds-c-button--primary" key="primary">
                 Dialog action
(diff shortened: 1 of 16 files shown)


line47 (commented):
LGTM!

nichia (approved):
LGTM

PR #986, opened 2021-03-30
[WNMGDS-874] Updates to tooltip design

line47:
## Summary
- Updated the styling of the tooltip based on direction from @mliupaeter 
- Removed the inverse option for the tooltip as styling of the tooltip is going to be the same no matter what the background is 
- Updated test to remove inverse test

### Added
- Added styles and variables to allow for brand theming by Healthcare and Medicare 

## How to test or demo
- Check out the live URL here 
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-874-update-tooltip-design/components/tooltip/

- Ensure the design matches the Sketch file https://www.sketch.com/s/9d0a4d12-4820-4cd7-984b-44080a4ad4ab


line47:
@kimieann you should be able to view the Sketch file to review now https://www.sketch.com/s/9d0a4d12-4820-4cd7-984b-44080a4ad4ab/a/zxEgGva

packages/design-system/src/components/Tooltip/Tooltip.jsx
```diff
@@ -310,7 +306,6 @@ Tooltip.propTypes = {
    * Sets the size of the invisible border around interactive tooltips that prevents it from immediately hiding when the cursor leaves the tooltip.
    */
   interactiveBorder: PropTypes.number,
-  inversed: PropTypes.bool,
```
kimieann, line 313: Also needs to be removed from Tooltip.d.ts file

kimieann (commented):
Comment below on updating the type file & I also think we can update Tooltip.example.jsx to not pass in "inversed" anymore in the example

packages/design-system/src/components/Tooltip/Tooltip.jsx
line47: I'm not sure how to get that updated because `Tooltip.d.ts` gets generated when `yarn build` is run. 

There is still an inverse prop for TooltipIcon that needs to still be part of the component.

nichia (commented):
Styling change needed for the links on interactive tooltip content Screen Shot 2021-04-05 at 8 46 07 AM

nichia:
It's probably using the inverse link styling right now.
Also, not related to this PR, but I'm getting the warning message when running locally about `<div> cannot be a descendant of <p>` which is cause from the external dependency `CSSTransition`. Not sure if this would be an issue that we need to look into.
Screen Shot 2021-04-05 at 9 13 03 AM

line47:
From our bi-weekly check in 

- [ ] Update documentation to talk about when to user certain positioning - consider removing some options to ensure the tooltip follows our usability, accessibility, and responsive guidelines

- [ ] Remove most of the customizable options - keep color 

- [ ] Document how font style gets pulled from different place and you don't need to change it here

- [ ] Explore update the color to vs gray which makes it seems disabled - use blue we have in other component examples

line47:
@nichia and @kimieann this is ready for another pass 👍

packages/design-system-docs/src/pages/components/Tooltip/_Tooltip.docs.scss
```diff
@@ -53,14 +53,10 @@ Style guide: components.tooltip.tooltip-icon
 
 The following Sass variables can be overridden to theme choice fields:
```
nichia, line 55:
```suggestion
```

nichia: Need to remove that extra line as with the extra line, subsequent variables are not being displayed by the docs site.

packages/design-system/src/components/Tooltip/Tooltip.test.jsx
```diff
@@ -24,11 +24,6 @@ describe('Tooltip', function () {
     expect(tooltip.wrapper).toMatchSnapshot();
   });
 
-  it('renders inverse tooltip', () => {
-    const tooltip = render({ inversed: true });
-    expect(tooltip.wrapper).toMatchSnapshot();
-  });
-
```
nichia, line 31: Add this test back as we're having `inversed` tooltip flag for the different tooltip color

nichia (approved):
Aside from a couple of small changes mentioned, the tooltip design updates LTGM!

packages/design-system/src/components/Tooltip/Tooltip.jsx
kimieann: If I delete `Tooltip.d.ts` file, it does not get re-generated when I run `yarn build` - Are you sure we can't edit directly, or is there something else to run to get them to re-generate?

line47: Oh I could be wrong @nichia can you weigh in here?

nichia: Yes, typedef files are auto-generated for Typescript codes only (eg. InlineError, FormControl, Accordion). We manually add the typedef files for other React codes like `Tooltip` etc. 
However, since  we're keeping the `inversed` prop here for passing to `TooltipIcon`, looks like we're good here.

kimieann (approved):

PR #995, opened 2021-04-05
[WNMGDS-876] Only include aria-describedby on non-interactive content Tooltips

kimieann:
## Summary
There was an issue noted by DK in [this pr](https://github.com/CMSgov/design-system/pull/979#issuecomment-808540772) where the screenreader was announcing the hidden content of interactive tooltips.

This is due to the fact that `aria-describedby` will announce the contents of items, even if they are hidden and have `aria-hidden` = true.

The non-interactive tooltips need `aria-describedby` because as you tab through them, the tooltip appears and we need the content of the tooltip to be read. However, for interactive tooltips, we do not want the content announced until the element is un-hidden, so we do not need `aria-describedby`

### Changed
`aria-describedby` will now only be included on items that do not include the `dialog` prop (non-interactive tooltips)

## How to test
- FIrst view the tooltip locally on master with the screenreader on
- As you tab through, the content of the interactive tooltip is announced, even though the tooltip is not displayed
- View http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNGMDS-876/Tooltip-screenreader-issue
- The content of the non-interactive tooltips should still be announced as you tab through
- The content of the interactive tooltip should not be announced until you hit spacebar to show the tooltip

### DIFF
--- a/packages/design-system/src/components/Tooltip/Tooltip.jsx
+++ b/packages/design-system/src/components/Tooltip/Tooltip.jsx
@@ -161,7 +161,7 @@ export class Tooltip extends React.Component {
       <TriggerComponent
         type={TriggerComponent === 'button' ? 'button' : undefined}
         aria-label={ariaLabel || ''}
-        aria-describedby={this.id}
+        aria-describedby={dialog ? null : this.id}
         className={triggerClasses}
         ref={this.setTriggerElement}
         {...others}
--- a/packages/design-system/src/components/Tooltip/Tooltip.jsx
+++ b/packages/design-system/src/components/Tooltip/Tooltip.jsx
@@ -200,7 +200,7 @@ export class Tooltip extends React.Component {
         };
 
     const tooltipContent = () => (
-      <span
+      <div
         id={this.id}
         tabIndex={dialog ? '-1' : null}
         ref={this.setTooltipElement}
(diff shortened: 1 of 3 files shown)


nichia (commented):
Tested on Safari using voiceover and the fix works great.
@davidakennedy this PR will omit the `aria-describeby` for interactive tooltip - is it okay to still omit the `aria-describedby` when this tooltip is activate and`aria-hidden='false'`? 
Is `aria-describedby` value required when the interactive tooltip is activated?

nichia:
Also, not related to this PR, but I noticed `undefined` on the class attribute of `<button>` element when the interactive tooltip is activated (see image). Should I add a ticket on Jira for this or would it be okay to put in a fix here?
Screen Shot 2021-04-05 at 2 08 59 PM

davidakennedy (approved):
I tested this, and it works as expected in all major screen readers we test in:

- Safari/VoiceOver
- JAWS/Edge
- JAWS/IE 11
- JAWS/Chrome
- Firefox/NVDA

This is the right move to make. First, this seems like how it works in all screen readers with hidden off screen content that uses `aria-describedby`. I made a simple test case, and it works the same way across screen readers: https://codepen.io/davidakennedy/pen/oNBeJoG

Second, this is something I missed when first testing the tooltip dialog. We don't need the `aria-describedby` here because we shift focus to the container once a user clicks the button.

> this PR will omit the aria-describeby for interactive tooltip - is it okay to still omit the aria-describedby when this tooltip is activate and aria-hidden='false'?

@nichia Good question! Yes, see above. We shift focus to the container, so we don't need to announce the content this way. The user will hear it.

> Is aria-describedby value required when the interactive tooltip is activated?

No, see above.

nichia (approved):
🚀

PR #996, opened 2021-04-07
[WNMGDS-867] First draft of adding FilterChip

kimieann:
## Summary
Adding new component for FilterTags

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-867/dismissible-badge


kimieann:
@line47 @nichia I changed this PR based on conversations Friday with CMS to create new FilterTag component. For now it copies the design of PET shared components FilterTag. I'm wondering, should it be more generic gray instead of blue, rounded instead of rectangular, etc.?

packages/design-system-docs/src/pages/components/FilterTag/_FilterTag.docs.scss
```diff
@@ -0,0 +1,45 @@
+/*
+Filter Tag
+
+Filter Tags are used to display dismissible filter chips
+
+Markup: filtertag.example.html
+
+Style guide: components.filtertag
```
line47, line 8: We should mark this component as draft by putting `@status Draft` on the doc page

packages/design-system-docs/src/pages/components/FilterTag/_FilterTag.docs.scss
```diff
@@ -0,0 +1,45 @@
...
+
+### When to consider alternatives
+
+- When a badge is sufficient in place of a filter tag
+
+### Usage
+
+-
+
+### Accessibility
```
line47, line 36: Do we have documentation for these sections?

packages/design-system/src/styles/components/_FilterTag.scss
```diff
@@ -0,0 +1,77 @@
+$filter-tag-button-size: 14px;
+$filter-tag-text-color: $color-base !default;
+$filter-tag-background-color: $color-primary-alt-lightest !default;
+$filter-tag-border-color: $color-primary !default;
+$filter-tag-text-color-inverse: $color-white !default;
```
line47, line 5: Can we rename these `color-inverse` variables to be `color-hover` ?

packages/design-system-docs/src/pages/components/FilterTag/_FilterTag.docs.scss
```diff
@@ -0,0 +1,45 @@
...
+
+-
+
+### Future research
+
+-
```
line47, line 42: We should add a customizations section here with SCSS variable names and what they do

packages/design-system-docs/src/pages/components/FilterTag/_FilterTag.docs.scss
kimieann: @line47 So I left them blank for now because I was going to ask - do I just write this up or do we get official content from elsewhere?

line47: It might be good to work on some guidance for these and then get feedback from folks after we've written something up

kimieann: @line47 I've added some more guidance. But admittedly, writing code is more my thing than text for humans to read :)

packages/design-system-docs/src/pages/components/FilterTag/filtertag.example.html
```diff
@@ -0,0 +1,23 @@
+<div class="ds-c-filter-tag">
+  <button class="ds-c-filter-tag__button" id="Example filter tag-tag">
+    <span class="ds-u-visibility--screen-reader">Remove Example Filter Tag </span
```
line47, line 3: I think we can remove this screen reader span and instead add `aria-label="Remove Example Filter Tag"` to the span with the clear icon on it

packages/design-system/src/components/FilterTag/FilterTag.tsx
```diff
@@ -0,0 +1,104 @@
...
+   */
+  value: string;
+  /**
+   *  For screenreaders, text to read for removal
+   */
+  removeText: string;
+  /**
+   * When false, filter tag is not removable
+   */
+  readOnly?: boolean;
```
line47, line 33: I'm wondering about if we change `readOnly` to something like `notDismissible` or something that lets folks know they can't remove or dismiss it

packages/design-system-docs/src/pages/components/FilterTag/_FilterTag.docs.scss
```diff
@@ -0,0 +1,59 @@
...
+
+### Accessibility
+
+- Use `removeText` prop so that screenreaders can accurately read out the close icon action
+- `Enter`, `Backspace`, `Delete`, and `Spacebar` will remove the tag when it is in focus
+
+### Future research
+
+### Customization
+The following Sass variables can be overridden to theme alerts:
```
line47, line 48:
```suggestion
The following Sass variables can be overridden to theme the filter tag:
```

packages/design-system-docs/src/pages/components/FilterTag/_FilterTag.docs.scss
```diff
@@ -0,0 +1,59 @@
...
+
```
line47, line 49: I think this empty line needs to be removed so the list shows up - currently none of the variables are showing 
```suggestion

```

packages/design-system/src/components/FilterTag/FilterTag.tsx
```diff
@@ -0,0 +1,104 @@
+import { ClearIcon } from '../ClearIcon';
+import React from 'react';
+import classNames from 'classnames';
+
+const actionableKeys = ['Enter', 'Backspace', 'Delete'];
+const idFromValue = (value: string) => `${value}-tag`;
+
```
nichia, line 7: To stay consistent with other components, can we use `uniqueId` from lodash instead. See `Alert.jsx` for an example.

packages/design-system/src/components/FilterTag/FilterTag.tsx
```diff
@@ -0,0 +1,104 @@
...
+              onClick={this.handleClick}
+              onKeyDown={this.handleKeyDown}
+            >
+              {!readOnly && <span className="ds-u-visibility--screen-reader">{removeText} </span>}
+              <span className="ds-c-filter-tag__label">{label}</span>
+              {!readOnly && (
+                <span className="ds-c-filter-tag__clear-icon">
+                  <ClearIcon />
+                </span>
+              )}
```
nichia, line 95: If `readOnly` filter tag is similar to a disabled button, can we set the `disable` attribute on this button so that we don't get keyboard focus for it.

packages/design-system/src/components/FilterTag/FilterTag.tsx
```diff
@@ -0,0 +1,104 @@
...
+  label: string;
+  /**  
+   *
+   Value for filter tag 
+   */
+  value: string;
+  /**
+   *  For screenreaders, text to read for removal
+   */
+  removeText: string;
```
pwolfert, line 29: I hadn't done it in the pet-shared-components version, but this is technically only necessary when `readOnly` is true. This is a good opportunity to correct this. Maybe something like
```ts
type RemoveProps =
  | { readOnly: true; removeText: string; }
  | { readOnly?: false; removeText?: string; }; 
```
I think that'll work. I tested it [here](https://www.typescriptlang.org/play?#code/C4TwDgpgBAShC2B7AbhACgJ0WAzlAvALABQUUAPlAN5QYQCGAJgPIB2ANiAFxTAYCuEANy0EKCABUIAD2A8cfAJasA5iIC+JMpRp0mbTgH4eAM3rscw0UlRTZxqAozK1UdUJIkT-VgGNgioisUABGiCEAFGBYuDxwNugxOACU1J7EmsQkYZG6DCwc3LwCEAA01uJ2clAARAAWijVuydnhEXn6habmls2tuc1AA)

nichia:
Also noticed that the doc site script displays typeError for the ClearIcon.tsx props - not sure why it's doing that as this component is not exported
```
🌪   Generating React propType documentation and grabbing raw example code
react-docgen TypeError: Cannot read property 'props' of undefined
react-docgen /Users/tchutnichia/Code/design-system/packages/design-system/src/components/ClearIcon/ClearIcon.tsx
 📝  Generating documentation pages
 ```

File: packages/design-system/src/components/FilterTag/FilterTag.tsx (hunk 6 above)
Comment (pwolfert) [reply packages/design-system/src/components/FilterTag/FilterTag.tsx:33]: Or if the `removeText` prop requirement becomes conditional, maybe the name could even change to something related to that prop, like `removeable`. Or this could actually be removed entirely, and it becomes removable if `onClear` is defined. And maybe `onClear` gets renamed to `onRemove`? These names don't need to match pet-shared in order for us to feel comfortable migrating away from our implementation 🙂

File: packages/design-system/src/components/FilterTag/FilterTag.tsx [hunk 12]
```diff
@@ -0,0 +1,104 @@
...
+  className?: string;
+  /**
+   * Text for the filter tag
+   */
+  label: string;
+  /**  
+   *
+   Value for filter tag 
+   */
+  value: string;
```
nichia, line 25: I wonder how the `value` for filter tag is used because if a function is passed in `onClear` prop, why would the calling function want to get that value back? Wouldn't we just run that function on handle click?

packages/design-system/src/components/FilterTag/FilterTag.tsx
```diff
@@ -0,0 +1,104 @@
...
+   */
+  readOnly?: boolean;
+  /**
+   * Function to call when filter tag is dismissed, will return value
+   */
+  onClear?: (value: string) => void;
+}
+
+interface FilterTagState {
+  tagDismissed: boolean;
```
pwolfert, line 41: What kind of use case would this have? The only way we use it so far in Enroll apps is where the this component's parent is in charge of managing this kind of state. I foresee this self-managed visibility resulting in a bunch of zombie filter tag instances.

packages/design-system/src/components/FilterTag/FilterTag.tsx
```diff
@@ -0,0 +1,104 @@
...
+  render(): React.ReactNode {
+    const { id, label, removeText, readOnly, className } = this.props;
+    const rootClassNames = classNames('ds-c-filter-tag', className);
+    const readOnlyTag = readOnly ? 'ds-c-filter-tag__button--read-only' : '';
+    const buttonClassNames = classNames('ds-c-filter-tag__button', readOnlyTag);
+    return (
+      <>
+        {!this.state.tagDismissed && (
+          <div className={rootClassNames}>
+            <button
```
nichia, line 83: I'm not sure about this, just want to check with you all, is <div> tag needed here or can we just have <button> tag for this filter tag component?

packages/design-system-docs/src/pages/components/FilterTag/_FilterTag.docs.scss
kimieann: I tried to add this, but it's not showing up on the demo URL or locally. Do you know why it's not working? I copied format from a different file.

packages/design-system-docs/src/pages/components/FilterTag/_FilterTag.docs.scss
kimieann: Got it!

packages/design-system/src/components/FilterTag/FilterTag.tsx
nichia: or change `onClear` to  `onClick?

kimieann: @line47 @nichia What do you think about removing the readOnly prop, renaming `onClear` to `onDelete`, and only having the X icon show if `onDelete` function is passed on? That is the way this filter chip behaves: https://material-ui.com/components/chips/

Can anyone think of a use case where we would need this to be removable, but a callback onDelete function wouldn't be necessary? I can't..

packages/design-system/src/components/FilterTag/FilterTag.tsx
kimieann: @davidakennedy Would we still want to tab and have the screenreader announce the items that are not dismissible, or should we skip over those?

packages/design-system/src/components/FilterTag/FilterTag.tsx
kimieann: Because the React component using this probably has a state being managed of which items are currently filtered. We would call back their onDelete function with the value so they could then remove this item from its state.

packages/design-system/src/components/FilterTag/FilterTag.tsx
nichia: Yes, that would work well.

packages/design-system/src/components/FilterTag/FilterTag.tsx
nichia: Ah, managing the state, thanks.

davidakennedy:
Hey all,

I ran this through my normal accessibility checks:

- Keyboard testing
- Color contrast and color blindness testing
- Screen reader testing
- Axe checks

It works really well, but there are some more things to consider from an accessibility standpoint.

**Screen reader text**

Right now, we have some hidden text in a span that says "Remove full button text" and then another span with the full button text. This results in screen readers hearing everything twice. To fix this, as @line47 suggested, we can either have "Remove" in the hidden text. Or use `aria-label` to override the visual label and say "Remove full button text" Using the hidden text is nice because you can throw a lang attribute on it and it shows up if CSS doesn't load for some reason. But using `aria-label` is probably more consistent with the rest of our components. Nice work on hiding the icon from screen readers!

**Target size**

Even though we're not required to meet WCAG 2.1 yet, one of the AAA (we follow AA) requirements is to make sure the target size for clickable elements is at least 44px by 44px. This is a good practice to follow because it helps people with low vision and motor dexterity issues be able to see and click the buttons more easily.

For the most part, width should be covered here because of text length in the buttons. But I'm wondering if we can add a `min-height: 44px` to the button to hit this guideline. It's especially useful for a component like this that will likely be a row of buttons next to each other. This could use some design input/feedback.

**Spacing between buttons**

This isn't a WCAG requirement, and I don't think something we can control via the component. However, one thing I'd like to encourage is adequate spacing between these filter buttons. This could probably go into the Accessibility Guidelines section. I'd word it how I've heard @SparrowSpit say it. Something like, "Allow adequate spacing between each button, as much as the design will allow. This helps people with low vision and motor disabilities be able to perceive and interact with the buttons."

**Focus management**

This is another item that may be difficult to control via the component: managing focus. This type of button disappears when interacted with, so teams need to be mindful of where focus goes after that. That depends on the design and situation.

I'd love to encourage teams to think this through. It could be making a prop with a function that sends focus to the ID specified. Or this could be a note in the Accessibility Guidelines. Curious to hear other thoughts on this.

**Focus styles**

We can't address this yet, but @line47, I wanted to flag this component as something worth testing with the new focus styles we're working through. Because it has rounded corners, not all focus styles will look good.

Let me know if you all have any questions on the above.

davidakennedy:
One more small thing I noticed. Seems like the `id` for the buttons are pulling from the button name, so it's resulting in invalid IDs. Like: `id="Readonly example-tag"`.

packages/design-system/src/components/FilterTag/FilterTag.tsx
kimieann: I think you're right. I'm going to remove & just change the example

packages/design-system/src/components/FilterTag/FilterTag.tsx
kimieann: @pwolfert Actually... is this really true above? Or could it all be handled on the parent component side without needing to be passed back the value?

pwolfert: It's technically true that it can be handled without passing any value. I had made it this way in so that I didn't have to create a new handler function for each of these filter tag items. Most of the time I was iterating over an array when rendering them. If it seems too arbitrary to pass the value, we can work around it 🤷‍♂️ . It's not a deal breaker without it.

nichia: We probably should keep the `value` prop as there's an existing use case to help track the filter tag items

kimieann: I think I'm going to remove value. It seems like an unnecessary extra layer.

nichia: How would the applications track it then? Would returning the `id` on `onClick` be helpful?

packages/design-system/src/components/FilterTag/FilterTag.tsx
davidakennedy: @kimieann @nichia Good question. What's the use case for that type of filter tag? Do you have an example?

packages/design-system/src/components/FilterTag/FilterTag.tsx
kimieann: The applications themselves can just handle the value in onDelete without us needing to take it in and return it

packages/design-system/src/components/FilterTag/FilterTag.tsx
kimieann: Based on conversations with Patrick I'm going to remove the read only / not dismissible functionality. He created the original version of this in `pet-shared-components` and put the functionality in case it was needed, but it's not actually being used. I think if someone were to want a read only filter chip, they'd just use a badge

kimieann:
I've made a good bit of changes based on the thorough reviews from everyone. Below I'll try to summarize the changes I just made.

- I renamed `onClear` renamed to `onDelete` just for clarity
- Removed `readOnly` prop and functionality. Based on conversations with Patrick, he'd originally put this in `pet-shared-components`, but it doesn't look like it's being used so it seems to be an unnecessary functionality. A team could use a badge, or use CSS to make this look read only.
- Removed the state being managed in the component .The parent component should manage the state and the actual dismissing of the chip.
- Renamed from FilterTag to FilterChip to be more consistent with current design speak. 
- Removed value prop. It seems to be an unnecessary extra layer. Because there is an onDelete function, it's easy enough for the parent component to handle knowing the value itself.
- Based on DK's recommendation, removed the min-height 44px
- Based on DK & Scott's recommendations, used `aria-label` for the remove text.

**Issues not addressed yet that need more input**
This comment from DK. Anyone have thoughts?
>I'd love to encourage teams to think this through. It could be making a prop with a function that sends focus to the ID specified. Or this could be a note in the Accessibility Guidelines.

packages/design-system/src/components/FilterChip/FilterChip.tsx
```diff
@@ -0,0 +1,75 @@
...
+   */
+  className?: string;
+  /**
+   * Text for the filter chip
+   */
+  label: string;
+  /**
+   *  For screenreaders, text to read for removal
+   */
+  removeText: string;
```
nichia, line 24: To stay consistent with other components (eg. Text field, Tooltip etc), can we change `removeText` prop name to `ariaLabel` instead

kimieann: Do you think that might be confusing that someone might think it's for the label itself instead of the close icon?

line47: Maybe something like `removeAriaLabel` ?

nichia: Ah, looks like Autocomplete has `ariaClearLabel` - how have consistent format and use `ariaRemoveLabel` here?

pwolfert:
Three of our healthcare.gov apps now have a blue background behind the filter sections, so this new color for the filter chips may make them hard to distinguish in those contexts.

Screenshots of those apps:
Screen Shot 2021-04-19 at 10 57 33 AM
Screen Shot 2021-04-19 at 10 57 59 AM

Screenshot of what the existing filter tag would look like given the background color of these proposed tags:
Screen Shot 2021-04-19 at 11 00 17 AM

packages/design-system/src/components/FilterChip/FilterChip.tsx
kimieann: @nichia I actually think we could use `ariaClearLabel` since it's the label for the Clear Icon

kimieann: Changed here - https://github.com/CMSgov/design-system/pull/996/commits/994327d10654571e12637cb90a9294faae5de72e

davidakennedy:
@kimieann This is looking good! One thing that wasn't clear in my original comment: The text for the `aria-label` needs to have everything in there. Like: `aria-label="Remove label text"`. This is because the `aria-label` overrides the visual label, so if we have "Remove" only, it doesn't provide enough context.

> This comment from DK. Anyone have thoughts?

I see we have a `onDelete` function. Maybe this is the best place to manage where focus next? I guess it depends on the design and functionality of the product that uses the chip. That may not make sense in all cases though. It might be worth talking to teams about the use cases for this component, and seeing where they plan on sending focus and how they're handling it. That could guide us.

PR #1011, opened 2021-04-20
[WNMGDS-885] Update error highlighting guidance for DateField

kimieann:
## Summary
Product teams (App3) have requested more guidance on how to highlight DateField errors.

Currently App3 highlights all fields (day, month, year) no matter what the date error is, as long as there is an error.

The design system documentation does not give guidance on whether if we know the error is for the year, should we only highlight the year field (and same for month and day).

### Added
Added guidance that if the field with an error is known, only that field should be highlighted instead of all date fiels.

## How to test
Read guidance at http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-885/add-highlighting-guidance


line47 (approved):
LGTM!

nichia (approved):

PR #1030, opened 2021-05-11
[WNMGPCII-1384] Provide e2e-test helpers to child design systems

pwolfert:
## Summary
See https://github.com/CMSgov/hcgov-design-system/pull/35 for more context

### Changed
- Move e2e test helpers to the scripts package where it will be available to child design systems but not application projects
- Move the step-list test helpers to the `StepList` directory so we can get rid of the `helpers` directory in the core package and remove its build exception

### Fixed
- Fixed ES modules import syntax mixed in with CommonJS require syntax in the Jest config that made e2e tests fail in child design systems

## How to test
Clone this and the hc.gov design system repository so that they share a parent directory, then modify the `package.json` to use these modified dependency entries:
```
"@cmsgov/design-system": "file:../design-system/packages/design-system",
"@cmsgov/design-system-docs": "file:../design-system/packages/design-system-docs",
"@cmsgov/design-system-scripts": "file:../design-system/packages/design-system-scripts",
```
Then run `yarn test:e2e` in the child design system.


line47 (approved):
Works for me 🚀

nichia (approved):
:tada:

PR #1035, opened 2021-05-13
[WNMGDS-865] Deprecate danger button

line47:
## Summary
- Deprecating the danger button variation from the documentation and adding warnings in the console and CSS when the danger button is used

### Deprecated
- Use of the danger button is deprecated and will show a warning in the console and in CSS however it will still work until fully deprecated
Demo URL
http://design-system-demo.s3-website-us-east-1.amazonaws.com/deprecate-danger-button 

## How to test
- View the button page at the demo URL and see that the danger button is no longer on the documentation site
- Add the `danger` variation prop or the `ds-c-button--danger` class to a button and see that it will still work but will show deprecated in the console and css if you use the inspector.


packages/design-system-docs/src/scripts/components/PageHeader.jsx
```diff
@@ -8,7 +8,7 @@ class PageHeader extends React.PureComponent {
       const classes = classNames('ds-c-badge ds-u-margin-left--1 ds-u-text-transform--capitalize', {
         'ds-c-badge--success': this.props.status === 'Ready',
         'ds-c-badge--warn': this.props.status === 'Draft',
-        'ds-c-badge--error': this.props.status === 'Deprecated',
+        'ds-c-badge--error': this.props.status === 'Do not use',
```
nichia, line 11: Use `ds-c-badge--alert` for the red background badge: 
```suggestion
        'ds-c-badge--alert': this.props.status === 'Do not use',
```

packages/design-system/src/components/Button/Button.jsx
```diff
@@ -159,7 +164,7 @@ Button.propTypes = {
   /**
    * A string corresponding to the button-component variation classes
    */
-  variation: PropTypes.oneOf(['primary', 'danger', 'success', 'transparent']),
+  variation: PropTypes.oneOf(['primary', 'success', 'transparent']),
```
nichia, line 167: Removing `danger` variation will break existing code that are using it. We can remove it when it's fully deprecated.
We want to keep `danger` variation and just throw console.warn deprecated for now (like what you have further up).

nichia: Maybe add comment line above the `variation` propstype?

/* The `danger` variation is deprecated and will be removed in a future release

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -1,5 +1,10 @@
 @import '../settings/index.scss';
 
+// Danger button is marked for deprecation //
+.ds-c-button--danger {
+  --deprecated: 'WARNING: Use .ds-c-button or .ds-c-button--primary instead';
+}
```
nichia, line 6: Just curious, would I be able to see this `deprecated` message on the browser? Or it's placed here so coders are aware of the deprecation?

packages/design-system/src/types/Button/Button.d.ts
```diff
@@ -6,7 +6,7 @@ export type ButtonSize = 'small' | 'big';
 
 export type ButtonType = 'button' | 'submit';
 
-export type ButtonVariation = 'primary' | 'danger' | 'success' | 'transparent';
+export type ButtonVariation = 'primary' | 'success' | 'transparent';
```
nichia, line 9: Same here, keep the `danger` variation for backwards-compatibility with typescript code

packages/design-system/src/styles/components/_Button.scss
line47: This will show in the code - it was a way to show it in the CSS side of things 👍

nichia (approved):
Just a minor change and :rocket:

kimieann:
I'm going to try to push a little fix for this PR. I am noticing when trying to use this in create-react-app-typescript I get an error for the below:

image

It looks like there is an extra space in the `danger` definition in the Button.d.ts file:
image

kimieann (approved):
Working for me now.

scheul93 (approved):

PR #1038, opened 2021-05-17
[WNMGDS-916] add Secondary and Tertiary button

line47:
## Summary
Adding secondary and tertiary button styles and examples

### Added
- Added `ds-c-button--secondary` button styles
- Added `ds-c-button--tertiary` button styles

Screen Shot 2021-05-20 at 10 28 44 AM

## How to test
Demo URL 
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-865-add-secondary-and-teriary-button/components/button/


laurellawrence:
Looks good to me @line47

packages/design-system-docs/src/pages/components/Button/button.example.html
```diff
@@ -1,4 +1,5 @@
 <h6 class="preview__label">Default button</h6>
+<p><code>.ds-c-button</code></p>
```
nichia, line 2: Placing the class names for each button group does make it easier for coders to reference, but the layout looks nicer without it. Maybe remove the class names code since people can view the code when they expand the `Code snippet`

packages/design-system-docs/src/pages/components/Button/button.example.html
```diff
@@ -67,26 +107,30 @@ <h6 class="preview__label">Success button</h6>
...
+  <h6 class="preview__label">Size variant buttons</h6>
+  <p><code>.ds-c-button .ds-c-button--small</code></p>
+  <button type="button" class="ds-c-button ds-c-button--small">
+    Small
+  </button>
+  <p><code>.ds-c-button .ds-c-button--big</code></p>
+  <button class="ds-c-button ds-c-button--big">
+    Big
+  </button>
+</div>
```
nichia, line 128: For size variant buttons, I do like the previous version which shows the `small`, `default`, `Big` buttons as that gives a good comparison of the size variant

packages/design-system-docs/src/pages/components/Button/button.example.html
```diff
@@ -32,7 +34,44 @@ <h6 class="preview__label">Primary button</h6>
...
+</button>
+<button type="button" class="ds-c-button ds-c-button--secondary ds-c-button--hover">
+  Hover
+</button>
+<button type="button" class="ds-c-button ds-c-button--secondary ds-c-button--active">
+  Active
+</button>
+<button type="button" class="ds-c-button ds-c-button--secondary ds-c-button--focus">
+  Focus
+</button>
```
nichia, line 50: The test on HCgov design system has a grayed out Secondary Focus button. Not an issue here, just want to note it here as we may need to add an override over at the HCgov ds

packages/design-system-docs/src/pages/components/Button/button.example.html
line47: I've always found it more difficult to look through the code blocks because there is often so much in there. We pull the classes out for all the utility pages and I was thinking it would be more helpful to see the class next to the visual. 

I think this would be something to ask folks to see which version they prefer?

packages/design-system-docs/src/pages/components/Button/button.example.html
line47: Hmm, I'll look back into that

packages/design-system-docs/src/pages/components/Button/button.example.html
line47: Getting a lot of feedback in Slack and folks so far overwhelmingly prefer to see the classname inline next to the element.

line47: https://adhoc.slack.com/archives/C201U8U7N/p1622038453066200 
https://adhoc.slack.com/archives/CL3CQ9LPR/p1622038580058400

line47: Going to leave the classes inline based on feedback from the polls in Slack

packages/design-system-docs/src/pages/components/Button/button.example.html
nichia: I'm seeing this on HCgov DS
Screen Shot 2021-05-26 at 1 24 12 PM

line47: That's probably because we would need to have unique theming for HCgov and Mgov

packages/design-system-docs/src/pages/components/Button/button.example.html
nichia: Yes, go with that which has better feedback, thanks for doing the extra polling.

nichia (approved):
:rocket:

scheul93 (approved):

PR #1040, opened 2021-05-19
[WNMGDS-871] Focus styles 2

line47:
## Summary
Adding themed focus styles to the design system 

### Added
- Added focus style to accordion 
- Added focus style to filter chip
- Added focus style to filter chip
- Added focus style positioning to tabs, autocomplete

### Removed
- old gold focus styles

#### Transparent button has a slight visual issue with the focus style when there is a transparent background used (Modal)

Screen Shot 2021-05-19 at 2 49 41 PM

- We could set the background to white or the default bg color for the design system 
- We could set the corner radius on transparent buttons (links) to be 0px and look like non button links

#### Setting position on focused styles might not be the best approach across the board because this could impact items that are positioned in a different way (IE absolute)

https://user-images.githubusercontent.com/1449852/118868054-bdebc700-b8b1-11eb-9b9c-ba63df4cb8be.mov

#### Links with background color appear on top of the focus styles where there is not a position set on the focused element

Screen Shot 2021-05-19 at 2 53 35 PM

## How to test
- click around on all the things as well as use the keyboard to set focus on elements
- turn off the focus styles by setting the build variable `$ds-include-focus-styles` to `false` in packages/design-system/src/styles/settings/variables/_build.scss

Demo URL 
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-871-focus-styles-2

Mgov PR for testing
https://github.com/CMSgov/mgov-design-system/pull/33 

Hcgov PR for testing 
https://github.com/CMSgov/hcgov-design-system/pull/40


laurellawrence:
Screen Shot 2021-05-19 at 3 17 55 PM
Screen Shot 2021-05-19 at 3 17 51 PM

When testing out our focus styles for the vertical nav, we have a strange overlap

davidakennedy:
This is looking great, @line47! 🎉 

## Feedback on the issues

> Transparent button has a slight visual issue with the focus style when there is a transparent background used (Modal)

Seems like setting a white background would be the least disruptive. If we do that, does it become more difficult to create modals with inverse backgrounds? Do we have modals with inverse backgrounds that this would affect? That's the only drawback I can think of there.

> Setting position on focused styles might not be the best approach across the board because this could impact items that are positioned in a different way (IE absolute)

Is this only needed for the tabs, autocomplete and choice components? Seems like if so, maybe we could talk with product teams about this to see if it would be a problem. I'd love to not have this, but that doesn't seem to be an option. Another idea would be a flag in Sass to turn the `focus-styles-position` on or off in certain cases. I know that complicates things, but it's an option.

> Links with background color appear on top of the focus styles where there is not a position set on the focused element

After some research, this is how HTML works when it comes to stacking contexts. I didn't know this! Things later in the DOM sit on top without interference from CSS. That said, maybe this is a case that wouldn't come up too often?

> By default, HTML elements will be stacked according to their DOM order.

[Source article](https://www.joshwcomeau.com/css/stacking-contexts/); [Reduced test case](https://codepen.io/davidakennedy/pen/wvJoQQp)

## Other notes

*Windows High Contrast Mode*

I can add support for this. I've already tested it locally since I have a Windows machine.

*Default browser styles in some spots*

I wanted to note that within our components, we still have browser default styles in two spots:

1. The modal when focus is shifted to the modal container

Focus on modal container with blue outline

2. The skip nav when focus is shifted to the main container (this one we can't really address, as it's up to products)

Main content container with blue outline

I don't think we need to address these, as we can leave it up to products, but wanted to point it out in case it comes up down the road.

*Thick white lines in some cases*

There are a few cases where a white border is next to the white inner box shadow, making a super thick line. I don't think this is a dealbreaker, but wanted to point it out.

Checkbox on dark inverse background, focused with larger white border and purple focus color

Inverse button on dark inverse background, focused with larger white border and purple focus color

Let me know if there are any questions on this.

davidakennedy:
@laurellawrence 

> When testing out our focus styles for the vertical nav, we have a strange overlap

I think this is the same thing as @line47 noted with the "Links with background color appear on top of the focus styles where there is not a position set on the focused element" See my last comment, it seems like HTML works that way, related to the stacking context.

line47:
> Seems like setting a white background would be the least disruptive. If we do that, does it become more difficult to create modals with inverse backgrounds? Do we have modals with inverse backgrounds that this would affect? That's the only drawback I can think of there.

I agree with changing the transparent background to a  white background for the transparent close button on the Modal - All other versions of transparent buttons use a background color so this would actually conform to the rest of the design system usage for this type of button. 

>Is this only needed for the tabs, autocomplete and choice components? Seems like if so, maybe we could talk with product teams about this to see if it would be a problem. I'd love to not have this, but that doesn't seem to be an option. Another idea would be a flag in Sass to turn the focus-styles-position on or off in certain cases. I know that complicates things, but it's an option.

Currently position is applied to tabs, autocomplete and vertical nav and I'm thinking it's OK to leave those in because these are complex components made up of multiple elements. I do not think we should apply positioning to single elements. For example, setting focus position styles on buttons by default would impact products that use some sort of positioning on a button. 

> focus styles for Modal and Skip nav

We could use something like this to target focus for those cases. This might be something to ask CMS folks about 
```
*:focus {
  @include focus-styles;
}
```

> Thick white lines in some cases

I think healthcare and medicare are moving away from darker background colors for content areas so I don't think this is a concern for these new styles. 👍

davidakennedy:
@line47 All this makes sense. 👍 

> We could use something like this to target focus for those cases. This might be something to ask CMS folks about

I'm okay with leaving it as is, and revisiting if we think it's worth it. I don't think it makes sense to add the focus indicator on a `<main>` element because it's so wide on the screen. The modal may be improved with it, but perhaps not.

packages/design-system/src/styles/components/_Autocomplete.scss
```diff
@@ -15,6 +15,9 @@
       display: block;
     }
   }
+  input:focus {
+    @include focus-styles-position;
+  }
```
nichia, line 20: Do we want to apply this mixin only when focus-styles is enabled? ie: check for `@if $ds-include-focus-styles` first?

line47: Good catch and yes I think so!

davidakennedy: I think this would be a good idea.

davidakennedy:
I retested, and this is looking good.

The one additional thing I found is something I missed last time:

Screenshot of filter chip with focus ring and black x that's hard to see.

The black "x" when it's not circled is too dark. Should that be white like the other "x" with the circles around them? It would meet contrast this way.

line47:
I think this was missed in the filter chip work as that is the same as what is live on https://design.cms.gov/components/filterchip/

We might want to put up a quick PR to fix and not add to this one, I can do that

davidakennedy:
Just a quick note that I tested with HCGov design system and was able to change the colors, and turn off the focus styles without issue.

kimieann (approved):
Everything I reviewed looks good!

nichia (approved):
:rocket:
Tested on core DS, Hcgov DS and Mgov DS and focus styles worked great! 
Easy to override the focus styles light/dark color and turn off the setting on each of the DS.

PR #1041, opened 2021-05-21
[WNMGDS-760] Convert DateField and ChoiceList components to typescript

kimieann:
## Summary
- Updated DateField and ChoiceList components to typescript
- There were issues in the DateField component because of defaultDateFormatter being exported and the typescript doc generator thinking that it needed docs generated. Pulled this out into a separate file to avoid conflicts
- The typescript doc generator also had issues with typescript classes that extend HTML props. At first it was printing all props, the components and the HTML props. I had to add special options to pass into the typescript doc generator so that it would only include the component props

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-760/convert-datefiled-choicelist-to-typescript
- Review doc pages for DateField and ChoiceList and compare to existing doc site. Make sure there are no mismatches
- Test DateField and ChoiceList in create-react-app and create-react-app-typescript and make sure behavior is as expected


packages/design-system-scripts/gulp/docs/extractReactData/parseReactFile.js
```diff
@@ -113,8 +113,16 @@ function parseComponent(file, options) {
     if (!options.typescript) {
       docs = [{}];
     } else {
+      // Filter so that when Components extend standard HTML, those props do not also get printed in the docs page
```
kimieann, line 116: What this is doing is saying we only want props that are the components, so that we don't print all HTML props when we extend HTML components.

packages/design-system/src/components/DateField/DateField.tsx
```diff
@@ -1,19 +1,18 @@
-import * as React from 'react';
+import { FormControl, FormControlPropKeys } from '../FormControl/FormControl';
+import defaultDateFormatter from './defaultDateFormatter';
```
nichia, line 2: I just realized that Eslint doesn't automatically check the Typescript codes like it does for the React codes during the `git commit` stage. However, github actions does automatically does the eslint check here. This would be be fixed by moving the `import defaultDateFormatter from './defaultDateFormatter';` below the `import React from 'react';` line

packages/design-system/src/components/DateField/DateField.tsx
```diff
@@ -23,14 +22,16 @@ export interface DateFieldProps {
    * Additional classes to be added to the root fieldset element
    */
   className?: string;
+  disabled?: boolean;
```
nichia, line 25: Good pickup on `disabled` field as it's the react file but was missing from the original typedef file.
The `disabled` line can be moved after the `dateFormatter` prop.

nichia: `lableId` is also on the react file, but was missing from the original typedef file - need to be added here too.

nichia (commented):
Looks good so far! 
Testing on Core, Hcgov DS worked, but getting an error on Mgov DS:
`mgov-design-system/node_modules/@cmsgov/design-system/dist/types/index.d.ts(41,15): error TS2307: Cannot find module './Choice/Choice'.`

Looks like the type definitions for `./Choice/Choice` is not getting generated by cmsds scripts as I cannot find it on the `dist` folder.

nichia:
The branch can be formatted with Prettier and eslint by manually running `yarn lint`

kimieann:
@nichia Everything should be fixed now. Can you let me know if you're still seeing an error on mgov? Choice/Choice doesn't exist, it's ChoiceList/Choice

line47:
@kimieann When I compare the demo site with the live site I'm seeing differences in the React prop types 

For example on the demo site I see `choices` type info is 
```
Pick<ChoiceProps, "className" | "disabled" | "hint" | "requirementLabel" | "label" | "labelClassName" | "children" | "checked" | "checkedChildren" | "uncheckedChildren" | "inputClassName" | "defaultChecked" | "id" | "value">[]
```

And on the live site i see this type info for `choices`
```
arrayOf[custom]
```

I also see this for other props where the content in type is different between the demo and live site

kimieann:
> @kimieann When I compare the demo site with the live site I'm seeing differences in the React prop types
> 
> For example on the demo site I see `choices` type info is
> 
> ```
> Pick<ChoiceProps, "className" | "disabled" | "hint" | "requirementLabel" | "label" | "labelClassName" | "children" | "checked" | "checkedChildren" | "uncheckedChildren" | "inputClassName" | "defaultChecked" | "id" | "value">[]
> ```
> 
> And on the live site i see this type info for `choices`
> 
> ```
> arrayOf[custom]
> ```
> 
> I also see this for other props where the content in type is different between the demo and live site

@line47 In some cases like this it will be different, just because PropTypes is more ambiguous than TypeScript. In cases like this above where `arrayOf[custom]` really tells you... absolutely nothing, and `Pick<ChoiceProps, "className" | "disabled" | "hint" | "requirementLabel" | "label" | "labelClassName" | "children" | "checked" | "checkedChildren" | "uncheckedChildren" | "inputClassName" | "defaultChecked" | "id" | "value">[]` tells you exactly which values are allowed in the array of ChoiceProps, I would think it would be okay. 

My worry is more making sure there are no missing proptypes, or proptypes where we gave information about which field was allowed before and now give something ambiguous. Open to discussion here though. @nichia

line47:
Thanks for explaining @kimieann - here is another scenario i noticed when comparing prop types 

I'm noticing a few places the prop for something like `errorMessage` type on demo is `ReactNode` but on the live site it is `string, number, element, or array` which to meet seems like it is more specific. 

Do we want to use `ReactNode` for examples like this?

kimieann:
> Thanks for explaining @kimieann - here is another scenario i noticed when comparing prop types
> 
> I'm noticing a few places the prop for something like `errorMessage` type on demo is `ReactNode` but on the live site it is `string, number, element, or array` which to me seems like it is more specific and we would not want to replace with `ReactNode`.
> 
> Do we want to use `ReactNode` for examples like this?

In these cases, I think `ReactNode` is correct. If you look at the old PropTypes, they had `PropTypes.node,` which for some reason converted to `string, number, element, or array`

The Typescript docgen seems to keep ReactNode as ReactNode.

line47:
### For the `<choice> props
- `className` - `Additional classes to be added to the root div element` is missing from the demo 
- `disabled ` is missing from the demo

kimieann:
> ### For the `<choice>` props
> * `className` - `Additional classes to be added to the root div element` is missing from the demo
> * `disabled ` is missing from the demo

@line47 They should be there with latest commit

line47:
I'm getting an issue when trying to add  <DateField/> to the typescript example 
```TSX
TypeScript error in /Users/scottweber/Documents/github/design-system/examples/create-react-app-typescript/src/components/App.tsx(76,9):
JSX element type 'ReactNode' is not a constructor function for JSX elements.
  Type 'undefined' is not assignable to type 'Element | null'.  TS2605

    74 |         </div>
    75 | 
  > 76 |         <DateField></DateField>
       |         ^
```

nichia:
I'm getting the same issue of `JSX element type 'ReactNode' is not a constructor` on `DateField` component using typescript app

packages/design-system/src/components/DateField/DateInput.tsx
```diff
@@ -1,223 +1,233 @@
...
+    this.handleBlur = this.handleBlur.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  monthInput: any;
+  dayInput: any;
+  yearInput: any;
+
+  formatDate(): DateObject {
+    console.log(this.monthInput, this.dayInput, this.yearInput);
```
nichia, line 147: Can remove `type OmitChoiceProp = 'inversed' | 'name' | 'onBlur' | 'onChange' | 'size' | 'type' | 'inputRef';`

packages/design-system/src/components/DateField/DateField.tsx
```diff
@@ -150,6 +154,32 @@ export interface DateFieldProps {
   yearValue?: DateFieldYearValue;
 }
 
-export default class DateField extends React.Component<DateFieldProps, any> {
-  render(): JSX.Element;
+export function DateField(props: DateFieldProps): React.ReactNode {
```
nichia, line 157: Change to return type React.Element
`export function DateField(props: DateFieldProps): React.ReactElement {`

packages/design-system/src/components/DateField/DateInput.tsx
nichia, line 147: Can remove this `console.log(this.monthInput, this.dayInput, this.yearInput);`

nichia:
> > @kimieann When I compare the demo site with the live site I'm seeing differences in the React prop types
> > For example on the demo site I see `choices` type info is
> > ```
> > Pick<ChoiceProps, "className" | "disabled" | "hint" | "requirementLabel" | "label" | "labelClassName" | "children" | "checked" | "checkedChildren" | "uncheckedChildren" | "inputClassName" | "defaultChecked" | "id" | "value">[]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > And on the live site i see this type info for `choices`
> > ```
> > arrayOf[custom]
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I also see this for other props where the content in type is different between the demo and live site
> 
> @line47 In some cases like this it will be different, just because PropTypes is more ambiguous than TypeScript. In cases like this above where `arrayOf[custom]` really tells you... absolutely nothing, and `Pick<ChoiceProps, "className" | "disabled" | "hint" | "requirementLabel" | "label" | "labelClassName" | "children" | "checked" | "checkedChildren" | "uncheckedChildren" | "inputClassName" | "defaultChecked" | "id" | "value">[]` tells you exactly which values are allowed in the array of ChoiceProps, I would think it would be okay.
> 
> My worry is more making sure there are no missing proptypes, or proptypes where we gave information about which field was allowed before and now give something ambiguous. Open to discussion here though. @nichia

The previous version of `arrayOf[custom]` would require us to add additional info/links at the comment area to reference the ChoiceProps.
I like that this version display `ChoiceProps` - however, there are 21 Choice Props, not sure how that cell would look

kimieann:
> > > @kimieann When I compare the demo site with the live site I'm seeing differences in the React prop types
> > > For example on the demo site I see `choices` type info is
> > > ```
> > > Pick<ChoiceProps, "className" | "disabled" | "hint" | "requirementLabel" | "label" | "labelClassName" | "children" | "checked" | "checkedChildren" | "uncheckedChildren" | "inputClassName" | "defaultChecked" | "id" | "value">[]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > And on the live site i see this type info for `choices`
> > > ```
> > > arrayOf[custom]
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > I also see this for other props where the content in type is different between the demo and live site
> > 
> > 
> > @line47 In some cases like this it will be different, just because PropTypes is more ambiguous than TypeScript. In cases like this above where `arrayOf[custom]` really tells you... absolutely nothing, and `Pick<ChoiceProps, "className" | "disabled" | "hint" | "requirementLabel" | "label" | "labelClassName" | "children" | "checked" | "checkedChildren" | "uncheckedChildren" | "inputClassName" | "defaultChecked" | "id" | "value">[]` tells you exactly which values are allowed in the array of ChoiceProps, I would think it would be okay.
> > My worry is more making sure there are no missing proptypes, or proptypes where we gave information about which field was allowed before and now give something ambiguous. Open to discussion here though. @nichia
> 
> The previous version of `arrayOf[custom]` would require us to add additional info/links at the comment area to reference the ChoiceProps.
> I like that this version display `ChoiceProps` - however, there are 21 Choice Props, not sure how that cell would look when all the props are included (right now, it's only showing 14 props)

@nichia  - These are specifically the ChoiceProps that you can pass into the component. Look here:https://github.com/CMSgov/design-system/pull/1041/files#diff-2c0d6b9485b31b860117c1d76c02727f7b6f0113daf3e71d7a798cd117b7da07R8

We filter out the other ChoiceProps because we will overwrite them with the fields from ChoiceLIst (ie, inversed). The other 7 aren't shown because passing them in will not do anything.

kimieann:
@nichia The other items should be addressed now also

nichia:
Doing a last check with coverage tools frontend and getting an issue with the `type` prop:
Screen Shot 2021-05-28 at 9 11 30 AM

packages/design-system/src/components/ChoiceList/__snapshots__/ChoiceList.test.tsx.snap
```diff
@@ -5,29 +5,11 @@ exports[`ChoiceList Radio buttons and Checkboxes is a radio button group 1`] = `
...
-        "disabled": undefined,
-        "inputClassName": "",
-        "inputRef": [Function],
-        "inversed": undefined,
         "label": "Choice 1",
-        "name": "spec-field",
-        "onBlur": undefined,
-        "onChange": undefined,
-        "size": undefined,
-        "type": "radio",
```
nichia, line 17: Hm, would this snapshot test change be related to the `type` prop issue on typescript app like coverage tools frontend?

nichia (approved):
Great work!!!

scheul93 (approved):

PR #1042, opened 2021-05-24
[WNMGDS-642] Add USWDS spacing design tokens

nichia:
## Summary
This PR adds USWDS spacing design tokens for margin/padding in the design system. 
Note: USWDS spacing tokens are documented in px, but they output values in rem which may break some sass calculation using mixed units of 'px' and 'rem' as a property value, which will produce an incompatible error. The fix is to change these property value to USWDS spacing token via the `units()` function

### Added
- Add USWDS as dependency
- Add USWDS sass and theme setting

### Changed
- Update the places where padding and margin is declared in the CMS Design System to use USWDS spacing design tokens

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-642/use-uswds-spacing-design-tokens/

- Test this branch on both the Mgov and HCgov child design systems (spacing-token branch) by updating the package.json to point to this branch. Ensure Child DS work as expected.
  - [Hcgov DS test branch: WNMGDS-642/use-uswds-spacing-tokens](https://github.com/CMSgov/hcgov-design-system/tree/WNMGDS-642/use-uswds-spacing-tokens)
  - [Mgov DS test branch: WNMGDS-642/use-uswds-spacing-tokens](https://github.com/CMSgov/mgov-design-system/tree/WNMGDS-642/use-uswds-spacing-tokens)

```
"@cmsgov/design-system": "file:../design-system/packages/design-system",
"@cmsgov/design-system-docs": "file:../design-system/packages/design-system-docs",
"@cmsgov/design-system-scripts": "file:../design-system/packages/design-system-scripts",
```


packages/design-system/src/styles/settings/variables/_layout.scss
```diff
@@ -44,20 +51,22 @@ $border-radius-pill: 9999px;
...
-$grid-gutter-width: $spacer-4 !default;
-$grid-form-gutter-width: $spacer-2 !default;
+$grid-gutter-width: units(4) !default;
+$grid-form-gutter-width: units(2) !default;
 
 // Z-index
 $z-dialog: 1000 !default;
 
 // Site variables
 $site-margins: 3rem !default;
```
line47, line 61:
```suggestion
$site-margins: units(6) !default;
```

packages/design-system/src/styles/settings/uswds/_index.scss
```diff
@@ -0,0 +1,81 @@
...
+*/
+
+// -------------------------------------
+// Import individual theme settings
+
+@import 'uswds-theme-general';
+@import 'uswds-theme-typography';
+@import 'uswds-theme-spacing';
+@import 'uswds-theme-color';
+@import 'uswds-theme-utilities';
```
line47, line 18: should we comment out the other imports other than spacing since we are not using colors, or typography, etc...

line47:
I was wondering if you were planning on adjusting things like `Width`, `Height` `Border-width`, etc... in this PR as well. I saw a few places where new USWDS spacing was used for things other than margin and padding but also wanted to ask so we don't make this PR too big

nichia:
My preference is to focus on updating the margin and padding to use uswds spacing token. Only update other variables if they result in mixed units errors.

PR #1047, opened 2021-05-25
[NO-TICKET] Put analytics behind feature flags

pwolfert:
## Summary
Put sending of analytics behind separate feature flags for each component that sends them while we get the issues worked out on the application side.

### Changed
- Analytics-sending is now behind feature flags for each component that sends them. They all OFF by default for now.

### DIFF
--- a/packages/design-system-docs/src/pages/components/Alert/_Alert.docs.scss
+++ b/packages/design-system-docs/src/pages/components/Alert/_Alert.docs.scss
@@ -191,13 +191,20 @@ Style guide: components.alert.guidance
 /*
 Google Analytics
 
-Analytics event tracking is disabled by default. To enable this feature, import and set the feature flag in your application's entry file:
+**Analytics event tracking is disabled by default.**
+
+### Enable event tracking
+
+- Import and set the `setAlertSendsAnalytics` feature flag to `true` in your application's entry file:
 ```
 import { setAlertSendsAnalytics } from "@cmsgov/<design-system-package>";
 setAlertSendsAnalytics(true);
 ```
-On application where the page has `utag` loaded, the data goes to Tealium which allows it to route to Google Analytics or the currently approved data analytics tools.
-- To disable the automatic event tracking on certain cases, pass this prop to the component
+On applications where the page has `utag` loaded, the data goes to Tealium which allows it to route to Google Analytics or the currently approved data analytics tools.
+
+### Disable event tracking
+
+- Pass the `onComponentDidMount` prop set to `false` to the component:
 ```
 analytics={
   {
@@ -205,7 +212,9 @@ analytics={
   }
 }
 ```
-- To override the event tracking, pass changes via the analytics prop
+### Override  event tracking
+
+- Pass changes via any of the [available analytics props](#components.alert.analytics).
 
 Style guide: components.alert.guidance-analytics
 */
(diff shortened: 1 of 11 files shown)


pwolfert:
I just tested this in PlanCompare, and it worked. Alerts and Dialogs did not send unless I turned the flag on with the flag-setting functions.

packages/design-system/src/components/Alert/Alert.jsx
```diff
@@ -51,10 +52,12 @@ export class Alert extends React.PureComponent {
             : '';
       }
 
-      sendAnalyticsEvent(
-        get(this.props.analytics, eventAction),
-        get(defaultAnalytics(this.eventHeadingText, this.props.variation), eventAction)
-      );
+      if (alertSendsAnalytics()) {
```
nichia, line 55: Ah, just noticed - Ok if I move ` if (alertSendsAnalytics()) {` further up the compponentDidMount() function?

packages/design-system/src/components/Dialog/Dialog.jsx
```diff
@@ -64,20 +65,24 @@ export class Dialog extends React.PureComponent {
           : '';
     }
 
-    /* Send analytics event for dialog open */
-    sendAnalyticsEvent(
-      get(this.props.analytics, eventAction),
-      get(defaultAnalytics(this.eventHeadingText), eventAction)
-    );
+    if (dialogSendsAnalytics()) {
```
nichia, line 68: Same for this

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -60,20 +61,24 @@ export class HelpDrawer extends React.PureComponent {
           : '';
     }
 
-    /* Send analytics event for helpdrawer open */
-    sendAnalyticsEvent(
-      get(this.props.analytics, eventAction),
-      get(defaultAnalytics(this.eventHeadingText), eventAction)
-    );
+    if (helpDrawerSendsAnalytics()) {
```
nichia, line 64: .. and for this

packages/design-system/src/components/Alert/Alert.jsx
pwolfert: Oh yep, thanks. I skimmed over that code above and didn't realize it was all for analytics. Good catch 👍

packages/design-system/src/components/Dialog/Dialog.jsx
pwolfert: There is _some_ risk that between `componentDidMount` and `componentWillUnmount` that the feature flag could be turned on, and then the class properties assigned here wouldn't exist. However, app teams can just be careful not to do this. I think it would be kind of weird.

line47 (approved):
Looks good to me

packages/design-system-docs/src/pages/components/Alert/_Alert.docs.scss
```diff
@@ -191,16 +191,32 @@ Style guide: components.alert.guidance
 /*
 Google Analytics
 
-On application where the page has `utag` loaded, analytics event tracking is enabled by default. The data goes to Tealium which allows it to route to Google Analytics or the currently approved data analytics tools.
-- To disable the automatic event tracking, pass this prop to the component
+**Analytics event tracking is disabled by default.**
```
line47, line 194: I updated the formatting of the content to use headers  making the sections easier to scan

PR #1052, opened 2021-05-26
[NO-TICKET] Output useful information for a11y violations

pwolfert:
## Summary
Our current suite of end-to-end tests includes a basic automated accessibility check on each component. I recently tried it out by intentionally lowering the contrast and verifying that the a11y tests would fail. Unfortunately the error did not actually give enough information to pinpoint what element was in violation of the color contrast rule. Before we were console-logging the top-level violations array, which doesn't expand objects deep enough to print useful information. I've added code to print each violation in a way that is legible and drills down enough to show the actual issues Axe found.

Relevant axe-core docs: https://github.com/dequelabs/axe-core/blob/master/doc/API.md#results-object

### Fixed
- Automated a11y tests now print useful information for debugging a11y violations

## How to test
Introduce a color-contrast issue by making white labels on a white background or something, and then run `yarn test:e2e`

### Before
Screen Shot 2021-05-26 at 9 27 21 AM

### After
Screen Shot 2021-05-26 at 9 22 19 AM


line47:
[I changed the text color for default buttons from `color-primary` to `color-white`](https://github.com/CMSgov/design-system/blob/pwolfert/format-axe-violations/packages/design-system/src/styles/components/_Button.scss#L9) on a white background but the a11y violation wasn't found when I ran  `yarn test:e2e` Is there something I'm missing? 

Screen Shot 2021-05-26 at 2 42 53 PM

pwolfert:
@line47, I would expect that to be caught. I made a similar CSS change in hc.gov ds, and it was caught. This will take some deeper investigation 🤔

nichia:
- I changed the text color for default buttons from $color-primary to $color-gray-light on a white background, 7 a11y violation was found when I ran yarn test:e2e. 
- However, if I changed the text color for default buttons from $color-primary to $color-white on a white background - looks like a11y cannot detect the white text on white background

I did notice the script failed with `TypeError: Cannot read property 'violations' of null` - not sure if that was intentional?
Screen Shot 2021-05-27 at 6 08 56 PM

nichia:
Ah, just checked that before this PR change, the `yarn test:e2e` would just throw the failed error of `TypeError: Cannot read property 'violations' of null`.

pwolfert:
@nichia and @line47, I wonder if that was an intentional feature on Axe's part to make the analyzer ignore equal text and background colors. Maybe it assumes we're intentionally trying to hide text in that scenario.

nichia:
That's what I was thinking too, that if this feature of Axe is to inspecting the visual page only, it would not discern text/image which has the same color as the background.

nichia (approved):
:rocket:

PR #1057, opened 2021-06-08
[WNMGDS-914] Add component examples to CRA and CRA-typescript

kimieann:
## Summary
Adding every component as an example to create-react-app and create-react-app-typescript to make testing of changes to components easier.

### Added
Examples folder under Components folder for both CRA and CRA-typescript example.

### Fixed
When adding the Accordion component to create-react-app-typescript, there was an error that no Accordion type was exported from the design system, which is why you see the fixes in the design-system package included.For this reason, when testing the create-react-typescript section, you will have to update the package.json to point to the local design-system and not the actual design-system for this PR.

## How to test
- cd to `examples/create-react-app`, yarn install, yarn build, yarn start - You should see a new Examples tab with an example of every component
- cd to `examples/create-react-app-typescript` - You can try the same steps and you'll see an error that there is no exported Accordion from the type folder. Change the package.json to point to the local design system package. Then run `(cd ../.. && yarn install && yarn build) && rm -rf node_modules && yarn install && yarn start` and it should compile and you should see examples tab


packages/design-system/src/components/Accordion/Accordion.tsx
```diff
@@ -31,7 +31,7 @@ const handleKeyDown = (e) => {
 };
 
 // eslint-disable-next-line react/prop-types
-const Accordion: FunctionComponent<AccordionProps> = ({ bordered, children, className }) => {
+export const Accordion: FunctionComponent<AccordionProps> = ({ bordered, children, className }) => {
```
kimieann, line 34: This and the change below were needed to be able to import the Accordion component into the typescript example

scheul93:
[question] both package.json files now have the `@cmsgov/design-system` dependency hard-coded to version `2.6.1` are there concerns about future releases not getting the latest design system updates because the version is hard-coded?

Or is the expectation that the version will have to be manually increased each time a design system update is made?

kimieann:
@scheul93 - No I just accidentally changed it to the current version of the design system when I was changing it back from the file:../ local pointing when I was testing. Most of the time we test these pointing to the local design system, so it really doesn't matter anyway but I changed back!

nichia (commented):
This is great to be able to test all the components via the CRA and CRA-typescript apps.
Aside from the minor changes noted, a couple of thoughts:
- Can we keep the existing CRA and CRA-typescript apps and rename the new apps for testing to CRA-test and CRA-typescript-test? This way the existing CRA and CRA-typescript apps which is referenced by our documentation would remain simple and internally for our testing, we can run the new apps?
- I noticed that using the editor `vscode`, formatting for the examples files are not aligned correctly. Can do a once-off `yarn lint` run (removing the `examples` folder from the ignoring paths) to fix the formatting
Screen Shot 2021-06-09 at 10 10 39 AM

kimieann:
@nichia - I ran the yarn lint and made the other two changes to remove the comma and add back centering.

Per our conversation on Friday, I think we should keep the examples in create-react-app and create-react-app-typescript, as if anyone does actually use the examples it will give them an example of how every component should be used.

nichia (commented):
The formatting looks good.

nichia (approved):

line47 (approved):
Nice work! 🚀

PR #1063, opened 2021-06-17
[WNMGDS-949] - alert component typescript conversion

scheul93:
[Link to ticket](https://jira.cms.gov/browse/WNMGDS-949)

## Summary
Converting the Alert component to typescript. I had to update a few of the Alert tests for typing.

I also updated some other tests that were failing.

## Outstanding Questions
* for the Analytics event object, should any of those properties be optional?

## How to test
* Build this project and confirm that there aren't any build errors
* I tested in mgov child system and locally linked and ran a build -- it was successful


packages/design-system/src/components/Alert/Alert.tsx
```diff
@@ -68,28 +136,33 @@ export class Alert extends React.PureComponent {
...
 
-  render() {
+  render(): React.ReactNode {
     const {
       children,
       className,
       autoFocus,
       heading,
-      headingId,
-      headingLevel,
```
nichia, line 89: Can keep the `headingId` and `headingLevel` const here, otherwise the spread operator for `{...alertProps}` will get passed these variables.
Also noticed that the current code is missing `analytics` variable here as well, can you help add that here as well.

nichia (commented):
Looks good, just noted a change below.

packages/design-system/src/components/Alert/Alert.tsx
nichia: `for the Analytics event object, should any of those properties be optional?` - good question, these properties would be optional for when applications passes these properties to the Alert component

nichia (approved):
:rocket:

line47 (approved):
built locally and on Mgov for me without issue!

PR #1064, opened 2021-06-21
[NO-TICKET] Remove analytics prop from modelProps

nichia:
## Summary
Add `analytics` to the list of props destructuring as we don't want to pass this prop to `modelProps` which uses the spread operator to get all the unspecified props.

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/ni/remove-prop-from-modalprops
Test that Dialog continue to work as per normal.


scheul93 (approved):
Good catch. Thanks for correcting this

line47 (approved):

PR #1067, opened 2021-06-24
[WNMGDS-933] github actions config for the pre-release scripts

scheul93:
## Summary
Part 1 of a series of PRs to automate parts of the release process. This PR is specifically setting up a new Github Actions workflow to do everything that the `pre-release.sh` script does. I went line by line through the original script and translated it into this .yml file. For now, this only replaces the `yarn pre-release` part of the RELEASE-PROCESS.md.

This workflow can be triggered manually from the 'Actions' tab, once merged.
Here's an example of what that looks like from a personal repo I was using for testing

Screen Shot 2021-06-24 at 10 55 49 AM

### Added
1. New github actions workflow called 'Pre-release'

## How to test
Unfortunately, I think we're going to have to test this PR by merging it and running the workflow manually and checking for errors. I tested some of it on one of my personal private repos by making sure that the syntax was correct and that the workflow was able to make and push branches.

## Next Steps
If this works, we can replicate on child systems any pre-release steps.

I also want to setup a separate workflow that will automatically publish to NPM once we publish a release on Github which make it so we don't have to manually run `yarn release`


.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          node-version: 13.7.0
+      - name: Install dependencies
+        run: yarn install
+      - name: Build project
+        run: yarn build
+  # Runs tests on project if build passes
+  test:
+    name: Running tests
+    needs: build
```
scheul93, line 29: This will only run if the build portion is successful

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,72 @@
...
+        with:
+          node-version: 13.7.0
+      - name: Run unit tests
+        run: yarn test:unit
+      - name: Run e2e tests
+        run: yarn test:e2e
+  # if build and tests pass, it begins pre-release steps
+  pre-release:
+    name: Pre-release steps
+    needs: test
```
scheul93, line 43: This will only run if the build & test portions are successful

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,72 @@
...
+          npm pack ./packages/design-system/
+          npm pack ./packages/design-system-docs/
+          npm pack ./packages/design-system-scripts/
+
+        # This saves .tgz files and artifacts in 'actions' tab
+      - name: Archive zips
+        uses: actions/upload-artifact@v2
+        with:
+          name: release-artifacts
+          path: ./*.tgz
```
scheul93, line 72: This section takes all the zipped files created in the previous step, and saves them as workflow artifacts that can be accessed in the 'Actions' tab when looking at a specific workflow run.

Screen Shot 2021-06-24 at 11 01 34 AM


Then, when we cut a Github release, we can pull the .tgz files from the workflow run and upload with the release

PR #1077, opened 2021-06-29
[WNMGDS-958] bash script for testing DS changes locally in other projects

scheul93:
## Summary
Based off a script that Kimberly used frequently to test design system changes in other applications, I created a bash script where it installs & build local dependencies for child system or application testing.

### Added
Script for local testing

## How to test
_Please note that this script assumes your core ds project, child systems, and applications all share a parent directory._

To test with script default, run `./test-locally.sh` in the root directory of the core DS. This should start a build that will
1. install and build core
2. install and build the mgov design system
3. install and build coverage tools project

This script also has the option to overwrite the app or child system paths when running the script. For example, if I wanted to test hcgov child system, I could run 
```
./test-locally.sh -c hcgov-design-system -a windowshop
```
where `-c` tells the directory name for the child system & `-a` tells the directory name for the application.

## Outstanding questions
1. Is this a helpful script?
2. What would make it more helpful?
3. Does the setup make sense?


test-locally.sh
```diff
@@ -0,0 +1,93 @@
...
+#   2. Run `./test-locally.sh -c hcgov-design-system
+
+# Similarly, if you want to change the application path your are testing with, you can:
+#   1. Update the APP_NAME path variable in this file
+#   2. Run `./test-locally.sh -a [directory name of application]
+#     a. If you only want to test in child systems and not in an application, you can run `./test-locally.sh -a ""`
+
+#!/bin/sh
+
+set -e
```
nichia, line 40: When I run `./test-locally.sh`, nothing happens unless I comment out this line 40

test-locally.sh
```diff
@@ -0,0 +1,93 @@
...
+# USING THIS SCRIPT
+# In the root directory of the design-system project, run `./test-locally.sh`
+#   This will install dependencies and run the core DS.
+#   Then, it will navigate to the child DS project. It will remove node_modules, install dependencies & build child system.
+#   Then, it will navigate to an application project. It will remove node_modules, install dependencies & build the application.
+
+# This script will try to use the default variables defined in this file, but you can override these for additional testing.
+# For example, if you want to test the hcgov child system, when the default CHILD_DS_NAME variable is mgov child system, you can:
+#   1. Update the CHILD_DS_NAME variable in this file
+#   2. Run `./test-locally.sh -c hcgov-design-system
```
nichia, line 31: Missing closing back quote:  2. Run `./test-locally.sh -c hcgov-design-system`

test-locally.sh
```diff
@@ -0,0 +1,93 @@
...
+echo "${GREEN}Building child design system with local dependencies at path $CHILD_DS_NAME...${NC}"
+cd ../$CHILD_DS_NAME
+rm -rf node_modules
+yarn install
+# yarn build
+
+if [ "$APP_NAME" != "" ]; then
+    echo "${GREEN}Building application with local dependencies at path $APP_NAME...${NC}"
+    cd ../$APP_NAME
+    nvm use
```
nichia, line 83: For applications without a .nvmrc file (eg. `windowshop` & `taxtools`), could we hard code nvm to a valide version eg `nvm use 10.22.1`? Otherwise it'll error out with:
```
error postcss@8.2.6: The engine "node" is incompatible with this module. Expected version "^10 || ^12 || >=14". Got "13.12.0"
error Found incompatible module.
````

**Review by nichia — COMMENTED** [review]
This script is super helpful for local test!

File: test-locally.sh (hunk 3 above)
Comment (line47) [reply test-locally.sh:83]: When I ran the script for coverage tools i ran into an issue 
```
N/A: version "10.16.0 -> N/A" is not yet installed.

You need to run "nvm install 10.16.0" to install it before using it.
```

Wondering if there is a way to prevent this for other applications or if we just add the missing version and move on?

File: test-locally.sh (hunk 1 above)
Comment (scheul93) [reply test-locally.sh:40]: 🤔 interesting... I copied this from the prerelease script, so I'll double check what it does and remove appropriately

**Comment by line47** [comment]
Do we need to be on a particular branch for the child system or application or do anything with that for setup?

File: test-locally.sh (hunk 3 above)
Comment (scheul93) [reply test-locally.sh:83]: That's a good question. I'm wondering if I can do some better error handling for errors... When I briefly googled this the other day, it made it seem like bash is not as robust as Javascript for error handling

**Comment by scheul93** [comment]
> Do we need to be on a particular branch for the child system or application or do anything with that for setup?

No. this script will run the child system & application for whatever branch you are currently on in those projects. The only real pre-req for those is that their package.json files have the link to the local dependency using `file:[relative path]`

**Comment by scheul93** [comment]
@nichia updated the script to now check if nvmrc exists. If it doesn’t script prompts for version number to use.

@line47 , I haven't found a good way to install the necessary version automatically. So right now, it will log that error you saw and exit the script

File: test-locally.sh [hunk 4]
```diff
@@ -0,0 +1,109 @@
...
+    else 
+        echo "${NC}No .nvmrc file found. Please enter node version you want to use for application build:"
+        read node_version
+        nvm use $node_version
+    fi
+    
+    rm -rf node_modules
+    yarn install
+    # This next line may need to be updated depending on the application project's local build command
+    yarn start
```
nichia, line 103: `coverage-tools-frontend` : yarn start is not working unless we remove one of the node_modules. We can add below code just before running `yarn start`:
```
    if [ “$APP_NAME” == “coverage-tools-frontend” ]; then
        rm -rf node_modules/@cmsgov/ds-medicare-gov/node_modules
    fi
```

nichia (approved):
:rocket:

line47 (approved):

Issue #1043, opened 2021-05-24
Consider options for stackable, responsive tables without column headers


Josh68:
I am working on a prototype in MCT with `Table` and its child elements. I'm liking it a lot but seeing room for more flexibility. One issue I'm running into is that I'm trying to mate it with [react-table](https://react-table.tanstack.com/), and if I want my first column to have no heading text, e.g., I get the error output defined here (in development): 

https://github.com/CMSgov/design-system/blob/f7642b0c7fcfec22f8d2b35e44457f7598c132b7/packages/design-system/src/components/Table/TableCell.jsx#L19

This is baked in because of the assumption that every cell in the `table-body` must have an associated `th` corresponding to that cell's column in the `thead`, if one exists. Also, the `data-title` attribute of a cell in the `table-body` is set to the content of that header. However, particularly in stacked, responsive mode, I want to just output the value of the row header, with no extra, blank space. I'm able to do this, but only with the warnings. I think there should be a way to do this without warnings, assuming it's a clear use-case, mostly (or only) for a column of row headers.

If I have a table where I don't want to show a `thead` at all, I don't add `Heading`s to any of the columns (in my `react-table` data). I still have a `thead` for referencing unique header IDs, but set it to `display:none`. The DS component then doesn't try to set a `data-title` on any of the cells in my `tbody`.

Please let me know if any of this isn't clear, thanks.

Screenshot showing the use-case, rough design:

Desktop layout
![image](https://user-images.githubusercontent.com/771447/119385365-e04b5f00-bc7a-11eb-9291-47d08a90851a.png)

Desired mobile layout (achieved, but with warnings in development mode):
![image](https://user-images.githubusercontent.com/771447/119385418-f22d0200-bc7a-11eb-9147-1ea9882d28b8.png)

line47:
https://jira.cms.gov/browse/WNMGDS-922

pwolfert:
Looks like the work was completed, but this issue wasn't automatically closed.

PR #1082, opened 2021-07-07
[WNMGDS-922] Update stackable table

nichia:
## Summary
Update stackable tables:
- disable console warning message of 'the header prop is required' and 'the stackedTitle prop is required' for a column of row headers (issue: #1043)
- disable console warning message of 'the id prop is required' for a column header without any children value

Re-instate the use of Context for Table component now that consuming applications have upgraded to newer version of React.

## How to test
- Check that http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-922/stackable-table-updates remains unchanged when compared to design.cms.gov

- Update the stackable table component example `Table.example.jsx`:
   - change line 52:  `<TableCell />`
   - change line 59 and 71:   `<TableCell component="th" scope="row">`

- Run `yarn start` on master and see console warnings
- Run `yarn start` on this branch and see there are no errors


scheul93:
Do you have links / could you explain why Context is needed for this table component? What is it allowing the Table to do now that it couldn't do before?

nichia:
Context simplifies the current multiple level of cloning or prop drilling of `_stackable` variable as it maps over an array of child components, passing down props to them using the React.cloneElement api. 
These components Table, TableBody, TableHead, TableRow current clones the `_stackable` variable.

scheul93 (approved):
It looks good to me. Thanks for teaching me about Context API and for using it to simplify some of these components

line47 (approved):
This is working for me

packages/design-system/src/components/Table/TableCell.jsx
```diff
@@ -13,18 +14,25 @@ export const TableCell = ({
...
-  if (process.env.NODE_ENV !== 'production' && _stackable) {
+  const stackableTable = useContext(TableContext);
+
+  let Component;
+  if (component) {
+    Component = component;
+  } else {
+    Component = _isTableHeadChild ? 'th' : 'td';
+  }
+  if (process.env.NODE_ENV !== 'production' && stackableTable) {
```
Josh68, line 27: @nichia, I don't have time to test locally with MPF at the moment, but I wonder if you can add a boolean prop (or 2, but I think 1 might be fine) that allows the users to just disable these warnings altogether. The reason I want to suppress them is that they can greatly slow things down during development, and I think we have cases where we may "violate the rules" because we need to implement things differently for a stacked table, particularly when headers contain JSX and not just simple strings of text.

So, I think these changes look fine, but would like to have the option to suppress warnings via (a) prop(s), if possible. Thanks

Josh68 (commented):
Adding a request, but don't want to block the merge if the team decides against it.

packages/design-system/src/components/Table/TableCell.jsx
nichia: Yes, we can add a disable warning prop which would helpful with the cases that you mentioned, especially when working with large datasets. One thing to note, since the id prop associates header and data cells for screen readers in stacked format, it's recommended that accessibility testing with screen readers is done to make sure we're not loosing out on that.

Josh68: Thanks. Hopefully after we get through our current work on Plan Details redesign, we can work on refining our implementation to ensure it's accessible, and I can test with the updated `Table`. If there are still issues to be worked out, we can address at that time. The point of the prop(s) would be to allow the user to acknowledge they know what they're doing and that they will do their best to ensure good accessibility. For that matter, you can annotate the prop(s) to say as much.

Josh68 (commented):
@nichia, I'm going to try to test this by updating latest Mgov DS and using a local version in MPF, but we've had bad luck with that in the past (blame our Docker setup?). I'll see what I can get tested and provide any feedback.

Note for reference: For tables that have no `rowheader`s  or `columnheader`s, cells are rendered without a `headers` attribute at all, per a decision by our UX team. Not sure how this affects various warnings. 

I was also never able to debug fully, but I got errors if I ever tried to conditionally render a child component, e.g. `TableHead`. So, in cases where it shouldn't be displayed, I resorted to setting `display:none` rather than not rendering. Again, this may affect warnings.

Josh68 (approved):
@nichia, I have just confirmed that running with the `test-locally` script and updating all of our tables to include the `warningDisabled` prop, warning are correctly suppressed in development with the inspector open. We will be continuing to review our implementation for A11Y, but this option will allow us to keep what we currently have and remove temporary overrides of `console.warn`. Thanks for the updates.

PR #1084, opened 2021-07-09
[WNMGDS-921] Add link variables for theming

line47:
## Summary
Adding variables to use to style links to make the design system more theme-able by child design system themes.

### Added
- Added component variables for styling default and inverse links

### Changed
- Changed previous component styles to use new component variables. For example, `color: $color-primary;` becomes `color: $link-color;`

## How to test
- Go to http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-921-add-link-theming and compare links to the live design.cms.gov website
- Test in child design system themes (there should be no visual impact)


packages/design-system/src/styles/base/_link.scss
```diff
@@ -1,17 +1,41 @@
...
+$link-underline-height-hover: 1px !default;
+$link-underline-offset: 0 !default;
+
 %link {
-  color: $color-primary;
+  color: $link-color;
   cursor: pointer;
   text-decoration: underline;
+  text-decoration-color: $link-color;
+  text-decoration-thickness: $link-underline-height;
```
scheul93, line 27: For my own education, how does the design system handle CSS properties that are not supported in legacy browsers. For example `text-decoration-thickness` [doesn't look to be supported in IE11](https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-thickness#browser_compatibility). Is that okay?

Like for consuming applications, do they know there might be some stylistic differences in IE for things not supported?

line47: This is a great question, we aim for graceful degradation and try to strike a balance between using modern practices where applicable while ensuring the design doesn't break. 

As far as what applications teams know or don't know - I'm unsure on that.

scheul93 (approved):

babsmiles (approved):

PR #1085, opened 2021-07-12
[WNMGDS - 967] autocomplete typescript conversion

scheul93:
## Summary
Converting the Autocomplete component (and relevant files) to typescript

### Changed
Updated the Autocomplete component and relevant files to typescript. As part of this, I had to refactor small bits of the component to be TS compliant. 

I also had to update one of the tests to test for prop defaults in a different way. This is because defaultProps and TS don't play nicely together. This also caused me to update the snapshot test for this component.

## How to test
Pull down this branch. Run `yarn start`. Confirm that there are not any build errors.
On the doc site, navigate to Components > Autocomplete. In the React example, confirm that there are not any runtime errors.

Additionally, all unit and e2e tests should still pass.


line47 (approved):
LGTM! Tested and ran locally

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -111,7 +234,7 @@ export class Autocomplete extends React.PureComponent {
           'aria-expanded': isOpen,
           'aria-labelledby': null,
           'aria-owns': isOpen ? this.listboxId : null,
-          autoComplete: this.props.autoCompleteLabel,
+          autoComplete: this.props.autoCompleteLabel || 'off',
```
nichia, line 237: Can remove the ` || 'off'` as it's now defined under static defaultProps

packages/design-system/src/components/Autocomplete/index.ts
```diff
@@ -0,0 +1,2 @@
+export { default as Autocomplete } from './Autocomplete';
+export { default as AutocompleteWrapperDiv } from './WrapperDiv';
```
nichia, line 2: Hm, do we want to expose the `AutocompleteWrapperDiv` to applications? It's currently just needed by the core ds Autocomplete right?

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -25,20 +24,128 @@ import { errorPlacementDefault } from '../flags';
...
 
-export class Autocomplete extends React.PureComponent {
-  constructor(props) {
+export class Autocomplete extends React.Component<AutocompleteProps, any> {
+  static defaultProps = {
+    ariaClearLabel: 'Clear search to try again',
+    autoCompleteLabel: 'off',
+    clearInputText: 'Clear search',
+    clearSearchButton: true,
+    itemToString: (item) : string => (item ? item.name : ''),
```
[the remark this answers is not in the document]
scheul93: In this situation, my intention was for `(item) : string` to say that the whole function returns a value of type string.  I shared your concern that `item` could be anything so right now, the `item` parameter does not have a type defined for it.

Here's the resource I referenced: https://www.tutorialsteacher.com/typescript/arrow-function
Does that make sense? Or should I still try to define some sort of type for item? My argument is that `any` doesn't provide a lot more info than the way it is currently written, but open to suggestions.

packages/design-system/src/components/Autocomplete/index.ts
scheul93: Will it still generate the type definition if it is not exported here? Only asking because I deleted the type file previously found at `packages/design-system/src/types/Autocomplete/WrapperDiv.d.ts`

Although... maybe I should keep the naming consistent?

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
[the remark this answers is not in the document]
nichia: Ah, thanks for the reference and explanation, you're right, looks like I'm getting my typescript syntax mixed up.

packages/design-system/src/components/Autocomplete/index.ts
nichia: Just tried removing the `export WrapperDiv` line from `packages/design-system/src/components/Autocomplete/index.ts` and looks like it works as the WrapperDiv typescript definitions is being generated by the cmsds scripts (eg. check for `packages/design-system/dist/types/Autocomplete/WrapperDiv.d.ts)

nichia (approved):

PR #1088, opened 2021-07-19
[WNMGDS-977] Table typescript conversion

nichia:
## Summary
Convert Table component to typescript

### Changed
- Update the Table component and test files to typescript.
- Remove the `<div>` wrapper from root table element for non-scrollable table. The `<div>` wrapper is needed for scrollable table only

## How to test
- Check http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-922/stackable-table-updates to confirm table component works as expected


packages/design-system/src/components/Table/TableCell.tsx
```diff
@@ -1,9 +1,65 @@
...
+  stackedClassName?: string;
+  /**
+   * Table data cell's corresponding header title, this stacked title is displayed as the row header
+   * when a responsive table is vertically stacked.
+   */
+  stackedTitle?: string;
+  /**
+   * @hide-prop This gets set from the parent `TableHead` component
+   */
+  _isTableHeadChild?: boolean;
```
zarahzachz, line 53: Seeing `_stackable?: boolean;` from the TableCell definition. Not sure if it's needed here.

packages/design-system/src/components/Table/TableHead.tsx
```diff
@@ -1,9 +1,19 @@
-import PropTypes from 'prop-types';
 import React from 'react';
 
-export const TableHead = ({ children, ...tableHeadProps }) => {
+export interface TableHeadProps {
+  /**
+   * The table head contents, usually `TableRow`.
+   */
+  children?: React.ReactNode;
```
zarahzachz, line 7: Not sure if `_stackable?: boolean;` was needed here. I see it in the definition file.

packages/design-system/src/components/Table/TableRow.tsx
```diff
@@ -1,9 +1,23 @@
...
-export const TableRow = ({ children, _isTableHeadChild, ...tableRowProps }) => {
+export interface TableRowProps {
+  /**
+   * The table row contents, usually `TableCell`.
+   */
+  children?: React.ReactNode;
+  /**
+   * @hide-prop This gets set from the parent `TableHead`
+   */
+  _isTableHeadChild?: boolean;
```
zarahzachz, line 11: `_stackable?: boolean;` is in the definition file but not here. Not sure if it's needed.

zarahzachz (commented):
I had a few comments but also, do we need to delete the Table definition files after they've been converted to TSX?

packages/design-system/src/components/Table/TableBody.tsx
```diff
@@ -1,7 +1,17 @@
-import PropTypes from 'prop-types';
 import React from 'react';
 
-export const TableBody = ({ children, ...tableBodyProps }) => {
+export interface TableBodyProps {
+  /**
+   * The table body contents, usually `TableRow`.
+   */
+  children?: React.ReactNode;
```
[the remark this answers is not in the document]
nichia: `_stackable` has been replaced with react context api [PR WNMGDS-922](https://github.com/CMSgov/design-system/pull/1082)

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -58,27 +121,31 @@ export class Table extends React.PureComponent {
...
-  componentWillUnmount() {
+  componentWillUnmount(): void {
     if (this.props.scrollable) {
       window.removeEventListener('resize', this.debounceHandleResize);
     }
   }
 
-  handleResize() {
+  captionID: string;
+  container: any;
```
scheul93, line 138: [question] could this variable have a type other than `any`. Looking through the code, it looks like it could be `boolean | React.RefObject<HTMLDivElement>`

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -93,7 +160,7 @@ export class Table extends React.PureComponent {
     });
   }
 
-  render() {
+  render(): React.ReactNode {
```
scheul93, line 163: Based on the [original type definition file](https://github.com/CMSgov/design-system/pull/1088/files#diff-3afd608cb36c07146ec6a790493d6b24b8e1341d6661aed7498a1ae5be468f5aL59), should this be `JSX.Element`?

packages/design-system/src/components/Table/TableCaption.tsx
```diff
@@ -1,15 +1,37 @@
...
+   * @hide-prop This gets passed from the parent `Table` component when the table `scrollable` prop is set.
+   */
+  _scrollActive?: boolean;
+  /**
+   * @hide-prop This gets passed from the parent `Table` component when the table `scrollable` prop is set.
+   */
+  _scrollableNotice?: React.ReactNode;
+}
+
+export const TableCaption: React.FC<TableCaptionProps> = ({
```
scheul93, line 27: [question] in what situation does the component need to omit the `children` prop (like in TableRow component) vs this component where it is not omitted?

nichia: In the case of `TableCaption` we don't use `React.ComponentPropsWithRef` and just refer to TableCaptionProps, so there's no need to omit any attributes.

For components which has `React.ComponentPropsWithRef` or `React.ComponentPropsWithoutRef` of an html element, and if the html element has the same attribute as what's on our component's prop (eg. TableRowProps has children), then we want to omit these attribute and use ours instead. 

Also if we don't omit `children` in the case of TableRow, the children prop (see below image) will not show up on the props table of our document site.
Screen Shot 2021-07-20 at 1 11 35 PM

packages/design-system/src/components/Table/Table.tsx
nichia: I've noticed that our other typescript conversion uses `React.ReactNode` and have used the same. Also this [slackoverflow](https://stackoverflow.com/questions/58123398/when-to-use-jsx-element-vs-reactnode-vs-reactelement#58123882) has more info.

packages/design-system/src/components/Table/Table.tsx
scheul93: Typing `container` led to an error on line 142. Leaving `any` for now

scheul93 (approved):

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -44,7 +107,7 @@ export class Table extends React.PureComponent {
       if (
         props.scrollable &&
         Array.isArray(props.children) &&
-        !props.children.some((child) => isTableCaption(child))
+        !props.children.some((child: React.ReactElement) => isTableCaption(child))
```
Josh68, line 110: @nichia, I have noticed issues with `Table` where if I want to have an optional child, like `TableCaption`, I get errors, because assumptions are made that all children are `React.ReactElement`s. I just tried having a function that returns properly formatted markup inside a `TableCaption` optionally return `null` if there's no caption content to render, and this blows up my table. Instead, for every child, I've had to optionally return a "shell" component with `.ds-u-display--none`.

I understand that an HTML `table` can only have prescribed child nodes, in the correct order, but in React, we should be able to accept and bypass `null` for conditionally rendered children. What do you think?

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -58,27 +121,31 @@ export class Table extends React.PureComponent {
...
+  handleResize(): void {
     const { scrollWidth, clientWidth } = this.container;
     const scrollActive = scrollWidth > clientWidth;
     this.setState({ scrollActive });
   }
 
-  renderChildren() {
-    return React.Children.map(this.props.children, (child) => {
+  renderChildren(): React.ReactNode[] {
+    return React.Children.map(this.props.children, (child: React.ReactElement) => {
```
Josh68, line 148: Referring to https://github.com/CMSgov/design-system/pull/1088/files#r673582273, it's somewhere in this block where I start getting errors if I try to optionally render any child of `Table`.

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -1,19 +1,70 @@
...
+   */
+  compact?: boolean;
+  /**
+   * @hide-prop [Deprecated] Use compact instead.
+   */
+  dense?: boolean;
+  /**
+   * Applies a horizontal scrollbar and scrollable notice on `TableCaption` when the `Table`'s contents exceed the container width.
+   */
+  scrollable?: boolean;
```
Josh68, line 33: Should optional boolean props have a default? This allows you to avoid casting to a boolean in code.

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -126,85 +185,42 @@ export class Table extends React.PureComponent {
...
     };
+    const contextValue = { stackable: !!stackable, warningDisabled: !!warningDisabled };
 
-    return (
+    return scrollable ? (
       <div
         ref={(container) => {
           this.container = container;
         }}
+        aria-live="polite"
```
Josh68, line 206: Can you annotate the reasons for adding `aria-live` and `aria-relevant` here? Is it because the scroll container is added and scrollbars forced to show? In my component that wraps the DS `Table`, I add aria-live for `stackable`, because it rearranges row headers.

packages/design-system/src/components/Table/TableBody.tsx
```diff
@@ -0,0 +1,24 @@
+import React from 'react';
+
+export interface TableBodyProps {
+  /**
+   * The table body contents, usually `TableRow`.
```
Josh68, line 5: Can a `tbody` legally have any direct descendants other than a `tr`? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/tbody

packages/design-system/src/components/Table/TableHead.tsx
```diff
@@ -1,15 +1,23 @@
-import PropTypes from 'prop-types';
 import React from 'react';
 
-export const TableHead = ({ children, _stackable, ...tableHeadProps }) => {
+export interface TableHeadProps {
+  /**
+   * The table head contents, usually `TableRow`.
```
Josh68, line 5: Same question as https://github.com/CMSgov/design-system/pull/1088/files#r674172226

packages/design-system/src/components/Table/TableRow.tsx
```diff
@@ -0,0 +1,38 @@
+import React from 'react';
+
+export interface TableRowProps {
+  /**
+   * The table row contents, usually `TableCell`.
+   */
+  children?: React.ReactNode;
```
Josh68, line 7: Essentially the same question as https://github.com/CMSgov/design-system/pull/1088/files#r674172226. A `tr`s only permitted children are supposed to be `<th>` or `<td>`, also `<script>` and `<template>`.

packages/design-system/src/components/Table/TableRow.tsx
```diff
@@ -0,0 +1,38 @@
...
+  _isTableHeadChild?: boolean;
+}
+
+type OmitProps = 'children';
+
+export const TableRow: React.FC<
+  Omit<React.ComponentPropsWithoutRef<'tr'>, OmitProps> & TableRowProps
+> = ({ children, _isTableHeadChild, ...tableRowProps }: TableRowProps) => {
+  const renderChildren = () => {
+    return React.Children.map(children, (child: React.ReactElement) => {
```
Josh68, line 20: Am I wrong, or are you here getting more specific about the allowed types in `children`, since `React.ReactNode` also includes things like strings.

Josh68 (commented):
@nichia, this looks good. I left some questions and comments.

nichia:
Thanks @Josh68 for reviewing.

packages/design-system/src/components/Table/TableRow.tsx
nichia: Yes, permitted children are supposed to be <th> or <td>, also <script> and <template>. 
Do you think the current type definition for `children` is appropriate?

packages/design-system/src/components/Table/TableBody.tsx
nichia: Yes, the MDN states that it should be `tr` only. I find that if a html page with `tbody` that contains children other then `tr`, console errors/warning are usually shown

packages/design-system/src/components/Table/Table.tsx
nichia: Hi @Josh68 - thanks for pick up the discrepancies. I'm not that familiar with typescript, what type definition would you suggest we use for `child` & `children`?

packages/design-system/src/components/Table/TableRow.tsx
nichia: Thanks for picking this up

PR #1089, opened 2021-07-19
Feature/pagination

line47:
- https://jira.cms.gov/browse/WNMGDS-930
- https://jira.cms.gov/browse/WNMGDS-1037
- https://jira.cms.gov/browse/WNMGDS-1038
- https://jira.cms.gov/browse/WNMGDS-1054
- https://jira.cms.gov/browse/WNMGDS-1090
- https://jira.cms.gov/browse/WNMGDS-1093
- https://jira.cms.gov/browse/WNMGDS-1095
- https://jira.cms.gov/browse/WNMGDS-1096

### Added
- New pagination component

## How to test
Demo URL
http://design-system-demo.s3-website-us-east-1.amazonaws.com/feature/pagination/components/pagination/


packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,187 @@
+import Ellipses from './Ellipses'
+import Page from './Page'
+import React, {useState} from 'react'
+
+const overflow = 1;
+const maxVisiblePages = 7;
+
+function paginationBuilder(currentPage, totalPages) {
```
scheul93, line 8: [fix] there are a few scenarios where the pagination component is only populating 5 slots instead of 7. 
Screen Shot 2021-08-19 at 4 56 09 PM

I tried some different conditionals in this functions and came up with this to replace the conditionals currently here:
```js
// The number of slots needing to be calculated - so total - 2 since we now the first slow will always be 1 and the last will always be total pages
const availableSlots = maxVisiblePages - 2;
/**
   * if the current page is < 5, add 1 - 5 pages
   */
  if (currentPage < availableSlots) {
    start = 1;
    end = availableSlots;
  }

  /**
   * if current page is total pages - 1, make sure start begins one earlier
   */
  if (currentPage === totalPages - 2) {
    start -= 1;
  }

  /**
   * if end page is two from the end, make sure the last page shows instead of ellipsis
   */
  if (end === totalPages - 2) {
    end += 1;
  }

  /**
   * If `end` > `totalPages`, add last pages to `range[]`
   */
  if (end >= totalPages) {
    start = totalPages - (availableSlots - 1); 
    end = totalPages;
  }

  /**
   * If `totalPages` is 5 or fewer, all pages added to `range[]`
   */
  if (totalPages <= maxVisiblePages) {
    start = 1;
    end = totalPages;
  }
```

scheul93 (commented):
So far, really good job with the pagination logic. This is more complicated than it seemed! I left some comments / questions / suggestions

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,187 @@
...
+              xmlns="http://www.w3.org/2000/svg"
+              viewBox="0 0 320 512"
+            >
+              <path
+                fill="currentColor"
+                d="M34.52 239.03L228.87 44.69c9.37-9.37 24.57-9.37 33.94 0l22.67 22.67c9.36 9.36 9.37 24.52.04 33.9L131.49 256l154.02 154.75c9.34 9.38 9.32 24.54-.04 33.9l-22.67 22.67c-9.37 9.37-24.57 9.37-33.94 0L34.52 272.97c-9.37-9.37-9.37-24.57 0-33.94z"
+              />
+            </svg>
+          </span>
+          Previous
```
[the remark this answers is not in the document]
line47: That's a good call

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  return (
+    <nav 
+      className="ds-c-pagination" 
+      role="navigation" 
+      aria-label="Pagination Navigation"
+    >
+      { currentPage !== 1 && (
+        <a
```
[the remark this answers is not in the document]
line47: I agree - I'll look into how to address this stylistically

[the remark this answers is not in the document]
line47: I'm thinking we could use - `visibility: hidden; to hide the previous/next buttons or other elements. From what I understand that would also be effective for screen readers as well

packages/design-system/src/components/Pagination/Pagination.tsx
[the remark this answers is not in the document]
zarahzachz: I like it - adding that logic!

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,187 @@
...
+  }
+
+  for (let i = start; i <= end; i++) {
+    range.push(i);
+  }
+  
+  return range;
+}
+
+export default function Pagination({compact, customUrl, page, totalPages}) {
```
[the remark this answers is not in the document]
zarahzachz: Love it! Implementing!

packages/design-system/src/components/Pagination/Pagination.tsx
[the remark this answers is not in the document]
zarahzachz: I agree this could be solved with style updates, but I recommend holding off on implementing any fix just yet - there's some a11y stuff I need to chat with DK about once this wraps that might change the markup.

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,239 @@
...
+  /**
+   * Class to be applied to parent `<nav>` element of Pagination component. Optional.
+   */
+  className?: string,
+  /**
+   * Renders compact layout. Optional.
+   */
+  compact?: boolean,
+  /**
+   * Defines initial active page in Pagination. Optional.
```
scheul93, line 16: [suggestion] Consider updating this to `Defines active page in Pagination. Optional` since the component is not stateless and there is no "initial" active page

zarahzachz: Like it - changing!

packages/design-system/src/components/Pagination/Pagination.tsx
zarahzachz: Thanks for wiring up this logic!

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,280 @@
...
+    paginationRange.push(i);
+  }
+  
+  return paginationRange;
+}
+
+function Pagination({className, compact, currentPage, customUrl, onPageChange, leftLabel, rightLabel, totalPages}: PaginationProps): React.ReactElement {
+  const pageChange = React.useCallback(page => (evt: React.MouseEvent) => onPageChange(evt, page), [onPageChange])
+  
+  const pageRange = paginationBuilder(currentPage, totalPages)
```
scheul93, line 109: [question] if the component is in mobile or compact mode is it worth skipping this paginationBuilder calculation since that info is not used?

I also understand that the calculation isn't very intensive, but just an idea.

packages/design-system/src/components/Pagination/Page.tsx
```diff
@@ -0,0 +1,46 @@
...
+  onPageChange?: (evt: React.MouseEvent) => void,
+}
+
+export default function Page({ customUrl, index, isActive, onPageChange }: PageProps): React.ReactElement {
+  return (
+    <li>
+      { isActive ? (
+        <span
+          className="ds-c-button ds-c-button--transparent ds-c-pagination__current-page"
+          aria-label={`current page, page ${index}`}
```
davidakennedy, line 28: I'm wondering why have the `aria-label` in addition to the `aria-current`? They're both doing the same job. Although, in my screen reader testing, they don't seem to conflict.

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,294 @@
...
+          index={totalPages}
+          isActive={currentPage === totalPages}
+          onPageChange={pageChange(totalPages)}
+        />
+      );
+    }
+  }
+
+  return (
+    <nav className={classes} role="navigation" aria-label="Pagination Navigation">
```
davidakennedy, line 216: A few thoughts here:

`aria-label="Pagination Navigation"`: Let's remove "Navigation" since most screen readers will announce the `<nav>` element as navigation by default.

`aria-label="Pagination`: What do you think about making this a prop? We don't know what type of pagination will be created with this component. But sometimes, a more detailed label could be helpful. Like "Plan pagination" for Healthcare.gov or Medicare.gov.

davidakennedy: One more thought on this part, `role="navigation"`: I'm leaning toward removing this. I know this came up recently in our accessibility channel though. It's good for older browsers/assistive tech. But I know the W3C HTML validator lists it as a warning if you put roles on HTML5 elements that already map to that role.

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,294 @@
...
+      {currentPage !== 1 && (
+        <Button
```
davidakennedy, line 218: It looks like in the output, this is a link with a `role="button"`. Is that happening because of the button component? I seem to remember something related to that.

davidakennedy: I think it might be good to add a prop here for an `aria-label` because "Next" and "Previous" aren't specific enough in most cases. Maybe the default is "Next page" and "Previous page", but teams could change that as necessary.

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,294 @@
...
+          {leftLabel}
+        </Button>
+      )}
+
+      {isMobile || compact ? (
+        <span className="ds-c-pagination__page-count">
+          Page <strong>{currentPage}</strong> of <strong>{totalPages}</strong>
+          <span
+            className="sr-only"
+            aria-live="assertive"
```
davidakennedy, line 249: The Axe checker throws an error here.  See: https://dequeuniversity.com/rules/axe/4.3/aria-allowed-attr I think because of the combination of `aria-live` and `aria-label`. It looks like you're trying to make the wording more specific, yeah? I think Page # of # is enough. Why the `aria-live`? Is it because this will likely update without a page refresh?

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,294 @@
...
+
+  /**
+   * `useState` and `useEffect` determine if
+   * mobile layout of component is rendered.
+   */
+
+  const [isMobile, setIsMobile] = React.useState(false);
+  React.useEffect(() => {
+    // Mobile media query derived from: https://design.cms.gov/guidelines/responsive/
+    const media = window.matchMedia('(max-width: 543px)');
```
LouisFettet, line 127: Just a heads-up, I think that this line might prevent us from rendering the component server-side. To remedy this, we should check the existence of `window` beforehand; if it doesn't exist we should default to the desktop or mobile layout (either one is probably fine).

packages/design-system/src/components/Pagination/Page.tsx
zarahzachz: I think the ARIA label was added to make the readout nicer ("page something" instead of just "number of number"). But I do wonder if a `<span>` is the correct markup here? Should the current page be able to receive focus on tab or is that unnecessary? There's also a very likely chance I'm just not using VoiceOver correctly here.

packages/design-system/src/components/Pagination/Pagination.tsx
zarahzachz: Re: `role="button"`, it does look like [that's baked into our Button component](https://github.com/CMSgov/design-system/blob/development/packages/design-system/src/components/Button/Button.tsx#L130). 

Good catch on the `aria-label` on nav elements!

packages/design-system/src/components/Pagination/Pagination.tsx
zarahzachz: `aria-live` was me goofin'. I've refactored it to just include ARIA label on the parent span element.

packages/design-system-docs/src/pages/components/Pagination/_Pagination.docs.scss
```diff
@@ -46,8 +69,15 @@ The following Sass variables can be overridden to theme alerts:
 - `$pagination-page-count` - Used to set the page count element text color.
 
 ### Accessibility
-#### Keyboard support
 
+- Use a wrapping `<nav>` element to identify Pagination as a navigation section.
+- Pagination **must** have an ARIA label attribute defined on the `<nav>` element that describes its purpose.
+- Use an unordered list for the navigation items as this allows screen readers to voice the number of elements in the Pagination component.
```
scheul93, line 75: [suggestion] this documentation says to use an `unordered` list, but the actual HTML is using an ordered list element. Consider updating either the documentation or the HTML to be consistent

pwolfert:
When looking at whether or not this would be able to replace the component that the Enroll team uses in their products, I see that unfortunately it does not. This component requires that there be a known page count, but there are use cases where we do not know how many pages there are—just that there are more results.

pwolfert:
Another thing I was looking at with regards to pagination in Enroll products is whether this component can to visually match the width of its container like so:
Screen Shot 2021-09-13 at 2 54 13 PM

It doesn't look like it does that right now. That's a design requirement we currently have.

packages/design-system/src/components/Pagination/Page.tsx
davidakennedy: I tested this a bit more, and realized that this part isn't getting announced in most browser/screen reader combinations we test in: `aria-label={`on page ${currentPage} of ${totalPages}`}`

Axe also [throws an error here](https://dequeuniversity.com/rules/axe/4.3/aria-allowed-attr). It's related to using `aria-label` on an element without a `role`. For more on this, [check out this post](https://www.davidmacd.com/blog/does-aria-label-override-static-text.html).

It's announced in Safari/VoiceOver, but not in Edge/Jaws, Chrome/Jaws or Firefox/NVDA. If we want to keep this copy for screen readers, I think a better approach would be to nest a span with some hidden text:

```<span className="ds-u-visibility--screen-reader">On </span>```

**Comment by pwolfert** [comment]
Regarding the above comment about expansion to fill the parent container, it can be solved by adding `className="ds-u-display--flex ds-u-justify-content--between"`, but I wonder if it should be the default behavior.

File: packages/design-system-docs/src/pages/components/Pagination/_Pagination.docs.scss [hunk 13]
```diff
@@ -0,0 +1,83 @@
...
+
+- `$pagination-current-page-color` - Used to set the current page element text color.
+- `$pagination-current-page-color-bg` - Used to set the current page element baclground color.
+- `$pagination-overflow-color` - Used to set the overflow element text color.
+- `$pagination-page-count-color` - Used to set the page count element text color.
+
+### Accessibility
+
+- Use a wrapping `<nav>` element to identify Pagination as a navigation section.
+- Pagination **must** have an ARIA label attribute defined on the `<nav>` element that describes its purpose.
```
davidakennedy, line 74: You really only need the `aria-label` on the `nav` element if there are more than one `nav` elements on a page. Chances are with this component, there will already be another nav of some kind. Like a main navigation. Should we be more specific? I'm leaning toward yes.

packages/design-system-docs/src/pages/components/Pagination/_Pagination.docs.scss
```diff
@@ -0,0 +1,83 @@
...
+- Use an unordered list for the navigation items as this allows screen readers to voice the number of elements in the Pagination component.
+- Use `aria-current="page"` on the current page's element to properly voice the current page for screen readers.
+- Voice the word "page" before the page numbers.
+- If *Previous* navigation link has keyboard focus, pressing <kbd>TAB</kbd> key **must** move focus to the first pagination page link.
+- If pagination page has keyboard focus, pressing <kbd>TAB</kbd> key **must** move focus to the next pagination page link.
+- If last pagination page has keyboard focus, pressing <kbd>TAB</kbd> key **must** move focus to the *Next* pagination navigation link.
```
davidakennedy, line 80: Does this refer to the pages that are links or the "Next" link at the end?

packages/design-system-docs/src/pages/components/Pagination/_Pagination.docs.scss
```diff
@@ -0,0 +1,83 @@
...
+- `$pagination-nav-link-color-hover` - Used to set color of next/previous and page number links on hover.
+- `$pagination-nav-link-color-active` - Used to set color of next/previous and page number links when active.
+- `$pagination-nav-link-color-focus` - Used to set color of next/previous and page number links on focus.
+
+- `$pagination-current-page-color` - Used to set the current page element text color.
+- `$pagination-current-page-color-bg` - Used to set the current page element baclground color.
+- `$pagination-overflow-color` - Used to set the overflow element text color.
+- `$pagination-page-count-color` - Used to set the page count element text color.
+
+### Accessibility
```
davidakennedy, line 71: Is it worth adding a guideline in here about shifting focus to a `h1` or other element after the links are clicked? It's tangental to the component, but could be helpful.

davidakennedy:
@zarahzachz Gave this another test. Left one related comment on the code and a few on the docs.

Great job on the first pass of the docs! This is shaping up real nice.

line47:
@pwolfert related to Enroll being able to use the pagination - this design requirement was not brought up in our original audit so I'm unsure if that requirement is still valid or not. We might want to check in with our stakeholders on that. We also did decide to backlog some functionality to get the basic component out and get feedback before we iterate or increase complexity. https://app.mural.co/t/cmsoc4625/m/cmsoc4625/1624644307658/05926eba16cf00b7faff435ffe943d80e19fe2c5?sender=u7ecc05658ea6811619bd7413

packages/design-system-docs/src/pages/components/Pagination/_Pagination.docs.scss
zarahzachz: The last page in the visible pagination set. So if Pagination has 20 pages and 12 is active (accounting for ellipses, it'd be 1 ... 11 **[12]** 13 ... 20), if you're tabbing through this set, you'd need to tab from 20 to Next to see the next set of pages (1 ... 12 **[13]** 14 ... 20) - the idea being users should be able to tab from the last page to Next. Not sure if I wrote it to be that clear though

packages/design-system/src/components/Pagination/Page.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+describe('Page', () => {
+    const onPageChange = jest.fn();
+
+    it('should render interactive el if not current', () => {
+        const wrapper = shallow(<Page index={1} isActive={false} onPageChange={onPageChange} />);
+        expect(wrapper).toMatchSnapshot();
+    });
+
+    it('should render static el if not current', () => {
```
pwolfert, line 13: Should it say `if current`?

packages/design-system/src/components/Pagination/Page.tsx
```diff
@@ -0,0 +1,50 @@
...
+        <span
+          className="ds-c-button ds-c-button--transparent ds-c-pagination__current-page"
+          aria-current="true"
+        >
+          {index}
+        </span>
+      ) : (
+        <Button
+          variation="transparent"
+          href={customUrl ? `${customUrl}/${index}` : `#${index}`}
```
pwolfert, line 41: Due to some technical limitations, all the single-page apps in one of our child sites use hash routing, so this default url could be problematic. Do these have to be actual links with hrefs, or could they be buttons styled like links? @davidakennedy, are they better as links?

packages/design-system-docs/src/pages/components/Pagination/_Pagination.docs.scss
davidakennedy: Oh, I see now. That wasn't clear to me on my first pass as my brain read "Next" as just next, and not the "Next" link. Could be just me though! If no one else is confused, we're good.

packages/design-system/src/components/Pagination/Page.tsx
zarahzachz: My assumption is Pagination updates the URL, which causes the browser history stack to update, making link the correct option. I'm still digging into this

zarahzachz:
> Regarding the above comment about expansion to fill the parent container, it can be solved by adding `className="ds-u-display--flex ds-u-justify-content--between"`, but I wonder if it should be the default behavior.

@pwolfert @line47 would either of you be opposed to me adding the rules of these utility classes to our `.ds-c-pagination` class? Asking because using util classes with `!important` declared on components makes it difficult for our users to overwrite specific styles if needed.

pwolfert:
@zarahzachz

> would...you be opposed to me adding the rules of these utility classes to our .ds-c-pagination class?

Nope, not at all. Good idea 👍

packages/design-system/src/components/Pagination/Page.tsx
zarahzachz: Per DK
> In the case of the pagination component, I’d advocate for links over buttons. Because when you interact with these the route will change. You’re going to a place vs. performing an action.

So links it is!

zarahzachz: @pwolfert I just pushed a fix up for this href problem, though I'm not seeing it on our demo link. React code examples should have a prop for `renderHref` that takes a function.

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,344 @@
...
+          {startLabelText}
+        </span>
+      ) : (
+        <Button
+          variation="transparent"
+          href={renderHref(currentPage - 1)}
+          onClick={pageChange(currentPage - 1)}
+          aria-label={startAriaLabel}
+          style={{ visibility: currentPage === 1 ? 'hidden' : 'visible' }}
+          aria-hidden={currentPage === 1}
```
scheul93, line 289: [question] are these two lines still needed? If not, consider removing

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,344 @@
...
+          </span>
+        </span>
+      ) : (
+        <Button
+          variation="transparent"
+          href={renderHref(currentPage + 1)}
+          onClick={pageChange(currentPage + 1)}
+          aria-label={endAriaLabel}
+          style={{ visibility: currentPage === totalPages ? 'hidden' : 'visible' }}
+          aria-hidden={currentPage === totalPages}
```
scheul93, line 321: [question] are these two lines still needed? If not, consider removing

packages/design-system/src/components/Pagination/Pagination.tsx
zarahzachz: you are correct! removing!

packages/design-system/src/components/Pagination/Pagination.tsx
zarahzachz: yup!

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,344 @@
...
+          key={`page-${totalPages}`}
+          index={totalPages}
+          isActive={currentPage === totalPages}
+          onPageChange={pageChange(totalPages)}
+        />
+      );
+    }
+  }
+
+  const startIcon = (
```
scheul93, line 237: [musing] for both startIcon & endIcon, I wonder if these could be separated out into individual files, exported, and reused in other components.
Doesn't seem like something that needs to be done now, just a thought.

zarahzachz: agreed - i was kinda wishing we had an icon component in place for this because it would make the markup a little tighter

scheul93 (approved):
LGTM

packages/design-system/src/components/Pagination/Page.tsx
```diff
@@ -0,0 +1,50 @@
...
+   */
+  isActive: boolean;
+  /**
+   * A callback function used to handle state changes.
+   */
+  onPageChange?: (evt: React.MouseEvent) => void;
+  /**
+   * Defines application-specific routing in url for links.
+   */
+  renderPageHref: string;
```
pwolfert, line 20: I just noticed that this is a string. Since it's not a function, could this be named `href` instead?

packages/design-system/src/components/Pagination/Pagination.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+      it('should call onPageChange when a page is pressed', () => {
+        const wrapper = render({ currentPage: 2 }, true);
+        wrapper.find('ul').childAt(0).find(pageButtonSelector).simulate('click', {});
+
+        expect(onPageChange).toHaveBeenCalledTimes(1);
+        expect(onPageChange).toHaveBeenCalledWith(expect.anything(), 1);
+      });
+    });
+
+    describe('href', () => {
```
pwolfert, line 75: I like these two unit tests. Should there be one more for the href on one of the page buttons in the middle?

pwolfert (approved):
Aside from the above two suggestions, LGTM as well 👍

davidakennedy:
Retested in Axe, Wave and all the usual browser and screen reader combinations. LGTM.

One note: Do we want to advise teams on how to handle the contrast failures as a result of the disabled state text? That will come up in any automated testing done with Axe in the pipeline. We can figure this out in a future iteration.

packages/design-system/src/components/Pagination/Page.tsx
zarahzachz: Updated!

PR #1090, opened 2021-07-20
[NO-TICKET] Add Husky to lint files on commit

line47:
## Summary
Adding husky and removing lint-staged to use for linting files on commit. 

### Added
- husky npm package - You can use it to lint your commit messages, run tests, lint code, etc... when you commit or push. Husky supports all Git hooks.
- linting for - typescript, javascript, Sass

### Removed
- Removed the previous husky and lint-staged from package.json as it's no longer needed

## How to test
Make a commit with some sort of linting issue - double quotes would work or ALL CAPS for a SCSS classname - then try and commit. The linter could catch the file that needs linting on commit but won't prevent a commit from happening.

### DIFF
--- a/.husky/pre-commit
+++ b/.husky/pre-commit
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+yarn lint
--- a/package.json
+++ b/package.json
@@ -11,13 +11,13 @@
     "build": "yarn cmsds build-docs",
     "pre-release": "./prerelease.sh",
     "release": "yarn lerna publish from-git --dist-tag latest",
-    "precommit": "lint-staged",
     "test": "yarn test:unit && yarn test:e2e",
     "test:unit": "yarn cmsds test ./packages --filePath",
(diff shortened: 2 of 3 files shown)


zarahzachz (approved):
Looks great!

scheul93:
[question] I followed your testing suggestions and if I'm understanding correctly, husky doesn't prevent the commit from happening, but could it...?

Not sure how @zarahzachz feels about this, but I feel like even though it could be frustrating, I'd rather have husky either 1) Tell me there's a linting error and make me fix it before committing or 2) make the linting update and commit it for me automatically.

Not sure if either of those are possible or even good ideas, so interested on your thoughts. (The actual file changes look good though 😄 )

line47:
In doing some exploration with @zarahzachz we discovered that the yarn lint utilized some CMSDS scripts and the husky and lint-staged version of linting do not. More exploration is needed to figure out what CMSDS scripts are doing and if we can ties that into the husky and lint-staged process... 

Closing this PR until we have some more exploration done

PR #1092, opened 2021-07-26
[WNMGDS 968] convert button component to typescript

zarahzachz:
## Summary
Converting `<Button>` from JS to TS.

### Changed
- Changed file extension from `.jsx` to `.tsx` in docs
- Changed file extension from `.jsx` to `.tsx` in unit test, removed `.jsx` from `<Button>` import
- Changed file extension from `.jsx` to `.tsx` in snapshot test
- `<Button>` types imported from component file instead of type definition file
- Added `onClick` to `OmitProps` as our ButtonProps interface overrides the `onClick` prop defined on the button element

### Removed
- Button type definition file

## How to test
Navigate to [Button component](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-968/convert-button-to-typescript/) and confirm there are no console errors and documentation renders as expected.


packages/design-system/src/components/Button/Button.tsx
```diff
@@ -63,20 +120,24 @@ export class Button extends React.PureComponent {
...
+
     if (component === 'button' && this.props.href) {
       // If `href` is provided and a custom component is not, we render `<a>` instead
       component = 'a';
     }
     return component;
   }
 
-  classNames() {
+  classNames(): any {
```
scheul93, line 133: [fix] consider using a more specific type here. Looking at the return statement, this function returns a `string`

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -1,9 +1,66 @@
...
+   */
+  type?: 'button' | 'submit';
+  /**
+   * A string corresponding to the button-component variation classes.
+   * The `'danger'` variation is deprecated and will be removed in a future release.
+   */
+  variation?: 'primary' | 'danger' | 'success' | 'transparent';
+}
+
+type OmitProps = 'size' | 'onClick';
```
scheul93, line 52: [fix] related to our conversation about Omit... looking at the documentation site, some of the props aren't showing up. It seems like type, disabled, className and children are not showing up in the doc table, and a new prop (ref) is there.

Screen Shot 2021-07-27 at 10 13 23 AM


The missing props can be resolved by changing this line to
`type OmitProps = 'size' | 'onClick' | 'type' | 'disabled' | 'className' | 'children';`
Not sure how or why `ref` is there

zarahzachz: Good callout! 

I wonder if I need to check the props table when implementing a button link by default as well. I think under the hood, if you provide an `href` it turns `<button>` to an `<a>`, which would make it inherit the `HTMLAnchorElement` interface props instead, possibly introducing even more props to omit. I'm going to test this out to confirm.

zarahzachz: I resolved this by adding both 'button' and 'a' to the Omit arg and added all missing props from prop table to OmitProps type. Appear to have discovered a bug with how union types are rendered on the docs site and filed a bug to track the work on that.

scheul93 (approved):
Looks good to me!

PR #1093, opened 2021-07-29
[WNMGDS 834] button with analytics

scheul93:
## Summary
Adding optional Google Analytics to button component

### Added
Google Analytics data to button component that is sent when button is clicked

### Fixed
Updated typescript definitions for prop types

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-834/button-with-analytics

1. On line 17 of file `design-system/packages/design-system/src/components/flags.ts`, change value to `true` to turn on analytics for button component
2. Run the site locally and navigate to the Button component and the React example section
3. Ensure that you have the Omnibug Chrome extension installed. Click a button and ensure that the event is captured by the Omnibug extension
4. Ensure that values for `Event Action`, `Event Category` and `Event Label` are all what you'd expect
5. Check out the Button component documentation and ensure that all analytics information is available and understandable.


zarahzachz (approved):
Looks good! Ran it locally and saw the events just fine.

nichia (approved):
The documentation looks great and the events are firing correctly.

nichia:
I like how you simplify the collection of analytics `text` field to only contain a value if the child node is a string within 2 levels only.

nichia:
Question - is the intended event label for nested child is to default to `button text:` ? example below: 
Screen Shot 2021-07-30 at 12 06 24 PM

nichia:
The other components currently use [`ref`](https://github.com/CMSgov/design-system/blob/0597a5bbee07af4fc3caac7f1e83ac4cfdfefb55/packages/design-system/src/components/Dialog/Dialog.jsx#L64) to find the text. 
The text length for those components could be very long, so it only send the first 80 characters.

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -1,7 +1,43 @@
...
 import classNames from 'classnames';
 
+export type ButtonComponent = React.ReactElement<any> | any | ((...args: any[]) => any);
+export type ButtonSize = 'small' | 'big';
+export type ButtonType = 'button' | 'submit';
+/**
+ * A string corresponding to the button-component variation classes.
+ * The danger variation is deprecated and will be removed in a future release.
+ */
+export type ButtonVariation = 'primary' | 'danger' | 'success' | 'transparent';
```
scheul93, line 17: note to self: need to add `'secondary' | 'tertiary'` to this list once [PR](https://github.com/CMSgov/design-system/pull/1106/files) is merged

scheul93:
Converted this PR to a draft while we figure out timing of releasing this feature. I will move any TS typing to a separate PR and leave this PR to be button analytics-focused

PR #1097, opened 2021-07-29
Patch/2.7.2 alert and choice list

scheul93:
## Summary
Updating the Alert, ChoiceList, and Choice components

## How to test
Original type definition files can be found here:
- [Alert](https://github.com/CMSgov/design-system/blob/354cd70ef0f69e235459a705a8637d76e5638a54/packages/design-system/src/types/Alert/Alert.d.ts)
- [ChoiceList](https://github.com/CMSgov/design-system/blob/660d51251cd9c7936013ebeb119023e287ada9ce/packages/design-system/src/types/ChoiceList/ChoiceList.d.ts)
- [Choice](https://github.com/CMSgov/design-system/blob/660d51251cd9c7936013ebeb119023e287ada9ce/packages/design-system/src/types/ChoiceList/Choice.d.ts)

Pull down the branch locally and run a `yarn build`. Confirm that type definitions in `dist/types` matches the original

### DIFF
--- a/packages/design-system/src/components/Alert/Alert.tsx
+++ b/packages/design-system/src/components/Alert/Alert.tsx
@@ -1,5 +1,4 @@
 import { EVENT_CATEGORY, MAX_LENGTH, sendAnalyticsEvent } from '../analytics/SendAnalytics';
-// import PropTypes from 'prop-types';
 import React from 'react';
 import { alertSendsAnalytics } from '../flags';
 import classNames from 'classnames';
--- a/packages/design-system/src/components/Alert/Alert.tsx
+++ b/packages/design-system/src/components/Alert/Alert.tsx
@@ -26,9 +26,6 @@ export interface AnalyticsObjectShape {
   onComponentDidMount?: boolean | AnalyticsEventShape;
 }
 
-// Omit props that we override with values from the Alert
-type OmitAlertProps = 'role';
-
 export type AlertHeadingLevel = '1' | '2' | '3' | '4' | '5' | '6';
 
 export type AlertRole = 'alert' | 'alertdialog' | 'region' | 'status';
@@ -97,6 +94,9 @@ const defaultAnalytics = (heading = '', variation = '') => ({
     type: variation,
   },
 });
+
+// Omit props that we override with values from the Alert
+type OmitAlertProps = 'role' | 'children' | 'className' | 'ref';
 
 export class Alert extends React.PureComponent<
   Omit<React.ComponentPropsWithRef<'div'>, OmitAlertProps> & AlertProps,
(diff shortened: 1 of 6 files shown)


packages/design-system/src/components/Alert/Alert.tsx
```diff
@@ -114,7 +131,7 @@ export class Alert extends React.PureComponent<AlertProps, any> {
 
     if (alertSendsAnalytics()) {
       const eventAction = 'onComponentDidMount';
-      const eventHeading = this.props.heading || this.props.children;
+      const eventHeading: string | React.ReactNode = this.props.heading || this.props.children;
```
zarahzachz, line 134: Nice cleanup work here!

packages/design-system/src/components/ChoiceList/Choice.tsx
```diff
@@ -4,6 +4,9 @@ import React from 'react';
 import classNames from 'classnames';
 import uniqueId from 'lodash.uniqueid';
 
+export type ChoiceSize = 'small';
+export type ChoiceType = 'checkbox' | 'radio';
+export type ChoiceValue = number | string;
 export interface ChoiceProps {
```
zarahzachz, line 10: I think there may be a conflict with the use of `ChoiceProps` here. At least in what I'm seeing in Aaron's screenshot, it looks like this actually gets exported from the ChoiceList file instead of Choice. I found a reference to what I think they're using, but not quite sure: https://github.com/CMSgov/design-system/blob/core-2.6.1/packages/design-system/src/types/ChoiceList/ChoiceList.d.ts#L10

packages/design-system/src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -4,6 +4,10 @@ import React from 'react';
 import classNames from 'classnames';
 import pick from 'lodash/pick';
 
+export type ChoiceListSize = 'small';
+export type ChoiceListType = 'checkbox' | 'radio';
+export type ChoiceListErrorPlacement = 'top' | 'bottom';
+
 // Omit props that we override with values from the ChoiceList
 type OmitChoiceProp = 'inversed' | 'name' | 'onBlur' | 'onChange' | 'size' | 'type' | 'inputRef';
 type ChoiceProps = Omit<ChoiceComponentProps, OmitChoiceProp>;
```
zarahzachz, line 13: I believe this `ChoiceProps` is what Aaron's team was importing

packages/design-system/src/components/ChoiceList/Choice.tsx
scheul93: From what I'm seeing, I think `ChoiceProps` is exported from both files. It is exported here (so it can be imported into `ChoiceList` and also from `ChoiceList` (which is where Aaron's team is importing it from)

Here's where I'm seeing it exported from the original file: https://github.com/CMSgov/design-system/blob/core-2.6.1/packages/design-system/src/types/ChoiceList/Choice.d.ts

zarahzachz (approved):

PR #1105, opened 2021-08-03
[WNMGDS-1002] updating doc site to display TS prop types correctly

scheul93:
## Summary
In situations like the `Dropdown` component, the doc site prop table was displaying `ReactText` instead of the acceptable types (`string | number`);

### Changed
I used [this prop](https://github.com/styleguidist/react-docgen-typescript#shouldextractvaluesfromunion-boolean) in the `react-docgen-typescript` library to help ensure the doc site has more prop info.

Now the Dropdown component has some appropriate looking props:
Screen Shot 2021-08-03 at 4 08 35 PM

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1002/ts-docsite-bugfix

1. Navigate to the Dropdown component and ensure that all prop types seem correct
2. Navigate to other Typescript components and confirm they look okay

Notes (some things I noticed while looking around the doc site):
- The `choices` prop on the `Choice` component is a little wonky. This is probably because of the way this prop is referencing an Interface but also omitting things?
Screen Shot 2021-08-03 at 4 05 34 PM
- The `required` doesn't show up on all required TS props. Particularly, it does not like the `children` prop when it is defined as type `ReactNode`. I troubleshot this and confirmed it is happening with `children` but I couldn't find a good reason why.


line47:
I'm noticing the `className` prop is no longer showing and children is not showing at all for the alert component if you compare these 2 links:  
- http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1002/ts-docsite-bugfix/components/alert/
- https://design.cms.gov/components/alert/


>`children` prop when it is defined as type `ReactNode`. I troubleshot this and confirmed it is happening with `children` but I couldn't find a good reason why.

I saw required children working on the table component - not sure if there is something there that is different?

zarahzachz (approved):
Great find with `shouldExtractValuesFromUnion`!

PR #1106, opened 2021-08-05
[WNMGDS-916] Feature/secondary tertiary buttons

line47:
## Summary
Adding secondary and tertiary button options to the core design system and also adding variables for customization as the child design system theme level 

### Added
- secondary button styles and variables
- tertiary button styles and variables
- Added class names to examples 

### Deprecated
- Danger button is marked as deprecated

## How to test
take a look at the buttons page on the demo site 
http://design-system-demo.s3-website-us-east-1.amazonaws.com/feature/secondary-tertiary-buttons


packages/design-system/src/styles/settings/variables/_forms.scss
```diff
@@ -25,7 +29,111 @@ $choice-border-radius: 0 !default;
 $choice-size: $spacer-4 !default;
 $choice-size-small: 20px !default;
 
+// Buttons
+$button-line-height: 1.3 !default;
 $button-border-width: 1px !default;
+$button-font-weight: $font-bold !default;
```
line47, line 35: Added font weight to be used at the mgov level - they don't use bold for their button text

line47:
@mliupaeter singed off on this ([in Slack](https://cmsgov.slack.com/archives/CT169N26L/p1628517959016600?thread_ts=1628517464.016200&cid=CT169N26L)) and it's ready for review 🎉

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -90,207 +79,136 @@
...
-/* stylelint-enable scss/at-extend-no-missing-placeholder */
-
-/**
- *  Status variants
- */
 $button-color-variants: (
-  primary: $button-primary-bg $button-primary-bg--hover $button-primary-bg--active,
-  success: $color-success $color-success-dark $color-success-darker,
-  danger: $color-error $color-error-dark $color-error-darkest,
+  button: $button-default-bg $button-default-bg--hover $button-default-bg--active
```
scheul93, line 101: [question] Is it standard for variables like this to use lists like this and then reference the list position in the styles below?

Do we ever do something more explicit like
```
$button-color-variants: (
  button: (
    background-color: $button-default-bg,
    background-hover: $button-default-bg--hover,
    background-active: $button-default-bg--active
    ...
  )
)
```
and then reference it below like `map-get($color, "background-color")`?

I'm rusty on sass so I don't know if:
1. that's possible
2. that's a convention we use in other places in the project and is something to do here
3. that's something worth doing to be more explicit instead of relying on list position 🤷🏻‍♀️

scheul93 (approved):
@line47 One comment that's more of a code-style thing. Otherwise, seems good to me

packages/design-system/src/styles/components/_Button.scss
line47: I'd definitely be interested in simplifying and making this more readable - I was extending what was used previously. I can look to see if there is a way we can make it more readable

scheul93: Okay! And if it's not possible or it's taking hours to refactor, it's not a big deal. 🙂

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -89,208 +78,195 @@
...
     &:hover,
     &:focus,
     &:active,
     &.ds-c-button--hover,
     &.ds-c-button--focus,
     &.ds-c-button--active {
-      border-color: transparent;
-      color: $color-white;
+      border-color: map-get($map, border-color);
+      color: map-get($map, text-color);
```
scheul93, line 219: [question] what is the point of this rule section? It seems like the `border-color` & `color` are all getting overwritten for specific states in the subsequent rules, so when does this ruleset ever get used?

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -89,208 +78,195 @@
...
-/* stylelint-enable scss/at-extend-no-missing-placeholder */
-
-/**
- *  Status variants
- */
-$button-color-variants: (
-  primary: $button-primary-bg $button-primary-bg--hover $button-primary-bg--active,
-  success: $color-success $color-success-dark $color-success-darker,
-  danger: $color-error $color-error-dark $color-error-darkest,
+$buttons-map: (
```
scheul93, line 81: [comment] this definitely got longer, but much easier to follow! 😄

packages/design-system/src/styles/components/_Button.scss
line47: Thanks and you're right - I'll removed this section

zarahzachz (approved):

PR #1109, opened 2021-08-06
WNMGDS-1008 aligning doc site footer links

scheul93:
## Summary
Updating the doc site CSS to remove transparent border on buttons in link list.

I wanted to make doc site specific styles instead of messing with the button component styling in case that broke current usage of transparent buttons

## How to test
1. Navigate to demo site: http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1008/doc-site-footer-fix
2. Scroll to footer and confirm that the links are aligned regardless of their HTML element
_This change should trickle down to the child systems after this is released and child systems update their dependency version_

Core:
Screen Shot 2021-08-06 at 2 02 30 PM

Hgov:
Screen Shot 2021-08-06 at 2 08 05 PM

Mgov:
Screen Shot 2021-08-06 at 2 05 34 PM

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -11,6 +11,6 @@
     "packages/eslint-config-design-system/**"
   ],
   "tagVersionPrefix": "core-",
-  "version": "2.7.1",
+  "version": "2.7.2",
   "useWorkspaces": true
 }
--- a/packages/design-system-docs/package.json
+++ b/packages/design-system-docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/design-system-docs",
-  "version": "2.7.1",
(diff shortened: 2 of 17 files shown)


line47:
This works and I'm wondering about instead using the utility class on the button `ds-u-border--0` to remove the border then we wouldn't need to create a new SCSS file?

scheul93:
> This works and I'm wondering about instead using the utility class on the button `ds-u-border--0` to remove the border then we wouldn't need to create a new SCSS file?

I missed that! Updated -- which makes this PR pretty straightforward 🙂

line47 (approved):

zarahzachz (approved):

PR #1110, opened 2021-08-10
[NO-TICKET] Remove getting started and guideline pages from backstop

line47:
## Summary
Remove getting started and guideline pages from visual regression checks


scheul93:
This is definitely a good start, but I think our backstop tests could test even less, specifically in relation to Components.

For example, the button component tests these URLS
1. - http://localhost:3000/components/button/
2. - http://localhost:3000/example/components.button/
3. - http://localhost:3000/example/components.button.icons/
4. - http://localhost:3000/example/components.button.inline-field/
5. - http://localhost:3000/example/components.button.react/
Items 2 - 5 are all the code examples for the Button component and should continue to be tested. Item 1 is the overall Button documentation and, in my opinion, doesn't need visual regression testing.

My opinion is that as a design system, we should care most about any visual changes to components that our users would be consuming. I don't really want to test & review all documentation changes - we are the ones making those updates and there is less chance of unanticipated style changes in the docs.
---
The same reduction could also be applied to all Components, Utilities & Patterns sections

line47:
I wonder if we could use our code page view to simplify to one URL?
https://design.cms.gov/components/button/?view=code 
This removes most of the guidance however there are a few sections that remain.

scheul93:
Maybe? The `view=code` still also shows the props table, which I don't think needs visual regression

line47:
Yeah that makes sense, I've updates to only reference the code & style examples

scheul93 (approved):
I think this takes us from 540 tests to varying usefulness to 327 tests focused on components. I like it!

PR #1111, opened 2021-08-10
[WNMGDS-963] updating vertical nav for improved accessibility

scheul93:
## Summary
Vertical Nav accessibility update

### Changed
Vertical Nav component to have improved accessibility by wrapping it in a `<nav>` element and adding an optional `aria-label` prop.

## How to test
Access the demo site at http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-963/vertical-nav-accessibility-update

Navigate to the Vertical Nav component

Examine the React iframe and confirm that the <nav> element now wraps the component & an aria-label is present (only because it is configured that way)

The docs for the VerticalNav should also reflect the prop addition. Confirm the information provided is acceptable for consumers


packages/design-system/src/components/VerticalNav/VerticalNav.jsx
```diff
@@ -50,6 +53,11 @@ VerticalNav.defaultProps = {
 };
 
 VerticalNav.propTypes = {
+  /**
+   * An optional arial label for the nav element in this component.
```
line47, line 57:
```suggestion
   * An optional arial label for the `<nav>` element in this component.
```

packages/design-system/src/components/VerticalNav/VerticalNav.jsx
```diff
@@ -50,6 +53,11 @@ VerticalNav.defaultProps = {
...
+   * This prop is necessary is there is more than one nav element on a page.
```
line47, line 58:
```suggestion
   * This prop is necessary when there is more than one nav element on a page.
```

packages/design-system-docs/src/pages/components/VerticalNav/VerticalNav.example.jsx
```diff
@@ -16,6 +16,7 @@ Link.propTypes = { children: PropTypes.node };
 ReactDOM.render(
   <VerticalNav
     selectedId="team"
+    ariaNavLabel="menu"
```
davidakennedy, line 19: Can we change this label to something like: "Main" or "Primary"? Most screen readers will know it's a menu because it has the `<nav>` element. The label provides a bit more context. I'm going to include this point in the docs too.

davidakennedy:
@scheul93 This looks good. Made one minor change request. And also committed some updates to the docs. Give those a read and see if they make sense to you. Otherwise, looks good!

scheul93:
Thank you both for the review! I updated your suggestions and also updated the HTML example, which I had missed before

packages/design-system-docs/src/pages/components/VerticalNav/_VerticalNav.docs.scss
```diff
@@ -58,8 +58,10 @@ if you're using the React components.
 - If your navigation list is longer than 3 items, consider using a skip navigation link. This allows screen reader and keyboard users to skip to the main content area(s).
 - If a skip navigation link is not an option, consider using:
   * A valid, descriptive page header
-  * Landmark regions like `<header>`, `<nav>`, `<main>`, `<aside>`, `<footer>`
-  * A `role` attribute like `<div role="navigation">`
+  * Landmark regions like `<header>`, `<main>`, `<aside>`, `<footer>`
+- When naming the optional `ariaNavLabel` prop, keep in mind:
+    * This prop is necessary only when there is more than one nav element on a page.
```
zarahzachz, line 63: Thinking out loud here...

I'm curious about the usage of this `aria-label`. This feels like something that will almost certainly never be used even with this information in the documentation. Is there a way to add more emphasis here so it stands out more? Or maybe have a default `aria-label` text that can be changed or turned off?

zarahzachz (commented):
This looks good, but I have a question about `aria-label` on this. It's more of a conversation starter than anything.

packages/design-system-docs/src/pages/components/VerticalNav/_VerticalNav.docs.scss
davidakennedy: Good callout, @zarahzachz. I struggled with this a bit too when thinking about it. I don't know how many of the applications/sites that use this component have another menu on the page.

I would be fine going the route of having it there by default because even if you only have one menu, having it named isn't a bad thing. It only provides more context for people who use screen readers.

Where I wondered if providing a default might be difficult is: What do we name it? We don't have any context of what will be in the navigation. We could call it something like "Main" or "Primary" and hope that people update it to fit their needs.

scheul93: Per a Slack conversation, it seems like it is better to have a region with no name than a region that is mis-named. So we should not provide a default for aria-label, for now.

PR #1124, opened 2021-08-30
Feature/pagination unit tests

scheul93:
### Added
Unit tests for pagination. I took the original declarations that were made and expanded upon them. I also made some small tests for the `Page` component.

I also updated the logic of the pageBuilder to match what I had commented in the original PR. I _think_ it works.

## How to test
Run unit tests to make sure everything looks good. Running this command will only run the new Pagination tests: `yarn test:unit packages/design-system/src/components/Pagination/Pagination.test.tsx`

Ensure there are no build errors.

Ensure the updated pageBuilder logic works as expected

### DIFF
--- a/packages/design-system/src/components/Pagination/Page.test.tsx
+++ b/packages/design-system/src/components/Pagination/Page.test.tsx
@@ -1,0 +1,17 @@
+import Page from './Page';
+import React from 'react';
+import { shallow } from 'enzyme';
+
+describe('Page', () => {
+    const onPageChange = jest.fn();
+
+    it('should render interactive el if not current', () => {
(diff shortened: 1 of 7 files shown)


packages/design-system/src/components/Pagination/Page.test.tsx
```diff
@@ -0,0 +1,17 @@
+import Page from './Page';
+import React from 'react';
+import { shallow } from 'enzyme';
+
+describe('Page', () => {
+    const onPageChange = jest.fn();
+
+    it('should render interactive el if not current', () => {
```
zarahzachz, line 8: I like the use of interactive vs. static in describing these tests. Makes it way less brittle!

zarahzachz (approved):

PR #1127, opened 2021-09-01
Bugfix/master ts export updates

scheul93:
## Summary
Applying changes from these two TS export update PRs to master. Have to do this manually since other PRs would introduce development code that is not ready

PRs for reference:
https://github.com/CMSgov/design-system/pull/1118
https://github.com/CMSgov/design-system/pull/1099/files

## How to test
Tests and build should pass. You can compare file changes to those in the PRs to ensure nothing got missed.


scheul93:
Note: There's also a bunch of other commits on here from the last two times I did releases. I don't think they changed any files, but kind of weird and I'm not sure why they're here

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -24,7 +24,7 @@ import { errorPlacementDefault } from '../flags';
 import get from 'lodash/get';
 import uniqueId from 'lodash.uniqueid';
 
-export interface AutocompleteItem {
+export interface AutocompleteItems {
```
zarahzachz, line 27: I never noticed the original was singular instead of plural. Good catch!

zarahzachz (approved):

line47 (approved):

PR #1131, opened 2021-09-07
[WNMGDS-843] updated spinner markup for improved accessibility

scheul93:
## Summary
Overview of changes

## How to test
Navigate to spinner component at demo site: http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-843/spinner-update. If you have the axe browser extension installed, confirm that the error "Spinner component doesn't have an accessible name" is not longer present.

Compare that spinner to what is on https://design.cms.gov/components/spinner/. Confirm that stylistically, nothing has changed.

Run `yarn test` on branch and confirm that all tests pass

### DIFF
--- a/packages/design-system/src/components/Spinner/Spinner.tsx
+++ b/packages/design-system/src/components/Spinner/Spinner.tsx
@@ -40,12 +40,16 @@ export const Spinner: React.FunctionComponent<SpinnerProps> = (props: SpinnerPro
     props.className
   );
 
-  return <span className={className} aria-valuetext={props['aria-valuetext']} role={props.role} />;
+  return (
+    <span className={className} role={props.role}>
+      <span className="ds-u-visibility--screen-reader">{props['aria-valuetext']}</span>
+    </span>
+  );
 };
 
 Spinner.defaultProps = {
   'aria-valuetext': 'Loading',
-  role: 'progressbar',
+  role: 'status',
 };
 
 export default Spinner;
--- a/lerna.json
+++ b/lerna.json
@@ -11,6 +11,6 @@
     "packages/eslint-config-design-system/**"
   ],
   "tagVersionPrefix": "core-",
-  "version": "2.7.2",
+  "version": "2.7.4",
   "useWorkspaces": true
 }
--- a/packages/design-system-docs/package.json
+++ b/packages/design-system-docs/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/design-system-docs",
-  "version": "2.7.2",
+  "version": "2.7.4",
   "publishConfig": {
     "access": "public",
     "tag": "latest"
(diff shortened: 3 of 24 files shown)


scheul93:
@davidakennedy, when you get time, can you please confirm that this PR accurately updates your recommendation for the Spinner component? Particularly, on the doc site, could you confirm that the markup for "Spinner inside a button" is correct?

line47 (approved):
LGTM!

davidakennedy:
@scheul93 This looks good to go, and tests well in all the major screen readers we test in.

My only other caveat/thought that I had was the possibility of making the hidden "Loading" text a prop that can be changed or left empty. This could be helpful in two cases:
1. The case where the spinner is on a button and the button's label does the work of labeling the loading state. Does this happen in the wild? I think you looked at current use cases. Because in the case of the example on the docs site where the button says "Saving" a screen reader would say "Loading Saving" and that's a bit odd.
2. The case where we want the label to be more specific. Like "Loading health plans" or something similar.

Just a thought, and we should optimize if we have those use cases, but it's okay not to if we don't.

scheul93:
This component currently supports overriding the "Loading" text which can be achieved by passing a value to the `aria-valuetext` prop.

If there should not be any text displayed (in the example of a spinner being used inside a button), a user can pass an empty string to the `aria-valuetext` prop like `<Spinner className="ds-u-margin-right--1" aria-valuetext="" />` and this markup will be rendered
```
<button class="ds-c-button ds-c-button--primary" disabled="" type="button">
    <span class="ds-c-spinner ds-u-margin-right--1" role="status">
        <span class="ds-u-visibility--screen-reader"></span>
    </span>
    Saving
</button>
```
Does that work?

davidakennedy (approved):

packages/design-system/src/components/Spinner/Spinner.tsx
```diff
@@ -40,12 +40,16 @@ export const Spinner: React.FunctionComponent<SpinnerProps> = (props: SpinnerPro
...
+    <span className={className} role={props.role}>
+      <span className="ds-u-visibility--screen-reader">{props['aria-valuetext']}</span>
+    </span>
+  );
 };
 
 Spinner.defaultProps = {
   'aria-valuetext': 'Loading',
-  role: 'progressbar',
+  role: 'status',
```
pwolfert, line 52: Do we want role to even be a prop? What other role would they want? If they wanted `progressbar` again, they wouldn't have the ability to add the `aria-label` or whatever was needed to make it accessible, right?

pwolfert (approved):
LGTM 👍 

One question about whether or not it's necessary to have `role` as a prop anymore, but you can ignore it for now. @davidakennedy, if you have thoughts on it, we could iterate later.

PR #1133, opened 2021-09-08
[WNMGWINSHOP-934] Add more customization options to Autocomplete items

pwolfert:
## Background
One of our applications needs to support a new "View all..." link at the bottom of the autocomplete results. It will need to be an item whose markup is generated differently from the other items. We actually customize the rendering of our items currently by passing in our own `itemToString` function that returns the markup we wish to generate (a non-string, I know). This is not ideal. Now that we need to render some items differently than others, it would make sense to support both use cases by implementing more straightforward support for custom-rendered items. That's what this does.

Also bundled in this PR is the addition of an optional `className` property for items. While I'd normally keep the changes in separate PRs, it made sense to me to bundle the change with this one so I could showcase an example that was more true-to-life.

### Added
- Added optional `children` property to Autocomplete `items`, which can be used to supply a custom contents for the result's `<li>` in the form of a React Node
- Added optional `className` property to Autocomplete `items`, which can be used to apply CSS classes to the result's`<li>`

## How to test
1. View the Autocomplete example that was added


packages/design-system-docs/src/pages/components/Autocomplete/Autocomplete.example.jsx
```diff
@@ -2,6 +2,17 @@ import { Autocomplete, TextField } from '@design-system';
...
+// Right now the documentation files don't actually get parsed for styles, so
+// this is a hack to add custom CSS classes for this example.
+const styleElement = document.createElement('style');
+styleElement.innerText = `
+  .ds-c-autocomplete__search-all.ds-c-autocomplete__list-item--active a {
+    color: #fff;
+    text-decoration-color: #fff;
+  }
+`;
+document.querySelector('head').appendChild(styleElement);
```
pwolfert, line 14: This is pretty hacky, and I'm not proud of it. Anyone have ideas to work around this limitation for now? Or ideas for any other kind of alternative? (I realize it's also not necessary for the example to even have the `ds-c-autocomplete__search-all` class because I'm applying the style to the link element, but it's showcasing the custom classes.

scheul93: Could you put a css class on the `<div>` (line 17) like `example-autocomplete`
And then in the file `packages/design-system-docs/src/styles/components/_example.scss`, you could move the styles to be
```
.example-autocomplete{
  .ds-c-autocomplete__search-all.ds-c-autocomplete__list-item--active a {
      color: #fff;
      text-decoration-color: #fff;
    }
}
```

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -176,16 +178,14 @@ export class Autocomplete extends React.Component<AutocompleteProps, any> {
...
-          }
+          className={classNames(item.className, 'ds-c-autocomplete__list-item', {
+            'ds-c-autocomplete__list-item--active': highlightedIndex === index,
+          })}
           key={item.id}
           role="option"
           {...getItemProps({ item })}
         >
-          {itemToString(item)}
+          {item.children ?? itemToString(item)}
```
scheul93, line 188: Nifty!

packages/design-system-docs/src/pages/components/Autocomplete/Autocomplete.example.jsx
pwolfert: That's great that there's a canonical place for these! So much better

pwolfert:
Another thing I'd like input on is the naming of that new property. The best I could come up with at the time was `children`, but can anyone think of a more appropriate name that wouldn't conflict with other naming patterns in the design system?

pwolfert:
@line47, I just added you as a reviewer to this too because I've added an example (`List with custom item markup`) and want your input on the details of that. Is the example text okay? Also the designs have actually changed today to treat that bottom link visually as having the same kind of hover/active state as the other items, because the focus halo on the link looked weird when using a pointer to hover. Would it be better if I updated the example to look like that instead of having that somewhat inconsistent style in an example?

line47:
@pwolfert Thanks for the tag - I think you image looks accurate and we don't want to use the focus styles in that dropdown list! 

We might want to think about some different example text - maybe something like option 1, option 2, select all options

packages/design-system-docs/src/pages/components/Autocomplete/Autocomplete.example.jsx
```diff
@@ -76,6 +76,56 @@ ReactDOM.render(
       />
     </Autocomplete>
 
+    <Autocomplete
+      items={[
+        {
+          id: '1',
+          name: 'Carrots (1)',
+          children: (
```
line47, line 84: Thinking about naming here - what about something like `autoCompleteAll` `autoCompleteGroup`?

pwolfert: `item.children` is for providing React elements for individual results rather than leaving the autocomplete component in charge of rendering that item based on `item.name`. Are you talking about replacing the `item.children` with `item.autoCompleteAll/Group`? I imagined alternatives like `item.element` or `item.content`.

davidakennedy:
@pwolfert Tested this out in the usual cadre of screen readers, and it works well.

It's doing what I mentioned in Slack, counting the "Search all snacks" part as the fourth of four results. So screen readers hear how many results are available: Four. But really there are three results with the "Search all snacks" link. Does that make sense?

It does create a mismatch in the true number of results. It may not be a huge deal. But I'm curious if that's possible to change, or if it's a big technical lift?

pwolfert:
@davidakennedy, hmm 🤔. One of our products provides _Downshift_ with a custom `getA11yStatusMessage` function so we can have more specific messages. You've probably tested it. It _would_ be possible to add our own `getA11yStatusMessage` implementation in the design system to accommodate things that shouldn't be listed as results.

pwolfert:
Looks like I actually might be able to modify the [`resultsCount`](https://github.com/downshift-js/downshift/blob/1c7698080805e05f5180dd223cbc17932a85fd42/src/downshift.js#L1026) passed to that `this.props.getA11yStatusMessage ` by setting [`itemCount`](https://github.com/downshift-js/downshift#itemcount)

pwolfert:
@davidakennedy, I've added a fix for the results count issue if you could please test it out again.

For other engineers, I'd like your feedback about the solution itself and if there are better ways you can think of to solve that issue.

davidakennedy:
@pwolfert This works much better! The results are accurate. A screen reader will still say, for example, "Two of four..." when moving through the results. It still counts that last link as the results from a markup perspective. But I think that's okay.

One side effect that I noticed in the custom markup example: The results are announced not only at the beginning, but also after each individual result. The part that says: "Three results available. Use up and down arrows to navigate..." That's extra noisy from a screen reader perspective. Are there any changes in the code that could cause this?

pwolfert:
@davidakennedy, thanks for bringing this bug to my attention. It highlights how hacky my solution is. Downshift's default `getA11yStatusMessage` function does a check between a `resultCount` and `previousResultCount` argument before announcing that "X results are available" message. I was modifying the `resultCount` but not the `previousResultCount`. Unfortunately to modify the `previousResultCount` would require some extra state management that would get even hackier. I don't think I should do it. In fact, I don't even like the first solution, but it's the best idea I've got so far.

pwolfert:
Regarding my previous comment, I actually came up with some code that will fix the bug without introducing extra state management or side effects: https://github.com/CMSgov/design-system/pull/1133/commits/5142f7699700f41c3d1aaf9272cf75a0c47fef4e#diff-d2902bdfecf5438a3e627f3b010ca640e6a471dd917ed3463f285d201ade4d61R282-R293

Still don't love it, but it should fix the bug.

packages/design-system/src/components/Autocomplete/Autocomplete.test.tsx
```diff
@@ -40,6 +40,70 @@ describe('Autocomplete', () => {
...
+
+    const tree = renderer
+      .create(
+        <Autocomplete items={items} isOpen>
+          <TextField label="autocomplete" name="autocomplete_field" />
+        </Autocomplete>
+      )
+      .toJSON();
+
+    // TODO: Clean this up by using another rendering library or figure out how to use this one to
```
scheul93, line 86: [suggestion] Consider using enzyme for this test to be more consistent with other unit tests. Starting at line 78, it would look like
```
    const wrapper = mount(
        <Autocomplete items={items} isOpen>
          <TextField label="autocomplete" name="autocomplete_field" />
        </Autocomplete>
    );
    const ul = wrapper.find('ul');
    expect(ul).toMatchSnapshot();
```
I tested ☝🏻 locally and I think it works, but if it gives you any issues, your current tests works.

davidakennedy:
@pwolfert Thanks for iterating on this! I think this is an improvement. I notice some odd behavior with Edge and Jaws announcing the "Three results available. Use up and down arrows to navigate..." part twice at the start after results are populated. However, that doesn't happen in any other browser/screen reader combination. Not even Chrome. So it makes me think this is a quirk with Edge and Jaws.

packages/design-system/src/components/Autocomplete/Autocomplete.test.tsx
scheul93: Also, separate issue, not for this PR, but I don't understand why we use both `react-test-renderer` & `enzyme`. What does one provide that the other doesn't?

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -27,6 +27,9 @@ import uniqueId from 'lodash.uniqueid';
 export interface AutocompleteItems {
   id?: string;
   name?: string;
+  children?: React.ReactNode;
+  className?: string;
+  isResult?: boolean;
```
scheul93, line 32: [question] is there any sort of documentation that is needed for the `AutocompleteItems` type? Specifically the `isResult` prop? Asking because would another dev know how to use this property for correct accessibility?

packages/design-system/src/components/Autocomplete/Autocomplete.test.tsx
pwolfert: > I don't understand why we use both react-test-renderer & enzyme

In this case I think we have both because when enzyme wrappers are serialized, they aren't human-readable. In other products, we use `enzyme-to-json` to get an HTML-looking snapshot. We could add it. What do you think? You can opt-in to using it in specific modules rather than adopting it for all.

pwolfert: I take that back! It looks like the serialized Enzyme objects are pretty now. That must have changed since I last tried using them in snapshots natively.

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
pwolfert: d013bde

packages/design-system-docs/src/pages/components/Autocomplete/_Autocomplete.docs.scss
```diff
@@ -18,6 +18,52 @@ The `Autocomplete` component is a parent component that adds autocomplete functi
...
+      <td><code>string</code></td>
+      <td>
+        <p>Displayed value of the item. May alternatively provide a <code>children</code> value</p>
+      </td>
+    </tr>
+    <tr>
+      <td><code class="ds-u-font-weight--bold">children</code></td>
+      <td><code>string</code></td>
+      <td>
+        <p>Additional classes to be added to the root element. Useful for adding utility classes.</p>
```
scheul93, line 51: [fix] there's a mis-match here. The description seems to be for the `className` prop which is not currently present in the docs. Please add the `className` prop and update the description for the `children` prop.

packages/design-system-docs/src/pages/components/Autocomplete/_Autocomplete.docs.scss
```diff
@@ -18,6 +18,52 @@ The `Autocomplete` component is a parent component that adds autocomplete functi
...
+      </td>
+    </tr>
+    <tr>
+      <td><code class="ds-u-font-weight--bold">isResult</code></td>
+      <td><code>string</code></td>
```
scheul93, line 56: [fix] Please update to `boolean` to match TS definition

packages/design-system-docs/src/pages/components/Autocomplete/_Autocomplete.docs.scss
```diff
@@ -18,6 +18,52 @@ The `Autocomplete` component is a parent component that adds autocomplete functi
...
+    <tr>
+      <td><code class="ds-u-font-weight--bold">name</code></td>
+      <td><code>string</code></td>
+      <td>
+        <p>Displayed value of the item. May alternatively provide a <code>children</code> value</p>
+      </td>
+    </tr>
+    <tr>
+      <td><code class="ds-u-font-weight--bold">children</code></td>
+      <td><code>string</code></td>
```
scheul93, line 49: [fix] Please update to `React.ReactNode` to match TS definition

packages/design-system-docs/src/pages/components/Autocomplete/_Autocomplete.docs.scss
pwolfert: Good catch 👍

scheul93 (approved):
Nice!

PR #1135, opened 2021-09-08
[NO-TICKET] Add clearInputOnBlur prop to Autocomplete

pwolfert:
## Summary
All of the instances of Autocomplete that I know of on hc.gov require the behavior implemented in this PR. DM me for details and more context.

### Added
- Added `clearInputOnBlur` default prop to `Autocomplete` that can be set to `false` to keep input after blur, instead of the default behavior which is to clear input if no selections are made.

## How to test
Modify one of the examples by setting `clearInputOnBlur` to false and observe the new behavior.

### DIFF
--- a/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
@@ -22,6 +22,7 @@ import WrapperDiv from './WrapperDiv';
 import classNames from 'classnames';
 import { errorPlacementDefault } from '../flags';
 import get from 'lodash/get';
+import keepInputDownshiftStateReducer from './keepInputDownshiftStateReducer';
 import uniqueId from 'lodash.uniqueid';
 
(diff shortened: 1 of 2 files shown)


zarahzachz (commented):
This appears to work as intended. Is there a way to write a test to support this functionality?

pwolfert:
@zarahzachz, so far I haven't been successful verifying that functionality with a unit test with the testing libraries we're using, and for the e2e tests, I was hesitant to add it to one of the examples in the docs pages just so I could test this case (because those two things are coupled). I know it's not a good alternative to tests—especially if there were to be regressions—but I can say this code has been running in production for a long time.

pwolfert:
I think my problem was that I couldn't figure out how to simulate a blur event for the entire component. Simulating blur events on the input wasn't enough to trigger the default behavior, so the unit test for this modified behavior wasn't valuable. If I can reproduce the default behavior, then we'd be good to go.

PR #1136, opened 2021-09-10
[NO-TICKET] Update chromedriver from version 91 to 93

pwolfert:
## Summary
Updated chrome driver version used in the e2e tests because the chromium version in our CI stack changed and e2e tests were failing.


zarahzachz (approved):
Ran it locally and it looks impeccable.

Mergey merge merge.

PR #1142, opened 2021-09-16
Simplify analytics API

pwolfert:
## Summary
In anticipation of implementing analytics in [hcgov-design-system](https://github.com/CMSgov/hcgov-design-system), simplify the analytics API for components as well as the helper function used to implement analytics. As far as we know, nobody is currently using the exposed `sendAnalytics` function or the `analytics` prop on any of the three components with analytics. This gives us freedom to pare it down to the only property of the event that product teams would actually consider overriding.


packages/design-system/src/components/analytics/SendAnalytics.ts
```diff
@@ -47,50 +35,38 @@ interface AnalyticsEventProps {
   [additional_props: string]: unknown;
 }
 
-export function sendAnalytics(event: EventType, props: AnalyticsPayload, retry = 0): string {
-  if (window.utag && window.utag[event]) {
+const MAX_RETRIES = 3;
+const TIMEOUT = 300;
+
+window.utag = { link: console.log };
```
scheul93, line 41: [fix] I'm getting a build error that `window is not defined` at this line
Not sure if this is happening because this was moved out of a function?

pwolfert: Whoops, that was for my debugging

packages/design-system/src/components/analytics/index.ts
```diff
@@ -1 +1 @@
-export { default as sendAnalytics } from './SendAnalytics';
+export { default as sendAnalytics } from './sendAnalytics';
```
scheul93, line 1: [nitpick] does this need to stay `SendAnalytics` to match the file name casing?

scheul93:
Overall, I think this is a good change because it simplifies the analytics. I guess, the assumption is that no one outside the DS was using the exported functions in `SendAnalytics` so renaming these shouldn't break anything?

packages/design-system/src/components/analytics/index.ts
pwolfert: Whoops, that was left over from a time when I thought I'd change the case of some files, and it was a terrible idea.

scheul93 (approved):

pwolfert:
> I guess, the assumption is that no one outside the DS was using the exported functions in SendAnalytics so renaming these shouldn't break anything?

That is my assumption, yes. I've verified it with HC.gov teams but not M.gov

packages/design-system/src/components/analytics/SendAnalytics.test.ts
```diff
@@ -1,6 +1,6 @@
-import { sendAnalyticsEvent } from './SendAnalytics';
+import { sendLinkEvent } from './sendAnalytics';
```
scheul93, line 1: [fix] the casing on the file name is causing unit tests to fail

pwolfert: 🤦‍♂️ I really thought TS would catch these for me.

PR #1149, opened 2021-09-20
[NO-TICKET] Update link to Jenkins job

pwolfert:
## Summary
The deploy script broke due to some recent changes in a shared script. That script doesn't support spaces in the job name or parent folder, so I renamed the job and its parent folder.


zarahzachz (approved):
Looks great!

PR #1153, opened 2021-09-22
[NO-TICKET] Update table example to fix right aligned header

line47:
## Summary
This was brought up in Slack as something to fix about the plain HTML table example. The header is right aligned and it looks off. 

### Changed
Removed the right align class from the table headers and cells. I think a better right aligned example would be needed to show when and why to align table content to the right.

## How to test
Go to the demo site table page and see that the html table header for `Rate` is no longer right aligned. 
http://design-system-demo.s3-website-us-east-1.amazonaws.com/sw-remove-table-alignment


pwolfert (approved):
Sounds good to me 👍 They did look a little weird

PR #1155, opened 2021-09-23
core-2.8.0

pwolfert:
Forgot to make a PR for this after last release

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -5,12 +5,14 @@
     }
   },
   "npmClient": "yarn",
-  "packages": ["packages/*"],
+  "packages": [
+    "packages/*"
+  ],
   "ignoreChanges": [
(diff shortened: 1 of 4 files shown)


lerna.json
```diff
@@ -5,12 +5,14 @@
     }
   },
   "npmClient": "yarn",
-  "packages": ["packages/*"],
```
zarahzachz, line 8: Just a comment - but I think I see the formatting of this line change in every PR. What rule should we enable to make this consistent?

zarahzachz (approved):

lerna.json
pwolfert: I think this is lerna modifying it vs prettier formatting it. Not sure what to do at the moment to fix it, but good to flag it

PR #1156, opened 2021-09-24
[WNMGDS-1094] Tagged release docs and scripts

pwolfert:
https://jira.cms.gov/browse/WNMGDS-1094

### Changed
- There are now two scripts for releasing and publishing
  1. `yarn release` creates the commit with the version bump and tags it as a release
  2. `yarn publish-release <release tag>` pulls that release, builds it, and publishes it to npm. I couldn't call it `yarn publish` because it conflicts with yarn's built-in commands
- Updated release-process docs. Read those docs for more details.

## How to test
1. Check out 82ee08c, which removes the teeth of the scripts so you don't accidentally push anything to origin or publish anything to npm
2. Run `yarn release` and note the command in the `NEXT STEPS` output
3. Run the next script with the command copied from the previous step
4. The result should be that you're on a release branch, you have extra zip files, and you have an extra local tag.
5. You can clean up that mess locally by
    1. Checking out another branch like `master`
    2. Deleting the release branch you were just on (`git branch -D release-2.8.0-alpha.0` or whatever you called it)
    3. Deleting the local tag (`git tag -d core-2.8.0-alpha.0` or whatever you called it)


zarahzachz (approved):
This is super slick! I love how the scripts prompt you what your next steps are! What an improvement!

zarahzachz:
Also, learned the difference between `-d` and `-D`. Always used `-D` and didn't know there was a less forcey option.
> The -d option is an alias for --delete, which only deletes the branch if it has already been fully merged in its upstream branch. You could also use -D, which is an alias for --delete --force, which deletes the branch "irrespective of its merged status." [Source: man git-branch]

pwolfert:
Oh, I never knew about the `-d/-D` difference either!

PR #1159, opened 2021-09-27
[NO-TICKET] Reduce OS notifications for build process

pwolfert:
## Summary
There are many steps in the build process, and they each have an OS-level popup notification associated with them. This is actually really obnoxious during development, so let's only use the notification for success messages. Even that has questionable utility.

### Changed
- (Internal) Restrict the OS notifications for the build process to only show for success messages

## How to test
Run `yarn start` or `yarn build`


scheul93 (approved):
Tried it locally and it's much less disruptive!

PR #1161, opened 2021-10-01
[RFC] [WNMGDS 461] icons future state

scheul93:
## Summary
Introducing a Request for Comments (RFC) directory with README + template along with a proposal for moving forward with DS icons.

### Added
* RFC directory
* README.md for what RFC's are
* template for future RFCs
* RFC for a plan for icon work

### What this PR needs from you
1. Feedback on the README.md. Does it make sense? Should there be more details?
2. Feedback on the template. Are there other sections you'd want to see when discussing DS proposals
3. Feedback on the Icon RFC. Does this path forward make sense? Are there other things missing or need to be considered?


rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+The .svg files in the `images/` directory are referenced in the code in two main ways:
+
+1. In CSS as background images
+2. Using the library `inline-react-svg`, .svg files can be imported and used as a component (Note: this does not work with Typescript components)
+
+## Proposal
+
+Icons in the core DS should be moved to a centralized location and each icon should be standardized in size, accessibilty feature, and exported individually as a React component. This will provide more flexibility to consuming applications by allowing them to import individual SVG icons, adjust the styling on them with more ease, and create icon consistency among products.
+
+This ideal state sill be achieved in a few small phases:
```
pwolfert, line 27: > ...state *sill* be...

📝

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+
+1. [Create icon components](#phase-1-create-icon-components)
+2. [Clean & improve markup](#phase-2-clean-and-improve-svg-markup)
+3. [Update icon references in components](#phase-3-update-component-icon-references)
+4. [Update icon references in CSS](#phase-4-update-css-icon-references)
+5. [Beyond](#beyond)
```
pwolfert, line 33: I love the use of internal links here. Makes it really easy to follow

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+
+### Phase 1: Create Icon Components
+
+In this phase:
+
+- Make `Icons` directory in the design-system component folder
+- Create a component for each icon in the `images` directory
+- Move the `ClearIcon` and `ClearIconAlternate` components into new directory
+- Move inlined SVG markup for `<` & `>` from Pagination to new directory
+- Ensure all icons are exported
```
pwolfert, line 43: That's a good thing to note 😂

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+#### Considerations
+
+There was an idea to create one `Icon` component and pass in a `name` prop that would return the proper svg markup. However, this approach would not work well for child design systems and could inflate bundle sizes.
+
+### Phase 2: Clean and Improve SVG Markup
+
+In this phase:
+
+- Ensure SVGs have accessibility features
+  - See [Smashing Magazine](https://www.smashingmagazine.com/2021/05/accessible-svg-patterns-comparison/) or [Deque](https://www.deque.com/blog/creating-accessible-svgs/) for guidance
```
pwolfert, line 58: Love this

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+- Standardize sizing of SVGs either using viewbox or CSS
+- Remove any inline `fill` properties from SVG and move to CSS for easier overriding
```
pwolfert, line 60: These are also wonderful ideas 😄

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+- Ensure SVG markup is optmizine - run through optimizer
+
+### Phase 3: Update Component Icon References
+
+In this phase:
+
+- Update any current DS component that is using an inline SVG.
+  - This includes, but is not limited to: `Pagination`, `FilterChip`, `TooltipIcon`, `UsaBanner`
+
+### Phase 4: Update CSS Icon References [Breaking Change]
```
pwolfert, line 70: > [Breaking Change]

Good call-out

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+
+In this phase:
+
+- Update any current DS component that is referencing an SVG in its CSS to use inlined icon components
+  - This includes, but is not limited to: `Accordion`, `Alert`, `Dialog`, `Dropdown`, `StepList`, `UsaBanner`, `VerticalNav`
+- Create documentation for consuming applications. This should explain change and instruct on how application styles may be affected
```
pwolfert, line 76: "explain ^the change"?

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+
+_**Note:** This phase will rely deeply on visual regression testing to ensure component styles translate correctly_
```
pwolfert, line 78: Another good call-out

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+
+This phase will include a breaking change to consuming applications' styles and potentially build systems. Will need to coordinate timing and communication.
+
+### Beyond
+
+Future icon work that needs consideration:
+
+- Incorporate configurable icons into some components (like Accordion, FilterChip, Modal, Naviation, etc)
```
pwolfert, line 86: This is a really interesting idea!

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+3. Included in a component's markup unable to be re-used individually.
+
+The .svg files in the `images/` directory are referenced in the code in two main ways:
+
+1. In CSS as background images
+2. Using the library `inline-react-svg`, .svg files can be imported and used as a component (Note: this does not work with Typescript components)
+
+## Proposal
+
+Icons in the core DS should be moved to a centralized location and each icon should be standardized in size, accessibilty feature, and exported individually as a React component. This will provide more flexibility to consuming applications by allowing them to import individual SVG icons, adjust the styling on them with more ease, and create icon consistency among products.
```
pwolfert, line 25: > ...should be standardized in size, accessibilty feature, and exported individually...

That second item in the list is a little confusing. Should it be "standardized in size and accessibility features and exported..."? Does the "standardized" apply to "accessibility features"?

rfcs/README.md
```diff
@@ -0,0 +1,27 @@
...
+- Submit the PR with the `[RFC]` PR title prefix and tag
+- Commence discussion, adjusting the RFC based on feedback
+- Once all questions defined in the RFC are satisfied, and PR is accepted, merge it
+
+### RFC Status Options
+
+- WIP: The author is still drafting the RFC and it’s not ready for review.
+- Review: The Review label is used when the RFC is ready for comments and feedback.
+- Approved: When the RFC is for the purpose of making a decision, the Approved label indicates that the decision has been made.
+- Implemented: When the RFC is for the purpose of making a decision, the Implemented label indicates that the RFC’s proposal has been implemented.
```
pwolfert, line 21: I don't know for our use case if the `WIP` status is very useful because we have GitHub PR tags that overlap with that functionality, but I do not have a strong opinion. I think it might be something I personally would forget to change or just not use.

pwolfert: On second thought, it might take longer to explain an alternative process of taggingn PRs and assigning reviewers than to just include this as a status, and people can use it or not.

pwolfert (approved):
This is excellent! You've thought thoroughly about this, and this will translate well into tickets.

rfcs/README.md
```diff
@@ -11,7 +11,7 @@ RFCs are useful when team feedback is required before making architecture or des
 - Fill out the template accordingly
 - Submit the PR with the `[RFC]` PR title prefix and tag
 - Commence discussion, adjusting the RFC based on feedback
-- Once all questions defined in the RFC are satisfied, and PR is accepted, merge it
+- Once all questions defined in the RFC are satisfied, and PR is accepted, merge it. Note: the goal is to have the RFC reviewed by multiple people and perspectives. The RFC PR should not be merged at first approval in order to give others time to provide feedback.
```
pwolfert, line 14: 💯

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+3. Included in a component's markup unable to be re-used individually.
+
+The .svg files in the `images/` directory are referenced in the code in two main ways:
+
+1. In CSS as background images
+2. Using the library `inline-react-svg`, .svg files can be imported and used as a component (Note: this does not work with Typescript components)
+
+## Proposal
+
+Icons in the core DS should be moved to a centralized location. Each icon should be exported individually as a React component and updated with a standard size and consistent accessibilty features. This will provide more flexibility to consuming applications by allowing them to import individual SVG icons, adjust the styling on them with more ease, and create icon consistency among products.
```
zarahzachz, line 25: Is it worth calling out that these new icons should be optimized for performance using a tool like SVGOMG? I don't know if we currently have something like this in our workflow.

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+5. [Beyond](#beyond)
+
+### Phase 1: Create Icon Components
+
+In this phase:
+
+- Make `Icons` directory in the design-system component folder
+- Create a component for each icon in the `images` directory
+- Move the `ClearIcon` and `ClearIconAlternate` components into new directory
+- Move inlined SVG markup for `<` & `>` from Pagination to new directory
```
zarahzachz, line 42: While I was writing Pagination, I was thinking I could've just used 1 chevron icon and used CSS to rotate it the direction I wanted. Maybe we only need 1 of these chevron icons?

rfcs/README.md
zarahzachz: Do we update the date when the status changes?

rfcs/2021-10-icons-in-core-DS.md
scheul93: I like that. We could honestly have a variation for up, down, left and right if needed. I'll update with clarification

rfcs/2021-10-icons-in-core-DS.md
scheul93: Sure, I can add this to the "Clean & improve markup" section

rfcs/README.md
scheul93: 🤷🏻‍♀️ What is the benefit of updating the date in the doc vs referencing git history?
Additionally... is the date actually useful info?

rfcs/2021-10-icons-in-core-DS.md
scheul93: Oh wait. I already have it in that section under
```
Ensure SVG markup is optimized - run through optimizer
```

rfcs/README.md
pwolfert: I think the date is useful just for understanding at a glance when it was started without looking at the log. For this reason, I don't think we'd update the date when changes are made—just keep it as the date of origination

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+1. In CSS as background images
+2. Using the library `inline-react-svg`, .svg files can be imported and used as a component (Note: this does not work with Typescript components)
+
+## Proposal
+
+Icons in the core DS should be moved to a centralized location. Each icon should be exported individually as a React component and updated with a standard size and consistent accessibilty features. This will provide more flexibility to consuming applications by allowing them to import individual SVG icons, adjust the styling on them with more ease, and create icon consistency among products.
+
+This ideal state can be achieved in a few phases:
+
+1. [Create icon components](#phase-1-create-icon-components)
```
line47, line 29: Not sure in which phase, but we should also ensure that icons represented in code are also available in the UI kit

rfcs/2021-10-icons-in-core-DS.md
```diff
@@ -0,0 +1,102 @@
...
+- Move inlined SVG markup for `<` & `>` from Pagination to new directory. Consider consolidating to one Icon with variations.
+- Ensure all icons are exported
+- Create base icon component styles to be re-used in each component
+- Update doc site with component details
+
+Exporting each SVG icon individually allows consuming applications to only import the icons they are using.
+
+#### Considerations
+
+There was an idea to create one `Icon` component and pass in a `name` prop that would return the proper svg markup. However, this approach would not work well for child design systems and could inflate bundle sizes.
```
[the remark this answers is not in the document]
scheul93: * I think the biggest difference in using one component vs may is that is one component, you would be building out a single `Icon` component that basically has a giant conditional statement to figure out which SVG to load. This wouldn't be so bad with our current amount of icons, but if we continue to grow our icon library, it would cause a bunch of SVG markup to be part of the component when a team may only really need one icon and a small amount of markup. Additionally, the conditional to determine the markup is part of the component, so it would be difficult for a child system to use this `Icon` component with a child DS-specific icon.
* In my opinion, I don't think either is "easier" or "harder". Like mentioned before, having only one icon could inflate JS bundle sizes unnecessarily.
* If the design systems were unified, the one icon approach _could_ be extended because it would have easier access to child DS files, and therefore include them as part of the large component conditional. I still don't think that is the way to go because then you run the risk of having mgov specific icons as part of the Icon component that gets pulled into hgov and have a bunch of code that is not relevant to one child system

All this being said... We could do both approaches. I still think that individual icons are a scalable path forward for now, but if people want a unified singular `Icon` component, we could make it and reference the individual icon components as a future enhancement 🤷🏻‍♀️

rfcs/2021-10-icons-in-core-DS.md
scheul93: What is the UI kit?

scheul93: I added this as part of "phase 2" and I'll make sure to add a ticket to track!

PR #1166, opened 2021-10-04
Wnmgds 1105 helpdrawer util classes

zarahzachz:
## Summary
Refactored CSS util classes, consolidating rules into component name-spaced classes. I needed to create some new classes to accommodate this work. Also addressed, [WNMGDS-1073](https://jira.cms.gov/browse/WNMGDS-1073) so the HTML example matched the React example.

## How to test
You can see the changes here (you actually shouldn't see any change outside of the HTML example footer behavior): http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1073-helpdrawer-util-classes

I ran Backstop on this work and 1 issue popped up relating to HelpDrawer. However, it was unclear what was wrong with it. The diff looked like HelpDrawer text was off by maybe 1px, which didn't seem like a real issue.

### DIFF
--- a/packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
+++ b/packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
@@ -90,11 +90,11 @@ export class HelpDrawer extends React.PureComponent {
            * separation between our sticky header, and the flex container
            * so things display as expected when the body content overflows
            */}
-          <div className="ds-u-fill--gray-lightest ds-u-padding--2 ds-u-display--flex ds-u-align-items--start">
+          <div className="ds-c-help-drawer__header-container">
             <Heading
               ref={(el) => (this.headingRef = el)}
               tabIndex="0"
-              className="ds-u-text--lead ds-u-margin-y--0 ds-u-margin-right--2"
+              className="ds-c-help-drawer__header-heading"
             >
               {
                 // TODO: make heading required after removing title
@@ -103,7 +103,7 @@ export class HelpDrawer extends React.PureComponent {
             </Heading>
             <Button
               aria-label={ariaLabel}
-              className="ds-u-margin-left--auto ds-c-help-drawer__close-button"
+              className="ds-c-help-drawer__header-button ds-c-help-drawer__close-button"
               size="small"
               onClick={onCloseClick}
             >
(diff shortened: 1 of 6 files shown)


packages/design-system/src/styles/components/_HelpDrawer.scss
```diff
@@ -54,6 +62,21 @@
...
 }
 
+.ds-c-help-drawer__header-container {
+  align-items: flex-start;
+  background-color: $help-drawer-header-fill;
+  display: flex;
+  padding: $help-drawer-padding;
+}
+
+.ds-c-help-drawer__header-heading {
```
pwolfert, line 72: The utility classes you replaced with this includes a `ds-u-text--lead`. Does that need to be added here?

packages/design-system/src/styles/components/_HelpDrawer.scss
```diff
@@ -72,5 +97,21 @@
...
 
 .ds-c-help-drawer__footer {
+  background-color: $help-drawer-footer-fill;
   flex-shrink: 0;
+  font-size: $small-font-size;
+  line-height: 1.5;
+  padding: $help-drawer-padding;
+
+  @media (min-width: $width-lg) {
+    font-size: $base-font-size;
```
pwolfert, line 107: It looks like the media queries that you replaced actually applied to the `help-drawer__content` and not the footer. I can see the difference when I look at it on small screens:

footer-size

pwolfert (commented):
It looks identical for larger screen sizes! There seems to be a difference for smaller screens (see comments below). When I compare your branch and what's on master in two tabs on a larger window, I don't see a difference.

packages/design-system/src/components/HelpDrawer/HelpDrawer.jsx
```diff
@@ -103,7 +103,7 @@ export class HelpDrawer extends React.PureComponent {
             </Heading>
             <Button
               aria-label={ariaLabel}
-              className="ds-u-margin-left--auto ds-c-help-drawer__close-button"
+              className="ds-c-help-drawer__header-button ds-c-help-drawer__close-button"
```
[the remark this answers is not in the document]
zarahzachz: You're probably right. Updating!

packages/design-system/src/styles/components/_HelpDrawer.scss
zarahzachz: Turns out `ds-u-text--lead` wasn't a real class. I think the class that was wanted was `ds-text--lead`, but since the style wasn't applied, I assumed this class was unnecessary. 

On that point, it was a little confusing seeing some text helper classes that appeared to be utilities not have that `-u` identifier. I encountered similar confusion font-sizing classes for various media breakpoints. It's probably 100% my lack of familiarity with our util classes, but it wasn't immediately intuitive what the class was accomplishing. Is this a problem? I dunno. Just wanted to point it out.

packages/design-system/src/styles/components/_HelpDrawer.scss
zarahzachz: So it is... wonder why VisReg didn't pick it up 🤔

packages/design-system/src/styles/components/_HelpDrawer.scss
pwolfert: That's a good point. It totally seems like a utility class, but they appear to be in a category of their own and treated as first-class citizens. @line47, what do you think about this? Does it make sense to you for these typography classes to omit the `-u` prefix? The other thing I can think to compare it to are the layout classes with `-l`. If we did want typography to be separate, shouldn't we continue the pattern with something like `-t` for typography? All that said, I'm not eager to make breaking changes 😆

pwolfert (approved):
LGTM 👍

line47 (changes requested):
Looks good, I have one small change to the SCSS variable naming

line47 (approved):
LGTM!

PR #1182, opened 2021-10-11
Refactor utils out of StepList.

zarahzachz:
## Summary
[Jira 1111](https://jira.cms.gov/browse/WNMGDS-1111)

### Changed
Consolidated margin-top util into the `ds-c-step-list` class.

## How to test
Compare [test url](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1111/steplist-utils/patterns/step-list/) to [prod](https://design.cms.gov/patterns/step-list/) - there should be no difference.

I ran Backstop and saw no errors.

### DIFF
--- a/packages/design-system/src/types/Dialog/Dialog.d.ts
+++ b/packages/design-system/src/types/Dialog/Dialog.d.ts
@@ -1,26 +1,9 @@
+/* eslint-disable filenames/match-exported */
 import * as React from 'react';
 
 export type DialogCloseButtonSize = 'small' | 'big';
 
 export type DialogSize = 'narrow' | 'wide' | 'full';
-
-export interface AnalyticsEventShape {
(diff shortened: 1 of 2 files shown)


zarahzachz:
@line47 There were 2 util classes I made the call to keep throughout this process: `.ds-u-visibility--screen-reader` and `.ds-u-clearfix` (I think that's the correct class name for that one).

With the screen-reader class, I chose to keep it around since a) this class is so widely used throughout our system for the same reason that it didn't make much sense to recreate these same styles for each component and bloat our CSS unnecessarily and b) it only exists to support a11y and we don't want folks overwriting those styles anyway.

Clearfix was kept because the styles in that class don't have the `!important` flag, so folks could overwrite them even though this is a util class.

PR #1185, opened 2021-10-12
[WNMGDS-762] lodash import update

scheul93:
## Summary
Standardizing the way we import lodash modules. This is for consistency but to also preemptively remove [per package modules](https://lodash.com/per-method-packages) which will be removed in a future lodash release

### Added
`lodash` as a dependency of the `design-system` package

### Changed
Any usage of the `uniqueid` lodash module was updated to import like other lodash modules

## How to test

- Ran tests and nothing broke
- Ran a build and no build errors
- I ran both mgov and hgov locally and didn't get any build errors. From what I observed, I also didn't see any regressions in the doc sites


scheul93:
Fixed the package name. I'm a little worried that it passed locally. I must have had cached node modules.
Either way, should be good to review now

PR #1186, opened 2021-10-13
Refactor Table utils, update VisReg ref images.

zarahzachz:
## Summary
[Jira 1112](https://jira.cms.gov/browse/WNMGDS-1112)

### Added
Added `.ds-c-table__scroll-alert` to consolidate table alert styles.

### Changed
Replaced the text alignment util class with `.ds-c-table__cell--align-${align}`. This caused some CSS issues as the table head cell elements had text-alignment="left" style explicitly set. The only way to overwrite this style was to use our utils, which use `!important` flags. 

Since the goal of this work is to remove !importants from our styles, I had to move that defined style to something with less specificity than the `<th>` element (`ds-c-table`). My reasoning being, our React component defaults text alignment to "left," so I can assume all text within a table should be left aligned _unless_ the new text-alignment class is applied to overwrite it. 

Manual spot checking and running visual regression testing showed no style regression in this move.

## How to test
<p style="color:red!important">**Please test this one rigorously. I had to move a style for specificity reasons and don't wanna be the one that breaks tables.**</p>

Visit [test url](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1112/table-util-classes/components/table/) and [prod](https://design.cms.gov/components/table/) - there should be no differences between examples.

I ran Backstop before and after refactor - the only error that returned both times was a text-alignment change on the basic table example. This error was expected as the alignment changed in that example recently but reference images weren't updated. I've updated the references for that example for future testing.

### DIFF
--- a/packages/design-system/package.json
+++ b/packages/design-system/package.json
@@ -24,7 +24,7 @@
     "downshift": "^3.2.10",
     "ev-emitter": "^1.1.1",
     "focus-trap-react": "^6.0.0",
-    "lodash.uniqueid": "^4.0.1",
+    "lodash": "^4.17.21",
     "prop-types": "^15.7.2",
     "react-aria-modal": "^2.11.1",
     "react-transition-group": "^2.9.0"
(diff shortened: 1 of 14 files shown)


line47:
We should update the docs page for tables to update this usage guidance because it refers to using a utility class and that's no longer necessary! 👍 

> Usage
Right-align numerical data in tables. You can use a text alignment utility class for this. This makes it easier to scan and compare values.

line47:
@zarahzachz - let's update it to refer to using the prop

PR #1193, opened 2021-10-25
[WNMGDS-1141][WNMGDS-1175] Add the ds-healthcare-gov package to the monorepo

pwolfert:
https://jira.cms.gov/browse/WNMGDS-1141
https://jira.cms.gov/browse/WNMGDS-1175

## Summary
We're moving the existing child design systems (`@cmsgov/ds-healthcare-gov` and `@cmsgov/ds-medicare-gov`) into our monorepo so they are easier to maintain and can be released at the same time as the core design system package. In this first branch, we will move just the healthcare.gov child design system into the monorepo. In a subsequent pull request, we will move the medicare.gov child design system.

This base pull request will be a collection point for other, smaller pull requests that will add functionality incrementally. Do not merge this right away.

- [x] Files copied from https://github.com/CMSgov/hcgov-design-system
- [x] Build scripts updated to support ds-healthcare-gov package in the monorepo for local development and production
- [x] Test scripts updated to support ds-healthcare-gov package in the monorepo
- [x] Set up GitHub Pages to host the healthcare.gov design system documentation
- [x] Update release and publish scripts

### Changed
- Migrated the healthcare.gov child design system (`@cmsgov/ds-healthcare-gov`) into the monorepo

PR #1201, opened 2021-10-28
[WNMGDS-1175] Update testing scripts to support child design systems in monorepo

pwolfert:
https://jira.cms.gov/browse/WNMGDS-1175

See #1193 for context. This PR updates testing scripts to work with the ds-healthcare-gov package.

To test, run `yarn test` or any of the commands that make it up.


packages/ds-healthcare-gov/src/components/Footer/Logo.jsx
```diff
@@ -35,6 +35,7 @@ function Logo(props) {
         'hc-c-footer__logo ds-u-display--inline-block',
         props.className
       )}
+      // eslint-disable-next-line react/no-danger
```
scheul93, line 38: [question] if we're comfortable with `dangerouslySetInnerHTML` used when necessary, can we change the linting rule to be a warning? Or does it make sense to keep it as an error and disable individual files like here

pwolfert: Hmm, looks like apart from the child design system we only use it in the docs package. I think it would be worth going back to these and figuring out if we need it, because I do think we should try to avoid it. How do you feel about it?

package.json
```diff
@@ -14,9 +14,10 @@
     "release": "./release.sh",
     "publish-release": "./publish.sh",
     "precommit": "lint-staged",
-    "test": "yarn test:unit && yarn test:e2e",
+    "test": "yarn test:unit && yarn test:e2e && yarn test:e2e:healthcare",
```
scheul93, line 17: Does healthcare use a relative path to core resources or the official released version?
If it's the latter, I have a question about coupling healthcare tests with core tests...

scheul93 (approved):

package.json
pwolfert: ds-healthcare-gov references the core package through its package name. Lerna hoists the packages up to a root node_modules, and the package is symlinked there

PR #1202, opened 2021-10-29
[WNMGDS-1133] e2e test cleanup

scheul93:
## Summary
Removing browser tests that just rendered a component to the screen. Renamed files to be `.a11y.test`

### Added
* Some components didn't have a11y tests, so I added files for `Accordion`
_Note: some components (`StepList` & `FilterChip`) still don't have a11y tests because violations were flagged when the test files were added. I split the work to add a11y tests and resolve the violations into separate tickets_

### Changed
* Updated build scripts and documentation to reflect updated `test:a11y` command
* Updated file names from `.e2e.test` --> `.a11y.test`
* Updated `chromedriver` dependency so selenium would run

### Removed
* browser tests that just rendered a component to the page. This can be done through visual regression

## How to test
1. Pull down branch
2. Run `yarn install` to get updated chromedriver
3. Run `yarn test` and/or `yarn test:a11y` to confirm that accessibility tests run & pass


scheul93:
Woops. I've got a wonky config somewhere for tests. Looking into it

packages/design-system/src/components/__tests__/smoke.e2e.test.js
```diff
@@ -1,22 +0,0 @@
-/* global driver */
-import { getElementById, getElementByXPath } from '@cmsgov/design-system-scripts/helpers/e2e';
```
scheul93, line 2: @pwolfert Moving our Slack conversation to the PR...
I have deleted this file instead of trying to support one singular e2e test file. My reasoning is that if you run `yarn test:a11y`, as part of the 'global-setup' file, `yarn build` is run. If the doc site generation fails (which is what this test is testing), won't it get caught in that setup file?

pwolfert: Yeah, that's true. This tests that generating all the example pages works. What's not necessarily tested is whether the docs site pages are generated properly, but if we really want that coverage, we can come back later and address it. It's not often that we break the docs site without knowing it. When it happened last week, the whole build command failed, so the `a11y` tests certainly do cover that casse 👍

pwolfert (approved):
LGTM 👍

PR #1204, opened 2021-10-29
[WNMGDS-1175] Run prettier on ds-healthcare-gov files

pwolfert:
See https://github.com/CMSgov/design-system/pull/1201. This PR is separate because it touches a lot of files. When running the linter, it automatically updates files that changed after running prettier.


scheul93 (approved):
If these are just linting resolutions, then ✅

PR #1206, opened 2021-11-01
[WNMGDS-1175] Update browser tests in ds-healthcare-gov

pwolfert:
Reconciling master and https://github.com/CMSgov/design-system/pull/1202 with this branch. Converts existing e2e tests in ds-healthcare-gov to a11y tests.


scheul93:
When I run `yarn test:a11y:healthcare` I see the error
```
design-system/packages/design-system-scripts/gulp/docs/generatePages/generateReactExample.js:46
      const exampleScripts = stats.compilation.assets['bundle.js'].source();
                                                                   ^

TypeError: Cannot read property 'source' of undefined
at design-system/packages/design-system-scripts/gulp/docs/generatePages/generateReactExample.js:46:68
```
Are you experiencing this too? Did I miss a dependency install or some other setup?

pwolfert:
Oh, @scheul93, this is because the core actually has to be built first. Options I can think of are:

1. Document that you need to run `yarn test:a11y` before `yarn test:a11y:healthcare` can be run
    - Problem: Core could be out of date if people forget to do this
2. Add `yarn build` as a first step in `yarn:test:a11y`
    - Problem: Combining it means if you're running them all in a CI job, there's an extra core build step that isn't necessary
    - Problem: It takes longer every time you run it

pwolfert:
Update: We determined that your error was unrelated to the problem I described above.

However, I did look into removing the prerequisite of building the core package by updating the build system to use the core's `src` files instead of `dist` files when building the child design systems. Unfortunately in the case of the Sass files, the fact that the sass files that we distribute need to reference the `dist` folder of the core package (see below) means that in order for styles to compile we're always going to need to have a compiled copy of core. If you can think of any workarounds, please let me know.
```scss
@import '@cmsgov/design-system/dist/scss/base/index';
```

pwolfert:
The Sass compiler doesn't have any options for aliasing different import paths. If we were using something like Webpack, we could alias the path to the file at that level. I also tried https://www.npmjs.com/package/node-sass-alias-importer, but it did not seem to work.

pwolfert:
This is something I'd like to come back and iterate on. In this case moving to the monorepo is not an improvement, but it's not worse than it was before. I will come back in a follow-up PR to streamline the build (or do a larger refactor of the build system) after these child design systems are in master.

PR #1207, opened 2021-11-03
[WNMGDS-1139]helpdrawer rename

zarahzachz:
## Summary
[JIRA ticket](https://jira.cms.gov/browse/WNMGDS-1139)

Created new SlidingPanel components based on HelpDrawer.

### Added
SlidingPanel and SlidingPanelToggle components (includes docs and tests)

### Changed
How HelpDrawer components work internally - they now reference the newly created SlidingPanel components, adding their component-specific stuff on top of it.

## How to test
Review [SlidingPanel](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1139/helpdrawer-rename/components/sliding-panel/) docs to see that it looks and behaves like a generic HelpDrawer. HelpDrawer documentation exists on this page as well and should document differences between HelpDrawer and SlidingPanel.


forrestbaer (approved):

packages/design-system/src/components/SlidingPanel/SlidingPanelToggle.jsx
```diff
@@ -0,0 +1,63 @@
...
+      'ds-c-sliding-panel__toggle',
+      inline && 'ds-c-sliding-panel__toggle--inline',
+      className
+    );
+
+    return (
+      <Button
+        className={classes}
+        inputRef={(el) => (this.buttonRef = el)}
+        onClick={() => showPanel()}
```
scheul93, line 28: [nitpick] I know this is how it was setup in the original panel, but consider updating this to be
```
onClick={showPanel}
```
I think the extra anonymous function is unnecessary here

packages/design-system/src/types/SlidingPanel/SlidingPanel.d.ts
```diff
@@ -0,0 +1,47 @@
...
+   */
+  headingLevel?: SlidingPanelHeadingLevel;
+  /**
+   * Enables "sticky" position of SlidingPanel header element.
+   */
+  isHeaderSticky: boolean;
+  /**
+   * Enables "sticky" position of SlidingPanel footer element.
+   */
+  isFooterSticky: boolean;
```
scheul93, line 41: [question] are `hasFocusTrap`, `headingId`, `isHeaderSticky` & `isFooterSticky` all required props? They don't appear to be in the original HelpDrawer, so wanted to check

zarahzachz: Oh you're right! Adding `headingId` to that list as we have a lodash backup in case this isn't set.

scheul93 (approved):
LGTM

PR #1208, opened 2021-11-03
[WNMGDS-1141] Update release scripts

pwolfert:
### Release script
- Now allows the `lerna version` command to do the tagging and committing. This results in tags for each of the packages that changed
- Creating the version bump branch now happens before the `lerna version` command. This doesn't result in any overall behavioral changes for the script; it's just what I had to do in order to cleanly get the automatically generated bump commit happen on a separate branch instead of master in a way that was easy to revert if someone exited out of the `lerna version` command or said "No" to the final question
- We now print out all the version tags and choose the first one in the example `yarn publish-release` command at the end of the script

## How to test

1. Check out fe13e57ff56e7f3c8c98de0fbc643f9d4551eafe, which removes the teeth of the script. It won't push anything to origin or publish. 
2. Run `yarn release` and then run the command that it gives you at the end
3. Clean up by removing the tags it made with `git tag -d [tags...]` and removing the branch it made. Both of these things are listed in the initial release script output. Example:
    ```git tag -d @cmsgov/design-system-docs@3.0.0 @cmsgov/design-system-scripts@3.0.0 @cmsgov/design-system@3.0.0 @cmsgov/ds-healthcare-gov@7.0.0```

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -26,8 +26,8 @@ if [ "$PRE_VERSION_HASH" = "$POST_VERSION_HASH" ]; then
 fi
 
 echo "${GREEN}Pushing tag and release commit to Github...${NC}"
-git push --set-upstream origin $BRANCH
-git push origin --tags
+# git push --set-upstream origin $BRANCH
+# git push origin --tags
 
 # Grep the last commit message for package versions
 PACKAGE_VERSIONS=$(git log -1 --pretty=%B | grep -o "@.*$")
--- a/release.sh
+++ b/release.sh
@@ -14,10 +14,12 @@ BRANCH="release-${DATE}"
 git checkout -b $BRANCH
 
 echo "${GREEN}Bumping version...${NC}"
+PRE_VERSION_HASH=$(git rev-parse HEAD)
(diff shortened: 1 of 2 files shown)


release.sh
```diff
@@ -8,35 +8,36 @@ YELLOW='\033[0;33m'
...
-TAG_PREFIX=$(node -pe "require('./lerna.json').tagVersionPrefix")
-TAG="$TAG_PREFIX$PACKAGE_VERSION"
-BRANCH="release-$PACKAGE_VERSION"
-
-# Create and push release branch containing the updated package versions
-git checkout -b $BRANCH
-git add --all
-git commit -m "Bump package version to $PACKAGE_VERSION"
 git push --set-upstream origin $BRANCH
+git push origin --tags
```
pwolfert, line 30: Comment this out when testing. It'll create local tags, which you will have to delete with `git tag -d [tags...]`. It'll also create a branch, which you'll have to move away from and then delete.

pwolfert: Whoops, also comment the line above it that pushes the branch

PR #1210, opened 2021-11-04
[WNMGDS-1178] icon rfc update for mgov

scheul93:
## Summary
As part of some icon work discovery, it was called out that Mgov overwrites icons in some components using CSS. Updated the RFC to propose a plan to address that

## Questions
* Are there other considerations that need to be accounted for?


pwolfert (approved):
LGTM 👍

forrestbaer (approved):
👍

PR #1213, opened 2021-11-05
[WNMGDS-1142] Get tests working for ds-medicare-gov

pwolfert:
https://jira.cms.gov/browse/WNMGDS-1142

## Summary
Fixes the tests for https://github.com/CMSgov/design-system/pull/1212

### Changed
- Updated the dev dependencies so we use the latest version of the scripts package
  - Did not update the core package to the latest, and it seems to work. That'll give us more time to make necessary updates and reconcile the child design system to the latest core version
- Updated the cmsds cli config and moved it to the root
- Added yarn scripts for everything
- Updated the browser tests to match the current pattern

## How to test
Run `yarn test:unit` and `yarn test:a11y:medicare`. Linting does not work yet.


forrestbaer (approved):
looks good

scheul93 (approved):
Ran `yarn start:medicare, yarn test & yarn test:a11y:medicare` without issue.
Thanks for updating the a11y tests here

PR #1214, opened 2021-11-05
[WNMGDS-1142] Configure linters for ds-medicare-gov

pwolfert:
https://jira.cms.gov/browse/WNMGDS-1142

## Summary
Get lint checks passing for https://github.com/CMSgov/design-system/pull/1212

### Changed
- Make it so the only `eslint` config that matters for `ds-healthcare-gov` is its own. This means I won't have to fix a bunch of lint errors right now, and we can circle back to standardize our linters in the future
- Completely ignore `ds-healthcare-gov` for `stylelint`, because nobody de-linted any style errors before and there are a lot. Again, this will have to be a separate initiative
- Address the three error-level issues for `eslint` that remained

## How to test
Run `yarn lint`

### DIFF
--- a/.stylelintignore
+++ b/.stylelintignore
@@ -1,0 +1,3 @@
+# Ignore lint errors for ds-medicare-gov until we have a chance to address
+# them separately. The linter has never been run on these files.
+packages/ds-medicare-gov/src/styles/**/*.scss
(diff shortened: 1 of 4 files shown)


forrestbaer (approved):

forrestbaer (approved):
No linter errors, just several warnings, assuming those are OK

PR #1216, opened 2021-11-08
[WNMGDS-1136] Fix box-shadow on HCgov header

zarahzachz:
## Summary
https://jira.cms.gov/browse/WNMGDS-1136

Added style `isolation: isolate` to HC header to force new stacking context. This style is [supported in all browsers (except IE)](https://caniuse.com/?search=isolation).

## Test
TBD


forrestbaer (approved):
I tested by validating the style change worked in safari and chrome on osx, and it looked good. I don't currently know what other cross platform/device test platforms are to be utilized so can take that test for what it's worth 👍

line47:
I think we also need to add a white background to the `.hc-c-header` class - this was an issue that came up a while ago about the drop shadow not showing up when appearing onscreen above a hero graphic

(the div bg color is set on this screenshot and as you can see it bleeds through to the header b/c we are not setting a background color. 
Screen Shot 2021-11-08 at 11 33 04 AM

zarahzachz:
@line47 If you're cool with it, I'm going to add the background color update to another PR to keep the change separate from this one (instantiating a new stacking context on header). I assume we're good with setting it to our white Sass var?

PR #1220, opened 2021-11-09
[WNMGDS 1144] Adding storybook for local development

scheul93:
## Summary
Initial add of storybook for core DS

### Added
- Storybook for core DS local testing
- And example story using the Button component

## How to test
Run `yarn install` at the project root to get all the new dependencies
Run `yarn storybook` to start the storybook instance. This should open `localhost:6006` where you will see 

Screen Shot 2021-11-09 at 3 35 43 PM

There are a few Button stories to see some variations. The prop table is generated automatically, but I did some configuration to make the "controls" more specific

### Future PR Work ###
- Setting up child DS's in storybook. Talking to Patrick, we will want each DS to be separate in storybook. This will help to keep configuration & assets separate
- Setting up other add-ons
- Other component stories


.storybook/CmsTheme.js
```diff
@@ -0,0 +1,36 @@
+import { create } from '@storybook/theming';
```
scheul93, line 1: We can update this and "brand" it more. I used a few colors from core.
Branding will be more important if we host storybook somewhere

package.json
```diff
@@ -22,19 +22,38 @@
     "posttest": "yarn lint",
     "lint": "yarn cmsds lint ./ --ignorePatterns '**/node_modules/**' '**/dist/**' '**/helpers/**' '**/__tests__/**' 'tmp/**' '**/types/**' 'examples/create-react-app-typescript/**' 'examples/create-react-app/**'",
     "prepare": "husky install",
-    "gh-pages": "yarn build:healthcare && gh-pages -d './packages/ds-healthcare-gov/docs/dist' --dest 'healthcare'"
+    "gh-pages": "yarn build:healthcare && gh-pages -d './packages/ds-healthcare-gov/docs/dist' --dest 'healthcare'",
+    "storybook": "start-storybook -p 6006 -s ./packages/design-system/src",
```
scheul93, line 26: `-s ./packages/design-system/src` serves the files at that location as static assets for storybook to reference.
I did this so we could load our fonts properly, but I don't feel great about serving the whole src directory as static assets.

I'm not sure if this is something I should worry about (like for performance), so open to thoughts or ideas

scheul93:
Patrick graciously pointed out that my local build was also using node v14.15.0, so we may need to upgrade our supported node version in the DS for this.

.storybook/main.js
```diff
@@ -0,0 +1,53 @@
...
+          include: path.resolve(__dirname, '../'),
+        },
+        {
+          test: /\.scss$/,
+          use: [MiniCssExtractPlugin.loader, 'css-loader?url=false', 'sass-loader'],
+          include: path.resolve(__dirname, '../'),
+        }
+      );
+
+    return config;
```
pwolfert, line 51: In the next iteration, we could alias these core paths that reference `dist`: https://github.com/CMSgov/design-system/blob/master/packages/ds-healthcare-gov/src/styles/index.scss#L12. That way, when we're live editing core styles and viewing the child design system storybook, we don't have to run `yarn build` before we see those style changes. This is the problem with the CDS migration that I wanted to fix by moving our build scripts to use webpack, but that is a big undertaking. Since we want to use Storybook for local development anyway, having that functionality here means we can put off adding that functionality to the docs site generator for quite a while.

pwolfert: https://webpack.js.org/configuration/resolve/#resolvealias

package.json
```diff
@@ -64,5 +83,6 @@
   "browserslist": [
     "> 2%",
     "Last 2 versions"
-  ]
+  ],
+  "dependencies": {}
```
pwolfert, line 87: Is this needed?

packages/design-system/src/components/Button/Button.stories.jsx
```diff
@@ -0,0 +1,69 @@
...
+      defaultValue: 'Your button text here',
+      type: { name: 'string', required: true },
+    },
+    component: {
+      control: false,
+    },
+    disabled: {
+      control: { type: 'boolean' },
+    },
+    // hiding deprecated props
```
pwolfert, line 23: Good thinking

packages/design-system/src/components/Button/Button.stories.jsx
```diff
@@ -0,0 +1,69 @@
...
+
+const Template = ({ data, ...args }) => <Button {...args} />;
+
+export const DefaultButton = Template.bind({});
+export const InverseButton = Template.bind({});
+InverseButton.args = {
+  inversed: true,
+};
+InverseButton.parameters = {
+  backgrounds: { default: 'dark' },
```
pwolfert, line 59: Cool!

packages/design-system/src/components/Button/Button.stories.jsx
```diff
@@ -0,0 +1,69 @@
+/* eslint-disable react/prop-types */
```
pwolfert, line 1: I wonder if we can disable this at the `.eslintrc` level for all `*.stories.jsx` so we don't have to copy this for all stories.

pwolfert (approved):
Woot! We're off to a great start!

pwolfert:
Oh, and we probably shouldn't merge this until Thursday

forrestbaer (approved):
This is really cool! Love it.

.storybook/main.js
scheul93: Yep! I'll include this with the child system storybook setup because it seems most relevant to that work.

package.json
scheul93: it's not! removed

packages/design-system/src/components/Button/Button.stories.jsx
scheul93: Updated!

forrestbaer (approved):

line47 (approved):
Woohoo, so excited to see this!

forrestbaer (approved):
Changes all look good!

PR #1221, opened 2021-11-09
[WNMGDS-1142] Fix footer SVG rendering

pwolfert:
Context: https://github.com/CMSgov/design-system/pull/1212#issuecomment-964526632

This can either be merged into the existing PR for medicare.gov or rebased after that one is merged, depending on which one is approved first.

Fixes the healthcare.gov `Footer` component, which previously rendered what looked like garbage (the svg markup) rather than logos. The babel config specific to hc.gov was removed, so we needed to switch to using the svg loader plugin that the root babel config uses.

To compare, view the Footer component with `yarn start:healthcare` on both master and this branch.


scheul93 (approved):
LGTM

forrestbaer (approved):

PR #1223, opened 2021-11-10
[NO-TICKET] Rename SlidingPanel to Drawer 👖

zarahzachz:
## Summary
No ticket.

Renamed SlidingPanel to Drawer as it makes more sense when HelpDrawer exists in the system.

## How to test
Visit the [Drawer documentation](http://design-system-demo.s3-website-us-east-1.amazonaws.com/rename-slidingpanel/components/drawer) and make sure everything looks good and makes sense.


packages/design-system/src/components/Drawer/Drawer.jsx
```diff
@@ -49,41 +49,37 @@ export class SlidingPanel extends React.PureComponent {
...
     /* eslint-disable jsx-a11y/no-noninteractive-tabindex, react/no-danger */
-    const slidingPanelMarkup = () => (
-      <div
-        aria-labelledby={this.id}
-        className={classNames(className, 'ds-c-sliding-panel')}
-        role="dialog"
-      >
-        <div className="ds-c-sliding-panel__window">
-          <div className="ds-c-sliding-panel__header">
+    const drawerMarkup = () => (
```
pwolfert, line 52: Not related to the name change, so out of scope, but I just noticed this function and that it could be flattened to just be a value instead of a function that gets called later, since it is always called in every logical branch.

packages/design-system/src/components/Drawer/Drawer.test.jsx
```diff
@@ -1,4 +1,4 @@
-import SlidingPanel from './SlidingPanel.jsx';
+import Drawer from './Drawer.jsx';
```
pwolfert, line 1: Here's another thing that isn't new. Why the file extension?

packages/design-system/src/components/Drawer/DrawerToggle.test.jsx
```diff
@@ -1,56 +1,56 @@
-import SlidingPanelToggle from './SlidingPanelToggle.jsx';
+import DrawerToggle from './DrawerToggle.jsx';
```
pwolfert, line 1: Same here (extension)

pwolfert (commented):
I love how much more consise this name is! I found a few unrelated things that are out of scope that I'm noting below. I'll approve this after the demo builds and I can test it out.

packages/design-system/src/components/Drawer/Drawer.jsx
zarahzachz: You're right. I created a separate PR here #1224

pwolfert:
Found this reference to `sliding-panel` here: https://github.com/CMSgov/design-system/blob/rename-slidingpanel/packages/design-system/src/components/HelpDrawer/HelpDrawer.a11y.test.jsx#L9

pwolfert (approved):

forrestbaer (approved):
Tested and reviewed 👍

PR #1224, opened 2021-11-10
[NO-TICKET] Drawer/flatten markup function

zarahzachz:
Flatten function for Drawer markup.


scheul93 (approved):
Should this branch be merging into https://github.com/CMSgov/design-system/pull/1223?

Maybe it doesn't matter, but the function flattening looks good

PR #1225, opened 2021-11-11
[WNMGDS-1144] storybook for child systems

scheul93:
## Summary
Configuring storybook for child systems

### Added
New commands for starting storybook development for child systems
- `yarn storybook:healthcare` will spin up storybook with all stories from core & healthcare. It will also use healthcare styles
- `yarn storybook:medicare` will spin up storybook with all stories from core & medicare. It will also use medicare styles

Added one child-specific story for each system to test the build.

### Changed
Bumped the .nvmrc node version to the latest version of 14

## How to test
Try out the ☝🏻 commands (make sure to do `nvm use` first!)
I compared the button styles to what I see on the doc site and they looked similar, so I think it's working


line47 (approved):
I was able to load them up and the specific components and styles were there!

packages/design-system/src/components/Button/Button.stories.jsx
```diff
@@ -55,7 +61,7 @@ InverseButton.args = {
   inversed: true,
 };
 InverseButton.parameters = {
-  backgrounds: { default: 'dark' },
+  backgrounds: { default: process.env.STORYBOOK_DS === 'mgov' ? 'Mgov dark' : 'Hcgov dark' },
```
pwolfert, line 64: Nice! Clever

pwolfert (approved):
This is elegant. It turned out less complicated than I thought it might be 👍

PR #1226, opened 2021-11-11
[WNMGDS 1166] Add lightweight alert

line47:
## Summary
Adding a lightweight alert weight option to our alerts

### Added
- Added a new prop `weight` to the alert with the current only option being `"lightweight"`.
- Added new styling to set the background color to white for the lightweight alerts

### Fixed
List any bug fixes.

## How to test
- Add `weight="lightweight"` to an alert component and see that the alert gets the lightweight styling with a color left border and white background
- Demo URL http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1166-lightweight-alert/components/alert/


packages/design-system/src/styles/components/_Alert.scss
```diff
@@ -76,6 +76,9 @@ $alert-icon-size: $spacer-5;
       margin-top: 0;
     }
   }
+  &.ds-c-alert--lightweight {
```
scheul93, line 79: [question] Does this class need extra specificity? Is that why the selector needs to be `.ds-c-alert.ds-c-alert--lightweight`?
If not, consider moving the `ds-c-alert--lightweight` selector to below/outside the `.ds-c-alert` block scope. It removes one degree of specificity and allows easier overwriting down stream

packages/design-system/src/components/Alert/Alert.test.tsx
```diff
@@ -37,6 +37,12 @@ describe('Alert', function () {
     expect(wrapper.hasClass('ds-c-alert--error')).toBe(true);
   });
 
+  it('appears as an lightweight alert', () => {
```
scheul93, line 40: 🎉 A test! 🎉

pwolfert (approved):
Nice work! I tried it with all the variations, and it looks good.

packages/design-system/src/styles/components/_Alert.scss
pwolfert, line 79: Only thing I might change is putting a line break above this

scheul93:
Can an alert be both "lightweight" & "error" or "lightweight" & "success"?

packages/design-system/src/styles/components/_Alert.scss
scheul93: Nevermind. I think I found my answer below AND why the lightweight needed extra specificity 🙃

scheul93 (approved):

PR #1228, opened 2021-11-11
[NO-TICKET] Update m.gov to use latest version of core package

pwolfert:
## Summary
Update the medicare.gov child design system package to use the latest version of the core package. I built the docs site and looked around for issues manually. I found none 🤷‍♂️ . I'm also not getting any errors in the automated tests.

## How to test
This will be most easily tested in a beta release. We'll want to work with our medicare.gov users to work on moving components out of the design system that are specific to individual apps. Since they're quite a ways behind anyway, I think we have time to iterate on this if we find problems and incrementally get the m.gov child design system to a more stable point.


zarahzachz (commented):
I pulled your branch and found some inconsistencies between components:
- `/components/card/` vs https://cmsgov.github.io/mgov-design-system/components/card/
- `/components/list/` vs https://cmsgov.github.io/mgov-design-system/components/list/
- `/components/table/` vs https://cmsgov.github.io/mgov-design-system/components/table/ (section header font-size is larger in monorepo version)
- `/components/tooltip/` vs https://cmsgov.github.io/mgov-design-system/components/tooltip/ (missing examples)
- `/components/usa-banner/` vs https://cmsgov.github.io/mgov-design-system/components/usa-banner/ (icon and link colors don't match)

I don't know if I caught everything, or if these were even introduced with this branch. I really wish we had visual regression testing on Medicare so I could be sure of the differences I saw.

zarahzachz:
Reviewed style differences with Scott and this can be contributed to [some style updates that haven't been released on the Medicare side](https://github.com/CMSgov/mgov-design-system/releases/tag/untagged-68d770b109e07f55821a).

zarahzachz (approved):

PR #1229, opened 2021-11-11
No ticket: Preserve border width for input fields in error state

Josh68:
It looks like the override for the unmodified border class is resetting the width of the border on inputs in error. It should need to override only the `border-color`, not all border properties.

For reference, see [CMSDS TextField CSS](https://github.com/CMSgov/design-system/blob/6757cb3f9b676c0bf010b37b120b1f928de527ef/packages/design-system/src/styles/components/_TextField.scss#L61) and [MCT-6020](https://jira.cms.gov/browse/MCT-6020)

## Summary
Re-applies form field error-state border width of 3px, from the CMSDS

## How to test
The easiest thing to do is look at the live component library and change `border: 1px solid $black-200` to `border-color: $black-200`, for the `.ds-c-field` selector (nested in source)


line47 (approved):
This looks good to me!

PR #1230, opened 2021-11-11
[WNMGDS-1097] Separate our build commands in package.json

pwolfert:
## Summary
Related to work under https://jira.cms.gov/browse/WNMGDS-1097, I want there to be a separate command for just building the packages vs building the docs site. Our underlying `cmsds` command-line utility supports two separate commands, so I'm making our `package.json` script aliases reflect that.

## How to test
Run the `build` and `build-docs` commands to make sure everything works.


forrestbaer (approved):
Tested functional here, LG

PR #1231, opened 2021-11-12
[WNMGDS-1188] filterchip a11y

zarahzachz:
## Summary
https://jira.cms.gov/browse/WNMGDS-1188

### Added
- `pointer-events: none` to `<button>` to prevent events from targeting wrong element
- `aria-live` and instructions for use to our documentation
- `aria-relevant` and instructions for use to our documentation

### Changed
Documentation
- Improved existing info
- Removed bad examples
- Made `ariaClearLabel` optional since we now provide a sensible default

### Fixed
Included Space in list of keycodes that control removing chip.

## How to test
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1188/filterchip-a11y/components/filterchip/

Run our React FilterChip example on Safari + VoiceOver and use keyboard to navigate. `aria-label` should give you a clear idea what interacting with the filter will do, and you should hear an announcement once element leaves the page.


zarahzachz:
@davidakennedy I've updated our FilterChip component to remove the `aria-label` on the span. Could you please review this to make sure it aligns with your expectations? (Adding comment to Jira ticket as well.)

packages/design-system/src/components/FilterChip/FilterChip.tsx
```diff
@@ -56,25 +60,29 @@ export class FilterChip extends React.Component<FilterChipProps> {
...
     );
+
     return (
       <>
         <button
           className={buttonClassNames}
           id={id || uniqueId(`filter_`)}
           onClick={this.handleClick}
           onKeyDown={this.handleKeyDown}
+          aria-label={`. ${ariaClearLabel} ${label}`}
```
scheul93, line 82: [question] is the `.` intentional here?

ericwbailey: Yup! That was my suggestion. The period is intended to provide a pause for a screen reader so the distinction between the filter chip's name and the control to close it are clearer, so something like:

`New York` *pause* `Remove filter` instead of `New Yorkremove filter`.

pwolfert: When I use VoiceOver and Safari and focus on one of these buttons, I hear only the aria-label. What you said above, Eric, makes me think that it should announce `Example big filter chip. Remove Small Example big filter chip`, but what I hear is `. Remove Small Example big filter chip, button`. Are you and Sarah getting different results?

packages/design-system-docs/src/pages/components/FilterChip/FilterChip.example.jsx
```diff
@@ -35,17 +35,13 @@ ControlledFilterChip.propTypes = {
...
       useAlternateIcon
       ariaClearLabel="Remove Alternate"
     />
-    <ControlledFilterChip
-      label="Example filter chip with long text that will wrap Example filter chip with long text that will wrap Example filter chip with long text that will wrap Example filter chip with long text that will wrap Example filter chip with long text that will wrap"
-      ariaClearLabel="Remove Long"
-    />
     <ControlledFilterChip
       label="Example big filter chip"
       ariaClearLabel="Remove Small"
```
pwolfert, line 47: This clear label value doesn't really make sense either

pwolfert: In general these examples don't really match how they'd be used in the real world. I think usually the contents would be much more succinct when used as filters.

pwolfert: Like it's not super clear how the `ariaClearLabel` should relate to the content of the filter

zarahzachz: Oh yeah, I don't think I reviewed the language in our current examples. I think I need direction on what the expected readout for this label is before I can address.

packages/design-system/src/components/FilterChip/FilterChip.tsx
zarahzachz: I'm not entirely sure what the AX should be for the readout. I tried following Target's example, knowing they invest pretty heavily in a11y, and their chips read out: "Remove filter [filter name ]". So Patrick's readout is the intended way.

pwolfert: It doesn't really make sense, though, because don't we want that pause to be here instead `Remove Small . Example big filter chip, button` ?

pwolfert: But the period is at the beginning and there's no break between the `ariaClearLabel ` and the `label` content

zarahzachz: I'd put the `.` break there at @ericwbailey direction. I understood this as providing a break between the previous filter chip read and the current one. I might be wrong about that though.

pwolfert: I'm just telling ya'll what I hear and see in VoiceOver. Do you get a different result with VoiceOver on your machine?

davidakennedy:
> I've updated our FilterChip component to remove the aria-label on the span. Could you please review this to make sure it aligns with your expectations? (Adding comment to Jira ticket as well.)

@zarahzachz Thanks for the ping! And as usual, I appreciate you all thinking about accessibility from the start. 🎉 

A few notes of feedback on this:

- We can remove the `aria-live` region here. Since these are buttons, consumers will be taking an explicit action to remove them by interacting with the button, so they don't need that announced. Also, in most cases, after the button is removed from the DOM, we'll need to send focus elsewhere, so that will announce to screen readers as well. Plus, in some cases, pages may already have another `aria-live` on the page related to filters, so adding that could complicate things. I do like adding one in the future, if this component were part of something larger, like a filter panel. It's a good idea for the design system to handle it whenever possible.
- For the `aria-label`, it reads: `aria-label="Remove Filter Chip Example Filter Chip"` Adding periods is helpful, but I wonder how helpful here at the beginning of the string? I don't hear a pause, and would that matter as much since the element is self-contained? So the screen reader would pause naturally when it moves to a different element. Curious to hear more about this, @ericwbailey. My thought here is to make the `aria-label` read out "Remove [visible text label]". This way, we include the text from the visible label, which is an [important WCAG requirement](https://www.w3.org/TR/WCAG21/#label-in-name). This helps people who use voice control. Also, removing the extra remove filter chip decreases the verbosity for screen readers. @ericwbailey Curious to hear your thoughts here too.

ericwbailey:
@davidakennedy Yeah, I think that could be a direction worth exploring. I agree with @pwolfert's comments about the AX. 

My thinking is it would be a construct that ultimately creates an announcement of `{Visible filter name}. Remove {visible filter name} filter`. So, a more real-world example would be something like, `Avocados. Remove avocado filter.` The reasoning behind this is:

1. Communicate the presence of the filter chip before the suggestion that you can remove it, in case its addition is not announced to AT.
2. Communicate that the filter chip can then be removed after its presence is announced.
3. Specifically don't use the word "chip", as that is jargon and may be confusing.

We might want to investigate using `aria-describedby` in place of `aria-label` so as to not destroy the original accessible name, especially if order of announcement is important.

```html
<button 
  class="ds-c-filter-chip__button" id="filter-chip-{label}" type="button" >
  <span class="ds-c-filter-chip__label" aria-describedby="filter-chip-{label}-instructions">{label}</span>
  <span class="ds-u-visibility--screen-reader" id="filter-chip-{label}-instructions">Remove {label} filter</span>
  <span class="ds-c-filter-chip__clear-icon-container">
    …
  <span>
</button>
```

Here's [a CodePen reduced test case](https://cdpn.io/ericwbailey/debug/xxLNZoR), as well.

zarahzachz:
> Communicate the presence of the filter chip before the suggestion that you can remove it, in case its addition is not announced to AT.

I like this approach tbh - because it makes it more clear that this is a filter before announcing this filter can be removed. But I don't know what the best practice is for this pattern, so I'll defer to that.

scheul93 (approved):

forrestbaer (approved):
lgtm, tested in Safari, responded as expected 👍

PR #1232, opened 2021-11-12
[WNMGDS-1089] Set error-placement flag default in ds-healthcare-gov

pwolfert:
## Summary
We tried to do this a long time ago, but when apps would build in production mode, the flag setting wouldn't work. We discovered this was because our package declared having no side effects, so their build tools were optimizing under the assumption that our modules had no side effects. However, setting that state within our index.js file is indeed a side effect. After testing different `sideEffects` values, we determined that setting it to true had no negative effects on the tested application's overall JavaScript bundle size.

https://jira.cms.gov/browse/WNMGDS-1089

### Changed
- The `ds-healthcare-gov` package now defaults error placement to be below fields

### Fixed
- Fixed flags set in the child design system not holding their value in a production build

## How to test
Make sure errors are placed below fields on the [docs site](http://design-system-demo.s3-website-us-east-1.amazonaws.com/pwolfert/ds-healthcare-flag-defaults)

## Further explanation
Explanation of the root cause as shared in the ticket, pasted here for historical purposes:

At the time this bug was reported, we were trying the flags inside `ds-healthcare-gov`'s `components/index.js` file so that when an application imported anything from the design system, the following code would be executed and the flag would be set:

```js
import { setErrorPlacementDefault } from '@design-system';
setErrorPlacementDefault('bottom');
```

However, the problem with this is that we declare our packages in package.json as having no side-effects via

```json
"sideEffects": false,
```

Code that executes and changes some kind of state when a module is loaded is the definition of a side effect. I have tested and confirmed that changing our ds-healthcare-gov package to

```json
"sideEffects": true,
```

fixes the bug in the downstream app.

In order to determine whether `"sideEffects": true` should be considered as a solution, I did some tests to find out the effect of changing that value on the total JavaScript bundle sizes in this downstream application. Here are the results:

```
HDS sideEffects: false; CMSDS sideEffects: false:

  306.96 KB  build/static/js/2.2ca90886.chunk.js

HDS sideEffects: true; CMSDS sideEffects: false:

  303.11 KB  build/static/js/2.6208dab3.chunk.js

HDS sideEffects: true; CMSDS sideEffects: true:

  313.5 KB  build/static/js/2.673c4939.chunk.js
```

This means that 1) whether or not tree-shaking is enabled for our packages has very little effect on the overall bundle size of this app and 2) there's no negative effect in declaring the healthcare.gov DS as having side effects (assuming this holds true for other applications as well).


forrestbaer (approved):

packages/ds-healthcare-gov/src/components/index.js
```diff
@@ -11,9 +11,13 @@
 
 import './Button';
 
+import { setErrorPlacementDefault } from '@design-system';
```
scheul93, line 14: [question] Can you explain again why it's `@design-system` here instead of `@cmsgov/design-system` like on line 16?

pwolfert: Doh! It shouldn't be. Good catch. It's copypasta.

pwolfert: The docs site builder has a `@design-system` alias that always points to the current design system package (core or child). When I tested this after remaking-this change in our package, I didn't test it in a downstream app, and it just happens to work in my local docs site because of this alias.

scheul93: Ah, thanks. From the pairing session, I wasn't sure if one was needed vs the other in this situation. Thanks for resolving

scheul93 (approved):

PR #1222, opened 2021-11-10
[WNMGDS-1206] mgov/helpdrawer overrides

zarahzachz:
**Note:** This work is dependent on Medicare being part of the monorepo.

## Summary
https://jira.cms.gov/browse/WNMGDS-1206

Updated the Medicare HelpDrawer style override file to match the latest markup for HelpDrawer. Bumped the design system version so Medicare has access to the latest HelpDrawer updates (they needed sticky header/footer props).

## Test

Locally, run:
`yarn install && yarn build && yarn build:medicare && yarn start:medicare`

Then navigate to the Help Drawer page to see the HelpDrawer examples aren't broken (should match https://design.cms.gov/components/help-drawer/)

The override makes use of a wrapping `mct-c-help-drawer` class that was present before the update. I assume a consuming application is using this pattern somewhere?

PR #1233, opened 2021-11-15
[WNMGDS-1206] mgov helpdrawer overrides

zarahzachz:
## Summary
Created a clean branch with cherry-picked commits from #1222 

https://jira.cms.gov/browse/WNMGDS-1206

Updated the Medicare HelpDrawer style override file to match the latest markup for HelpDrawer. Bumped the design system version so Medicare has access to the latest HelpDrawer updates (they needed sticky header/footer props).

## Test

Locally, run:
`yarn install && yarn build && yarn start:medicare`

Then navigate to the Help Drawer page to see the HelpDrawer examples aren't broken (should match https://design.cms.gov/components/help-drawer/)

The override makes use of a wrapping `mct-c-help-drawer` class that was present before the update. I assume a consuming application is using this pattern somewhere?

### DIFF
--- a/packages/design-system-docs/src/pages/components/Drawer/DrawerDefault.example.jsx
+++ b/packages/design-system-docs/src/pages/components/Drawer/DrawerDefault.example.jsx
@@ -1,5 +1,5 @@
 /* eslint no-alert: 0 */
-import { Drawer, DrawerToggle } from '@design-system';
+import { Drawer, DrawerToggle } from '@cmsgov/design-system';
 import React from 'react';
 import ReactDOM from 'react-dom';
 
--- a/packages/design-system-docs/src/pages/components/Drawer/DrawerDefault.example.jsx
+++ b/packages/design-system-docs/src/pages/components/Drawer/DrawerDefault.example.jsx
@@ -1,5 +1,5 @@
 /* eslint no-alert: 0 */
-import { Drawer, DrawerToggle } from '@cmsgov/design-system';
+import { Drawer, DrawerToggle } from '@design-system';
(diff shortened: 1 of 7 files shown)


packages/design-system-docs/src/pages/components/Drawer/DrawerDefault.example.jsx
```diff
@@ -1,5 +1,5 @@
 /* eslint no-alert: 0 */
-import { Drawer, DrawerToggle } from '@design-system';
+import { Drawer, DrawerToggle } from '@cmsgov/design-system';
```
pwolfert, line 2: Why did these need to change?

zarahzachz: Hmmm... So when I ran this locally before the change, there were errors in the console. Forrest commented with finding the same thing and saw changing this import fixed it - but now I'm not able to recreate the original issue. I'm gonna undo this change now.

forrestbaer: Maybe something that was happening before the medicare core version bump I wager

forrestbaer (approved):

packages/design-system-docs/src/pages/components/Drawer/DrawerDefault.example.jsx
zarahzachz: It was certainly odd. If not for the screenshots @forrestbaer took on #1222  I don't know if I wouldn't noticed tbh

PR #1235, opened 2021-11-16
[WNMGDS-1179] - Standardize healthcare.gov icons

forrestbaer:
## Summary / Changed

- HhsLogo, WhiteHouseLogo, UsaLogo and MenuIcon added to core icon library
- Removed those items from subsystem and updated dependencies
- Added flag for use of error icon in InlineError component which utilizes the AlertCircleIcon component

## How to test

- Build and Start medicare to review the HhsIcon and the SimpleFooter in the docs
- Build and Start Healthcare to review the Footer content for logo updates, Header for menu icon updates, and all pages with errors to confirm they include the error icon.
- Build and Start core to confirm the addition of the menu icon to the icon docs


packages/ds-healthcare-gov/src/styles/components/_FormLabel.scss
```diff
@@ -14,3 +14,9 @@ $error-message-icon-size: 28px;
 .ds-c-field__error-message--inverse::before {
   background-image: url('#{$image-path}/error-light-red.svg');
 }
+
```
forrestbaer, line 17: This was added to give backward compatibility to anyone who might be using this class on healthcare.gov without using the component. The override will only be applied to the component version in order to not display the icon twice.

packages/design-system/src/components/InlineError/InlineError.tsx
```diff
@@ -9,24 +11,31 @@ import classNames from 'classnames';
...
 }
 
 export function InlineError({
   children,
   className,
+  displayErrorIcon,
   id,
   inversed,
 }: InlineErrorProps): React.ReactElement {
+  getInlineErrorIconDisplay() ? (displayErrorIcon = true) : (displayErrorIcon = false);
```
scheul93, line 26: [fix] with this ternary, if another component uses `<InlineError>` and passes the prop `displayErrorIcon`, the value of the prop will never get used, it will always be overwritten by the flag.

Consider doing either one of the following
1. Not make `displayErrorIcon` a prop and always use the value of the flag like `const displayErrorIcon = getInlineErrorIconDisplay()`
2. Update this line to be more of an "order of precedence" like use the value of the prop, if that doesn't exist use the flag. Something like `const displayIcon = displayErrorIcon || getInlineErrorIconDisplay()`

packages/ds-healthcare-gov/src/styles/components/_FormLabel.scss
scheul93: I think this works for backwards compatibility. Hopefully we can remove it in the future 🤞🏻

packages/design-system/src/components/Icons/MenuIcon.tsx
```diff
@@ -1,24 +1,22 @@
...
   return (
-    <svg
-      aria-hidden="true"
-      data-prefix="fas"
-      data-icon="bars"
-      className={`svg-inline--fa fa-bars fa-w-14 ${className}`}
-      xmlns="http://www.w3.org/2000/svg"
-      viewBox="0 0 448 512"
-    >
+    <SvgIcon {...defaultProps} {...props} className={iconCssClasses}>
```
scheul93, line 14: [fix] Because the `className={iconCssClasses}` attribute is after the `defaultProps` are spread, the default value for `className` will never be used. <SvgIcon> will always use the value of `iconCssClasses`.

Consider either:
1. Updating `iconCssClasses` to include `svg-inline--fa fa-bars fa-w-14`, if those classes are needed
2. Making the default value of `className: ''` if they are not

forrestbaer: Oh shoot I forgot to update this file thats a copy-paste-forgot to finish thing, will fix!

pwolfert: This is a place where the `classNames` function can be used to avoid messy or complicated interpolated strings.

packages/design-system/src/components/Icons/index.ts
```diff
@@ -9,14 +9,18 @@ export { default as CheckIcon } from './CheckIcon';
 export { default as CloseIcon } from './CloseIcon';
 export { default as CloseIconThin } from './CloseIconThin';
 export { default as DownloadIcon } from './DownloadIcon';
+export { default as HHSLogo } from './HhsLogo';
```
pwolfert, line 12: Nice that this can be shared now 👍

packages/ds-healthcare-gov/src/styles/components/_FormLabel.scss
pwolfert: Maybe we should explain this in the comments so we know when it can be removed.

packages/ds-medicare-gov/docs/src/pages/components/HHSLogo/HHSlogo.example.tsx
```diff
@@ -1,5 +1,5 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { HHSLogo } from '@design-system';
+import { HHSLogo } from '@cmsgov/design-system';
```
pwolfert, line 3: This is intentionally `@design-system`. It's aliased in webpack

packages/design-system/src/components/InlineError/InlineError.tsx
```diff
@@ -9,24 +11,31 @@ import classNames from 'classnames';
...
+
   const classes = classNames(
+    'ds-c-inline-error',
     'ds-c-field__error-message',
     { 'ds-c-field__error-message--inverse': inversed },
     className
   );
+  const viewbox = '36 -6 186 186';
```
pwolfert, line 34: This seems to have done nearly a perfect job of matching the old one. Looks like we're maybe 1 pixel offset between https://cmsgov.github.io/design-system/healthcare/example/components.form-label.react/ and http://localhost:3000/example/components.form-label.react/

pwolfert: I like the new spacing better

packages/ds-medicare-gov/docs/src/pages/components/HHSLogo/HHSlogo.example.tsx
forrestbaer: gotcha, my vscode intellisense marks it as unknown but it works fine on compilation

scheul93 (approved):
LGTM

pwolfert (commented):
I noticed that the HTML example for `FormLabel`/`InlineError` still uses the CSS-based error icon. Should that be updated?

forrestbaer:
> I noticed that the HTML example for `FormLabel`/`InlineError` still uses the CSS-based error icon. Should that be updated?

Yes I left that and made the css override in case that shouldn't be removed because it's being utilized somewhere and could be considered a breaking change? We could remove it from the example to denote that without using the InlineError component you will not get an error icon with your error messages?

pwolfert:
Since we're not changing any existing team's implementation of an HTML-only form label, I don't think it should be considered a breaking change. If we update the example, we're saying _this is how you do it now_. But, like you said, you still left the CSS class for backwards compatibility

forrestbaer:
Gotcha re: breaking change. Issue is that class just adds a background image to ds-c-field__error-message, which contains the other styles for the error message, so we don't want to remove those things.  The html example uses the markup correctly.  I could change the class in _FormLabel.scss to something different but then we'd need to let them know where to use the new class if they still needed that background svg somewhere.. What do you advise?

PR #1236, opened 2021-11-16
Release 2.11.0-beta.1

pwolfert:
This is core 2.11.0-beta.1 and child design system betas with https://github.com/CMSgov/design-system/commit/af78d66666ed52176ca0eb0c0f4d0e216488cad2 cherry-picked onto it so the publish job works


packages/ds-healthcare-gov/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/ds-healthcare-gov",
-  "version": "6.2.0-beta.1",
+  "version": "6.3.0-beta.1",
```
pwolfert, line 3: This jumped from beta to beta because it was migrated prior to the full `6.2.0` release and never had its version number updated

packages/ds-medicare-gov/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/ds-medicare-gov",
-  "version": "3.4.0",
+  "version": "3.4.0-beta.1",
```
pwolfert, line 3: Instead of releasing a 3.4.0, we're going to be releasing a 3.4.0 beta that will live on until we've got the green light to release it next year.

pwolfert:
Ah, it's failing because the beta versions required by some of the packages don't exist yet on npm 😆 . I guess actually merging this has to wait until it's published.

scheul93 (approved):

PR #1240, opened 2021-11-22
Fixing dependabot.yml indentation

scheul93:
## Summary
I'm seeing this error for the dependabot yml file.
Screen Shot 2021-11-22 at 9 36 38 AM

I think it could be two things
1. I ran the yml through a validator and there was some indentation issues that this PR addresses
2. Other thing I'm guessing could be a problem is that maybe dependabot doesn't like duplicate `package-ecosystem: 'npm'`?


pwolfert (approved):
🤞

PR #1279, opened 2021-11-23
[RFC] Design Tokens in Core

forrestbaer:
An RFC for the implementation of Design Tokens in the design system. Discuss!

Invite reviewers as required.


rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+2. There can be a reductive quality to design tokens in that there might be a more limited set of colors/spacing options/etc.. to work with for the consuming applications. Someone might have to sell consuming systems on the advantages to brand consistency, which they hopefully already understand.
+
+## Scope/Outline
+
+**An outline for how this process could proceed:**
+
+1. The SSOT should be decided on, at the moment I see two options:
+
+   1. **The DSM becomes the source of truth**. Designers and developers work together to consolidate the current set of design rules residing in the codebase/Sketch and place them into the DSM. In this case, with that work done, an application can be written in the design system to utilize the [DSM Token API](https://support.invisionapp.com/hc/en-us/articles/360049950931) to periodically update it's set of scss variables with data from these design tokens and store them in a new NPM package.
```
ericwbailey, line 48:
```suggestion
   1. **The DSM becomes the source of truth**. Designers and developers work together to consolidate the current set of design rules residing in the codebase/Sketch and place them into the DSM. In this case, with that work done, an application can be written in the design system to utilize the [DSM Token API](https://support.invisionapp.com/hc/en-us/articles/360049950931) to periodically update it's set of Sass variables with data from these design tokens and store them in a new NPM package.
```

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+   2. **Tokens are modified directly as sets of JSON files**. Designers and developers work together to consolidate tokens in a hierarchy which makes the most sense for all platforms. These tokens could be stored in a new NPM package. Changes to tokens would require a pull request and have a release cadence.
+
+2. Once the SSOT is determined, tokens should be created in the appropriate format and at minimum, populated to form a set required to serve the current consuming applications. Ongoing work could be ticketed to fill out the entire token library. Items which should be tokenized can be decided on by the teams. These first tokens would represent the most basic items such as spacing, color, type, breakpoints, animation, border.. i.e..
```
ericwbailey, line 51:
```suggestion
2. Once the SSOT is determined, tokens should be created in the appropriate format. At minimum tokens should also be populated to form a set required to serve the current consuming applications. Ongoing work could then be ticketed to fill out the entire token library. Items which should be tokenized can be decided on by the teams. These first tokens would represent the most basic items such as spacing and color values, font families and type sizes, breakpoints, animation easing values, border thickness, etc. An example of a color token setup could be:
```

Minor tweaks to make the example basic tokens more discrete.

ericwbailey: >Items which should be tokenized can be decided on by the teams

Should we include this? It suggests to me that different teams can create tokens in a vacuum, but I'm assuming we as the Experience team all collectively decide on what tokens get shared with consuming sources.

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+   			"50": "rgba(43, 199, 252, 50)",
+   			"75": "rgba(43, 199, 252, 75)"
+   		}
+   	}
+   }
+   ```
+
+   > The aliases design tokens are mapped to in the various consuming systems need not match up directly one another initially. The token is a standard atomic unit which can be referenced in many systems using many aliases but should be applied as high up in the dependency tree as possible.
+
+3. When the first dataset is completed and ready for use, a tool stored in the NPM package would need to be created to convert these tokens into new formats, initially SASS. There are several existing tools which can do this such as [Style Dictionary](https://github.com/amzn/style-dictionary) (format is [easily utilized by the DSM](https://support.invisionapp.com/hc/en-us/articles/360049483032)), [Theo](https://github.com/salesforce-ux/theo), [JSON -> CSS](https://github.com/rlapoele/json-to-scss) and others.
```
ericwbailey, line 70:
```suggestion
3. When the first dataset is completed and ready for use, a tool stored in the NPM package would need to be created to convert these tokens into new formats, initially Sass. There are several existing tools which can do this such as [Style Dictionary](https://github.com/amzn/style-dictionary) (format is [easily utilized by the DSM](https://support.invisionapp.com/hc/en-us/articles/360049483032)), [Theo](https://github.com/salesforce-ux/theo), [JSON -> CSS](https://github.com/rlapoele/json-to-scss) and others.
```

Sorry to [be a nudge about this](https://sassnotsass.com/) 😅

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+
+   I believe the best option would be to write our own [Webpack Loader](https://webpack.js.org/contribute/writing-a-loader/) to convert these tokens to top level SASS files.
```
ericwbailey, line 72:
```suggestion
   I believe the best option would be to write our own [Webpack Loader](https://webpack.js.org/contribute/writing-a-loader/) to convert these tokens to top level Sass files.
```

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+
+   1. In the case of the DS-Core, we would convert these tokens into variables which would be utilized by files in our /src/styles/settings/variables scss files and font.scss file initially. Since we would not be changing these files aside from referencing the token variable names in the tokens NPM package instead of providing hard coded values, the compiled output of these files should be compared to detect any breaking changes. Token variables would reside in scss files, checked into the token package. One advantage to using a custom webpack loader in our case is that we could do things like generate the list of @font-face includes based on what tokens are available, slimming overall DS package size.
```
ericwbailey, line 74:
```suggestion
   1. In the case of the DS-Core, we would convert these tokens into variables which would be utilized by files in our `/src/styles/settings/variables` Sass files and `font.scss` file initially. Since we would not be changing these files aside from referencing the token variable names in the token's NPM package instead of providing hard coded values, the compiled output of these files should be compared to detect any breaking changes. Token variables would reside in Sass files, checked into the token package. One advantage to using a custom Webpack loader in our case is that we could do things like generate the list of `@font-face` includes based on what tokens are available, slimming overall DS package size.
```

ericwbailey: >Token variables would reside in scss files, checked into the token package

I'm a little unsure of what this means. Could we clarify?

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+   2. In the case of Sketch and the DSM, the tool chosen/created would be extended to output libraries for the DSM (if needed) and Sketch and these libraries would also be versioned and live in the token package. These would be made available as release download libraries with which to work from. The [Sketch file format](https://developer.sketch.com/file-format/) is very easy to work with as it is already made up of JSON files.
```
ericwbailey, line 75:
```suggestion
   2. In the case of Sketch and the DSM, the tool chosen/created would be extended to output libraries for the DSM (if needed). Sketch and these libraries would also be versioned and live in the token package. These would be made available as release download libraries with which to work from. The [Sketch file format](https://developer.sketch.com/file-format/) is very easy to work with, as it is already made up of JSON files.
```

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+
+4. A Jenkins task for creating these releases would need to be created and tested for CI of changes to the tokens package.
```
ericwbailey, line 77: What would Jenkins do?

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+5. Future work to structure similarities between component level variables in sketch and the codebase would be ongoing from this point. These variables could be added and utilized on a component by component basis, and would make use of the highest level tokens for spacing, color, etc.. i.e.
```
ericwbailey, line 78:
```suggestion
5. Future work to structure similarities between component level variables in Sketch and the codebase would be ongoing from this point. These variables could be added and utilized on a component-by-component basis, and would make use of the highest level tokens for spacing, color, etc.. i.e.
```

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+   				"highlight": {
+   					"color": "ds.core.colors.lightblue.300"
+   				}
+   			}
+   		}
+   	}
+   }
+   ```
+
+6. Documentation would need to be generated from the token build task which could be another option in our existing site docs filed under "Design Tokens." Component level variables could be included on the doc pages for each component.
```
ericwbailey, line 98:
```suggestion
6. Documentation would need to be generated from the token build task. This could be another option in our existing site docs, filed under "Design Tokens." Component level variables could be included on the doc pages for each component.
```

ericwbailey: What would the documentation cover?

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+
+### Other Notes
+
+- There are several ways to transform values from one format to another. If we agree on standard units (px, rem, rgb) for each token type, we can utilize mixins to transform these values to other types dynamically. [A codepen example](https://codepen.io/jakob-e/pen/AHunv).
```
ericwbailey, line 102:
```suggestion
- There are several ways to transform values from one format to another. If we agree on standard units (`px`, `rem`, `rgb`) for each token type, we can utilize Sass Mixins to transform these values to other types dynamically. [A CodePen example](https://codepen.io/jakob-e/pen/AHunv).
```

ericwbailey (changes requested):
I really enjoyed reading this. A couple of high-level questions and a few formatting suggestions, but overall I think it's in really good shape.

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+3. The set of design tokens can be transposed, updated and extended easily and quickly across all systems at once
+4. The set of design tokens can be versioned and changes in that system can be tracked and monitored
+
+By deciding on a [Single Source Of Truth](https://en.wikipedia.org/wiki/Single_source_of_truth) (SSOT) for our design rules, we can organize and transform them for any system which utilizes them.
+
+## Benefits
+
+1. **Consistency** - Between design and code, utilizing a single set of variables everywhere allows a design proof to accurately describe the visual representation of a component as a set of variables. This provides better parity between what's designed and what is seen in the rendered usage on all platforms.
+
+   Looking at the [CSSStats for healthcare.gov](https://cssstats.com/stats?url=healthcare.gov), there are currently 70 color variations, many of which are nearly identical. Utilizing design tokens could bring this number down considerably and provide better brand consistency.
```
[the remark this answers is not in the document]
pwolfert: I don't think the landing page of healthcare.gov currently uses the design system exclusively. From what I've seen at sprint demos, migrating to the design system is work that is currently being done. So I don't think it's fair to use `healthcare.gov` as an example. An app like `healthcare.gov/see-plans` does use the design system exclusively, though, and it has different stats (29 colors).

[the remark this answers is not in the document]
forrestbaer: Ah good point, I didn't realize it didn't use it exclusively!

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+
+Presently, the inconsistency of design rules between the DSM, Designers making and consuming changes, and the design system codebase lead to **confusion**, **duplicated work**, **inconsistency** and **disorganization** within all three systems.
+
+## Proposal
+
+[Design tokens](https://www.invisionapp.com/inside-design/design-tokens/) can help us solve this problem.
+
+> Design tokens are all the values needed to construct and maintain a design system — spacing, color, typography, object styles, animation, etc. — represented as data. These can represent anything defined by design: a color as a RGB value, an opacity as a number, an animation ease as Bezier coordinates. They’re used in place of hard-coded values in order to ensure flexibility and unity across all product experiences. - [Adobe Spectrum](https://spectrum.adobe.com/page/design-tokens/)
+
+Design tokens can be packaged from the atomic level to the application level by utilizing them in a dependency tree where one value inherits a value from the trunk above it.
```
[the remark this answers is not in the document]
forrestbaer: A single key/pair without children.. so a single color, single unit of measure..

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+
+   Looking at the [CSSStats for healthcare.gov](https://cssstats.com/stats?url=healthcare.gov), there are currently 70 color variations, many of which are nearly identical. Utilizing design tokens could bring this number down considerably and provide better brand consistency.
+
+2. **Organization** - By starting from the atomic level, we can use naming conventions for our tokens that are simple and specific to the variable they represent. By keeping these tokens separate from the systems they are utilized in, they can be managed as a discrete unit. Agreement on naming conventions of variables could eventually create parity between Sketch libraries and DS SASS variables.
+3. **Interoperability** - Design tokens allow a huge amount of flexibility for moving a design system to a new platform. Should an iOS/Android app ever be built, a new web application created, etc.. That system would have a framework ready to utilize for its theming which would be consistent with the other brand applications.
```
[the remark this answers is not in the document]
forrestbaer: Yes this is goot 👍

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+4. **Opportunity for theming** - Decoupling design rules from the codebase allowing the transposition-ing of these static design decisions in various ways, i.e. theming/high contrast. By replacing all tokens in a system with new tokens you can essentially theme swap, assuming the consuming system is utilizing tokens at all levels appropriately.
+
+## Risks
+
+1. There is a small risk that implementing tokens could break many things at once, as many design rules would be referencing these across an application. With tools like storybook, visual regression and QA this would be a part of every applications standard testing procedure and should be easy to resolve.
```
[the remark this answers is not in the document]
forrestbaer: This one really isn't a good risk.. I was struggling to come up with them.. We would test changes against the existing variables so it wouldn't be an issue really

rfcs/2021-11-22-design-tokens.md
forrestbaer: I was thinking handling some of the building of the sketch libraries but those should be checked in during pr.. Removing..

rfcs/2021-11-22-design-tokens.md
forrestbaer: Superfluous and confusing, good point.

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Benefits
+
+1. **Consistency** - Between design and code, utilizing a single set of variables everywhere allows a design proof to accurately describe the visual representation of a component as a set of variables. This provides better parity between what's designed and what is seen in the rendered usage on all platforms.
+
+   Looking at the [CSSStats for healthcare.gov/see-plans](https://cssstats.com/stats?url=healthcare.gov/see-plans), there are currently 29 color variations, many of which are nearly identical. Utilizing design tokens could bring this number down considerably and provide better brand consistency.
+
+2. **Organization** - By starting from the atomic level, we can use naming conventions for our tokens that are simple and specific to the variable they represent. By keeping these tokens separate from the systems they are utilized in, they can be managed as a discrete unit. Agreement on naming conventions of variables could eventually create parity between Sketch libraries, DS Sass variables, and future design and development tooling.
```
line47, line 33: I'm really curious to get into how code and Sketch naming can and should be in lock step in terms of naming

rfcs/2021-11-22-design-tokens.md
line47: > Items which should be tokenized can be decided on by the teams.

I had the same thought here as Eric. I don't think product teams should be deciding on what gets tokenized - I think product teams should help inform what should be tokenized through what is needed in terms of flexibility

forrestbaer: Oh that's just poor phrasing, I'll update that.. I meant between design and dev on our team.. But i'll likely just remove that line. 👍

rfcs/2021-11-22-design-tokens.md
forrestbaer: Our conversation Monday should help but also you can impromptu hit me up anytime to chat unless I'm marked as away/busy in slack.

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+   		}
+   	}
+   }
+   ```
+
+5. Documentation would need to be generated from the token build task. This could be another option in our existing site docs, filed under "Design Tokens." Component level variables could be included on the doc pages for each component.
+
+### Other Notes
+
+- There are several ways to transform values from one format to another. If we agree on standard units (`px`, `rem`, `rgb`) for each token type, we can utilize Sass Mixins to transform these values to other types dynamically. [A CodePen example](https://codepen.io/jakob-e/pen/AHunv).
```
line47, line 100: > If we agree on standard units (`px`, `rem`, `rgb`) for each token type

There is interest across CMS to move away from px in favor of rem - this might be a good opportunity to set that standard

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+- There are many tickets currently in the backlog that would be resolved with this work automatically.
+
+### Questions and Requested Feedback
```
line47, line 103: I'm curious about utilize a token set from another design system like USWDS? https://designsystem.digital.gov/design-tokens/ 

Is there a path where we use their token set?

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+
+   1. In the case of the DS-Core, we would convert these tokens into variables which would be utilized by files in our /src/styles/settings/variables scss files and font.scss file initially. Since we would not be changing these files aside from referencing the token variable names in the tokens NPM package instead of providing hard coded values, the compiled output of these files should be compared to detect any breaking changes. One advantage to using a custom webpack loader in our case is that we could do things like generate the list of @font-face includes based on what tokens are available, slimming overall DS package size.
+   2. In the case of Sketch and the DSM, the tool chosen/created would be extended to output libraries for the DSM (if needed). Sketch and these libraries would also be versioned and live in the token package. These would be made available as release download libraries with which to work from. The [Sketch file format](https://developer.sketch.com/file-format/) is very easy to work with, as it is already made up of JSON files.
```
line47, line 74: > The [Sketch file format](https://developer.sketch.com/file-format/) is very easy to work with, as it is already made up of JSON files.

I'm curious to know more about this and how the high level tokens could map to color variables and layer styles in Sketch?

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+   		}
+   	}
+   }
+   ```
+
+   > The aliases design tokens are mapped to in the various consuming systems need not match up directly one another initially. The token is a standard atomic unit which can be referenced in many systems using many aliases but should be applied as high up in the dependency tree as possible.
+
+3. When the first dataset is completed and ready for use, a tool stored in the NPM package would need to be created to convert these tokens into new formats, initially Sass. There are several existing tools which can do this such as [Style Dictionary](https://github.com/amzn/style-dictionary) (format is [easily utilized by the DSM](https://support.invisionapp.com/hc/en-us/articles/360049483032)), [Theo](https://github.com/salesforce-ux/theo), [JSON -> CSS](https://github.com/rlapoele/json-to-scss) and others.
+
+   I believe the best option would be to write our own [Webpack Loader](https://webpack.js.org/contribute/writing-a-loader/) to convert these tokens to top level Sass files.
```
scheul93, line 71: Would we have to change our build tools in order to utilize this custom loader since I don't believe our current build tools use Webpack (aside from Storybook)

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+
+1. There can be a reductive quality to design tokens in that there might be a more limited set of colors/spacing options/etc.. to work with for the consuming applications. Someone might have to sell consuming systems on the advantages to brand consistency, which they hopefully already understand.
+
+## Scope/Outline
+
+**An outline for how this process could proceed:**
+
+1. The SSOT should be decided on, at the moment I see two options:
+
+   1. **The DSM becomes the source of truth**. Designers and developers work together to consolidate the current set of design rules residing in the codebase/Sketch and place them into the DSM. In this case, with that work done, an application can be written in the design system to utilize the [DSM Token API](https://support.invisionapp.com/hc/en-us/articles/360049950931) to periodically update it's set of Sass variables with data from these design tokens and store them in a new NPM package.
```
scheul93, line 47: Does the DSM's design tokens system have some sort of approval process?

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Risks
+
+1. There can be a reductive quality to design tokens in that there might be a more limited set of colors/spacing options/etc.. to work with for the consuming applications. Someone might have to sell consuming systems on the advantages to brand consistency, which they hopefully already understand.
+
+## Scope/Outline
+
+**An outline for how this process could proceed:**
+
+1. The SSOT should be decided on, at the moment I see two options:
```
scheul93, line 45: Is this an internal decision by the team? Or does it involve outside folks?

scheul93 (commented):
Overall, I really enjoyed this and am looking forward to this work 💪🏻

rfcs/2021-11-22-design-tokens.md
forrestbaer: I'm envisioning the tokens being in their own npm package with custom tooling. I have an interest in moving our system [off gulp and onto webpack](https://jira.cms.gov/browse/WNMGDS-1239) so it should slot in nicely over the next.. 6 months?

rfcs/2021-11-22-design-tokens.md
forrestbaer: That's a big ticket but basically it involves transforming the token JSON into JSON usable by Sketch and bundling it up in a zip file (the basic definition of what a Sketch file is) There are a lot of details there to work out before I can answer that. There are typescript and vanilla JS libraries to write the settings and there are even methods to export React components to Sketch, so that could be worked out over two sprints or so.

rfcs/2021-11-22-design-tokens.md
forrestbaer: It seems like that would be tough to work out, it would be a larger shift. That would require a large shift at many levels in our system, what I'm proposing can be done incrementally with smaller changes and deprecations along the way.

rfcs/2021-11-22-design-tokens.md
forrestbaer: Maybe @line47 or @ericwbailey  can answer that? I don't believe the DSM's system offers any advantages over storing these in our own package/repo with git versioning.

rfcs/2021-11-22-design-tokens.md
forrestbaer: This goes back to what we started to talk about tangentially yesterday, it would be great if we could do this without the DSM, I offer it as an option because it exists and someone might force the decision? Unsure.

line47: I'd expect the core team has a proposal and we get that validated by or CMS partners

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
...
+
+2. **Organization** - By starting from the atomic level, we can use naming conventions for our tokens that are simple and specific to the variable they represent. By keeping these tokens separate from the systems they are utilized in, they can be managed as a discrete unit. Agreement on naming conventions of variables could eventually create parity between Sketch libraries and DS SASS variables.
+3. **Interoperability** - Design tokens allow a huge amount of flexibility for moving a design system to a new platform. Should an iOS/Android app ever be built, a new web application created, etc.. That system would have a framework ready to utilize for its theming which would be consistent with the other brand applications.
+4. **Opportunity for theming** - Decoupling design rules from the codebase allowing the transposition-ing of these static design decisions in various ways, i.e. theming/high contrast. By replacing all tokens in a system with new tokens you can essentially theme swap, assuming the consuming system is utilizing tokens at all levels appropriately.
```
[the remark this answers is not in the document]
forrestbaer: Yes it would be easier to evaluate what tokens are being used and which are not with tokens assuming the system utilizes them in some way everywhere.  The inheritance should always lead back to a top level token somewhere.

rfcs/2021-11-22-design-tokens.md
line47: The DSM doesn't have much in terms of an approval process things are either in a draft state or published. I guess you could review things in draft and then publish via the DSM.

rfcs/2021-11-22-design-tokens.md
forrestbaer: We could certainly look at using the USWDS's approaches to the way it handles things [like using REM](https://designsystem.digital.gov/design-tokens/typesetting/overview/), as it seems they've worked a lot of that out already.

rfcs/2021-11-22-design-tokens.md
line47: How would you feel about leveraging a Sketch plugin like Puzzle tokens to pull tokens into sketch? https://github.com/ingrammicro/puzzle-tokens

forrestbaer: I don't like that it adds a software dependence where we could generate native Sketch files that would offer all that functionality and more with our own solution. I'm not against it, I just don't like the dependency.

forrestbaer:
Somewhat related, while going through our Sass dependencies I ran a visualizer and it gave me this, fun to look at:
image

rfcs/2021-11-22-design-tokens.md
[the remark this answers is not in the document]
ericwbailey: Gotcha, cool. It might be helpful to include that as a parenthetical, to help steer the thoughts of a reader who isn't as familiar.

rfcs/2021-11-22-design-tokens.md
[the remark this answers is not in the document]
ericwbailey: Gotcha. Thanks for clarifying!

rfcs/2021-11-22-design-tokens.md
ericwbailey: Nice. I've used functions before to do px-to-rem conversion in Sass, but in the spirit of this RFC, the more we can move these measurements into tokens that ultimately spit out rem values the better 🙂

rfcs/2021-11-22-design-tokens.md
pwolfert: Could you explain how this would work a little more? Would this loader only be used in an intermediary stage where the token files exist in our core package and are not compiled before the design system's own build step? Would there be an import in our Sass files that looks something like
```scss
@import "../tokens/colors.json";
```
and the loader would basically convert `colors.json` into a Sass file that has Sass variables?

pwolfert (commented):
This is looking good! Here are notes from my first pass reading it.

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,106 @@
...
+
+## Scope/Outline
+
+**An outline for how this process could proceed:**
+
+1. The SSOT should be decided on, at the moment I see two options:
+
+   1. **The DSM becomes the source of truth**. Designers and developers work together to consolidate the current set of design rules residing in the codebase/Sketch and place them into the DSM. In this case, with that work done, an application can be written in the design system to utilize the [DSM Token API](https://support.invisionapp.com/hc/en-us/articles/360049950931) to periodically update it's set of Sass variables with data from these design tokens and store them in a new NPM package.
+   2. **Tokens are modified directly as sets of JSON files**. Designers and developers work together to consolidate tokens in a hierarchy which makes the most sense for all platforms. These tokens could be stored in a new NPM package. Changes to tokens would require a pull request and have a release cadence.
```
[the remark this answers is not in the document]
forrestbaer: Interesting, yeah this is a good point, and the sketch file format has a typescript library as well, typescript everything!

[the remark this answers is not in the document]
forrestbaer: It's not necessary and I can understand wanting to put this all in the ds repo. I'm on the fence.

rfcs/2021-11-22-design-tokens.md
zarahzachz: What benefits does writing our own converter have over the previously mentioned options? Is there a maintenance cost on our end for this custom tool? 

My thinking here is that we encountered problems using our custom built docs site to build our components (custom solution) and now we're migrating to Storybook because it's an industry standard and we don't have to spend time maintaining it (third party solution). Is there a risk we end up doing the same thing with tokens?

forrestbaer: @pwolfert yes essentially the loader just converts the json to sass variables we can reference in our DS variable files. This gives us the option to gradually replace the currently used vars with token values. Those variable names would match up to the naming convention designers use in sketch and anyone else utilizing our token library.

@zarahzachz the converter would be very basic for the DS, simply giving us more Sass vars to work with, but those vars and naming conventions would be unified with anyone else using the token library in name/value. The conversion would be a very simple set of transforms. This change shouldn't effect application downstream and just gives us a palette of pre-defined, commonly structured values to work with.

forrestbaer: Does that make sense? Should I expand that to make it more clear?

rfcs/2021-11-22-design-tokens.md
```diff
@@ -0,0 +1,108 @@
+# RFC: Implementation of design tokens
+
+**Date** : Nov 22, 2021
+**Status** : Review
+
+## Problem
+
+Presently, the inconsistency of design rules between the DSM, Designers making and consuming changes, and the design system codebase lead to **confusion**, **duplicated work**, **inconsistency** and **disorganization** within all three systems.
```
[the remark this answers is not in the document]
forrestbaer: Hows this?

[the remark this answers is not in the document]
ericwbailey: I think it's great!

forrestbaer:
I have updated the RFC with some cuts to make it more concise and revisions to make it more exacting. The details of technical implementation are purposefully left off to be included later as a part of this sprints work. Please review when you have a chance. Thanks!

line47:
Just read through and this looks 💯

rfcs/2022-1-10-design-tokens.md
```diff
@@ -0,0 +1,66 @@
...
+[Design tokens](https://www.invisionapp.com/inside-design/design-tokens/) can help us solve this problem.
+
+> Design tokens are all the values needed to construct and maintain a design system — spacing, color, typography, object styles, animation, etc. — represented as data. These can represent anything defined by design: a color as a RGB value, an opacity as a number, an animation ease as Bezier coordinates. They’re used in place of hard-coded values in order to ensure flexibility and unity across all product experiences. - [Adobe Spectrum](https://spectrum.adobe.com/page/design-tokens/)
+
+Design tokens can be codified in many ways using a variety of data storage formats. Since the CMSWDS is implemented in Javascript/Typescript, we can utilize the modularity and type safety of Typescript to manage the current set of styles.
+
+A design token solution is needed in which:
+
+1. The set of design tokens remain finite and equal
+2. The set of design tokens maintain consistency between application (DSM, Sketch, React and other future platforms such as iOS, Android, Web Components)
```
ericwbailey, line 25: Maybe we mention net-new design systems, as well as external design systems we inherit? Those are probably more likely than iOS/Android app work in the short term.

rfcs/2022-1-10-design-tokens.md
```diff
@@ -0,0 +1,66 @@
...
+
+**An outline for how this process could be rolled out for our current system:**
+
+1. A subset of the total token library along with helpers will be created in Typescript and will live in the CMSWDS repository within the packages folder. The following tasks could be completed over the course of two development sprints. This subset will include:
+
+   - The full set of colors used by each system represented in a typescript module.
+   - Naming convention determined with collaboration with the design team.
+   - Type definitions for colors and color format transform helper functions.
+   - _Theme_ typescript modules defined for Core, Hc.goc and M.gov which define Sass variable outputs that map token values to the current color definitions
+   - A webpack/grunt build script which can be implemented in the current build process toolchain to compile these modules into Sass style variables to be imported into each package.
```
ericwbailey, line 55: >A webpack/grunt build script

Maybe genericize this to "- Build scripts" to futureproof it?

ericwbailey (approved):
Two comments to think about, but I don't see them as blocking. I'm really excited for this work!

rfcs/2022-1-10-design-tokens.md
```diff
@@ -0,0 +1,66 @@
...
+   - A set of tests scripts to validate data integrity and compare the variable output to the current variable settings to ensure data parity.
+
+2. Once this concept has been validated and released, work on other token types can progress with: measure, spacing, font, animation and other values, utilizing this toolkit. One set of tokens can be ported at a time, broken up into single-sprint long tasks.
+
+3. A build tool will be developed to export tokens from the token library to Sketch variables which can be imported by designers to ensure they are working with the correct set of values. If the choice to support other design tools ever occurs, a new export process would be architecturally easy to implement for almost any API. This could be done over the course of one development sprint.
```
pwolfert, line 60: This `Scope` section is great. These are separated into pretty discrete tasks. For this third one, I wonder if
1. This could be part of our MVP, which would move it higher in priority than number 3
2. If we should maybe try to tackle a more basic token like _spacing_ first so we can squeeze something like the sketch integration in sooner and prove out more of the whole system before expanding.
Thoughts?

rfcs/2022-1-10-design-tokens.md
```diff
@@ -0,0 +1,66 @@
...
+- **Developers** could eliminate confusion and the hacking together of solutions when presented with a visual change by utilizing the same set of elements as everyone else.
+
+## Proposal
+
+[Design tokens](https://www.invisionapp.com/inside-design/design-tokens/) can help us solve this problem.
+
+> Design tokens are all the values needed to construct and maintain a design system — spacing, color, typography, object styles, animation, etc. — represented as data. These can represent anything defined by design: a color as a RGB value, an opacity as a number, an animation ease as Bezier coordinates. They’re used in place of hard-coded values in order to ensure flexibility and unity across all product experiences. - [Adobe Spectrum](https://spectrum.adobe.com/page/design-tokens/)
+
+Design tokens can be codified in many ways using a variety of data storage formats. Since the CMSWDS is implemented in Javascript/Typescript, we can utilize the modularity and type safety of Typescript to manage the current set of styles.
```
pwolfert, line 20: CMSWDS -> CMSDS

rfcs/2022-1-10-design-tokens.md
```diff
@@ -0,0 +1,66 @@
...
+
+A design token solution is needed in which:
+
+1. The set of design tokens remain finite and equal
+2. The set of design tokens maintain consistency between application (DSM, Sketch, React and other future platforms such as iOS, Android, Web Components)
+3. The set of design tokens can be transposed, updated and extended easily and quickly across all systems at once
+4. The set of design tokens can be versioned and changes in that system can be tracked and monitored
+
+By deciding on a [Single Source Of Truth](https://en.wikipedia.org/wiki/Single_source_of_truth) (SSOT) for our design rules, we can organize and transform them for any system which utilizes them. The design tokens, stored in our codebase and transpilable to any format can be our source of truth for the CMSWDS.
```
pwolfert, line 29: CMSWDS -> CMSDS

rfcs/2022-1-10-design-tokens.md
```diff
@@ -0,0 +1,66 @@
...
+
+## Risks
+
+There are few to no risks involved in implementing a token system. Implementation would be incremental and with 1 to 1 testing of output against the current system.
+
+## Scope/Outline
+
+**An outline for how this process could be rolled out for our current system:**
+
+1. A subset of the total token library along with helpers will be created in Typescript and will live in the CMSWDS repository within the packages folder. The following tasks could be completed over the course of two development sprints. This subset will include:
```
pwolfert, line 49: CMSWDS -> CMSDS

pwolfert (commented):
This looks really nice

rfcs/2022-1-10-design-tokens.md
forrestbaer: Yes makes more sense for #3 to be in place of #2, i've updated that. 

Regarding the type of token to start with, I think color will have the biggest impact/wow factor that would be a good way to kick things off.  I've got the basics complete now I don't think it will be too much to work to possibly even have much of this done this sprint.  I'll grab some of your time tomorrow or monday so you can see where i'm at.

pwolfert: Regarding colors: Okay, cool!

rfcs/2022-1-10-design-tokens.md
```diff
@@ -0,0 +1,66 @@
...
+   - _Theme_ typescript modules defined for Core, Hc.goc and M.gov which define Sass variable outputs that map token values to the current color definitions
+   - Build scripts which can be implemented in the current build process toolchain to compile these modules into Sass style variables to be imported into each package.
+   - A set of tests scripts to validate data integrity and compare the variable output to the current variable settings to ensure data parity.
+
+2. A build tool will be developed to export tokens from the token library to Sketch variables which can be imported by designers to ensure they are working with the correct set of values. If the choice to support other design tools ever occurs, a new export process would be architecturally easy to implement for almost any API. This could be done over the course of one development sprint.
+ 
+4. Once this concept has been validated and released, work on other token types can progress with: measure, spacing, font, animation and other values, utilizing this toolkit. One set of tokens can be ported at a time, broken up into single-sprint long tasks.
```
pwolfert, line 60: Doesn't matter when md is rendered, but this list jumps from 2 to 4

pwolfert (approved):
Excellent work on this, Forrest! This LGTM

PR #1289, opened 2021-11-23
Release 2021-11-23

pwolfert:
I had to create this release by hand. Unfortunately the existing script using `lerna version` doesn't want to bump anything because the packages did not change between the full release and the first beta. I'll need to update the script to account for this flow where we basically just bump the version and re-tag everything.


forrestbaer (approved):
👍

scheul93 (approved):

PR #1292, opened 2021-11-24
[NO-TICKET] Update release script to support bumping unchanged betas to full releases

pwolfert:
## Summary
See the comments in the code. I need to leave now but wanted to throw this up here for feedback. I have not tested this script yet, but the problem I encountered that this is trying to solve is that when checking out an old beta tag and trying to `yarn release`, the `lerna version` command looks at the repo and says that nothing has changed. In fact, what we want to do is force it to bump and re-release the packages that were released in the beta.

"Retag" might not be a good word for it. I just don't have time at the moment to think about it too deeply but wanted to document my findings and my idea for a fix.

### Fixed
- ~Adding a `--retag` argument when running the release script fixes the problem where when trying to make a full release off of beta release tags lerna's bump tool will not recognize any of the packages as having been changed and will not allow you to perform a version bump.~
- Adding a `--force-publish` argument that allows you to specify which packages will be released in the case where they haven't changed since the last beta (which is most of the time)

## How to test
1. Check out latest beta tag with `git checkout tag/@cmsgov/ds-healthcare-gov@6.4.0-beta.4`
2. Comment out these lines:
    ```
    git push --set-upstream origin $BRANCH
    git push origin --tags
    ```
3.  Must commit`git add --all; git commit -m "testing"` because the release script won't let you move forward with uncommitted changes
4. Run the release script with the new argument ~`yarn release --retag`~ and the names of the packages that are being released `yarn release --force-publish @cmsgov/design-system @cmsgov/design-system-docs @cmsgov/ds-healthcare-gov @cmsgov/ds-medicare-gov`
5. Bump all the versions to non-beta and confirm
6. It should have created all the tags locally (see with `git tag`). You'll have to clean this up locally by deleting the tags with `git tag -d <tag-name>`


zarahzachz (commented):
I tried walking through the testing steps and `yarn release --retag` failed on me ☹️

```bash
yarn run v1.22.17
$ ./release.sh --retag
Creating release branch...
Switched to a new branch 'release-2021-12-22'
Bumping version...
$ /Users/sarahannsachs/Documents/Sites/design-system/node_modules/.bin/lerna version --no-push
lerna notice cli v3.21.0
lerna info versioning independent
lerna info Looking for changed packages since @cmsgov/ds-healthcare-gov@6.4.0-beta.4
lerna info ignoring diff in paths matching [
lerna info ignoring diff in paths matching   'packages/stylelint-config-design-system/**',
lerna info ignoring diff in paths matching   'packages/eslint-config-design-system/**'
lerna info ignoring diff in paths matching ]
lerna success No changed packages to version 
No bump commit detected. Removing release branch and exiting...
Note: switching to 'f44eb7ce8b41a0d64ff86e47bec368b9f3211f70'.
```

pwolfert:
The way I was generating the list of packages to force publish was by pulling them out of the last commit message. That’s problematic in our specific testing scenario, because our last commit was not a publish commit, but it’s also problematic for the normal use case of this script, because the latest beta might not of updated all packages (like in the case of beta 4, which only applied to hc.gov)

zarahzachz (approved):

PR #1298, opened 2021-11-26
[WNMGDS-1218] Create RFC for WHCM

zarahzachz:
## Summary
https://jira.cms.gov/browse/WNMGDS-1218


rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+  - Reliance on icons only to convey meaning is lost, especially for Alerts, Buttons with icons, and displaying active states in navigation
+  - Icons should inherit the appropriate text color as mapped to the WHCM theme
+- Border issues
+  - Thin/no borders make it hard to see Badge, FilterChip, HelpDrawerToggle, Tooltip, and the Modal close button
+  - Content in Accordion and Tabs aren't visually grouped together as they are in non-WHCM
+- Buttons (_needs a11y/design direction_)
+  - Use of "ghost style" buttons results in weird padding/margins and/or broken borders
+  - Should buttons and links styled as buttons look the same?
+
+## Proposal
```
scheul93, line 62: A few questions:
1. Apart from the technical implementations for WHCM, is design/UX/accessibility expert input needed to implement fixes?
2. It is unclear to me how much time it would take to implement updates for this. What kind of order would you recommend implementing changes? Would you update for WHCM one component at a time? Or would you address multiple components and focus on resolving/updating each outage category ?

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+### Benefits
+
+- We get to be Good Guy Gregs by supporting a widely used accessibility feature
+- Our design system delivers a consistent experience regardless of user setting
+- Current unusable components will be usable to our users and the end users
+
+### Risks
+
+- Overrides between the media queries outlined may not be a perfect 1:1 match - visual testing will be required
+- Support for `forced-colors` exists in all modern browsers, but the spec is still in draft mode so there's a risk the spec will change
```
scheul93, line 103: So there's a bit of a maintenance risk here? Do you think a big one?

And are there some sorts of tests we could put in place that could catch if a change in spec affects our implementation?

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+And as [MSN says](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors#accessibility_concerns):
+
+> [...] its intended usage is to make small tweaks to improve usability or legibility when the default application of forced colors does not work well for a given portion of a page.
+
+I believe the outages identified above (with maybe the exception of Buttons) fall into this definition.
+
+### Benefits
+
+- We get to be Good Guy Gregs by supporting a widely used accessibility feature
+- Our design system delivers a consistent experience regardless of user setting
```
scheul93, line 97: Maybe expand with the idea that non-WHCM look of the DS should not change with these updates. These updates should be expanding consistent experience...?

I guess, to emphasize that these changes are scoped to a specific mode/setting

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+- Current unusable components will be usable to our users and the end users
+
+### Risks
+
+- Overrides between the media queries outlined may not be a perfect 1:1 match - visual testing will be required
```
pwolfert, line 102: Another risk could be that if there isn't a good way to automate the testing of this, a regression with WHCM could exist for a long time before detection.

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+- Support for `forced-colors` exists in all modern browsers, but the spec is still in draft mode so there's a risk the spec will change
+- Does our team have the bandwidth to tackle these issues?
+
+## Questions and Requested Feedback
+
+TBD
```
line47, line 108: Should we think about documenting developing components that work in WHCM to ensure new components meet our standards?

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
+# RFC: Windows High Contrast Mode (WHCM) Support
+
+**Date**: November 29, 2021
+**Status**: `WIP`
+
+## Background
+
+Windows High Contrast Mode (WHCM) is a unique accessibility feature originally designed for Windows users with contrast sensitivity issues. Other reasons a user might enable WHCM include:
```
ericwbailey, line 8:
```suggestion
Windows High Contrast Mode (WHCM) is a unique accessibility feature originally designed for Windows users with contrast sensitivity and low vision issues. Other reasons a user might enable WHCM include:
```

There's a non-trivial amount of HCM users with low vision who use it along screen zooming magnification to get content into a perceivable state.

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+
+- Reducing visual noise to improve focus
+- Easing eye strain, migraines, or light sensitivity
+- Preferring a specific color scheme
```
ericwbailey, line 12: 💯

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+  - Needs `not-allowed` cursor for additional context
+- Form fields lack usable success, error, and hint text/states
+  - Due to the limited color palette...
+    - Field error and success states look indistinguishable from default field state
+    - Hint and error text look identical to each other
+    - It's not entirely obvious an error has happened in a form field without reading the text
+  - The solution is to avoid relying on color alone when conveying different messages/states (_needs a11y/design direction_)
+    - Different border styles/widths?
+    - Use of italic or bold text?
+    - Label error text with "Error:" language?
```
line47, line 48: Something to note: I think we are moving in the direction of using an Icon next to error messages

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Background
+
+Windows High Contrast Mode (WHCM) is a unique accessibility feature originally designed for Windows users with contrast sensitivity issues. Other reasons a user might enable WHCM include:
+
+- Reducing visual noise to improve focus
+- Easing eye strain, migraines, or light sensitivity
+- Preferring a specific color scheme
+
+With WHCM, users can select a theme of colors for a scoped number of semantic elements. This theme is then applied to user interfaces and applications.
```
ericwbailey, line 14: I might mention how these colors are mapped to a subset of [CSS2 System Color keywords](https://www.w3.org/TR/css-color-3/#css2-system), which tie into Windows OS user interface components. This is the bit that allows web content coloration to be dynamic!

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Proposal
+
+CSS media queries exist to detect if WHCM is enabled and can help developers modify how its styles are applied to specific elements.
+
+`[-ms-high-contrast](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/-ms-high-contrast)` is a Microsoft extension that targets WHCM for IE and older versions of Edge. It supports `none`, `active`, `black-on-white`, and `white-on-black` as values. But it's recommended at this point to only use the `active` value.
+
+`[forced-colors](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/forced-colors)` is a new media query supported by Chrome, Edge, and Firefox. It supports `none` and `active` as values. It should also be noted that this media feature spec is actively being worked on.
+
+[Windows recommends using both media queries](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/) (aka ["The Frankenquery"](https://adrianroselli.com/2021/02/whcm-and-system-colors.html#Frankenquery)) to guarantee adequate support for those using WHCM. This would look something like:
```
ericwbailey, line 70: >The Frankenquery

🤣

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+}
+```
+
+While our design system no longer technically supports IE (as [it falls below our 2% support threshold](https://analytics.usa.gov/health-human-services/)), a good percentage of those who rely on WHCM continue to use it. And because the effort to support IE is low, I agree with Microsoft's recommendation - just be certain your system color keywords work across both queries and the cascade doesn't impact how those styles render (meaning, test the changes). If more targeted styles are needed between IE and Edge, this query may need to be split.
+
+We have access to a SauceLabs account and they have [good documentation for how to do local testing](https://docs.saucelabs.com/secure-connections/sauce-connect/quickstart/).
+
+**And to note, before any work begins:**
+
+It's not the intention of this work to create a separate design for WHCM users. If semantic HTML is used correctly, there should be very little need to overwrite styles. The overall recommendation is to use a light touch when applying fixes. As [Adrian Roselli writes](https://adrianroselli.com/2021/02/whcm-and-system-colors.html#FQs):
```
ericwbailey, line 84: One hundred percent this.

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+
+> [...] its intended usage is to make small tweaks to improve usability or legibility when the default application of forced colors does not work well for a given portion of a page.
+
+I believe the outages identified above (with maybe the exception of Buttons) fall into this definition.
+
+### Benefits
+
+- We get to be Good Guy Gregs by supporting a widely used accessibility feature
+- Our design system delivers a consistent experience regardless of user setting
+- Current unusable components will be usable to our users and the end users
```
ericwbailey, line 98: I like this bullet point as a way to prioritize HCM work. Take the components that are flat-out unusable, sort by prevalence, then chip away at them. Things that technically work, but could benefit from a small tweak or two follow afterwards.

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,108 @@
...
+
+[Windows recommends using both media queries](https://blogs.windows.com/msedgedev/2020/09/17/styling-for-windows-high-contrast-with-new-standards-for-forced-colors/) (aka ["The Frankenquery"](https://adrianroselli.com/2021/02/whcm-and-system-colors.html#Frankenquery)) to guarantee adequate support for those using WHCM. This would look something like:
+
+```css
+@media (-ms-high-contrast: active), (forced-colors: active) {
+  /* CSS overrides here */
+}
+```
+
+While our design system no longer technically supports IE (as [it falls below our 2% support threshold](https://analytics.usa.gov/health-human-services/)), a good percentage of those who rely on WHCM continue to use it. And because the effort to support IE is low, I agree with Microsoft's recommendation - just be certain your system color keywords work across both queries and the cascade doesn't impact how those styles render (meaning, test the changes). If more targeted styles are needed between IE and Edge, this query may need to be split.
```
ericwbailey, line 78: Another thing I might mention is forced colors is a standard, meaning that Chrome, Safari, and Firefox should hypothetically also all roll out support. It'd be good to have our content in good shape for if and when that happens.

I think Chrome in particular will do it sooner than later, as they tend to hoover up all of Edge's Webkit-level accessibility work.

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,185 @@
...
+  .c-swatch__preview {
+    -ms-high-contrast-adjust: none; /* For IE */
+    forced-color-adjust: none; /* For everything else * /
+  }
+}
+```
+
+### Recommendation
+
+**WHCM updates should support IE in addition to modern browsers.**
```
pwolfert, line 134: I think these should be headings like your examples, and it would also be helpful to number them in my opinion

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,185 @@
...
+
+**New and refactored components should be checked in WHCM as part of our "definition of done".**
+
+We have access to a SauceLabs account and they have [good documentation for how to do local testing](https://docs.saucelabs.com/secure-connections/sauce-connect/quickstart/), so we should use it.
+
+**Collaboration between developer and designer may be needed for some issues, but probably not all.**
+
+There are some issues in WHCM that I would consider low severity and I'd think design could trust development to implement a good solution. I'm thinking border-widths, margin/padding on our buttons, and ensuring System Colors are being implemented correctly (like in the disabled state mentioned above). For these kinds of issues, I can envision a developer making an adequate fix and maybe sharing a screenshot with design if needed.
```
pwolfert, line 144: In light of Scott's recent point about inclusive language and making sure we aren't forcing designers and engineers into siloed roles, what if you reframed this to be more about what amount of design effort goes into these different kinds of problems. The ones that "design could trust development to implement" are ones that are straightforward and don't require redesigning the component. The ones that require "designer/developer collaboration" are the ones that need to have their designs revisted or require a more thoughtful approach to coming to a solution.

This is not a fully baked thought. It's just something that was in my head, so take it with a grain of salt.

CC @ericwbailey

rfcs/2021-11-whcm-support.md
```diff
@@ -0,0 +1,185 @@
...
+
+- We get to be Good Guy Gregs by supporting a widely used accessibility feature
+- Our design system delivers a consistent experience regardless of user setting
+- Current unusable components will be usable to our users and the end users
+
+### Risks
+
+- Overrides between the media queries outlined may not be a perfect 1:1 match - visual testing will be required
+- Support for `forced-colors` exists in all modern browsers, but the spec is still in draft mode so the spec may change
+- Does our team have the bandwidth to tackle these issues?
```
pwolfert, line 178: I don't think we need to mention this. We'll either prioritize it or not

pwolfert (approved):
These three examples are really helpful! You did a good job picking scenarios that are really informative. I have a few suggestions below, but otherwise this looks good to me. Nice work!

rfcs/2021-11-whcm-support.md
zarahzachz: Agreed. I was thinking of @line47's comments here as well, but it was hard for me to articulate generic roles. Ideally, collaboration would happen between those who drive the visual/functional language of our system and those who've implemented it. This could be a collab between designer/developer, product/designer, product/developer, etc. I just don't know how to phrase this at all.

ericwbailey: I like design/dev collaboration and I like where your head is at with this, @pwolfert!

PR #1300, opened 2021-11-29
[WNMGDS-1244] Story Tabs Component

forrestbaer:
### Summary
Added a storybook story for the Tabs component

Went around in circles trying to decide if I should be representing submodules as stories then discovered the `submodules` arg.. 🤦 

### Added
Single story with submodules listed under "docs"

### How to test
run `yarn storybook` and see the Tab story, should be at the root level as it doesn't have variations


scheul93 (approved):
Looks good! 👍🏻

pwolfert (approved):
LGTM too, but make sure to update your PR title to use the `[WNMGDS-1244] Title` format before merging.

As a side note, this story made it more apparent to me how badly the tabs behave in mobile. We have guidance on avoiding overflow, but this pattern doesn't seem great in general. I wonder how much it's actually used in downstream apps.

forrestbaer:
Ah good to remember 👍  .. and yes converting tabs on mobile to a dropdown seems like a good pattern, or something more usable there.. I'll put in a ticket

PR #1301, opened 2021-11-29
[NO-TICKET] Remove package lockfiles because yarn workspaces uses only one

pwolfert:
## Summary
When dealing with dependabot pull requests for individual packages, we realized that `yarn` doesn't actually use or care about the package yarn lockfiles. Because we are using _yarn workspaces_, it actually only uses one central lockfile at the root. According to https://classic.yarnpkg.com/lang/en/docs/workspaces/,

> Yarn will use a single lockfile rather than a different one for each project, which means fewer conflicts and easier reviews.

The fact that our packages had lockfiles was just an artifact of pre-monorepo days—a 1.7MB vestigial tail!

### Removed
- Remove package lockfiles because yarn workspaces uses only one

## How to test
Run `yarn install` and make sure nothing changes


scheul93 (approved):

forrestbaer (approved):
nice!!

PR #1315, opened 2021-11-29
[WNMGDS-1148] stories for FilterChip, Drawer & FormLabel

scheul93:
## Summary
Added storybook stories for FilterChip, Drawer & FormLabel

## How to test
Run `yarn storybook` and see newly added stories. I tried to encapsulate some of the major examples while relying on controls for more variations


scheul93:
I'll look into the docs page bug, thanks for catching that!

The toggle intentionally does not open the drawer. I did this because those two components are not actually associated to one another. Yes, they are in the same directory, but there is nothing in the DrawerToggle component that explicitly opens the Drawer. I can update the story, but I intentionally kept the components separate to accurately show their default functionality

forrestbaer:
Ahh gotcha, I wasn't aware of that, sounds good!

forrestbaer (approved):
Woot that fixed it! 👍

pwolfert (commented):
Apparently I never submitted this review yesterday 😆, sorry

packages/design-system/src/components/Drawer/Drawer.stories.jsx
```diff
@@ -0,0 +1,60 @@
...
+  </Drawer>
+);
+
+export const DrawerDefault = Template.bind({});
+export const DrawerWithStickyPositioning = Template.bind({});
+DrawerWithStickyPositioning.args = {
+  isFooterSticky: true,
+  isHeaderSticky: true,
+};
+export const DrawerToggleDefault = () => (
```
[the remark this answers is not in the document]
pwolfert: I see what you've said above about the components being independent of each other, but I would argue that the toggle component is worthless out of context, and it would be good to have an example of the complete thing working as we intended.

packages/design-system/src/components/Drawer/Drawer.stories.jsx
```diff
@@ -60,17 +60,43 @@ const Template = ({ data, ...args }) => (
...
-    Drawer Toggle{' '}
-  </DrawerToggle>
+
+function WithState({ children }) {
+  const [state, setState] = useState({});
+  return <div>{children(state, setState)}</div>;
+}
+export const DrawerToggleWithDrawer = () => (
+  <WithState>
+    {(state, setState) => (
```
pwolfert, line 81: Love that this uses hooks now. Is the `WithState` wrapper function necessary?

pwolfert: Could it just use state inside `DrawerToggleWithDrawer`?

forrestbaer: Fyi I'm working through the same kind of issue right now with pagination but using this: https://medium.com/urbandataanalytics/changing-args-values-state-in-storybook-without-usestate-hook-15cc096716f7

args instead of state

forrestbaer: works great! here is my snippet that works if it helps:

```
const Template = ({...args}) => {
  const [{currentPage}, updateArgs] = useArgs();
  const handleSetPage = (evt, page) => {
    evt.preventDefault();
    updateArgs({ currentPage: page })
  }
  return (
    <>
      <div className="ds-u-text-align--center">
        Current page: {currentPage} / {args.totalPages}
      </div>
      <Pagination
        {...args}
        onPageChange={handleSetPage}
        renderHref={(page) => `#/results/${page}`}
      />
    </>
  );
}```

scheul93: Updated using args for state management!

forrestbaer (approved):
looks goot!

pwolfert (approved):
LGTM too 👍

PR #1316, opened 2021-11-29
[WNMGDS-1238] Add Storybook stories for Tooltip

forrestbaer:
### Summary
Added a storybook story for the Tooltip component

### Added
Icon Trigger Tooltip Story
Inline Trigger Tooltip Story
Interactive Content Tooltip Story
Small update to Tooltip component to update props on componentDidUpdate()

### How to test
run `yarn storybook` and see the new Tooltip stories, play with args to confirm they update


packages/design-system/src/components/Tooltip/Tooltip.jsx
```diff
@@ -47,6 +47,10 @@ export class Tooltip extends React.Component {
     });
   }
 
+  componentDidUpdate() {
+    this.popper.setOptions(this.props);
```
scheul93, line 51: A few questions because I don't remember the inner workings of this component
1. Do all the props for this component apply to popperjs? Or do we need to worry about only passing down a subset?
2. Should there be a conditional in this function that `setOptions` is only called if there is a difference between certain props? Or should `setOptions` be called any time any prop changes?

forrestbaer: Yes all the props deal with popper in some way. It should't effect anything other than modifying layout/attributes/etc  on the popper tooltip if changed.

packages/design-system/src/components/Tooltip/Tooltip.stories.jsx
```diff
@@ -0,0 +1,66 @@
...
+    },
+  },
+};
+
+const Template = ({ data, ...args }) => (
+  <div className="ds-u-display--flex ds-u-align-items--center ds-u-margin-x--4 ds-u-margin-y--3">
+    {data}
+    <Tooltip {...args} />
+  </div>
+);
```
pwolfert, line 25: I noticed that since the tooltip trigger is near the top of the window, there's no way to actually see the `top` positioning. Even if you select `top`, it Popper will never render it on the top because there's no room. I like that it's smart like that, but maybe our example should change to allow people to see that option.

pwolfert (approved):
These look great! One comment below

packages/design-system/src/components/Tooltip/Tooltip.jsx
pwolfert: At some point this will be a functional component, and this will be handled a little more elegantly.

packages/design-system/src/components/Tooltip/Tooltip.stories.jsx
forrestbaer: Good call, updated to add more spacing, can now see top.

PR #1321, opened 2021-12-02
text field storybook stories

scheul93:
## Summary
Added stories for the TextField component that represent
* Single line state
* Multiline state
* error state
* disabled state
* success state
* Masked field

Should the Masked input be its own section? Should there be more variations?

## How to test
Run `yarn storybook` and verify that TextField stories look okay.

### DIFF
--- a/packages/design-system/src/components/TextField/TextField.stories.jsx
+++ b/packages/design-system/src/components/TextField/TextField.stories.jsx
@@ -1,0 +1,77 @@
+import React from 'react';
+
+import TextField from './TextField';
+
+export default {
+  title: 'Components/Text Field',
+  component: TextField,
+  argTypes: {
+    defaultValue: {
+      control: { type: 'text' },
+    },
+    errorMessage: {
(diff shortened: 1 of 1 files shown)


forrestbaer (approved):
👍

packages/design-system/src/components/TextField/TextField.stories.jsx
```diff
@@ -0,0 +1,77 @@
...
+  multiline: true,
+  rows: 3,
+};
+export const MaskedField = Template.bind({});
+MaskedField.args = {
+  mask: 'currency',
+  label: 'Currency Field',
+  inputMode: 'numeric',
+};
+export const ErrorField = Template.bind({});
```
zarahzachz, line 66: Should the ErrorField also have `hint` set? Based on our docs, it looks like error examples always have hint text, but I don't know if that's coincidental.

scheul93: I'll add it!

PR #1361, opened 2021-12-03
[WNMGDS-1145] Add Storybook stories for Accordion, Alert and Badge

forrestbaer:
### Summary
Added storybook stories for the Accordion, Alert and Badge components

### How to test
run `yarn storybook` and see the new stories, play with args to confirm they update


zarahzachz:
Question for these compound components - since there are props for the nested components, is it worth somehow adding those to the wrapping Storybook instance? I'm thinking about Accordion and AccordionItem. There are quite a few AccordionItem props that seem like they'd be worth documenting/allowing users to play with.

packages/design-system/src/components/Alert/Alert.stories.jsx
```diff
@@ -0,0 +1,61 @@
...
+export default {
+  title: 'Components/Alert',
+  component: Alert,
+  argTypes: {},
+  args: {
+    hideIcon: false,
+  },
+};
+
+const Template = (args) => (
```
zarahzachz, line 14: There's examples of Alerts without headings in the docs that should probably be represented. I don't know if you've heard otherwise though.

forrestbaer: No just shooting from the hip with these, that sounds good i'll add one without heading for each type to round them out 👍

forrestbaer:
> Question for these compound components - since there are props for the nested components, is it worth somehow adding those to the wrapping Storybook instance? I'm thinking about Accordion and AccordionItem. There are quite a few AccordionItem props that seem like they'd be worth documenting/allowing users to play with.

I didn't think the AccordionItem props were really usefully adjustable so I've only exposed them through th docs page under the AccordionItem tab using subcomponent. To make these composite components adjustable in that way exposes those elements as new components/stories the sidebar, which to me adds some confusion when they don't always work outside the context of their container.  Which props do you think would be helpful to adjust as args?

zarahzachz:
> > Question for these compound components - since there are props for the nested components, is it worth somehow adding those to the wrapping Storybook instance? I'm thinking about Accordion and AccordionItem. There are quite a few AccordionItem props that seem like they'd be worth documenting/allowing users to play with.
> 
> I didn't think the AccordionItem props were really usefully adjustable so I've only exposed them through th docs page under the AccordionItem tab using subcomponent. To make these composite components adjustable in that way exposes those elements as new components/stories the sidebar, which to me adds some confusion when they don't always work outside the context of their container. Which props do you think would be helpful to adjust as args?

Oh, I missed the AccordionItem props tab on the initial review, so disregard that!

packages/design-system/src/components/Alert/Alert.stories.jsx
forrestbaer: Added an alert without a heading to the default template

scheul93 (approved):
lgtm

zarahzachz (approved):

PR #1362, opened 2021-12-03
[WNMGDS-1149] Create List and MonthPicker stories

zarahzachz:
## Summary
[Jira](https://jira.cms.gov/browse/WNMGDS-1149)

### Added
Stories for List and MonthPicker.

List is more of a utility CSS class and is the first of the components converted. There's still a question about how these kind of components should be organized in the future (do we create a "Utilities/List" section in Storybook?) but because our current docs site has List grouped with other components I've kept it there. 

## How to test
Run `yarn storybook` and compare to [List](https://design.cms.gov/components/list/) and [MonthPicker](https://design.cms.gov/components/month-picker/) doc pages.

### DIFF
--- a/packages/design-system/src/components/List/List.stories.jsx
+++ b/packages/design-system/src/components/List/List.stories.jsx
@@ -4,9 +4,9 @@ import React from 'react';
 export default {
   title: 'Typography/List',
   parameters: {
-    previewTabs: {
-      'storybook/docs/panel': {
-        hidden: true,
+    docs: {
+      source: {
+        type: 'dynamic',
       },
     },
   },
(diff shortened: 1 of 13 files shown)


packages/design-system/src/components/List/List.stories.jsx
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+
+export default {
+  title: 'Components/List',
```
scheul93, line 4: One idea for separating out List is that you could change the title to be 'Utilities/List' and that would put it in a separate grouping than other React components

I'm not sure if "utilities" is the term we want to use, but we could use the title to organize things that aren't quite components

scheul93 (approved):
One idea for the List organization, but the stories look good functionally

forrestbaer:
The MonthPicker looks great but the list I'm not sure we should use the typical component rendering. Under the code tab for each nothing shows up, so essentially it's just a view of a list. Maybe we should change it to just render the html markup like the docs site?

packages/design-system/src/components/List/List.stories.jsx
zarahzachz: Agree. I talked to @line47 and he mentioned that this should go under a "Typography" section - so I'm going to change the title to reflect that.

forrestbaer (approved):
Lgtm! 👍

PR #1366, opened 2021-12-06
[WNMGS-919] Adding Medicare focus styles

line47:
Medicare is inheriting the focus styles from the core design system!

- Removed :focus styles that are no longer needed because they were duplicative of the SCSS set up in core

## How to test
Run the M.gov doc site locally with `yarn start:medicare` and see that the eggplant color is used


zarahzachz (approved):
I'm seeing the purple focus color just fine!

Side note: looks like we have some `tabindex="0"` set on code blocks, so the focus style isn't being applied there. I don't know why these code blocks are focusable to begin with though. To see what I'm talking about, scroll to the bottom of `/components/alert/`.

line47:
> Side note: looks like we have some tabindex="0" set on code blocks, so the focus style isn't being applied there. I don't know why these code blocks are focusable to begin with though. To see what I'm talking about, scroll to the bottom of /components/alert/.

That is weird i don't know why that would be set on the code blocks 🤔

PR #1367, opened 2021-12-06
[WNMGDS-1205] help drawer toggle icon replacement

scheul93:
## Summary
Replacing the info icon in the mgov HelpDrawerToggle component with an inline svg instead of a CSS background image

## How to test
Run `yarn build` to build core. Run `yarn build:medicare` to build medicare and get updated core code. Run `yarn storybook:medicare` and you should see the new HelpDrawer story in storybook

Outstanding questions:
* Can we get a new icon with separate elements for easier CSS styling?
* Can this CSS workaround be a temporary solution?
* How do we deal with the inverse version of this button? (When the button is on a dark background, can we set a color to be the outline & "i" color that will work for all cases)


forrestbaer:
I mean, it works, and it's only in med-gov. So long as we comment the scss that it will be removed in a future release I don't have any problem with it.  The Icon work can be updated when we have the new icons.

scheul93:
More context: The trickiest part is that the info icon in the original implementation is actually two different svgs with different transparent areas.

I tried to replicate that with a singular icon component that would swap the SVG on hover. However, since the icon is in a button, the svg would not swap if the button text is being hovered vs the svg element itself/

Patrick suggested a wrapper component for the Toggle in mgov to help with this functionality, but we'd want to export this component as `HelpDrawerToggle` to not break any apps and I'm not sure if there would be competing component names

forrestbaer:
Makes sense, I don't know of a clean way to do this easily and it seems like doing it at the 'theme' level in scss is most sensible for the time being.

ericwbailey (approved):
I think this is okay for an interim solution. The big win for me is exposing it as an inline SVG and not a background image.

I'll see what I can do about creating a SVG version of this icon with more targetable SVG element classes. I'm also wondering if the SVG class names can have consistency across icons in some way?

line47:
I agree with the folks here I think this is the best path forward fo now. Having all of the icons using our icon component outweighs the non ideal styling.

forrestbaer (approved):

.eslintrc.js
```diff
@@ -22,6 +22,7 @@ module.exports = {
       files: ['*.stories.jsx'],
       rules: {
         'react/prop-types': 'off',
+        'react/no-multi-comp': 'off',
```
pwolfert, line 25: What does this fix?

packages/ds-medicare-gov/src/styles/components/_override.help_drawer.scss
```diff
@@ -35,22 +35,48 @@
...
 .ds-c-help-drawer__toggle {
-  background: url('#{$image-path}/InfoOutline.svg') center right no-repeat;
-  padding-right: $spacer-3;
+  .icon-wrapper {
+    position: relative;
+  }
+
+  .ds-c-icon--info-circle-outline {
+    width: 18px;
+    height: 18px;
```
pwolfert, line 46: Did the original one change size on hover as well? It looks like this one shrinks

pwolfert: I see that it's impractical to fix this right now.

pwolfert (approved):

pwolfert:
Oh, except the PR title format should be updated

.eslintrc.js
scheul93: This is an eslint error that was thrown in a bunch of `.stories` files because it didn't like multiple components being exported, which is what stories need

scheul93:
> Oh, except the PR title format should be updated

Updating!

PR #1374, opened 2021-12-07
[WNMGDS-1151] storybook review skipnav spinner

zarahzachz:
## Summary
[Jira](https://jira.cms.gov/browse/WNMGDS-1151)

### Added
Stories for Review, SkipNav, and Spinner.

Updated Button to include SpinnerButton.

## How to test
Pull branch and run `yarn storybook`. Compare stories to the docs pages for [Spinner](https://design.cms.gov/components/spinner/), [SkipNav](https://design.cms.gov/components/skip-nav/), and [Review](https://design.cms.gov/patterns/review/) to make sure all UI cases accounted for.

### DIFF
--- a/packages/design-system/src/components/Button/Button.stories.jsx
+++ b/packages/design-system/src/components/Button/Button.stories.jsx
@@ -2,6 +2,7 @@ import React from 'react';
 
 import Button from './Button';
 import { NextIcon } from '../Icons';
+import Spinner from '../Spinner/Spinner';
 
 export default {
   title: 'Components/Button',
@@ -30,7 +31,7 @@ export default {
     },
   },
   args: {
-    children: 'You button text here',
+    children: 'Your button text here',
   },
   parameters: {
     backgrounds: {
@@ -71,4 +72,14 @@ IconButton.args = {
       Button with icon <NextIcon />{' '}
     </>
   ),
+};
+
+export const SpinnerButton = Template.bind({});
+SpinnerButton.args = {
+  children: (
+    <>
+      <Spinner /> Loading...
+    </>
+  ),
+  variation: 'primary',
 };
--- a/packages/design-system/src/components/Review/Review.stories.jsx
+++ b/packages/design-system/src/components/Review/Review.stories.jsx
@@ -1,0 +1,60 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
+import React from 'react';
+
+import Review from './Review';
+
+export default {
+  title: 'Patterns/Review',
+  component: Review,
+  argTypes: {
+    headingLevel: {
+      control: {
+        type: 'select',
+      },
+      options: [1, 2, 3, 4, 5, 6],
+      defaultValue: 3,
+    },
+  },
+};
+
+const Template = ({ data, ...args }) => <Review {...args} />;
+
+export const DefaultReview = Template.bind({});
+DefaultReview.args = {
+  children: 'This is an example of a default Review component.',
+};
+
+export const SingleReview = Template.bind({});
+SingleReview.args = {
+  heading: 'A single Review component',
+  children: 'This is an example of a single Review component.',
+  editHref: 'javascript:void(0)',
+};
+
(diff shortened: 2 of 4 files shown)


packages/design-system/src/components/Button/Button.stories.jsx
```diff
@@ -72,3 +73,13 @@ IconButton.args = {
     </>
   ),
 };
+
+export const SpinnerButton = Template.bind({});
+SpinnerButton.args = {
+  children: (
+    <>
+      <Spinner /> Loading...
```
zarahzachz, line 81: I had initially imported the `DefaultSpinner` story for this, but since it appears the Button overwrites Spinner props (can't change sizing or color), I imported the Spinner component itself.

forrestbaer: will it work without the fragment wrapper? It's adds the <Fragment> to the code example.. no big if not.

forrestbaer:
I wasn't able to tab to the skipnav, am I doing something wrong?

packages/design-system/src/components/SkipNav/SkipNav.stories.jsx
```diff
@@ -0,0 +1,39 @@
+/* eslint-disable jsx-a11y/anchor-is-valid */
+
+import React from 'react';
+
+import SkipNav from './SkipNav';
+
+export default {
+  title: 'Components/Skip Nav',
+  component: SkipNav,
```
scheul93, line 9: I'm seeing the console error 
```
The prop `href` is marked as required in `SkipNav`
```
Consider resolving this with a default arg passed to the SkipNav like
```
args: {
    href: '#'
  }
```

packages/design-system/src/components/SkipNav/SkipNav.stories.jsx
```diff
@@ -0,0 +1,39 @@
...
+      {Story()}
+      <nav className="ds-u-padding--3">
+        <h1 className="ds-u-margin--0">Navigation</h1>
+        <p className="ds-u-margin--0">Clicking on the Skip Nav will skip over this section.</p>
+        <ul>
+          <li>
+            <a href="javascript:void(0);">Navigation link 1</a>
+          </li>
+          <li>
+            <a href="javascript:void(0);">Navigation link 2</a>
```
scheul93, line 29: I see a console error that reads
```
react-dom.development.js:88 Warning: A future version of React will block javascript: URLs as a security precaution. Use event handlers instead if you can. If you need to generate unsafe HTML try using dangerouslySetInnerHTML instead. React was passed "javascript:void(0);".
```
Consider updating the href of these to something like `#` or `google.com`

packages/design-system/src/components/Review/Review.stories.jsx
```diff
@@ -0,0 +1,60 @@
...
+
+export default {
+  title: 'Patterns/Review',
+  component: Review,
+  argTypes: {
+    headingLevel: {
+      control: {
+        type: 'select',
+      },
+      options: [1, 2, 3, 4, 5, 6],
```
scheul93, line 14: I'm seeing a console error about invalid prop type. I think the component is expecting a string but is getting a number for the `headingLevel` value

packages/design-system/src/components/Button/Button.stories.jsx
zarahzachz: Unfortunately not, because of the sibling "Loading..." text you'd get a warning that there needs to be a wrapping container. I experimented with [changing `docs.source.type` with a parameter](https://storybook.js.org/docs/react/writing-docs/doc-blocks#docspage-1), but it seems like the fragment's going to be stuck there.

packages/design-system/src/components/SkipNav/SkipNav.stories.jsx
zarahzachz: Thanks for catching this - I went with "#".

forrestbaer (approved):
👍

PR #1375, opened 2021-12-08
[NO-TICKET] Dependency major bumps

pwolfert:
These are all the major version bumps that Dependabot wants to make that don't seem to require any changes on our end. I've left out several of the major version bump branches because they caused errors or other problems.


zarahzachz (approved):
Not sure if this is ready for review, but I built and ran Core, HC and Mgov locally and didn't run into any issues.

pwolfert:
ESLint failed, so I removed the bump from 6.8.0 to 8.4.0. I think there are a lot of problems there that we'd be better served addressing in one linter-updating ticket. Our lint rules need love anyway.

PR #1377, opened 2021-12-09
[WNMGDS-980] - Convert UsaBanner component to typescript

forrestbaer:
## Summary
Converted UsaBanner to a typescript FunctionComponent, updated test snapshots and docs site reference.

### Changed
Added json importing setting to global .tsconfig

### Removed
Removed jsx version of component

## How to test
Run `yarn build && yarn start` to confirm component compiles and works as before on docs site
Run `yarn storybook` to confirm component works as expected in storybook
Run `yarn test` to confirm no tests are broken


packages/design-system/src/components/UsaBanner/UsaBanner.tsx
```diff
@@ -0,0 +1,107 @@
+import React, { useState } from 'react';
+import { Button } from '../Button';
+import classNames from 'classnames';
+import uniqueId from 'lodash/uniqueId';
+import { LockCircleIcon, LockIcon, UsaFlagIcon, BuildingCircleIcon } from '../Icons';
+import EnglishTranslations from '../../locale/en.json';
+import SpanishTranslations from '../../locale/es.json';
+
+type LocaleLanguage = 'en' | 'es';
```
scheul93, line 9: Should this be exported so it could potentially be used in consuming applications?

packages/design-system/src/components/UsaBanner/UsaBanner.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+interface UsaBannerProps {
```
scheul93, line 11: Should this be exported?

tsconfig.json
```diff
@@ -3,6 +3,7 @@
     "target": "es2015",
     "jsx": "react",
     "moduleResolution": "node",
+    "resolveJsonModule": true,
```
pwolfert, line 6: 👍

scheul93 (approved):

pwolfert (approved):
LGTM 👍 

It doesn't seem like our props table generator does a good job picking up on the details of the types we use (locale doesn't show actual values it accepts, just the TS type name). Has anyone found a workaround for other TS components?

Screen Shot 2021-12-09 at 9 56 17 AM

Looks like Storybook does a better job at it:

Screen Shot 2021-12-09 at 9 57 35 AM

forrestbaer:
> It doesn't seem like our props table generator does a good job picking up on the details of the types we use (locale doesn't show actual values it accepts, just the TS type name). Has anyone found a workaround for other TS components?

Hmm, we would probably need to work on ReactProp docs to get that working. Maybe another ticket?

pwolfert:
We might just wait until a day when the docs generated by Storybook are used in the docs site 🤷‍♂️

PR #1382, opened 2021-12-09
[WNMGDS-978] tabs typescript conversion

scheul93:
## Summary
Converting `Tabs`, `Tab` & `TabPanel` into Typescript

## How to test
Run `yarn storybook` and ensure component functions as expected.

Run `yarn start` to ensure doc site for Tabs is rendering as expected.


scheul93:
Several props in these components are functions. I took a chance at explicitly defining what the function would look like instead of using `(any : any[]) => any`. Are there any objections? This would be a breaking change

forrestbaer:
I was under the impression we would be moving to all function components instead of class based. Was I incorrect, maybe that's something to move towards in the future? So far I've been converting them to functionComponents.. 

In any event, whatever pattern we choose, it would be good to get on the same page as far as our patterns.  Apparently after some reading there is some debate over using the verbose React.FC/React.FunctionComponent vs the more terse (): type pattern because of how React.FC assumes children props. @zarahzachz, @pwolfert, Anna, what do you think would be best?

pwolfert:
I've never bothered explicitly typing a component as `React.FC`, and it has worked for me. What are the benefits?

forrestbaer:
The difference is that FunctionComponent comes with one property by default: children, also using React.FunctionComponent enables some editor codesense stuff..  It's not different really.  The main question I have is, are we moving all components eventually to using function components and hooks or are we ok with a mix of class based components and functional components?  If we're doing functional components I think the pattern you outlined is best as its very simple to read and direct.

scheul93:
I'll update and convert to functional components. 1) because now's a good time to do so & 2) because I really dislike class components, so I'm happy to reduce our usage of them.

forrestbaer:
Cool sounds good, I'm going to open another ticket to update the one I did yesterday as well to convert it to the shorter syntax, looks cleaner.

packages/design-system/src/components/Tabs/Tab.tsx
```diff
@@ -1,48 +1,95 @@
...
-  render() {
+  render(): JSX.Element {
     const classes = classnames('ds-c-tabs__item', this.props.className);
     const sharedTabProps = {
       role: 'tab',
       className: classes,
       id: this.props.id,
       ref: (tab) => {
         this.tab = tab;
+        console.log();
```
zarahzachz, line 84: Is this `console.log()` needed?

scheul93: Sure isn't! It's a remnant from some testing, so I will clean it up

packages/design-system/src/components/Tabs/TabPanel.tsx
```diff
@@ -1,67 +1,64 @@
...
   // this id within the Tabs component. Otherwise React will yell at you even
   // though it's ultimately being passed in.
   /**
    * The `id` of the associated `Tab`. Used for the `aria-labelledby` attribute.
    */
-  tabId: PropTypes.string,
+  tabId?: string;
+}
+
+export function TabPanel(props: TabPanelProps): JSX.Element {
```
forrestbaer, line 40: Wondering, for consistency could we write TS functions like:

`export const TabPanel = (props: TabPanelProps) => {`

And perhaps leaving off the return type (unless it's more specific like a string/bool/int) and allowing TS to infer it in these cases?  It seems like strongly typing some of these might cause issues down the line, especially with JSX.Element as it's not as permissive as React.ReactNode.

I am mainly suggesting this because I prefer this syntax, feel free to ignore 🖖

packages/design-system/src/components/Tabs/Tabs.tsx
```diff
@@ -1,23 +1,56 @@
...
+  return child && (child.type === TabPanel || componentName === 'TabPanel');
+}
+
 /**
  * Get the id of the first TabPanel child
  * @param {Object} props
  * @return {String} The id
  */
-function getDefaultSelectedId(props) {
+function getDefaultSelectedId(props): string {
```
forrestbaer, line 53: Similar to above, in this file we mix function syntax, it would be great to agree on one so we can all adopt a pattern for consistency, my vote in the last comment

packages/design-system/src/components/Tabs/TabPanel.tsx
forrestbaer: @zarahzachz @pwolfert as I commented here and below, maybe we can agree on a function syntax and when to use return values as a way of structuring a new team norm? Thoughts?

forrestbaer:
All looks good and functional beyond my hijacking of your PR to discuss coding patterns 🤦

forrestbaer (approved):

packages/design-system/src/components/Tabs/TabPanel.tsx
scheul93: I'm happy to convert to the `export cont TabPanel = () => {}` syntax for all functions

With regard to the return type, I kind of agree that JSX.Element is a strict typing and could potentially cause problem eslint throws a warning if a function doesn't have a return type. 😞

forrestbaer: Right on! I think that's one of the eslint warnings we should turn off in the future

pwolfert: Agreed

zarahzachz (approved):

PR #1387, opened 2021-12-10
[WNMGDS-973] - Convert Review component to typescript

forrestbaer:
## Summary
Updated Review component to use condensed functional typescript syntax

### Added
Added an ignore to our extractReactData gulp runner so it would not try to extract props from storybook stories

## How to test
run `yarn build && yarn start`, run `yarn test`, ensure no changes or new errors


packages/design-system-scripts/gulp/docs/extractReactData/index.js
```diff
@@ -21,7 +21,7 @@ async function extractReactProps(sourceDir, options) {
     gulp
       .src([
         `${sourcesGlob}/**/*.{jsx,tsx}`, // React props
-        `!${sourcesGlob}/**/*{.test,.spec,.d}.{js,jsx,ts,tsx}`,
+        `!${sourcesGlob}/**/*{.stories,.test,.spec,.d}.{js,jsx,ts,tsx}`,
```
forrestbaer, line 24: Added this to resolve these, doesn't effect anything else, ok to keep in here?

image

forrestbaer: We don't need to run .stories files through the docs site prop doc generator

packages/design-system-docs/src/pages/components/Review/Review.example.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { Review } from '@design-system';
+import { Review } from '@design-system/packages/design-system';
 
-const editButtonHref = 'javascript:void(0);';
+const editButtonHref = '#!';
```
forrestbaer, line 5: Fixes JS error, react deprecating use of scripts in hrefs

pwolfert: What does the `!` do?

packages/design-system-scripts/gulp/docs/extractReactData/index.js
pwolfert: Nice! Thank you

packages/design-system-docs/src/pages/components/Review/Review.example.tsx
```diff
@@ -1,8 +1,8 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { Review } from '@design-system';
+import { Review } from '@design-system/packages/design-system';
```
pwolfert, line 3: `@design-system` is an alias, so this shouldn't need to be changed. Does it break without this change?

packages/design-system/src/components/Review/Review.tsx
```diff
@@ -0,0 +1,73 @@
...
+  onEditClick?: (...args: any[]) => any;
+}
+
+export const Review = (props: ReviewProps) => {
+  const getHeading = () => {
+    const Heading = (`h${props.headingLevel}` || `h3`) as keyof JSX.IntrinsicElements;
+    if (props.heading) {
+      return <Heading className="ds-c-review__heading">{props.heading}</Heading>;
+    }
+  };
```
pwolfert, line 45: Up to you, but this could be moved out of the component with a parameter now that you've made it into a nice little function.

packages/design-system/src/components/Review/ReviewLink.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+export const ReviewLink = (props: ReviewLinkProps) => {
+  const handleClick = useCallback(
+    (event) => {
+      if (props.onClick) {
+        props.onClick(event, props.href);
+      }
+    },
+    [props.onClick, props.href]
+  );
```
pwolfert, line 23: I think `useCallback` [is overkill](https://kentcdodds.com/blog/usememo-and-usecallback#why-is-usecallback-worse) here because it's unlikely that those two prop dependencies will change after first render.

forrestbaer: good to know!

packages/design-system-docs/src/pages/components/Review/Review.example.tsx
forrestbaer: It does, having trouble understanding why

packages/design-system-docs/src/pages/components/Review/Review.example.tsx
forrestbaer: It references the id = ! which doesn't exist so it's similar to # but doesn't cause anything to happen, where using # on it's own still throws an error in the console based on the react dep warning.. so it's a bit of a hack to make it work without an error.. best way I know of!

packages/design-system-docs/src/pages/components/Review/Review.example.tsx
forrestbaer: @cmsgov/design-system also works

packages/design-system-docs/src/pages/components/Review/Review.example.tsx
pwolfert: Oh that's interesting! TIL

packages/design-system-docs/src/pages/components/Review/Review.example.tsx
pwolfert: Hmm, we're going to be running into this problem a lot then, moving forward. Maybe we should avoid converting the example files for now

forrestbaer: Sounds good, I'll see if I can figure out why first and if I can't will put this back to JSX

packages/design-system/src/components/Review/Review.tsx
```diff
@@ -0,0 +1,74 @@
...
+  heading?: React.ReactNode;
+  /**
+   * Heading type to override default `<h3>`.
+   */
+  headingLevel?: ReviewHeadingLevel;
+  /**
+   * An optional function that is executed on edit link click. The event and
+   * props.editHref value are passed to this function.
+   */
+  onEditClick?: (...args: any[]) => any;
```
scheul93, line 36: General question that I also asked in my Tabs PR -- should we be more strictly typing these callback functions? Like with set parameters?

scheul93 (approved):
LGTM, one question about our TS conversions in general

packages/design-system/src/components/Review/Review.tsx
forrestbaer: I don't know that we should unless the function is defined with a particular set of params that should be typed.. in this case it's typed generically as a func so the caller can do anything they want with onEditClick.

packages/design-system-docs/src/pages/components/Review/_Review.docs.scss
```diff
@@ -14,9 +14,9 @@ Style guide: patterns.review
 
 The `Review` component wraps up the styles and markup required to make a single reviewable row. They are designed to be used either alone, as a single row, or multiple rows can be combined together one after the other to present more information. The component can take text or HTML as row content.
 
-@react-example Review.example.jsx
+@react-example Review.example.tsx
```
zarahzachz, line 17: I don't think we've been converting our example files to Typescript. Not sure why though.

packages/design-system/src/components/Review/Review.tsx
zarahzachz: If we opt to more strictly type these props, it should probably happen after the initial TS conversion (with a major release). I'm personally OK with keeping the type interfaces as they are and revising later since I imagine other interfaces need reviewed as well.

packages/design-system/src/components/Review/Review.tsx
```diff
@@ -0,0 +1,74 @@
...
+}
+
+const getHeading = (heading, headingLevel) => {
+  const Heading = (`h${headingLevel}` || `h3`) as keyof JSX.IntrinsicElements;
+  if (heading) {
+    return <Heading className="ds-c-review__heading">{heading}</Heading>;
+  }
+};
+
+export const Review = (props: ReviewProps) => {
```
zarahzachz, line 46: Should this function return something like `React.ReactElement` or `React.ReactNode`?

packages/design-system/src/components/Review/ReviewLink.tsx
```diff
@@ -0,0 +1,36 @@
...
+   * explain the context of the link.
+   */
+  ariaLabel?: string;
+  className?: string;
+  children: React.ReactNode;
+  href: string;
+  onClick?: (...args: any[]) => any;
+}
+
+export const ReviewLink = (props: ReviewLinkProps) => {
```
zarahzachz, line 15: Should this return `React.ReactElement` or something?

packages/design-system/src/components/Review/ReviewLink.tsx
```diff
@@ -0,0 +1,36 @@
...
+  const handleClick = (event) => {
```
zarahzachz, line 16: Should the event be typed?

packages/design-system/src/components/Review/Review.tsx
forrestbaer: yep agreed, could run into troubles with integration

packages/design-system/src/components/Review/Review.tsx
forrestbaer: No the pattern I suggested in Anna's PR that I think we're going with is to leave react component returns off and allow ts to infer this, they can only cause possible problems at this point. We're only setting return types on simple values like bool, string, etc..

packages/design-system/src/components/Review/ReviewLink.tsx
forrestbaer: sure it can be typed as void 👍

PR #1388, opened 2021-12-10
[NO-TICKET] Revert major version upgrade of css-loader

pwolfert:
## Summary
This is not strictly a revert of what was done in https://github.com/CMSgov/design-system/pull/1375, as I've made the version number `^5` instead of the previous `5`. This seems to work after testing with Storybook. Prior to this change, Storybook would not build.

### Fixed
Fixes Storybook failing to build after https://github.com/CMSgov/design-system/pull/1375

## How to test
`yarn install && yarn storybook`


forrestbaer (approved):
👍

PR #1391, opened 2021-12-13
[WNMGDS-981] VerticalNav TS/function component conversion

zarahzachz:
## Summary
[Jira ticket](https://jira.cms.gov/browse/WNMGDS-981)

### Changed
I refactored the `VerticalNav` and related components from JSX to TSX. Also migrated from class to function components.

## How to test
Build this branch locally and compare against the [VerticalNav docs](https://design.cms.gov/components/vertical-nav/).


packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
```diff
@@ -1,137 +1,195 @@
...
-  }
+  const [collapsed, setCollapsed] = useState(props.defaultCollapsed);
+
+  // Refactor from `componentDidUpdate` to `isEffect` using this example:
+  // https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
+  React.useEffect(() => {
+    if (props.onSubnavToggle) {
+      props.onSubnavToggle(props.id, collapsed);
+    }
+  }, [collapsed]);
```
pwolfert, line 83: I don't actually know why this was in a `componentDidUpdate` in the first place and not just called in the `handleToggleClick` function. Do you know? That seems to be the only place we call `setCollapsed` anyway.

pwolfert: Maybe at some point in its history, there were multiple things that could set the collapsed state, and it seemed cleaner to make a different function that called this whenever that state changed

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
```diff
@@ -1,137 +1,195 @@
...
+   */
+  const isSelected = (): boolean => {
+    if (props.selected) return props.selected;
+
+    if (props._selectedId && hasSubnav()) {
+      return childIsSelected(props.items);
+    }
+
+    return false;
+  };
```
pwolfert, line 142: It looks like this is only called in one place, so it could be a plain value instead of a function

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
```diff
@@ -1,137 +1,195 @@
...
-          selected={this.isSelected()}
-          subnavId={this.subnavId}
-          url={this.props.url}
-        />
-        {this.renderSubnav()}
-      </li>
-    );
-  }
-}
+  };
```
pwolfert, line 171: This could also just become a value or be inlined at the bottom, since it is only called once. In the class components, it made sense to organize these into different functions, but now that it's all one function anyway, organizing them into functions isn't really necessary.

pwolfert (commented):
I haven't finished reviewing, and I know you're still working on unit tests, but this is looking good so far!

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
pwolfert: I tested it out in storybook, and I didn't notice a difference between just calling it at the end of `handleToggleClick` and doing it here.

PR #1392, opened 2021-12-14
[WNMGDS-972] Convert MonthPicker to TypeScript

pwolfert:
## Summary
https://jira.cms.gov/browse/WNMGDS-972

### Changed
- Converted `MonthPicker` component to TypeScript and removed old type definitions


zarahzachz:
I thought the base branch was supposed to be the `typescript-conversions` branch?

packages/design-system/src/components/ChoiceList/Choice.tsx
```diff
@@ -76,7 +76,7 @@ export interface ChoiceProps {
    */
   name: string;
   onBlur?: (...args: any[]) => any;
-  onChange?: (...args: any[]) => any;
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => any;
```
zarahzachz, line 79: Does this have a regression risk? Changing the type from `any` to something else more specifically types seems like it might?

zarahzachz (commented):
I didn't see an update to the scss docs pages (I thought they were pulling props from the former JSX files).

Also, not sure if this is something we're supposed to do because it's kinda happening sporadically - but the a11y tests aren't here and I don't know if those are supposed to be converted as well? They are in some components and not others.

packages/design-system/src/components/ChoiceList/Choice.tsx
pwolfert: We're preferring correctness over risk of breaking changes in these conversions. One set of breaking changes at once

pwolfert:
> the a11y tests aren't here and I don't know if those are supposed to be converted as well

We haven't been bothering converting those just yet. Converting them or not should have no effect on what we export

packages/design-system/src/types/index.d.ts
```diff
@@ -28,6 +27,7 @@ export * from './FilterChip';
 export * from './FormControl';
 export * from './FormLabel';
 export * from './InlineError';
+export * from './MonthPicker';
```
scheul93, line 30: Does the `MonthPicker.d.ts` file need to be deleted?

packages/design-system/src/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,217 @@
...
+
+export const MonthPicker = (props: MonthPickerProps) => {
+  const months = getMonthNames(props.locale);
+  const monthsLong = getMonthNames(props.locale, false);
+  const isControlled = props.selectedMonths !== undefined;
+  const [selectedMonthsState, setSelectedMonthsState] = useState(props.defaultSelectedMonths ?? []);
+  const selectedMonths = isControlled ? props.selectedMonths : selectedMonthsState;
+  const disabledMonths = props.disabledMonths ?? [];
+
+  function handleChange(event: ChangeEvent<HTMLInputElement>) {
```
scheul93, line 113: For consistency, consider changing this to arrow function notation

packages/design-system/src/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,217 @@
...
+      if (newSelectedMonths.includes(month)) {
+        newSelectedMonths.splice(newSelectedMonths.indexOf(month), 1);
+      } else {
+        newSelectedMonths.push(month);
+      }
+      setSelectedMonthsState(newSelectedMonths);
+    }
+  }
+
+  function handleSelectAll() {
```
scheul93, line 130: For consistency, consider changing this to arrow function notation

packages/design-system/src/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,217 @@
...
+    if (props.onSelectAll) {
+      props.onSelectAll();
+    }
+
+    if (!isControlled) {
+      setSelectedMonthsState(monthNumbers.filter((m) => !disabledMonths.includes(m)));
+    }
+  }
+
+  function handleClearAll() {
```
scheul93, line 140: For consistency, consider changing this to arrow function notation

scheul93 (approved):
Left a few comments because I'm not sure if we're trying to commit to a consistent use of arrow functions. But everything else seems good.

packages/design-system/src/types/index.d.ts
pwolfert: Good catch!

packages/design-system/src/components/MonthPicker/MonthPicker.tsx
pwolfert: I honestly like the `function` notation better in TypeScript, but if that's how we want to do it, I'm open

pwolfert: I think there can be different room for different notations as long as we're consistent about what we use when, especially since they technically behave differently in JavaScript. To me, there's no functional need to make these arrow functions since there is no need for a `this` object to be referenced in any of them. Functions defined like this also get hoisted, so order doesn't matter. I'd listen to an argument, though, about everything being closed to mutation by using `const` everywhere inside these functions...because technically one can re-defined these functions as I've written them.

zarahzachz: +1 for hoisting powers.

packages/design-system/src/components/MonthPicker/MonthPicker.tsx
```diff
@@ -0,0 +1,217 @@
...
+   * `selectedMonths` property.
+   */
+  defaultSelectedMonths?: number[];
+  /**
+   * A callback function that's invoked when a month's checked state is changed.
+   * Note: This callback is not called when a month is selected or deselected
+   * via the "Select all" or "Clear all" buttons – use the `onSelectAll` and
+   * `onClearAll` event handlers for those instances.
+   */
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => any;
```
zarahzachz, line 92: Just a question - I think I've been setting my event types to `React.MouseEvent | React.KeyboardEvent`. What's the difference between that approach and what you're using? Does my approach have outages in what could be considered an event?

pwolfert: As far as I know, this is the type of the event that gets fired. I'm not really sure where those two events you mentioned would get used—maybe if you're specifically trying to target events from certain input sources? I know that `React.ChangeEvent` has `target` and `currentTarget`, whereas there's a more generic `SyntheticEvent` that is only guaranteed to have `currentTarget` 🤷

PR #1393, opened 2021-12-14
2.12.0-beta.1 Release

forrestbaer:
# Version Bumps for 12/14/21 beta release

 - @cmsgov/design-system-docs@2.12.0-beta.1
 - @cmsgov/design-system-scripts@2.12.0-beta.1
 - @cmsgov/design-system@2.12.0-beta.1
 - @cmsgov/ds-healthcare-gov@6.4.0-beta.1
 - @cmsgov/ds-medicare-gov@3.5.0-beta.1

### DIFF
--- a/packages/ds-medicare-gov/package.json
+++ b/packages/ds-medicare-gov/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/ds-medicare-gov",
-  "version": "3.4.0-beta.2",
+  "version": "3.5.0-beta.1",
   "publishConfig": {
     "tag": "latest",
     "access": "public"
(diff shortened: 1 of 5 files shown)


packages/ds-medicare-gov/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/ds-medicare-gov",
-  "version": "3.4.0-beta.2",
+  "version": "3.5.0-beta.1",
```
pwolfert, line 3: I should have mentioned this first, but for the medicare one, we're just incrementing the beta version number for now until PI-1

forrestbaer: Ok, am I correct in thinking I can just update that here, remove that particular tag then add the new tag? Just want to be sure there isn't anything I'm missing.

pwolfert: Unfortunately I think it would mean redoing all the tags, because the tags should all point to the same commit even though they have different names.

PR #1396, opened 2021-12-14
Wnmgds 981/verticalnav typescript

zarahzachz:
## Summary
[Jira ticket](https://jira.cms.gov/browse/WNMGDS-981)

### Changed
I refactored the `VerticalNav` and related components from JSX to TSX. Also migrated from class to function components.

## How to test
Build this branch locally and compare against the [VerticalNav docs](https://design.cms.gov/components/vertical-nav/).


packages/design-system/src/components/VerticalNav/VerticalNav.tsx
```diff
@@ -0,0 +1,114 @@
...
+  //         {...item}
+  //         component={props.component || item.component}
+  //         _selectedId={props.selectedId}
+  //         key={item.id + item.url + item.label}
+  //         onClick={onClick}
+  //         selected={selected}
+  //       />
+  //     );
+  //   });
+  // };
```
scheul93, line 70: If no longer needed, consider removing

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
```diff
@@ -0,0 +1,212 @@
...
+  const [collapsed, setCollapsed] = useState(props.defaultCollapsed);
+
+  // Refactor from `componentDidUpdate` to `useEffect` using this example:
+  // https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects
+  React.useEffect(() => {
+    if (props.onSubnavToggle) {
+      props.onSubnavToggle(props.id, collapsed);
+    }
+  }, [collapsed]);
+
```
scheul93, line 84: Since `collapsed` is only updated by one line in this component (line 96), consider moving the body of the use effect to directly after the `setCollapsed` call in `handleToggleClick`.

`useEffect` might be necessary if the component doesn't know when `collapsed` might be changing, but here it is extraneous because the component knows when `setCollapsed` is being called

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
```diff
@@ -0,0 +1,212 @@
...
+   * Whether or not the item's sub-navigation is in a collapsed state by default
+   */
+  defaultCollapsed?: boolean;
+  /**
+   * Called when the link is clicked, with the following arguments:
+   * [`SyntheticEvent`](https://facebook.github.io/react/docs/events.html),
+   * `id`, `url`.
+   * This takes precedence over the `VerticalNav` `onLinkClick` prop
+   */
+  onClick?: (...args: any[]) => any;
```
scheul93, line 41: Should the args be typed here?

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
```diff
@@ -0,0 +1,212 @@
...
+  //         selectedId={props._selectedId}
+  //         collapsed={collapsed}
+  //         id={subnavId}
+  //         items={subnavItems()}
+  //         component={props.component}
+  //         nested
+  //       />
+  //     );
+  //   }
+  // };
```
scheul93, line 171: If no longer needed, consider removing

forrestbaer:
lgtm with Annas comments addressed!

packages/design-system/src/components/VerticalNav/VerticalNavItemLabel.test.tsx
```diff
@@ -3,7 +3,7 @@ import VerticalNavItemLabel from './VerticalNavItemLabel';
 import { shallow } from 'enzyme';
 
 function shallowRender(customProps = {}) {
-  const props = Object.assign(
+  const props: any = Object.assign(
```
pwolfert, line 6: I think it would work to type this as `Partial<VerticalNavItemLabelProps>` if `customProps` also had that type

pwolfert (approved):
LGTM 👍  Great work! One idea below for the typing in the unit tests

forrestbaer (approved):

PR #1401, opened 2021-12-15
[WNMGDS-1146] - ChoiceList, Dropdown stories

forrestbaer:
## Summary
Added stories to storybook for ChoiceList (contains Choice as a subcomponent) and Dropdown.

## How to test
Run `yarn storybook` and check new components


packages/design-system/src/components/Dropdown/Dropdown.stories.jsx
```diff
@@ -0,0 +1,108 @@
...
+    </optgroup>
+  </>
+);
+
+const Template = (args) => <Dropdown {...args} />;
+
+export const DefaultDropdown = Template.bind({});
+DefaultDropdown.args = {
+  options: dropdownOptions,
+  label: 'Dropdown Example',
```
zarahzachz, line 67: Not to be a stickler, but this example's using title case when our docs don't ("Dropdown example"). Not sure if this is something we enforce as a design system.

packages/design-system/src/components/Dropdown/Dropdown.stories.jsx
```diff
@@ -0,0 +1,108 @@
...
+  errorMessage: 'Example error message',
+  hint: 'Helpful hint text',
+  label: 'Error example',
+  name: 'error_dropdown_field',
+};
+
+export const Disabled = Template.bind({});
+Disabled.args = {
+  options: dropdownOptions,
+  label: 'Disabled Example',
```
zarahzachz, line 83: Same title case comment here.

packages/design-system/src/components/Dropdown/Dropdown.stories.jsx
```diff
@@ -0,0 +1,108 @@
...
+  name: 'custom_dropdown_field',
+  children: optGroupData,
+};
+
+export const InverseOption = Template.bind({});
+InverseOption.args = {
+  options: dropdownOptions,
+  errorMessage: 'Example error message',
+  hint: 'Helpful hint text',
+  label: 'Inverse Example',
```
zarahzachz, line 102: Same title case comment here.

zarahzachz (commented):
I didn't see examples for different sized Dropdown components here. Is that something we should have?

forrestbaer:
> I didn't see examples for different sized Dropdown components here. Is that something we should have?

I don't think so, to me it adds unnecessary code when it's a radio option in the args area. I wasn't going to do this for Dialog that I'm finishing up now either..   What has been qualifying a new story for me has been cases where the props change enough that they're not settable with a single click or they are significantly different visually.  What do you think?

zarahzachz:
If the goal is to ultimately replace our docs site examples with embedded SB examples... I wonder if there's a way to embed examples with specific props set in the embed only?

forrestbaer:
replying for posterity, Sarah and I chatted more about this, going to be more discussion with team today

packages/design-system/src/components/Dropdown/Dropdown.stories.jsx
```diff
@@ -0,0 +1,108 @@
...
+import Select from './Select';
+
+export default {
+  title: 'Components/Dropdown',
+  component: Dropdown,
+  argTypes: {
+    children: { control: false },
+    options: { control: false },
+  },
+  parameters: {
```
scheul93, line 13: When I originally setup storybook, I didn't realize the backgrounds would be used so frequently. Since this code is repeated in a number of component story files, this config can be elevated to be global parameters.

for example in `preview.js`, this chunk of code would be moved to the parameters property
```
backgrounds: {
    default: 'light',
    values: [
      {
        name: 'light',
        value: '#fff',
      },
      {
        name: 'Hcgov dark',
        value: '#112e51',
      },
      {
        name: 'Mgov dark',
        value: '#146a5d',
      },
    ],
  },
```

scheul93: Not sure if you want to update that here, or if a separate PR is better, but doing this will prevent the copy/paste for every component that has an inverse example

scheul93 (approved):
Actual stories seem good. One question/comment about the background parameters

packages/design-system/src/components/Dropdown/Dropdown.stories.jsx
forrestbaer: Yeah I agree, it's a bad pattern we should squash early.. I can make the changes here for all components using it, or open a new pr with changes for everything.. I thing a new pr would be cleaner in case we ever had to back it out?

forrestbaer: I'm going to write a quick ticket for next sprint and do this for all modules using this pattern next sprint, will ping Krithi with it.

PR #1402, opened 2021-12-15
[WNMGDS-983] Remove base html ruleset variable 

line47:
## Summary
Removing the `$ds-include-base-html-rulesets: true !default;` setting because it's never been set to false and is not needed. This was specifically used for styling an A tag only. 

From the doc site 
> The only base HTML element the design system applies styling to is the `<a>` element. You can also use `.ds-c-link` and `.ds-c-link--inverse` for styling links.

### Removed
- Removed instances where the `ds-include-base-html-rulesets` was set to true
- Removed the instances where the `ds-include-base-html-rulesets` were wrapping an a tag

## How to test
- Run the doc site locally `yarn start` and see if links appear differently than the live site. do the same for Medicare and Healthcare with `yarn start:medicare` and `yarn start:healthcare`

### DIFF
--- a/examples/child-design-system/src/styles/settings/variables/_override.build.scss
+++ b/examples/child-design-system/src/styles/settings/variables/_override.build.scss
@@ -6,8 +6,5 @@
 $font-path: './fonts';
 $image-path: './images';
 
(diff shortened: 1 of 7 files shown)


forrestbaer (approved):
seems to do what it does! 👍

PR #1404, opened 2021-12-16
[WNMGDS-1099] Update alert link styling with more component variables

line47:
## Summary
The alert links are a bit darker than regular text links to ensure they pass color contrast of the various alert backgrounds. Error lightest and our default link color don't pass so we've used a darker shade of blue for the links found in the alerts. 
I've Removed the `%link-darker` placeholder selector to style the link directly. This simplifies the usage and overriding of the alert links. 

### Added
- Variables for the alert link and the various states
- Added variable descriptions to the alert docs page

### Deprecated

### Removed
- Removed the `%link-darker` placeholder selector
- Removed the duplicative code for the Medicare theme

## How to test
Demo http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1099
- See that the proper link color is used - primary-darker and primary-darkest
run `yarn start:medicare` to verify medicare
- See that the proper link color is used - primary-alt-dark, primary-alt-darker, primary-alt-darkest


packages/design-system/src/styles/components/_Alert.scss
```diff
@@ -3,43 +3,11 @@
...
-
-  &:visited {
-    color: $color-visited;
-  }
-
-  &:hover {
-    color: $color-primary-darkest;
-  }
-
-  @if $ds-include-focus-styles {
```
line47, line 21: Focus style blocks removed because we are no longer conditionally showing and hiding focus styles

scheul93 (approved):
Looks okay to me!

PR #1414, opened 2021-12-20
[WNMGDS-1147] Add Storybook stories for DateField and Dialog

forrestbaer:
### Added
Added new `DateField` and `Dialog` component Storybook stories.

Decided to only add one Dialog story as everything is configurable in props easily.  Note it does not automatically close when moving to the docs page, can't be solved for at the moment. 

## Changed

This update also includes a storybook version bump which resolves most of the many warnings in the cosole from the previous version.

## How to test
Run `yarn install` to update storybook dependencies
Run `yarn storybook` and view the new components. Do a small review of other components to ensure nothing else broke with the upgrade.. I've not notices any issues.


packages/design-system/src/components/DateField/DateField.stories.jsx
```diff
@@ -0,0 +1,76 @@
+import React, { useState } from 'react';
+import DateField from './DateField';
+import DateInput from './DateInput';
+
+export default {
+  title: 'Components/DateField',
+  component: DateField,
+  args: {
+    errorMessaage: 'Please enter a year in the past',
```
pwolfert, line 9: I was just tying this out, and there's both an `errorMessaage` (typo), and an `errorMessage` control field. I've tried clearing out both, but the error message is still showing up in the component.

pwolfert (commented):
The Dialog is working smoothly! I think it would also be valuable to have a story where it's just always rendered so people can play around with the different options and see them immediately without having to close it and click the button again.

I also ran into a little issue with the DateField

packages/design-system/src/components/DateField/DateField.stories.jsx
forrestbaer: Ah good catch, thats resolved, if you want to clear it out you have to enter an empty object like ""

Regarding the extra Story, it causes problems to have more than one of these in a story file so I'd need to create another file for it.. On the story page you can open the modal and change props while it's open, just not on the docs page which isn't really meant as a playground.  I think it's ok as is?

forrestbaer: Actually another story wont resolve anything but removing the button to open, it doesn't add any existing functionality or usefulness.

forrestbaer: Oh the issue with the errormessage was fixable by setting an argtype, my bad, fixed..

pwolfert: > On the story page you can open the modal and change props while it's open,

Oh, you're right. This is enough then 👍

pwolfert (approved):
It works! Nice job 🙌

PR #1416, opened 2021-12-20
[WNMGDS-1087] Update focus styles to fix link and remove unneeded conditional focus style SCSS

line47:
## Summary
Updating the focus styles to fix an issue where multi line links were not looking good - The solution was to implement outline instead of box shadow for links only. In addition, went ahead and removed all of the conditional styling for focus styles since we are all in for focus styles now 🎉 

### Added
- Added a new mixin for focus style links
- Added more specific focus styles for HCgov secondary button

### Removed
- `$ds-include-focus-styles` conditional blocks 

### Fixed
- Fixed the multi line link focus style issue 

## How to test
Run each doc site locally to see that the focus styles look as expected and that the links use an outline as compared to everything else which uses a box shadow for the focus styles.


examples/child-design-system/src/styles/settings/variables/_override.build.scss
```diff
@@ -5,6 +5,3 @@
 // Paths to images and fonts relative from the SCSS entry point
 $font-path: './fonts';
 $image-path: './images';
-
-// Set $ds-include-focus-styles to enable the CMSDS focus styles.
-$ds-include-focus-styles: true;
```
line47, line 10: No longer needed

packages/design-system-docs/src/pages/startup/theming.md
```diff
@@ -45,9 +45,6 @@ You can also [browse all Sass variable files on GitHub](https://github.com/CMSgo
 
 #### Focus styles
 
-The focus styles are enabled by default. However, if you'd like to disable them,
-set the `$ds-include-focus-styles` variable to `false` in your `_override.build.scss` file.
-
```
line47, line 50: Updates to documentation removing the ability to disable the focus styles

packages/design-system/src/styles/base/_link.scss
```diff
@@ -77,16 +69,10 @@ $link-underline-offset: auto !default;
...
-    }
-  } @else {
-    &:focus {
-      color: $link-inverse-hover-color;
-      text-decoration-color: $link-inverse-hover-color;
-    }
+  &:focus,
+  &:focus:visited {
+    @include focus-styles-link;
+    color: $color-base;
```
line47, line 75: Can't set the link color to white because the background is white so setting it to `color-base`

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -22,27 +22,26 @@
   text-align: center;
   text-decoration: none;
 
+  /* stylelint-disable selector-max-specificity */
+
   &:hover,
-  &.ds-c-button--hover {
+  &.ds-c-button--hover,
+  &:hover &:focus,
+  &.ds-c-button--hover &.ds-c-button--focus {
```
line47, line 30: Needed more specificity here

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -147,26 +143,20 @@
...
   &:focus,
   &.ds-c-button--focus {
-    @if $ds-include-focus-styles {
-      @include focus-styles;
-    } @else {
-      border-color: rgba($border-color-inverse, 0.8);
-      color: rgba($color-base-inverse, 0.8);
-    }
+    @include focus-styles;
+    background-color: $color-background-inverse;
```
line47, line 147: Needed to set the background color for inverse buttons to use the same as the default state of inverse buttons

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -259,31 +247,22 @@ $button-color-variants: (
...
-      }
-    } @else {
-      &:focus,
-      &.ds-c-button--focus {
-        background-color: nth($color, 2);
-      }
+    &:focus,
+    &.ds-c-button--focus {
+      @include focus-styles;
+      background-color: nth($color, 1);
```
line47, line 253: Needed to set the background color for focus to the same background color used for default states

packages/ds-healthcare-gov/src/styles/components/_Button.scss
```diff
@@ -15,7 +15,10 @@
     border-color: transparent;
     color: $color-white;
   }
-
+  &:focus,
+  &.ds-c-button--focus {
+    background-color: $bg-color;
+  }
```
line47, line 21: Setting the bg color for Healthcare secondary button focus

packages/design-system/src/styles/base/_link.scss
pwolfert: Why not `$link-inverse-visited-color`?

line47: Hmmm, `$link-inverse-visited-color` uses `$color-muted-inverse` which is a really light grey and wouldn't pass contract on a white background - We add white background for the focus to ensure the link is visible on dark backgrounds in addition to the outline

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -51,13 +50,11 @@
...
-    &.ds-c-button--active:focus {
-      background-color: $color-white;
-      box-shadow: none;
-    }
+  // Make sure active overrides focus styles
+  &:active:focus,
+  &.ds-c-button--active:focus {
+    background-color: $color-white;
+    box-shadow: none;
   }
```
pwolfert, line 58: Do we intentionally not show the focus ring in the active state?

packages/design-system/src/styles/components/_Button.scss
pwolfert: Can you talk about this more? I'm having trouble understanding what state this applies to

packages/ds-medicare-gov/src/styles/settings/_index.scss
```diff
@@ -1,5 +1,4 @@
 // Overrides and variable re-declarations that need to be in place *before* loading the core design system
-@import 'variables.build';
```
pwolfert, line 2: 💥

packages/design-system/src/styles/components/_Button.scss
line47: Yep that's correct 👍  I'm sure there is a closed PR somewhere explaining why but ATM I can't remember the reasoning

packages/design-system/src/styles/base/_link.scss
pwolfert: oh, I actually meant to copy and paste the non-inverse color variable

pwolfert: `$link-visited-color`

packages/design-system/src/styles/components/_Button.scss
line47: This applies to an edge case where an a tag is used with button classes to be styles like a button. We want the background color for focus to be the same color as in the default state. 

For links the default focus background color on hover is white. We don't want that behavior with buttons.

line47: Primary button `<a>` tag with no background set 
Screen Shot 2021-12-20 at 5 14 08 PM

Primary button `<a>` tag with background set 
Screen Shot 2021-12-20 at 5 15 02 PM

packages/design-system/src/styles/base/_link.scss
line47: I updated to use `link-color` - this is only for focused elements so we wouldn't want to use the visited color here

packages/design-system/src/styles/components/_Button.scss
pwolfert: Thank you, that makes sense to me 👍

pwolfert (approved):
I think I've tried out all the scenarios for each of the design systems. LGTM

PR #1419, opened 2021-12-21
[WNMGDS-1279] Fix definition of window.utag conflicting with consuming products

pwolfert:
## Summary
https://jira.cms.gov/browse/WNMGDS-1279

Also, https://github.com/CMSgov/design-system/pull/1375/commits/129ad9a45822cf75a4e97fadacc748e7bd5eeef9 caused another problem that I discovered while fixing this. The `translate` HOC function changed its name to `withTranslation`.

### Fixed
- Fixed type-definition for `window.utag` in the design system and consuming applications by no longer applying a global type definition for it in our library
- Fixed bad usage of the `react-i18next` library in `ds-healthcare-gov` after [this major version bump](https://github.com/CMSgov/design-system/pull/1375/commits/129ad9a45822cf75a4e97fadacc748e7bd5eeef9)

## How to test
1. Check out `d106cca294bed3dac463f10514ce67486047c921`
2. `yarn install && yarn build`
2. Go to `examples/create-react-app-typescript`
3. Clear `node_modules`
4. `yarn install && yarn start`; should compile without problems


forrestbaer (approved):
all works and looks good!

pwolfert:
Not sure why, but the unit tests think `withTranslation` doesn't exist, even though our yarn lockfile says to use `react-i18next` v11, and in my local node_modules `withTranslation` is definitely an exported higher-order-component

pwolfert:
Oh, it's because we mock `react-i18next` in the unit tests

PR #1424, opened 2021-12-21
[WNMGDS-821] Move lists to typography page

line47:
## Summary
Moved list html examples and guidance from it's own component page to the typography page

### Added
- Added list information to the typography page

### Removed
- Removed the list page from the components section 

## How to test
Check out the demo site and see the following
- Lists is no longer listed as a component
- List info can be found under Styles > Typography
http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-821


packages/design-system-docs/src/pages/styles/typography/_typography.docs.scss
```diff
@@ -26,6 +26,37 @@ Markup: typography-body.example.html
...
+List
+
+Lists organize written information for users.
+
+Markup: list.example.html
+
+### When to use
+
+- Use an ordered list when you need to display text in some ranking, hierarchy, or series of steps.
+- Use unordered lists to display text in no specific order.
```
zarahzachz, line 39: Curious - do we have guidance on when to use an unstyled list?

line47: No we currently do not

packages/design-system-docs/src/pages/styles/typography/_typography.docs.scss
```diff
@@ -26,6 +26,37 @@ Markup: typography-body.example.html
...
+
+- If you need to communicate long lists of narrative text.
+
+### Usage
+
+- Use sentence case and begin lists with a capital letter.
+- Use punctuation appropriate to the text. Do not leave sentences without periods.
+
+### Accessibility
+- When a list has a heading, consider adding an `id` attribute to the heading element, and an `aria-labelledby` attribute to the list element. Screen readers will then announce the heading text when users navigate to the list.
```
ericwbailey, line 51:
```suggestion
### Accessibility

- When a list has a heading, consider adding an `id` attribute to the heading element, and an `aria-labelledby` attribute to the list element. Screen readers will then announce the heading text when users navigate to the list.
```

ericwbailey, line 51: This advice is new to me. I'm not necessarily debating it, but I'm curious about its origin.

ericwbailey (changes requested):
One little suggested change for consistency, and a non-blocking question.

packages/design-system-docs/src/pages/styles/typography/_typography.docs.scss
line47: I'm unsure of the origin of this - there might be some info in the depths of closed PRs

ericwbailey (approved):

PR #1434, opened 2021-12-22
[NO-TICKET] Remove extra yarn.lock file that I must have missed before

pwolfert:
## Summary
This file shouldn't exist because it's redundant. With yarn workspaces, there's only supposed to be one root lockfile. [Dependabot created this PR to update this lockfile](https://github.com/CMSgov/design-system/pull/1429), but the lockfile shouldn't exist.

### Removed
- Removed redundant lockfile in one of our packages


scheul93 (approved):
Be gone, yarn.lock!

PR #1435, opened 2021-12-22
[WNMGDS-1121] usabanner refactor

scheul93:
## Summary
Refactoring the USA Banner component because it's easier to swap the icons with updated markup than to try to shoehorn this work into the existing solution

### Changed
Updated the markup for USA Banner. The previous markup was doing a bunch of CSS trickery for mobile vs larger screen views. I think this is because on mobile, the whole "header" become a clickable element, whereas on larger screens, it is just the button-that-looks-like-a-link text

Note: upon making this PR, I think the biggest risk here is that anyone not using react will have to implement their own mobile vs larger screen view. The original implementation kinda solved for that by having complex markup and CSS. Not sure if that's a concern

## How to test

Checkout storybook demo site here: http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1121/usabanner-refactor/core/storybook/?path=/story/components-usa-banner--english-banner
Compare to live site here: https://design.cms.gov/components/usa-banner/


packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -30,14 +32,52 @@ $usa-banner-font-family: $font-sans;
...
+
 .ds-c-usa-banner__header-flag {
   height: 11px;
   margin-right: $spacer-1;
   width: $spacer-2;
 }
 
 .ds-c-usa-banner__header-text {
-  display: table;
+  display: inline;
```
zarahzachz, line 80: It's curious this layout worked before using `table` - I was under the impression `table` had a block-like behavior instead of inline.

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -10,14 +17,89 @@ export class UsaBanner extends React.PureComponent {
...
+  // on mobile, the entire header needs to be a clickable element
+  renderMobileHeaderContent(t) {
+    return (
+      <button
+        onClick={this.handleToggleBanner}
+        className="ds-c-usa-banner__button"
+        aria-expanded={this.state.isBannerOpen}
+        aria-controls={this.id}
+      >
+        <p className="ds-c-usa-banner__header-text">
```
zarahzachz, line 48: Why does this icon need to be wrapped in a `<p>` with this class? Seems like it could work just as well without the wrapping element?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -10,14 +17,89 @@ export class UsaBanner extends React.PureComponent {
...
+        )}
+      </button>
+    );
+  }
+
+  // on larger screens, only cta needs to be clickable
+  renderHeaderContent(t) {
+    return (
+      <>
+        <p className="ds-c-usa-banner__header-text">
```
zarahzachz, line 76: Again, not sure this `<p>` is needed when a class could be applied to the SVG itself

zarahzachz (commented):
This gets the job done - just unfortunate there's not a lot of crossover between the two viewports. I wonder if this component could be refactored design-/code-wise in the future to remove the necessity for the 2 sets of code?

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
```diff
@@ -10,14 +17,89 @@ export class UsaBanner extends React.PureComponent {
...
-    this.state = { isBannerOpen: false };
+    this.media = window.matchMedia('(max-width: 543px)');
+    this.state = {
+      isBannerOpen: false,
+      shouldRenderMobileView: this.media.matches,
+    };
     this.handleToggleBanner = this.handleToggleBanner.bind(this);
+    this.onMediaChange = this.onMediaChange.bind(this);
+
+    this.media.addEventListener('change', this.onMediaChange);
```
pwolfert, line 28: Does this need to be cleaned up when the component goes away? I guess since this is in the header, that shouldn't really happen, should it?

scheul93: I'll add a cleanup function in here for safety! I think generally we'd be okay since the component is usually not mounting and un-mounting on a page, but also, it feels short sited to assume that.

packages/design-system/src/components/UsaBanner/UsaBanner.jsx
scheul93: Good question! I think this is left over from some copy/paste + the original markup, but while I'm refactoring, might as well clean this up too.

forrestbaer: looks like that class is used below as well to inline these items and set margins.. maybe `<span>` would be better.

scheul93: Looking into this and I think I can remove the `<p>` here altogether and I don't need to add the class to the SVG.

scheul93:
I think I addressed all feedback. I had to move the match media to the `componentDidMount` method because when running a11y tests, the doc site is being built in node and a `window is not defined` error kept popping up. I'm open to alternative ideas if anyone has them.

I'm also realizing that this work is going to create merge conflicts with the component conversion to Typescript 😬

pwolfert (approved):
This LGTM as it is now, but, like you said, it will conflict with the hooks implementation in the new TypeScript version

PR #1445, opened 2021-12-28
[WNMGDS-1272] Configure project for type-checking and fix errors

pwolfert:
## Summary
Overview of changes

### Added
- Added a `yarn type-check` command

### Changed
- TypeScript now checks our src files and ignores the manually-defined definition files found in `packages/**/src/types/`
- ESLint rule disabled that disallows the `@ts-ignore` directive. In certain cases it's not worth the time cost to solve every compiler error and have perfect type-checking. To see my attempt at fixing this error, look at https://github.com/CMSgov/design-system/commit/b5f986571b90338e5ca80d0fb79beb8d7efd8f34

### Fixed
- Fix errors in unit tests about `window.utag`
- Fix errors in Button tests by cleaning up Button prop types
- Fix errors in DateInput tests and remove an obsolete test
- Fix errors in maskHelpers tests by making optional function parameters optional
- Fix errors in TextInput tests by cleaning up TextInput prop types

## How to test
The first three commits configure TypeScript for true type-checking of our source code, and each subsequent commit fixes a set of TypeScript errors. If you want to see these errors, check out https://github.com/CMSgov/design-system/commit/094a72a3064de418e2ed6783f0aca9048eb4ebfa and then run `yarn type-check`. To see them fixed, check out the branch head again and run the same command.

### DIFF
--- a/packages/design-system/src/components/List/List.a11y.test.tsx
+++ b/packages/design-system/src/components/List/List.a11y.test.tsx
@@ -3,7 +3,7 @@ import { ROOT_URL } from '@cmsgov/design-system-scripts/helpers/a11y/constants';
 
 import assertNoAxeViolations from '@cmsgov/design-system-scripts/helpers/a11y/assertNoAxeViolations';
 
-const rootURL = `${ROOT_URL}/example/components.list/`;
+const rootURL = `${ROOT_URL}/example/styles.typography.list/`;
 
 describe('List component', () => {
   it('Should have no accessibility violations', async () => {
--- a/packages/design-system/src/components/TextField/TextInput.tsx
+++ b/packages/design-system/src/components/TextField/TextInput.tsx
@@ -9,7 +9,7 @@ export type TextInputSize = 'small' | 'medium';
 export type TextInputValue = string | number;
 export type TextInputErrorPlacement = 'top' | 'bottom';
 
-export interface TextInputProps {
+export interface CommonTextInputProps {
   /**
    * Apply an `aria-label` to the text field to provide additional
    * context to assistive devices.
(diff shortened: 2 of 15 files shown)


forrestbaer:
I see you didn't modify List but I am getting an accessibility error in List.a11y during `yarn test` in this branch.

pwolfert:
Looks like this error and others exist on the `typescript-conversions` branch. This PR cleans up most of them. I'll try to fix this one here too.

pwolfert:
Looks like it's because of https://github.com/CMSgov/design-system/commit/8dfdc3bb41f5fc259e2087e188b1668761c67365

packages/design-system/src/components/List/List.a11y.test.tsx
```diff
@@ -3,7 +3,7 @@ import { ROOT_URL } from '@cmsgov/design-system-scripts/helpers/a11y/constants';
 
 import assertNoAxeViolations from '@cmsgov/design-system-scripts/helpers/a11y/assertNoAxeViolations';
 
-const rootURL = `${ROOT_URL}/example/components.list/`;
+const rootURL = `${ROOT_URL}/example/styles.typography.list/`;
```
zarahzachz, line 6: Since List is a class applied to a native list element, do we even need to test the a11y for it? We don't test for other Typography elements.

packages/design-system/src/components/TextField/TextInput.tsx
```diff
@@ -51,14 +56,18 @@ export interface TextInputProps {
...
+  multiline?: MultilineValue;
   name?: string;
   /**
    * Sets `inputMode`, `type`, and `pattern` to improve accessiblity and consistency for number fields. Use this prop instead of `type="number"`, see [here](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/) for more information.
    */
   numeric?: boolean;
-  onBlur?: (...args: any[]) => any;
-  onChange?: (...args: any[]) => any;
+  onBlur?: (
+    e: React.FocusEvent<MultilineValue extends true ? HTMLTextAreaElement : HTMLInputElement>
```
zarahzachz, line 66: Didn't realize you could use ternaries in a <Type>!

zarahzachz (approved):
Looks good - code looks so clean!

packages/design-system/src/components/List/List.a11y.test.tsx
pwolfert: That's a fair point. It might be good to have a check to make sure our example is accessible so we're not encouraging bad practices. I'm going to leave it for now because that decision is probably out of scope for this PR.

packages/design-system/src/components/TextField/TextInput.tsx
pwolfert: I think TypeScript type annotations are Turing complete at this point 😆. Honestly it feels kind of icky, and there are probably several ways we could go about defining these types. I was spending too much time on this, though. We might want to go back at some point in the future and do another cleanup pass on the types (as long as the API doesn't change).

PR #1451, opened 2021-12-30
[WNMGDS-979] - Tooltip/TooltipIcon Typescript conversion

forrestbaer:
## Summary
Updated Tooltip/Tooltipicon components to functional TS components

## How to test
Run `yarn storybook' 'yarn build && yarn start' for each package, static builds as well. Ensure Tooltip components are functional.


packages/design-system/src/components/Tooltip/__snapshots__/Tooltip.test.jsx.snap
```diff
@@ -7,7 +7,7 @@ exports[`Tooltip renders custom trigger component 1`] = `
   onMouseLeave={[Function]}
 >
   <a
-    aria-describedby="trigger_3"
+    aria-describedby={null}
```
pwolfert, line 10: Were these ids supposed to disappear?

forrestbaer: Seems like our snapshot tool might need some lessons, I can revert these I think

forrestbaer: Actually, modifying these directly seems like a bad idea? Maybe another PR for jest typescript config work?

forrestbaer: Maybe get into something like : https://huafu.github.io/ts-jest/

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,337 @@
...
+    setTimeout(() => {
+      const focusedInsideTrigger = triggerElement && triggerElement.contains(event.target as Node);
+      const focusedInsideTooltip = tooltipElement && tooltipElement.contains(event.target as Node);
+      if (!focusedInsideTrigger && !focusedInsideTooltip && !isHover) {
+        setActive(false);
+      }
+    }, 10);
+  };
+
+  const handleTouch = (): void => {
```
pwolfert, line 129: I don't think these `: void` annotations are necessary

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,337 @@
...
+import 'core-js/features/promise';
+// TODO: Update react-transition-group once we update react peer dep
+import CSSTransition from 'react-transition-group/CSSTransition';
+import FocusTrap from 'focus-trap-react';
+import React, { useState, useRef, useEffect, useLayoutEffect } from 'react';
+import classNames from 'classnames';
+import { createPopper } from '@popperjs/core';
+import uniqueId from 'lodash/uniqueId';
+
+export interface ITooltipProps {
```
pwolfert, line 16: We haven't been using the `I{ComponentName}` pattern anywhere else

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,337 @@
...
+    const tooltipStyle = { maxWidth, zIndex };
+    const interactiveBorderStyle: React.CSSProperties = {
+      left: `-${interactiveBorder}px`,
+      top: `-${interactiveBorder}px`,
+      border: `${interactiveBorder}px solid transparent`,
+      zIndex: -999, // ensures interactive border doesnt cover tooltip content
+    };
+    const eventHandlers = dialog ? {} : { onBlur: (event) => handleBlur(event) };
+
+    const tooltipContent = () => (
```
pwolfert, line 263: This could be just a value instead of a function.

pwolfert: The parent function `renderContent` I kind of feel the same way about, but there's probably enough that has changed from the original already.

packages/design-system/src/components/Tooltip/TooltipIcon.tsx
```diff
@@ -1,9 +1,15 @@
 import { InfoCircleIconThin } from '../Icons';
-import PropTypes from 'prop-types';
 import React from 'react';
 import classNames from 'classnames';
 
-export const TooltipIcon = (props) => {
+export interface ITooltipIcon {
```
pwolfert, line 5: Same here. We should use `TooltipIconProps` for consistency

pwolfert (commented):
From manually testing this, it seems to behave just like the old one 🙌. Nice work! I do have some comments in the code about consistency and simplification.

packages/design-system/src/components/Tooltip/__snapshots__/Tooltip.test.jsx.snap
pwolfert: I was calling it out because I think it's actually a regression. I'm pretty sure these should have ids but they don't in the snapshot. The snapshot changes here are symptoms of another problem as far as I can tell

forrestbaer: Ah gotcha!

packages/design-system/src/components/Tooltip/Tooltip.tsx
forrestbaer: Agreed, I had actually flattened the whole thing out at one point, since this is re-used I figured I would leave it instead of wrapping the focus element in conditionals..

forrestbaer:
All set with those changes, thanks @pwolfert !

packages/design-system/src/components/Tooltip/__snapshots__/Tooltip.test.jsx.snap
pwolfert: I think it's null on first render because it's being set in a `useEffect` function, which runs after the first render technically https://github.com/CMSgov/design-system/pull/1451/files#diff-c9c653706166a0e772cbd7bceda432d6217948044581f3f43e489a388a3c8720R137

I tried moving the initialization value to the useRef call like
```ts
const id = useRef(props.id ?? uniqueId('trigger_'));
```
and that seemed to work in Storybook and also fix the unit test snapshots

pwolfert (approved):
Made a comment about my findings on the snapshot failures. Other than that, this LGTM 👍

packages/design-system/src/components/Tooltip/__snapshots__/Tooltip.test.jsx.snap
forrestbaer: excellent! thanks for that fix.. updated and tested, working now

PR #1452, opened 2021-12-30
[NO-TICKET] Upgrade TypeScript to 4.5.4

pwolfert:
## Summary
This newer version of TypeScript has nice features. Unfortunately some changes to the compiler forces me to relocate the `locales` directory. We could possibly revisit this later when we upgrade react-i18next.

### Added

### Changed
- Upgrade TypeScript from `3.8.3` to `4.5.4`
- Move `locale` folder into `components` so `components` can be the root of the compiled TypeScript files because the new TS compiler complains that `locale` and the json files in it are outside of the root directory. We want the `components` folder to be the root because its contents get compiled into `dist/components`, `dist/esnext`, and `dist/types`

## How to test
Make sure all build targets work and storybook works and `yarn type-check` works


forrestbaer (approved):
All builds, static and otherwise ran successfully, storybook and type checking with no problems. Tests all passed 👍

PR #1453, opened 2021-12-30
[WNMGDS-976] Convert StepList component to TypeScript

pwolfert:
## Summary
https://jira.cms.gov/browse/WNMGDS-976

### Changed
- Converted StepList and sub-components to TypeScript


packages/design-system/src/components/StepList/Step.tsx
```diff
@@ -23,7 +53,7 @@ export const Step = ({ step, ...props }) => {
     const label = isValidTemplate ? text.replace('%{step}', step.heading || step.title) : undefined;
     return { 'aria-label': label };
   };
-  const Heading = `h${step.headingLevel || '2'}`;
+  const Heading = `h${step.headingLevel || '2'}` as const;
```
pwolfert, line 56: This enables TypeScript to interpret the type of this as `h${1 | 2 | 3 | 4 | 5}` rather than as a generic string, which then in turn allows it to know what props to expect so it doesn't complain later when we try to use it as `<Heading...>`

Screen Shot 2021-12-29 at 1 58 30 PM

scheul93:
Please update the file `packages/design-system/src/types/index.d.ts` to `export * from './StepList';` and remove all StepList-related type definition files in the `types` directory

pwolfert:
@scheul93, I've updated this according to your feedback

scheul93 (approved):
Looks good

PR #1454, opened 2021-12-31
[NO-TICKET] Upgrade react-i18next again

pwolfert:
## Summary
Over the course of several TypeScript config fixes, I think all of the problems causing us to revert the upgrade in https://github.com/CMSgov/design-system/pull/1405 are now resolved. Commits that I think collectively fixed the issue:
- https://github.com/CMSgov/design-system/commit/85f8f866f869ea3476a217efde0e6323267912be
- https://github.com/CMSgov/design-system/commit/7bf9c52d5e3d434f58817d0680aa2bb9685b205d
- https://github.com/CMSgov/design-system/commit/5171b7ecfa09956e8dd09196245d8998c5658084

### Changed
- Upgraded `react-i18next` from `9.0.10` to `11.14.3`, which fixes TypeScript errors that pop up when we start trying to convert modules to TypeScript that use `react-i18next`

## How to test
Install dependencies and then try all the build targets, including build, build-docs, start, storybook, and both child design systems. Also try `yarn type-check`.

### DIFF
--- a/packages/ds-healthcare-gov/package.json
+++ b/packages/ds-healthcare-gov/package.json
@@ -26,7 +26,7 @@
     "classnames": "^2.0.0",
     "i18next": "^21.5.5",
     "js-cookie": "^2.2.0",
-    "react-i18next": "^9.0.10"
+    "react-i18next": "^11.14.3"
   },
   "devDependencies": {
     "@babel/core": "^7.0.0",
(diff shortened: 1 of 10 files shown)


forrestbaer (approved):
all tests ran and passed, builds, storybook, type-check, lgtm

PR #1462, opened 2022-01-03
[WNMGDS-969] Typescript Dialog Conversion

forrestbaer:
## Summary
Converted the Dialog component to Typescript

## How to test
`yarn test`, `yarn build && yarn start` each system to confirm docs site, `yarn storybook` to confirm storybook & `yarn type-check' to confirm no ts typing issues.


packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -1,49 +1,170 @@
 import { EVENT_CATEGORY, MAX_LENGTH, sendLinkEvent } from '../analytics/SendAnalytics';
 import AriaModal from 'react-aria-modal';
```
pwolfert, line 2: I can't find the type definitions for this package. I'm not sure how there are no type definitions and TypeScript isn't complaining about it.

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -1,49 +1,170 @@
...
+  title?: React.ReactNode;
+  /**
+   * Enable exiting the dialog when a user clicks the underlay.
+   */
+  underlayClickExits?: boolean;
+  /**
+   * Allow additional AriaModal props to be passed to Dialog
+   */
+  // eslint-disable-next-line camelcase
+  [additional_props: string]: unknown;
```
pwolfert, line 129: The reason I want to look at the type definitions for `react-aria-modal` is that I wonder if we can extend the props for that component here rather than having this generic string index with an unknown type.

forrestbaer: Gotcha, those are in `node_modules/react-aria-modal/src/react-aria-modal.js`

`
     7  static defaultProps = {
     8     underlayProps: {},
     9     dialogId: 'react-aria-modal-dialog',
    10     underlayClickExits: true,
    11     escapeExits: true,
    12     underlayColor: 'rgba(0,0,0,0.5)',
    13     includeDefaultStyles: true,
    14     focusTrapPaused: false,
    15     scrollDisabled: true
    16   };
`

plus the others listed there, `titleText`, `titleId`, `getApplicationNode`, `underlayProps`, `dialogStyle`

forrestbaer: No static definitions but yes I could create an interface for those

forrestbaer: There are some types here: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-aria-modal/index.d.ts we could include I suppose? Though they aren't a part of the core package it adds a dependency..

pwolfert: What I don't understand is how TypeScript is even allowing us to use this without type definitions. Normally it yells at us. If we add `@types/react-aria-modal` like you linked to, we could extend your props with `AriaModalProps`, and that would give us everything we need.

forrestbaer: That library contains a ton of definitions I don't think we want to mess with, I've extracted the type from the set and extended it in the dialog component here.. I actually tried to install that library and ran into a bunch of version issues with yarn it wouldn't let me do it without a struggle.. Also if you look at that libraries issues list it's like 500+ long, a little spooky.. let me know what you think

pwolfert: Yikes! In that case TypeScript mysteriously ignoring it is a blessing. 👀 I guess we should just ignore it too. Recreating the props type definition sounds good to me 👍

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,436 @@
...
+          ga_eventCategory: EVENT_CATEGORY.uiComponents,
+          ga_eventLabel: eventHeadingText,
+          heading: eventHeadingText,
+        });
+      }
+    }
+  },[])
+
+
+    const {
```
pwolfert, line 342: Did this get ignored by Prettier? It looks like Prettier hasn't formatted this file

pwolfert (approved):
It is working! Looks good to me except for a formatting issue.

Tested analytics code in Storybook by setting
```js
setDialogSendsAnalytics(true);
window.utag = { link: console.log };
```
and setting `analytics` to true

packages/design-system/src/components/Dialog/Dialog.tsx
forrestbaer: Interesting, worked this time, i've been playing around with a git tool on the command line, must have gotten around it.. i've updated going through the normal linter/prettier

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -0,0 +1,436 @@
...
+      );
+    }
+    if (props.closeText) {
+      console.warn(
+        `[Deprecated]: Please remove the 'closeText' prop in <Dialog>, use 'closeButtonText' instead. This prop has been renamed and will be removed in a future release.`
+      );
+    }
+  }
+
+  useEffect(() => {
```
zarahzachz, line 298: This is something I'm noticing on my HelpDrawer conversion... but refactoring the lifecycle methods used for analytics with `useEffect` breaks the analytics unit tests. I don't understand why this is the case though and can only get my own tests to pass by maintaining the class component structure.

I wonder if it has something to do with a class property in the constructor behaving different from a func let? It's fuzzy to me.

forrestbaer: the analytics issue is a race condition in the test for the analytics being ready in the event queue, adding a `setTimeout(() => { test },0)` sorts it out, updated the test, works now!

packages/design-system/src/types/index.d.ts
```diff
@@ -16,6 +15,7 @@ export * from './Button';
 export * from './ChoiceList';
 export * from './DateField';
 export * from './Dropdown';
+export * from './Dialog';
```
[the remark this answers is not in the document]
forrestbaer: you are permitted to nitpick :)

zarahzachz (approved):

PR #1448, opened 2021-12-29
[WNMGDS-1273] Convert Healthcare Footer to TypeScript

zarahzachz:
## Summary
https://jira.cms.gov/browse/WNMGDS-1273

### Changed
- Convert HC.gov Footer and all subcomponents to TypeScript

### Fixed
- Update the i18next-related code to be compatible with its new API
  - Unfortunately this means perhaps misusing its namespaces feature in order for us to be able to keep our own API unchanged. I have made a note about it in the code.

## How to test
View the storybook story for the footer to test it. Make sure switching between English and Spanish works.

PR #1467, opened 2022-01-04
[WNMGDS-1274] Convert Healthcare Header to TypeScript

pwolfert:
## Summary
https://jira.cms.gov/browse/WNMGDS-1274

### Changed
- Convert HC.gov Header and all subcomponents to TypeScript

### Fixed
- Update the i18next-related code to be compatible with its new API (See https://github.com/CMSgov/design-system/pull/1448)

## How to test
View the storybook story for the header to test it. When switching between English and Spanish in the story controls, the changes won't take until you refresh.


packages/ds-healthcare-gov/src/components/Header/ActionMenu.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * ActionMenu is displayed at the top-right of the header.
+ * There are two variations of this component: One for logged-out
+ * users, which displays a "Log in" & "Espanol" link on desktop.
+ * And another for logged-in users, which includes the user's first
+ * name. The logged-in variation always displays a "Menu" toggle
+ * button, and the logged-out variation only displays it on mobile.
+ */
+const ActionMenu = function (props) {
```
zarahzachz, line 44: Did these props not need to have the type `ActionMenu`?

zarahzachz (commented):
I tried building with Storybook and received an error:

`Cannot read properties of undefined (reading 'variation')`

Also ran `yarn type-check` and `yarn test` and got a few errors there as well...

packages/ds-healthcare-gov/src/components/Header/ActionMenu.tsx
```diff
@@ -0,0 +1,116 @@
...
+
+const menuId = 'hc-c-menu';
+
+export interface LinkProps {
+  href: string;
+  label: React.ReactNode;
+  onClick?: (event: SyntheticEvent) => any;
+}
+
+export interface ActionMenu {
```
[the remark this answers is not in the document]
pwolfert: Yes it does, _and_ we're not even using it!

packages/ds-healthcare-gov/src/components/Header/ActionMenu.tsx
pwolfert: Yup

pwolfert: Hooking this up properly spawned a few more TS errors that I'm now working through 👍

packages/ds-healthcare-gov/src/components/Header/Header.tsx
```diff
@@ -2,189 +2,65 @@ import ActionMenu from './ActionMenu';
...
-          submenuTop={this.props.submenuTop}
-          submenuBottom={this.props.submenuBottom}
-        />
-
-        {this.props.deConsumer && <DeConsumerMessage deBrokerName={this.props.deBrokerName} />}
-        {this.props.headerBottom}
-      </header>
-    );
-  }
+interface Link {
```
scheul93, line 13: [question] would it be helpful to export this interface since it is the type for the `links` prop?

packages/ds-healthcare-gov/src/components/Header/Menu.tsx
```diff
@@ -1,10 +1,29 @@
...
 import classnames from 'classnames';
 
-const Menu = function (props) {
+interface MenuProps {
+  links?: {
+    identifier?: string;
+    href: string;
+    label: React.ReactNode;
+    onClick?: (...args: any[]) => any;
+  }[];
```
scheul93, line 12: There's probably a little bit of reuse you could get here from the Links interface in Header, but you'd also have to extend it, so might not be worth

packages/ds-healthcare-gov/src/components/Header/MenuLinks.tsx
```diff
@@ -0,0 +1,51 @@
+import { LINK_IDENTIFIERS } from './defaultMenuLinks';
+import React from 'react';
+import { sendHeaderEvent } from './analytics';
+
+interface MenuLinksProps {
+  links: {
+    identifier?: string;
+    href: string;
+    label: React.ReactNode;
+    onClick?: (...args: any[]) => any;
```
scheul93, line 10: Same comment from Menu. There's probably a type that could be defined for reuse, but not sure if that's easy or worth right now

packages/ds-healthcare-gov/src/components/Header/defaultMenuLinks.ts
```diff
@@ -1,41 +1,32 @@
...
- * @returns {Object}
  */
 export function defaultMenuLinks(
+  t: TFunction,
   locale = 'en',
-  deConsumer,
-  subpath,
+  deConsumer?: boolean,
+  subpath?: string,
   primaryDomain = '',
```
scheul93, line 21: should primaryDomain have a `?` here? It seems odd that a required param is in the middle of a bunch of optional ones.

Maybe it's okay because it's defaulted, but seemed a little odd

packages/ds-healthcare-gov/src/types/Header/index.d.ts
```diff
@@ -10,33 +10,6 @@ interface Link {
...
-  submenuTop?: React.ReactNode;
-  submenuBottom?: React.ReactNode;
-  headerBottom?: React.ReactNode;
-}
-
-export class Header extends React.Component<HeaderProps, any> {
-  render(): JSX.Element;
-}
-
-export default Header;
```
scheul93, line 38: Does this whole file still need to exist? If yes, does the `defaultMenuLinks` definition need to be updated to match the updated params in defaultMenuLinks.ts?

packages/ds-healthcare-gov/src/components/Header/Header.tsx
scheul93: Or maybe this is covered by the Header/index.d.ts file?

pwolfert: I'll see what I can do

packages/ds-healthcare-gov/src/types/Header/index.d.ts
pwolfert: Nope, it should be deleted

packages/ds-healthcare-gov/src/components/Header/defaultMenuLinks.ts
pwolfert: TypeScript won't allow a question mark if there's an intiailizer, but I could be explicitly defining the type as string

scheul93 (approved):

PR #1471, opened 2022-01-05
[WNMGDS-1121] usabanner refactor ts

scheul93:
## Summary
Applying changes from PR from `master` to this branch: https://github.com/CMSgov/design-system/pull/1435

This will hopefully make merge conflicts easier


scheul93:
`packages/design-system/src/components/UsaBanner/UsaBanner.tsx` is the only code that's new. All others are direct copy/paste from original changes

packages/design-system/src/components/UsaBanner/UsaBanner.tsx
```diff
@@ -1,8 +1,16 @@
-import React, { useState } from 'react';
-import { Button } from '../Button';
+/* eslint-disable react/no-multi-comp */
+/* eslint-disable react/prop-types */
```
scheul93, line 2: ES lint was complaining about these errors. Specifically the `prop-types` error is confusing to me because it was complaining about `locale` not being in prop validation when it is included in the interface.

I wonder if it's because `locale` is an optional prop and maybe needs a default?

pwolfert: I went down a rabbit hole thinking that maybe the fact that our version of the eslint TypeScript plugin doesn't support our newer version of TypeScript has something to do with it. Unfortunately we can't update that plugin/parser without updating all of eslint, it seems. This error is certainly puzzling otherwise 🤔

packages/design-system/src/components/UsaBanner/UsaBanner.tsx
```diff
@@ -25,43 +33,104 @@ export interface UsaBannerProps {
...
+    setShouldRenderMobileView(evt.matches);
+  };
+
+  useEffect(() => {
+    if (window) {
+      const media = window.matchMedia('(max-width: 543px)');
+      media.addEventListener('change', onMediaChange);
+
+      setShouldRenderMobileView(media.matches);
+    }
```
pwolfert, line 53: Defining this inside the `useEffect` here might make the relationship clearer. It doesn't look like it's used anywhere else in the component

pwolfert (approved):
Are you going to also update the static HTML example too in this PR?

I noticed it wasn't updated when I saw this:

Screen Shot 2022-01-05 at 10 37 14 AM

Everything else LGTM 👍

PR #1473, opened 2022-01-06
[WNMGDS-1272] Clean up core types

pwolfert:
## Summary
Now that we've migrated all the individual files, we want to remove the root `src/types` folder and work out any issues through testing with our `create-react-app-typescript` project. Testing with that project found some issues which I've addressed here. See the full commit messages for details.


pwolfert:
After reading through https://stackoverflow.com/a/58123882, I realize that the annotations I removed in https://github.com/CMSgov/design-system/pull/1473/commits/4101b545efc78ab5c7544ba65f38f6c8d98c73b4 are not wrong, but they are still not necessary, so I'll move forward with the change.

PR #1489, opened 2022-01-10
[WNMGDS-1310] Autocomplete WHCM support

zarahzachz:
## Summary

[Jira](https://jira.cms.gov/browse/WNMGDS-1310)

### Added

Media query to target if WHCM is enabled. Inverting the background/foreground color palette when an option is selected.

## How to test

You can view the demo url through [our team's Sauce Labs account](http://saucelabs.com/) to ensure the changes are correct, or you can look at this pretty screenshot:
Screen Shot 2022-01-10 at 10 54 03 AM

If you opt to test via Sauce Labs, I'm using Edge 96 to view changes.

### DIFF
--- a/packages/design-system/src/styles/components/_Autocomplete.scss
+++ b/packages/design-system/src/styles/components/_Autocomplete.scss
@@ -53,6 +53,14 @@ $autocomplete-list-border: $autocomplete-list-border-width $autocomplete-list-bo
 .ds-c-autocomplete__list-item--active {
   background-color: $color-primary-alt-darkest;
   color: $color-white;
+
+  /* stylelint-disable scss/media-feature-value-dollar-variable */
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    background-color: CanvasText;
+    color: Canvas;
+    -ms-high-contrast-adjust: none;
+  }
+  /* stylelint-enable scss/media-feature-value-dollar-variable */
 }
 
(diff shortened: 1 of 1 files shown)


packages/design-system/src/styles/components/_Autocomplete.scss
```diff
@@ -53,6 +53,14 @@ $autocomplete-list-border: $autocomplete-list-border-width $autocomplete-list-bo
 .ds-c-autocomplete__list-item--active {
   background-color: $color-primary-alt-darkest;
   color: $color-white;
+
+  /* stylelint-disable scss/media-feature-value-dollar-variable */
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    background-color: CanvasText;
+    color: Canvas;
```
ericwbailey, line 60: What are your thoughts on using IE-specific values until forced colors is more globally supported? I say this because I think IE won't support the new values, but Edge supports legacy IE declarations.

ericwbailey (commented):
Just a question about IE and support.

pwolfert:
That's a beautiful screenshot!

packages/design-system/src/styles/components/_Autocomplete.scss
zarahzachz: That's a good call. Updating for IE compatibility now.

PR #1493, opened 2022-01-11
Publishing Mgov version 4.0.0

scheul93:
Version bumps for Mgov major version.

It also looks like the other packages were bumped, but I think those are just the current versions on `master`...?


zarahzachz (approved):
I can confirm these bumps already exist in master (except for Medicare, which is expected).

forrestbaer (approved):

PR #1495, opened 2022-01-11
[WNMGDS-1321] Netlify CMS demo

pwolfert:
## Summary
https://jira.cms.gov/browse/WNMGDS-1321

### Added
- Added a `/admin` content management interface using Netlify CMS
  - Configured it to be able to edit the _Guidelines_ and _Getting started_ collections of pages.
  - Uses GitHub to authenticate
  - Previews use the same page component as the actual docs pages, but it doesn't yet replace our special strings like `{{npm}}`, `{{root}}`, and `{{github}}`
  - Note that you're able to upload images, but our docs pages don't yet know how to interpret the image widgets generated by Netlify CMS

## How to test
1. `yarn install && yarn start`
2. Go to http://localhost:3000/admin/
3. Link your GitHub account
4. Try making an edit to a page and saving it. It will end up creating a pull request in your name with those content changes

http://design-system-demo.s3-website-us-east-1.amazonaws.com/pwolfert/netlify-demo/core/admin/


pwolfert:
For some reason this isn't working with the demo generator Jenkins job. Tried to fix it in `9420535`, but now it tries to load the admin script from a relative path that is actually the full path so that it looks like `http://design-system-demo.s3-website-us-east-1.amazonaws.com/pwolfert/netlify-demo/core/admin/pwolfert/netlify-demo/core/admin/index.js`

line47:
At first bush this looks really promising! I like the live preview a lot! Still some functionality I'm not sure how it works but this is really great!

pwolfert:
Right now this is working with a dummy site at https://naughty-benz-6d7e02.netlify.app/ and a config that points the git-gateway backend to the authentication endpoints there.

zarahzachz:
Love this so far. I have a couple questions/comments:
- What happens if you hit "Publish"? I was scared to try it because I didn't want to push something live (if that's what it did).
- I made a change to a page that was edited by you (#1503) and instead of creating a new PR with my changes, it force pushed it to yours. Is that expected behavior?
- If I manually add `/admin` to the local url (`localhost:3000/admin`) it adds a hash to the end and returns a 404 (I get `http://localhost:3000/admin#/`). I know this is because I didn't add the "/" to the end of my url, but wondering if this will prompt bug reports from other users?

packages/design-system-docs/src/admin/config.yml
```diff
@@ -0,0 +1,142 @@
...
+    folder: "packages/design-system-docs/src/pages/startup"
+    create: true
+    slug: "{{slug}}"
+    sortable_fields: ['weight']
+    fields:
+      - { label: "Title", name: "title", widget: "string" }
+      - { label: "Order", name: "weight", widget: "number" }
+      - { label: "Body", name: "body", widget: "markdown" }
+
+  # From example site, this is how we can configure individual pages that have uique structures
```
zarahzachz, line 42: Typo: "uique" should be "unique"

pwolfert:
@zarahzachz, thanks for the questions/comments!

## 1 & 2
Publishing will actually merge to master, which is maybe not what we want. We might want to tinker with that further. The way the "editorial_workflow" in Netlify CMS works (from their docs) is like so

Actions in Netlify UI | Perform these Git actions
-- | --
Save draft | Commits to a new branch (named according to the pattern cms/collectionName/entrySlug), and opens a pull request
Edit draft | Pushes another commit to the draft branch/pull request
Approve and publish draft | Merges pull request and deletes branch

## 3
I noticed that too, and you've got a good point. We should resolve that before we release this

forrestbaer:
Hmm I was able to make the change and saw the PR but it came through under your name, possibly because I created the netlify account a while back when we were first testing this and that's the way it was working?

forrestbaer:
> * If I manually add `/admin` to the local url (`localhost:3000/admin`) it adds a hash to the end and returns a 404 (I get `http://localhost:3000/admin#/`). I know this is because I didn't add the "/" to the end of my url, but wondering if this will prompt bug reports from other users?

Same thing happened for me at first, probably have to add a redirect to the docs site for that path or some config option.

pwolfert:
For now we've decided not to implement any kind of content management system, but I'm going to keep this branch in case we want to revisit it at a later date.

PR #1497, opened 2022-01-12
Publish core 3.0.0-beta, mgov 5.0.0-beta & hgvo 7.0.0-beta

scheul93:
- @cmsgov/design-system-docs@3.0.0-beta.1
 - @cmsgov/design-system-scripts@3.0.0-beta.1
 - @cmsgov/design-system@3.0.0-beta.1
 - @cmsgov/ds-healthcare-gov@7.0.0-beta.1
 - @cmsgov/ds-medicare-gov@5.0.0-beta.1


forrestbaer (approved):
lookin good!

packages/ds-healthcare-gov/package.json
```diff
@@ -20,10 +20,10 @@
     "src"
   ],
   "dependencies": {
-    "@cmsgov/design-system": "^2.13.0",
+    "@cmsgov/design-system": "^3.0.0-beta.1",
+    "@types/js-cookie": "^2.2.7",
```
zarahzachz, line 24: Was this needed as part of the version bump?

packages/ds-medicare-gov/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@cmsgov/ds-medicare-gov",
-  "version": "3.4.0-beta.7",
```
zarahzachz, line 3: Why does Medicare seem to not have the previous release version? I think this is the 2nd time I've seen this happen

packages/ds-healthcare-gov/package.json
scheul93: it's not a bump, the dependency name got moved around for unknown reasons. In the previous version of the file, you can see it on line 26

packages/ds-medicare-gov/package.json
scheul93: I think it's because the commit to bump the release happened after the cutoff for this release. Commit happened 1/11, the last commit in release was 1/7.
I'm not sure if you can cherry pick a commit forward or you cherry pick it onto the newly created tag here. 

That's why I think this is happening, open to ideas for solutions!

zarahzachz (approved):

PR #1498, opened 2022-01-12
[WNMGDS] Badges HCM support

ericwbailey:
## Summary

[JIRA ticket](https://jira.cms.gov/browse/WNMGDS-1307)

This PR updates [the Badge component](https://design.cms.gov/components/badge/) to support Windows High Contrast Mode

### Added

Media queries that:

- Update the badge's background color to define its shape and size
- Remove the backplate to ensure text is visible
- Component mixins to help keep declarations DRY

### Changed

The non-High Contrast Mode design utilizes color to indicate status (info, success, warning, alert). Since Windows High Contrast Mode uses a user-specified palette, color cannot be reliably utilized to indicate state.

In addition, we cannot know who is utilizing Badge components, so we cannot guarantee that their Badge label copy will indicate status via text. To get around this, we now prefix each status badge with CSS pseudo element content ("Notice:", "Success:", "Warning:", "Alert:"). 

While wordy, this ensures we do not violate [WCAG Success Criterion 1.4.1: Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html).

### Before

A 3 column, 5 row grid showing default and larger badge treatments in Windows High Contrast Mode for default, info, success, warning, and alert treatments. The background color that defines the shape of each badge is not present. Screenshot.

### After

A 3 column, 5 row grid showing default and larger badge treatments in Windows High Contrast Mode for default, info, success, warning, and alert treatments. The background color that defines the shape of each badge is set to text theme color, and the text 'Notice:', 'Success:', 'Warning:', and 'Alert:' prepend badges to indicate their status. Screenshot.

### DIFF
--- a/packages/design-system/src/styles/components/_Badge.scss
+++ b/packages/design-system/src/styles/components/_Badge.scss
@@ -12,14 +12,23 @@ $badge-font-size-big: $base-font-size;
   /* stylelint-enable scss/media-feature-value-dollar-variable */
 }
 
-@mixin badge-high-contrast-mode-label($label) {
-  /* stylelint-disable scss/media-feature-value-dollar-variable */
+@mixin badge-high-contrast-mode-label($label-en, $label-es: null) {
+  /* stylelint-disable */
   @media (-ms-high-contrast: active), (forced-colors: active) {
-    &::before {
-      content: $label + ': ';
+    [lang^='en'] & {
+      &::before {
+        content: $label-en + ': ';
+      }
+    }
+    @if ($label-es != null) {
+      [lang='es'] & {
+        &::before {
+          content: $label-es + ': ';
+        }
+      }
     }
   }
-  /* stylelint-enable scss/media-feature-value-dollar-variable */
+  /* stylelint-enable */
 }
 
 .ds-c-badge {
--- a/packages/design-system/src/styles/components/_Badge.scss
+++ b/packages/design-system/src/styles/components/_Badge.scss
@@ -21,7 +21,7 @@ $badge-font-size-big: $base-font-size;
       }
     }
     @if ($label-es != null) {
-      [lang='es'] & {
+      [lang^='es'] & {
         &::before {
           content: $label-es + ': ';
         }
--- a/packages/design-system/src/styles/components/_Badge.scss
+++ b/packages/design-system/src/styles/components/_Badge.scss
@@ -2,6 +2,26 @@
 
 $badge-font-size-big: $base-font-size;
 
+@mixin badge-high-contrast-mode-remove-color() {
+  /* stylelint-disable scss/media-feature-value-dollar-variable */
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    background-color: windowText;
+    color: window;
+    -ms-high-contrast-adjust: none;
+  }
+  /* stylelint-enable scss/media-feature-value-dollar-variable */
+}
+
+@mixin badge-high-contrast-mode-label($label) {
+  /* stylelint-disable scss/media-feature-value-dollar-variable */
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    &::before {
+      content: $label + ': ';
+    }
+  }
+  /* stylelint-enable scss/media-feature-value-dollar-variable */
+}
+
 .ds-c-badge {
   background-color: $color-gray;
   border-radius: $border-radius-pill;
@@ -15,6 +35,8 @@ $badge-font-size-big: $base-font-size;
   &:only-of-type {
     margin-right: 0;
   }
+
+  @include badge-high-contrast-mode-remove-color;
 }
 
 .ds-c-badge--big {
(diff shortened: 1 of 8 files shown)


packages/design-system/src/styles/components/_Badge.scss
```diff
@@ -23,17 +45,29 @@ $badge-font-size-big: $base-font-size;
 
 .ds-c-badge--info {
   background-color: $color-primary;
+
+  @include badge-high-contrast-mode-remove-color;
```
scheul93, line 49: Is this mixin included in each badge variation to increase the specificity of background-color styling?

Asking because, and I maybe reading this wrong, aren't the levels of specificity equal in this line as in line 39?

packages/design-system/src/styles/components/_Badge.scss
```diff
@@ -2,6 +2,26 @@
...
+@mixin badge-high-contrast-mode-remove-color() {
+  /* stylelint-disable scss/media-feature-value-dollar-variable */
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    background-color: windowText;
+    color: window;
+    -ms-high-contrast-adjust: none;
+  }
+  /* stylelint-enable scss/media-feature-value-dollar-variable */
+}
+
```
line47, line 14: Wondering if any of these mixins are generic enough to be applied to our WHCM fixes?

packages/design-system/src/styles/components/_Badge.scss
```diff
@@ -2,6 +2,26 @@
...
+@mixin badge-high-contrast-mode-label($label) {
+  /* stylelint-disable scss/media-feature-value-dollar-variable */
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    &::before {
+      content: $label + ': ';
```
zarahzachz, line 19: Very nice solution!

zarahzachz (approved):

packages/design-system/src/styles/components/_Badge.scss
ericwbailey: This is to enforce a HCM theme color in place of a color that might not be visible, and then signal to HCM that it shouldn't use a [backplate](https://www.gwhitworth.com/posts/2019/high-contrast-text-backplate/). 

I wrote this mixin mainly for DRY purposes, but we could also accomplish the same thing with the following selector:

``` css
.ds-c-badge,
.ds-c-badge--info,
.ds-c-badge--success,
.ds-c-badge--warn
.ds-c-badge--alert { … }
```

packages/design-system/src/styles/components/_Badge.scss
pwolfert: It's a good solution. Do we have an idea for how this could support Spanish, though? Maybe with a [language selector](https://developer.mozilla.org/en-US/docs/Web/CSS/:lang)?

scheul93 (approved):

packages/design-system/src/styles/components/_Badge.scss
ericwbailey: @pwolfert Oh, nice catch! I just pushed up an update that can utilize a Spanish language selector.

packages/design-system/src/styles/components/_Badge.scss
```diff
@@ -12,14 +12,23 @@ $badge-font-size-big: $base-font-size;
...
   @media (-ms-high-contrast: active), (forced-colors: active) {
-    &::before {
-      content: $label + ': ';
+    [lang^='en'] & {
+      &::before {
+        content: $label-en + ': ';
+      }
+    }
+    @if ($label-es != null) {
+      [lang='es'] & {
```
pwolfert, line 24: Do we want this to also be `^=` instead of `=`?

packages/design-system/src/styles/components/_Badge.scss
pwolfert: Nice, thank you! I added a comment below. Not sure if the difference between the two selectors was intentional. I could see our users using either `es-MX` or `es-US`. Like `es-US` would probably be more correct for our use cases, but I know of teams who are using `es-MX`.

packages/design-system/src/styles/components/_Badge.scss
ericwbailey: I was thinking of situations where the consuming design system has something like `lang="en-US"` set up.

pwolfert: I figured that, and I think `es` should get the same treatment

pwolfert: (See above comments about `es-MX` and `es-US`)

packages/design-system/src/styles/components/_Badge.scss
ericwbailey: Ah, I got my wires crossed there. I've updated the Spanish language selector to use `^=`, thank you!

pwolfert: Cool, LGTM 👍

PR #1501, opened 2022-01-14
[WNMGDS-1333] Bugfix - Steplist sublist on mobile

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1333)

### Changed

Changed the display type of our substep list items from `inline-block` to `inline`. 

**Note:** After testing locally and through the URL cited in that issue, this appears to be the right fix, but I am somewhat nervous about changing an elements display when it's nested within a component this deeply. What I'm getting at is that this component would be greatly improved by a refactor - there are a lot of nested `div`s and incorrect markup at play here that make it tough to maintain when style bugs pop up. And without vis regression in play, fixing these bugs gives me the heebie-jeebies.

## How to test

View [this demo URL](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1333/bugfix-steplist-list-styles/core/storybook/?path=/story/patterns-step-list--step-list) and change the step sublist to a ridiculously long string. Change the viewport to mobile and see that it's been fixed!

### DIFF
--- a/packages/design-system/src/styles/components/_StepList.scss
+++ b/packages/design-system/src/styles/components/_StepList.scss
@@ -201,7 +201,7 @@ $current-step-color: $color-primary !default;
 }
 
 .ds-c-substep__heading {
-  display: inline-block;
+  display: inline;
   margin-right: $spacer-1;
 }
 


pwolfert (approved):
This seems fine to me. I thought perhaps it was used to keep the edit button on the right side of the item in desktop-size screens, but that is being positioned absolute anyway and isn't affected by the `ds-c-substep__heading` element. You're probably right that this component is in need to some refactoring, but I also wonder if it's in need of some design love too.

PR #1514, opened 2022-01-18
[NO-TICKET] Clean up redundant dev dependencies

pwolfert:
## Summary

Right now we have 24 Dependabot PRs, and I'm hoping this eliminates 8 of them. These are all redundant devDependencies that are in the root-level package.json


forrestbaer (approved):
looks good, everything built/tested/ran as expected

PR #1516, opened 2022-01-18
[WNMGDS] Alerts HCM support

ericwbailey:
## Summary

[JIRA ticket](https://jira.cms.gov/browse/WNMGDS-1309).

This PR updates [the Alert component](https://design.cms.gov/components/alert/) to support Windows High Contrast Mode.

### Background

The non-High Contrast Mode design utilizes color to indicate status (info, success, warning, alert). Since Windows High Contrast Mode uses a user-specified palette, color cannot be reliably utilized to indicate state if the status icon is missing.

In addition, we cannot know who is utilizing Alert components, so we cannot guarantee that their copy will indicate status via text. To get around this, we now prefix the first content node with CSS pseudo element content ("Notice:", "Success:", "Warning:", "Alert:") if an icon is not present.

While wordy, this ensures we do not violate [WCAG Success Criterion 1.4.1: Use of Color](https://www.w3.org/WAI/WCAG21/Understanding/use-of-color.html).

Note: The SVG icon color issue will be addressed in [WNMGDS-1306](https://jira.cms.gov/browse/WNMGDS-1306).

### Added

Media queries and selectors that:

- Update the alert's borders to define an alert's boundaries.
- Conditionally apply or remove CSS pseudo element content labels.

### Changed

- The [`badge-high-contrast-mode-label` mixin](https://github.com/CMSgov/design-system/pull/1498/files#diff-dba3dcc7a4a70ee67de55afe9226d669f58d48373d3245fcd10df07d3ead0a07R15) was refactored to be a globally-available mixin, and moved to `settings/mixins/`.
    - The badge component is updated to use the renamed mixin.
- A companion mixin, `high-contrast-mode-label-remove`, is also included.

### Before

High Contrast Mode screenshot showing informative, error, warning, and success alert components with status icons. Additionally, a informative alert without an icon component variant is shown. 3 of the borders of the component are not visible and you cannot determine status type if an icon is present.

#### Icon present

High Contrast Mode screenshot showing informative, error, warning, and success alert components. A border has been added, delineating the alert component.

#### No icon present, heading present

High Contrast Mode screenshot showing informative, error, warning, and success alert components. The status icon has been removed, which prefixes the term "Notice:", "Error:", "Warning:" or "Success before the alert heading, depending on the alert's type.

#### No icon or heading present

High Contrast Mode screenshot showing informative, error, warning, and success alert components. The status icon and alert heading have been removed, which prefixes the term "Notice:", "Error:", "Warning:" or "Success before the alert's paragraph content, depending on the alert's type.

#### No icon or heading present, multiple paragraphs

High Contrast Mode screenshot showing informative, error, warning, and success alert components. The status icon and alert heading have been removed, which prefixes the term "Notice:", "Error:", "Warning:" or "Success before the alert's paragraph content, depending on the alert's type. A second paragraph has been added to each alert to demonstrate the prefixes are only applied to the first paragraph.


packages/design-system/src/styles/settings/mixins/_high-contrast-mode-label.scss
```diff
@@ -0,0 +1,31 @@
+@mixin high-contrast-mode-label-add($label-en, $label-es: null) {
```
scheul93, line 1: How will the Spanish translation get added in? I see that there's a parameter for it, but like who is in charge of providing those?

forrestbaer:
I wonder if it makes more sense to add the warning text at the component level to handle i8n stuff and not include logic in our styles..

packages/design-system/src/styles/settings/mixins/_high-contrast-mode-label.scss
ericwbailey: That's a good question. Maybe we should speak with a content person? I was hesitant to add the text in, as I don't speak Spanish (also thinking regional considerations?).

ericwbailey:
@forrestbaer I was thinking about going that direction, but decided against it because I was worried exposing it as a prop would mean it might be adjusted by an author and accidentally violate SC 1.4.1. 

I also don't super like using CSS as a source of content, as per @scheul93's comment. Maybe the approach is to add text to the component and have a `.s-u-visibility--hcm-only` utility class, kind of like our `.s-u-visibility--hidden-print` utility class?

forrestbaer:
> I also don't super like using CSS as a source of content, as per @scheul93's comment. Maybe the approach is to add text to the component and have a `.s-u-visibility--hcm-only` utility class, kind of like our `.s-u-visibility--hidden-print` utility class?

This seems like a good approach.

line47:
> I also don't super like using CSS as a source of content, as per @scheul93's comment. Maybe the approach is to add text to the component and have a .s-u-visibility--hcm-only utility class, kind of like our .s-u-visibility--hidden-print utility class?

I like this approach too 🥳

ericwbailey:
@zarahzachz just pushed up a pretty large refactor to conditionally add the text via `matchMedia`, matching the approach in https://github.com/CMSgov/design-system/pull/1537/.

I've also created [a Jira ticket](https://jira.cms.gov/browse/WNMGDS-1355) to discuss threading localization into this approach as a followup.

With this approach, I don't think we need the `ds-u-visibility--high-contrast` utility class, so I will be removing it. If we feel it'd be useful, I think that's something now out of scope of this ticket's ask, and that we could investigate it in its own ticket.

packages/design-system/src/styles/components/_Alert.scss
```diff
@@ -46,6 +46,15 @@ $alert-link-color-focus: $color-primary-darkest;
     }
   }
 
+  /* stylelint-disable */
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    border-top: 1px solid windowText;
+    border-right: 1px solid windowText;
+    border-bottom: 1px solid windowText;
+    border-left: $alert-bar-size solid windowText;
```
zarahzachz, line 54: I wonder if this could be consolidated into 2 lines?
```css
border: 1px solid windowText;
border-left-style: $alert-bar-size;
```

zarahzachz (approved):
I have a small suggested improvement, but approving!

packages/design-system/src/styles/components/_Alert.scss
ericwbailey: Ooh, I like that.

PR #1530, opened 2022-01-21
Working on fixing the Button prop types

pwolfert:
## Summary

Overview of changes

### Added

### Changed

List changes in existing functionality or design.
If the change was visual, include a comparison screenshot showing the before and after the visual change.

### Deprecated

### Removed

List deprecated features or components removed in this PR.

### Fixed

List any bug fixes.

## How to test

Instructions on how to test the changes. This is not exhaustive list of ways you should test this PR.

### DIFF
--- a/packages/design-system/src/components/Button/Button.test.tsx
+++ b/packages/design-system/src/components/Button/Button.test.tsx
@@ -11,7 +11,7 @@ const defaultProps = {
   children: 'Foo',
 };
 
-function renderButton(props: Partial<ButtonProps> = {}) {
+function renderButton(props: Partial<ButtonProps<any>> = {}) {
   return shallow(<Button {...defaultProps} {...props} />);
 }
 
--- a/packages/design-system/src/components/Button/Button.tsx
+++ b/packages/design-system/src/components/Button/Button.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
 import classNames from 'classnames';
 
-export type ButtonComponent = React.ReactElement<any> | any | ((...args: any[]) => any);
+export type CustomButtonComponentType = React.ComponentType<any> | React.FC;
+export type ButtonComponentType = React.ElementType<any> | CustomButtonComponentType;
 export type ButtonSize = 'small' | 'big';
 export type ButtonType = 'button' | 'submit';
 /**
@@ -10,7 +11,7 @@ export type ButtonType = 'button' | 'submit';
  */
 export type ButtonVariation = 'primary' | 'danger' | 'success' | 'transparent';
 
-type CommonButtonProps<T> = {
+type CommonButtonProps<T extends ButtonComponentType> = {
   /**
    * Label text or HTML
    */
@@ -26,6 +27,11 @@ type CommonButtonProps<T> = {
    */
   component?: T;
   disabled?: boolean;
+  /**
+   * When provided the root component will render as an `<a>` element
+   * rather than `button`.
+   */
+  href?: string; // Still optional because it's optional on the anchor tag
   /**
    * Access a reference to the `button` or `a` element
    */
@@ -55,27 +61,18 @@ type CommonButtonProps<T> = {
 
 type OmitProps = 'children' | 'className' | 'onClick' | 'ref' | 'size' | 'type' | 'href';
 
-type LinkButtonProps = CommonButtonProps<'a'> &
-  Omit<React.ComponentPropsWithRef<'a'>, OmitProps> & {
-    /**
-     * When provided the root component will render as an `<a>` element
-     * rather than `button`.
-     */
-    href?: string; // Still optional because it's optional on the anchor tag
-  };
+export type ButtonProps<T extends ButtonComponentType> = CommonButtonProps<T> &
+  (T extends React.ElementType ? Omit<React.ComponentPropsWithRef<T>, OmitProps> : unknown);
 
-type ButtonButtonProps = CommonButtonProps<'button'> &
-  Omit<React.ComponentPropsWithRef<'button'>, OmitProps>;
-
-export type ButtonProps = CommonButtonProps<ButtonComponent> | LinkButtonProps | ButtonButtonProps;
-
-export class Button extends React.PureComponent<ButtonProps> {
+export class Button<T extends ButtonComponentType = 'button'> extends React.PureComponent<
+  ButtonProps<T>
+> {
   static defaultProps = {
     type: 'button',
     component: 'button',
   };
 
-  constructor(props: ButtonProps) {
+  constructor(props: ButtonProps<T>) {
     super(props);
 
     if (process.env.NODE_ENV !== 'production') {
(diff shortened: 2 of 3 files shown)


pwolfert:
There are a lot of problems with the typing of button props.

If I use a generic for the props type, it needs to be able to infer the type of the `component` prop even though it's not required and its type could actually be `"a"` when it is either omitted or `"button"`. This means TypeScript can classify the `component` as a `"button"` when it should have the intrinsic properties of an `"a"`.

This was my first solution:
```tsx
export type CustomButtonComponentType = React.ComponentType<any> | React.FC;
export type ButtonComponentType = React.ElementType<any> | CustomButtonComponentType;

type CommonButtonProps<T extends ButtonComponentType> = {
  ...
  component?: T;
  href?: string;
  ...
};

export type ButtonProps<T extends ButtonComponentType> = CommonButtonProps<T> &
(T extends React.ElementType ? Omit<React.ComponentPropsWithRef<T>, OmitProps> : unknown);

export class Button<T extends ButtonComponentType = 'button'> extends React.PureComponent<ButtonProps<T>> {
  ...
```

This fixes the errors in `windowshop`, but we no longer have proper restrictions on intrinsic attributes like `size`, and it allows completely arbitrary props on `Button` like `bob`. It also became very difficult to type in the DrawerToggle because of the generic. It throws cryptic errors at the usage of `Button` in `DrawerToggle` (see `cfe67b5`).

I can forego the generic type and create a union of the intrinsic properties of both `a` elements and `button` elements, like below:

```tsx
export type ButtonProps = CommonButtonProps &
  Omit<React.ComponentPropsWithRef<'a'>, OmitProps> &
  Omit<React.ComponentPropsWithRef<'button'>, OmitProps>;
```

but if they pass a custom component for `component`, TypeScript won't know about any of the props that exist on that component. For instance,

```tsx
interface LinkProps {
  children: React.ReactNode;
  to: "string";
}

const Link = (props: LinkProps) => {
  return <div {...props}>{props.children}</div>;
};

const ThingA = () => (
  <Button 
    component={Link}
    to="nowhere"
  >
    Hello World
  </Button>
);
```

In this case TypeScript won't know that `Button` should allow for the prop `to`, which is passed on to the `Link`.

---

I've been exploring some alternatives in https://github.com/CMSgov/design-system/compare/pwolfert/fix-button-types-2?expand=1 too. Some of that is copied and pasted from https://www.benmvp.com/blog/polymorphic-react-components-typescript/

Look at this part:
```tsx
type DetermineButtonComponentType<Props extends { component: unknown }> = Props extends {
  component: CustomButtonComponentType;
}
  ? CustomButtonComponentType
  : Props extends { href: string }
  ? 'a'
  : Props['component'];
  ```
I'd like to use this logic to help TypeScript determine a more specific type for `component`—either a React component, `a`, or something else. I want to be able to tell TS, "If it's a React component, make sure to add that component's props; if it's a plain html tag name, make sure to add its intrinsic properties."

forrestbaer:
Let me re-state the problem with a few questions, as simple as I can, maybe it will help:

- The component prop in the button component can be one of:
    1. nothing, in which case it will become an <a> element (currently typed as any), which should include an href
    2. a button element with the current set of button props? (buttons all the way?)
    3. a button element wrapping an <a> with href and any other valid tags on an <a> element and a button element
    4. a custom component which may contain any number of type definitions for its props
- The interface from the inherited 'component' element should come along

Would defining the set of options as several types then using a [merge] (https://stackoverflow.com/a/53936938) of that set of types with the button props do the trick? Do the React.HTMLComponents types need to be included for the possible html attributes?

I'm sure I don't fully understand the problem!

pwolfert:
@forrestbaer, yeah, that's helpful to state the problem. I see the `component` prop having these possible values:
1. A custom React component
2. A string representing an element tag name
3. Nothing, in which case it will default to `button` but might still end up being rendered as an `a` if the `href` prop is a string, which is where things get confusing.
  - Maybe to avoid confusing ourselves and TypeScript we need to allow it to be undefined without a default value so that an `Button` with an `href` property doesn't end up having `button` as its `component` value. `component: undefined` becomes "figure it out for me", but if the user sets it, that becomes law. We should go down this path and see where it leads, I think.

Before seeing your comment and responding, I created [a third branch](https://github.com/CMSgov/design-system/compare/pwolfert/fix-button-types-3?expand=1) that goes back to intersection types where the props can only be one of three shapes. It worked in [this TypeScript Playground](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocpAD0ljhgE8wk4BhAVwGcZcAhDjF4A7NrkjCkwmABVmrALyJiMAHRjwESdLksAPCmGMAfHAA+y9GoBibCjTrwmLOAKFaNEqbPlwlyK1UAUQAbJBBvXSQDI1MLTh5+QRFPLUj5CnI0LR44AEEALldkj3E0nV8lAAoASj9TPQATYAA3YwALJBCQiD0Aema2rJz4PiK3FLLtHxclfBQySmzhLggw1R6Acyq9PLg+4xrhlbWkDYhtvT59w8pnVg0QLQmtAAUcMC4-OABvcjg4Gh2sAQo0iMIijwoMBhJtzJYMKoAhgAHIQRpICgA9pETAAfkhMGhsKxcA8IWAaAA1gS4FVVAyUFBNlwioZGABtAC6dQUpnZpK4wAAXkhafgAEbATb4eH4LggFDdRYAX0y9zgAHkQMAYO8IJ9vvggSCwVJZRZjSEUFwuCiUBELQRyZSqU7CEhMO6haL3fd3TjPYsNfrPprMHoZHAaDApI0vi9RFN0ixTEptbq9Mi1KlpqGuAB1XXtZARmTGAA0Wp1eo+XGM6t8ABkYVTE-nvhmYHpHs8SsJ81X8IGvaYAGS-f6A5PScULUkjwnEzYUFVwCf58N6eb4BvkDWJ9t17694RHg1fCd-AHZTTTcUS-uq9dwTcRyVPvcahK8EDnw1KKe-6XpON4zjARQ-kk7hJneKaYuQaqUA40BOL4wHfC2whtv2HYWIeuHHvE3C-sBmTLLkqwRPmACM4yERe3w-ICwKguCRQAESdN0EAcXASEUfAVFIPmABM9EwR2SjMSabFSJxwIcVWi5wFxXQ9HxKrHJRuAiXWADMEkiFJvwsaa7GqYwvHKbiRQcMIGKYDCSCNPxmQofQgnFDB3xVGAdZGW8da8qY15wGE8C5vB3z+Reqi3l40hwHieIEI+MGLACfR9HAACSmBwLFnzxeBcDAF8H4ZVWuplV8wgQPAKCAiRuAIvACXlGoeXCICNpIFwU7ZXAADunREAwnRwNwSBQHAzSNMI+BCSwaDAJgjBwE1aAtXgHXTJtDlzcArkwmgIQcBiB1wCOVbDUgg05cNhhOBAU1cKwAAGBjGB9cCYNAZIcLNXThN4qhTmtdJRRUswKHM6UiLKY4Tn5x42lN2H1cNPXo1hOGScFqgjnUYVZTlKJIC5DCvTa730Oyf0A1EADKaDQmA8ASkgaAoNNcBRuVDDc+0wiUkqIQbZViNTgC0MzIoBALJtXwClOWkAkQMBAz1PbgVExg-EVXDxaxZrCCq-Ry-riHkTkpznJcibGAAEupvR9E7RzafbWw7Im124goak8RxxgFtAoL9J7QA), but if you pull that branch and run the type checker, it throws errors when trying to use `<Button>` elsewhere.

pwolfert:
Closing in favor of https://github.com/CMSgov/design-system/pull/1532

PR #1520, opened 2022-01-19
[WNMGDS] Accordion HCM support

ericwbailey:
## Summary

[JIRA ticket](https://jira.cms.gov/browse/WNMGDS-1308).

This PR updates [the Accordion component](https://design.cms.gov/components/accordion/) to support Windows High Contrast Mode.

### Added

Media queries that update the Accordion's header button to better describe its size and interactable status.

### Before

Screenshot of the accordion component with Windows High Contrast Mode active. The accordion heading backgrounds are invisible.

### After

https://user-images.githubusercontent.com/634191/150204289-f32208dd-687f-47c1-ad87-7d0073eb9ca6.mp4

Accordion header button is visible in its resting state, and hover/focus interactions are now added.

PR #1531, opened 2022-01-21
[WNMGDS] Vertical nav HCM support

ericwbailey:
## Summary

[Jira ticket.](https://jira.cms.gov/browse/WNMGDS-1314)

This PR updates [the Vertical nav component](https://design.cms.gov/components/vertical-nav/) to support Windows High Contrast Mode.

### More info

The High Contrast Mode design has been updated to use the same [HCM visual treatment for the accordion component](https://github.com/CMSgov/design-system/pull/1520). Additionally, the hierarchal/nested subnavigation has been adjusted to be more apparent. This will be helpful for users with low vision and/or screen magnification concerns, where the relationships between nested levels is more obvious.

### Added

Media queries that update the Accordion's header button to better describe its interactable status, hierarchy, and currently active nav item.

### Before

Screenshot of the vertical nav component in Windows High Contrast Mode. Nested hierarchy is not apparent, as well as the ability to toggle subcategories open or closed.

### After

Screenshot of the vertical nav component in Windows High Contrast Mode. There is a more obvious accordion treatment for nested nav, as well as highlights for active navigation, and left margin insets to show nesting.


packages/design-system/src/styles/components/_VerticalNav.scss
```diff
@@ -75,6 +83,24 @@
     @extend %nav-item-focus;
   }
 
+  /* stylelint-disable scss/media-feature-value-dollar-variable */
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    /* stylelint-disable order/properties-alphabetical-order */
```
ericwbailey, line 88: I have to add this here to get `-ms-high-contrast-adjust: none;` to cascade properly on lines 89 and 94.

packages/design-system/src/styles/components/_VerticalNav.scss
```diff
@@ -100,9 +132,39 @@
...
+    }
+    /* stylelint-enable scss/media-feature-value-dollar-variable */
   }
 
   .ds-c-vertical-nav__subnav .ds-c-vertical-nav__label {
     padding-left: $spacer-4;
   }
 }
+
+/* stylelint-disable */
```
ericwbailey, line 148: I normally am loathe to do this, but needed to for the declarations we need to get the desired visual changes working.

packages/design-system/src/styles/components/_VerticalNav.scss
```diff
@@ -100,9 +132,39 @@
...
+.ds-c-vertical-nav__subnav {
+  @media (-ms-high-contrast: active), (forced-colors: active) {
+    margin-left: $spacer-2;
+    width: calc(100% - $spacer-2);
+  }
+}
+
+.ds-c-vertical-nav__subnav .ds-c-vertical-nav__item:last-of-type {
```
ericwbailey, line 156: This is a nasty selector, but I couldn't think of a more elegant way to do it.

packages/design-system/src/styles/components/_VerticalNav.scss
ericwbailey, line 148: I also placed this gnarly HCM-only styling at the end of the Sass partial because I was feeling like its addition, in addition to the Stylelint comments, was negatively affecting the partials legibility.

pwolfert:
The new screenshot looks so much better!

scheul93 (approved):

forrestbaer (approved):
looks great

PR #1532, opened 2022-01-22
[WNMGDS-1344][WNMGDS-1345] Fix Button types

pwolfert:
## Summary
https://jira.cms.gov/browse/WNMGDS-1344
https://jira.cms.gov/browse/WNMGDS-1345

### Changed
- Converted `Button` to a functional component because it was easier to type

### Fixed
- Fixed TS errors downstream where TypeScript was flagging buttons with the `secondary` variation in ds-healthcare-gov as invalid (`Type '"secondary"' is not assignable to type '"success" | "primary" | "danger" | "transparent" | undefined'`). See https://github.com/CMSgov/design-system/commit/0cd1b41213e243c5f5b674fc4bbc1671a19bb571
- Fixed TS error downstream where buttons with no explicit `components` prop and with `href` were flagged as invalid, because TypeScript believed they should have the attributes `button` elements instead of `a` elements.

### DIFF
--- a/packages/design-system/src/components/Button/Button.tsx
+++ b/packages/design-system/src/components/Button/Button.tsx
@@ -64,134 +64,104 @@ type OmitProps = 'children' | 'className' | 'onClick' | 'ref' | 'size' | 'type' 
 export type ButtonProps<T extends ButtonComponentType> = CommonButtonProps<T> &
   (T extends React.ElementType ? Omit<React.ComponentPropsWithRef<T>, OmitProps> : unknown);
 
-export class Button<T extends ButtonComponentType = 'button'> extends React.PureComponent<
-  ButtonProps<T>
-> {
-  static defaultProps = {
-    type: 'button',
-    component: 'button',
-  };
-
-  constructor(props: ButtonProps<T>) {
-    super(props);
-
-    if (process.env.NODE_ENV !== 'production') {
-      if (props.inverse) {
-        console.warn(
-          `[Deprecated]: Please remove the 'inverse' prop in <Button>, use 'inversed' instead. This prop has been renamed and will be removed in a future release.`
-        );
-      }
-      if (props.variation === 'danger') {
-        console.warn(
-          `[Deprecated]: Please remove the 'danger' variation prop in <Button>. This prop has will be removed in a future release.`
-        );
-      }
+export const Button = <T extends ButtonComponentType = 'button'>(props: ButtonProps<T>) => {
+  if (process.env.NODE_ENV !== 'production') {
+    if (props.inverse) {
+      console.warn(
+        `[Deprecated]: Please remove the 'inverse' prop in <Button>, use 'inversed' instead. This prop has been renamed and will be removed in a future release.`
+      );
     }
-
-    this.handleClick = this.handleClick.bind(this);
-    this.handleKeyPress = this.handleKeyPress.bind(this);
+    if (props.variation === 'danger') {
+      console.warn(
+        `[Deprecated]: Please remove the 'danger' variation prop in <Button>. This prop has will be removed in a future release.`
+      );
+    }
   }
 
-  // Get an object of props to pass to the rendered <Button> component
-  attrs(): any {
-    /**
-     * Since any number of arbitrary props can be passed into this component, we
-     * use a destructuring assignment to get only the props we want to pass to the
-     * rendered HTML element. For example, the "variation" prop is used to generate
-     * the classNames, but doesn't need to be passed to the rendered component, so
-     * we omit it here so that it's not included in the props object.
-     */
-    const {
-      className,
-      component,
-      inputRef,
-      inversed,
-      inverse,
-      onClick,
-      size,
-      variation,
-      ...props
-    } = this.props;
-
-    const attrs: any = {
-      className: this.classNames(),
-      ...props,
-    };
-
-    if (this.props.onClick) {
-      attrs.onClick = this.handleClick;
+  function handleClick(e: React.MouseEvent | React.KeyboardEvent): void {
+    if (!props.disabled && props.onClick) {
+      props.onClick(e);
     }
-
-    if (component !== 'button' || this.props.href) {
-      // Assume `component` is not a <button> and remove <button> specific attributes
-      attrs.role = 'button';
-      delete attrs.disabled;
-      delete attrs.type;
-    }
-
-    return attrs;
   }
 
-  componentType(): ButtonComponentType {
-    let component = this.props.component;
-
-    // If props.component is 'button', it is not a custom React component. In cases
-    // where the user didn't specify a custom component and did include an href, we
-    // want to use `<a>` for our element.
-    if (component === 'button' && this.props.href) {
-      // Need to assert any for TypeScript because T is technically 'button'
-      component = 'a' as any;
-    }
-    return component;
-  }
-
-  classNames(): string {
-    const variationClass = this.props.variation && `ds-c-button--${this.props.variation}`;
-
-    const disabledClass =
-      this.props.disabled && this.componentType() !== 'button' && 'ds-c-button--disabled';
-
-    const sizeClass = this.props.size && `ds-c-button--${this.props.size}`;
-
-    const inverseClass = (this.props.inversed || this.props.inverse) && 'ds-c-button--inverse';
-
-    return classNames(
-      'ds-c-button',
-      disabledClass,
-      variationClass,
-      inverseClass,
-      sizeClass,
-      this.props.className
-    );
-  }
-
-  handleKeyPress(e: React.KeyboardEvent): void {
+  function handleKeyPress(e: React.KeyboardEvent): void {
     // Trigger onClick on space key event for `<a>` elements
     if (e.key === ' ') {
-      this.handleClick(e);
+      handleClick(e);
     }
   }
 
-  handleClick(e: React.MouseEvent | React.KeyboardEvent): void {
-    if (!this.props.disabled && this.props.onClick) {
-      this.props.onClick(e);
-    }
+  /**
+   * Since any number of arbitrary props can be passed into this component, we
+   * use a destructuring assignment to get only the props we want to pass to the
+   * rendered HTML element. For example, the "variation" prop is used to generate
+   * the classNames, but doesn't need to be passed to the rendered component, so
+   * we omit it here so that it's not included in the props object.
+   */
+  const {
+    className,
+    component,
+    inputRef,
+    inversed,
+    inverse,
+    onClick,
+    size,
+    variation,
+    ...otherProps
+  } = props;
+
+  let ComponentType = component;
+  // If props.component is 'button', it is not a custom React component. In cases
+  // where the user didn't specify a custom component and did include an href, we
+  // want to use `<a>` for our element.
+  if (ComponentType === 'button' && props.href) {
+    // Need to assert any for TypeScript because T is technically 'button'
+    ComponentType = 'a' as any;
   }
 
-  public render() {
-    const attrs = this.attrs();
-    const ComponentType = this.componentType();
+  const variationClass = variation && `ds-c-button--${variation}`;
+  const disabledClass = props.disabled && ComponentType !== 'button' && 'ds-c-button--disabled';
+  const sizeClass = size && `ds-c-button--${size}`;
+  const inverseClass = (inversed || inverse) && 'ds-c-button--inverse';
+  const allClassNames = classNames(
+    'ds-c-button',
+    disabledClass,
+    variationClass,
+    inverseClass,
+    sizeClass,
+    className
+  );
 
-    return (
-      <ComponentType
-        ref={this.props.inputRef}
-        onKeyPress={ComponentType === 'a' ? this.handleKeyPress : undefined}
-        {...attrs}
-      >
-        {this.props.children}
-      </ComponentType>
-    );
+  const attrs: any = {
+    className: allClassNames,
+    ...otherProps,
+  };
+
+  if (props.onClick) {
+    attrs.onClick = handleClick;
   }
-}
+
+  if (component !== 'button' || props.href) {
+    // Assume `component` is not a <button> and remove <button> specific attributes
+    attrs.role = 'button';
+    delete attrs.disabled;
+    delete attrs.type;
+  }
+
+  return (
+    <ComponentType
+      ref={inputRef}
+      onKeyPress={ComponentType === 'a' ? handleKeyPress : undefined}
+      {...attrs}
+    >
+      {props.children}
+    </ComponentType>
+  );
+};
+
+Button.defaultProps = {
+  type: 'button',
+  component: 'button',
+};
 
 export default Button;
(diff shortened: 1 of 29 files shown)


packages/design-system/src/components/Button/Button.tsx
```diff
@@ -54,144 +43,120 @@ type CommonButtonProps<T> = {
...
-
-    if (this.props.onClick) {
-      attrs.onClick = this.handleClick;
+export type ButtonProps<T extends ButtonComponentType> = PropsOf<T> &
+  CommonButtonProps &
+  (
+    | LinkTypeButtonProps
+    | DefaultButtonTypeButtonProps
+    | (T extends NonAnchorButtonComponentType ? OtherTypeButtonProps<T> : never)
+  );
```
pwolfert, line 73: When coming across instances of `<Button>` in downstream apps, TypeScript wasn't able to determine which type of the intersection type `type ButtonProps = CommonButtonProps<ButtonComponent> | LinkButtonProps | ButtonButtonProps` was at play, so it would complain that props like `href` and `target` did not exist on `CommonButtonProps<ButtonComponent>` when it should have been using `LinkButtonProps`. The fix is to make the `Button` component itself a generic type, showing TypeScript which part of it is variable (`component`) and then try to help TypeScript narrow it down which specific type it is in this union (term is "discriminated union") by giving it hints about the `href` value and such.

The part about the last type in this union being `(T extends NonAnchorButtonComponentType ? OtherTypeButtonProps<T> : never)` is a later development, which I shall address in a follow-up comment.

pwolfert: From my https://github.com/CMSgov/design-system/pull/1532/commits/ceeee3154e26b987814ce7ce9ceded4b6244a58f commit message:

The significance of this is that sometimes instances of Button even though they should be identified as having `LinkTypeButtonProps` (where `href` is strictly required)—like when explicitly passing `component='a'`—TypeScript seemed to be thinking it was the more generic `OtherTypeButtonProps<T>`. I _think_ this is working now.

Error that flags a missing `href` prop if the component has an explicit `component='a'` prop:
```
packages/design-system/src/components/Button/Button.test.tsx:36:8 - error TS2322: Type '{ component: "a"; target: "_blank"; type: string; children: string; }' is not assignable to type 'IntrinsicAttributes & ButtonProps<"a">'.
  Property 'href' is missing in type '{ component: "a"; target: "_blank"; type: string; children: string; }' but required in type 'LinkTypeButtonProps'.

36       <Button
          ~~~~~~

  packages/design-system/src/components/Button/Button.tsx:56:3
    56   href: string;
         ~~~~
    'href' is declared here.``
```

Note that it is correctly identifying this as a `LinkTypeButtonProps`.

Before this change, no error was thrown for a missing `href`, but in a similar situation where we explicitly pass a `component='a'` prop but give an _invalid_ value for `href`, we'd get the following error which illustrates TypeScript's incorrect inference about the type of the component props:
```
packages/design-system/src/components/Button/Button.test.tsx:36:8 - error TS2322: Type '{ component: "a"; href: { hello: string; }; target: "_blank"; type: string; children: string; }' is not assignable to type 'IntrinsicAttributes & ButtonProps<"a">'.
  Type '{ component: "a"; href: { hello: string; }; target: "_blank"; type: string; children: string; }' is not assignable to type 'OtherTypeButtonProps<"a">'.
    Types of property 'href' are incompatible.
      Type '{ hello: string; }' is not assignable to type 'null'.

36       <Button
          ~~~~~~
```
Note that it thinks our button has `OtherTypeButtonProps<"a">` props. We don't want that more generic type to be used when the more specific `LinkTypeButtonProps` should be used!!

forrestbaer:
A really complex problem that may get more complex with growing needs of our users? Could we start a discussion about the tradeoff between strongly typing everything and future code complexity, maintainability and simplicity? I think this is a good example of some of these tradeoffs.

pwolfert:
For future reference, here is a [TypeScript playground link](https://www.typescriptlang.org/play?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wG4AocpAD0ljhgE8wk4BhAVwGcZcAhDjF4A7NrkjCkwmABVmrALyJiMAHRjwESdLksAPCmGMAfHAA+y9GoBibCjTrwmLOAKFaNEqbPlwlyK1UAUQAbJBBvXSQDI1MLTh5+QRFPLUj5CnJnVg0QLTcRAAUcMC4-OABvcjg4NAALYBCAEyJhAC44HihgYQBzc0sMVQCMADkIJqQKGo8Q4DQAawB+DoAKVQ2UKF6uDsNGAG0AXQBKP1N96c7gAC8kFYIAI2Be-AH8LhAUEJCycgBfTLZOAAWSQ22iIIANHBRqYlAB5EDAGB6aFwILUGBQKx6BZIRgQTCgmH4wnEuGmABksKBvgAMj0FlEClpihBSuUwRC9CM1KltDB2aUAOoourITB6fAofDGGG5fLJNklMo0qo1NDiNLSB4yv41OpETAdLo9XoUf7GCjAgAiSEwKA4IRgrOELOVwmFZSU3N60T56m1gu9YpgEod0senrlCtweWEbu9cHV1Vqwe8euj7mEBrgRodDw4wkmmB6SCaA2EzpClutWV8CPD4I9Oe9ehkcBoMCkTTKboF6RY8NB4P9vJUQc0IdVYYjUpk8vY8aVbdVKcqaa10+8HRkVwLmCLJYd5crFmrPzrFAA9DeGL5xsIAILCerQAcZnS+BTkO81AYEl4EBPx3b8WD-e8agsQNQnCIdokxNAQg4SY9DNPoYX1d5sxEOV63-YEmzqFt5CTVUOy7LFezKJ9X3fKBQK8cCkBHP1okggDA0HaRQ3FSU9CQlC0JkLDZRwmNjHlTiakVRNPWTVN-wA9MwJgPdbyggDD2PUszyrGtNIA-5OPrBxoCcXxyI5LhKO7GjXE9HifGHPw0wsRlhGZMiFNVdy4HtR1nVdT1WyKPzoLgYjSJYazSg7etyC1YQeE6XAkG9ABGDo4ts-BcK0OVygqWoGmaVoOgAIhIn4IEqmEYC2f11LgSqAH1HhCQwFkquBASSrRUq4dLvQAJhy3ybOlWURxK+pGhaKQqoaer82NZakFq1bGohFr2s67repM5KhpG1UAGYJrXKb8B6MBBCKpQ5rKxb2lawlVsPDpiz0yQmgapqkD2jquq81b6iQRYKw6bEOCQGFKq2YAUAAWghqGmkqmGoDhhr5CqrgOEeZEYCO0g4H-ABlOoIGdStwxwAB3LsoBwKBKHM+gTvgN1yjs6iS37Jyvxc1jVjAVUrvCqbFzOBRTA1dMUvgZyonKCWbNUbdmPgJYljgcXVVUQ84H17COnymMTkyGoiBgDgoGEOA9FV+RjAqDXSi1l7Wn+PQb1d4cAUyE6IDCVQQggXpVj0N1jAACU2yP-bjk4BpSsOkAjqOY95w8FGqpO6uMEVoGaFPPWMNPQ-DyPo9jz1VJ1gu7sEXrsgLwniZRSqS7LpoK5zKvKB6HsoEdNAcm4YDPIWZNFfm8qlsGNQ+XGSYrl4U1sXNS1KG59hp9wWfykNmyOiA4+mW9OWFbTO2HadvQmmAAA3SoNlUT2uCtD2jcX16fsbwv1fvWfq6dhq1xzg3HMTcdQwAUBUS+IBZ7-AYBATu6VGYkSIL3eOwBB4iCrkAA) that contains my type musings during this process. We might want to revisit these at some point

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -42,156 +45,119 @@ type CommonButtonProps<T> = {
...
-  };
-
-type ButtonButtonProps = CommonButtonProps<'button'> &
-  Omit<React.ComponentPropsWithRef<'button'>, OmitProps>;
-
-export type ButtonProps = CommonButtonProps<ButtonComponent> | LinkButtonProps | ButtonButtonProps;
-
-export class Button extends React.PureComponent<ButtonProps> {
-  static defaultProps = {
-    type: 'button',
```
scheul93, line 74: Not sure if there's an easy way to resolve this in the refactored component, but removing the default `type` for button could create a potential regression. In the case where a button is used inside of a form, the `type` should be specified to be `button` otherwise it will be defaulted to type=submit.

Not sure if this is a scenario in many of our consuming apps.

[From MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#notes)
```
If your buttons are not for submitting form data to a server, be sure to set their type attribute to button. Otherwise they will try to submit form data and to load the (nonexistent) response, possibly destroying the current state of the document.
```

pwolfert:
Moving this from the description to a comment:

> NOTE: I seem to have broken type-checking downstream. With the updated types, everything is permissible, even invalid values like href props that are objects instead of strings. In our own project, errors like that are found (try editing a unit test), but in the downstream app where they were found, I seem to have broken type-checking. We can't consider this solved until type-checking is restored; otherwise this was all for naught and we should just type the darn thing as any.

When I copy our generated typdef files into the downstream project that had the errors, I'm getting the same silent results. I wonder if it has to do with the version of the react types

packages/design-system/src/components/Button/Button.tsx
pwolfert: Ooh, good catch. I mistakenly thought they were optional 😬

pwolfert: For some reason when I try adding the default props externally like 
```tsx
Button.defaultProps = {
  type: 'button',
}
```
I get some cryptic errors downstream in DrawerToggle. I think for functional components it's encouraged to define them as default parameters to the function, but I kind of wanted the default value to show up on the generated props table in our docs so we're transparent about `type` defaulting to `button`. Still fiddling with it...

pwolfert: Pushed https://github.com/CMSgov/design-system/pull/1532/commits/79e3e9d3c192c50c1de7e76f7aa184f3ef316e61

scheul93 (approved):

pwolfert:
About testing this downstream: I discovered that if I explicitly pass the type to the JSX Button like `<Button<'a'> ...`, it will correctly flag bad prop values like objects for `href` and arbitrary extra props. So it doesn't seem to be able to correctly infer. I would at least think, though, that because we omit props like `href` from the `OtherProps`, it should still yell at us for sending non-strings for `href` even if the `component` type is `any`.

pwolfert:
I thought maybe it could have been the conversion to a functional component, but I applied the same changes to the original generated class-based TS definition files in the downstream project, and it still doesn't seem to infer.

pwolfert:
Here's my new theory on downstream type checking. I discovered that if I set a default type for `T` like `Button<T extends ButtonComponentType = 'button'>` (*and* revert back to class based types), it will flag those bad attributes but only because it thinks `component` is `button`. It also yells at me for saying `component="a"` because it thinks it should always be `button`. I also discovered that if I change
```ts
export declare type ButtonProps<T extends ButtonComponentType> = CommonButtonProps<T> & OtherProps<T>;
```
to
```ts
export declare type ButtonProps<T extends ButtonComponentType> = CommonButtonProps<T>// & OtherProps<T>;
```
it will correctly flag bad `href` values.

So here's my theory: TS infers the type of `component` to be ` JSXElementConstructor<any>` or something like that, which would have `any` for `props`.

forrestbaer:
Good discovery work! So the intersection with otherprops was allowing an any type sneak in because component was possibly being inferred as a strange type in that case? It seems like there is some variation in what the ts-lsp tells us with intellisense and what plays out via compilation and runtime use?

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -26,6 +25,10 @@ type CommonButtonProps<T> = {
    */
   component?: T;
   disabled?: boolean;
+  /**
+   * Button [`type`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#attr-type) attribute
```
zarahzachz, line 29: This comment doesn't match the `href` prop beneath, does it?

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -45,153 +48,118 @@ type CommonButtonProps<T> = {
...
-        );
-      }
+// Collect all the additional properties that one could supply to a button component
+// that will be passed down to whatever element or component is being used. This is
+// generally permissive in order to keep the typing simple at the expense of being
+// more accurate. `OtherProps` is generic so that we can extract any props from a
+// custom component that is being passed in. I'm trying to keep most of the complexity
+// in this section and leave the `ButtonProps` definition simpler. - PW
+//
+// Extend is a utility type that works like `Object.assign` where properties defined
```
zarahzachz, line 66: Love this comment

packages/design-system/src/components/Button/Button.tsx
zarahzachz: I think you can pass default props in the params you pass to the component, i.e., 
```js
export const Foo = ({bar, baz = "Default value"}) => {
    // logic
}
```

if `defaultProps` doesn't work in the future.

packages/ds-healthcare-gov/package.json
```diff
@@ -20,7 +20,7 @@
     "src"
   ],
   "dependencies": {
-    "@cmsgov/design-system": "^3.0.0-beta.1",
+    "@cmsgov/design-system": "file:../design-system",
```
zarahzachz, line 23: Should this point here?

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -45,153 +48,118 @@ type CommonButtonProps<T> = {
...
-      className: this.classNames(),
-      ...props,
-    };
-
-    if (this.props.onClick) {
-      attrs.onClick = this.handleClick;
+    if (variation === 'danger') {
+      console.warn(
+        `[Deprecated]: Please remove the 'danger' variation prop in <Button>. This prop has will be removed in a future release.`
+      );
```
zarahzachz, line 103: Is this a good time to yank out these deprecated props?

packages/design-system/src/components/Button/Button.tsx
pwolfert: Whoops! Good catch

packages/ds-healthcare-gov/package.json
pwolfert: Whoops!

packages/design-system/src/components/Button/Button.tsx
pwolfert: It would be, but we haven't discussed it prior to this change

pwolfert:
Update: I'm going to be merging this and then immediately following up with a revert via https://github.com/CMSgov/design-system/pull/1545. I'd like git to retain a history of this, which would be lost in a squashed commit.

packages/design-system/src/components/Button/Button.tsx
pwolfert: If we do, we should make it a separate PR

forrestbaer (approved):

PR #1538, opened 2022-01-24
Simplify Button types

pwolfert:
See notes in code. This needs more thorough testing.

### DIFF
--- a/packages/design-system/src/components/Button/Button.tsx
+++ b/packages/design-system/src/components/Button/Button.tsx
@@ -1,8 +1,7 @@
 import React from 'react';
 import classNames from 'classnames';
 
-export type CustomButtonComponentType = React.ComponentType<any> | React.FC;
-export type ButtonComponentType = React.ElementType<any> | CustomButtonComponentType;
+export type ButtonComponentType = React.ElementType<any> | React.ComponentType<any>;
 export type ButtonSize = 'small' | 'big';
 /**
  * A string corresponding to the button-component variation classes.
--- a/packages/design-system/src/components/Button/Button.tsx
+++ b/packages/design-system/src/components/Button/Button.tsx
@@ -1,6 +1,8 @@
 import React from 'react';
 import classNames from 'classnames';
 
+export type CustomButtonComponentType = React.ComponentType<any> | React.FC;
+export type ButtonComponentType = React.ElementType<any> | CustomButtonComponentType;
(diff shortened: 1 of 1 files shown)


forrestbaer:
I like this! A lot more readable, thank you for the time you put into this.

pwolfert:
The two basic sanity checks I've done in our `create-react-app-typescript` example is to try making `href` an object and to try adding an arbitrary attribute `bob`. Both problems are caught by the type-checker. In our own Button unit tests, omitting the `to` property of the custom `Link` example is also caught. That seems sufficient to me.

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -42,35 +53,25 @@ type CommonButtonProps = {
...
+  // Get all possible HTML attributes and override with any more specific props from
+  // the possibly custom component type `T`
+  Extend<React.HTMLAttributes<HTMLElement>, React.ComponentPropsWithRef<T>>,
+  // And omit any properties that we're defining on our own `CommonButtonProps`
+  keyof CommonButtonProps<T>
+>;
+
+export type ButtonProps<T extends ButtonComponentType> = CommonButtonProps<T> & OtherProps<T>;
 
 export const Button = <T extends ButtonComponentType>(props: ButtonProps<T>) => {
```
forrestbaer, line 76: Could this can be simplified to:

`export const Button = (props: ButtonProps<ButtonComponentType>) => {`

pwolfert: The hope here is to get specific props for the `component` component, like in https://github.com/CMSgov/design-system/blob/pwolfert/fix-button-types-5/packages/design-system/src/components/Button/Button.test.tsx#L53-L60

forrestbaer: I see gotcha!

scheul93 (approved):

forrestbaer (approved):
looks good!

packages/design-system/src/components/Button/Button.tsx
pwolfert: I also just double-checked to make sure the generic type on `CommonPropTypes` was necessary in order to infer the actual type of `component`, and it was. If I remove it, it allows the `Link` unit test to omit the `to` prop.

PR #1537, opened 2022-01-24
[WNMGDS-1307] WHCM badges

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1307)

hcm-badges

Implement a HCM query via React.

## How to test

Test [this link](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS/1307-whcm-badges/core/storybook/?path=/story/components-badge--default) via your favorite VM.

PR #1543, opened 2022-01-25
[NO TICKET] New HCM approach to include SR support

zarahzachz:
## Summary

This work is an enhancement to HCM work done for Alert and Badge (#1516, #1537) to include screen reader support. Refactor removes reliance on JS to toggle queries and instead switches a `<span>` labeling the component variation from not visible to visible.

## Testing

Preview [Badge](http://design-system-demo.s3-website-us-east-1.amazonaws.com/NO-TICKET/HCM-alert-badge/core/storybook/?path=/story/components-badge--default) and [Alert](http://design-system-demo.s3-website-us-east-1.amazonaws.com/NO-TICKET/HCM-alert-badge/core/storybook/?path=/story/components-alert--default) links (building).

### DIFF
--- a/packages/design-system/src/components/Alert/Alert.tsx
+++ b/packages/design-system/src/components/Alert/Alert.tsx
@@ -141,6 +141,12 @@ export class Alert extends React.PureComponent<
   headingId: string;
   eventHeadingText: string;
 
+  a11yLabel = {
+    error: 'Alert',
+    warn: 'Warning',
+    success: 'Success',
+  };
+
   heading(): React.ReactElement | void {
     const { headingLevel, heading } = this.props;
     const Heading = `h${headingLevel}`;
@@ -218,7 +224,19 @@ export class Alert extends React.PureComponent<
       >
         {this.getIcon()}
         <div className="ds-c-alert__body">
-          {this.heading()}
+          {heading ? (
+            <div className="ds-c-alert__header ds-c-alert__heading">
+              <span className="ds-c-alert__a11y-label ds-u-visibility--screen-reader">
+                {variation ? this.a11yLabel[variation] : 'Notice'}:{' '}
+              </span>
+              {this.heading()}
+            </div>
+          ) : (
+            <span className="ds-c-alert__a11y-label ds-u-visibility--screen-reader">
+              {variation ? this.a11yLabel[variation] : 'Notice'}:{' '}
+            </span>
+          )}
+
           {children}
         </div>
       </div>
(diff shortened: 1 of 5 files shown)


packages/design-system/src/styles/components/_Alert.scss
```diff
@@ -51,6 +51,29 @@ $alert-link-color-focus: $color-primary-darkest;
   }
 }
 
+.ds-c-alert__a11y-label {
+  margin-right: 0.5rem;
+
+  /* stylelint-disable */
+  // !important needed to overwrite screen reader utility styles
```
zarahzachz, line 58: It hurt adding these `!important` flags, but because we're using the SR utility class to visually hide the label, there was a specificity issue when targeting this class. And I couldn't target a `<span>` like I did with Badge since theoretically users could insert their own `<span>` as a child - and I don't want to break things for them.

packages/design-system/src/components/Alert/Alert.tsx
```diff
@@ -218,7 +224,19 @@ export class Alert extends React.PureComponent<
...
+              </span>
+              {this.heading()}
+            </div>
+          ) : (
+            <span className="ds-c-alert__a11y-label ds-u-visibility--screen-reader">
+              {variation ? this.a11yLabel[variation] : 'Notice'}:{' '}
+            </span>
+          )}
+
           {children}
```
zarahzachz, line 240: Children will always render as a block element beneath the label. I considered adding some CSS to make the label appear inline with the children, but not knowing what children a user could pass made me reconsider this. I opted to treat the label as a backup heading instead.

packages/design-system/src/styles/components/_Alert.scss
ericwbailey: This is kind of a tangent, but it's why I've personally moved away from declaring `!important` on utility classes. I like the idea of `!important` forcing immutability, which is why I like keeping it on a per-declaration basis.

packages/design-system/src/styles/components/_Badge.scss
```diff
@@ -45,6 +26,22 @@ $badge-font-size-big: $base-font-size;
...
+    @media (-ms-high-contrast: active), (forced-colors: active) {
+      border: none;
+      clip: auto;
+      height: auto;
+      overflow: visible;
+      padding: 0;
+      position: relative;
+      width: auto;
+      // Workaround for https://bugzilla.mozilla.org/show_bug.cgi?id=1241631
+      word-wrap: normal;
```
ericwbailey, line 40: I wonder if we should convert this to a mixin.

ericwbailey (approved):
I tested this on Storybook, using the latest version of NVDA and VoiceOver. I can confirm the announcement of each variation for both screen readers.

Additionally, I tested each component with High Contrast Mode active, and can confirm both the visual presence of each variation's label, as well as still being announced by NVDA.

PR #1552, opened 2022-01-26
NO TICKET - Resolution of several build, start and test warnings

forrestbaer:
## Summary

Please see commit messages for description of changes, a couple items may be jumping the gun but I don't believe any of these are breaking changes.  The only unresolved build/start warning I couldn't resolve was the highlight.js deprecated warning which comes from the KSS package for css documentation, perhaps this will go away with the new docs site anyway as I don't imagine we'd be using KSS there.

## How to test

Run all build/test/start/storybook tasks.. Evaluate code for anything that may look like a breaking change

Attached here is a comparison of the live heathcare.gov site and my local run with the fixes for the privacy settings dialog:

Screen Shot 2022-01-26 at 12 32 50 PM
Screen Shot 2022-01-26 at 12 34 18 PM


yarn.lock
```diff
@@ -7303,20 +7303,10 @@ caniuse-api@^3.0.0:
...
-caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001125:
-  version "1.0.30001278"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001278.tgz#51cafc858df77d966b17f59b5839250b24417fff"
-  integrity sha512-mpF9KeH8u5cMoEmIic/cr7PNS+F5LWBk0t2ekGT60lFf0Wq+n9LspAj0g3P+o7DQhD3sUdlMln4YFAWhFYn9jg==
-
-caniuse-lite@^1.0.30001251, caniuse-lite@^1.0.30001274:
-  version "1.0.30001277"
-  resolved "https://registry.yarnpkg.com/caniuse-lite/-/caniuse-lite-1.0.30001277.tgz#9416dae5e075f47eacd8e0475ae1dcc5a20e9ca5"
-  integrity sha512-J2WtYj2Pl6LBEG214XmbGw1gzZEsYuinQFPqYtpZDB3/vm49qNlrcbJrTMkHKmdRDdmXYwkG0tgOBJsuI+J12Q==
+caniuse-lite@^1.0.0, caniuse-lite@^1.0.30000971, caniuse-lite@^1.0.30001039, caniuse-lite@^1.0.30001109, caniuse-lite@^1.0.30001125, caniuse-lite@^1.0.30001219, caniuse-lite@^1.0.30001251, caniuse-lite@^1.0.30001274:
```
pwolfert, line 7306: Did you manually edit this?

packages/ds-healthcare-gov/src/components/Header/DeConsumerMessage.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import { TFunction } from 'react-i18next';
 
 interface DeConsumerMessageProps {
-  deBrokerName: string;
```
pwolfert, line 5: What about `deBrokerName?: string`

guides/CODING-GUIDELINES.md
```diff
@@ -30,7 +30,7 @@ JavaScript coding guidelines are enforced using [Nava's ESLint ruleset](https://
 - Build components that are safe to render server-side. This means components must be pure functions of their `props` and `state`, and you only have access to a subset of lifecycle methods:
   - `getInitialState()`
   - `getDefaultProps()`
-  - `componentWillMount()`
```
pwolfert, line 33: Did one of our tools catch this, or did you? There are probably some other things in here we could update.

packages/ds-medicare-gov/src/components/NavigationMenu/NavigationMenu.tsx
```diff
@@ -49,7 +50,7 @@ export default class NavigationMenu extends React.PureComponent<
             onClick={this.toggleMenu}
             variation="transparent"
           >
-            {this.state.menuOpen ? <Close /> : <Hamburger />}
```
pwolfert, line 52: 👍

pwolfert: Does this look good to you, @scheul93?

packages/ds-healthcare-gov/src/components/PrivacySettings/PrivacySettingsDialog.tsx
```diff
@@ -39,11 +39,10 @@ export const PrivacySettingsDialog = (props: PrivacySettingsDialogProps) => {
   return (
     <Dialog
       {...dialogProps}
-      title={t('privacy.dialogTitle')}
+      heading={t('privacy.dialogTitle')}
       size="full"
       closeButtonSize="small"
-      // @ts-ignore: TODO: We probably shouldn't be using secondary here, as it's supposed to be deprecated
-      closeButtonVariation="secondary"
+      closeButtonVariation="transparent"
```
pwolfert, line 45: I agree that it's time to rip the bandaid off with this one

pwolfert: We might also want to remove the `closeButtonSize` and let it be the default.

packages/design-system/src/components/Pagination/Ellipses.tsx
```diff
@@ -1,9 +1,11 @@
 import React from 'react';
 
-export default function Ellipses(): React.ReactElement {
+const Ellipses : React.FC<{}> = props => {
```
pwolfert, line 3: Seems odd that we'd have to define an unused `props` parameter in order to please the docs generator. It doesn't work with the type annotation alone and using `() => {`?

pwolfert (commented):
The unit tests are much less noisy!

When I run the linter on master, I get
`(0 errors, 310 warnings)`

When I run the linter on this branch, I get
`(3 errors, 311 warnings)`

Looks like the errors might come from the doc-gen fixes (`error    Don't use `{}` as a type. `{}` actually means "any non-nullish value".`)

forrestbaer:
> Looks like the errors might come from the doc-gen fixes (`error Don't use `{}`as a type.`{}` actually means "any non-nullish value".`)

Changed to 'unknown' resolves the linter errors and the tests from bellyaching.. Looking at the linter warnings, the majority are return types missing, even for void, etc.. I know we've decided on not typing using void in the past, but I do see this warning as useful in the future possibly to avoid returning the wrong types.. What do you think?  I would be ok with typing all return types, including void just to stick to idomatic TS

scheul93 (approved):
All these changes look good. Is there a future plan to remove the sass warning
```
Using / for division is deprecated and will be removed in Dart Sass 2.0.0.
``` 
that occurs when running storybook?

forrestbaer:
> Using / for division is deprecated and will be removed in Dart Sass 2.0.0.

Unfortunately that's not an easy one, updating to use the .div method just doesn't work for many of our uses of it, and upgrading the css package that's complaining about it causes other dependency problems.. I tried it a couple weeks ago and it didn't go anywhere.. Definitely something I want to resolve as well will give it another try sometime soon, likely in a new branch.

PR #1555, opened 2022-01-26
[NO-TICKET] Update version search tool

pwolfert:
## Summary

A limitation in the previous iteration of the version-scanning script was that the GitHub search result excerpts were not long enough to contain the version, so the tool could not report the version in use. This happened for the majority of projects last time I ran it. What I've done here is add a step for fetching the contents of the `package.json` files where they're found and reading the versions from there.

I've also made some organizational changes, like moving its dependencies to our root `devDependencies` and moving the tool into a root `scripts` folder.

### Changed
- Updated the `findDesignVersions` script and aliased it at `yarn scan-versions`. It now reliably finds the versions of the design system used by our dependent projects.

## How to test

Follow the tool's readme instructions to set up a personal access token, and then run `yarn scan-versions`


forrestbaer:
The code looks good, and since it's not really interacting with anything else I don't see the extra dependencies causing any troubles. I will run a full test of it tomorrow. 👍🏻

scripts/findDesignSystemVersions/index.ts
```diff
@@ -0,0 +1,125 @@
...
+import Table from 'cli-table3';
+import chalk from 'chalk';
+import yargs from 'yargs';
+import dotenv from 'dotenv';
+
+const designSystemPackageNames = [
+  '@cmsgov/design-system',
+  '@cmsgov/ds-healthcare-gov',
+  '@cmsgov/ds-medicare-gov',
+  '@cmsgov/design-system-core',
```
scheul93, line 11: Is this the old, DS version? Do we still need to query for that?

Or are we querying to figure out if anyone is still using this?

pwolfert: Yes, that's the old version. The list is pretty long still, so it might still be helpful to keep it around for a while. There's at least one there that I did not expect to see.

scripts/findDesignSystemVersions/README.md
```diff
@@ -0,0 +1,39 @@
+# Find Design System Versions Tool
+
+This tool gathers the versions of our design systems currently being used the CMS-WDS organization on github.cms.gov.
```
zarahzachz, line 3: I think you're missing a word here. Should it read:
> currently being used **by** the CMS-WDS organization

scripts/findDesignSystemVersions/README.md
```diff
@@ -0,0 +1,39 @@
...
+
+You can run this tool from the root of the repository directory with the following command:
+
+```
+yarn scan-versions
+```
+
+## Setup
+
+First, you'll need to create a personal access token through GitHub Enterprise.
```
zarahzachz, line 13: Would it make more sense to [link to the Github docs](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/creating-a-personal-access-token) for this in the instance they update the process/UI for creating a token?

pwolfert: Doh! Yes, good idea

forrestbaer (approved):
Worked well for me, interesting to see taxtools leading the charge on the implementation front. Almost everyone else so far behind!

PR #1561, opened 2022-01-29
[WNMGDS-1401] Make the StepList props with default values optional

pwolfert:
## Summary

StepList props became required to appease TypeScript, but the TypeScript definitions that were generated required those props from components that use `StepList`. We want the props that have default values to be optional to our users.

### Fixed

- Fixed optional `StepList` props that have default values being required

## How to test

1. `yarn type-check`
2. Can also `yarn build` and look at the type output


forrestbaer (approved):
seems ok in my testing

scheul93 (approved):

PR #1570, opened 2022-02-01
[WNMGDS-1330] initial markup, styling and small functionality for IdleTimeout

scheul93:
## Summary
PR pt 1 of the new IdleTimeout component. Description of component found here: https://jira.cms.gov/browse/WNMGDS-1330

As I was building this component, I realized the logic might be a bit complicated, so I wanted to section off the changes into several PRs. 
1. This PR (which will also be the "feature branch" that I will merge subsequent PRs into) which includes markup & styling for the warning modal, setup of Storybook for this component & some unit tests
2. A PR for timeout functionality & related unit tests
3. Remaining documentation and testing

You may noticed that I added some props and changed some prop names from the original ticket. I'm trying to keep this component flexible, but naming may also change as I get to the timeout functionality.

## How to test

Run `yarn storybook` and see the new IdleTimeout component. The current version shows what the warning dialog will look like

### DIFF
--- a/packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
+++ b/packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
@@ -1,0 +1,134 @@
+/* eslint-disable react/no-multi-comp */
+/* eslint-disable react/no-danger */
+import React from 'react';
+import { Dialog } from '../Dialog';
+import { Button } from '../Button';
+
+export interface IdleTimeoutProps {
+  /**
+   * The text for the 'close' button on the warning dialog
+   */
+  closeDialogText?: string;
+  /**
+   * The text for the 'continue session' button in warning dialog.
+   */
+  continueSessionText?: string;
+  /**
+   * The heading text for the warning dialog.
+   */
+  heading?: string;
+  /**
+   * The text for the button that ends the session in warning dialog.
+   */
+  endSessionButtonText?: string;
+  /**
+   * The message text for the warning dialog.
+   * Note that using the token `<timeToTimeout>` will be replaced in the message text with the number of minutes until timeout.
+   */
+  message?: string;
+  /**
+   * Optional function that is called when the warning dialog's close button is clicked
+   */
+  onClose?: (...args: any[]) => any;
+  /**
+   * Optional function that is called when the user chooses to keep the session alive.
+   * The IdleTimeout component will reset the countdown internally.
+   */
+  onSessionContinue?: (...args: any[]) => any;
+  /**
+   * Optional function that is called when the session is manually ended by user.
+   * If not provided, the behavior of `onTimeout` will be used.
+   */
+  onSessionForcedEnd?: (...args: any[]) => any;
+  /**
+   * Function that is called when the timeout countdown reaches zero.
+   */
+  onTimeout: (...args: any[]) => any;
+  /**
+   * Describes if the button to manually end session should be shown
+   */
+  showSessionEndButton?: boolean;
+  /**
+   * Defines the time (in minutes) until the session is timed out
+   */
+  timeToTimeout: number;
+  /**
+   * Defines the time (in minutes) until the warning message is shown. The default is 5 minutes.
+   */
+  timeToWarning?: number;
+}
+
+const IdleTimeout = ({
+  closeDialogText = 'Close',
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -0,0 +1,134 @@
...
+  heading?: string;
+  /**
+   * The text for the button that ends the session in warning dialog.
+   */
+  endSessionButtonText?: string;
+  /**
+   * The message text for the warning dialog.
+   * Note that using the token `<timeToTimeout>` will be replaced in the message text with the number of minutes until timeout.
+   */
+  message?: string;
```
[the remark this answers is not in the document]
forrestbaer: It could possibly be a function that takes a number and returns a string `message?: (t : number) => string;` with that number available to interpolate then down below it could be `div>{{message(timeToTimeout)}}</div>` which might be a nice way to present it and would allow the user to change that time to minutes:seconds, or just seconds or whatever they wanted..

forrestbaer:
Looks pretty good! I noticed type-errors during build but am guessing those will be addressed, made a comment about message.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -0,0 +1,134 @@
...
+   */
+  timeToWarning?: number;
+}
+
+const IdleTimeout = ({
+  closeDialogText = 'Close',
+  continueSessionText = 'Continue session',
+  heading = 'Are you still there?',
+  endSessionButtonText = 'Logout',
+  message = 'You’ve been inactive for a while. <br/>Your session will end in <timeToTimeout>. <br/><br/>Select "Continue session" below if you want more time.',
```
zarahzachz, line 66: This is clever, but I think our DS has styles (spacing) associated with true paragraphs that aren't available when `<br>` tags are used instead. To your comment above, you might need to make this a React.Node type to ensure our styles are applied to whatever message is passed - or we might need to add more robust styles on our end to account for using `<br>` tags

packages/design-system/src/components/IdleTimeout/IdleTimeoutDialog.tsx
```diff
@@ -0,0 +1,83 @@
...
+  /**
+   * The message text for the warning dialog.
+   * Note that using the token `<timeToTimeout>` will be replaced in the message text with the number of minutes until timeout.
+   */
+  message: string | React.ReactNode;
+  /**
+   * Optional function that is called when the user chooses to keep the session alive.
+   * The IdleTimeout component will reset the countdown internally.
+   */
+  onSessionContinue: (...args: any[]) => any;
```
zarahzachz, line 32: Is this optional? Not seeing `?` set.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -0,0 +1,233 @@
...
+  showSessionEndButton = false,
+  timeToTimeout,
+  timeToWarning,
+}: IdleTimeoutProps): JSX.Element => {
+  if (timeToWarning > timeToTimeout) {
+    console.error(
+      'Error in TimeoutManager component. `timeToWarning` is greater or equal to `timeToTimeout`'
+    );
+  }
+  const MS_BETWEEN_STATUS_CHECKS = 30000;
```
forrestbaer, line 103: small thing, the formatting on this const is different from the others.

packages/design-system/src/components/IdleTimeout/useInterval.js
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useRef } from 'react';
+
+// https://overreacted.io/making-setinterval-declarative-with-react-hooks/
```
forrestbaer, line 3: For when this disappears some day, is there any way usage could be synopsized in 1 to 2 lines?  😰

packages/design-system/src/components/IdleTimeout/IdleTimeoutDialog.tsx
scheul93: Ope! Not optional, I'll update the documentation

PR #1584, opened 2022-02-08
[WNMGDS-1330] countdown with storage

scheul93:
## Summary

Re-implementing the countdown functionality with local storage instead of settimeouts. This also includes unit tests and updates to the Dialog component to make the 'Close' optional

## How to test

Run `yarn storybook` to see the component in action. Run `yarn test` to run unit tests


packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -77,70 +78,148 @@ const defaultMessageFormatter = (timeTilTimeout: number): React.ReactNode => {
...
+  if (timeToWarning > timeToTimeout) {
+    console.error(
+      'Error in TimeoutManager component. `timeToWarning` is greater or equal to `timeToTimeout`'
+    );
+  }
+  // convert minutes to milliseconds
+  const msToTimeout = timeToTimeout * 60000;
+  const msToWarning = (timeToTimeout - timeToWarning) * 60000;
+  const [checkStatusTime, setCheckStatusTime] = useState<number>(null);
+  const [warningIntervalId, setWarningIntervalId] = useState<NodeJS.Timeout>(null);
```
scheul93, line 107: > I found this (https://stackoverflow.com/a/56239226) when searching for NodeJS.Timeout, seems like it may be better typing for this for platform independence?

From a different PR -- wanted to capture it here

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -21,15 +17,15 @@ export interface IdleTimeoutProps {
    * The text for the button that ends the session in warning dialog.
    */
   endSessionButtonText?: string;
+  /**
+   *
+   */
+  endSessionRedirectUrl?: string;
```
pwolfert, line 23: Does this need to say "redirect" if it's actually only used for a link href?

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -77,70 +78,148 @@ const defaultMessageFormatter = (timeTilTimeout: number): React.ReactNode => {
...
+// names for local storage variables
+const timeoutCookieName = 'CMS_DS_TIMEOUT';
+const timeoutWarningCookieName = 'CMS_DS_WARNING';
+
 const IdleTimeout = ({
-  closeDialogText = 'Close',
   continueSessionText = 'Continue session',
   heading = 'Are you still there?',
   endSessionButtonText = 'Logout',
+  endSessionRedirectUrl = '/logout',
```
pwolfert, line 89: Did you say you were going to remove the default value for this?

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
pwolfert: I think it would be good to use the return type, since it adapts to any environment and always stays accurate. I'm pretty sure I've actually used that method before. Could have even seen the same SO answer.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -77,70 +78,148 @@ const defaultMessageFormatter = (timeTilTimeout: number): React.ReactNode => {
...
+  };
+
+  const removeEventListeners = () => {
+    document.removeEventListener('mousemove', resetTimeouts);
+    document.removeEventListener('keypress', resetTimeouts);
+  };
+
+  const addEventListeners = () => {
+    document.addEventListener('mousemove', resetTimeouts);
+    document.addEventListener('keypress', resetTimeouts);
```
pwolfert, line 163: Did you look into the `passive` option? https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#parameters

forrestbaer:
One small issue during build: 
```
[11:08:07] react-docgen Error: No suitable component definition found.
[11:08:07] react-docgen /Users/forrestbaer/code/design-system/packages/design-system/src/components/IdleTimeout/useInterval.jsx
```
Happens during doc build, guess there needs to be a new doc page for this as well

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -77,70 +78,148 @@ const defaultMessageFormatter = (timeTilTimeout: number): React.ReactNode => {
...
+  };
+
+  const setTimeoutCookies = () => {
+    const timeoutTime = Date.now() + msToTimeout;
+    const warningTime = Date.now() + msToWarning;
+    localStorage.setItem(timeoutCookieName, timeoutTime.toString());
+    localStorage.setItem(timeoutWarningCookieName, warningTime.toString());
+
+    if (checkStatusTime === null) {
+      setCheckStatusTime(30000);
```
pwolfert, line 147: Can we pull this out into a constant?

forrestbaer:
May want to disable the component on the docs page with a [docs block](https://storybook.js.org/docs/react/writing-docs/docs-page#remixing-docspage-using-doc-blocks)

packages/design-system/src/components/IdleTimeout/useInterval.jsx
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useRef } from 'react';
+
+// https://overreacted.io/making-setinterval-declarative-with-react-hooks/
```
forrestbaer, line 3: Interesting read!

forrestbaer:
Looks good outside my two notes, component testing in storybook worked well though not a lot of open controls. Since the close time is effectively modifiable via `useTimeout` might be nice to have that as editable.. Regardless looks good, I envy your ability to write good tests!

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -77,70 +78,148 @@ const defaultMessageFormatter = (timeTilTimeout: number): React.ReactNode => {
...
+}: IdleTimeoutProps): JSX.Element => {
+  if (timeToWarning > timeToTimeout) {
+    console.error(
+      'Error in TimeoutManager component. `timeToWarning` is greater or equal to `timeToTimeout`'
+    );
+  }
+  // convert minutes to milliseconds
+  const msToTimeout = timeToTimeout * 60000;
+  const msToWarning = (timeToTimeout - timeToWarning) * 60000;
+  const [checkStatusTime, setCheckStatusTime] = useState<number>(null);
```
pwolfert, line 106: Why does this value to be saved in local state?

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -77,70 +78,148 @@ const defaultMessageFormatter = (timeTilTimeout: number): React.ReactNode => {
...
+  // when it's time to warn the user about idleness,
+  // set an interval that updates the modal message
+  const handleWarningTimeout = () => {
+    setShowWarning(true);
+    removeEventListeners();
+    let timeTilTimeout = timeToWarning - 1;
+    const intervalId = setInterval(() => {
+      setTimeInWarning(timeTilTimeout);
+      timeTilTimeout--;
+    }, 60000);
```
pwolfert, line 136: Could we simplify by actually using the same interval as the `checkWarningStatus` function? Like in that function can we `setTimeInWarning` based on subtracting the `warningTime` in that function from the current time?

pwolfert: I think then the `timeInWarning` could be derrived from the current time and the state we already track.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
scheul93: I'm not sure. I removed redirecting to logout as default behavior for the `onTimeout`, but for the end session url, I kind of assumed that the 'end session' option would only be used in logged in experiences?

That is an assumption though, because I couldn't think through what ending a session on an unauthenticated page would lead to

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
scheul93: I just realized I forgot about this after our conversation 🤦🏻 
The only thing that made me hesitate is that `passive` isn't supported in IE and I don't know if it's through a runtime error in IE if it's present in the code, so I may have to check if there's a way to test if a browser supports it and set that option conditionally. Thoughts?

scheul93:
> One small issue during build:
> 
> ```
> [11:08:07] react-docgen Error: No suitable component definition found.
> [11:08:07] react-docgen /Users/forrestbaer/code/design-system/packages/design-system/src/components/IdleTimeout/useInterval.jsx
> ```
> 
> Happens during doc build, guess there needs to be a new doc page for this as well

I'm separating the new doc page into a separate PR, and will address this there!

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
scheul93: I wanted to clear the checkStatus interval when the timeout occurs and using the custom hook, that happens by changing the interval time from a number to null.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
pwolfert: I was thinking when I said this that the presence of this value (OR the on-click handler) controlled whether that button got shown, but I realize now that there's a separate flag for that. The front-end-only (unauthenticated) apps that I know of that currently have one of these timeouts clears local browser storage on timeout, FYI.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
pwolfert: That docs page shows how to test for it. In the absense of any other optimizations like debouncing, it'd probably be better to have this kind of optimization for 99% of our users rather than none of our users. It would be annoying to have to test for support, but at least we could hide away that code in another module.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
scheul93: This is no longer needed since we aren't keeping track of any interval ids with the refactor!
But I will keep this in mind for the future. TIL 🌈

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
pwolfert: 👍 and it's being cleared before component destruction so that if app devs want to show some other messaging this modal doesn't pop up again? It would only be useful for unauthentiicated apps, right? I actually think in the example front-end-only app that we've been looking at, it re-routes to the landing page that doesn't render the idle timeout (because the end user hasn't entered any sensitive information yet). Maybe we can actually just let this clear on component destruction like `useInterval` does and leave the ball in product teams' court to handle what happens after a timeout.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -77,70 +79,146 @@ const defaultMessageFormatter = (timeTilTimeout: number): React.ReactNode => {
...
+    const now = Date.now();
+
+    if (now >= Number(timeoutTime)) {
+      handleTimeout();
+    } else if (!showWarning && now >= warningTime) {
+      handleWarningTimeout();
+    } else if (showWarning && now >= warningTime) {
+      // if the warning is showing, update the timeInWarning variable (in minutes)
+      const minutesLeft = Math.ceil((timeoutTime - now) / 60000);
+      setTimeInWarning(minutesLeft);
```
pwolfert, line 172: Doing it in this block may mean we don't have an initial value after `handleWarningTimeout()`, if I'm thinking about it correctly. If we moved this `minutesLeft` calculation into the main function body and provided it directly to the dialog message format functiion, we wouldn't even need to store `timeInWarning` in local state. I could be wrong

pwolfert: Another thing I just thought about in this branching logic here: Do we handle the case where another tab updates the `localStorage` values. Is there a way this component will read an updated value and _hide_ the warning modal? That is, this tab is starting to time out, but in another tab the end user is working away and updating `localStorage` with fresh timeout times. I haven't actually tested it yet in my browser

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -77,70 +79,146 @@ const defaultMessageFormatter = (timeTilTimeout: number): React.ReactNode => {
...
+  const handleWarningTimeout = () => {
+    setShowWarning(true);
+    removeEventListeners();
+  };
+
+  const setTimeoutCookies = () => {
+    const timeoutTime = Date.now() + msToTimeout;
+    const warningTime = Date.now() + msToWarning;
+    localStorage.setItem(timeoutCookieName, timeoutTime.toString());
+    localStorage.setItem(timeoutWarningCookieName, warningTime.toString());
```
pwolfert, line 136: If we start saving the _last active_ timestamp instead of the timeout and warning time, I think we'd only need to store and update one value. We can then calculate the difference between now and that last active time on intervals and see if it's greater than the warning time or greater than the timeout time.

packages/design-system/src/components/IdleTimeout/useInterval.js
```diff
@@ -0,0 +1,24 @@
+import { useEffect, useRef } from 'react';
+
+// https://overreacted.io/making-setinterval-declarative-with-react-hooks/
+function useInterval(callback, delay) {
```
pwolfert, line 4: Nice 👌 . I actually missed that you were adding this yourself the first time. It's a nice hook!

pwolfert (commented):
It's looking great! I have another round of thoughts on simplifying.

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
scheul93: >Doing it in this block may mean we don't have an initial value after handleWarningTimeout(), if I'm thinking about it correctly. If we moved this minutesLeft calculation into the main function body and provided it directly to the dialog message format functiion, we wouldn't even need to store timeInWarning in local state. I could be wrong

There is an initial value after `handleWarningTimeout` because the state variable is initialized with a value (`timeToWarning`)

Aside from not having a state variable, I'm not sure how this change would be much different. As I'm implementing it, in order to have all the information to calculate minutes left, I would have to do similar calculations in another function that this function is already doing and wouldn't it be performing the calculation for `minutesLeft` on every render?

forrestbaer (approved):
nice work!  I can't find anything else to really improve on now, gotta see how it fares in the wild 🌈

PR #1590, opened 2022-02-11
[WNMGDS-1415] focusing drawer toggle only when drawer goes from open to close

scheul93:
## Summary

Addressing reported bug where the drawer toggle is focused whenever the drawer is not open.

### Fixed

Added a custom hook that will keep track of the "previous" value of a prop so that can be used as a comparison point later in the component. It is similar to the `previousProps` param that the `componentDidUpdate` lifecycle had.

`usePrevious` was taken from [this helpful StackOverflow](https://stackoverflow.com/questions/53446020/how-to-compare-oldvalues-and-newvalues-on-react-hooks-useeffect)

I also updated the unit tests to better capture the behavior of the toggle and test for the reported bug

## How to test

1. `yarn storybook`
2. Navigate to the Drawer > Drawer Toggle with Drawer story
3. Confirm that on page load, the toggle does not have focus
4. Open the drawer & close the drawer. Confirm that on close, the drawer toggle gets focus

### DIFF
--- a/packages/design-system/src/components/Drawer/DrawerToggle.test.tsx
+++ b/packages/design-system/src/components/Drawer/DrawerToggle.test.tsx
@@ -53,7 +53,7 @@ describe('DrawerToggle', () => {
     expect(toggle.getAttribute('aria-label')).toBe(ariaLabel);
   });
 
-  it('focuses button when drawer is closed', () => {
+  it('focuses button when drawer goes from open to closed', () => {
     const { rerender } = renderDrawerToggle({ drawerOpen: true });
     rerender(
       <DrawerToggle {...defaultProps} drawerOpen={false}>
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/Drawer/DrawerToggle.test.tsx
```diff
@@ -63,4 +63,10 @@ describe('DrawerToggle', () => {
     const toggle = screen.getByRole('button');
     expect(toggle).toEqual(document.activeElement);
   });
+
+  it('does not focus button when drawer in initialized to close', () => {
+    renderDrawerToggle({ drawerOpen: false });
+    const toggle = screen.getByRole('button');
+    expect(toggle).not.toEqual(document.activeElement);
+  });
```
zarahzachz, line 71: Love the addition of this test

packages/design-system/src/components/Drawer/usePrevious.ts
```diff
@@ -0,0 +1,13 @@
...
+// similar to the old previousProps param from `componentDidUpdate`
+const usePrevious = <T>(value: T): T | undefined => {
+  const ref = useRef<T>();
+  useEffect(() => {
+    ref.current = value;
+  });
+  return ref.current;
+};
+
+export default usePrevious;
```
zarahzachz, line 13: I wonder if this custom hook should be moved to a new `/hooks` directory - it might be useful if we want to use it in any other components that need some kind of `componentdidupdate(prevprops)` equivalent.

zarahzachz (approved):
Worked like a charm. I had a small comment unrelated to functionality, so merge when you'd like

PR #1597, opened 2022-02-15
[WNMGDS-1330] warning time refactor

scheul93:
## Summary

Some updates to the IdleTimeout component

### Added

* an effect that updates the time-in-message as well as the timeout cookie if the `timeToWarning` or `timeToTimeout` prop changes
* unit tests to support changes

### Changed

* IdleTimeout story has some additional text to provide context for how the component works & a short time until warning modal is shown
* The `timeToWarning` prop so now its value represents how many minutes (from current time) of idle activity will trigger the warning modal
*   it used to be relative to the timeout time

### Fixed

* IdleTimeout story so now the component gets updated args properly

## How to test

`yarn storybook` and navigate to the idle timeout component

### DIFF
--- a/packages/design-system/src/components/IdleTimeout/IdleTimeout.test.tsx
+++ b/packages/design-system/src/components/IdleTimeout/IdleTimeout.test.tsx
@@ -5,9 +5,8 @@ import { render, fireEvent } from '@testing-library/react';
 import { mockTime, restoreTime } from './utilities/mockTime';
 
 describe('Idle Timeout', () => {
+  const MOCK_START_TIME = 1643811720; // setting starting date time to 2/22/2022 2:22
   const ADVANCE_TIMER_MS = 30000;
-  const WARNING_DATETIME = 1643991720; // start time + (3 * 60000ms) (3 minutes)
-  const TIMEOUT_DATETIME = 1644111720; // start time + (5 * 60000ms) (3 minutes)
   const onTimeout = jest.fn();
   const defaultProps = {
     timeToTimeout: 5,
@@ -15,6 +14,8 @@ describe('Idle Timeout', () => {
     onTimeout,
   };
   const timeTilWarningShown = defaultProps.timeToWarning * 60000;
+  const WARNING_DATETIME = MOCK_START_TIME + defaultProps.timeToWarning * 60000; // date time when warning should appear
+  const TIMEOUT_DATETIME = MOCK_START_TIME + defaultProps.timeToTimeout * 60000; // date time when timeout should occur
 
   const renderIdleTimeout = (overrideProps?) => {
     return render(<IdleTimeout {...defaultProps} {...overrideProps} />);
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -186,6 +186,11 @@ const IdleTimeout = ({
     };
   }, []);
 
+  useEffect(() => {
```
scheul93, line 189: I tried adding the props as dependencies to the `useCallback` of `resetTimeouts`, but it doesn't work how useEffect works. The `resetTimeouts` function would only trigger on re-render and I'm not sure why... so this is my workaround

Only downside here is that this effect runs on mount so technically `setTimeoutCookies` gets called twice. Once in the above effect & once here

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
```diff
@@ -103,10 +103,10 @@ const IdleTimeout = ({
   const MS_BETWEEN_STATUS_CHECKS = 30000;
   // convert minutes to milliseconds
   const msToTimeout = timeToTimeout * 60000;
-  const msToWarning = (timeToTimeout - timeToWarning) * 60000;
+  const msToWarning = timeToWarning * 60000;
   const [checkStatusTime, setCheckStatusTime] = useState<number>(null);
   const [showWarning, setShowWarning] = useState<boolean>(false);
-  const [timeInWarning, setTimeInWarning] = useState<number>(timeToWarning);
+  const [timeInWarning, setTimeInWarning] = useState<number>(timeToTimeout - timeToWarning);
```
pwolfert, line 109: I think we need to use the same ceiling function in all the places where we set this `timeInWarning`, because in the story, I see `1.5 minutes` in the message at first.

packages/design-system/src/components/IdleTimeout/IdleTimeout.test.tsx
```diff
@@ -118,18 +118,46 @@ describe('Idle Timeout', () => {
...
-      `You've been inactive for a while.Your session will end in 3 minutes.Select "Continue session" below if you want more time.`
+      `You've been inactive for a while.Your session will end in 2 minutes.Select "Continue session" below if you want more time.`
     );
   });
 
   it('default formatMessage should adjust message for singular minute vs multiple', () => {
-    const { getByRole } = renderIdleTimeout({ timeToWarning: 1 });
-    showWarning(1644051720);
+    const { getByRole } = renderIdleTimeout({ timeToWarning: 4 });
+    showWarning(1644051720); // start time + (4 * 60000ms) (4 minutes)
```
pwolfert, line 160: it might make this easier to follow if we actually used that math here instead of this raw number

packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx
pwolfert: I think that's fine for now. This is kind of a complex little component we've got here, and maybe over time we'll see ways to simplify it. For now, some comments could help a lot to explain the different useEffects and such

forrestbaer (approved):
This looks good to me!

PR #1603, opened 2022-02-16
[WNMGDS-1330] idle timeout documentation

scheul93:
## Summary

Documentation for the Idle Timeout component.

## How to test

If you want to see in use, run `yarn storybook` and navigate to Components > Idle Timeout


packages/design-system-docs/src/pages/components/IdleTimeout/_IdleTimeout.docs.scss
```diff
@@ -0,0 +1,31 @@
...
+---
+
+### Related patterns
+
+- [Dialog]({{root}}/components/dialog)
+
+### When to use
+
+- Use the IdleTimeout for authenticated sessions to force logout after a set amount of time
+
```
scheul93, line 29: I'm not sure what information is helpful here since this component is for a pretty specific use case

packages/design-system-docs/src/pages/components/IdleTimeout/_IdleTimeout.docs.scss
```diff
@@ -0,0 +1,31 @@
...
+Style guide: components.idleTimeout
+*/
+
+/*
+`<IdleTimeout>`
+
+@react-example IdleTimeout.example.jsx
+
+@react-props IdleTimeout.tsx
```
pwolfert, line 14: This type value is weird, but I don't think there's much we can do about it 😕 

Screen Shot 2022-02-16 at 4 00 40 PM

pwolfert (approved):
I have a couple suggestions, but other than that, this LGTM 👍

PR #1625, opened 2022-03-02
[WNMGDS-1443] updating tooltip to support close button

scheul93:
## Summary

In order to support some application teams, updating the tooltip to support an optional close button.

## How to test

`yarn storybook`. Navigate to Components > Tooltip > TooltipWithClose


packages/design-system/src/components/Tooltip/Tooltip.stories.jsx
```diff
@@ -80,3 +80,17 @@ InteractiveContent.args = {
   ),
   children: 'Tooltip with interactive content',
 };
+
+export const TooltipWithCloseButton = Template.bind({});
+TooltipWithCloseButton.args = {
+  dialog: true,
+  title: (
```
pwolfert, line 87: For a while I was thinking the new focus code wasn't working, but I just didn't notice the focus state. Can you add `className: 'ds-c-button',` on this Story as well?

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -160,6 +182,13 @@ export const Tooltip = (props: TooltipProps) => {
       props.onOpen && props.onOpen();
     } else {
       props.onClose && props.onClose();
+
+      // if tooltip goes from active to inactive and is the dialog version, focus the trigger
+      if (prevActiveStateVar && (props.dialog || isMobile) && props.showCloseButton) {
+        if (triggerElement && triggerElement.current) {
+          triggerElement.current.focus();
+        }
```
pwolfert, line 190:
```suggestion
        triggerElement?.current?.focus();
```

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -160,6 +182,13 @@ export const Tooltip = (props: TooltipProps) => {
       props.onOpen && props.onOpen();
     } else {
       props.onClose && props.onClose();
+
+      // if tooltip goes from active to inactive and is the dialog version, focus the trigger
+      if (prevActiveStateVar && (props.dialog || isMobile) && props.showCloseButton) {
```
pwolfert, line 187: Does the `props.showCloseButton` part matter?

scheul93: I guess I did that because _technically_ a tooltip could be `dialog='true'` but not have a close button. So I'm not sure if the trigger should be focused on when the user clicks elsewhere in the document

pwolfert: That's fair. I doubt anyone will complain about those two behaviors being inconsistent because no product will have two different configurations side by side.

packages/design-system/src/components/Tooltip/Tooltip.tsx
scheul93: eslint is not a fan
> Expected an assignment or function call and instead saw an expression.

packages/design-system/src/components/Tooltip/Tooltip.test.jsx
```diff
@@ -11,35 +11,108 @@ const defaultProps = {
...
+      const { getByLabelText } = renderTooltip({
+        dialog: true,
+        showCloseButton: true,
+        ariaLabel: 'tooltip trigger',
+      });
+      const tooltipTrigger = getByLabelText('tooltip trigger');
+      fireEvent.click(tooltipTrigger);
+      const closeButton = getByLabelText('Close', { selector: 'button' });
+      fireEvent.click(closeButton);
+      expect(tooltipTrigger).toEqual(document.activeElement);
```
pwolfert, line 105: Hey, this is cool! Didn't know we could test that

packages/design-system/src/components/Tooltip/Tooltip.test.jsx
```diff
@@ -11,35 +11,108 @@ const defaultProps = {
...
+        dialog: true,
+        showCloseButton: true,
+        ariaLabel: 'tooltip trigger',
+      });
+      const tooltipTrigger = getByLabelText('tooltip trigger');
+      fireEvent.click(tooltipTrigger);
+      const closeButton = getByLabelText('Close', { selector: 'button' });
+      fireEvent.click(closeButton);
+      const tooltipContent = queryByRole('dialog');
+      expect(tooltipContent).toBeNull();
```
pwolfert, line 92: Nice 👍

zarahzachz (approved):

packages/design-system/src/components/Tooltip/Tooltip.test.jsx
```diff
@@ -11,35 +11,108 @@ const defaultProps = {
...
     });
-    expect(tooltip.wrapper).toMatchSnapshot();
+    expect(asFragment()).toMatchSnapshot();
   });
 
   it('renders dialog tooltip', () => {
-    const tooltip = render({ dialog: true });
-    expect(tooltip.wrapper).toMatchSnapshot();
+    const { asFragment } = renderTooltip({ dialog: true });
+    expect(asFragment()).toMatchSnapshot();
```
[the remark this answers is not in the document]
scheul93: Updated the broader snapshot tests to be more focused

packages/design-system/src/components/Tooltip/__snapshots__/Tooltip.test.jsx.snap
```diff
@@ -1,245 +1,158 @@
...
     </div>
-  </CSSTransition>
-</div>
+  </div>
+</DocumentFragment>
 `;
+
+exports[`Tooltip tooltip with close renders heading element 1`] = `null`;
+
+exports[`Tooltip tooltip with close renders heading element and close button 1`] = `null`;
```
pwolfert, line 158: I don't think these are supposed to be null

pwolfert (approved):
Thanks for making those snapshots succint 👍 

I saw a possible problem below, but other than that LGTM

PR #1626, opened 2022-03-02
[WIP] [WNMGDS-1462] Create init web component sandbox

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1462)

This PR is an initial attempt at seeing what wrapping our React components using the web components API would look like.

### References

While I've found several projects using web components within React, it's been more difficult finding examples of the inverse. I relied heavily on [this post by Gil Fink](https://gilfink.medium.com/wrapping-react-components-inside-custom-elements-97431d1155bd) to provide some direction on how it would look (with some liberties taken).

[This page does a good job documenting the nitty-gritty of a web component](https://developers.google.com/web/fundamentals/web-components/customelements). I used it to get more info on the `constructor()` and lifecycle methods.

Figuring out [how to bundle styles](https://usefulangle.com/post/370/custom-elements-external-cs-stylesheet) is a question I have. In the web components I'd worked in, each component had their own stylesheet that was imported into each web component, but our styles are bundled into one. As a result, we're going to have a lot of unneeded CSS bundled in each component and that makes me a sad panda.

## Testing

Since I imagine this will be a package we eventually export, I put all this work in its own web component directory (`/ds-web-components`). `cd` into that directory and run `npm install && npm run dev` to see the sandbox in the browser. (Setting up Storybook is still a TODO.)

Fonts and images haven't been added to the project, so it looks a little wonky.


packages/ds-web-components/components/Alert.ts
```diff
@@ -0,0 +1,41 @@
...
+    // Sets Shadow DOM to open or closed
+    const shadowRoot = this.attachShadow({ mode: 'open' });
+    shadowRoot.appendChild(this.mountPoint);
+
+    // I think because React components and CSS are separate in DS,
+    // styles don't fully load when component initializes
+    // Need to bundle styles with each WC - bloat (unless we split up our component styles)
+    const style = document.createElement('link');
+    style.setAttribute('rel', 'stylesheet');
+    style.setAttribute('href', './node_modules/@cmsgov/design-system/dist/css/index.css');
```
forrestbaer, line 30: Interesting.. With the proposed Idea of having component styles available by theme it shouldn't be too hard to break component styles out individually into their own files in the future.

forrestbaer:
The result of `npm run build` is less than awesome.. The 242k bundle and the 140k css plus all the fonts make this a 1mb component.  I wonder if Preact is even an option here, and obviously including a small set of css for just this component would reduce the size as well.

zarahzachz:
> The result of `npm run build` is less than awesome.

Yeah, that's what I'm feeling too. Especially considering these component will ultimately be consumed by other JS frameworks. Just seems like a lot of bloat from the jump. Maybe there's ways to optimize the _very_ rough prototype more? But to bundle React + styles in each component is a lot of a lot.

Right now, I think if we wanted to use web components, we'd need to invert this experiment. Release web components as our primary library and use the different JS flavors as wrappers.

forrestbaer:
> Right now, I think if we wanted to use web components, we'd need to invert this experiment. Release web components as our primary library and use the different JS flavors as wrappers.

Definitely an undertaking, but a good idea. Preact is only 3k and with the CSS being very specific and not all the fonts being included, I'm thinking we could actually reduce these to around 20-100k.

PR #1631, opened 2022-03-04
[WNMGDS-1475] Add release script undo feature and pass args through to lerna command

pwolfert:
## Summary
https://jira.cms.gov/browse/WNMGDS-1475

Someone had the idea to make an undo feature, so I added that. I also wanted to be able to pass arbitrary arguments through to the `lerna version` command. Any unknown args will be passed through to that command now. Specific handling of the `--force-publish` command has been removed in favor of that pass-through.

## Changed
- Pass unknown args to `lerna version`
- Added an `--undo` (or `-u`) flag to undo last release (must be on release branch)
- Updated the next-steps instructions at the end of the script
- Ignore changes to the `design-system-tokens` package for now, because we're not ready to publish it

## How to test

1. Use `yarn release` to create a release and then immediately run `yarn release --undo` to undo it.
2. Instead of `yarn release --force-publish @cmsgov/design-system,@cmsgov/design-system-docs,@cmsgov/design-system-scripts,@cmsgov/ds-healthcare-gov,@cmsgov/ds-medicare-gov`, use an `=` between `--force-publish` and its arguments, [like `lerna version` expects](https://github.com/lerna/lerna/tree/main/commands/version#--force-publish)


forrestbaer:
Awesome! I actually wrote a ticket for this : https://jira.cms.gov/browse/WNMGDS-1475

Feel like doing any of the other items on there as well? I could jump in to this branch and finish the doc cleanup if you want to move onto something else.

forrestbaer (approved):
This works great, thank you, nice work! 💯

PR #1637, opened 2022-03-07
[WNMGDS-1451] and (WNMGDS-1362) Token plugin updates and New Token Types

forrestbaer:
## Description

I fit both [WNMGDS-1362](https://jira.cms.gov/browse/WNMGDS-1362) and [WNMGDS-1451](https://jira.cms.gov/browse/WNMGDS-1451) in here because the items in 1451 actually require we have one of the components with variable level style definitions be completed before it's finished.

For 1451 I've added the layer and text style API's and got the skeleton for the import plugin ready when we have the first component styles completed (Buttons). That work will continue in future tickets to be created for every component, so a part of those tickets will be to ensure those variables can be imported with the plugin.

The new token styles have been created here as well (1362), and some will need to be adjusted as we work through getting the components ready, I've added some additional values where I thought we were a bit light based on the [SalesForce Lightning token system](https://www.lightningdesignsystem.com/design-tokens) as I thought they had some good ideas.

As this is a moving target I don't have any real tests for you to run here, but look forward to any code comments you may have. This ticket does not fundamentally change anything currently implemented but only adds new styles and cleans up some of the code/functionality from the last Token PR.

`yarn build [tokens | themes] [json | scss | csv]` will give expected output in /dist


packages/design-system-tokens/src/lib/utility.ts
```diff
@@ -1,13 +1,127 @@
...
 export const rgbToHex = (r: number, g: number, b: number): HexValue => {
   const hex = ((1 << 24) + (r << 16) + (g << 8) + b).toString(16).slice(1);
   return `#${hex}`;
 };
 
+/* Table of Alpha Values for 10 bit Hex
+ *
+ * Alpha% 	Hex	Num
+ * ----------------
+ * 100%	    FF	255
```
pwolfert, line 15: If we were supplying this value as hex, this table would be useful, but aren't we only ever supplying the alpha value as a percentage?

packages/design-system-tokens/src/lib/exportScss.ts
```diff
@@ -18,32 +26,25 @@ export const exportScss = (fileDescriptors: FileDescriptor[], outPath: string):
...
-      Object.entries(importedModule.default).forEach(([name, value]) => {
-        vars += tokenFormat(name, value);
-      });
-
-      writeFile(filename, vars);
+      const tokens = flatten(importedModule.default, '', {}, '-', true);
+      output = setVars(tokens);
     }
+
+    writeFile(filename, output);
```
pwolfert, line 47: Is there anything I can test related to this change?

pwolfert: `yarn build tokens scss` and `yarn build themes scss`?

packages/design-system-tokens/src/lib/utility.ts
```diff
@@ -20,6 +134,28 @@ export const hexToRgb = (hex: HexValue): RGBValue | null => {
...
+    if (typeof val === 'object') {
+      if (prefixCategory) {
+        flatten(val, `${k}${delim}`, r, delim);
+      } else {
+        flatten(val, ``, r, delim);
+      }
+    } else {
+      res[k] = val;
+    }
+    return r;
```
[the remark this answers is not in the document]
forrestbaer: Ended up doing a big refactor to remove this complexity by changing token naming to be more concise.. this was originally set up as a place to do many things when it should just be doing one, good call out.

packages/design-system-tokens/src/themes/core/defaultTheme.ts
```diff
@@ -95,14 +95,19 @@ const DefaultTheme = makeTheme({
...
-  ],
+  components: {
+    'button-style-example': color.white,
+    'button-spacing-top': spacing['spacer-2'],
+    'dropdown-padding-left': spacing['spacer-4'],
+  },
+  shadows: {}
 });
 
+DefaultTheme.shadows = {
```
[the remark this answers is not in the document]
forrestbaer: Yeah this was the reasoning for the let above, this solution is cleaner

forrestbaer:
@pwolfert great comments, a lot of cleanup done, should be ready for another review when you have time.

packages/design-system-tokens/src/lib/exportJson.ts
```diff
@@ -26,9 +26,9 @@ export const exportJson = (fileDescriptors: FileDescriptor[], outPath: string):
     // theme files have a description prop token files do not
     if (importedModule.default.description !== undefined) {
       isTheme = true;
-      tokenOutput.tokenType = 'theme';
+      tokenOutput[`${file.fileBaseName}`].tokenType = 'theme';
```
pwolfert, line 29: Here's a template string that can probably be removed

packages/design-system-tokens/src/lib/exportJson.ts
```diff
@@ -26,9 +26,9 @@ export const exportJson = (fileDescriptors: FileDescriptor[], outPath: string):
...
       filename = `${outPath}/${file.parentDirectoryName}-${file.fileBaseName}.tokens.json`;
-      writeFile(filename, JSON.stringify(tokenOutput, null, 4));
+      writeFile(filename, JSON.stringify(tokenOutput[`${file.fileBaseName}`], null, 4));
```
pwolfert, line 31: Extra template string here too

packages/design-system-tokens/src/lib/exportScss.ts
```diff
@@ -1,10 +1,19 @@
...
   return `$${name}: ${value};\n`;
 };
 
+const setVars = (items: Record<string, any>, filename: string) => {
+  let vars = '';
+  Object.entries(items).forEach(([name, value]) => {
+    name = `${filename}-${name}`;
+    vars += tokenFormat(name, value);
+  });
+  return vars;
```
pwolfert, line 14: Such a tiny thing, but this would be a perfect use for a reducer

packages/design-system-tokens/src/lib/exportScss.ts
```diff
@@ -18,32 +27,26 @@ export const exportScss = (fileDescriptors: FileDescriptor[], outPath: string):
      */
     // eslint-disable-next-line @typescript-eslint/no-var-requires
     const importedModule = require(`${file.moduleImportName}`);
+    const filename = `${outPath}/${file.exportFileName}.scss`;
+    let output = '';
+    let type = 'tokens';
```
pwolfert, line 32: It looks like this could maybe be replaced with a boolean to simplify, but I really like that you pulled this out into a variable so it's clear when we get to the if/else what is happening.

pwolfert: Oh, but now I see that the string helps us understand what the other case is. It's a token or a theme, and a boolean name would only capture one of those cases.

pwolfert: Could maybe consolidate this to
```suggestion
    const type = importedModule.default.description ? 'theme' : 'tokens';
```
so we can see what the two possible values are right away.

pwolfert: Sorry, this is a nit-pick

packages/design-system-tokens/src/lib/utility.ts
```diff
@@ -134,27 +134,16 @@ export const hexToRgb = (hex: HexValue): RGBValue | null => {
...
-  obj: Record<string, any>,
-  prefix = '',
-  res: Record<string, any> = {},
-  delim = '.',
-  prefixCategory = false
-) =>
-  Object.entries(obj).reduce((r, [key, val]) => {
-    const k = `${prefix}${key}`;
+// flattens an object into one dimension
+export const flatten = (obj: Record<string, any>, initialObject: Record<string, any> = {}) =>
```
pwolfert, line 33: I love that this only takes one parameter now 😍

packages/design-system-tokens/src/tokens/font.ts
```diff
@@ -8,39 +8,29 @@ const makeFonts = <T extends FontTokens>(value: T) => {
...
-    narrow: '45ex',
-    base: '65ex',
-    wide: '80ex',
-  },
+  'family-open-sans': '"Open Sans", Helvetica, sans-serif',
+  'family-rubik': '"Rubik", sans-serif',
+  'family-montserrat': '"Montserrat", sans-serif',
+  'family-bitter': 'Bitter, Georgia, serif',
+  'size-base': '16px',
+  'size-sm': '14px',
```
pwolfert, line 16: Okay, yeah, in the final generated tokens it was just one string anyway, right? I think that's a good simplification

pwolfert (approved):
Love the refactor/simplifications! I haven't tested it out again, but I'm sure you have. LGTM 👍

PR #1638, opened 2022-03-07
[WNMGDS-1427] [WNMGDS-1426] Add consistent header documentation deprecate navbar and global header

line47:
## Summary
Add consistent header documentation deprecate navbar and global header

### Added
- Added pattern page for consistent header and took info from MAX team Google doc 

### Deprecated
- Deprecated Navbar page (deprecation notice points to Consistent header page)
- Deprecated Global header (deprecation notice points to Consistent header page)

## How to test
- run the Mgov doc site `yarn start:medicare`
- Navigate to the consistent header pattern page and see content and instructions for use
- Navigate to navbar page and se deprecation notice
- Navigate to global header page and se deprecation notice

## Questions
- How can we get a live example of the global header set on on this page?

### DIFF
--- a/packages/ds-medicare-gov/docs/src/pages/patterns/consistent-header.md
+++ b/packages/ds-medicare-gov/docs/src/pages/patterns/consistent-header.md
@@ -4,9 +4,9 @@ usage: |
   The Consistent Header is a shared component available for all of medicage.gov. The header has two main states and three screen size ranges. The information below provides guidance for the various versions of the header so you can easily integrate it into any experience.
 ---
 
-For detailed information on the header contents, see [this Confluence page](https://confluence.cms.gov/display/MAX/%5BMAX%5D+Consistent+Header+Contents).
+For detailed information on the header contents, see [this Confluence page](https://confluence.cms.gov/display/MAX/Consistent+Header+Contents).
 
-For guidance on how to use the consistent header in a variety of scenarios, see the [Additional Guidelines](https://cms.invisionapp.com/dsm/cms/medicare/nav/build/asset/componentContainers/5fa15ef85296b21c9eb14293/tab/design%20guidelines?mode=preview).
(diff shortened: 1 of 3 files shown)


packages/ds-medicare-gov/docs/src/pages/patterns/consistent-header.md
```diff
@@ -0,0 +1,125 @@
+---
+title: Consistent header
+usage: |
+  The Consistent Header is a shared component available for all of medicage.gov. The header has two main states and three screen size ranges. The information below provides guidance for the various versions of the header so you can easily integrate it into any experience.
+---
+
+For detailed information on the header contents, see [this Confluence page](https://confluence.cms.gov/display/MAX/%5BMAX%5D+Consistent+Header+Contents).
```
scheul93, line 7: Two things here:
1. It seems a little odd that we are linking to a Confluence page that requires a login on our doc site. Is that a practice we do other places?
2. If we do want to link to confluence here, can we get an updated link? When I follow this URL, I get a `Page not found`

packages/ds-medicare-gov/docs/src/pages/patterns/consistent-header.md
```diff
@@ -0,0 +1,125 @@
...
+//Imports a single CSS file
+function importCSSFile(url) {
+	var cssElement = document.createElement('link');
+	cssElement.href = url;
+	cssElement.rel = 'stylesheet';
+	cssElement.type = 'text/css';
+
+	var head = document.head;
+	head.insertBefore(cssElement, head.childNodes[0]);
+}```
```
scheul93, line 106: This should be four ticks in order to format properly on the doc site

Current rendering
Screen Shot 2022-03-07 at 11 55 06 AM

packages/ds-medicare-gov/docs/src/pages/patterns/consistent-header.md
line47: 1. I think it's odd as well. This is the first attempt to get some of the consistent header into the Doc site and is very MVP
2. Link updated!

scheul93 (approved):

PR #1639, opened 2022-03-07
[WNMGDS-1159] Set up loki for Storybook-based visual regression testing

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1159

This adds [loki](https://storybook.js.org/addons/loki), which is a low-configuration visual regression tool that uses our existing Storybook stories. You'll see that I had to disable two of the stories:
1. `SkipNav` component - This isn't primarily a visual component anyway, and it only becomes visible on focus. Not really worth visual regression testing
2. `IdleTimeout` component - I wanted to at least have the "ViewDialog" story working, but it kept timing out. I want to revisit this later. It makes sense that we can't take screenshots of the one that is time-sensitive, but the "ViewDialog" story should be visible right away

I will create a follow-up PR for removing Backstop (our old VRT tool).

## How to test

See my changes to the readme for instructions. If those instructions are insufficient, I need to address that in the readme.


forrestbaer (approved):
This looks like a really straightforward solution!

scheul93 (approved):
Wow!

PR #1642, opened 2022-03-08
[WNMGDS-1159] Remove Backstop

pwolfert:
## Summary

Now that we have visual regression tests through Loki, remove Backstop and its reference images.


forrestbaer (approved):
Quite a culling!

PR #1649, opened 2022-03-14
[WNMGDS-1413] Add spacing example to page

line47:
## Summary

Adding spacing variables and values to the spacing page

### Added

- Added examples of our spacing values 

Screen Shot 2022-03-14 at 2 58 59 PM

## How to test

- Run `yarn start` to run the doc site and go to the spacing page which is under styles in the left navigation. 
- See that the spacing examples are on the page

### DIFF
--- a/packages/design-system-docs/src/pages/styles/spacing/spacing.example.html
+++ b/packages/design-system-docs/src/pages/styles/spacing/spacing.example.html
@@ -63,17 +63,3 @@
     </td>
   </tr>
 </table>
-
-<!-- <% var tokens = [ '', '$spacer-1', '$spacer-2', '$spacer-3', '$spacer-4', '$spacer-5',
-'$spacer-6', '$spacer-7' ]; -%>
-<% var pixels = [ '4', '8', '16', '24', '32', '40','48', '56' ]; -%>
-<div class="">
-  <% tokens.forEach(token => { -%>
-   
-  <article class="ds-u-margin-bottom--1">
-    <code><%= token %></code>
-    <% pixels.forEach(pixel => { -%> <code><%= pixel %>px</code> <% }) -%>
-  </article>
- 
-  <% }) -%>
-</div> -->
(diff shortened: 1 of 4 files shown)


packages/design-system-docs/src/pages/styles/spacing/spacing.example.html
```diff
@@ -0,0 +1,79 @@
...
+  <tr>
+    <td><code>$spacer-7</code></td>
+    <td><code>56px</code></td>
+    <td>
+      <div class="ds-u-fill--primary-alt" style="height: 56px"></div>
+    </td>
+  </tr>
+</table>
+
+<!-- <% var tokens = [ '', '$spacer-1', '$spacer-2', '$spacer-3', '$spacer-4', '$spacer-5',
```
forrestbaer, line 67: Can just remove this, git will save it if we need it again.

line47: Doh, thanks!

forrestbaer (approved):
lgtm! 👍🏻

scheul93:
What you have here is good but I was wondering if there was a way to make the example values on this page tied to our actual `$spacer-X` variables in the event we ever decide that our base space number is "10" instead of "8".

So I played around and I think if you edit this file `packages/design-system-docs/src/styles/settings/_layout.scss` to include some doc site specific example classes
```
.example-spacing-half {
    height: $spacer-half;
}
.example-spacing-1 {
    height: $spacer-1;
}
.example-spacing-2 {
    height: $spacer-2;
}
.example-spacing-3 {
    height: $spacer-3;
}
.example-spacing-4 {
    height: $spacer-4;
}
.example-spacing-5 {
    height: $spacer-5;
}
.example-spacing-6 {
    height: $spacer-6;
}
.example-spacing-7 {
    height: $spacer-7;
}
```

Then you could update the .example.html page items to use those classes. Something like
```
<tr>
    <td><code>$spacer-half</code></td>
    <td><code>4px</code></td>
    <td>
      <div class="ds-u-fill--primary-alt example-spacing-half"></div>
    </td>
  </tr>
```

I think this could achieve the same result but be tied to our sass variables more intentionally.

line47:
@scheul93 updated and ready for another look 👀

scheul93 (approved):
Looks good!

PR #1650, opened 2022-03-14
[WNMGDS-1455] removing visited link color from hgov menu items

scheul93:
## Summary

Updating hgov header links (in logged out & logged in variations) to maintain their primary color when the link has been visited.

Note, the selector in the CSS includes some `:not()`s because we want the link to change color when in an active or hover state.

## How to test

You can run `yarn storybook:healthcare` and navigate to components > Header and play around with the links.

Here's a gif where the Spanish link has the visited state enabled in browser tools and the login link does not. Notice how both links behavior similarly for focusing, clicking & hovering
HgovHeaderLinks


line47:
This looks good to me, I did notice one small thing that is not part of this PR but might be a nice little fix to add. 

When clicking the menu button when logged in the button size changed because a margin-class is only applied to one of the images used. I think we could update to add a margin-right class when the `x` close icon is present. This will keep the button from collapsing 

### Example of button changing size 
https://user-images.githubusercontent.com/1449852/158389346-1afc2f90-d697-4b7d-b762-cae84d29cb76.mov

line47 (approved):
LGTM!

scheul93:
@line47 Thanks for that catch! Updated now to look like
HgovHeaderIconUpdate

zarahzachz (approved):

forrestbaer (approved):
lgtm! 👍🏻

PR #1651, opened 2022-03-15
[WNMGDS-1432] Add error icon to html error examples

line47:
## Summary

HTML examples of error messages didn't all use the SVG error Icon this PR updates the HTML examples to use the error icon. 

### Added

- Added error SVG to HTML examples where it was missing 

## How to test

- Run the doc site locally `yarn start` and see that error icons are showing for HTML examples for the following: Radio/checkbox, form label, dropdown, date field, text field


packages/ds-healthcare-gov/docs/src/pages/components/DateField/datefield.example.html
```diff
@@ -107,7 +121,21 @@
         </div>
       </div>
       <span class="ds-c-field__error-message ds-c-field__error-message--inverse" role="alert">
-        Please enter a year in the past
+        <svg
+          aria-labelledby="icon-2__title"
+          aria-hidden="true"
+          class="ds-c-icon ds-c-icon--alert-circle"
+          focusable="false"
+          id="icon-2"
```
scheul93, line 129: there are two elements on this page that have `id="icon-2"` consider updating one so that no two elements on the page have the same id.

packages/ds-healthcare-gov/docs/src/pages/components/DateField/datefield.example.html
```diff
@@ -107,7 +121,21 @@
...
+          role="img"
+          viewBox="36 -12 186 186"
+          xmlns="http://www.w3.org/2000/svg"
+        >
+          <title id="icon-2__title">Alert</title>
```
scheul93, line 134: there are two elements on this page that have `id="icon-2__title"` consider updating one so that no two elements on the page have the same id. Will also need to update line 125 as part of this change

packages/ds-healthcare-gov/docs/src/pages/components/DateField/datefield.example.html
scheul93: Same situation for the form-label & textfield examples

packages/ds-healthcare-gov/docs/src/pages/components/DateField/datefield.example.html
scheul93: Same situation for the form-label & textfield examples

zarahzachz (approved):
LGTM!

scheul93 (approved):
Thanks for the updates!

PR #1660, opened 2022-03-22
[WNMGDS-1516] Fix null-ref error for uncontrolled Choice components

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1516

A "Cannot read properties of null (reading 'checked')" error is being thrown due to a race condition in the code. The Choice component constructor adds a listener to the `dsChoiceEmitter`, and that listener assumes the `this.input` ref exists. Really this listener should be added in a `componentDidMount` React lifecycle function in order to ensure this ref exists.

### Fixed

- Fixed uncontrolled Choice component throwing errors thrown when input ref accessed before it existed

## How to test

I've tested it out this proposed fix in the context of the app, and it works.

See ticket for steps to reproduce.


forrestbaer (approved):
Had some trouble testing it but the change looks good based on review of the code.

PR #1661, opened 2022-03-23
[NO-TICKET] Update @types/node as alternative to #1658

pwolfert:
## Summary

In https://github.com/CMSgov/design-system/pull/1658, dependabot tries to update `@types/node` in `ds-medicare-gov`, but that package doesn't actually need that dev dependency. Instead, let's update it in the package files where it's relevant and remove it from `ds-medicare-gov`.


forrestbaer (approved):
All tests and builds worked well for me, seems good 👍🏻

PR #1662, opened 2022-03-23
Update tooltip display to inline-block.

zarahzachz:
## Summary

[Jira](https://jira.cms.gov/browse/WNMGDS-1537)

When text is added after the Tooltip component, it's pushed to the next line when tooltip is opened (via hover or toggle).

This appears to have been caused by the block display of the Tooltip.

## How to test

To test, add text after the [inline Tooltip](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1537/BUGFIX/tooltip-line-wrap/core/components/tooltip/) via browser dev tools. Hovering over inline Tooltip should not result in text moving to next line.

### DIFF
--- a/packages/design-system/src/styles/components/_Tooltip.scss
+++ b/packages/design-system/src/styles/components/_Tooltip.scss
@@ -159,28 +159,28 @@ $tooltip-close-min-width: 32px !default;
 // Tooltip transition styles
 // http://reactcommunity.org/react-transition-group/css-transition
 .ds-c-tooltip-enter {
-  display: block;
+  display: inline-block;
   opacity: 0;
 }
 
 .ds-c-tooltip-enter-active {
-  display: block;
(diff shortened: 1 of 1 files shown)


pwolfert:
I'm actually having trouble reproducing the buggy behavior on master. I've tried putting extra text after it in a span and not in a span, but hovering over the inline trigger doesn't make it wrap. I've also tried lots of text and a small amount of text.

zarahzachz:
> I'm actually having trouble reproducing the buggy behavior on master.

Were you building Storybook or docs? Storybook can't recreate due to the flex utils wrapping the example, but you should be able to recreate on master/PROD doc site

forrestbaer (approved):
LGTM, tested locally with change and on current docs site using method shown, breaks line on current site, does not break line with this change. 👍🏻

PR #1668, opened 2022-03-25
[WNMGDS-1355][WNMGDS-1551] Translate a11y labels for Alert and Badge and add new i18n module

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1355
https://jira.cms.gov/browse/WNMGDS-1551

### Added

- Added an internationalization module in core that replaces `i18next`, and `@cmsgov/ds-healthcare-gov` uses this core module now

### Fixed

- Accessibility text in `Alert` and `Badge` now support Spanish

## How to test

The following core components were affected by this change:
- Alert
- Badge
- UsaBanner
- MonthPicker

The following healthcare components were affected:
- Footer
- Header

For each of these components, if they previously supported some kind of language/locale prop, make sure that old prop still works and overrides the globally-set language.


packages/design-system/src/components/MonthPicker/fallbackLocale.ts
```diff
@@ -0,0 +1,14 @@
+/**
```
forrestbaer, line 1: Seems like a useful function, maybe should be in i18n.ts?

forrestbaer (approved):
Manual + `yarn test` testing all passed for me, learned a lot about our i18 setup by going through this code..

pwolfert:
Wanted to "Make a way of testing out the globally set language in Storybook", but I think I'll need to leave that for a follow-up PR.

PR #1671, opened 2022-03-28
[WNMGDS-1535] choice ref update

scheul93:
## Summary

Updating the ChoiceList component to pass through an 'inputRef' defined on a choice.

### Changed

* If an 'inputRef' is defined on a choice in the 'choices' prop, it now gets passed down to the Choice component
* Updating ChoiceList tests to use 'testing-library' instead of Enzyme
* Added unit test to check if an inputRef on a choice i called

## How to test

Run storybook with `yarn storybook`. Navigate to Components > ChoiceList. If you open the dev console, you can see a log that occurs from an inputRef defined on one of the `choices` items


forrestbaer (approved):
Tests all ran fine for me, and I see the callback happening for choice lists. 👍🏻

packages/design-system/src/components/ChoiceList/ChoiceList.test.tsx
```diff
@@ -26,143 +26,169 @@ function render(customProps = {}, choicesCount = 2, deep = true) {
...
 
-      expect(wrapper.find('Choice').first().prop('type')).toBe('checkbox');
+      expect(choiceEl.type).toBe('checkbox');
     });
 
     it('renders all choices', () => {
-      const { props, wrapper } = render();
-      const choice = wrapper.find('Choice').first();
+      const numChoices = 3;
+      const { queryAllByRole } = renderChoiceList({}, numChoices);
```
pwolfert, line 66: The other day when I was troubleshooting the tests I was creating, I was looking through this article and saw this heading: https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen. KCD recommends using screen over the destructured `render` return value. It's a tiny thing, but I thought I'd share that.

packages/design-system/src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -9,7 +9,7 @@ export type ChoiceListType = 'checkbox' | 'radio';
 export type ChoiceListErrorPlacement = 'top' | 'bottom';
 
 // Omit props that we override with values from the ChoiceList
-type OmitChoiceProp = 'inversed' | 'name' | 'onBlur' | 'onChange' | 'size' | 'type' | 'inputRef';
+type OmitChoiceProp = 'inversed' | 'name' | 'onBlur' | 'onChange' | 'size' | 'type';
```
pwolfert, line 12: Oh, I didn't realize that we were omitting this before. It being part of the expected props was the basis of my justification for calling this a bugfix 😆

pwolfert (approved):
I love the new tests, and the change works well!

packages/design-system/src/components/ChoiceList/ChoiceList.test.tsx
scheul93: Thanks for sharing! This had some good suggestions for several updates that I've now incorporated to the tests

PR #1672, opened 2022-03-29
[WNMGDS-1551] Remove translation parameter from defaultMenuLinks function

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1551

As part of our 7.0 release, I had updated the `defaultMenuLinks` function to accept a required `t` (`i18next` _translate_) function, but I did not realize this function was being used externally. The way the translations are set up to support the old component-specific language props, it is impractical to expect teams to supply this function, which needs to use our internal `i18next` instance and be configured to use the right namespace.

Additionally, the list of parameters for that function was way too long, and the unit tests for it were actually full of mistakes where the wrong parameters were being sent. I switched this long list of parameters with an object with named properties, and I've updated the unit tests and snapshots to be correct.

### Changed

- Changed the function signature of `defaultMenuLinks` from a long list of positional parameters to a single object parameter with named properties

### Fixed

- Remove required `t` (`i18next` _translate_) function parameter for the `defaultMenuLinks` function used in the healthcare.gov `Header` component

## How to test

Run the unit test and also make sure the Header component works in Storybook

## Discussion

This could potentially be considered a breaking change. I don't know of other teams using this function directly besides the App3 team. If there are, we might want to bump `ds-healthcare-gov` to `8.0.0`. Otherwise, maybe we can consider it only a bugfix even though the API is changing. It'd be safer to call it a breaking change.

### DIFF
--- a/packages/design-system/src/__mocks__/i18next.js
+++ b/packages/design-system/src/__mocks__/i18next.js
@@ -7,5 +7,9 @@ i18next.init = jest.fn();
 i18next.addResourceBundle = jest.fn();
 i18next.changeLanguage = jest.fn();
 i18next.language = 'en';
+i18next.t = function (key, params) {
+  const paramString = params ? ` | ${JSON.stringify(params)}` : '';
+  return `${key}${paramString}`;
+};
 
 export default i18next;
--- a/packages/ds-healthcare-gov/src/components/Header/__snapshots__/Header.test.jsx.snap
+++ b/packages/ds-healthcare-gov/src/components/Header/__snapshots__/Header.test.jsx.snap
@@ -29,12 +29,12 @@ exports[`Header renders Direct Enrollment banner 1`] = `
           Array [
             Object {
               "href": "https://www.cuidadodesalud.gov/es/",
-              "label": "header.español",
+              "label": "header.español | {\\"ns\\":\\"en\\"}",
             },
             Object {
               "href": "/login?check_de=1",
               "identifier": "login",
-              "label": "header.login",
+              "label": "header.login | {\\"ns\\":\\"en\\"}",
             },
           ]
         }
@@ -49,12 +49,12 @@ exports[`Header renders Direct Enrollment banner 1`] = `
       Array [
         Object {
           "href": "https://www.cuidadodesalud.gov/es/",
-          "label": "header.español",
+          "label": "header.español | {\\"ns\\":\\"en\\"}",
         },
         Object {
           "href": "/login?check_de=1",
           "identifier": "login",
-          "label": "header.login",
+          "label": "header.login | {\\"ns\\":\\"en\\"}",
         },
       ]
     }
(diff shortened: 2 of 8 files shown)


packages/ds-healthcare-gov/src/components/Header/__snapshots__/Header.test.jsx.snap
```diff
@@ -29,12 +29,12 @@ exports[`Header renders Direct Enrollment banner 1`] = `
           Array [
             Object {
               "href": "https://www.cuidadodesalud.gov/es/",
-              "label": "header.español",
+              "label": "header.español | {\\"ns\\":\\"en\\"}",
```
pwolfert, line 32: I updated the translation mock so that it prints out the extra options. That's why these snapshots are changing. It's showing that it's looking for this translation key in the `en` namespace

packages/design-system/src/__mocks__/i18next.js
```diff
@@ -7,5 +7,9 @@ i18next.init = jest.fn();
 i18next.addResourceBundle = jest.fn();
 i18next.changeLanguage = jest.fn();
 i18next.language = 'en';
+i18next.t = function (key, params) {
+  const paramString = params ? ` | ${JSON.stringify(params)}` : '';
+  return `${key}${paramString}`;
+};
```
pwolfert, line 13: Moved the existing translation mock to the `i18next` file instead of the `react-i18next` file so it can be used outside of react

packages/ds-healthcare-gov/src/components/Header/defaultMenuLinks.test.js
```diff
@@ -2,95 +2,71 @@ import defaultMenuLinks, {
...
   it('includes a named export', () => {
     // Some apps may need to import the default menu links in order to extend them
     expect(typeof namedExportDefaultMenuLinks).toBe('function');
   });
 
   it('leaves out login link if hideLoginLink true', () => {
-    expect(
-      defaultMenuLinks(t, undefined, undefined, undefined, undefined, undefined, true)
-    ).toMatchSnapshot();
+    expect(defaultMenuLinks({ hideLoginLink: true })).toMatchSnapshot();
```
pwolfert, line 12: These unit tests were wrong in several places because one of the parameters in this long list would be missing. It was so hard to read!

forrestbaer:
Tests run fine for me, and as for the code I find the interface a little confusing but understandable, perhaps there could be a little more documentation on how to utilize / manage our internationalization options in the new docs site, I find the current documentation a little light.

Ran into some build issues but nothing that seems to have broken the site, during the docs site build:
image

pwolfert:
@forrestbaer, are you finding the API change in this PR confusing or the i18n interface confusing for al of https://github.com/CMSgov/design-system/pull/1668 ?

As far as those build warnings go, I don't know that I can fix them. It actually does successfully export those types in the TypeScript-compiling step.

forrestbaer:
@pwolfert This is my first time looking closely at what we're doing with internationalization, and I think I'm of the opinion that we shouldn't be doing it.  It feels like something the consuming application should be worrying about and not the DS. I guess I'm confused about the whole thing from that angle, but if that ship has sailed and we're going this way, it's just that I haven't worked with this code yet so don't have a mature view of how it's operating.

forrestbaer (approved):

pwolfert:
@forrestbaer, noted your concerns about the technical solution of https://github.com/CMSgov/design-system/pull/1668 in general. I'm going to merge this into that branch for now (because it fixed that bug), but we'll continue that discussion outside of this PR

PR #1673, opened 2022-03-29
[WNMGDS-831] Refactor choice to use native input element

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-831).

This work refactors the Choice component to rely on native radio and checkboxes instead of creating a custom input. This solves two problems:

1. Relying on native inputs is good for accessibility
2. Updating the styling to use box-shadow instead of background color fixes print styles

## ~TODO~ DONE

- [x] Test solution in browsers
**Results:** Solution works for IE10 (couldn't get instance of docs site to load for IE11), and all modern browsers
- [x] Test HCM to make sure styles weren't wiped away
**Results:** Improved HCM experience - radio buttons receive HCM focus state (consistent with checkbox) and native select outline color now visible when user selects Choice
- [x] Update code examples to ensure they're all the same structure (i.e., add wrapping `div` to choice and label)
- [x] Finish styling labels to match PROD

## How to test

Instructions on how to test the changes. This is not exhaustive list of ways you should test this PR.

Compare [PROD Choice examples](https://design.cms.gov/components/choice/) and [updated Choice examples](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-831/checkbox-refactor/core/components/choice/) to see if they differ (they shouldn't).


packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -1,252 +1,194 @@
...
+    &::after {
+      opacity: 1;
     }
+
+    /* stylelint-disable max-nesting-depth, selector-max-specificity */
+    &:focus::before {
+      height: $choice-size - ($choice-border-width * 2);
+      width: $choice-size - ($choice-border-width * 2);
+    }
+    /* stylelint-enable max-nesting-depth, selector-max-specificity */
```
forrestbaer, line 129: Can remove all stylelint disables like this, and merge master which includes the eslint/stylelint updates.

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -1,252 +1,194 @@
...
-    color: $color-muted-inverse;
+.ds-c-choice--inverse {
+  background-color: $choice-background-color-inverse;
+  border-color: $choice-border-color-inverse;
 
-    &::before {
-      background-color: $choice-disabled-background-color-inverse;
-      border-color: $choice-disabled-border-color-inverse;
+  // Error
+  @if $ds-include-choice-error-highlight {
```
forrestbaer, line 149: This is interesting, I didn't realize we were using style toggles like this.. Feels a little awkward, like maybe anywhere this is used it should just be another style? 

Wondering about working these into theming now. It looks like this is the only variable that I can see used in this way.

zarahzachz: I agree. I sent a message to [the CMS Slack](https://cmsgov.slack.com/archives/CT169N26L/p1648567752752839) for more guidance since I don't know what purpose this serves.

packages/design-system/src/styles/components/_Choice.scss
zarahzachz: Merging master results in an error:
> Error: Could not find "stylelint-config-recommended". Do you need a `configBasedir`?

Is there a step I'm missing to get master updated?

forrestbaer: You'll have to do a yarn install after merging, should fix it.

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -1,240 +1,224 @@
...
+  position: relative;
+  width: $choice-size;
+
+  &::before {
+    box-shadow: inset 0 0 4em 1em $choice-checked-background-color;
+    content: '';
+    grid-area: 1/-1;
+    height: $choice-size;
+    opacity: 0;
+    transition: 120ms opacity ease-in-out;
```
zarahzachz, line 79: @line47 @ericwbailey @babsmiles This is the only visual change I implemented - it's a subtle animation when users select/deselect a choice. Lemme know if yinz think it's gross.

line47: I think this looks good and also have 2 thoughts on adding in this animation...


1. Should we use some animation settings from our variables? https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/styles/settings/variables/_animation.scss you can see a commented out example here. https://github.com/CMSgov/design-system/blob/7a0f6c1c2bfcc74b773f14bf63c156972a232a8c/packages/design-system/src/styles/components/_Tabs.scss#L44

2. My second thought is we've never established animation guidelines or a plan for how animation would be integrated into the design system which makes me think...should we establish a deliberate direction for our animations? This makes me want to take a step back and have a more defined approach to animation and how it could and would be applied to components that exists in the design system today.

zarahzachz: Good points. I can refactor the transition I currently have to use one of our pre-existing variables (or comment it out). 

I think #2 would be a worthwhile (and fun) exercise. I imagine there's a few components that could benefit from animation.

line47: I agree on the second point and think it could be a nice touch to the system overall, thanks for bringing this up! Should we backlog an animation plan?

pwolfert: I just thought of another consideration: I'm not sure how this change would affect teams' visual regression tests that include this component in a checked state. Would their testing suite know to wait for the animation end to snap the picture? If it's a race, how likely is it to get non-deterministic image results?

pwolfert: I do like the transition visually

pwolfert (approved):
These changes look ✨ _great_ ✨

I'm unsure about the animations at the moment, but it's something we could easily revert if we need to

packages/design-system/src/styles/components/_Choice.scss
zarahzachz: @line47 I tried using `$animation-speed-1` in this transition, but it was slower than the 120ms I currently have and made it look a little laggy. I'm going to comment it out until we have a better animation guidelines in place.

line47: Sounds good, thanks for testing!

PR #1674, opened 2022-03-30
No ticket: Prettier Update

forrestbaer:
## Summary

In the latest linting update I introduced the latest version of prettier.  This PR adds the ignore file for prettier and has been run with that ignore file and includes a few updates.

Let me know if there are other ignores that should be added to this file.

I've also removed the running of prettier from the lint step, do we need it there? I'm assuming the changes listed here somehow got around the auto husky running of prettier during git commit, or were pre-husky.

I like the idea of `yarn lint` just running linting tasks..

`yarn prettier` runs prettier.. Pretty straightforward

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,6 @@
+**/node_modules/**
+**/dist/**
+tmp/**
+**/types/**
+storybook-static/**
+packages/coverage/**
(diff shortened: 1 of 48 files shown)


scheul93 (approved):
Can't think of any other things to add to the ignore right now!

pwolfert (approved):

PR #1681, opened 2022-04-04
[NO-TICKET] Update Choice styles in mgov, core

zarahzachz:
## Summary

Medicare had a lot of style overrides for Choice, which impacted the MonthPicker component.

I updated the overrides to account for Medicare's:
- hover state on checkboxes and radios
- reduced spacing between Choice and label
- vertical alignment between Choice and label (label uses a different font, requiring a different translate value)
- size of a focused and selected checkbox (background color is edge-to-edge)

I also noticed the radio button border color in Core was errantly updated to black when it was supposed to stay blue, so I've update that as part of this work.

## Test

To test Mgov styles, compare [PROD](https://cmsgov.github.io/design-system/medicare/components/choice/) to Storybook.

To test Core styles, compare [PROD](https://design.cms.gov/components/choice/) to Storybook.


packages/ds-medicare-gov/src/styles/components/_override.choice.scss
```diff
@@ -1,58 +1,29 @@
-// positioning the radio
```
pwolfert, line 1: Might be good to have an updated comment in here to give context for the override

pwolfert (approved):
LGTM! It's nice that there's a bit less of the override code now

PR #1685, opened 2022-04-05
[DON'T MERGE] [WNMGDS-1507] dialog discovery

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1507)

Discovery work to replace current dialog implementation with native `dialog` HTML element.

## DoD
- [x] Tested on modern browsers
- [x] Tested on IE11 (progressive enhancement OK - means users have access to content, but may not be visually appealing)
- [x] Explore Modal dialog
- [x] Explore Drawer dialog
- [ ] Explore Tooltip dialog (this was a stretch goal)

## Test

[Storybook implementation](http://design-system-demo.s3-website-us-east-1.amazonaws.com/WNMGDS-1507/dialog-discovery/core/storybook/?path=/story/components-dialog--dialog) to test in modern browsers.

To test in IE, you'll have to build this project locally and run the react-app index.html file in a VM (I used Sauce Labs).

1. At project root, run `yarn build` to make sure the design system's compiled correctly
2. In another terminal, `cd examples/react-app` and `rm -rf dist node_modules` to make sure the previous build of the design system has been properly cleared
3. In that terminal window, run `yarn install && yarn build`
4. Run a local server on the react-app index.html file (I used VS Code's Live Server extension)

**Note:** You'll have to [create a tunnel](https://docs.saucelabs.com/secure-connections/sauce-connect/proxy-tunnels/) to run your local code in Sauce Labs


forrestbaer:
Looks cool! The storybook run works fine for me but having some trouble building docs site, haven't looked deeply into it.. Anyway like how it's being used in your demo.

examples/react-app/package.json
```diff
@@ -3,17 +3,19 @@
...
   },
   "devDependencies": {
     "@babel/core": "^7.8.4",
     "@babel/preset-env": "^7.8.4",
     "@babel/preset-react": "^7.8.3",
     "babel-loader": "^8.0.0",
+    "css-loader": "5.2.7",
     "gulp": "^4.0.2",
     "gulp-sass": "^3.1.0",
+    "style-loader": "2.0.0",
```
zarahzachz, line 18: `css-loader` and `style-loader` were required for this project to properly render the imported CSS file from our design system. These specific versions were required as they're the last to work with our version of Webpack.

**tldr; these packages were added to import polyfill CSS.**

examples/react-app/webpack.config.js
```diff
@@ -19,6 +19,10 @@ const config = {
           },
         ],
       },
+      {
+        test: /\.css$/i,
+        use: ['style-loader', 'css-loader'],
+      },
```
zarahzachz, line 25: The Webpack config needed special permissions to import the polyfill CSS correctly. I'm not a Webpack Wizard, so there's probably a better way to handle this, but since I was just proving an IE solution could be reached, I went with what worked.

tsconfig.json
```diff
@@ -2,6 +2,7 @@
   "compilerOptions": {
     "target": "es2015",
     "jsx": "react",
+    "module": "esnext",
```
zarahzachz, line 5: This line was needed, again, to support our polyfill CSS. I kept getting errors about not being able to import a module.

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -1,194 +1,19 @@
 import { EVENT_CATEGORY, MAX_LENGTH, sendLinkEvent } from '../analytics/SendAnalytics';
-import AriaModal from 'react-aria-modal';
 import Button, { ButtonVariation } from '../Button/Button';
-import React, { useEffect, useRef } from 'react';
+import React, { useEffect, useLayoutEffect, useRef } from 'react';
 import classNames from 'classnames';
 import { dialogSendsAnalytics } from '../flags';
 import { CloseIcon } from '../Icons';
+import dialogPolyfill from 'dialog-polyfill';
+import 'dialog-polyfill/dist/dialog-polyfill.css';
```
zarahzachz, line 8: Ideally, I'd like to conditionally import this CSS if `window.HTMLDialogElement` doesn't exist, but ran into issues getting that to work. Importing it here just to demo what an IE solution would look like.

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -1,194 +1,19 @@
...
-   * but as of focus-trap v4, the pausing and unpausing of hierachical traps is handled automatically.
-   */
-  focusTrapPaused?: boolean | undefined;
-
-  /**
-   * Customize properties of the focusTrapOptions prop that is passed to the modal dialog's focus trap.
-   * For example, you can use this prop if you need better control of where focus is returned.
-   */
-  focusTrapOptions?: Record<string, unknown> | undefined;
+declare const window: any;
```
zarahzachz, line 14: I had to add this line to skirt some TS errors. TS wasn't reading `HTMLDialogElement` as a valid type of window (even though it should be?). Maybe our types are outdated?

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -334,86 +148,90 @@ export const Dialog = (props: DialogProps) => {
...
-  } = props;
+  // Toggle dialog open
+  const dialogRef = useRef(null);
+  const lastActiveElement = React.useRef(null);
+  const firstRender = React.useRef(true);
+
+  useLayoutEffect(() => {
+    if (window.HTMLDialogElement === undefined) {
+      dialogPolyfill.registerDialog(dialogRef.current);
+      // import('dialog-polyfill/dist/dialog-polyfill.css');
```
zarahzachz, line 159: This is where I'd like to conditionally import the polyfill CSS. The error seemed to be this file couldn't find this CSS path and I didn't know how to fix that.

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -334,86 +148,90 @@ export const Dialog = (props: DialogProps) => {
...
+    }
+  }, []);
+
+  useLayoutEffect(() => {
```
zarahzachz, line 163: Using `useLayoutEffect` because I think this is modifying the DOM. Still a little unclear on the difference between this hook and `useEffect`

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -334,86 +148,90 @@ export const Dialog = (props: DialogProps) => {
...
-    size,
-    title,
-    ...modalProps
-  } = props;
+  // Toggle dialog open
+  const dialogRef = useRef(null);
+  const lastActiveElement = React.useRef(null);
+  const firstRender = React.useRef(true);
+
+  useLayoutEffect(() => {
```
zarahzachz, line 156: Using `useLayoutEffect` because I think this is modifying the DOM. Still a little unclear on the difference between this hook and `useEffect`

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -334,86 +148,90 @@ export const Dialog = (props: DialogProps) => {
...
+    if (window.HTMLDialogElement === undefined) {
+      dialogPolyfill.registerDialog(dialogRef.current);
+      // import('dialog-polyfill/dist/dialog-polyfill.css');
+    }
+  }, []);
+
+  useLayoutEffect(() => {
+    // prevents calling imperative methods on mount since the polyfill will throw an error since we are not using the `open` attribute
+    if (firstRender.current) {
```
zarahzachz, line 165: I had to add some extra `ref` hooks to address an IE bug where IE was expecting the `open` attribute to exist on the `<dialog>` element. Since this doesn't exist until after the modal's triggered, there was an issue.

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -334,86 +148,90 @@ export const Dialog = (props: DialogProps) => {
...
+  // https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/oncancel
+  // onCancel allows close/open of dialog using keyboard - state doesn't get tripped up
+  // Receiving TS error saying `onCancel` doesn't exist on HTMLDialogElement
+  useEffect(() => {
+    const dialogNode = dialogRef.current;
+    const handleCancel = (event) => {
+      event.preventDefault();
+      onExit();
+    };
+    dialogNode.addEventListener('cancel', handleCancel);
```
zarahzachz, line 192: TS kept yelling at me for adding `onCancel` to the dialog, so I'm adding it here

packages/design-system/src/components/Dialog/Dialog.tsx
zarahzachz: Also pointing out that adding the polyfill CSS at the top doesn't _hurt_ our dialog. So maybe this is fine

packages/design-system/src/components/Dialog/Dialog.tsx
pwolfert: This is only 37 lines of CSS unminified, so I'm not worried about it. Do you think when we implement it for real that it would work to import it from our SCSS files? That is, I wonder if we can avoid changing the webpack config if we forget about conditionally importing it.

pwolfert: That is, unless the concern is that we shouldn't be applying that CSS at all if it has native support

zarahzachz: My concern is if we import the polyfill CSS regardless of IE/not IE and the styles get updated (unlikely but not impossible), we might see visual regressions get into our component. We could skirt that issue altogether and copy those styles into our `dialog.scss` file (at the top) and not bother importing at all.

PR #1687, opened 2022-04-06
[NO-TICKET] Remove the `child-design-system` example

pwolfert:
It's starting to get outdated, and we don't need it anymore. We maintain all child design systems ourselves now, and they live in this repository. Removed the folder and all references to it. Those documentation pages and READMEs that reference it are going to be pretty outdated soon with the design-token work anyway.

Sarah, I was working on getting one of the examples working with IE and got a little side-tracked by deleting this. I didn't want to include this deletion in that work because it would be a huge diff 😂


scheul93 (approved):
👋🏻

PR #1688, opened 2022-04-06
[NO-TICKET] Loki.sh - script to automate test running with loki/storybook

forrestbaer:
## Summary

Makes it easier to run Loki visual regression tests!

_Doesn't check if docker is currently running, I suppose it could? Not sure how that would work on Jankins.. Whatcha think?_

### Added

Loki.sh script which takes a single argument, if none passed it defaults to working on core, otherwise `healthcare` and `medicare` will run their respective yarn jobs

## How to test

Have docker running.. 

Possible `args` are: `update` `test` `approve`

`./loki.sh args` `./loki.sh -t healthcare args` `./loki.sh -t medicare args` .. See that it runs the tests and kills the storybook process when complete regardless of success/failure.


loki.sh
```diff
@@ -0,0 +1,71 @@
...
+  fi
+  attempts=$(($attempts+1))
+  sleep 2
+done
+
+# storybook is up, start the appropriate loki process
+msg "[${GREEN}✓${NOFORMAT}] Storybook is up!"
+msg "${NOFORMAT}--- Starting ${CYAN}yarn loki:update${args}${NOFORMAT}"
+
+if yarn loki:update$args ; then
```
pwolfert, line 56: What if we pass extra args through to loki like `test` and `update`? So you could use it for both loki commands like `./loki.sh test` or `./loki.sh update` or `./loki.sh update --healthcare`? I guess there's a variety of ways it could be structured, but we have two use cases we can cover over three brands ({test, update} for {core, healthcare, medicare})

pwolfert: Actually, there's also a an `approve` command, which prunes old images. So this command could cover {test, update, approve} for {core, healthcare, medicare}.

pwolfert (commented):
This is a handy script! I have an idea below for how we could expand its use. I dig the output. It looks really nice in my terminal.

pwolfert:
I just tried sabotaging loki, too, and it gracefully shut down the Storybook instance! 😃

<details>
  <summary>Terminal output (without the fancy colors)</summary>

```
--- Starting yarn storybook with process id 88835
[✓] Storybook is up!
--- Starting yarn loki:update
yarn run v1.22.11
$ yarn loki update
$ /Users/patrickwolfert/Projects/design-system/node_modules/.bin/loki update
loki update v0.28.1
Unexpected identifier
/Users/patrickwolfert/Projects/design-system/loki.config.js:18
hello world
      ^^^^^

SyntaxError: Unexpected identifier
    at wrapSafe (internal/modules/cjs/loader.js:984:16)
    at Module._compile (internal/modules/cjs/loader.js:1032:27)
    at Object.Module._extensions..js (internal/modules/cjs/loader.js:1097:10)
    at Module.load (internal/modules/cjs/loader.js:933:32)
    at Function.Module._load (internal/modules/cjs/loader.js:774:14)
    at Module.require (internal/modules/cjs/loader.js:957:19)
    at module.exports (/Users/patrickwolfert/Projects/design-system/node_modules/cosmiconfig/node_modules/import-fresh/index.js:32:59)
    at loadJs (/Users/patrickwolfert/Projects/design-system/node_modules/cosmiconfig/dist/loaders.js:16:18)
    at ExplorerSync.loadFileContentSync (/Users/patrickwolfert/Projects/design-system/node_modules/cosmiconfig/dist/ExplorerSync.js:84:26)
    at ExplorerSync.createCosmiconfigResultSync (/Users/patrickwolfert/Projects/design-system/node_modules/cosmiconfig/dist/ExplorerSync.js:89:30)
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
[X] There was a problem with Loki
[✓] Successfully killed storybook instance
```
</details>

loki.sh
forrestbaer: Nice, thanks for the feedback! Yes this sounds like a good set of expansions.. easy enough to add flags, could probably even run multiple sequentially that would be nice.. I'll dig into it :construction\_worker:

forrestbaer: Made these updates, removed color (made code messy, not needed), run it by itself for usage.. updated test in pr desc

pwolfert (approved):
I kind of miss the colors, but wow is this script easier to read! It works really well, too.

To now integrate this into our package scripts, I propose the following:
1. We replace the existing `loki:*` commands in `package.json` with one `"loki": "./loki.sh",`
2. We update the commands inside this script to not use `yarn loki` but to execute `node ./node_modules/.bin/loki` directly so we don't have an infinite loop
    - Optionally this could be avoided if we alias `./loki.sh` as something else, like `yarn vrt` or `yarn test:vr` or `yarn test:browser`
3. To match our other commands, maybe alias the different child design systems with `yarn loki:healthcare` and `yarn loki:medicare`

loki.sh
```diff
@@ -0,0 +1,105 @@
...
+      -v | --verbose) set -x ;;
+      -t | --target) 
+        export LOKI_FILE_PREFIX="${2-}"
+        export STORYBOOK_DS="${2-}"
+        shift
+        ;;
+      *) break ;;
+    esac
+    shift
+  done
```
pwolfert, line 56: Does it make sense to pull this out of this function now that it is in charge of populating the env vars for both loki and storybook?

pwolfert (approved):
That makes sense to me. Looks great!

loki.sh
forrestbaer: I think we should leave it, in case someone wants to run it by itself on a new theme in the future before adding to scripts

forrestbaer: Oh wait i understand.. I'll just rename it to something more generic

pwolfert: Sorry, I just meant bumping this while loop code up to the script level and out of this function

pwolfert: Could have a `parse_args` and `run_storybook` function if you'd rather keep it in a function

forrestbaer:
```
./loki.sh: line 75: curl: command not found
[x] Something went wrong starting storybook .
```

Guess no curl over there :) Since the name was the same it ran my script instead... I'll mess around with the image and see what else might work.. or we could update the test script to download curl with :

```
apt-get -qq update

apt-get -qq -y install curl
```

loki.sh
forrestbaer: oh yes of course, good point.. updated

loki.sh
```diff
@@ -0,0 +1,98 @@
+#!/usr/bin/env bash
+
+set -Eeo pipefail
+trap cleanup SIGINT SIGTERM ERR
+
+script_dir=$(cd "$(dirname "${BASH_SOURCE[0]}")" &>/dev/null && pwd -P)
+
+usage() {
+  cat <<EOF
+Usage: $(basename "${BASH_SOURCE[0]}") [-h] [-v] [-t target] args
```
pwolfert, line 10:
```suggestion
Usage: $(basename "${BASH_SOURCE[0]}") [-h] [-v] [-t target] <...loki args>
```

PR #1684, opened 2022-04-05
[MCT-6718] Adding medicare specific icons

Afrothundr:
## Summary
Resolves #1683
Adds 4 new medicare specific icons

### Added
|name |screenshot  |
--- | --- |
|`<CheckShieldIcon />` |image|
|`<DrugsIcon />`|image|
|`<PiggyBankIcon />`|image|
|`<PharmacyIcon />`|image|

## How to test
- Boot up the Medicare Design System docsite
- Navigate to `/components/icon/` and see the new icons added at the top of the page

PR #1689, opened 2022-04-06
[MCT-6718] Adding medicare specific icons

pwolfert:
See https://github.com/CMSgov/design-system/pull/1684. Merged into a branch on this repo so the CI tests would run correctly


forrestbaer (approved):
All pass, LGTM 👍🏻

PR #1691, opened 2022-04-07
[DO-NOT-MERGE] Analytics simplification discovery

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1604

Right now our components that send content to analytics have a bunch of verbose code around pulling text content out of the rendered html in the DOM. This could potentially be consolidated into either a function that statically renders the content into a string or into a component that provides a callback prop that sends back a rendered string.

### Requested feedback

I've got two ideas that I'm showcasing in this PR, and I would like feedback.
1. **A React component**
    - Pro: It doesn't change the bundle size. It's a net savings on code
    - Con: I haven't figured out a way to get a ref to an arbitrary `children` without adding an element (`<span>`). I tried `React.Fragment`, but it doesn't take a ref
    - Con: In the case where content can come from two different places, the there's extra branching logic in the render function (e.g., when Alert analytics uses `heading` but falls back to `children`)
2. **A function that renders ReactNodes statically**
    - Pro: Can use it immediately and get a value right where you want it. So simple
    - Con: Increases the bundle size to import `react-dom/server` to use functions like `renderToString` or `renderToStaticHtml`. According to my calculations, it's about 20kb
    - Con: I'm still going to have to do some parsing in order to get the actual string content out of the html

After implementing it, I don't really like the first idea. I really like the simplicity of the second idea. It might be worth the bundle size increase, especially since we were just able to remove i18next as one of our dependencies


packages/design-system/src/components/analytics/AnalyticsContent.tsx
```diff
@@ -0,0 +1,27 @@
...
+/**
+ * For analytics, we often need to send a string version of the content that the user
+ * is being shown. However, most of our content props allow for React nodes, which
+ * must be rendered before it can be turned into a string. While there is an existing
+ * function in `react-dom/server` that will render JSX components to strings, using
+ * it here would add about 20KB to the bundle and would include html tags, which we
+ * don't care about. This component renders content and sends a string back to the
+ * parent through a callback.
+ */
+export const AnalyticsContent = ({ children, onContentStringRendered }: RenderToStringProps) => {
```
pwolfert, line 17: This is the first idea, and I show how it would be implemented in the `Alert` component

packages/design-system/src/components/analytics/getRenderedTextContent.tsx
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import ReactDOMServer from 'react-dom/server';
+
+export function getRenderedTextContent(reactNode: string | React.ReactNode): string {
```
pwolfert, line 4: This is the second idea, and I show how it would be implemented in the `HelpDrawer` component

packages/design-system/src/components/analytics/AnalyticsContent.tsx
```diff
@@ -0,0 +1,27 @@
...
+  const ref = useRef<HTMLSpanElement>();
+
+  useEffect(() => {
+    if (ref.current) {
+      onContentStringRendered(ref.current.textContent);
+    }
+  });
```
scheul93, line 24: [question] would you want to have a conditional variable here as the second param of `useEffect`? Or is the intention that this effect will run on every re-render?

packages/design-system/src/components/analytics/getRenderedTextContent.tsx
scheul93: I guess this works and is simple, but it feels like a slight mis-use of this library which seems to be intended to be used on a server 🤷🏻‍♀️

forrestbaer: That was my gut reaction as well, but if it works well and is simple I thought I'd put that feeling aside.. I am curious how it will be used in the wild, what an actual use case would look like

packages/design-system/src/components/analytics/AnalyticsContent.tsx
pwolfert: I thought about that but didn't know the answer off the top of my head. I figured that was a detail I could figure out if we actually wanted to implement it. What do you think of this option?

packages/design-system/src/components/analytics/getRenderedTextContent.tsx
zarahzachz: The 2nd idea has a cleaner implementation that I like more than the first idea

pwolfert:
Closing this to keep it as a record of our conversations around discovery. Will open a new pull request now that I've selected an option (third option not originally listed) and implemented it

PR #1698, opened 2022-04-08
[WNMGDS-1376] initial add of gatsby doc site

scheul93:
## Summary

Adding a new Gatsby-powered documentation site at `packages/docs`

I added some placeholder components and some styles to help establish a directory structure.

## How to test

* `yarn install`
* ~navigate to packages/docs and `yarn develop` to run Gatsby locally~ run `yarn start:gatsby` from the root directory. This should spin up localhost:8000
* You should see a simple page with header, main & footer text

## Notes
* the tsconfig is generated by Gatsby. If there are updates or a way to make it consistent with our larger system, please let me know
* I got our core styles imported and working (with fonts!). This section of code may update if/when we get a theme switcher


scheul93:
🤔 I may refactor so that you don't have to navigate to the directory to run the doc site and instead, make it a command from our root directory...

package.json
```diff
@@ -16,6 +16,8 @@
     "build-docs": "yarn cmsds build-docs",
     "build-docs:healthcare": "yarn cmsds build-docs --config cmsds.healthcare.config.js",
     "build-docs:medicare": "yarn cmsds build-docs --config cmsds.medicare.config.js",
+    "build:gatsby": "yarn --cwd ./packages/docs build",
```
forrestbaer, line 19: maybe build:docs, start:docs ?

packages/docs/README.md
```diff
@@ -0,0 +1,35 @@
...
+### Scripts
+
+The following scripts can be run from the `packages/docs` folder.
+
+`yarn install`
+Installs dependencies for the doc site. Only needs to be run on first spin up of the project, or if new dependencies are installed.
+
+`yarn develop`
+Runs the site locally at `http:localhost:8000/`
+Also runs the GraphiQL playground at `http://localhost:800/___graphql`.
```
forrestbaer, line 16: This is pretty cool!

forrestbaer (dismissed):
Still getting those annoying scss warnings about math.div, can clean all those up when we move off gulp fairly easily.. 

Love the setup, looks very simple to work with.

package.json
pwolfert: we can change this when we retire the legacy one

scheul93: I struggled with naming here because I didn't want to conflict with the current doc site. 
And to Patrick's point, I think maybe this can be updated right before we merge this feature work?

forrestbaer: Yeah no biggie, object:verb or verb:object! We use a little of both.. always something we can update in the future..

pwolfert:
On this branch, I'm unable to `yarn build` without TypeScript errors. Does anyone else get that?

forrestbaer:
Oof yeah I didn't see that good catch.. Didn't expect to see that come up.. Something to do with Lerna, I deleted package.json and tsconfig.json and they went away.\\

forrestbaer (approved):
LGTM 👍🏻

PR #1699, opened 2022-04-08
[NO-TICKET] resolve SCSS deprecation warnings

### LINKED ISSUE #1602: Sass deprecation warnings
Getting sass deprecation warnings in the cms design package . We're using gatsby 

```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.

Recommendation: math.div($spacer-1, 2) or calc($spacer-1 / 2)

More info and automated migrator: https://sass-lang.com/d/slash-div
   ╷
21 │ $spacer-half: $spacer-1 / 2; // 4px
   │               ^^^^^^^^^^^^^
   ╵
    node_modules/@cmsgov/design-system/dist/scss/settings/variables/_layout.scss 21:15  @import
    node_modules/@cmsgov/design-system/dist/scss/settings/variables/_index.scss 6:9     @import
    node_modules/@cmsgov/design-system/dist/scss/settings/_index.scss 2:9               @import
    node_modules/@cmsgov/design-system/dist/scss/components/_Table.scss 1:9             @import
    src/components/Markdown/Markdown.scss 1:9                                           root stylesheet

```

Packages:
```
  "gatsby": "^3.8.1",
  "gatsby-plugin-sass": "^4.1.0",
    "sass": "^1.49.7",
    "sass-loader": "^12.1.0",

dev:
   "node-sass": "^6.0.1",
```

DavidMCarek:
## Summary

Resolves #1602

Convert inline division to either decimal multiplication or use the `math` library's division function (`math.div(<param1>, <param2>)`) to resolve the deprecation warning. https://sass-lang.com/documentation/breaking-changes/slash-div

Changes were made using the automatic Sass migrator mentioned in the documentation
```
$ npm install -g sass-migrator
$ sass-migrator division **/*.scss
```
### Changed

Simple division like `($var / 2)` has been updated to `($var * 0.5)` and more complex division now done like `math.div($var, 107)`

## How to test

Just make sure the components that have been updated still have the same calculated sizes as before by comparing with what's currently in production.


forrestbaer:
@DavidMCarek Unfortunately this is not a simple fix for us as there are different sass preprocessors running between storybook and the design system.  If you run `yarn build` you'll see an error right away, `yarn test` will also fail.. At the moment we can't change the preprocessor on the DS side until we've phased out gulp.

DavidMCarek:
@forrestbaer that's unfortunate lol. There's only 1 division operation that would actually require the sass upgrade which is the `flexbox-col` mixin in the `_layout.scss`. Would it be worth while to resolve the other deprecation warnings except that one?

forrestbaer:
I'm open to that @DavidMCarek, I just ran a visual regression against these changes and things look fine.. A few less warnings is always a good thing.

forrestbaer (dismissed):

forrestbaer:
Hmm, actually looks like there was a couple regressions in healthcare, looks like to our current gulp preprocessor, these changes actually can produce different results.. I remember running into issues like this in the past when trying to solve this one..

I kinda think we should just close this one until we finish the new docs site and can phase out all the old tooling which will allow us to resolve things like this for all our various bits of tooling.

PR #1701, opened 2022-04-08
yarn.lock was out of date in main

forrestbaer:
yarn.lock was out of date in master for some reason.. 

This one was created from a fresh clone of the ds with yarn install..

Also some visual regressions in healthcare?


pwolfert:
For some reason these visual differences don't show up for me on master with loki. @zarahzachz, could you look at the image updates in this PR and determine if they're more correct? The ones that change from black to blue make sense, but I'm not sure about the error-state borders. Ohh, I'm now seeing that these image reference changes are specifically for healthcare.

zarahzachz (approved):

PR #1705, opened 2022-04-11
[WNMGDS-1582] gatsby docs site frame

scheul93:
## Summary

Created a site frame for Gatsby doc site with
* Skip Nav Link
* USA Banner
* Header
  * Download code button will download the files we attach with the latest release
  * View on Github links to our Github repo

Screen Shot 2022-04-11 at 10 17 46 AM

## How to test

`yarn start:gatsby` will spin up the local build of the new doc site

## Notes
I tried to re-use but also simplify any doc site specific styles

Right now, I have environment variables for the current release version number. This points the 'download code' link to the appropriate place. I was thinking that as part of the Jenkins job, we could be provided the latest release number then update the .env.production file? If there is another way to get this info I'm open, but since the design system is not defined as a dependency for this project (instead we're using relative paths to access assets), this was my first idea


forrestbaer (approved):
Looks Good!

packages/docs/.env.production
```diff
@@ -0,0 +1 @@
+GATSBY_CORE_PKG_VERSION=3.2.2
```
pwolfert, line 1: Is the idea that in the actual production version we'd provide this via the command line instead of by this file?

packages/docs/src/components/DocSiteHeader.tsx
```diff
@@ -1,7 +1,34 @@
...
+            className="ds-c-button ds-c-button--primary ds-c-button--inverse ds-u-font-weight--normal"
+          >
+            {' '}
+            Download Code{' '}
+          </Button>
+          <Button
+            href="https://github.com/CMSgov/design-system"
+            className="ds-c-button ds-c-button--inverse ds-u-font-weight--normal ds-u-margin-left--2"
+          >
+            View on Github
```
pwolfert, line 26:
```suggestion
            View on GitHub
```

pwolfert (approved):

packages/docs/.env.production
scheul93: > Right now, I have environment variables for the current release version number. This points the 'download code' link to the appropriate place. I was thinking that as part of the Jenkins job, we could be provided the latest release number then update the .env.production file? If there is another way to get this info I'm open, but since the design system is not defined as a dependency for this project (instead we're using relative paths to access assets), this was my first idea
So it was either command line (which I'm not sure how Gatsby handles that) or I wasn't sure if our job could update this file based on the job's param

scheul93: I'll do some experimenting with the command line var...

pwolfert: Sorry, should have read the whole PR description 🤦‍♂️

pwolfert: Anything defined in this file should be definable in the shell, and Gatsby should automatically read the shell's environment variable

scheul93: 🤔 Upon further reading, I don't know if Gatsby works like that. I see [this documentation](https://www.gatsbyjs.com/docs/how-to/local-development/environment-variables/#:~:text=Gatsby%20has%20built%2Din%20support,variables%20from%20a%20file%20named%20.) where the `.env` files are recommended.

I also tried to update the start command to define an environment variable in the terminal command like
```
CORE_PKG_VERSION=3.2.1 yarn --cwd ./packages/docs develop
```
And got undefined in a component

scheul93: This may take some experimentation

packages/docs/package.json
```diff
@@ -35,5 +34,9 @@
     "@types/react": "^17.0.10",
     "@types/react-dom": "^17.0.10",
     "typescript": "^4.5.4"
+  },
+  "peerDependencies": {
+    "react": ">=16.0.0",
+    "react-dom": ">=16.0.0"
   }
```
scheul93, line 41: If react & react dom are listed in the `dependencies` section, React 18 gets installed and causes errors in other parts of the monorepo

PR #1706, opened 2022-04-12
Pwolfert/ie example

zarahzachz:
This is a working IE native modal example. 

To test in IE, you'll have to build the project locally and run the `react-app` index.html file in a VM (I used SauceLabs).

Steps:
1. At project root, run `yarn build` to make sure the design system's compiled correctly
2. In another terminal, cd into `examples/react-app` and `rm -rf dist node_modules` to make sure the previous design system files have been properly cleared
3. In the react-app example, run `yarn install && yarn build`
4. Run a local server on the react-app index.html file (I used VS Code's Live Server extension)

**Note:** You'll have to [create a tunnel](https://docs.saucelabs.com/secure-connections/sauce-connect/proxy-tunnels/) to run your local code in Sauce Labs


packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -334,86 +148,90 @@ export const Dialog = (props: DialogProps) => {
...
-    size,
-    title,
-    ...modalProps
-  } = props;
+  // Toggle dialog open
+  const dialogRef = useRef(null);
+  const lastActiveElement = React.useRef(null);
+  const firstRender = React.useRef(true);
+
+  useLayoutEffect(() => {
```
zarahzachz, line 156: Using `useLayoutEffect` because I think this is modifying the DOM. Still a little unclear on the difference between this hook and `useEffect`

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -334,86 +148,90 @@ export const Dialog = (props: DialogProps) => {
...
+    if (window.HTMLDialogElement === undefined) {
+      dialogPolyfill.registerDialog(dialogRef.current);
+      // import('dialog-polyfill/dist/dialog-polyfill.css');
+    }
+  }, []);
+
+  useLayoutEffect(() => {
+    // prevents calling imperative methods on mount since the polyfill will throw an error since we are not using the `open` attribute
+    if (firstRender.current) {
```
zarahzachz, line 165: I had to add some extra `ref` hooks to address an IE bug where IE was expecting the `open` attribute to exist on the `<dialog>` element. Since this doesn't exist until after the modal's triggered, there was an issue.

PR #1709, opened 2022-04-12
[NO-TICKET] resolving ts errors unearthed by Gatsby

scheul93:
## Summary

With the installation of Gatsby, some new TS errors were unearthed when running `yarn build`.

Mostly, it looks like some types needed to be defined. I tried my best for the i18n file, but the other components I used `any`. I'm sure there's some more specific typing that could happen.

## How to test

Run `yarn install` & `yarn build` and the doc site should build without error.


packages/docs/package.json
```diff
@@ -25,17 +25,15 @@
...
   },
   "peerDependencies": {
+    "@types/node": "^17.0.22",
     "@types/react": "^17.0.10",
     "@types/react-dom": "^17.0.10",
     "react": ">=16.0.0",
-    "react-dom": ">=16.0.0"
+    "react-dom": ">=16.0.0",
+    "sass": "^1.43.4",
+    "typescript": "^4.5.4"
```
pwolfert, line 37: I don't think we can technically put this one here in `peerDependencies` because the root-level package file doesn't have it as a dependency. If it is moved back, does it break things?

pwolfert (approved):
Thanks for fixing this! I looked into the `translate` errors and saw that it was catching a case when we could get unexpected results, so I added some code and unit tests around that.

packages/docs/package.json
scheul93: I don't think it will break anything, but I'll confirm

PR #1710, opened 2022-04-12
[WNMGDS-1557] Autocomplete theme

forrestbaer:
## Summary

Adds theme variables to the core theme for the Autocomplete Component, HC and Medicare inherit these styles.

## How to test

I think just a code review is fine. I've run full visual regressions on these changes and ran all tests, no difference between here and `component-variable-integration` where I recently updated styles based on the new color setup for tokens.


packages/design-system/src/styles/components/_Autocomplete.scss
```diff
@@ -30,29 +22,29 @@ $autocomplete-list-border: $autocomplete-list-border-width $autocomplete-list-bo
...
+  background-color: $autocomplete-list__background-color;
+  border: 1px solid $autocomplete-list__border-color;
   box-sizing: border-box;
   max-height: ($base-font-size + $spacer-2) * 10;
   overflow-y: auto;
-  padding: $autocomplete-space;
+  padding: $spacer-1;
   position: absolute;
   width: 100%;
   z-index: $z-dialog;
```
scheul93, line 37: [question] are only color variables being replaced? If not, this z-index seems like it could be a theme variable

forrestbaer: In most cases I'm just planning on color to start, I don't want to get too deep into customization right off the bat with all the changes happening at once and all..

forrestbaer: Do you think I should expose more to theming? I know in our audit we identified several things that could be themed but as I'm looking at an MVP for this month it seems like color (and cleaning up some of the existing styles to use tokens where they do not currently) will be the easiest way to get things started, I figure once this effort is launched we can start adding more customization on the theme level where needed without as big of a lift.

PR #1716, opened 2022-04-13
[WNMGDS-1588] gatsby desktop nav

scheul93:
## Summary

Adding dynamically generated Desktop nav using Vertical Nav component.
The nav is generated based on the structure of the `content` directory.

This includes creating some placeholder content files and using a `NUMBER_` preface to support ordering.

## How to test

`yarn install` + `yarn start:gatsby`

Screen Shot 2022-04-13 at 3 41 43 PM

## Still to come
* Once content is migrated, update nav component to show selected state
* Add urls / linking to each item


forrestbaer (approved):
👍🏻

packages/docs/src/components/DocSiteNav.tsx
```diff
@@ -1,9 +1,87 @@
...
+ * queries all MDX files and generates the proper format to create a vertical nav
+ * @returns {React Element}
+ * @todo figure out which item is currently selected and mark & expand appropriately
+ * @todo Determine URL string for each item once content is integrated
+ */
 const DocSiteNav = () => {
+  const data = useStaticQuery(graphql`
+    query SiteNavQuery {
+      allFile(
+        filter: { absolutePath: {}, extension: { eq: "mdx" } }
```
zarahzachz, line 24: I'm not well-versed in graphql, but I'm seeing `allMdx` as an option next to `allFile`. Would the be a more efficient query to start with since you're filtering the mdx files?

packages/docs/src/components/DocSiteNav.tsx
```diff
@@ -1,9 +1,87 @@
...
+        sort: { fields: [relativeDirectory, name] }
+      ) {
+        nodes {
+          id
+          name
+          relativeDirectory
+          relativePath
+        }
```
pwolfert, line 32: I was playing around with the GraphQL playground (didn't try Sarah's `allFile` -> `allMdx` idea though) and saw that there was a way to group the results. It wouldn't take away the need to do some kind of transformation to get valid `VerticalNavItemProps`, but it could possibly reduce the aggregating code:
```graphql
query SiteNavQuery {
  allFile(sort: {fields: [relativeDirectory, name]}) {
    group(field: relativeDirectory) {
      fieldValue
      edges {
        node {
          id
          name
          relativeDirectory
          relativePath
        }
      }
    }
  }
}
```

In this case, the result will have a `fieldValue` of `""` for the `intro` file where there is no relative directory, and all others will have the directory name inside `fieldValue`.

packages/docs/src/components/DocSiteNav.tsx
scheul93: I'm now well-versed either 😅 
Looking into it, it seems like `allMdx` is used to help bring back the content of MDX files along with some other information. The one piece of info that seems to be harder to retrieve in this query is the `relativeDirectory`. I think it's possible to include, but several layers deep.
Also related to Patrick's comment, I don't see a way to group by relativeDirectory, so I'm not sure if this query is both more efficient AND gives all the required info -- looking into it!

packages/docs/src/components/DocSiteNav.tsx
```diff
@@ -1,9 +1,87 @@
...
+   */
+  const formatNavItemLabel = (name: string): string => {
+    let newName = name.replace(/-/g, ' ');
+    newName = newName.replace(/\d+_/g, '');
+    return newName;
+  };
+
+  const formatNavItemData = ({ name, id }: NavItem) => ({
+    label: formatNavItemLabel(name),
+    url: '',
```
pwolfert, line 50: Will this part be filled in later?

pwolfert (approved):
LGTM ⛵🧭

packages/docs/src/components/DocSiteNav.tsx
scheul93: Yes, I'm not quite clear on how URLs will work on the site yet, so once we start adding content pages & content, we should be able to fill this in

PR #1717, opened 2022-04-13
[BETA-FIX] Fix missing Header export

pwolfert:
## Summary

One of the integration tests found a bug in this current beta release (healthcare version `7.3.0-beta.1`).

### Fixed

- Fixed missing export for `<Header>` component in `@cmsgov/ds-healthcare-gov`

## How to test

If you try to build the healthcare doc site, it outputs a bunch of warnings about missing exports, like
```
WARNING in ./packages/ds-healthcare-gov/dist/esnext/i18n.js 15:0-90
"export 'Language' was not found in '@cmsgov/design-system'
```
On master, `Header` is included in this list. I'm not really sure what to do about most of these warnings, but I was hoping they'd get resolved in the Gatsby version of the doc site. I was able to add some code to the `examples/create-react-app-typescript` example to import the `Language` type and use it, and it compiled, so I don't think it's actually a problem for downstream projects.


pwolfert:
I thought it might be helpful when looking at the git log to prefix it with `[BETA-FIX]`. Thoughts?

forrestbaer (approved):

PR #1723, opened 2022-04-14
[WNMGDS-1604] Simplify gathering of analytics content

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1604

Right now our components that send content to analytics have a bunch of verbose code around pulling text content out of the rendered html in the DOM. It's a lot of branching code that is copied between components and not very compatible with functional components. We need a much easier way of implementing analytics in components.

### Added

- Added a `useAnalyticsContent` hook

### Changed

- Updated unit tests for several components to
    1. Move to `testing-library`
    2. Make sure we cover analytics cases where content passed into components is made of React elements
- Implement the new hook where it's appropriate. (It's mostly for simplifying the process when we add it to other elements that don't currently have analytics.)

### Fixed

- Fixed analytics events not firing when `HelpDrawer` `heading` value is non-string

## How to test

- Run unit tests
- You can also modify Storybook stories to enable analytics by
    - Calling appropriate flag-setting function
    - Removing `analytics: false` arg on story if it exists
    - Defining `window.utag = { link: console.log }`


forrestbaer (approved):
Tests and all other system tasks ran without issue, analytics events fired in storybook testing, can't find anything to speak of with code, nice execution and comments! 👍🏻

PR #1729, opened 2022-04-18
Wnmgds 1556 theme alert

forrestbaer:
## Summary

- Added theme styling for the Alert Component
- Updated loki snapshots for medicare, the Alert components were slightly off
- Removed choice reference items from medicare as it inherits these from core 

## How to test

`yarn loki test` `yarn loki:healthcare test` you can compare the output of `yarn storybook:medicare` with the same in `component-variable-integration` or just look at the screenshots and see they aren't really different from where they're going.


pwolfert (approved):
Reviewed all the loki reference changes 👍

PR #1733, opened 2022-04-19
[WNMGDS-1583] Gatsby footer

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1583)

Implement and style Gatsby footer to match current footer in PROD (desktop and mobile).

Some small design tweaks have been made:

Footer exists as two parts, gray and blue
- Gray part changes
    - Mobile: Feedback buttons have been adjusted to be full width (previously were width of content, appearing staggered)
    - Desktop: Feedback buttons now fill their column. This addresses the odd gaps that previously existed on them
- Blue part changes
    - Legal/address text adheres to line-length rules
- For entire footer, changed case of links to sentence case (unless it's referencing a law or something)

## How to test

Pull down this branch and run `yarn install && yarn start:gatsby`. Compare footers in local and PROD environments.


packages/docs/src/helpers/githubUrl.js
```diff
@@ -0,0 +1,10 @@
+import join from 'url-join';
+
+const cmsdsGithubUrl = 'https://github.com/CMSgov/design-system';
+
+function githubUrl(pathname = '') {
+  const ghUrl = process.env.githubUrl || cmsdsGithubUrl;
```
scheul93, line 6: [suggestion] Consider removing this `ghUrl` variable. It seems like this might have been used originally when the child DS's were separate repos (see [this example](https://github.com/CMSgov/hcgov-design-system/blob/master/cmsds.config.js#L12)). Now that we're in a monorepo, this url should always be `'https://github.com/CMSgov/design-system'`

packages/docs/src/components/DocSiteFooter.tsx
```diff
@@ -1,7 +1,128 @@
...
+      </dd>
+    );
+  });
+}
 
 const DocSiteFooter = () => {
-  return <footer className="c-footer">Placeholder footer content here</footer>;
+  return (
+    <footer className="c-footer">
+      <section className="c-footer__content--top">
```
scheul93, line 50: [nitpick] Consider updating this class name to `c-footer__content c-footer__content--top` to be more inline with BEM. 

The current class name makes the most sense if the class `c-footer__content` also existed on this element since `--top` is implying a modification of the `c-footer__content` class.

[In the scss file](https://github.com/CMSgov/design-system/pull/1733/files#diff-c9decc8d17c1f72ac7135030f6d0ba091178b70dc5137580606b8e1bc63d5578R9), it looks like there are styles that apply to both `c-footer__content--bottom` & `c-footer__content--top`. These styles could be instead scoped to the class `c-footer__content`

packages/docs/src/components/DocSiteFooter.tsx
```diff
@@ -1,7 +1,128 @@
...
+              target="_blank"
+              rel="noreferrer"
+              className="ds-c-button"
+            >
+              Propose something new
+            </a>
+          </div>
+        </div>
+      </section>
+      <section className="c-footer__content--bottom ds-base--inverse">
```
scheul93, line 87: [nitpick] Consider updating this class name to `c-footer__content c-footer__content--bottom` to be more inline with BEM. 

The current class name makes the most sense if the class `c-footer__content` also existed on this element since `--bottom` is implying a modification of the `c-footer__content` class.

[In the scss file](https://github.com/CMSgov/design-system/pull/1733/files#diff-c9decc8d17c1f72ac7135030f6d0ba091178b70dc5137580606b8e1bc63d5578R9), it looks like there are styles that apply to both `c-footer__content--bottom` & `c-footer__content--top`. These styles could be instead scoped to the class `c-footer__content`

packages/docs/src/styles/components/footer.scss
```diff
@@ -1,2 +1,86 @@
 .c-footer {
+  .ds-footer__container {
+    margin-inline: auto;
+    max-width: $site-max-width;
+    width: 100%;
+  }
+}
+
+.c-footer__content--top,
+.c-footer__content--bottom {
```
scheul93, line 10: As a continuation of my comments in the React component, consider scoping these styles to a `c-footer__content` class. This would be more inline with BEM naming

packages/docs/src/styles/components/footer.scss
```diff
@@ -1,2 +1,86 @@
...
+
+    @media (min-width: $width-md) {
+      grid-template-columns: repeat(3, minmax(max-content, 33.3%));
+    }
+  }
+}
+
+.c-footer__content--bottom {
+  background-color: $color-primary-darkest;
+  color: $color-white;
```
scheul93, line 37: [question] how do these styles differ from the styles that `ds-base--inverse` provide? Can one or the other be used?

scheul93 (approved):
LGTM

PR #1734, opened 2022-04-20
[NO-TICKET] Gatsby path prefix

pwolfert:
## Summary

Read an optional `PATH_PREFIX` environment variable inside the Gatsby config file to allow for using a custom path in our demo sites.

## How to test

Run `PATH_PREFIX='/hello-world' PREFIX_PATHS=true yarn build:gatsby` and check to make sure the resources in `docs/public/index.html` are prefixed with `/hello-world`

### DIFF
--- a/packages/docs/gatsby-config.ts
+++ b/packages/docs/gatsby-config.ts
@@ -38,4 +38,8 @@ const config: GatsbyConfig = {
   ],
 };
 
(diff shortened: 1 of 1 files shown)


scheul93 (approved):
I think something was cached because this change didn't work on my machine first-try, but after opening the `gatsby-config`, it worked 🤷🏻‍♀️

PR #1739, opened 2022-04-21
[WNMGDS-1589] gatsby mobile nav

scheul93:
## Summary

Adding styling for the mobile nav for Gatsby doc site.

### Added

* Mobile menu button

### Changed

* Renaming `DocSiteNav` to `DocSiteSidebar`
* Made a `GithubLinks` component for reuse in desktop & mobile 

## How to test

`yarn install` & `yarn start:gatsby`. Change the screen size and confirm that nav switches between mobile and desktop views
DocSiteMobileMenu


packages/docs/src/components/GithubLinks.tsx
```diff
@@ -0,0 +1,37 @@
...
+   */
+  onDark?: boolean;
+}
+
+/**
+ * GithubLinks - a reusable component for Github links for downloading code & the Github repo
+ */
+const GithubLinks = ({ onDark }: GithubLinks) => {
+  return (
+    <>
```
forrestbaer, line 17: Might be nice to have the github logo in here? I guess the icon isn't a part of our set maybe not :shrug: Just always something I look for instinctively.

packages/docs/src/components/GithubLinks.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
+import { Button } from '@cmsgov/design-system';
+import classnames from 'classnames';
+
+interface GithubLinks {
+  /**
+   * Describes if the links are on a dark background
+   */
+  onDark?: boolean;
```
forrestbaer, line 9: I'm thinking this might be defined at the theme/css level during a swap and may not be necessary?

packages/docs/src/styles/components/footer.scss
```diff
@@ -32,6 +32,12 @@
 }
 
 .c-footer__content--bottom {
+  padding-bottom: $spacer-7 + $spacer-2;
```
forrestbaer, line 35: Wonder if I should add more spacer values to tokens :thinking:

packages/docs/src/styles/components/sidebar.scss
```diff
@@ -1,8 +1,41 @@
 .c-sidebar {
-  border-right: 1px solid $color-gray-lighter;
-  max-width: 15.625rem;
+  // leaving space for menu button
+  bottom: $spacer-2 + $spacer-2 + ($input-line-height * $base-font-size);
```
forrestbaer, line 3: `$spacer-4` here?

forrestbaer (approved):
All looks good, works well for me in testing, just a few comments. 👍🏻

packages/docs/src/components/GithubLinks.tsx
scheul93: I agree. Right now, I'm porting over the exact button from the current site, but I think there have been (multiple) requests to use the GH logo when we redesign!

packages/docs/src/components/GithubLinks.tsx
scheul93: I included this because this component is used both on in the desktop header (which has a dark background) and in the mobile menu (which has a light background). I needed a way to apply slightly different button styles based on those two uses.

It's not related to the whole site's theme.

packages/docs/src/components/GithubLinks.tsx
```diff
@@ -0,0 +1,37 @@
+import React from 'react';
```
zarahzachz, line 1: Looks like we're using React 17 for this project? According to docs, you [don't have to import React](https://reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html#removing-unused-react-imports) if you aren't using hooks, so this could probably be removed

zarahzachz: (This might apply to other files in this PR.)

packages/docs/src/styles/components/footer.scss
zarahzachz: Not sure - this looks like a 1-off thing in our designs, so I don't know if it warrants tokenizing (or using the spacer vars at all)

packages/docs/src/components/GithubLinks.tsx
scheul93: I may be missing it, but where do you see that we are using react 17? In the yarn.lock I see the react versions listed as
```
react@16.14.0, react@^16.14.0:
  version "16.14.0"
  resolved "https://registry.yarnpkg.com/react/-/react-16.14.0.tgz#94d776ddd0aaa37da3eda8fc5b6b18a4c9a3114d"
  integrity sha512-0X2CImDkJGApiAlcf0ODKIneSwBPhqJawOa5wCtKbu7ZECrmS26NvtSILynQ66cgkT/RJ4LidJOc3bUESwmU8g==
  dependencies:
    loose-envify "^1.1.0"
    object-assign "^4.1.1"
    prop-types "^15.6.2"
```

scheul93: Also, when I try to remove `import React from 'react'` I see this eslint error
```
'React' must be in scope when using JSXeslint react/react-in-jsx-scope
```
Is that an eslint rule we should change for just the `docs` directory?

zarahzachz: My bad - thought I saw it in the package file, but it was for React _types_, not React. Disregard!

zarahzachz (approved):

PR #1742, opened 2022-04-21
[BETA-FIX] Fix errant TS type in compiled esnext modules

pwolfert:
## Summary

#1740 failed to fix the actual error. It wasn't a type-checking error, which explains why I wasn't able to reproduce it with TypeScript. If I were to have looked closer at the following error, I would have seen that it was coming from a plain JS module the `esnext` directory.
```
./node_modules/@cmsgov/ds-healthcare-gov/dist/esnext/i18n.js
Attempted import error: 'Language' is not exported from '@cmsgov/design-system'.
```
After messing with the build a bit, I got it to print the following error:
```
The exported identifier "Language" is not declared in Babel's scope tracker
as a JavaScript value binding, and "@babel/plugin-transform-typescript"
never encountered it as a TypeScript type declaration.
It will be treated as a JavaScript value.

This problem is likely caused by another plugin injecting
"Language" without registering it in the scope tracker. If you are the author
 of that plugin, please use "scope.registerDeclaration(declarationPath)".
The exported identifier "Language" is not declared in Babel's scope tracker
as a JavaScript value binding, and "@babel/plugin-transform-typescript"
never encountered it as a TypeScript type declaration.
It will be treated as a JavaScript value.

This problem is likely caused by another plugin injecting
"Language" without registering it in the scope tracker. If you are the author
 of that plugin, please use "scope.registerDeclaration(declarationPath)".
 [15:43:53] compileJs SyntaxError in plugin "gulp-babel"
Message:
    /Users/patrickwolfert/Projects/design-system/packages/ds-healthcare-gov/src/components/i18n.ts: Exporting local "Language", which is not declared.
  24 | addTranslations('es', es);
  25 |
> 26 | export { Language, TFunction, getLanguage, languageMatches, t, translate, tWithLanguage };
     |          ^^^^^^^^
  27 |
 ```
What seemed to fix it was separating out `Language` and `TFunction` into its own import statement, using the `type` keyword. It's not supposed to be necessary in modern TypeScript, but that seems to have fixed it. I've added steps for verifying below.

### Fixed

- Fix attempted type import from our compiled plain JavaScript modules, which would cause a build-time error downstream

## How to test

1. Check out dd3d650d
2. `yarn install && yarn build && yarn build:healthcare`
3. Go to `examples/create-react-app-typescript`
4. `yarn install && yarn build`. Notice that it will throw an error
5. Clear `examples/create-react-app-typescript/node_modules` and then check out 3412232bd8260307434d2b8556a84492fef2a683 (the head of this branch)
6. In the root, `yarn install && yarn build:healthcare`
7. Back in `examples/create-react-app-typescript`, `yarn install && yarn build`. There should be no errors this time

### DIFF
--- a/examples/create-react-app-typescript/package.json
+++ b/examples/create-react-app-typescript/package.json
@@ -3,14 +3,14 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@cmsgov/design-system": "^2.0.0",
+    "@cmsgov/ds-healthcare-gov": "file:../../packages/ds-healthcare-gov",
     "@types/node": "^12.0.0",
     "@types/react": "^16.9.0",
     "@types/react-dom": "^16.9.0",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
     "react-scripts": "3.4.3",
-    "typescript": "^3.9.10"
+    "typescript": "^4.2.4"
   },
   "scripts": {
     "start": "react-scripts start",
--- a/examples/create-react-app-typescript/package.json
+++ b/examples/create-react-app-typescript/package.json
@@ -3,14 +3,14 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
-    "@cmsgov/ds-healthcare-gov": "file:../../packages/ds-healthcare-gov",
+    "@cmsgov/design-system": "^2.0.0",
     "@types/node": "^12.0.0",
     "@types/react": "^16.9.0",
     "@types/react-dom": "^16.9.0",
     "react": "^16.13.1",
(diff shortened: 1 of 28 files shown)


forrestbaer:
I'm not getting the build error during building of the example project in dd3d650:shrug: Just a few minor warnings. `yarn type-check` ran fine at that commit as well. At the second commit I didn't notice any difference aside from a slightly smaller set of build files. Not sure what I'm missing that may have caused it to compile correctly in the first commit but the second commit passes all my tests and works just as well so I'll throw a :+1: for this one!

forrestbaer (approved):

PR #1751, opened 2022-04-22
[WNMGDS-1595] info content migration

scheul93:
## Summary

* Copying content pages from `getting started` & `guidelines` sections of core doc site
  * Content is the same, I did update internal links to use relatives path to be used in a gatsby `<Link>` component and if there was markup for an alert, I used an actual DS `<Alert>` component 
* I added an `InfoPage` component that is used to format information pages
* I updated `gatsby-node.js` to generate pages using Gatsby APIs. These seemed like the best option for [creating routes](https://www.gatsbyjs.com/docs/reference/routing/creating-routes/#creating-routes)
* Now that pages are being generated, I updated linking in the vertical nav so that you can. click through and access real page content.

## Still TODO
* I need to include syntax highlighting for code snippets included in content -- either with prismjs or a gatsby plugin

## How to test

`yarn start:gatsby`. Click around and view content pages!


packages/docs/content/1_getting-started/1_using-sass-and-css.mdx
```diff
@@ -0,0 +1,113 @@
+---
+title: Using Sass and CSS
```
pwolfert, line 2: I noticed that in the nav menu the case of "CSS" isn't maintained
Screen Shot 2022-04-22 at 3 13 21 PM

packages/docs/content/1_getting-started/1_using-sass-and-css.mdx
```diff
@@ -0,0 +1,113 @@
...
+
+```
+SASS_PATH="node_modules"
+```
+
+If you have your own custom Webpack configuration, you need to use [`file-loader`](https://webpack.js.org/loaders/file-loader/) and [`resolve-url-loader`](https://github.com/bholloway/resolve-url-loader) to copy assets into your build folder and inject the correct URLS to your build files.
+
+## Applying styles to your page
+
+Once your page is loading the design system's CSS, you can then begin applying its styling to your pages. Below is an example of a project applying the [base-level of styles](/styles/base) and a [utility class](/utilities/overview).
```
pwolfert, line 96: These relative urls are so much nicer to deal with ✨

packages/docs/content/1_getting-started/2_using-components.mdx
```diff
@@ -0,0 +1,85 @@
...
+  <div class="ds-c-alert__body">
+    <h3 class="ds-c-alert__heading">Warning status</h3>
+    <p class="ds-c-alert__text">Alert warning body text</p>
+  </div>
+</div>
+```
+
+More HTML/CSS examples and guidance can be found on component documentation pages.
+
+<h2 id="examples" class="ds-h2 ds-u-color--primary-darker">Examples</h2>
```
pwolfert, line 74: Any idea why these two headings are special?

pwolfert: It looks like the old page was this way, but I think we should probably normalize them. Not sure why they have the primary-darker color applied

packages/docs/content/2_guidelines/3_design-principles.mdx
```diff
@@ -0,0 +1,70 @@
+---
+title: Design Principles
+parentNavItem: guidelines
```
pwolfert, line 3: Is there a way to avoid having to include this `parentNavItem` information here?

packages/docs/content/2_guidelines/5_internationalization.mdx
```diff
@@ -0,0 +1,48 @@
...
+import { getLanguage, setLanguage } from '@cmsgov/design-system';
+
+// Set the design system language to something other than the document's detected language
+setLanguage('es');
+
+// Get the design system's current language
+console.log(getLanguage());
+```
+
+<Alert heading="Deprecated per-component langauge props" variation="warn" className="ds-u-margin-top--3">
```
pwolfert, line 44: This is pretty cool

packages/docs/gatsby-node.js
```diff
@@ -13,3 +13,43 @@ exports.onCreateWebpackConfig = ({ stage, actions }) => {
...
+  `).then((result) => {
+    if (result.errors) {
+      throw result.errors;
+    }
+
+    // Create blog post pages.
+    result.data.allMdx.edges.forEach((edge) => {
+      createPage({
+        // Path for this page -- the slug with positioning markers removed
+        path: edge.node.slug.replace(/\d+_/g, ''),
```
pwolfert, line 46: When we loop through these, are these _edges_ guaranteed to be in alphanumeric order? And that's why we don't need to extract that order prefix from this? And does the order in which we call `createPage` matter?

packages/docs/src/components/DocSiteSidebar.tsx
```diff
@@ -26,11 +29,23 @@ interface DocSiteNavProps {
 }
 
 /**
- * DocSiteSidebar
+ * A wrapper for the gatsby link to handle internal sight navigation
```
pwolfert, line 32:
```suggestion
 * A wrapper for the gatsby link to handle internal site navigation
```
🙂

packages/docs/src/components/DocSiteSidebar.tsx
```diff
@@ -100,7 +115,7 @@ const DocSiteSidebar = ({ isMobileNavOpen }: DocSiteNavProps) => {
         'c-sidebar--open': isMobileNavOpen,
       })}
     >
-      <VerticalNav className="c-nav" items={navItems} />
+      <VerticalNav className="c-nav" items={navItems} component={GatsbyLink} />
```
pwolfert, line 118: Wow, the `component` prop came in handy! Clever

pwolfert (approved):
It looks like a finished site now! I have some questions and a couple comments below, but everything else LGTM 👍

packages/docs/content/2_guidelines/3_design-principles.mdx
scheul93: Ah! This was leftover from a previous implementation attempt. It looks like I can safely remove this now

packages/docs/gatsby-node.js
scheul93: I don't think the order that we call `createPage` matters since it is just to create routes for the site. The ordering matters more in `DocSiteSidebar`.

This string replace is to ensure that the alphanumeric ordering of the file names (which is most important in the nav creation) doesn't end up in the url path

packages/docs/content/1_getting-started/1_using-sass-and-css.mdx
scheul93: Fixed. I think that was because these nav labels are using the file name and it wasn't capitalized there

PR #1754, opened 2022-04-25
[NO-TICKET] RFC - Using native dialog

zarahzachz:
## Summary

Write a summary for using the `<dialog>` element within `<Drawer>` and `<Dialog>`. 

Related to discovery work done in #1685.


rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Benefits
+
+By relying on the browser's implementation, the Design System stands to inherit several benefits:
+
+- Basic functional and accessibility enhancements become the browser's problem
+- Less reliance on third-parties, results in
+  - Publishing less code (better performance)
+  - A more flexible implementation (no longer beholden to someone else's API)
+  - Less Engineering time and effort spent updating
```
scheul93, line 78: 🙌🏻

rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,102 @@
...
+
+### 1. Polyfill support
+
+[Windows support of IE will end on June 15, 2022](https://docs.microsoft.com/en-us/lifecycle/announcements/internet-explorer-11-end-of-support). In alignment with other websites, and to encourage Users to update their browsers, I recommend removing all IE-specific polyfills following this date.
+
+**tldr; The addition of this polyfill is a temporary concern.**
+
+### 2. One-off bugs may be discovered
+
+The discovery of using the `<dialog>` element in the Design System was done using the pre-existing Storybook examples and documentation pages. As a result, there is a chance consuming teams have specific use-cases not documented in our system that this implementation will expose.
```
scheul93, line 95: Should it be mentioned that teams will have the ability to test these changes with our Beta release cycle to give them the opportunity to flag items early? It will be on teams to do this testing, but they have the opportunity to flag issues earlier rather than later

rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,102 @@
...
+
+## Questions and Requested Feedback
+
+- How does the Design System demarcate internal, not for public-consumption components?
+  - Is prefixing a component with an underscore, as in `<_Dialog>`, acceptable? Or do we prefix the name with something like, `<InternalDialog>`?
+  - Should the system have a special directory for these components?
```
scheul93, line 101: I don't mind the `_Dialog` naming convention. The `_` seems like a standard way the industry denotes something that is internal.
I also wouldn't mind having an `Internal` directory that this (and future internal components) live in to really drive the point home.

rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,102 @@
...
+
+A robust [polyfill](https://github.com/GoogleChrome/dialog-polyfill), maintained by Chrome, is available to ensure support of `<dialog>` for browsers that aren't actively being maintained (IE) or haven't yet been updated (older versions of Firefox and Safari).
+
+Given the pain points listed above and the overlap of functionality between `<Dialog>` and `<Drawer>`, it's suggested the Design System lessen its reliance on third-party libraries by using the native `<dialog>` element for these components instead.
+
+The recommended implementation is to create one internal wrapping `<_Dialog>` component, which houses the `<dialog>` element and provides props that control styling and User interaction (i.e., if background content remains interactive). This component would also manage elevation and focus trapping behavior.
```
scheul93, line 30: Would the internal `_Dialog` contain any default styling? Or would that be left to the exported `<Dialog>` & `<Drawer>` to handle?

pwolfert: Yeah ☝️ that's good to call out.

rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,102 @@
...
+
+Implementation of this component should not change the current component APIs for `<Drawer>` and `<Dialog>`.
+
+The reasoning behind creating one parent component is as follows:
+
+- Code is maintained in a singular file; DRY implementation
+- The `<dialog>` spec has a limited number of methods available, making the complexity of this component minimal
+  - `close()` will close the dialog
+  - `show()` will open the dialog, but background elements will remain interactive
+  - `showModal()` will open the dialog as a modal, background elements will **not** be interactive
```
pwolfert, line 40: This bullet point has another reason baked into it: We want to turn an imperative API into a declarative API. That might be worth calling out as well, even though it's implied when say we're wrapping it with React.

rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,102 @@
...
+
+Simplified example of implementation:
+
+```js
+const _Dialog = (props) => {
+  <dialog className={customClass} {...dialogProps}>
+    {children}
```
pwolfert, line 47: I like these examples. They're concise and clear.. I think it might be worth adding more of your comments in this top one here about what all might be implemented in this component like "/* Hooks for mapping props to imperative dialog API functions */"

rfcs/2022-04-using-native-dialog.md
pwolfert: Also, to avoid confusion between our component called `Dialog` and because it will live in a top-level folder as a sibling of `Dialog`, I do think we actually name it something like `NativeDialog`. I don't have a _strong_ opinion, but I think underscored internal things should only exist within a module. I don't think there's any reason why we couldn't expose `NativeDialog` in our package, especially if it could be useful for child design systems.

pwolfert (commented):
Nice work!

rfcs/2022-04-using-native-dialog.md
zarahzachz: So should the component structure/name look like the following:

```
design-system/
|-- src/
|-- components/
|   |-- Dialog/
|   |-- Internal/
|   |   |-- NativeDialog/
|   |   |   |-- NativeDialog.tsx
```

scheul93: Per Patrick's comment, if we are going to expose `NativeDialog` in our package, then I think it should just live along with the other components, like
```
design-system/
|-- src/
|    |-- components/
|    |    |-- Dialog/
|    |    |-- NativeDialog/
|    |    |    |-- NativeDialog.tsx
```

my original proposal was only if we weren't going to make `NativeDialog` available to consumers

pwolfert (approved):
This looks ready for implementation to me 👍

scheul93 (approved):

rfcs/2022-04-using-native-dialog.md
forrestbaer: Really simple implementation I love it

rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,107 @@
...
+### 2. One-off bugs may be discovered
+
+The discovery of using the `<dialog>` element in the Design System was done using the pre-existing Storybook examples and documentation pages. As a result, there is a chance consuming teams have specific use-cases not documented in our system that this implementation will expose.
+
+To mitigate this risk, consuming teams can test this update using the Design System's beta release and can flag any potential problems to be resolved before this implementation is officially released.
+
+## Questions and Requested Feedback
+
+- ~How does the Design System demarcate internal, not for public-consumption components?~ Resolved in PR feedback.
+  - ~Is prefixing a component with an underscore, as in `<NativeDialog>`, acceptable? Or do we prefix the name with something like, `<InternalDialog>`?~ Resolved in PR feedback.
```
forrestbaer, line 105: This works for me.

rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,107 @@
...
+
+### 2. One-off bugs may be discovered
+
+The discovery of using the `<dialog>` element in the Design System was done using the pre-existing Storybook examples and documentation pages. As a result, there is a chance consuming teams have specific use-cases not documented in our system that this implementation will expose.
+
+To mitigate this risk, consuming teams can test this update using the Design System's beta release and can flag any potential problems to be resolved before this implementation is officially released.
+
+## Questions and Requested Feedback
+
+- ~How does the Design System demarcate internal, not for public-consumption components?~ Resolved in PR feedback.
```
forrestbaer, line 104: Doc site big red alert, next major release removal.

rfcs/2022-04-using-native-dialog.md
```diff
@@ -0,0 +1,107 @@
...
+  - ~Is prefixing a component with an underscore, as in `<NativeDialog>`, acceptable? Or do we prefix the name with something like, `<InternalDialog>`?~ Resolved in PR feedback.
+  - ~Should the system have a special directory for these components?~ Resolved in PR feedback.
```
forrestbaer, line 106: I think they should go with all the others.

forrestbaer (approved):
Good stuff 👍🏻

PR #1755, opened 2022-04-25
[NO-TICKET] gatsby code snippet formatting

scheul93:
## Summary

Implemented prismjs (what we are using on current doc site) to do syntax highlighting in code blocks coming from markdown.

The changes to the `syntax highlighting` styles use Eric's accessible syntax highlighting themes from [this repo](https://github.com/ericwbailey/a11y-syntax-highlighting)

## How to test

`yarn install` & `yarn start:gatsby`

In this example, you can see syntax highlighting for javascript & html

Screen Shot 2022-04-25 at 1 14 38 PM


packages/docs/src/styles/components/syntax-highlighting.scss
```diff
@@ -1,60 +1,180 @@
...
 pre[class*='language-'] {
+  padding: 1em;
+  margin: 0.5em 0;
   overflow: auto;
+  border-radius: 0.3em;
+}
+
+:not(pre) > code[class*='language-'],
+pre[class*='language-'] {
+  background: #fefefe;
```
pwolfert, line 36: @ericwbailey, I like that we're using your theme! I noticed on the white pages that there isn't really any contrast between the normal page background and our code blocks (`#fff` vs `#fefefe`). Do you have any design ideas for how we might show that it's a code block? Maybe an outline with a "Code" label or something? Also @line47

pwolfert (commented):
It's so cool that we're just rendering this to HTML with CSS and don't have to have a whole big package in a JS bundle that can parse various languages and create ASTs and stuff in the client. We're actually compiling the code at build time 😂

packages/docs/src/styles/components/syntax-highlighting.scss
line47: I like the idea of adding a 1px border using `$color-gray-lighter` 1px solid #D6D7D9

ericwbailey: Plus one to @line47's idea!

pwolfert (approved):
It looks great! I like that soft border

PR #1756, opened 2022-04-25
[WNMGDS-1580] Add default Spanish content

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1580

This is a larger feature branch for the work of migrating our components that have default English content to use our i18n module. That combined with the addition of Spanish translations will give us default content that can be in English or Spanish based on the global language setting.

Check the following boxes as PRs get merged into this branch:
- [x] All the icons
- [x] `Accordion`
- [x] `Autocomplete`
- [x] `DateField`
- [x] `Dialog`
- [x] `Drawer` and `HelpDrawer`
- [x] `FilterChip`
- [x] `MonthPicker`
- [x] `Pagination`
- [x] `Review`
- [x] `SkipNav`
- [x] `Spinner`
- [x] `VerticalNav`

### Added

- Added default Spanish content for the following components that had default English content:
    - All the icons
    - `Accordion`
    - `Autocomplete`
    - `DateField`
    - `Dialog`
    - `Drawer` and `HelpDrawer`
    - `FilterChip`
    - `MonthPicker`
    - `Pagination`
    - `Review`
    - `SkipNav`
    - `Spinner`
    - `VerticalNav`


forrestbaer (approved):
Looks good! Everything passes.. I added `document.documentElement.setAttribute('lang', 'es')` to the `.storybook/preview-body.html` in the script tag and the spanish translations showed up in storybook.

pwolfert:
I do want to add a way to easily change the language in Storybook. I've been looking into it on and off, but I haven't come up with a good solution yet. I know there are plugins out there, but last I checked they seemed to have implementation that was too specific.

pwolfert:
The only way to get it in storybook may be to use the React context API. Maybe we need to have a context-API option for setting language, and there's an order of precedence for the language setting like Global > Context > Component (where Component is currently deprecated)

forrestbaer:
> The only way to get it in storybook may be to use the React context API. Maybe we need to have a context-API option for setting language, and there's an order of precedence for the language setting like Global > Context > Component (where Component is currently deprecated)

Would a storybook plugin help? https://storybook.js.org/addons/storybook-addon-i18n  would that require changing our current i18 api? I haven't looked deeply at this yet..

pwolfert:
I actually have something working now on `pwolfert/storybook-lang-switcher`. I looked into that plugin and several others

PR #1757, opened 2022-04-25
[WNMGDS-1580] Use i18n for components A-F

pwolfert:
## Summary

## How to test

`yarn test:unit`

### DIFF
--- a/packages/design-system/src/components/FilterChip/FilterChip.tsx
+++ b/packages/design-system/src/components/FilterChip/FilterChip.tsx
@@ -2,6 +2,7 @@ import { CloseIcon, CloseIconThin } from '../Icons';
 import React from 'react';
 import classNames from 'classnames';
 import uniqueId from 'lodash/uniqueId';
+import { t } from '../i18n';
 
 const actionableKeys = ['Enter', 'Space', 'Backspace', 'Delete'];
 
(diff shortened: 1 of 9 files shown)


packages/design-system/src/components/FilterChip/FilterChip.tsx
```diff
@@ -81,6 +78,9 @@ export class FilterChip extends React.Component<FilterChipProps> {
         className={buttonClassNames}
         onClick={this.handleClick}
         onKeyDown={this.handleKeyDown}
+        aria-label={`${label} . ${
+          ariaClearLabel ?? t('filterChip.ariaClearLabel')
+        } ${label} filter`}
```
scheul93, line 83: does the word 'filter' need to be localized too?

pwolfert: Ah, yep, good catch

pwolfert: The word order might even be different in Spanish for the phrase "_something_ filter"

pwolfert: Got translations for this and updated the code

scheul93 (approved):

PR #1758, opened 2022-04-25
[WNMGDS-1580] Use i18n for icon components

pwolfert:
## Summary

## How to test

`yarn test:unit`


scheul93 (commented):
There's also two sneaky icons used in accordion that may need localization. Not sure if you want to include it as part of this PR, but they are

[Remove Icon](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/Accordion/AccordionItem.tsx#L104)
[Add Icon](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/Accordion/AccordionItem.tsx#L108)

pwolfert (commented):
@scheul93, nice job catching those sneaky icons. I'll address those in a new PR

scheul93 (approved):

PR #1760, opened 2022-04-25
[WNMGDS-1580] Use i18n in Pagination and migrate tests to testing-library

pwolfert:
## Summary

## How to test

`yarn test:unit`


scheul93 (approved):
Yay testing library!

PR #1773, opened 2022-04-28
[NO-TICKET] Storybook language switcher

pwolfert:
## Summary

After implementing https://github.com/CMSgov/design-system/pull/1756, I decided to go back and try to set up an easy way to switch between languages in our Storybook.

### Added

Added a language-switching button in our Storybook toolbar

## How to test

- `yarn storybook` - Go to the `UsaBanner` story
- `yarn build && yarn storybook:healthcare` - Check out the `UsaBanner` story, but also check out the healthcare.gov `Header`

## Feedback

My implementation does a React no-no in that it calls a function with side-effects inside a render function. Until https://github.com/CMSgov/design-system/commit/e2736bee7872f71a22be71ebdb8ab4935e251bf6, I was doing it "the right way", but the right way was more complicated, more verbose, and less performant (I surmise at least). I decided to simplify. But would ya'll have kept it as it was?

### DIFF
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -1,5 +1,5 @@
 import './storybookStyles.scss';
-import React, { useEffect } from 'react';
+import React, { useLayoutEffect } from 'react';
 import { setLanguage } from '../packages/design-system/src/components/i18n';
 import { setLanguage as setLanguageFromPackage } from '@cmsgov/design-system';
 import { addons } from '@storybook/addons';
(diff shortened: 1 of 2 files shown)


pwolfert:
I figured out how to do this after reading https://storybook.js.org/docs/react/essentials/toolbars-and-globals.

Other plugins I looked into were:
- https://storybook.js.org/addons/storybook-i18n
- https://storybook.js.org/addons/storybook-addon-i18n
- And a few that turned out to be i18n-library specific

forrestbaer (approved):
I like this! Works well and keeps language between pages nicely, I prefer this implementation even though it's a 'no-no' .. 👍🏻

PR #1774, opened 2022-04-28
[WNMGDS-1570] StepList component theme

forrestbaer:
## Summary

Added StepList component styling and updated steplist to do away with all the computations as those would be replaced with single token config variables in the future and they were unnecessary. Also set up usage of spacer tokens for padding/margin and other token variables where able.

## How to test


packages/design-system/src/styles/components/_StepList.scss
```diff
@@ -28,36 +14,36 @@ $current-step-color: $color-primary !default;
...
-  margin-bottom: $step-margin;
-  padding-bottom: $step-margin;
+  margin-bottom: $spacer-2;
+  padding-bottom: $spacer-2;
   position: relative;
 
   &::before {
-    border: $step-list-number-border-size solid $color-muted;
-    border-radius: 50%;
+    border: 2px solid $steplist-step__border-color--default;
```
pwolfert, line 24: Should the `2px` here use a theme variable?

packages/design-system-tokens/src/themes/medicare/defaultTheme.ts
```diff
@@ -248,6 +248,16 @@ const components = {
   'pagination-page-count__color': themeColors.gray,
   // review
   'review__border-color': themeColors['border'],
+  // steplist
+  steplist__color: themeColors.muted,
+  'steplist__color--current': themeColors.primary,
+  'steplist__background-color--current': themeColors.primary,
+  'steplist-step__color': themeColors.muted,
+  'steplist-step__border-color': themeColors.border,
+  'steplist-step__border-color--default': themeColors.muted,
```
pwolfert, line 257: It took me a while to figure out that these border color variables had nothing to do with each other—that they're styling two different kinds of elements (the dividing line and the step number). I don't think there's anything you can do to reduce that confusion, but I'm just noting that. The `--state` naming convention does give you a clue because there are other variables with that postfix.

packages/design-system/src/styles/components/_StepList.scss
pwolfert: I guess we only need to add that customizability if someone actually _wants_ to change it.

pwolfert (approved):

packages/design-system-tokens/src/themes/medicare/defaultTheme.ts
forrestbaer: Yeah in this case since all the \`--default\` variables go together it makes a little sense but isn't super understandable.. I think when I get to the a plugin for storybook possibly to theme these it won't matter as much. I was thinking last night as I drifted off to sleep about how creating a new theme would look, as it is now it's a little rough, lots of variables, lots of testing.. a simple editor of some type would improve that a lot.. On my mind :thinking:

packages/design-system/src/styles/components/_StepList.scss
forrestbaer: I'm trying to do mostly color right now with a few spacing variables tossed into the mix for simplicity sake, but yes I definitely see us expanding the variable list in the future when someone wants a change that goes outside the 'norm'

PR #1780, opened 2022-04-29
[WNMGDS-1610] Implement button analytics.

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1610)

Implement button analytics.

Analytics event should fire on click.

Links have different analytics requirements than buttons (links send URL info, buttons don't).


zarahzachz:
I'm running into two issues:

**Omnibug shows our analytics events firing twice per interaction (includes other components like `<Dialog />` and `<HelpDrawer />`).** 

I don't know if this is intentional, but did notice events fire once on `<Button />` when the underlying element is an anchor tag.

**Can't seem to get the event to fire on the close button within the `<Dialog />`**

I added `console.log`s to the handles in the `<Button />` dealing with click events and can't seem to apply the analytics event if a click event prop isn't set on components with underlying button tag.

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -109,8 +129,75 @@ export const Button = ({
...
+          if (typeof child === 'string') {
+            return child;
+          }
+        })
+        .join(' ');
+
+      buttonText = labelStrings;
+    }
+
+    console.log('=== BUTTON LABEL === ', buttonText);
```
zarahzachz, line 159: This seems to fire for all `<Button />` components

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -148,6 +235,11 @@ export const Button = ({
     attrs.onClick = handleClick;
   }
 
+  if (!onClick) {
+    console.log('=== TEST NO ONCLICK DEFINED ===');
```
zarahzachz, line 239: Never fires

pwolfert:
Knowledge transfer complete 🖨️ 🤖 . I will open a new PR once I'm ready

pwolfert:
Knowledge transfer complete 🖨️ 🤖 . I will open a new PR once I'm ready

PR #1781, opened 2022-04-29
[WNMGDS-1637] Implement native dialog component

zarahzachz:
## Summary

[JIRA ticket](https://jira.cms.gov/browse/WNMGDS-1637)

- [x] Create `<NativeDialog />`
- [x] Add polyfill support
- [x] Implement in `<Drawer />`
    - [x] Markup
    - [x] Styles
    - [x] A11y
    - [x] Behavior (focus trapping, interaction with background content, etc.)
    - [x] Tests (if applicable)

### Testing (building)

- Checkout the [docs site](https://design-system-demo.s3.amazonaws.com/WNMGDS-1637/implement-native-dialog/core/components/drawer/index.html) to make sure Drawer examples look/behavior like [PROD](https://design.cms.gov/components/drawer/) (Note: You'll need to add `index.html` to the end of each iframe URL example to see them 😢 )
- Check out [Storybook](https://design-system-demo.s3.amazonaws.com/WNMGDS-1637/implement-native-dialog/core/storybook/index.html?path=/story/components-drawer--drawer-toggle-with-drawer) for more look/behavior checks (optional)


examples/react-app/package.json
```diff
@@ -3,9 +3,9 @@
   "version": "1.0.0",
   "private": true,
   "dependencies": {
-    "@cmsgov/design-system": "^2.0.0",
-    "react": "^16.2.0",
-    "react-dom": "^16.2.0"
+    "@cmsgov/design-system": "file:../../packages/design-system/",
```
zarahzachz, line 6: Updated to point to local build of design system to test in IE using Sauce Labs

packages/design-system/package.json
```diff
@@ -27,6 +27,7 @@
     "@types/react-dom": "^17.0.10",
     "classnames": "^2.2.5",
     "core-js": "^3.6.5",
+    "dialog-polyfill": "^0.5.6",
```
zarahzachz, line 30: Needed to support IE and older versions of Safari and Firefox

zarahzachz: Actually I think Patrick has a branch up addressing a `window not defined bug` that might need merged here

packages/design-system/src/components/Drawer/Drawer.stories.jsx
```diff
@@ -3,7 +3,7 @@ import { Title, Subtitle, Description, ArgsTable, PRIMARY_STORY } from '@storybo
 import { useArgs } from '@storybook/client-api';
 
 import Drawer from './Drawer';
-import DrawerToggle from './DrawerToggle';
+import { Button } from '../Button';
```
zarahzachz, line 6: NativeDialog removes need for DrawerToggle component

packages/design-system/src/components/NativeDialog/NativeDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+  /*
+   * @hide-prop Spreading props
+   */
+  [x: string]: any;
+}
+
+const NativeDialog = ({ children, exit, showModal, ...dialogProps }: NativeDialogProps) => {
+  const dialogRef = useRef(null);
+
+  // useLayoutEffect(() => {
```
zarahzachz, line 25: Needs to be uncommented if testing for IE

packages/design-system/src/components/NativeDialog/NativeDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+  //   if ((window as any).HTMLDialogElement === undefined) {
+  //     dialogPolyfill.registerDialog(dialogRef.current);
+  //   }
+  // })
+
+  useEffect(() => {
+    const dialogNode = dialogRef.current;
+    showModal ? dialogNode.showModal() : dialogNode.show();
```
zarahzachz, line 33: `showModal` will call a method that makes an `open` attribute appear dynamically. Depending on which method gets called, you either get a focus trap or you don't.

packages/design-system/src/components/NativeDialog/NativeDialog.tsx
```diff
@@ -0,0 +1,59 @@
...
+
+  useEffect(() => {
+    const dialogNode = dialogRef.current;
+    const handleCancel = (event) => {
+      event.preventDefault();
+      exit();
+    };
+    dialogNode.addEventListener('cancel', handleCancel);
+    return () => {
+      dialogNode.removeEventListener('cancel', handleCancel);
```
zarahzachz, line 48: Chrome offers an `onCancel` event for dialog that's not fully supported. This fills in gaps for that for other browsers.

packages/design-system/src/styles/components/_Dialog.scss
```diff
@@ -2,6 +2,53 @@
 
 $dialog-spacer: $spacer-4;
 
+// Start polyfill styles
+dialog {
```
zarahzachz, line 6: I inlined the polyfill CSS as discussed in #1685 but am somewhat concerned about impacting the cascade, mainly because I can't guarantee when this CSS will load in relation to styles for Drawer.

A better spot might be to add this in a CSS higher up the cascade, like an internal DS reset. Do we have something like that?

packages/design-system/src/styles/components/_Drawer.scss
```diff
@@ -55,10 +55,23 @@ $drawer-toggle-space: 0;
...
+  max-width: 100%;
+  width: 100%;
+  max-height: 100%;
+  height: 100%;
+
+  // If `hasFocusTrap` enabled, `background` should
+  // not be visible.
+  &::backdrop {
+    background: transparent;
+  }
```
zarahzachz, line 74: Styles added to overwrite user-agent styles provided by browser

zarahzachz (commented):
~FYI: Tests are currently a work in progress (JSDOM doesn't have the latest browser API for dialog, so methods aren't understood).~ No longer an issue

packages/design-system/package.json
pwolfert: If you've looked at those changes and like them, can you merge them or cherry-pick them into this branch?

zarahzachz: Will do!

packages/design-system/src/styles/components/_Dialog.scss
pwolfert: We have control over the order because we import these files explicitly in the components index, and they're alphabetical. However, I don't think it'll matter because we use `.ds-c-dialog*` and `.ds-c-drawer*` selectors, which have a higher specificity than the `dialog` type selector. Hmm, I guess there's the `dialog.fixed` further down 🤔 . I don't actually see the `fixed` class referenced anywhere in the polyfill code. Do you know what that's about?

If we don't need to care about order, we could possibly include this in its own `_NativeDialog.scss` file.

forrestbaer (commented):
I'm having some trouble running storybook in my branch, it's actually failing because of LabelMask not being resolved in the TextField component. Are you also having this issue? 

If you are/once that's resolved you'll probably want to run yarn loki update yarn loki:healthcare update and yarn loki:medicare update to update reference shots as I imagine there will be small differences..

packages/design-system/src/components/NativeDialog/NativeDialog.tsx
```diff
@@ -0,0 +1,62 @@
+import React, { useRef, useEffect, useLayoutEffect } from 'react';
+import dialogPolyfill from './polyfill';
+
+interface NativeDialogProps {
+  children: React.ReactNode;
+  /**
+   * Function called to close dialog.
+   */
+  exit: any;
```
forrestbaer, line 9: Could maybe be a little more strict? In a few other places we use:

`(...args: any[]) => any`

packages/design-system/src/styles/components/_Dialog.scss
```diff
@@ -2,6 +2,53 @@
...
+  width: -webkit-fit-content;
+  width: fit-content;
+  height: -moz-fit-content;
+  height: -webkit-fit-content;
+  height: fit-content;
+  margin: auto;
+  border: solid;
+  padding: 1em;
+  background: white;
+  color: black;
```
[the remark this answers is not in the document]
forrestbaer: I think in the case of black and white it's not a big priority but definitely could use `$color-white` and `$color-black` as eventually all these kinds of definitions will switch over.

packages/design-system/src/components/NativeDialog/NativeDialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+   * is opened with `showModal`, focus will be trapped inside the dialog, and it will
+   * need to be closed before the rest of the page can be interacted with. See the
+   * [`showModal` MDN documentation](https://developer.mozilla.org/en-US/docs/Web/API/HTMLDialogElement/showModal)
+   * for more details.
+   */
+  showModal?: boolean;
+  /*
+   * @hide-prop Spreading props
+   */
+  [x: string]: any;
```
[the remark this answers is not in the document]
forrestbaer: `interface NativeDialogProps extends Omit<DialogHTMLAttributes<HTMLElement>, 'children'> {`

Worked for me here

PR #1782, opened 2022-04-29
[WNMGDS-1565] button theme

forrestbaer:
## Summary

🗑️ 🔥 😓 

Adds component theme definition variables for buttons. There is a lot of duplication of values under specific names where it may not be necessary to allow some flexibility for a future button style re-write. I imagine when it comes time to re-work buttons they can be simplified by half or more.

I've tested hover and active states as well using storybook.

## How to test

`yarn loki test` `yarn loki:healthcare test` `yarn loki:medicare test` and if you want to test hover/active styles you can do so in storybook for each.


packages/design-system/src/styles/components/_Button.scss
```diff
@@ -26,8 +26,9 @@
   &.ds-c-button--hover,
   &:hover &:focus,
   &.ds-c-button--hover &.ds-c-button--focus {
-    border-color: $color-primary-darker;
-    color: $color-primary-darker;
+    background-color: $button__background-color--hover;
```
pwolfert, line 29: Looking at my local doc site, the hover state for default and transparent buttons appears to be broken.

Screen Shot 2022-05-02 at 9 37 09 AM

If I remove the `ds-c-button--hover` class, the true hover state is still that way.

forrestbaer: yikes, an extra background attribute snuck in after my testing.. should be resolved, just went through all 3 child systems to be sure.

pwolfert (approved):
LGTM 👍

PR #1786, opened 2022-05-02
[WNMGDS-1642] Make sure Alerts always have an `aria-labelledby`

pwolfert:
I also moved the id for the header up one level so it would capture our a11y label as well. I discovered that it didn't include the a11y label when I was writing new unit tests.

## Summary

https://jira.cms.gov/browse/WNMGDS-1642

### Fixed

- Previously Alerts without headings had no `aria-labelledby` attribute. The `aria-labelledby` attribute now falls back to the a11y label so that we satisfy the requirement for all `role="region"` elements to have an aria label.

## How to test

<details>
  <summary>Look at the accessibility tree in devtools:</summary>
  
  Screen Shot 2022-05-02 at 11 46 51 AM
</details>

## Discussion

There's maybe an opportunity to simplify the HTML here.

```html
<div class="ds-c-alert__header ds-c-alert__heading">
  <span
    class="ds-c-alert__a11y-label ds-u-visibility--screen-reader"
    id="alert_a11y_label_6"
  >
    Notice:
  </span>
  <h2 class="ds-c-alert__heading">This is a simple heading</h2>
</div>
```

Right now we actually have two components with that `ds-c-alert__heading` class. I propose that we actually move the `a11yLabel` inside the `<h#>` heading element. However, [@zarahzachz might have some insight into why it was structured like that](https://github.com/CMSgov/design-system/commit/c549eb59a8e86f92a8c10ded80c5926573aff312#diff-37f97c547f54f69ff76e16809c0f91ced438648da53436cd7dc23b5848286bf9L221) and if it was done for a particular reason.

### DIFF
--- a/packages/design-system/src/components/Alert/Alert.test.tsx
+++ b/packages/design-system/src/components/Alert/Alert.test.tsx
@@ -101,6 +101,21 @@ describe('Alert', function () {
       renderAlert({ variation: 'warn' });
       expect(screen.getByText('Warning:')).toBeInTheDocument();
     });
+
+    it('points aria-labelledby to heading', () => {
+      const heading = 'Elvis has left the building';
+      renderAlert({ heading, variation: 'error' });
+      const alert = screen.getByRole('region');
+      const id = alert.getAttribute('aria-labelledby');
+      expect(alert.querySelector(`#${id}`).textContent).toContain(`Alert: ${heading}`);
+    });
+
+    it('falls back aria-labelledby to a11y label when no heading is provided', () => {
+      renderAlert();
+      const alert = screen.getByRole('region');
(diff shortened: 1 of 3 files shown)


zarahzachz (commented):
I [commented on that file you linked to me](https://github.com/CMSgov/design-system/commit/c549eb59a8e86f92a8c10ded80c5926573aff312#r72778892) without realizing it was part of the closed HCM PR.

forrestbaer:
Looks good to me, ChromeDriver is out of date so the a11y tests aren't running for me with Chrome v101.  Looks like it was removed from our packages.json at some point and wasn't upgrading `yarn add chromedriver -W -D` added the new version and all looked good.

zarahzachz (approved):

PR #1787, opened 2022-05-02
[WNMGDS-1648] Bundle JS for the CDN

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1648

Updates the build process to create JS bundles for the CDN. Also adds an example project, which will need to be updated once we actually have files on the CDN.

### Added

- Added JavaScript bundling to our build process, which outputs files to `dist/js/`
- Added a `cdn` example project

## How to test

1. Run `yarn build` and then copy the `dist` folder to `examples/cdn`. (In the future, the example will point directly to our files on the CDN.)
2. Open `examples/cdn/index.html` in your browser


forrestbaer:
Looks good, total bundle size of around 1.4M with fonts/css/js only :grimacing:

forrestbaer (approved):

PR #1789, opened 2022-05-03
[WNMGDS-1596] Styles > Base content migration

scheul93:
## Summary

* Migrating the Styles/Base page content
  * Updated the content for MDX
* Add an `EmbeddedExample` component that will render some elements as well as the HTML for them.
  * I did update the arrow used for the 'Code Snippet' portion to use one of our DS arrows

## How to test

`yarn start:gatsby`. Navigate to Styles > Base and click through page

Code snippets closed:
Screen Shot 2022-05-03 at 11 38 07 AM

Code snippets open:
Screen Shot 2022-05-03 at 11 38 22 AM

### DIFF
--- a/packages/docs/src/components/EmbeddedExample.tsx
+++ b/packages/docs/src/components/EmbeddedExample.tsx
@@ -1,0 +1,42 @@
+/* eslint-disable react/no-danger */
+import React, { useState } from 'react';
+import ReactDOMServer from 'react-dom/server';
+import Prism from 'prismjs';
+import { ArrowIcon } from '@cmsgov/design-system';
+
+interface EmbeddedExampleProps {
+  children: React.ReactElement;
+}
+
+/**
+ * Shows a code example as its rendered form as well as the HTML that is rendered for it
+ *
+ * Using ReactDOMServer's renderToString function to extract the HTML from a React child
+ * @see https://reactjs.org/docs/react-dom-server.html#rendertostring
+ *
+ * @todo HTML formatting is not great -- look into making it better
+ */
+const EmbeddedExample = ({ children }: EmbeddedExampleProps) => {
+  const [open] = useState(false);
+  const html = ReactDOMServer.renderToString(children);
+  const highlightedContent = Prism.highlight(html, Prism.languages.html);
+
+  return (
+    <section className="c-embedded-example">
+      <div className="ds-u-border--1 ds-u-padding--2">{children}</div>
+      <details open={open} className="c-embedded-example__details">
+        <summary className="ds-u-margin-y--1 ds-c-button ds-c-button--small ds-c-button--transparent ds-u-padding--1 ds-u-text-decoration--none">
+          <ArrowIcon direction="right" /> Code snippet
+        </summary>
+        <pre className="ds-u-margin-bottom--4 ds-u-overflow--auto ds-u-padding--2 language-html">
+          <code
+            dangerouslySetInnerHTML={{ __html: highlightedContent }}
+            className="language-html"
+          />
+        </pre>
+      </details>
+    </section>
+  );
+};
+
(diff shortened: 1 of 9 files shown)


packages/docs/src/components/EmbeddedExample.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <details open={open} className="c-embedded-example__details">
+        <summary className="ds-u-margin-y--1 ds-c-button ds-c-button--small ds-c-button--transparent ds-u-padding--1 ds-u-text-decoration--none">
+          <ArrowIcon direction="right" /> Code snippet
+        </summary>
+        <pre className="ds-u-margin-bottom--4 ds-u-overflow--auto ds-u-padding--2 language-html">
+          <code
+            dangerouslySetInnerHTML={{ __html: highlightedContent }}
+            className="language-html"
+          />
+        </pre>
```
scheul93, line 36: I know we talked about not generating the HTML from a React component but instead just having a code block in the content, but upon implementation, I didn't feel great about that. 1) Because we're generating lorem ipsum text, you'd have to copy the rendered markup from the browser 2) _if_ these examples ever change, both the example & and HTML code block would need to be updated.

So I attempted this implementation. Current biggest con is that the formatting of the code snippet is kind of rough and I haven't put a lot of effort into finding a way to fix that.

I also think this implementation may help with other, more complicated pages like [Flexbox](https://design.cms.gov/utilities/flexbox/) which has multiple, long examples. Being able to hide the code snippet helps save page space and may also help us break up these examples into smaller chunks / code snippets when we get there.

I am open to other thoughts!

pwolfert: This is actually much simpler than I expected. I really like this component! You've set it up nicely, and MDX seems to make it much easier.

pwolfert: I bet we could [use prettier](https://prettier.io/docs/en/api.html#prettierformatsource-options)

forrestbaer: I think static examples are a good idea for formatting and the ability to use inline comments.

forrestbaer:
Looks pretty good! I would prefer the code examples be expanded by default, seems like work to click another thing, especially when a lot of folks using this will want to see those right away for copy/paste.

pwolfert:
There's a good feature idea for next PI: A copy-to-clipboard button

scheul93:
> Looks pretty good! I would prefer the code examples be expanded by default, seems like work to click another thing, especially when a lot of folks using this will want to see those right away for copy/paste.

@forrestbaer I half agree. On some pages in the 'utilities' section, there are a lot of examples and all the expanded code blocks may muddy up the page content. I'm wondering if part of the re-deisgn, we could rearrange the page a bit to put the content at the top and the examples in a section towards the bottom. Then we could have the expanded snippets but also allowing people to focus on the guidelines if that's what they're looking for 🤷🏻‍♀️ 

I also like the copy-to-clipboard idea @pwolfert !

packages/docs/src/components/EmbeddedExample.tsx
scheul93: Now with prettier!
Screen Shot 2022-05-03 at 1 19 39 PM

forrestbaer:
Good point, maybe there is a configuration cookie or setting near the theme switcher to expand them by default or not in the future.

forrestbaer (approved):

packages/docs/src/components/EmbeddedExample.tsx
pwolfert: 👨‍🍳👌

pwolfert (approved):

PR #1796, opened 2022-05-06
adding formlabel component theme

forrestbaer:
## Summary

Adds form label component styling items, theme variable names based loosely on class naming convention, there are several other form theme items that will all go under the `form` category

## How to test


packages/design-system/src/styles/components/_FormLabel.scss
```diff
@@ -35,23 +35,23 @@
...
 }
 
 .ds-c-field__hint--inverse {
-  color: $color-muted-inverse;
+  color: $form-hint__color--inverse;
 }
 
 .ds-c-field__error-message {
-  color: $color-error;
+  color: $form-field__color--error;
```
pwolfert, line 48: Looking at `form-hint__color` vs `form-field__color--error`, I almost feel like the error should have the same treatment to be `form-error__color`. It's not changing the color of the form field when it's in an error state; the error message is a sub-element of the form label, just like the form hint.

forrestbaer: Good call! updated/tested

pwolfert (approved):

PR #1799, opened 2022-05-06
[WNMGDS-1577] vertical nav theme

forrestbaer:
## Summary

Adds color theme items for vertical navigation, note some transparent items were left transparent. At the moment I would prefer to keep most transparent items as transparent until a theme requires otherwise, in most cases transparent is used for a purpose where it seems it should stay that way.

## How to test


pwolfert (approved):
LGTM 🧭

PR #1801, opened 2022-05-06
[WNMGDS-1596] responsive example

scheul93:
## Summary

* Made a `ResponsiveExample` component that shows a list of breakpoint buttons as well as an iframe to a storybook story with the example
  * Much of the responsive part of the iframe was taken from the original doc site implementation
  * Note, that iframed storybook stories will not show up when running `yarn start:gatsby`. This has something to do with gatsby building in production vs local dev mode
* Made a storybook story for responsive headings used on the typography page
* Created new yarn commands
  * `yarn build-storybook:gatsby` which builds storybook in production mode and puts it into the folder `packages/docs/static` which makes the files available by relative url to gatsby pages
  * `yarn serve:gatsby` serves the production build of the gatsby site

## How to test

To see the storybook items in action, run `yarn build-storybook:gatsby` to put storybook into gatsby's static folder. This only needs to be run once since storybook stories aren't changing.

To see gatsby typography page, run `yarn build:gatsby && yarn serve:gatsby`

ResponsiveExample


zarahzachz (approved):
LGTM!

scheul93:
> This looks great! Something I was wondering about is could we utilize the viewport settings in storybook so we don't have to create our own responsive settings? https://storybook.js.org/docs/react/essentials/viewport

@line47 I looked into this! The trouble I found is that if we embed the story so that the viewport controls are visible, it doesn't scale very well on smaller screens. This is something I would love to revisit in the redesign because 1) how useful is this responsive feature in general and 2) if it is useful, do we need to support it on smaller screens

packages/docs/src/components/ResponsiveExample.tsx
```diff
@@ -0,0 +1,130 @@
+import React, { useState, useRef, useEffect } from 'react';
+import classnames from 'classnames';
+
+const breakpointOpts = {
```
forrestbaer, line 4: FYI I have these in tokens as :

```
  'width-xs': '0px',
  'width-sm': '544px',
  'width-md': '768px',
  'width-lg': '1024px',
  'width-xl': '1280px',
```

Which is inherited [from our css](https://github.com/CMSgov/design-system/blob/7e1c2bf193fcd62d461071073b69cc6d05c2d757/packages/design-system/src/styles/settings/variables/_layout.scss#L38), I wonder why `width-xs` is zero..

packages/docs/src/styles/components/responsiveExample.scss
```diff
@@ -0,0 +1,60 @@
...
+
+.c-responsive-example__iframe {
+  border: 0;
+  width: 100%;
+}
+
+.c-responsive-example__iframe-wrapper--width-xs {
+  width: 360px;
+}
+.c-responsive-example__iframe-wrapper--width-sm {
```
forrestbaer, line 18: These will eventually be changed to `$media-width-xx` I wonder if a TODO could be added or something to reference later when tokens go in.

forrestbaer: I suppose a search/replace later will bring this up, can disregard.

PR #1812, opened 2022-05-09
[HOTFIX][WNMGDS-1665] Remove max-width constraint on Choices' checked/unchecked children

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1665

#1673 introduced a new [`ds-c-wrapper` element](https://github.com/CMSgov/design-system/commit/43ed3e6dc96394db37b57072b323acd43e5803fa#diff-8e0b8a5ca7fdb96a14410bbe70bfe0d324d681a2a63382f2deadc729619e1307) that wrapped itself around a Choice's `input` and `label`, allowing it to be styled in a way that is consistent with the design system but now using the actual `input`. This was to improve the experience for screenreader users. However, there was a problem with the under-tested `checkedChildren`/`uncheckedChildren` feature in that the new `ds-c-wrapper` in that its flexbox layout was being applied to those elements, when previously they always flowed below the input + label. This could result in unexpected behavior for checked and unchecked children. Additionally, we did not call out the max-width constraint that was now being applied to the checked and unchecked children.

### Fixed

- Fixed `checkedChildren` and `uncheckedChildren` elements being siblings of the Choice input and label in the `ds-c-choice-wrapper` element, which
  1. Caused the checked and unchecked children to float to the right of the input and label when the content was narrow enough
  2. Caused the checked and unchecked children to have a max-width applied to them, which was unexpected

## How to test

1. See the new story in Storybook
2. If you want to reproduce the bug on master, add some narrow `checkedChildren` content to one of the stories' choices and see it sitting on the same line as the input and label.


zarahzachz (approved):
This looks like it fixes the problem without introducing other regressions - merge!

PR #1819, opened 2022-05-10
[WNMGDS-1626] Masked label input

zarahzachz:
## Summary

[Jira](https://jira.cms.gov/browse/WNMGDS-1626)

Implementing a LabelMask component to facilitate live masking atop TextInputs.

## How to test

Navigate to Storybook instance (_building_), update field, witness magic.


packages/design-system-docs/src/pages/components/TextField/LabelMask.example.jsx
```diff
@@ -0,0 +1,44 @@
...
+const Example = () => {
+  return (
+    <div>
+      <TextField
+        label="Phone number"
+        labelMask="phone"
+        inputMode="tel"
+        name="phone_example"
+        type="tel"
+        helpText="(123) 456-7890"
```
zarahzachz, line 14: Just noticing this prop and it'll probably break a site build, but it reminded me of a question - the ticket references a `placeholder` prop. Is that meant for the visible, masked, text that appears above the input? If so, I was thinking of using something like `helpText` instead since `placeholder` has another meaning.

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,68 @@
...
+  const textRef = useRef();
+  const modifiedTextField = React.cloneElement(field(), {
+    defaultValue: undefined,
+    onChange: (e) => handleChange(e),
+    onBlur: () => setFocus(false),
+    onFocus: () => setFocus(true),
+    type: 'text',
+    inputMode: 'numeric',
+    ref: textRef,
+    // pattern: labelMaskPattern[labelMask],
```
zarahzachz, line 55: Temporary comment - adding the wildcard string option to `labelMask` caused an error

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * @param {Object} evt
+   * @param {React.Element} field - Child TextField
+   */
+  const handleChange = (evt: React.ChangeEvent<HTMLInputElement>) => {
+    // Eval use of state here
+    setValue(evt.target.value);
+    // if (typeof field().props.onChange === 'function') {
+    //   value = field().props.onChange(evt);
+    // }
```
zarahzachz, line 43: Having a heck of a time storing the onChange value without saving it to state.

Issue I've encountered so far is not being able to see this text appear over the input as it's typed in.

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,68 @@
...
+  };
+
+  const textRef = useRef();
+  const modifiedTextField = React.cloneElement(field(), {
+    defaultValue: undefined,
+    onChange: (e) => handleChange(e),
+    onBlur: () => setFocus(false),
+    onFocus: () => setFocus(true),
```
zarahzachz, line 51: Since I brought in `useState` for the onChange stuff, I figured I'd use it for onBlur/onFocus while I was at it. This seems to work really well in showing/hiding the `helpText` or masked value without too much JS. Thoughts?

packages/design-system-docs/src/pages/components/TextField/LabelMask.example.jsx
pwolfert: I don't think we'll need a placeholder prop if we craft our masking function to print the unfilled mask when it's given an empty string. For example, if we're masking a date:
`myMaskFn("")` =`DD/MM/YYYY`
`myMaskFn("1.02")` = `01/02/YYYY`

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
pwolfert: It's up to the parent component to store that value, which means your Storybook story or your doc site example will have to do the state management. This component itself doesn't need any internal state.

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,68 @@
...
+    inputMode: 'numeric',
+    ref: textRef,
+    // pattern: labelMaskPattern[labelMask],
+  });
+
+  return (
+    <div className={`ds-c-field-mask ds-c-field-mask--${labelMask}`}>
+      {/* mask in render */}
+      {/* {formatValue()} */}
+      {focus ? <span>{value}</span> : <span>(xxx) xxx-xxxx</span>}
```
pwolfert, line 62: What I was meaning here was that we should let CSS determine if we're focused or not

```suggestion
<span className="ds-c-field-labelmask__mask--active">{labelMask(value)}</span>
<span className="ds-c-field-labelmask__mask--inactive">{labelMask('')}</span>
```

```css
.ds-c-field-labelmask__mask--active {
  display: none;
  &:focus {
    display: block;
  }
}

.ds-c-field-labelmask__mask--inactive {
  display: block;
  &:focus {
    display: none;
  }
}
```
This way we don't have to use any state to determine if this element has focus within JavaScript

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,68 @@
...
+    day_month: '[0-9.,-]*',
+  };
+
+  /**
+   * Get the child text field. Called as a method so that
+   * updates to the field cause the mask to re-render
+   * @returns {React.ReactElement} Child TextField
+   */
+  const field = (): React.ReactElement => React.Children.only(children as React.ReactElement);
+  const initialValue = field().props.value || field().props.defaultValue || '### ### ###';
```
pwolfert, line 29: This component won't support `defaultValue` unless we add that feature later. We'll always assume this is a controlled component.

pwolfert: (Which also means this whole line will be gone)

packages/design-system-docs/src/pages/components/TextField/LabelMask.example.jsx
pwolfert: `myMaskFn("1/2/2022")` = `01/02/2022`
`myMaskFn("01-02-2022")` = `01/02/2022`
`myMaskFn("0201")` = `01/02/YYYY`

packages/design-system-docs/src/pages/components/TextField/LabelMask.example.jsx
```diff
@@ -0,0 +1,44 @@
+import { TextField } from '@design-system';
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+const Example = () => {
+  return (
+    <div>
+      <TextField
```
pwolfert, line 8: We could pair down this example for now to be just for Date and do state management for a single input like
```jsx
const Example = () => {
  const [value, setValue] = useState("");
  return (
    <div>
      <TextField
        label="Enter the last day of your coverage"
        hint="If you don't have it, give your best estimate. For example: 01/02/2022"
        labelMask={LabelMask.DATE}
        value={value}

Comment (pwolfert) [reply packages/design-system-docs/src/pages/components/TextField/LabelMask.example.jsx:8]: We want a chance to showcase the simpler API anyway, where we want users to store the raw value from `onChange` without having to do masking or unmasking or anything.

File: packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx (hunk 5 above)
Comment (zarahzachz) [reply packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx:62]: This isn't working because `<span>` isn't a focusable element. 

The HTML structure is kinda like this:
```html
<div>
    <div class="label-mask>
        <span class="active">Foo</span>
        <span class="active">Bar</span>
    </div>
    <input class="text-input" />
</div>
```

I attempted using sibling selectors to target if the input has focus, but [this selector relies on the structure of the markup](https://developer.mozilla.org/en-US/docs/Web/CSS/Adjacent_sibling_combinator). Since the label comes before the input (both visually and structurally), I'm unable to target something like 

```css
.text-input:focus + .label-mask .active { display: inline-block; }
.text-input + .label-mask .active { display: none; }
```

afaik the only way to target the sibling of a focusable element, regardless of DOM order, is via the `:has` attribute, which is only available in Safari [sad trombone].

zarahzachz: Alternatively... I might be able to change the visual appearance of the DOM order using `flex` and `order`, which would allow the sibling selector to work - but it _might_ be a [WCAG violation](https://www.w3.org/TR/WCAG20-TECHS/C27.html). I say might because we are intending to hide the label mask from AT anyways, so I'm not entirely sure it matters if it's visually before or after... 🤔 

I'm going to attempt this and discuss with a11y folks.

pwolfert: Ah, that's too bad 😞

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,84 @@
...
+    if (labelMask === BuiltInMask.DATE) {
+      const charCount = str.length;
+
+      hint = 'MM/DD/YYYY';
+      const hintSub = hint.replace(/\//g, '').substring(charCount);
+
+      let val = str + hintSub;
+      // const dateRegex = /^(?<month>\d{1,2})(\s|-|\/)?(?<day>\d{0,2})(\s|-|\/)?(?<year>\d{0,4}).*/;
+      // const dateRegex = /^(?<month>\d{1,2})(?<day>\d{0,2})(?<year>\d{0,4}).*/;
+      const dateRegex = /^(?<month>[\S]{2})(?<day>[\S]{2})?(?<year>[\S]{1,4}).*/;
```
zarahzachz, line 52: I'm not crazy about this regex, but it was the only way I could allow the mixed number/letters format that happens while typing, ie, seeing "10/1D/YYY" in the hint text area. Without it, I'd get "10/1," which kinda defeats the purpose of a super intuitive mask while you type solution imo.

My regex-fu isn't very good though so I'm hopeful there's a better approach.

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  function format(str): any {
+    if (labelMask === BuiltInMask.DATE) {
+      const charCount = str.length;
+
+      hint = 'MM/DD/YYYY';
+      const hintSub = hint.replace(/\//g, '').substring(charCount);
+
+      let val = str + hintSub;
+      // const dateRegex = /^(?<month>\d{1,2})(\s|-|\/)?(?<day>\d{0,2})(\s|-|\/)?(?<year>\d{0,4}).*/;
```
zarahzachz, line 50: I initially wanted to do something with named groups, since it makes the code more readable imo, but couldn't crack that one. I know there's a way to use names with `replace`. I wanted `str.replace(regex, "$<month>/$<day>/$<year>")` to work so badly :(

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
pwolfert: You should be doing the regex on the input string before applying the mask

pwolfert: Given `str` as your raw input (`props.value`):
1. Parse `str` for month, day, and year using one or more regular expression
    - There are two main scenarios: 1) they inputted delimiters or 2) they did not input delimiters
    - In both cases, something like `/^(\d{1,2})[\D]?(\d{1,2})?[-\D]?(\d{1,4})?/` should match, but in the second case we will not be able to detect if they omitted the leading zeros for the first two groups. That's just not something we can do 🤷 
    - Here's a sandbox using the regular expression above and a bunch of example input that I made up: https://regex101.com/r/9vRHLO/1
2. Once we have values month, day, and year (with the possibility that some are empty), we can format them into a string that matches our mask with something like `const date = `${month}/${day}/${year}`
3. We then mix our mask and our formatted input like you were doing before: `hint.replace(/\//g, '').substring(date.length);`

pwolfert: This should get you your possibly-undefined variables for each of the groups
```js
const [month, day, year] = /^(\d{1,2})[\D]?(\d{1,2})?[-\D]?(\d{1,4})?/
      .exec(str)
      .slice(1)
      .filter(s => s)
```
and if you want them to default to empty strings, you can do
```js
const [month = '', day = '', year = ''] = ...
```

pwolfert: I always have to look it up, but [there's a handy table for the returned array for RegExp's exec function here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/exec#description). The first element is the matched string, and we can ignore it (`.slice(1)`), and then we filter on only strings that exist (`.filter(s => s)`)...which I guess is not really necessary since we're destructuring the array...yeah, I guess you can remove that part; that was a straight copy from my phone number example in the ticket where I was iterating through the matched groups later.

pwolfert: So more like
```
const [month = '', day = '', year = ''] = /^(\d{1,2})[\D]?(\d{1,2})?[-\D]?(\d{1,4})?/
      .exec(str)
      .slice(1);
```

pwolfert: or
```js
const RE_DATE = /^(\d{1,2})[\D]?(\d{1,2})?[-\D]?(\d{1,4})?/;
// ...
const [month = '', day = '', year = ''] = RE_DATE.exec(str).slice(1);
```

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,93 @@
...
+  children: React.ReactNode;
+  labelMask?: (rawInput: string, valueOnly?: boolean) => string;
+}
+
+const LabelMask = ({ children, labelMask }: LabelMaskProps) => {
+  if (typeof labelMask === 'string') {
+    if (labelMask === BuiltInMask.DATE) {
+      labelMask = maskDate;
+    }
+  }
```
pwolfert, line 45: This can be removed now

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,93 @@
...
+export const BuiltInMask = {
+  DATE: maskDate,
+};
+
+export interface LabelMaskProps {
+  /**
+   * Must contain a `TextField` component
+   */
+  children: React.ReactNode;
+  labelMask?: (rawInput: string, valueOnly?: boolean) => string;
```
pwolfert, line 37: We'll want to add a doc comment here to explain how the function is used and what the two props are about

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
pwolfert: Bump

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,93 @@
...
+    }
+  }
+
+  /**
+   * Get the child text field. Called as a method so that
+   * updates to the field cause the mask to re-render
+   * @returns {React.ReactElement} Child TextField
+   */
+  const field = (): React.ReactElement => React.Children.only(children as React.ReactElement);
+  const value = field().props.value;
```
pwolfert, line 53: Rather than carrying over this pattern from mask, I think we could simplify to not use a function and to pull out the props that we need, like
```suggestion
  const field = React.Children.only(children as React.ReactElement);
  const { onBlur, onChange, value } = field.props;
```

packages/design-system/src/styles/components/_LabelMask.scss
```diff
@@ -0,0 +1,30 @@
...
+
+// Changing DOM order of label mask and input
+// Maintaining expected visual order
+.ds-c-label-mask {
+  display: flex;
+  flex-direction: column-reverse;
+}
+
+.ds-c-field + .ds-c-label-mask__mask {
+  font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
```
scheul93, line 11: [question] Is this a font that is used elsewhere in the design system?

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,93 @@
+import React from 'react';
+
+function maskDate(rawInput = '', valueOnly = false): any {
+  const RE_DATE = /^(\d{1,2})[\D]?(\d{1,2})?[\D]?(\d{1,4})?/;
```
scheul93, line 4: nice

scheul93 (commented):
This looks slick when using it in Storybook!

packages/design-system/src/styles/components/_LabelMask.scss
zarahzachz: Oh you're right, looks like there's a `$font-monospace` being downloaded/used for our docs site (using webfonts) that could be used here. That's a bummer - the OS fonts work just as good here without the additional file download 😞 

Updating now.

zarahzachz: OH! Seems like this font rule is specific to our docs site and can't be used in our component library! Disregard my previous comment

packages/design-system-docs/src/pages/components/TextField/LabelMask.example.jsx
```diff
@@ -0,0 +1,21 @@
+import React, { useState } from 'react';
+import { TextField, BuiltInMask } from '@design-system';
+import ReactDOM from 'react-dom';
+
+const Example = () => {
+  const [date, setDate] = useState('');
+  return (
+    <>
+      <TextField
+        name="labelMask-date"
```
pwolfert, line 10: Nit pick: `labelMask-date` mixes two styles

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,102 @@
...
+  }
+  if (valueOnly) {
+    return formattedDate;
+  }
+  const hint = 'MM/DD/YYYY';
+  const hintSub = hint.substring(formattedDate.length);
+  return formattedDate + hintSub;
+}
+
+export const BuiltInMask = {
```
pwolfert, line 37: I'm thinking that this could use a better name. This was kind of a placeholder name until we came up with something better. `Mask` is already taken. Ideas:
- `MaskFunction.DATE`
- `LabelMask.Date` (if we applied them to the `LabelMask` component object itself...which devs don't actually use directly)
- `TextField.DATE_MASK` (if we applied them to the `TextField` component object itself)
- `TextField.MASK_DATE`
- `DATE_MASK` directly without an enum/object

@scheul93 and @forrestbaer, would like your input on it too

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,102 @@
...
+};
+
+export interface LabelMaskProps {
+  /**
+   * Must contain a `TextField` component
+   */
+  children: React.ReactNode;
+  /**
+   * Applies date format masking to the input value entered
+   * and renders to a text field above the input.
```
pwolfert, line 48: This is nice and succinct. Maybe instead of "to a text field above the input" it would be more accurate to say "to the text field's label"? And maybe we should add a note along the lines of "See documentation page for 'Label mask'"

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,102 @@
...
+   * updates to the field cause the mask to re-render
+   * @returns {React.ReactElement} Child TextField
+   */
+  const field = React.Children.only(children as React.ReactElement);
+  const { onBlur, onChange, value } = field.props;
+
+  /**
+   * @param {Object} evt
+   * @param {React.Element} field - Child TextField
+   */
```
pwolfert, line 67: This doc comment might have been helpful before our TypeScript migration, but it isn't useful anymore.

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,102 @@
...
+   * formatted value entered.
+   */
+  labelMask?: (rawInput: string, valueOnly?: boolean) => string;
+}
+
+const LabelMask = ({ children, labelMask }: LabelMaskProps) => {
+  /**
+   * Get the child text field. Called as a method so that
+   * updates to the field cause the mask to re-render
+   * @returns {React.ReactElement} Child TextField
```
pwolfert, line 59: Because this isn't a function anymore, it doesn't need to be formatted like a JSDOC and include the `@returns` line. It's still a useful comment overall though.

packages/design-system/src/components/TextField/LabelMask/LabelMask.test.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+      const mask = container.querySelector('.ds-c-label-mask__mask--active');
+      const input = container.querySelector('input');
+
+      fireEvent.change(input, { target: { value: data } });
+
+      expect(input).toHaveValue(data);
+      expect(mask.textContent).toBe(maskText);
+
+      expect(asFragment()).toMatchSnapshot();
```
pwolfert, line 68: Is the snapshot needed since you're already testing for the expected `mask.textContent`? There are a lot of snapshots to wade through in the snapshot file, and I don't know if all of them are helpful. Can it be pared down to one or two?

pwolfert (commented):
The component code looks great! The only remaining questions and suggestions I have are about naming and documentation

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
zarahzachz: I was trying to avoid saying "label" since that has semantic meaning in HTML and this isn't updating in a label tag. Is that something we should reconsider?

I like adding a comment to check out our docs page.

packages/design-system/src/styles/components/_LabelMask.scss
zarahzachz: Added a comment to the Jira ticket pointing this out.

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
pwolfert: You're right. I forgot it wasn't actually in the label. It's not a "text field" either, though. While that doesn't have semantic meaning in HTML, it does have meaning in our design system. Maybe just "displays it above the input"?

zarahzachz: Good catch! Updating that now!

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
```diff
@@ -0,0 +1,97 @@
...
+
+/**
+ * Takes the string value from an input and returns a string
+ * with appropriate date format masking applied concatenated
+ * with the date hint text, MM/DD/YYYY.
+ *
+ * `valueOnly` defaults to false. If true, returns formatted
+ * string without additional hint text.
+ */
+function maskDate(rawInput = '', valueOnly = false): any {
```
forrestbaer, line 11: Maybe add a return type of `String` ?

forrestbaer: Or even a more complex type we could expand on in the future for other mask types:

```
export type Mask<T> = <T>(rawInput: String, valueOnly: Boolean): T => T
```

Dunno maybe just optimizing for a case that doesn't exist yet.. String would also be fine just some thoughts

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
pwolfert: What makes sense to me about `DATE_MASK` by itself is that unless we applied them as properties to `TextField`, developers are still going to have to import some other named thing from our package. It might as well be the constant that they want to use.

packages/design-system/src/components/TextField/LabelMask/LabelMask.tsx
pwolfert: You're right, this should be a string and always a string

forrestbaer (approved):
👍🏻  looks good

PR #1836, opened 2022-05-12
[ no ticket ] token file cleanup / reorganization

forrestbaer:
## Summary

Basically a formatting update but includes what I believe is a better system for organizing / nesting components and colors in particular. 

- Removes prettier for token files so we can place them in columns for readability
- Splits component themes out into their own file
- Organizes component theme variables by component
- Organizes color tokens by primary color
- Several removals and items renamed for clarity
- Token export functions now export a single file with all tokens instead of multiple files
- JSON exporter exports all themes and component definitions

## How to test

`yarn token:dist`
`yarn build && yarn build:healthcare && yarn build:medicare`


.prettierignore
```diff
@@ -4,3 +4,4 @@ tmp/**
 **/types/**
 storybook-static/**
 packages/coverage/**
+packages/design-system-tokens/src/**
```
pwolfert, line 7: Could this just be the `themes` and `tokens` folders? It would be nice to keep the code files in `lib` formatted in a standard way.

packages/design-system-tokens/src/themes/core-components.ts
```diff
@@ -0,0 +1,244 @@
+/*
+ * Core CMSDS Component Theme
+ */
+
+import { default as t } from './core'
+import { hexOpacity } from '../lib/utility'
+
+export const components = {
```
pwolfert, line 8: Should this have a type annotation?

packages/design-system-tokens/src/themes/core.ts
```diff
@@ -0,0 +1,161 @@
...
+  global,
+  measure,
+  media,
+  radius,
+  shadow,
+  spacer,
+  z,
+};
+
+export default core_theme;
```
pwolfert, line 161: Why the `snake_case` here?

packages/design-system-tokens/src/themes/healthcare-components.ts
```diff
@@ -0,0 +1,254 @@
...
+ */
+
+import { default as t } from './healthcare'
+import { hexOpacity } from '../lib/utility'
+
+export const components = {
+  'alert': {
+    '__background-color':                         t.color['primary-alt-lightest'],
+    '__background-color--error':                  t.color['error-lightest'],
+    '__background-color--lightweight':            t.color['white'],
```
pwolfert, line 12: This formatting and grouping is much easier to read, and I also like the succinct `t.*` shorthand.

packages/design-system-tokens/src/themes/healthcare-components.ts
```diff
@@ -0,0 +1,254 @@
+/*
+ * Healthcare.gov CMSDS Component Theme
+ */
+
+import { default as t } from './healthcare'
```
pwolfert, line 5: At first I was unsure about having this separate file that is directly handled by the exporter instead of importing this components file into the base theme file and spreading it on the exported theme object. But now I realize how much nicer this is because the components file can _consume_ the base tokens for the theme. It's not like any other tokens will need access to the tokens defined in this components file, so this flow of information makes sense.

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -118,7 +118,7 @@
   &.ds-c-button--disabled {
     background-color: $button-transparent__background-color--disabled;
     border-color: transparent;
-    color: $button-transparend__color--disabled;
+    color: $button-transparent__color--disabled;
```
pwolfert, line 121: Nice catch 👍

pwolfert (commented):
This is so much easier to read!

forrestbaer:
@pwolfert thank you for the comments, all good points, I've made updates as recommended.

pwolfert (approved):
LGTM 👍

PR #1839, opened 2022-05-13
[NO-TICKET] conditional font path for storybook styles

scheul93:
## Summary

In the Storybook webpack config, I created a conditional to set a different font path for design system styles. This is necessary because directory structures between dev & production builds are different (prod is flatter). This also resolves an issue with the new doc site where font files weren't loading.

## How to test

`yarn build-storybook:gatsby` to build storybook with conditional font path
`yarn build:gatsby && yarn serve:gatsby` to build doc site in production mode and serve it

Notice that before this change, you can see console errors about missing font files from storybook
Screen Shot 2022-05-13 at 11 53 02 AM

After change, no console errors and font files can be seen in network tab
Screen Shot 2022-05-13 at 11 47 34 AM


scheul93:
@pwolfert, not sure if this resolves previous storybook build oddities you've encountered in the past

.storybook/webpack.config.js
```diff
@@ -56,7 +67,7 @@ module.exports = async (webpackConfig) => {
         {
           loader: 'sass-loader',
           options: {
-            additionalData: `@import '${getStylesPath()}';`,
+            additionalData: `${getFontPath()} @import '${getStylesPath()}';`,
```
pwolfert, line 70: Do we need to do the same thing for `./images` and `$image-path`?

pwolfert (approved):
🔡 🎉

pwolfert:
> not sure if this resolves previous storybook build oddities you've encountered in the past

I think this fixes it! That was so simple!

.storybook/webpack.config.js
scheul93: Added, because it IS needed 🙂

PR #1840, opened 2022-05-16
[WNMGDS-1592] component page support

scheul93:
## Summary

Added support in doc site page headers to have things like
* Badge for page status (mostly used for components in draft phase)
* "Jump to guidance" link
* A link to related uswds guidance
These things are all configured per content page in the frontmatter of the page

## How to test

`yarn start:gatsby`. Navigate to Component > Accordion to see result

Screen Shot 2022-05-16 at 10 31 28 AM


packages/docs/content/4_utilities/text-transform.mdx
```diff
@@ -1,5 +1,6 @@
 ---
 title: Text transform
+showJumpToGuidance: true
```
pwolfert, line 3: Is there a way to detect if there's a guidance section when rendering the header? That's probably pretty complicated. Just an idea

pwolfert (approved):

packages/docs/content/4_utilities/text-transform.mdx
scheul93: I don't think so unless we start querying a `table of contents` property from the graphQL query and then traversing that to check

pwolfert: Hmm, table of contents, you say? Won't we be wanting that for the in-page navigation in the future?

scheul93: And right now we don't have a purpose for the table of content data... However! In Scott's re-design, there is one, so maybe we could throw in some logic then?

pwolfert: 🥤

pwolfert: Awesome!!

PR #1848, opened 2022-05-18
[WNMGDS-1664] Deprecate nav menu Mgov component

line47:
## Summary

This is no longer in use due to the Consistent header provided by the max team. We can mark this for deprecation on the site.

### Deprecated

Nav menu

## How to test

Run the medicare doc site locally with `yarn start:medicare` and see that the navigation menu page has a deprecation notice

### DIFF
--- a/packages/ds-medicare-gov/docs/src/pages/components/NavigationMenu/NavigationMenu.docs.scss
+++ b/packages/ds-medicare-gov/docs/src/pages/components/NavigationMenu/NavigationMenu.docs.scss
@@ -1,6 +1,16 @@
 /*
 Navigation Menu
 
+<div class="ds-c-alert ds-c-alert--error">
+    <div class="ds-c-alert__body">
+      <h2 class="ds-c-alert__heading">Deprecation Notice</h2>
+      <p class="ds-c-alert__text">
+        This component has been marked as 'deprecated' and will be removed in a future release. Please use the component `Consistent Header` instead.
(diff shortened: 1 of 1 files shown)


packages/ds-medicare-gov/docs/src/pages/components/NavigationMenu/NavigationMenu.docs.scss
```diff
@@ -1,6 +1,16 @@
 /*
```
forrestbaer, line 1: Can delete these lines too?

line47: I thought I needed these comments to format the page?  - `/*`

forrestbaer: Oh, I was thinking this whole page was being removed, is that not the case? If the whole page is being removed wouldn't this whole file go?

line47: No not removing the page - adding the deprecation notice only

forrestbaer: Omg I'm sorry I'm color blind this morning, for some reason I saw the green as red.. :facepalm:

forrestbaer (approved):

PR #1849, opened 2022-05-19
[NO-TICKET] Add script for finding React versions used by dependents

pwolfert:
## Summary

I wrote this a while ago, and it has been sitting on my computer. We're starting to evaluation whether we need to do an upgrade to React 17 soon, so this is valuable information. I reorganized a bit, but I'm actively avoiding doing too much code deduplication because it's simpler to keep the logic for these separate, and it allows us to continue to modify them independently in the future if we need to.

### Added

- Added a script for finding the versions of react used by our dependents on github.cms.gov

### Changed

- Changed the `yarn scan-versions` command to `yarn scan-dependents` so it can hold more subcommands

## How to test

- `yarn scan-dependents`
- `yarn scan-dependents --react`

Note that this still doesn't scan public github.com. That could be an improvement for the future


scripts/scanDependents/index.ts
```diff
@@ -0,0 +1,53 @@
...
+  })
+  .help().argv;
+
+dotenv.config();
+
+const accessToken = argv.token ?? process.env.GHE_ACCESS_TOKEN;
+if (!accessToken) {
+  console.error(
+    chalk.red(
+      'A GitHub Enterprise personal access token is required. Please provide this as an argument or assign it to GHE_ACCESS_TOKEN in a .env file.'
```
forrestbaer, line 33:
```suggestion
      'A GitHub Enterprise personal access token is required. Please provide this as an argument or set a GHE_ACCESS_TOKEN environment variable using an .env file or by exporting that variable within your current shell.'
```

forrestbaer (approved):
Works & Looks great!

PR #1855, opened 2022-05-20
[WNMGDS-1597] component content pt 2

scheul93:
## Summary

Content migration for following components:
* Button
* Checkbox / Radio
* Date Field
* Drawer
* Dropdown
* Filter chip
* Form label

Also had to update the PropTable component because it wasn't showing simple types like `string` or `boolean`. But now it is!

## How to test

### DIFF
--- a/packages/docs/content/5_components/alert.mdx
+++ b/packages/docs/content/5_components/alert.mdx
@@ -110,7 +110,7 @@ The following Sass variables can be overridden to theme alerts:
 Import and set the `setAlertSendsAnalytics` feature flag to `true` in your application's entry file:
 
 ```js
-import { setAlertSendsAnalytics } from "@cmsgov/<design-system-package>";
+import { setAlertSendsAnalytics } from '@cmsgov/<design-system-package>';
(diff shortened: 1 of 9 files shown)


forrestbaer:
All looks good to me except the Drawer page, the 'open in storybook' link prevents the drawer from reaching the bottom and it feels like the component is broken in some way. I'm not sure what the best solution is. Any thoughts @line47 ?

line47:
@forrestbaer the easiest thing I can think of is to maybe shorten the body content in the help drawer example?

forrestbaer:
> @forrestbaer the easiest thing I can think of is to maybe shorten the body content in the help drawer example?

Hmm, that would still leave the big white gap under still.. If we changed the background color of the wrapping div like:



It wouldn't look as bad in my opinion, this could be done globally with all examples or possibly with this one example @scheul93  @line47?  Or could just expand the open in storybook link to be the whole width.

line47:
That's an interesting point I am wanting to move the storybook link down next to the code link something like this in the updated design

PR #1864, opened 2022-05-26
[WNMGDS-1627] Upgrade to Jest 27 so we can use userEvent.type correctly

pwolfert:
## Summary

### Changed

- Upgrade to Jest 27 so we can use `userEvent.type` correctly in some new unit tests

## How to test

`yarn install && yarn test`


packages/design-system/src/components/analytics/events.test.ts
```diff
@@ -67,9 +71,12 @@ describe('sendLinkEvent', () => {
       (window as any as UtagContainer).utag = { link: undefined };
       expect(sendLinkEvent(gaEventProps)).toBe(undefined);
 
-      jest.runAllTimers();
-      jest.runAllTimers();
-      jest.runAllTimers();
+      // jest.runAllTimers();
+      // jest.runAllTimers();
+      // jest.runAllTimers();
```
scheul93, line 76: are these still used? If not, can they be removed?

forrestbaer (approved):
👍🏻

PR #1867, opened 2022-05-26
[WNMGDS-1627] Refactor FormControl and LabelMask into hooks

pwolfert:
## Summary

### Added

- Added `useFormLabel` hook as an alternative to `FormControl`, which allows the implementing component to decide how to structure the HTML and where to place the label and field elements relative to each other.
    - It's also arguably more straightforward to understand than the render function that `FormControl` uses. At the time when `FormControl` was created, we did not have the ability to use React Hooks in the design system
    - `FormControl` now uses `useFormLabel` under the hood, and we should consider deprecating `FormControl`
    - Also added a new `FormFieldProps` type to be used in building the public-facing component props without all the duplication. It defines all the props that should exist on a component that is utilizing the `useFormLabel` hook.
- Added `useLabelMask` hook, which captures all the core logic of the label mask but allows the implementing component to decide how to structure the HTML and where to place the label-mask element and the field that drives it relative to each other. It will be used in the new `SingleInputDateField` component.

### Changed

- Changed the `TextField` implementation to move the `Mask` and `LabelMask` logic up to the `TextField` and out of the `TextInput` component, because there was some circular logic there that was really hard to follow and reason about
- Updated `MonthPicker` to use `useFormLabel` directly as an example. Only functional components can use hooks.

## How to test

In Storybook, try out the following stories:
- http://localhost:6006/?path=/story/components-month-picker--default-month-picker
- http://localhost:6006/?path=/story/components-text-field--single-line-field
- http://localhost:6006/?path=/story/components-text-field--label-masked-field
- And any other stories that seem appropriate


pwolfert:
We're going to have a meeting to walk through this tomorrow

forrestbaer:
Getting some test test failures, `TypeError: Cannot read property 'textContent' of null`

pwolfert:
Added some tests for the mask function and the `useLabelMask` hook

packages/design-system/src/components/FormLabel/useFormLabel.tsx
```diff
@@ -0,0 +1,138 @@
...
+      <>
+        {label}
+        {showHiddenError && <div className="ds-u-visibility--screen-reader">{errorMessage}</div>}
+      </>
+    ),
+    className: labelClassName,
+    component: labelComponent,
+    errorMessage: bottomError ? undefined : errorMessage,
+    errorMessageClassName: bottomError ? undefined : errorMessageClassName,
+    errorId: errorId,
```
zarahzachz, line 107: Could this just be `errorId`?

packages/design-system/src/components/FormLabel/useFormLabel.tsx
```diff
@@ -0,0 +1,138 @@
...
+    // Avoid using `for` attribute for components with multiple inputs
+    // i.e. ChoiceList, DateField, and other components that use `fieldset`
+    fieldId: wrapperIsFieldset ? undefined : id,
+    hint: hint,
```
zarahzachz, line 111: Could this just be `hint`?

packages/design-system/src/components/TextField/useLabelMask.test.tsx
```diff
@@ -0,0 +1,79 @@
+import React from 'react';
+import useLabelMask, { DATE_MASK } from './useLabelMask';
+import { render, screen, cleanup } from '@testing-library/react';
+import { renderHook, act } from '@testing-library/react-hooks';
```
zarahzachz, line 4: Nice! Didn't realize there was testing available for hooks

zarahzachz:
I might've found a bug - looks like the `id` (and all attrs using that value) update onChange. This only seems to happen with the Label Masked Field.
https://user-images.githubusercontent.com/5159392/170770555-82b1e835-6b09-4b7f-a1b1-acc7f24a61e9.mov

pwolfert:
Good catch, @zarahzachz! Fixed it

pwolfert:
@forrestbaer or @scheul93, since Sarah is out, could one of you continue reviewing this PR?

PR #1869, opened 2022-05-27
[WNMGDS-1573] Tabs component theme

forrestbaer:
## Summary

Adds `Tabs` component theme

## How to test

`yarn build && yarn loki test && yarn loki:healthcare test && yarn loki:medicare test`


packages/ds-medicare-gov/src/styles/settings/variables/_medicare-theme.scss
```diff
@@ -53,7 +53,7 @@ $color-error-darkest: #720e1f;
 $color-error-light: #e8b3b3;
 $color-error-lighter: #f7bbc5;
 $color-error-lightest: #f7e6e6;
-$color-focus: #3e94cf;
+$color-focus: #c97532;
```
forrestbaer, line 56: comes from master

packages/design-system/src/styles/components/_Tabs.scss
```diff
@@ -26,16 +26,16 @@
...
-  border-top: 1px solid $border-color;
-  color: $color-base;
+  border-top: 1px solid $tabs__border-color;
+  color: $tabs__color;
   display: inline-block;
   font-size: $font-size-sm;
-  font-weight: $font-bold;
-  line-height: $reset-line-height;
+  font-weight: $font-weight-bold;
+  line-height: $font-line-height-reset;
```
zarahzachz, line 38: This will probably be revisited with later work, but should font styles like these be defined under the "tabs" theme?

zarahzachz (approved):
LGTM

packages/design-system/src/styles/components/_Tabs.scss
forrestbaer: I'm focusing on colors primarily right now as a pilot, eventually will add more if we decide to make these more customizeable here.

forrestbaer: So these are just changes to reference the new token name values.

PR #1870, opened 2022-05-27
[WNMGDS-1576] USABanner component theme

forrestbaer:
## Summary

- Adds `USABanner` component theme
- Colors were off-brand and hard-coded previously, updated to use brand colors
- Updated loki references

## How to test

Can look at screenshots to see updates to references, the references aren't expanded in storybook automatically, can `yarn install && yarn build && yarn run` to compare to https://design.cms.gov/


packages/design-system-tokens/src/themes/healthcare-components.ts
```diff
@@ -241,6 +241,16 @@ export const components: AllTokenValues = {
     '__border-radius':                            '0',
   },
 
+  'usaBanner': {
```
pwolfert, line 244: Have we used `camelCase` on these names before? Looks like the one below uses `kebab-case`

forrestbaer: Need to switch that over to camel case as well, for parsing these later it's easier to split on `-` so camel case keeps things better organized.

forrestbaer: For now neither is needed but I will clean it up in a big once-over in the main branch before this feature branch gets merged.

pwolfert (approved):
All the screenshot changes LGTM

PR #1871, opened 2022-05-27
NO-TICKET/ Update visual regression docs

zarahzachz:
## Summary

Updated vis regression docs to remove references to Docker (no longer needed) and troubleshooting info. Also updated our script to remove Docker language.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,14 +84,19 @@ These scripts can all be run from the root level of the repo:
 
 ### Visual regression testing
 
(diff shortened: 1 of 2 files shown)


forrestbaer:
Docker is actually required for loki to run locally, in the CI env it's available via the CI docker instance that runs with the jenkins pipeline. I think this just needs to be the part about upgrading bash if you get that error because the default bash installed with OSX is out of date.

forrestbaer (approved):
👍🏻

PR #1872, opened 2022-05-27
[WNMGDS-1598] patterns content

scheul93:
## Summary

* Content migration for:
  * Documentation Page pattern
  * Review pattern
  * Step List pattern
  * Third party, external link pattern

## How to test

### DIFF
--- a/packages/docs/content/6_patterns/step-list.mdx
+++ b/packages/docs/content/6_patterns/step-list.mdx
@@ -1,0 +1,126 @@
+---
+title: Step List
+---
+
+import { Badge } from '@cmsgov/design-system';
+
+A step list represents a user's progression through an application or multi-page
+form. It serves as a table of contents and a way to quickly see where they are
+and what they should be working on next.
+
+## Start, Resume, and Edit links
+
+A user interacts with the steps through "Start", "Resume", and "Edit"
+links.
+
+In the React component, the links' `href` properties are determined by the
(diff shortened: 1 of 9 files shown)


packages/docs/content/6_patterns/step-list.mdx
```diff
@@ -0,0 +1,126 @@
...
+  storyId="patterns-step-list--step-list-example"
+/>
+
+### Props
+
+<PropTable componentName="StepList" />
+
+## `<Step>`
+
+### Step Object
```
scheul93, line 42: This is a weird situation because the `StepObject` is an interface that we export that is used to configure StepList data. However, it is not the same thing as the props for the `Step` component, so we cannot use react-docgen to grab this data programmatically.

In the current doc site, this data is hard coded into the content, so I did the same here until an alternative solution can be found

packages/docs/content/6_patterns/third-party-external-link.mdx
```diff
@@ -0,0 +1,30 @@
+---
+title: Third party external link
+---
+
+The Third party external link pattern is used to notify a user that the action they are taking will take them to a new external website.
```
zarahzachz, line 5: I know you're just copying, but the capitalization of "Third" is weird.

zarahzachz (approved):
LGTM

PR #1875, opened 2022-05-31
[WNMGDS-1675] component content pt 3

scheul93:
## Summary

* Content migration for:
  * Icon
  * Idle Timeout
  * Masked field
  * Modal dialog
  * Month Picker
  * Pagination
  * Skip Nav

## How to test


forrestbaer:
These all look good! :+1: 

Two questions relating to the stories two of these represent that maybe could be covered here? out of scope?:

*   Skip Nav: seems like we should focus the hidden element in the story? tabbing to it can take a long time
*   Idle Timeout: default is 30 seconds, maybe it should be like, 5?

scheul93:
1. I'm not sure if we should autofocus the skip nav in the story on page load. It does take a long time to tab to but that is the same in the current doc site. I think I'm mostly concerned with potential accessibility issues of forcing focus in the middle of page content
2. I tried making the time to show warning shorter and found an interesting bug (or feature?). The IdleTimeout checks the "status" of the page every 30s, meaning it checks if there has been any user interactions. This is hard coded. Even if I change the time to show warning prop to be < 30s, the modal won't show until at least 30s because of [this](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx#L106) & [this](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/components/IdleTimeout/IdleTimeout.tsx#L200)

PR #1876, opened 2022-05-31
[WNMGDS-1627] Single-input date field

pwolfert:
## Summary

This is still a work in progress, but I'd like to get some feedback on it. What's left to do is finalize the API around when we use the picker button and when we don't and whether they should be two different components to our users.

### Added

- Added a `SingleInputDateField` component
- Added `useClickOutsideHandler` and `usePressEscapeHandler` hooks

### Changed

- Renamed `DateField` to `MultiInputDateField` with an alias for the old name
- Updated the `useFormLabel` and `useLabelMask` hooks a little bit to make them easier to work with and more consistent

### Deprecated

### Removed

List deprecated features or components removed in this PR.

### Fixed

List any bug fixes.

## How to test

- http://localhost:6006/?path=/story/components-singleinputdatefield--with-picker
- http://localhost:6006/?path=/story/components-singleinputdatefield--default


packages/design-system/src/components/utilities/useClickOutsideHandler.ts
```diff
@@ -0,0 +1,25 @@
+import { RefObject, useEffect } from 'react';
+
+type ClickOutsideEvent = MouseEvent | TouchEvent;
+
```
forrestbaer, line 4: Nice hook! Maybe could use a little documentation

forrestbaer (commented):
Great work! Tests look good, feels like we need a form/input usage visual guide but maybe that's just me 😁 

I like this picker, really easy to use with the keyboard, seems accessible by the little testing I did using voiceover and it looks great.  It is a bit of a hassle if you need to pick a date in 1970 or so but easily resolved by just typing the date.

packages/design-system/src/components/DateField/SingleInputDateField.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+  const dayPickerRef = useRef();
+  const calendarButtonRef = useRef();
+  useClickOutsideHandler([dayPickerRef, calendarButtonRef], () => setPickerVisible(false));
+  usePressEscapeHandler(dayPickerRef, () => setPickerVisible(false));
+
+  // Validate the date string (value) and make date null if it's invalid. We don't want to pass
+  // a bizarre date to DayPicker like `new Date('01/02')`, which is interpreted as `Jan 02, 2001`
+  const dateString = DATE_MASK(props.value, true);
+  const validDateString = isMatch(dateString, 'MM/dd/yyyy');
```
scheul93, line 81: [question] I think I'm behind on some of the requirements, but is the date format always going to be `MM/dd/yyyy`?

packages/design-system/src/components/DateField/SingleInputDateField.stories.jsx
```diff
@@ -0,0 +1,38 @@
...
+
+const Template = ({ ...args }) => {
+  const [dateString, setDateString] = useState('');
+  return <SingleInputDateField {...args} value={dateString} onChange={setDateString} />;
+};
+
+export const Default = Template.bind({});
+
+export const WithPicker = Template.bind({});
+WithPicker.args = {
```
scheul93, line 36: Is the styling for this component complete? Because in Storybook I'm not seeing an outline around the calendar popup like in the mocks
Storybook:
Screen Shot 2022-06-01 at 10 06 35 AM

Mocks:
Screen Shot 2022-06-01 at 10 07 04 AM

packages/design-system/src/components/DateField/SingleInputDateField.stories.jsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+const Template = ({ ...args }) => {
+  const [dateString, setDateString] = useState('');
+  return <SingleInputDateField {...args} value={dateString} onChange={setDateString} />;
+};
+
+export const Default = Template.bind({});
+
+export const WithPicker = Template.bind({});
```
scheul93, line 35: Should this story contain some kind of `onChange` to better simulate what happens when a date is selected from the picker?

scheul93: Nevermind! I think my branch was behind because I see something happening when a date is selected in the picker

scheul93 (approved):
Looks good! Really like the new hooks

pwolfert:
@SparrowSpit, could you try out this new component in the storybook environment linked to in the description and let us know what you think? I'm going to go ahead and merge this with the assumption that we'll iterate on it before releasing it, but I'm tagging you here for now. One thing I know we'll need that I flagged in the description is some more descriptive content for the calendar button. I was thinking something like "Open calendar picker dialog" / "Close calendar picker dialog". Thoughts?

SparrowSpit:
@pwolfert Mostly it's working great! A few things to consider:

1. It would be helpful for a screen reader user if the description of the date format is associated with the text input via` aria-describedby`, making it available to assistive technologies as an accessible description. Currently the hint/format text is set to `aria-hidden`. What’s the reason for this?
2. To the calendar grid dialog I’d suggest adding more context for screen reader. Add `role=dialog` and (as you mentioned in your comment) an `aria-label` on the calendar button that provides an accessible name/text like “Open calendar”
3. After I close the calendar dialog (using ESC key or by selecting a date), the focus should be returned to the calendar icon button
4. Not sure how helpful the update while you type of the format/hint text is. Might be more confusing than just leaving it static to understand the correct formatting. I'd suggest leaving the format/hint text as MM/DD/YYY if possible. 

(Note: My PC isn't cooperating right now so I'll have to test with JAWS/NVDA when I get it working.)

pwolfert:
@SparrowSpit, thanks for reviewing!

1. We made a decision to disassociate this information because the live changes as the user types would be distracting. There is a conversation we had in our accessibility channel that I can share with you. Essentially we wouldn't gain any of the a11y advantages of this model over a classic input mask if we announced all those changes. The guidance should explain that the hint text provided should give an example of how to format the date.
2. I can add the `role=dialog`. If we don't have focus trapping, is that a problem? I'll also send "Open calendar picker dialog" / "Close calendar picker dialog" for content approval and translations unless you have exact text you'd like to see there.
3. Will do 👍 
4. This is the whole point of the new label-mask component and one of the pillars of the new date input designs. If you strongly believe this is unhelpful, we can talk about scrapping it. This is the behavior that our stakeholders approved.

PR #1882, opened 2022-06-03
[WNMGDS-1630] SingleInputDateField updates

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1630

This is a follow-up to https://github.com/CMSgov/design-system/pull/1876#issuecomment-1145219469

- [x] Needs to be associated with a ticket
- [x] Needs Spanish translations

### Added

- Added doc comments for each `SingleInputDateField` prop
- Added `role="dialog"` to the calendar picker
- Added a more descriptive title for the icon in the calendar-picker button (BUT RIGHT NOW IT'S ONLY IN ENGLISH)

### Changed

- The inert mask-only portion of the label mask is now visible to screen readers and linked to the input via an `aria-describedby` attribute. This is to provide more instruction to assistive tech users about the expected format (`MM/DD/YYYY`)
- Closing the calendar picker by either clicking selecting a date or pressing ESC will shift focus back to the text input

### Fixed

- Fixed `aria-describedby` in `TextInput` not including the error id if the `aria-describedby` prop is defined
- Fixed styling of calendar picker button in Safari
- Fixed potential to render a stray `0` in `SingleInputDateField` if the `fromYear` is year 0 AD 😂 

## How to test

https://design-system-demo.s3.amazonaws.com/pwolfert/date-field-updates/core/storybook/index.html?path=/story/components-singleinputdatefield--with-picker&args=fromYear:0


packages/design-system/src/styles/components/_SingleInputDateField.scss
```diff
@@ -397,6 +397,7 @@
     border-bottom-left-radius: 0;
     color: $color-primary;
     cursor: pointer;
+    margin: 0;
```
scheul93, line 400: I can't leave a comment because the change is not part of this PR, but on line 368, I see a warning that says
> start value has mixed support, consider using flex-start instead

pwolfert:
Some feedback I'm particularly interested in is what ya'll think about the RDP styles. I wrote [a comment about my dilemma here](https://github.com/CMSgov/design-system/blob/pwolfert/single-input-date-field/packages/design-system/src/styles/components/_SingleInputDateField.scss#L3-L8). Should we just manipulate the styles until they work for us? Or should we make it a practice not to touch the original styles?
- Upside of leaving the original styles
  - Clear separation between our code and theirs
  - Could make it easier to upgrade because our overriding rules encode our intentions
- Downside of leaving the original styles as is and only overriding:
  - More CSS in our bundle
  - More complicated override rules
  - No guarantees that doing it this way will make it any easier when we need to upgrade the library and it comes with new styles

I'm starting to lean towards just modifying the original styles at this point because it'll simplify, and we'll end up with rules that make sense in our final bundle.

packages/design-system/src/components/locale/en.json
```diff
@@ -85,6 +85,11 @@
   "review": {
     "editText": "Edit"
   },
+  "singleInputDateField": {
+    "close": "Close calendar picker dialog",
+    "open": "Open calendar picker dialog",
+    "arrowKeyInstructions": "Arrow keys can navigate dates"
```
pwolfert, line 91: This isn't being used yet, but it will be in the next PR

packages/design-system/src/components/DateField/SingleInputDateField.tsx
```diff
@@ -40,14 +91,22 @@ const SingleInputDateField = (props: SingleInputDateFieldProps) => {
...
-  const withPicker = fromDate || fromMonth || fromYear;
+  const withPicker = !!(fromDate || fromMonth || Number.isInteger(fromYear));
   const [pickerVisible, setPickerVisible] = useState(false);
 
+  // Set up change handlers
   function handleInputChange(event) {
     const updatedValue = event.currentTarget.value;
     onChange(updatedValue, DATE_MASK(updatedValue, true));
   }
+  function handlePickerChange(date: Date) {
```
zarahzachz, line 102: I'm not sure, but I _think_ this function is producing a lot of warnings on Storybook. This only happens when the calendar/picker is on the component. I'm guessing there isn't a default date value when this component loads and it's causing this to happen?

Screen Shot 2022-06-07 at 12 47 08 PM

zarahzachz:
To your CSS question, I'm leaning more towards overriding than modifying their styles - but that's me being overly cautious. From my perspective, it's nice to know what CSS came in from a third party and what was overwrote by us - especially if we need to update this package at some point.

Tangent - Looking at the mocks referenced, it doesn't look like this implementation currently supports what the mocks imply, i.e., clicking the month or year will allow you to choose a different month or year. Is this an issue?

pwolfert:
@zarahzachz, about the styling, that'll be covered in https://jira.cms.gov/browse/WNMGDS-1641

packages/design-system/src/components/DateField/SingleInputDateField.tsx
zarahzachz: Oh that makes sense! Then I'm good with this PR - approving!

zarahzachz (approved):

scheul93 (approved):

PR #1889, opened 2022-06-09
[WNMGDS-1641] Implement styled custom controls for DayPicker

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1641

### Changes

- Added `CustomDayPicker` and `CustomDayPickerCaption`, which implement the custom controls
- Styled the new controls
- Styled the component's error state
- Made some changes to `useFormLabel` so that error-related info would be automatically passed down to the `TextInput`
- Created a Sass variable for the error border width, which we actually need to talk about because we recently filed a related bug
- Start exporting the Calendar icon, and include them in the example story
- Update visual regression tests

## How to test

Check out the stories and try all hover, focus, and active states, and interact with the new controls.

https://design-system-demo.s3.amazonaws.com/pwolfert/day-picker-custom-controls/core/storybook/index.html?path=/story/components-singleinputdatefield--with-picker


pwolfert:
You reminded me that I didn't write any unit tests specifically for `SingleInputDateField`. Unfortunately when I try to open the picker dialog in the tests, I get a bunch of logged errors. It's taking too long to figure it out though.

zarahzachz (changes requested):
Styles look good! 

Tested in `forced-color` mode in Edge and found we have a HCM bug that should be solved by applying the [`Highlight` value](https://www.w3.org/TR/css-color-3/#css2-system) to the selected date.

Screen Shot 2022-06-10 at 9 46 53 AM

I also wonder if it's worth increasing the height of the datepicker examples in Storybook? The dialog gets cut off when the calendar is open. I think we increased the height on similar stories like Tooltip.

packages/design-system/src/components/Icons/icons.stories.tsx
```diff
@@ -68,6 +69,11 @@ const iconData = [
     component: <BuildingCircleIcon />,
     name: 'BuildingCircleIcon',
   },
+  {
```
scheul93, line 72: Did you also want to add the Calendar icon to the gulp doc site in the file `design-system-docs/src/pages/components/Icon/IconComponent.example.jsx` or is it okay to leave out until we migrate?

pwolfert: In either case there's a good chance I'll forget to move it over later. I'll add it to the example now and put it on my to-do list to add it to the gatsby one when that's merged

scheul93 (approved):

PR #1892, opened 2022-06-09
[WNMGDS-1301] Color primary-alt to secondary

forrestbaer:
## Summary

- Updates color-secondary to be used where color-primary-alt was previously being used
- Updates component theme variables to use color-secondary
- Added notes for deprecation of color-primary-alt
- Updated current examples
- Looks like this one gets loki snapshot updates for the datepicker updates in master also
- Adds a color-secondary-lighter as color-secondary-darker exists, I think lighter should also, but let me know, it's a value between light and lightest.

**Note:** Updates to the new doc site will need to happen once we get the token feature branch merged, I think it's best to just do those all at once. I can write a new ticket for new doc site token cleanup for future work but this get's the base ready.

## How to test

`yarn install && yarn build && yarn build:healthcare && yarn build:medicare`
`yarn loki test && yarn loki:healthcare test && yarn loki:medicare test`
`yarn start each` check text colors, background colors, color swatches, buttons, drawers


pwolfert:
https://github.com/CMSgov/design-system/pull/1889 will take care of the loki reference updates

line47:
Do we need to take into consideration updating the color naming in Sketch files as well?

packages/design-system-tokens/src/themes/core.ts
```diff
@@ -76,13 +76,22 @@ export const themeColors: ColorTokens = {
   'muted-inverse':              color['lapis-50'],
   //
   'primary':                    color['ocean-500'],
+  'primary-darker':             color['ocean-700'],
+  'primary-darkest':            color['ocean-800'],
+  // @TODO: deprecate primary-alt as secondary
```
zarahzachz, line 81: Could the references to `primary-alt` be deprecated as part of this work? Or are they being referenced elsewhere still?

scheul93 (approved):

PR #1894, opened 2022-06-10
[WNMGDS-1670] Gatsby 404 Page

forrestbaer:
## Summary

- Added 404 page markup
- Added 404 page content
- Added new folder `not-in-sidebar` to `content` folder to allow for content items that do not currently live in the sidebar and updated sidebar to exclude that data in it's query

## How to test

`yarn install && yarn build && yarn build:storybook-gatsby && yarn build:gatsby && yarn start:gatsby`

Hit any missing page and since you're in `developer` mode it will give you a button to click to preview the 404 page, in production it will just show this page normally but in development mode you have to click to it.. Notice 404 page content does not show up in sidebar, making it available to edit more easily.

image

### DIFF
--- a/packages/docs/content/not-in-sidebar/404.mdx
+++ b/packages/docs/content/not-in-sidebar/404.mdx
@@ -1,0 +1,5 @@
+---
+title: Page not found
+---
+
+Sorry, the page you're looking for couldn't be found. It's possible that this page has moved, or the address may have been typed incorrectly.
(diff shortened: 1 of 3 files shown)


scheul93:
What is the benefit of having the 404 page content in a mdx file vs hard coded into the `404.tsx` file? I know that it's consistent with other content, but would this content need to be updated regularly, if not at all?

forrestbaer:
> What is the benefit of having the 404 page content in a mdx file vs hard coded into the `404.tsx` file? I know that it's consistent with other content, but would this content need to be updated regularly, if not at all?

Seems like there should be the capability to have non-sidebar pages, I just figured it would provide that opportunity and make all pages editable.  In my opinion all content should be editable.

PR #1896, opened 2022-06-13
[NO-TICKET] Merge master into Gatsby branch

pwolfert:
## Summary

Updates the Gatsby branch with master and fixes some TypeScript errors that weren't present on master. Still not sure why they happen on this branch only.

Here are the actual changes:
- b15d1d6ed8222890d8432c995c393437a28470ad
- 4386cead6e0d480f46ec25c85fba29c91bf8e70d - Removes the generic typing and the union type for the `TextInput` component's props. This more specific typing wasn't even carried over into the public type, so it wasn't doing much for us. We should revisit the type definition for `TextField` at some point to make it more accurate. Right now we're only extending the attributes of the `input` element and not `textarea`.
- 5281116aa591be374bdc1b854a3c0fccb54ccd15


scheul93:
build is succeeding now, so that's great! 🎉 

One clarification, it looks like these previously exported types are no longer exported
```
export type MultilineTextInputProps = CommonTextInputProps<true>;
export type SingleLineTextInputProps = CommonTextInputProps<false | undefined>;
```
That shouldn't be a problem because those were just added as part of this date field work, right?

pwolfert:
@scheul93, thank you for calling out that risk. There is a possibility that users were referencing types from this file directly. I think it's probably a low risk. At least these types aren't actually exported in `TextField/index.ts`

forrestbaer (approved):
Success all around as far as I can tell, enough to get us to the next PR at least with less clutter.

forrestbaer (approved):
🏆

PR #1903, opened 2022-06-14
[NO-TICKET] Use trailing slashes for all internal links in Gatsby

pwolfert:
## Summary

On the new doc site, when running performance audits that URLs without the trailing slash we were being redirected to include a trailing slash. All internal links in the Gatsby site do not currently contain a trailing slash. However, even though the server redirected to include a trailing slash, the client side router was stripping that trailing slash back out of the address bar when the page hydrated. This back and forth creates an unnecessary hop for each request and was confusing to reason about.

Here's a relevant discussion on the Gatsby project that brought us to a solution: https://github.com/gatsbyjs/gatsby/discussions/27889#discussioncomment-254854

We found that we don't need the plugin that they mention, nor do we need the new `trailingSlash` config. Those seemed to exacerbate the issue. We only needed to update the URLs to include trailing slashes. I surmise that this is because the `@reach/router` indexes those URLs at build time when rendering each page, so we just need to make sure the links include the slash when rendered. I'm only guessing that that's how the client-side router works based on reading this:

> ...prepare (at build time) the embedded [@reach/router](https://github.com/reach/router) for trailing slash paths

### Changed

- Updated all internal Gatsby links to include a trailing slash

## How to test

```
yarn build-storybook:gatsby && yarn build:gatsby && yarn serve:gatsby
```


scheul93:
After building storybook, building Gatsby & serving gatsby, I'm seeing 404s for storybook iframe. Is that expected or caused by this change?
Screen Shot 2022-06-14 at 12 16 36 PM

forrestbaer (approved):
Yeah!

PR #1905, opened 2022-06-14
[WNMGDS-1698] Hide the instructions for the day picker visually

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1698

Our stakeholders would like to hide these instructions visually for now. We still plan to come back and iterate on the design for this to make sure it's an inclusive experience.

### Changed

- Visually hide the instructions for the day picker that are in the footer


forrestbaer (approved):
Looks good 👍🏻

PR #1908, opened 2022-06-15
[WNMGDS-1737] Update the download link on Gatsby

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1737)


pwolfert (approved):
Tried with both the gatsby dev server and statically built gatsby 👍

PR #1911, opened 2022-06-17
[NO-TICKET] Legacy docs background color page bug

forrestbaer:
## Summary

- Fixes a small bug that came in with the token branch, old doc site 'background-color' page was broken because of a formatting issue
- Looks like prettier doesn't do a very good job with EJS, it's ugly but it's not broken now, and we won't have to deal with it for long / likely again.

## How to test

`yarn install && yarn build && yarn build:healthcare && yarn start:healthcare`
[http://localhost:3000/utilities/background-color/](http://localhost:3000/utilities/background-color/)


pwolfert (approved):
Nice catch!

PR #1922, opened 2022-06-23
[WNMGDS-1723] Update sidebar

zarahzachz:
## Summary

Update the sidebar for Gatsby to match the Mural designs in [this Jira ticket](https://jira.cms.gov/browse/WNMGDS-1723).


packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -136,22 +165,51 @@ const DocSiteSidebar = ({ isMobileNavOpen, location }: DocSiteNavProps) => {
   const navItems: VerticalNavItemProps[] = formatNavData(data?.allFile?.group);
 
   return (
-    <nav
-      className={classnames('ds-l-md-col--3 ds-u-padding--2 ds-u-fill--white c-sidebar', {
-        'c-sidebar--open': isMobileNavOpen,
+    <div
+      className={classnames('ds-l-md-col--3 ds-u-padding--0 ds-u-md-padding--2 c-navigation', {
```
pwolfert, line 169: I think the side bar width should be fixed rather than a 3/12 grid column. It looks too wide on some screens and too narrow on others. What do you think, @line47?

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -136,22 +165,51 @@ const DocSiteSidebar = ({ isMobileNavOpen, location }: DocSiteNavProps) => {
...
+          className="c-navigation__link-list"
+          items={navItems}
+          component={GatsbyLink}
+          selectedId={location ? location.pathname : ''}
+        />
+        <a
+          href="https://github.com/CMSgov/design-system"
+          className="c-navigation__github-link ds-c-link"
+        >
+          View code on GitHub
```
pwolfert, line 208: Are we going to include the GitHub logo here?

pwolfert (commented):
It's happening!

packages/docs/src/components/Layout.tsx
```diff
@@ -67,9 +61,11 @@ const Layout = ({
       </Helmet>
 
       <SkipNav href="#main" />
-      <Header />
+
+      <UsaBanner className="ds-u-display--none ds-u-md-display--block" />
+
       <div className="ds-l-row ds-u-margin--0">
-        <Sidebar isMobileNavOpen={isMobileNavOpen} location={location} />
+        <Navigation location={location} />
```
scheul93, line 68: In the mural, it looks like the nav extends all the way to the bottom of the page as its own column instead of stopping at the footer like it does now. Is that something that should be included in this PR or was that part of a future ticket?
Screen Shot 2022-06-23 at 4 02 21 PM

line47: I'm open to interpretation here - it might be nice for the footer to go full width. This will then align with the homepage design Melissa did.

packages/docs/src/components/DocSiteNavigation.tsx
line47: I agree, fixed width for the left nav

packages/docs/src/components/DocSiteNavigation.tsx
line47: We can for sure! Let me know if you need me to grab the logo from font awesome

packages/docs/src/components/Layout.tsx
pwolfert, line 68: I like that this "Navigation" component now handles all of it. I think that's a good boundary for these components and responsibilities.

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -0,0 +1,221 @@
...
+      })}
+    >
+      <header className="c-navigation__header">
+        <Button
+          className="ds-u-md-display--none ds-u-padding-left--0 ds-u-padding-right--1"
+          variation="transparent"
+          aria-expanded={isMobileNavOpen}
+          aria-controls="c-mobile-navigation"
+          onClick={toggleMenu}
+        >
```
pwolfert, line 180: One tiny thing that I think could help make this look really polished is to normalize the gap around this focus ring:
menu-button-focus

pwolfert (approved):
This LGTM 👍 

I have one minor suggested improvement, but it's optional.

packages/docs/src/components/DocSiteNavigation.tsx
zarahzachz: By "normalize," do you mean center position the icon within the button? 

My one concern with that is if the button doesn't have focus, the icon will look off-kilter with the container width on the page. I kept the right padding in place to give users more click affordance, but maybe it's best to reduce it to the same width as the left-padding to visually center it? 

@line47 do you have a preference on this?

pwolfert: I didn't have a plan completely thought out, but I figured you could do something with negative margin to get the button to appear in the right place but still have even spacing around the icon. (Yes, I mean centering the icon)

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -0,0 +1,221 @@
...
+      {props.children}
+    </Link>
+  );
+};
+
+/**
+ * DocSiteNav
+ * queries all MDX files and generates the proper format to create a vertical nav
+ * @returns {React Element}
+ * @todo figure out which item is currently selected and mark & expand appropriately
```
forrestbaer, line 48: Love that we're all starting to use this doc format!

packages/docs/src/styles/components/navigation.scss
```diff
@@ -0,0 +1,111 @@
+// Mobile
+.c-navigation {
+  background-color: $color-white;
+  width: 100%;
+
+  .c-navigation__header {
+    background-color: $color-primary-alt-lightest;
```
forrestbaer, line 7: `primary-alt` is now `secondary` so anywhere you use primary alt you can substitute with `secondary` going forward.

forrestbaer (commented):
I think this looks great, just a small comment about `primary-alt` change and the having the focus on the menu icon be uniform as patrick mentioned would be nice.

forrestbaer (approved):
Looks good! 🥇

PR #1926, opened 2022-06-23
[WNMGDS-1720] css variable exporter

scheul93:
## Summary

This work is step one on our path to a theme switcher. This PR contains:
- Updates to the design-tokens directory to export to 'css'. This creates
  - a `.map.scss` file that maps sass variables to css variables
  - a `.css` file that maps css variables to a token value. These variables are also scoped to a css selector.
- Updates to the `yarn dist` command for tokens which will copy the above files into the docs site directory and the storybook directory
- Updates to the doc site directory to integrate these new variables for testing
- Updates to some component styles. I was getting an error where sass could not compute a calculation that used but a css var & a sass variable, so I updated those calculations to use the css function `calc()` and sass interpolation for variables. I think this works for both the internal tools like the doc site but without breaking the styles for the DS package itself

### Creating Css variables
To create the sass to css map and css variables, run `yarn build themes css` in the tokens directory. You should see the relevant files generated in the `dist` tokens folder.

### Copying these files into internal tool directories
To ensure that these updated files get to the doc site & storybook directories, run `yarn dist` in the tokens directory. If there are other scss files in the `dist` folder, you should not see those transferred to the doc site & storybook directories. Additionally, you should not see the `.map` & `.css` files moved into DS package directories

### See the variables in action
Run the doc site with `yarn start:gatsby` from the project root. In the `<Layout>` component, you'll see the data attribute `data-theme="core"` on one of the elements. You can update the value to be `"healthcare"` or `"medicare"` and you should see things like buttons on the doc site change.
CssVars

## Outstanding questions
1. I setup the export css functionality to only work for themes. Is this functionality needed for tokens?
2. Are there any other gotchas for the tokens directory that I'm missing?


packages/design-system-tokens/src/index.ts
```diff
@@ -21,6 +22,8 @@ const tokenExporter = (inputType: string, exportType: string): number => {
       return exportCsv(fileData, OUTPUT_PATH);
     case 'json':
       return exportJson(fileData, OUTPUT_PATH);
+    case 'css':
```
forrestbaer, line 25: Nitpick, this makes sense but I think `css-vars` would be more specific.

forrestbaer (approved):
This is awesome! 🥇 Can't wait till it's all wired up, glad it seems to work so seamlessly and fast!

zarahzachz (approved):
The switching worked like a charm! 

This probably has nothing to do with this PR, but I'm curious about the Medicare buttons. I assume they don't match our current Mgov DS site header because of the button architecture work?

scheul93:
I ended up adding an ignore list because some sass vars cannot be turned into css vars since they are used in sass loops. Most of these variables seem to be items that are standard across all three brands, so I think we'll be okay in terms of theme switching.

I also had to update a few more files to use `calc()` and sass interpolation. I found that even though sass wasn't throwing errors, these values weren't being calculated correctly in the browser without these changes. We'll probably catch more as we go too.

Anyway, here's a gif of the update. I think the Buttons are even displaying correctly with these changes.
CssVarsRedux

forrestbaer (approved):
So good, love the changes 💯

packages/docs/src/styles/components/navigation.scss
```diff
@@ -79,7 +79,20 @@
...
+            position: absolute;
+            top: 0;
+            bottom: 0;
+            left: 0;
+            right: 0;
+            content: '';
+            background-color: #{$vertical-nav-item__color--hover};
+            filter: opacity(0.05);
+          }
         }
```
scheul93, line 96: @zarahzachz , I had to make an update here because I was getting compilation errors when trying to use the css vars with `scale-color`. Instead, I had to do a bit of a workaround.

PR #1930, opened 2022-06-24
[No-ticket] Fix for node-sass deprecation warnings

forrestbaer:
## Summary

Resolves Sass warnings about the slash operators being deprecated and replaced with `math.div` when building storybook/gatsby/etc.. `node-sass` in the `design-system-scripts` was the culprit and I never thought to just throw it out and use the version of Sass we include at root, until I did think that and then it worked!

- Upgrades Sass & Sass Loader
- Upgrades gulp-sass and utilizes Sass as interpreter
- Fixes slash warnings in components

Using `es-check` to check to make sure we don't break es5 again with these changes, think I will do another PR that does this on `yarn test` so we can detect that if it happens again.

## How to test

`npm i es-check -g`
`yarn install && yarn build && yarn build:healthcare && yarn build:medicare`
`yarn start:healthcare` `yarn start:medicare`
`yarn build-storybook:gatsby && yarn build:gatsby`
`yarn storybook` `yarn storybook:healthcare` `yarn storybook:medicare`
`es-check es5 ./packages/design-system/dist/js/bundle.js`
`es-check es5 ./packages/ds-medicare-gov/dist/js/bundle.js`
`es-check es5 ./packages/ds-healthcare-gov/dist/js/bundle.js`
`yarn test`


pwolfert (approved):
Wow, no more errors! This is great!

Also that es-check tool is handy.

PR #1931, opened 2022-06-24
[No-ticket] Update release branch with commit hash

forrestbaer:
## Summary

Adds the current commit short hash to the release name to prevent situations where two releases are needed in a day (main and beta patch releases for example).

## How to test

`yarn release` note release name, can stage a release with default options then `yarn release -u` to undo, note release branch deleted successfully.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ These scripts can all be run from the root level of the repo:
   - `yarn loki:medicare test` to run the Medicare.gov visual regression tests
 - `yarn lint`
   - Runs just the linting portion of the tests, eslint and stylelint
+- `yarn deploy-demo`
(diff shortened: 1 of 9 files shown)


pwolfert (approved):
It worked for me. That undo feature is handy to have. Do we actually care about the date now that we have that unique commit hash? I personally don't think we do. Kind of just looks like a mess of numbers and hyphens to me

forrestbaer:
> It worked for me. That undo feature is handy to have. Do we actually care about the date now that we have that unique commit hash? I personally don't think we do. Kind of just looks like a mess of numbers and hyphens to me

I tend to agree, let's make it `release-hash`

PR #1934, opened 2022-06-27
[WNMGDS-1773] Fix long choice hint text wrapping the whole label in Safari

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1773

The `flex-basis: min-content` declaration is not supported by Safari yet, so we can't rely on it to build our layout

### Fixed

- Fixed choices with long, wrapping label content pushing the whole label to the next line beneath the checkbox or radio button

## How to test

1. Open up one of the choice list stories and modify it to have some really long hint text (or any text inside the label)
2. View it in Safari. Notice that on master the whole label will wrap below the checkbox or radio button


forrestbaer (approved):
This works for me in safari and chrome at various breakpoints, looks good! 👍🏻

PR #1938, opened 2022-06-29
[WNMGDS-1562] information architecture nav updates

scheul93:
## Summary

Updated a few things to support new information architecture
1. Changed way we are querying graphql to use `allMdx` instead of `allFiles`. This changes our ability to group things, so a little more work in the component. However, it does allow us to use the frontmatter title as the text shown in the nav bar
2. Had to make some updates into the data conversion. I tried to comment every step

## How to test

`yarn start:gatsby` to build and run the demo site locally.

[Demo site link](https://cmsgov.github.io/design-system/branch/WNMGDS-1562/ia-nav-updates/)

### DIFF
--- a/packages/docs/content/4_components/1_overview.mdx
+++ b/packages/docs/content/4_components/1_overview.mdx
@@ -1,5 +1,5 @@
 ---
-title: Overview
+title: Components Overview
 ---
 
 Component class names follow the format: `ds-c-[BLOCK]__[ELEMENT]--[MODIFIER]`
--- a/packages/docs/content/5_patterns/1_overview.mdx
+++ b/packages/docs/content/5_patterns/1_overview.mdx
@@ -1,5 +1,5 @@
 ---
-title: Patterns
+title: Patterns Overview
 ---
 
 Patterns are solutions and researched best practices that solve recurring user interface design problems:
--- a/packages/docs/content/6_utilities/1_overview.mdx
+++ b/packages/docs/content/6_utilities/1_overview.mdx
@@ -1,5 +1,5 @@
 ---
-title: Utilities
+title: Utilities Overview
 ---
 
 A utility class modifies a single trait, typically a single CSS property. To apply a trait, or a combination of traits to an element, add the corresponding utility class directly to the HTML element.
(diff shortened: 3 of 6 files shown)


scheul93:
@line47 , I didn't add any new pages outlined in the IA because I don't think we have content for them, so I just tried to re-organize what we already had.

pwolfert:
> it does allow us to use the frontmatter title as the text shown in the nav bar

Would it also allow us to use info in the frontmatter to determine the ordering of files so we don't have to prefix some of the filenames with numbers?

pwolfert:
It looks like we'll need to adjust the sidebar width to account for the architecture changes:

<details>
  <summary>Screen capture of changing width</summary>
  
  sidebar-resize
</details>

scheul93:
> Would it also allow us to use info in the frontmatter to determine the ordering of files so we don't have to prefix some of the filenames with numbers?

I guess so. Would going into every file and changing the frontmatter value be easier or more difficult than the current file structure approach?
I guess it simplifies this data formatting and the git history, but you have to remember to change every file in that directory if adding a new file that needs a specific position.

scheul93:
> It looks like we'll need to adjust the sidebar width to account for the architecture changes:

Is there a recommendation for what that width should be? Or should we use "best judgment"?

pwolfert:
Regarding width, that's a @line47 question

pwolfert:
Regarding frontmatter vs filename, one way I've seen it done before is having a _weight_ in the frontmatter, where higher numbers or lower numbers get sorted to the top, and they can be arbitrary. In that scenario, we could space them out by 10s or 100s, and then things could still be inserted without reordering everything. It might be less straightforward than filenames though.

Pros of filenames:
- You can see the display order when looking at a directory listing
Cons of filenames:
- Messier git history when we want to change page order
- More visual noise when looking for the file you want to edit and not alphabetized, which might be confusing for our non-dev contributors

I lean towards frontmatter personally. @zarahzachz and @forrestbaer can also chime in here.

line47:
@pwolfert we certainly could do that! I think we still would want to set a width explicitly to prevent left nav with expansion in the future

scheul93:
I'm going to close this PR because, after reading & reflection, I think it makes more sense to order based on frontmatter. I'll have 1 PR to implement all the content files changes and then re-open a PR similar to this with the nav updated. I will also implement the nav width update too

PR #1939, opened 2022-06-30
[WNMGDS-1425] Implement new Button API and styles

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1425)

Implement new API for styling buttons.

### DIFF
--- a/packages/design-system/src/components/Button/Button.stories.jsx
+++ b/packages/design-system/src/components/Button/Button.stories.jsx
@@ -16,7 +16,6 @@ export default {
     disabled: {
       control: { type: 'boolean' },
     },
-    // hiding deprecated props
     inverse: {
       table: {
         disable: true,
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/Button/Button.stories.jsx
```diff
@@ -64,13 +61,303 @@ SpinnerButton.args = {
       <Spinner /> Loading...
     </>
   ),
-  variation: 'primary',
+  variation: 'solid',
 };
 
 export const LinkButton = Template.bind({});
```
scheul93, line 67: [question] Should we remove or update this story name? I only ask because it seems a bit confusing that we have a button variation that is a "link" variation, but we also have this story called Link Button that is showcasing an anchor tag styled as a button. 

I'm not sure how useful this story so maybe we either a) remove it or b) rename it to something like "Anchor Styled as Button" 🤷🏻‍♀️

scheul93 (approved):
Looks lovely!

I ran the gatsby doc site and it built & ran and I think will be fine once the documentation pages are updated in a future PR -- so that's great 🙂

packages/design-system/src/components/Button/Button.stories.jsx
zarahzachz: I went ahead and updated the Story (and tests) to use "anchor" to better indicate what it's showing/testing.

PR #1942, opened 2022-07-05
[WNMGDS-1652] moving ordering from filename to frontmatter

scheul93:
## Summary

Updates to content files to move from a filename-based ordering to having ordering defined in MDX's frontmatter. This will help preserve version history.

Right now, I have not updated the nav to account for these changes, I wanted to get all the content files renamed in one PR. Ordering values are also subject to change once nav functionality begins.


packages/docs/content/getting-started/developers/examples.mdx
```diff
@@ -1,5 +1,6 @@
 ---
 title: Examples
+order: 16
```
pwolfert, line 3: What determines whether we count by ones or tens?

pwolfert (approved):

packages/docs/content/getting-started/developers/examples.mdx
scheul93: So I did 10s for "level 2" and 1s for level 3. I'm open to another scheme, but I needed a way to 1) make sure level 3 could be ordered & 2) that level 2 items that had level 3 items could also be ordered and since those types of level 2 items don't have an mdx file associated with them, this is what I came up with

zarahzachz (approved):
LGTM!

PR #1943, opened 2022-07-05
[WNMGDS-1778] Add error-message support to Choice component

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1778

### Added

- Added `errorMessage` and `errorMessageClassName` props to the `Choice` component. Previously an individual `Choice` was not able to have its own error message

## How to test

- View the new story
- You can also modify the ChoiceList story to pass an `errorMessage` as one of the `ChoiceList` `choices`


packages/design-system/src/components/ChoiceList/__snapshots__/Choice.test.tsx.snap
```diff
@@ -13,7 +13,7 @@ exports[`Choice applies className to label 1`] = `
 exports[`Choice has a hint and requirementLabel 1`] = `
 <FormLabel
   component="label"
-  fieldId="checkbox_foo_10"
+  fieldId="checkbox_foo_11"
```
pwolfert, line 16: These unit tests really need to be refactored so they don't have so many snapshots, but that's out of scope for a last-minute feature like this

zarahzachz (approved):
LGTM!

PR #1944, opened 2022-07-05
[WNMGDS-1779] Do not show analytics content error when analytics are disabled

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1779

### Fixed

- Fixed Dialogs without headings printing an analytics content error to the console even when dialog analytics are disabled

## How to test

1. Open the storybook story for Dialog
2. In the story controls, delete the `heading` prop text
3. In the analytics settings above the story, turn on `Log to console`
4. Click the "Click to show modal" and check the console. On this branch, no errors should be printed


zarahzachz (approved):
Ran locally - I see an error only when analytics are enabled, but not when its disabled. Appears to be working as expected!

PR #1945, opened 2022-07-06
[WNMGDS-1652] ia nav updates

scheul93:
## Summary

Updated the doc site nav bar to accommodate our new information architecture. This sorts the nav items by the order variable defined in the mdx content files. If no order is defined, the files are sorted alphabetically.

## How to test

[Demo site found here](https://cmsgov.github.io/design-system/branch/WNMGDS-1562/ia-nav-updates/) to see the changes in action


packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -123,57 +132,117 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
...
-    const retVal: VerticalNavItemProps[] = [];
+  /**
+   * Need to nest level2 items that have subnav items under the relevant level1 items
+   */
+  const nestSubNavs = (dataList): VerticalNavItemProps[] => {
+    const level1Items: { string?: VerticalNavItemProps } = {};
+    const level2Items = [];
+
+    // sort level1 vs level2 items
     dataList.forEach((dataItem) => {
```
pwolfert, line 169: This `dataItem` is being interpreted by TypeScript as `any` because the type isn't defined in the function signature above. However, I think the `GraphQlNavItem[]` type that `formatNavData` expects—and what this function expects by extension—is not accurate. If you were to apply that type to `dataList`, it'd start throwing errors. Each of those `GraphQlNavItemNode` objects should have an `edges` and `fieldValue` property.

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -123,57 +132,117 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
...
 
   /**
-   * Updating a name to remove kebab case & get rid of numeric ordering
+   * Updating a name to remove kebab case
    */
   const formatNavItemLabel = (name: string): string => {
-    let newName = name.replace(/-/g, ' ');
-    newName = removePositioning(newName);
-    return newName;
+    return name.replace(/-/g, ' ');
```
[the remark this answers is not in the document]
pwolfert: I see now that items with sub-items do not have a markdown file to specify a `title`. Looks like even if it were removed from `formatNavItemData` or made into a fallback, this function would still need to be used in `restructureAsVerticalNavItem`

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -123,57 +132,117 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
...
+        selected: isSelected,
+        id: dataItem.fieldValue,
+        order: orderVal,
+      });
+    });
+
     return retVal;
   };
 
+  const formatNavData = (graphQlData: GraphQlNavItemNode[]): VerticalNavItemProps[] => {
```
pwolfert, line 239: These functions are now substantial enough to consider moving out of the component. It looks like they get all the data they need through parameters.

pwolfert (commented):
Okay, I've gone through everything

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -123,57 +132,117 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
...
+    level1Items.forEach((level1Item) => {
+      // sort items based on the order defined in frontmatter then id which is the file path for the item
+      level1Item.items = sortBy(level1Item.items, ['order', 'id']);
+    });
+
+    level1Items.sort((itemA, itemB) => {
+      return level1ItemOrder.indexOf(itemA.label) - level1ItemOrder.indexOf(itemB.label);
+    });
+    return level1Items;
+  };
```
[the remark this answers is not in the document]
pwolfert: It might also be worth noting in the doc comments for the `restructureAsVerticalNavItem` that the level1 and level2 items (or all items that have sub-items) are all mixed together at that point and still need to be organized. That's what took me a while to understand.

forrestbaer:
The frontmatter data organization is a complicated problem! I've tested it out and it works great..

Dare I say it ... do we need to engineer this to be dynamic? Hard-coding the menu options would be simple to read and modify and re-order even. Adding new content would require editing this file, but that seems ok to me. I feel like this is a case where we don't need to have a lot of complicated code. We could simply have a comment that describes how to determine the correct URL for the resource.

What do you think @scheul93 @zarahzachz @pwolfert ?

packages/docs/src/components/ContentRenderer.tsx
```diff
@@ -87,6 +88,7 @@ const customComponents = {
   EmbeddedExample,
   StorybookExample,
   PropTable,
+  ResponsiveExample,
```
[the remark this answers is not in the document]
scheul93: There are content files using this component, the removal of their imports happened during one of the content file renaming PRs

pwolfert:
> Do we need to engineer this to be dynamic? ... What do you all think?

I don't think I agree. I think there's too much risk of becoming out of sync, especially as we start to encourage content contribution. I think that a top-level sorting list is the furthest I'd go, personally, down the hard-coding road in this case. I think the benefits of having it dynamic outweigh the cognitive load for us, especially if we can streamline or comment this transformation pipeline a bit to make it easier to understand.

forrestbaer:
I see the order system as more of a possible issue to contribution. And since there is only one docs site, I don't quite understand how this would get out of sync. I see it as a preemptive optimization.  I'll review this as-is tomorrow morning regardless!

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -123,57 +132,117 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
...
+
+  /**
+   * re-structures graphQL object to VerticalNavItem structure
+   */
+  const restructureAsVerticalNavItem = (dataList): VerticalNavItemProps[] => {
```
forrestbaer, line 213: Maybe this could be moved above `nestSubNavs` so it mirrors the calling of these functions below?

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -88,8 +96,8 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
   const data = useStaticQuery(graphql`
     query SiteNavQuery {
       allFile(
-        sort: { fields: [relativeDirectory, name] }
         filter: { ext: { eq: ".mdx" }, relativeDirectory: { ne: "not-in-sidebar" } }
+        sort: { fields: [relativeDirectory, name] }
```
forrestbaer, line 100: Not super important but we could remove `name` `relativeDirectory` and both `id`'s from this graphql query.

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -123,57 +132,117 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
    * Checks sub nav items to see if any are currently selected
    */
   const isSubNavItemSelected = (subNavItems) => {
-    return subNavItems.some((navItem) => isItemSelected(navItem.url));
+    return subNavItems.some((navItem) => navItem.selected);
```
forrestbaer, line 135: This could be put inline as it's only being used once.

forrestbaer (commented):
So I couldn't sleep, and I needed to puzzle so I attempted to re-write the formatter and ran into all kinds of road blocks as you must have with the data that makes this complex.  Used a ton of reducers and maps and eventually gave up, in the end it wasn't any easier to read than what you've done. It's unfortunate gatsby can't do data transforms like SQL. 😭  Spent a lot of time with the graphql interface looking for ways to improve the data but didn't find anything. I noticed a few things that could be re-organized but overall it looks good. Nice work on a difficult problem! 🙇🏻 

One thought about `order` could always use floating point numbers in the future if we ever ran into issues with the gaps.

I still think we should just make this basic JSX though, there is so much complexity in the data and organizing it. It would just be so much easier to manage and reason about. The levels of complexity in codebase is expanding rapidly.

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -123,57 +132,117 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
...
+      const subNavItems = dataItem.edges.map((subNavItem) => formatNavItemData(subNavItem.node));
+      // in order to sort properly, need to get the smallest leve l3 order value and assign to level 2
+      const subNavItemsOrderVals = subNavItems.map((subNavItem) => subNavItem.order);
+      const orderVal = Math.min(...subNavItemsOrderVals);
+
+      const labelText = formatNavItemLabel(dataItem.fieldValue);
+      const isSelected = isItemSelected(dataItem.fieldValue);
+
+      // add level 1 item & sub items
+      retVal.push({
```
forrestbaer, line 226: As documentation `retVal` could be `sortedNavItems`?

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -88,8 +96,8 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
   const data = useStaticQuery(graphql`
     query SiteNavQuery {
       allFile(
-        sort: { fields: [relativeDirectory, name] }
         filter: { ext: { eq: ".mdx" }, relativeDirectory: { ne: "not-in-sidebar" } }
```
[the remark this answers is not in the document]
scheul93: I think it would be valuable, but I'm not sure it is possible. Because we are using a component query and the `useStaticQuery` hook, no variables are able to be used inside the static query

packages/docs/src/components/DocSiteNavigation.tsx
scheul93: I added some more specific type checking in the new functions

packages/docs/src/components/DocSiteNavigation.tsx
scheul93: Done! Moved in a utils file

packages/docs/src/components/DocSiteNavigation.tsx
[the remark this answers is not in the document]
scheul93: Done. I tried to update the docs and types a bit to be more descriptive

packages/docs/src/helpers/navDataFormatUtils.ts
```diff
@@ -0,0 +1,132 @@
+import sortBy from 'lodash.sortby';
```
scheul93, line 1: Moved all the data restructuring to this file with some updated typing and docs.

packages/docs/src/components/DocSiteNavigation.tsx
```diff
@@ -112,69 +86,10 @@ const DocSiteNavigation = ({ location }: DocSiteNavProps) => {
...
-          retVal.push(formatNavItemData(navItemLvl1.node));
-        });
-      }
-    });
-
-    return retVal;
-  };
-
-  const navItems: VerticalNavItemProps[] = formatNavData(data?.allFile?.group);
+  const navItems: VerticalNavItemProps[] = useMemo(() => {
```
scheul93, line 89: Wanted to call out this fun use of `useMemo`. I did some logging and it seems like the data restructuring was happening on certain re-renders, so by using this memoization, the data restructuring only happens if the data or the window location data changes

pwolfert: Nice! 🙌

pwolfert (approved):
Looks great!

PR #1948, opened 2022-07-07
[WNMGDS-1721] Storybook Consolidation with Theme Capabilities for SB & Gatsby

forrestbaer:
Initial PR, don't merge until sprint 3.3. Will be adding loki updates.

## Summary

- Consolidates Storybook into a single themable instance! Still buildable separately. There will be a future ticket in 3.3 added to this to show/hide extraneous stories based on environment variable, this would help with loki testing issues
- A few small style fixes to stories
- Adds theme switching capability to both storybook & gatsby via query param, localStorage & environment variables
- Adds medicare fonts to gatsby / storybook

## Questions

- Should `localStorage` be swapped out for `sessionStorage`?

## How to test

- Make sure you're on node 16.
- `yarn install && yarn build-storybook` `http-server storybook-static` to check out a static build, can also `yarn build-storybook:healthcare` etc.. to see the static build of the other sites that way, check out the theme switching option.. when you build a child system you should see that systems theme selected by default, look at a component like monthpicker and note that the inverse background is set correctly
- `yarn storybook` will give you the same thing as above
- `yarn build-storybook:gatsby && yarn build:gatsby && yarn start:gatsby` - add `?theme=medicare` `?theme=healthcare` to the url when loaded, note localStorage item updates. check out pages with iframes and note a theme change updates the iframe content
-


packages/design-system-tokens/src/lib/exportCssVars.ts
```diff
@@ -9,6 +9,8 @@ import { writeFile } from './file';
  * Instead, these values will have a direct sass variable -> value definition
  */
 const variableIgnoreList: string[] = [
+  'font-sans',
```
forrestbaer, line 12: These were causing recursion loops

packages/design-system/src/components/MonthPicker/MonthPicker.stories.jsx
```diff
@@ -35,7 +35,7 @@ export default {
   },
 };
 
-const Template = ({ data, ...args }) => <MonthPicker {...args} />;
+const Template = (args) => <MonthPicker {...args} />;
```
forrestbaer, line 38: In storybook 6 args are passed first, there are some other places where this could be updated but I was in here messing with passing globals and just left it as it's more correct.

packages/docs/src/components/ComponentThemeOptions.tsx
```diff
@@ -17,12 +17,11 @@ const componentThemes = {
 };
 const masterThemes = { core: CoreTheme, healthcare: HealthcareTheme, medicare: MedicareTheme };
 
-type ThemeNames = 'core' | 'healthcare' | 'medicare';
```
forrestbaer, line 20: In an effort to allow some flexibility for adding future themes without needing to update a bunch of files I just changed this to string as generally references to this type are undefined until runtime.

packages/ds-healthcare-gov/src/components/Accordion/Accordion.stories.jsx
```diff
@@ -4,14 +4,21 @@ import AccordionItem from './AccordionItem';
 import { useArgs } from '@storybook/client-api';
 
 export default {
-  title: 'Components/Accordion',
+  title: 'Healthcare/Accordion',
   component: Accordion,
   argTypes: {},
   args: {
-    bordered: true,
```
forrestbaer, line 11: A bug, healthcare accordion doesn't use bordered.

scheul93:
When I did the css token variable work, I updated the token dist command to export all those map and css files to the .storybook directory since I wasn't sure what the plan was for theme switching.

I see that you removed those unused files 🎉 , can you update the `copyThemes.sh` file to remove the copying to the storybook directory. ([This line](https://github.com/CMSgov/design-system/blob/master/packages/design-system-tokens/src/copy_themes.sh#L22) & [this line](https://github.com/CMSgov/design-system/blob/master/packages/design-system-tokens/src/copy_themes.sh#L26))

packages/docs/src/styles/theme-variables/_medicare-scss-to-css.map.scss
```diff
@@ -130,7 +130,7 @@ $color-warn-light: var(--color-warn-light);
 $color-warn-lighter: var(--color-warn-lighter);
 $color-warn-lightest: var(--color-warn-lightest);
 $color-visited: var(--color-visited);
-$font-sans: var(--font-sans);
+$font-sans: 'Rubik', sans-serif;
```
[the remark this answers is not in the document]
forrestbaer: Updated `copyThemes.sh`! And yeah you're right, setting that back to what it originally was fixes that, I was wrong about the loop I thought that was creating.. going to fix that and remove the ignore for now and test.

scheul93 (approved):
Looks good after the latest updates to fonts. Really great work with the consolidation

PR #1954, opened 2022-07-11
BUGFIX: Calc issue with multiplication in old sass

forrestbaer:
## Summary

- see: https://adhoc.slack.com/archives/C645GLLNN/p1657538365283359


packages/design-system-scripts/gulp/sass.js
```diff
@@ -9,7 +9,7 @@ const gulpIf = require('gulp-if');
 const path = require('path');
 const postcss = require('gulp-postcss');
 const postcssImport = require('postcss-import');
-const sass = require('gulp-sass')(require('sass'));
+const sass = require('gulp-sass');
```
pwolfert, line 12: I think we can avoid the gulp-sass downgrade if we supply node-sass here with `require('gulp-sass')(require('node-sass'))`. I remember I had done that on a branch that got merged with yours. I think I might have had to upgrade because of node v16.

forrestbaer: I'm getting strange issues when building with any configuration here, need to look into it further, only finding 4 sass files when building :no\_good:

pwolfert (approved):

PR #1955, opened 2022-07-11
[NO-TICKET] Fix scan scripts after node 16 upgrade

pwolfert:
## Summary

On master, our `yarn scan-dependents` script was broken. This fixes it. I wasn't able to get our root tsconfig working with our scripts without messing it up for our ds packages, so I just made a new tsconfig.


scripts/tsconfig.json
```diff
@@ -0,0 +1,4 @@
+{
+  // This is an alias to @tsconfig/node16: https://github.com/tsconfig/bases
+  "extends": "ts-node/node16/tsconfig.json"
```
forrestbaer, line 3: Interesting didn't know about this.

forrestbaer (approved):
Working great for me now with node 16

scripts/tsconfig.json
pwolfert: Me neither. I didn't really want to have two separate configs, but it was taking too long to troubleshoot the other one

pwolfert: At least with this extends feature we're not actually _maintaining_ a separate set of configs

PR #1917, opened 2022-06-22
[WNMGDS-1687] NewRelic monitoring for doc site

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1687

### Added

- Added NewRelic monitoring agent in Gatsby site

### Changed

- Changed node version from `14.17` to `16.15.1`, which `gatsby-plugin-newrelic` wanted
  - Which also meant we needed to upgrade `gulp-sass` because the version of `node-sass` that `gulp-sass` required was outdated and didn't support node 16. Even though our scripts package was installing the newer version of `node-sass` (v7), that gulp plugin wanted an older one.

- [x] Need to update our Jenkins scripts

## How to test

1. `yarn build-storybook:gatsby && yarn build:gatsby && yarn serve:gatsby`
2. Log into NewRelic and view our browser application. You should see your traffic being reported

Screen Shot 2022-06-21 at 7 50 06 PM

PR #1957, opened 2022-07-11
[NO-TICKET] Revert node 16 upgrade

pwolfert:
## Summary

In order for the [current beta release](https://github.com/CMSgov/design-system/releases/tag/%40cmsgov%2Fdesign-system%403.7.0-beta.1) to not break downstream builds for teams that still build their Sass with the `node-sass` package, we need to roll back our Sass updates that fixed deprecated syntax. The compiler that ships with `node-sass` cannot interpret the newer syntax, and we cannot use the old syntax with the newer compiler that ships with the `sass` package. Therefore we have to downgrade `sass` to `node-sass`, revert Sass syntax updates, and downgrade Node 16 to 14 because 16 is incompatible with `node-sass`...AND remove a library that required us upgrading to node 16 in the first place. After making these changes, we will patch the current beta release.

### Changes

- Reverts https://github.com/CMSgov/design-system/pull/1917, which 
  - Reverts node version change so that version 16 is back down to 14
  - Reverts new feature of NewRelic monitoring for doc site
- Partially reverts https://github.com/CMSgov/design-system/pull/1930 by removing syntax that is incompatible with the older `node-sass` compiler

- [x] Downgrade version used in Jenkins scripts


pwolfert:
I think I found an alternative to all this here https://github.com/CMSgov/design-system/pull/1958

PR #1958, opened 2022-07-11
[NO-TICKET] Make Sass syntax compatible with node-sass compiler

pwolfert:
## Summary

This is an alternative to https://github.com/CMSgov/design-system/pull/1957. I've changed the parts of our Sass code that were incompatible with `node-sass` (the old one) while still avoiding division-syntax deprecation warnings in dart `sass` (the new one).

### Fixed

- Fixes incompatibility of our Sass code with downstream projects that are still using `node-sass`. Everywhere that performed division using the `math.div` function has been replaced with `calc()`, which [according to the Sass documentation](https://sass-lang.com/documentation/breaking-changes/slash-div) might actually be evaluated during compile time when all the information is available.


forrestbaer (approved):
This seems to work! Nice job!

PR #1961, opened 2022-07-12
[BUGFIX] Restore a reference to an old sass variable

pwolfert:
### Fixed

- Fixes backward compatibility for token update


forrestbaer (approved):
Good find!

PR #1965, opened 2022-07-14
[WNMGDS-1719] theme switcher

scheul93:
## Summary

Adding theme switcher dropdown to the doc site. When the dropdown value changes, the page reloads with the new theme applied.

Additionally, building off Forrest's previous work, if a theme value is saved in session storage, but no theme query param is present, the query param is added to the url (without reloading the page) to keep the two in sync.

## How to test

`yarn start:gatsby` or [demo site](https://cmsgov.github.io/design-system/branch/WNMGDS-1719/theme-switcher/)

You should see the site change if you change the dropdown. You should also see the url query param update as the drop down updates. If you leave the page and come back, the site should maintain your previous theme selection


forrestbaer (approved):
This is great! Refactors look good. 💯

pwolfert (changes requested):
This is really cool! I did just notice an issue on the demo site. Not sure if anyone else can recreate it. I've included a screen capture of what's happening below. The dropdown value doesn't seem to be in sync.

<details>
  <summary>Screen capture</summary>
  
select-out-of-sync
</details>

forrestbaer:
> This is really cool! I did just notice an issue on the demo site. Not sure if anyone else can recreate it. I've included a screen capture of what's happening below. The dropdown value doesn't seem to be in sync.
> 
> Screen capture

Oh interesting I didn't notice this happening locally.

scheul93:
I'll take a look at what's happening

scheul93:
@pwolfert I think I fixed it and updated the demo site.

For the curious, it seems that on building for production, the `getTheme` function was running before the window object was available, so 'core' was always being returned as the theme for the default of the dropdown. I converted `getTheme` to a custom hook `useTheme` and wrapped the functionality in a `useEffect` which runs once in the browser where window should be available. I believe this fixes the issues while keeping the intended functionality.

pwolfert (approved):
Can confirm that it works now 👍. Nice fix!

PR #1966, opened 2022-07-14
[NO-TICKET] Refactor scan scripts and add arbitrary dependency-scanning feature

pwolfert:
## Summary

In order to more thoroughly research the effects of certain changes we make on our dependent projects downstream, I'm expanding the scan scripts a bit. This isn't all of what I want to accomplish feature-wise, but it's a first step in refactoring what we have there so multiple scan strategies can be used with the same output functions so we aren't duplicating code as much. Duplicating code was fine—and even preferred—at first when we didn't have much of a plan for these scripts.

### Added

- Added the ability to scan for aribtrary dependencies of our dependents with `yarn scan-dependents --dependency <dependency-package-name>`, which replaces `yarn scan-dependents --react`
- Added the ability to restrict the search to a subset of design systems

### Changed

- Refactored scan scripts to pull out common patterns into shareable functions

## How to test

There are some instructions in the readme in the `scripts/scanDependents` folder


forrestbaer (approved):
This works great, so cool, and export as csv in the future could be really useful if we check our main dependencies against some of theirs during some part of build or upgrade, etc.. 🥇

PR #1968, opened 2022-07-14
[NO-TICKET] Adding 2 white and 2 black alpha tokens to theme files

line47:
Adding 2 white and 2 black alpha tokens to theme files to utilize in designs in Sketch


forrestbaer (commented):
Just need to remove the added files to `.storybook` and if you could format the `.ts` files so they are in a column layout like the rest of the file that would make this one ready.

forrestbaer (approved):
👍🏻 FYI in the future will also have to update loki snapshots but that's in flux right now so we can cross that bridge when everything is done and merged.

line47:
@forrestbaer I tried to update the formatting of the `.ts` files - let me know if I didn't get it

PR #1971, opened 2022-07-14
[NO-TICKET] Introduce `scan-dependents` sub-commands and implement code search

pwolfert:
## Summary

This adds a wishlist item from https://github.com/CMSgov/design-system/pull/1966 and reorganizes the `scan-dependents` command to have sub-commands.

### Changed

- New commands:
  - `yarn scan-dependents versions` - was previously `yarn scan-dependents`
  - `yarn scan-dependents dependency <dependencyName>` - was previously `yarn scan-dependents --dependency <dependencyName>`
  - `yarn scan-dependents search <searchString>` - a new feature
  - Run `yarn scan-dependents --help` to see commands and options

## How to test

See the readme and `yarn scan-dependents --help`. Try some common formulas in the readme. I can't test anymore because I ran it _a lot_ and have reached my API rate limit 😂 

Something that might be nice in the future if we use this a lot is local caches of the repo lists (result of the `getDependentPackageData` function) saved by day in `scripts/scanDependents/cache` so we can skip 1-4 API calls for every search executed


scripts/scanDependents/README.md
```diff
@@ -32,32 +32,26 @@ Pass it to the command like this
...
-yarn scan-dependents --dependency <dependency-package-name>
+yarn scan-dependents --dependency react
 ```
 
-For example, this would scan for which version of React our dependencies are using:
+### Scan only a subset of our design systems
 
 ```
-yarn scan-dependents --dependency react
+yarn scan-dependents --designSystems @cmsgov/ds-medicare-gov @cmsgov/ds-healthcare-gov --dependency node-sass
```
forrestbaer, line 46: These docs don't work for me now that `--dependency` has changed to just `dependency..`

this works:

`yarn scan-dependents dependency node-sass --designSystems @cmsgov/ds-medicare-gov @cmsgov/ds-healthcare-gov`

scripts/scanDependents/README.md
```diff
@@ -32,32 +32,26 @@ Pass it to the command like this
...
-
-### Scanning only a subset of our design systems
+This will scan only dependents of the medicare and healthcare design systems to find out what version of `node-sass` they're using if any.
 
-Example:
+### Search JavaScript files for certain strings
 
 ```
-yarn scan-dependents --designSystems @cmsgov/design-system @cmsgov/ds-healthcare-gov --dependency node-sass
+yarn scan-dependents search ChoiceList --extensions js jsx tsx
```
forrestbaer, line 54: It doesn't look like extensions was a part of this change, I don't see it in the available options in index, though it is presented in the search function as an optional prop, something to add in the future?

scripts/scanDependents/scans.ts
```diff
@@ -84,3 +106,99 @@ export async function scanDependentDependencyVersions(
...
+
+  const repositoryResults: { [repoName: string]: number } = {};
+  let page = 1;
+  const perPage = 50;
+  let itemsProcessed = 0;
+  let searchResponse;
+  do {
+    searchResponse = await apiRequest('GET /search/code', { q, page, per_page: perPage });
+    for (const item of searchResponse.data.items) {
+      const repo = item.repository.full_name;
```
forrestbaer, line 147: I don't know what this would look like but would be nice to get the URL's back from the request so we could click around to the references and maybe suggest changes in those cases.. Not in-scope for this one but maybe a future addition!

forrestbaer (commented):
Looks and works great, just a few comments about the docs.

scripts/scanDependents/README.md
pwolfert: Is it listed when you run it with the help flag? Should be coming from https://github.com/CMSgov/design-system/pull/1971/files#diff-2aca4f9afa500c949a5d976373ed3303a8c6b2ec6bc12ad271b11239f7c35012R72

forrestbaer: Hmm no it doesn't show up with `--help` for me, though running `yarn scan-dependents search ChoiceList --extensions js jsx tsx` did work for me this time around.

forrestbaer (approved):
💯

PR #1972, opened 2022-07-18
[WNMGDS-1782] Storybook/Gatsby Consolidation Loki Updates

ghost:
## Summary

- Loki snapshot updates with most recent color changes
- Loki runs correctly now with `yarn loki:system test` for each system
- Gatsby `.utility-example` rendering correct colors for each system now
- Storybook loading correct child system files now for storybook static and dynamic rendering
- Removed `<div data-theme=theme>` wrapper from element and moved `data-theme` to `html` tag of example iframe to fix rendering of `.ds-base` which has been moved to the `body` tag.
- Moved deprecated theme styles for medicare to the bottom of the declaration to resolve issue with wrong colors being picked up by `ComponentThemeOptions`
- Several small fixes to color names to use theme sass values instead of to-be deprecated values
- Updated `loki.sh` to use `STORYBOOK_DS` environment variable
- Fixed a small bug with `ResponsiveExample` that would sometimes throw an error when value wasn't avaliable
- Child system specific stories only shown when their storybook is rendered, or when using build-storybook:gatsby so they are all available for gatsby.. Gatsby uses a different method for picking up static assets (in it's webpack config), storybook instances utilize the static dirs defined in `.storybook/main.js`.. This also allows for loki testing without including extraneous stories for a theme.
- Default theme set to child system when their respective storybook is started/built

## **Note**

- The Medicare Navbar seems off, I messed with the calc for it's top position a bit but I would consider that a bug that I didn't introduce with this PR, calc doesn't need to interpolate the variable here so `-($font-size-base + 21px + 6px)` would work but then puts the button up pretty high.. And still doesn't look good on mobile either way.. I don't know if this is an old bug or what.. Also could just make that 27px 🚀  
- I didn't do work here to hide the default `Accordion` for healthcare as I think that's being handled in another ticket?

## How to test

`yarn install && yarn build`
`yarn loki test, yarn loki:healthcare test, yarn loki:medicare test`
`yarn storybook, etc.. yarn build-storybook, etc..`
`yarn build-storybook:gatsby && yarn build:gatsby && yarn start:gatsby + serve:gatsby`

### DIFF
--- a/packages/docs/content/components/accordion.mdx
+++ b/packages/docs/content/components/accordion.mdx
@@ -10,11 +10,21 @@ An accordion is a list of headers that hide or reveal additional content when se
 
 ## `<Accordion>`
 
-<StorybookExample
-  componentName="accordion"
-  sourceFilePath="components/Accordion/Accordion.tsx"
-  storyId="components-accordion--default"
-/>
+<ThemeContent neverThemes={['healthcare']}>
+  <StorybookExample
+    componentName="accordion"
+    sourceFilePath="components/Accordion/Accordion.tsx"
+    storyId="components-accordion--default"
+  />
+</ThemeContent>
+<ThemeContent onlyThemes={['healthcare']}>
+  <StorybookExample
+    componentName="accordion"
+    sourceFilePath="components/Accordion/Accordion.tsx"
+    sourcePackageName="healthcare"
+    storyId="healthcare-accordion--default"
+  />
+</ThemeContent>
 
 ### Props
 
--- a/packages/docs/src/components/StorybookExample.tsx
+++ b/packages/docs/src/components/StorybookExample.tsx
@@ -24,6 +24,10 @@ interface StorybookExampleProps {
    */
   sourceFilePath?: string;
   /**
+   * package where the source comes from
+   */
+  sourcePackageName?: string;
+  /**
    * Current theme
    */
   theme: string;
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -34,17 +34,17 @@
...
     "prettier": "npx prettier --write 'packages/**/*.{js,jsx,tsx,scss,html,md,mdx,json}'",
-    "loki": "./loki.sh",
-    "loki:healthcare": "./loki.sh -t healthcare",
-    "loki:medicare": "./loki.sh -t medicare",
+    "loki": "STORYBOOK_DS=core ./loki.sh",
+    "loki:healthcare": "STORYBOOK_DS=healthcare ./loki.sh",
+    "loki:medicare": "STORYBOOK_DS=medicare ./loki.sh",
     "prepare": "husky install",
     "storybook": "STORYBOOK_DS=core start-storybook -p 6006",
     "storybook:healthcare": "STORYBOOK_DS=healthcare start-storybook -p 6006",
```
scheul93, line 42: [question] Not directly related to your changes, but do we still need to support the `storybook:healthcare`, `storybook:medicare`, `build-storybook:healthcare` & `build-storybook:medicare` commands?

forrestbaer: I know we at least need the `storybook:healthcare/medicare` for loki to work appropriately, I don't know that we need to support the build commands though, good question.

PR #1974, opened 2022-07-18
[NO-TICKET] updated mobile component theme options table

scheul93:
## Summary

I noticed that on mobile, the table for the sass variable "customizations" was overflowing and causing the whole page to extend
Before 1:
Screen Shot 2022-07-18 at 10 28 01 AM

Before 2:
Screen Shot 2022-07-18 at 10 28 21 AM

## How to test

Updated to use our table component so that we can take advantage of the stacking

After:
Screen Shot 2022-07-18 at 10 27 28 AM


forrestbaer (approved):
Very nice 💅🏻

PR #1977, opened 2022-07-19
[WNMGDS-1722] component page header

scheul93:
## Summary

Updating the page header to support new designs for component pages.

This PR contains:
- Moving Github logo to separate component for reusability
- Making utility functions to help with github, sketch and storybook links
- Removing 'related USWDS guidance' from content
- Updating the Tooltip page's content with proposed structure for these link values

## TODO

- [X] Remove 'View Source' component since that link is moved to the page header
- [X] Update content pages with appropriate links for sketch, storybook and github
- [X] Update content pages with intro moved into frontmatter

## How to test
[Demo site](https://cmsgov.github.io/design-system/branch/WNMGDS-1722/component-content-update/)

`yarn start:gatsby`
Screen Shot 2022-07-19 at 3 59 50 PM


packages/docs/content/components/tooltip.mdx
```diff
@@ -1,12 +1,23 @@
...
+  sketchLink: B8CDD372-C06C-4C5F-9D08-3E78B5AD6CC2
+  storybookLink: components-tooltip--icon-trigger
+healthcare:
+  githubLink: design-system/src/components/Tooltip
+  sketchLink: 6B9E0E70-93E2-4068-B7A5-317FADC175AE
+  storybookLink: components-tooltip--icon-trigger
+medicare:
+  githubLink: design-system/src/components/Tooltip
+  sketchLink: 9D306E81-ED90-4FA3-B298-D7D44F4405AF
+  storybookLink: components-tooltip--icon-trigger
```
scheul93, line 16: What do you think of this structure. In yaml, it is equivalent to an object of key/value pairs.

The idea is that every component page would have the three 'core' values defined + the sketch links for healthcare and medicare. There are a few components that are brand-specific, so they may need different github or storybook urls than core.

Similar to other places in our system, it's a bit like a cascade, where if a medicare storybook link is not defined, the core one will be used.

Thoughts?

zarahzachz: I was going to ask about a cascade option - since the GH and SB links are the same for all three it looked like a value was required

zarahzachz (commented):
We might need a mobile friendly solution for these links. I checked in in iPhone SE and there is some funny wrapping that happens that splits the icon from the link. At the very least, maybe the wrapping is OK but the icon and link should always stay together?

Screen Shot 2022-07-20 at 3 38 06 PM

packages/docs/content/components/tooltip.mdx
scheul93: I did that for this PR, but it is confusing. I'll update.
I also talked to Scott and sketch links will not cascade, but I think the other two can (for now)

forrestbaer (approved):
Looks good to me 👍🏻

zarahzachz (approved):

pwolfert (approved):

PR #1981, opened 2022-07-20
[WNMGDS-1795] Add aria-current to active link in VerticalNav

### LINKED ISSUE #1979: Vertical Navigation: selected link within the list item needs aria-current attribute added
**Describe the bug**
The [MDCT Team alerted the team](https://cmsgov.slack.com/archives/CHH0381RD/p1658248622172569) that the [vertical navigation component](https://design.cms.gov/components/vertical-navigation) would be improved by adding the [aria-current](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current) attribute to the currently selected link.

**To Reproduce**
1. Go to [CMS DS Vertical Navigation component](https://design.cms.gov/components/vertical-navigation/)
2. Open browser dev tools
3. Inspect the currently selected list item
4. You will notice that there is no aria-current attribute applied

**Expected behavior**
As a screen reader user I would expect to be able to hear which item from the list is currently selected. 

**Desktop**
 - OS: Mac
 - Browser: Chrome
 - Version: 103.0.5060.114

**Additional context**
Info regarding aria-current:
- The aria-current attribute indicates the element on which it is set
- Only mark one element in a set of elements as current with aria-current.
- The aria-current attribute accepts a limited list of [values](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-current#values) including page, step, location, date, time, true, and false. I think to avoid confusion we should limit this to "true" as the given value. I'm open to discussion on this but hoping to help developers avoid having to decide this on their own.

The output html should look something like this with aria-current added to the `<a>`:
`<li class="ds-c-vertical-nav__item"><a class="ds-c-vertical-nav__label ds-c-vertical-nav__label--current special-link" href="javascript:void(0);" aria-current="true">Grandchild link</a></li>`

zarahzachz:
Fix #1979

## Test
Visit [demo site](https://cmsgov.github.io/design-system/branch/WNMGDS-1795/add-ariaCurrent-to-VerticalNav/storybook/?path=/story/components-vertical-nav--default-vertical-nav), open dev tools and "Grandchild link" should have `aria-current="true"` set.

### DIFF
--- a/packages/design-system/src/components/VerticalNav/VerticalNavItemLabel.tsx
+++ b/packages/design-system/src/components/VerticalNav/VerticalNavItemLabel.tsx
@@ -66,6 +66,7 @@ export const VerticalNavItemLabel = (props: VerticalNavItemLabelProps): React.Re
   } else if (LabelComponent !== DEFAULT_COMPONENT_TYPE) {
     // Apply href if <a> or custom component type
     otherProps = {
+      'aria-current': props.selected,
       href: props.url,
     };
   }


scheul93 (approved):
LGTM

PR #1985, opened 2022-07-21
[WNMGDS-1722] component content update

scheul93:
## Summary

* Removing 'View Source Link' component from docs site since link to Github will now be in the page header
* Updating links to sketch, github & storybook for all components
* Moving intro sentence into fronmatter for components
* Also found an instance in Mask scss where the value was not being calculated for css vars, so updated that for display on the doc site

## How to test

[Demo site](https://cmsgov.github.io/design-system/branch/WNMGDS-1722/component-content-update)


forrestbaer (approved):
Nice! This looks great 👍🏻

packages/docs/content/components/accordion.mdx
```diff
@@ -1,26 +1,31 @@
 ---
 title: Accordion
 status: draft
+intro: An accordion is a list of headers that hide or reveal additional content when selected.
+core:
+  githubLink: design-system/src/components/Accordion/Accordion.tsx
```
zarahzachz, line 6: Curious - should we link to the tsx file or the directory of the component? Or does that even matter?

scheul93: I could go either way. I think for some, I linked directly to a file and others I linked to the directory. I chose files when the whole directory felt like it had an overwhelming amount of files. But I'm happy to update to whatever folks like

scheul93: Updating to link to the directory for v1.

pwolfert (approved):
I ran through a fair number of these pages and checked all the links in the different themes, and I didn't see any mistakes 👍

PR #1986, opened 2022-07-21
[NO-TICKET] Update Loki to fix problems on M1 Macs

pwolfert:
## Summary

Our Loki tests were failing on Macs with M1 chips. [This issue](https://github.com/oblador/loki/issues/344) tracks the issue and resolution, which was for the Loki team to update the Docker image they were using. I've updated Loki to get that fix, but it also means most, if not all, reference images have changed due to trivial differences.

### Fixed

- Fixed M1 support for our Loki tests

## How to test

`yarn loki test`, `yarn loki:healthcare test`, `yarn loki:medicare test`


forrestbaer:
Wondering what it changed that required all the reference images to be updated? Since none of those diffs will load in the browser (because there are too many, wish github would add pagination) can you confirm the new reference images are correct?

zarahzachz:
I ran it locally and it looks like it's showing all our components as 1px smaller? Super weird.

I wonder if Loki has a threshold we can set, or if it was adjusted during this fix? I kinda don't think a 1px difference is worth pointing out

pwolfert:
If the image itself is larger, I don't know if the threshold will matter, but it might. I can try with 1% and see if it catches things. Forrest, they all looked identical to me; the components themselves did not change.

pwolfert:
@zarahzachz, I tried up to 50% with `yarn loki test --chromeTolerance 50`, but they still failed. I think the changed pixels still have to fit within the dimensions of the reference image.

forrestbaer (approved):
LGTM 👍🏻

PR #1987, opened 2022-07-22
Release 3.5.3

zarahzachz:
@cmsgov/design-system-docs@3.5.3
@cmsgov/design-system-scripts@3.5.3
@cmsgov/design-system@3.5.3
@cmsgov/ds-healthcare-gov@7.5.3
@cmsgov/ds-medicare-gov@5.5.3


forrestbaer:
I think we want to cherry pick the fix onto the @cmsgov/design-system@3.5.3 tag, and release from there, don't see any changes.

pwolfert (approved):
Looked at the release tag and the fix commit and bump commits, and they look good to me 👍

forrestbaer:
Another benefit of release branches, easier to see what's being changed for a patch on an older version! :)

PR #1989, opened 2022-07-22
[NO-TICKET] - Lowercase font filenames to optimize deploys to Akamai NetStorage.

LouisFettet:
## Summary

This has been something I've been wanting to look into for a while. Long story short, when we deploy any file to Akamai NetStorage, its name will be transformed to all lowercase. This has never caused any issues on the clientside, since any requests for a file with incorrect casing are resolved by Akamai. However, this discrepancy causes some efficiency problems with our deploys, since `rsync` doesn't recognize that the files are the same, which means that during every deploy we are deleting all of the DS font files in NetStorage and reuploading them from scratch. This only takes about 10 seconds, but since it impacts almost every application, it does add up.

### Changed 

* Renamed all DS font files under `packages/design-system/src/fonts/` so that they are all lowercase.
* Made minor edits to the font stylesheet, `packages/design-system/src/styles/_font.scss`, to reflect the above changes.

## How to test

To ensure that this didn't cause any issues in downstream applications, I tested this change in both [Learn](https://github.cms.gov/CMS-WDS/learn-frontend-gatsby) and [FLH](https://github.cms.gov/CMS-WDS/flh-consumer). It would be wise to test this out for apps that use `ds-medicare-gov` as well (although I admit I'm not familiar enough with them).

These are the rough steps I took to run this in FLH locally (Learn is similar but has a dependency on a running Drupal instance which makes it difficult to spin up):
```bash
# from this project's root, /design-system/
# build the parent and create a symlink to the dependency
yarn build
yarn link
# make sure that the child is using the changes made in the parent
cd packages/ds-healthcare-gov
yarn link "@cmsgov/design-system"
# build the child
cd ../..
yarn build:healthcare
# create another symlink
cd packages/ds-healthcare-gov
yarn link
# navigate to downstream application, removing all installed dependencies (FLH uses npm instead of yarn)
cd /path/to/flh-consumer
rm -rf node_modules
# link to local upstream packages and install
yarn link "@cmsgov/design-system"
yarn link "@cmsgov/ds-healthcare-gov"
yarn install
# if everything went well you should be able to build the app
yarn run build-production
# check that fonts are all lowercased in the build
ls dist/fonts
```
I did run into some issues regarding hooks when trying to run FLH locally, but I'm relatively certain that this has nothing to do with the font filename changes. Learn built and ran just fine. 

There's likely a better way to accomplish the above; it would probably be good to add some information like this to either the `README.md` or `CONTRIBUTING.md` in the future.


forrestbaer:
This is a little confusing to me, is there no way this can be resolved via configuration on the Akamai side of things? I don't think that we want to convert these to all lowercase on our side as references to these files exist in our storybook and doc site applications as well, those would also need to be validated. 

I can't find any references on the web to Akamai requiring lowercase filenames. Could you provide more context?

pwolfert (approved):

PR #1990, opened 2022-07-22
[NO-TICKET] Swap Medicare's main and alternate ghost button color

pwolfert:
## Summary

I was updating `button-architecture` with `master` and making sure the loki reference images were up-to-date, and I noticed that the _main_ version of the ghost button was green instead of blue. They were flipped. [Here is a link to the Sketch mockups.](https://www.sketch.com/s/d7543fe5-171a-4302-a19c-85ea9fc438f8/a/VrO0anr)

### Fixed

- Fixed Medicare's _main ghost_ button being green instead of blue. Also reverted it to the blue used for links, which is what was being used in the old medicare transparent buttons.

## How to test

1. `yarn build && yarn storybook:medicare`
2. Visit http://localhost:6006/?path=/story/components-button--default-button&args=variation:ghost


forrestbaer (approved):
Good catch, looks good 👍🏻

PR #1992, opened 2022-07-25
[WNMGDS-1781] adding hover, active, focus styling to tooltip trigger

scheul93:
## Summary

Pre-storybook, our tooltip trigger examples had states for focus, active and hover. It's because our previous examples were using the `<Tooltip>` component's prop `activeClassName`. Instead of updating our examples, we can use CSS pseudo-selectors to add this styling instead of a whole class. We still need to support the `activeClassName` for backwards compatibility.

In this PR, I:
- Added a story for inverse tooltips to check this change
- Added selectors so that the active styling of tooltip triggers shows up
- Updated the logic for the `activeClassName` css class. Previously, if that optional prop was not provided, the tooltip trigger would get `undefined` added to its list of classes. Now, that value is just ignored.

## How to test

`yarn start-storybook` to test changes in Tooltip stories

Before:
Screen Shot 2022-07-25 at 11 35 35 AM

After (on light & on dark):
Screen Shot 2022-07-25 at 11 35 00 AM

Screen Shot 2022-07-25 at 11 34 44 AM

### DIFF
--- a/packages/design-system/src/components/Tooltip/Tooltip.stories.jsx
+++ b/packages/design-system/src/components/Tooltip/Tooltip.stories.jsx
@@ -95,3 +95,16 @@ TooltipWithCloseButton.args = {
   showCloseButton: true,
   className: 'ds-c-button',
 };
+
+export const InversedTrigger = Template.bind({});
+InversedTrigger.parameters = {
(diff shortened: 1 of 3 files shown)


pwolfert (approved):
Woot! Looks great!

PR #1995, opened 2022-07-25
[NO-TICKET] Update release script to bump exact

pwolfert:
## Summary

We recently patched an old version of the design system, but when that patch for an older version of a child design system was installed in a project, the core version resolved to the latest because npm/yarn deemed the versions compatible because of the caret `^` in our child design systems' package files. This is something we manually fixed once, but Lerna re-added the carets.

### Changed

- Update release script to use exact versions of our design system packages during lerna version bump

## How to test

You can use the release script if you want and then run `yarn release -u` to undo


forrestbaer (approved):
Looks good!

PR #1999, opened 2022-07-26
[NO-TICKET] Table of Contents update

scheul93:
## Summary

Updating the Table of Contents for the doc site to include h2 & h3 elements. Includes some updated styling.

Additionally, I removed the custom heading components we were using in `ContentRenderer` that were adding ids to headings so that we could link them. Instead, I am adding a [gatsby plugin](https://github.com/gatsbyjs/gatsby/tree/master/packages/gatsby-remark-autolink-headers) that adds these ids. This plugin is necessary because it uses 'remark' under the hood to generate these ids so that they will be in sync with the url being returned by graphql for the TOC data.

## How to test
Since I added a package, you may need to `yarn install` first
`yarn start:gatsby`


pwolfert (approved):
Thank you for updating this! The table of contents is much more informated with the extra level.

I noticed some strange behavior that we might want to investigate: When I click one of the links in the table of contents, it seems to re-render the page, which causes layout shifting because of the loading iframe examples. Try it out on the Alert page to see what I mean.

scheul93:
It might have to do with the TOC links not preserving the theme query param. Looking into this...

PR #2001, opened 2022-07-27
[WNMGDS-1805] Deprecate btn state classes

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1805)

Removed CSS classes styling button states:
- `ds-c-button--focus`
- `ds-c-button--active`
- `ds-c-button--disabled`
- `.ds-c-button--hover`

Buttons (and links styled as buttons) have pseudo selectors that do this (and are already styled to do so). These classes are irrelevant.

I ran unit tests and tested buttons locally and saw now change in behavior. I searched our `/packages` dir and found no reference to these classes in place.

[Storybook demo](https://cmsgov.github.io/design-system/branch/WNMGDS-1805/deprecate-btn-state-classes/storybook/?path=/story/components-button--all-anchor-buttons&globals=theme:core)

[Hgov storybook demo (for @line47)](https://cmsgov.github.io/design-system/branch/WNMGDS-1805/deprecate-btn-state-classes/storybook/?path=/story/components-button--all-anchor-buttons&globals=theme:healthcare)


zarahzachz:
I also got a few Loki updates and didn't see any visual differences even though it's marked as different.

forrestbaer (approved):
Oh yeah this is nice, fun game too, trying to find differences in the loki snapshots.. Reminded me of those magazines in the dentist office .. I couldn't find any.

scheul93 (approved):

packages/design-system-tokens/src/themes/healthcare-components.ts
```diff
@@ -57,7 +57,7 @@ export const components: AnyTokenValues = {
 
   'button': {
     '__background-color--active':                 t.color['primary-darkest'],
-    '__background-color--disabled':               'transparent',
+    '__background-color--disabled':               t.color['white'],
```
zarahzachz, line 60: @line47 Requested updated to disabled background. Change from `transparent` to `white` for all contexts (light/dark).

packages/design-system/src/components/Button/Button.stories.jsx
```diff
@@ -94,6 +94,297 @@ VariationsOnDark.decorators = [
   ),
 ];
 
+export const AllAnchorButtons = () => {
```
zarahzachz, line 97: Created new story to reflect how the new Button API looks/behaves when there's an `<a>` under the hood. Turned out it was very bad! CSS fixes further down.

packages/design-system/src/components/Button/Button.stories.jsx
```diff
@@ -94,6 +94,297 @@ VariationsOnDark.decorators = [
...
+          <Button href="#!" size="big">
+            Anchor button
+          </Button>
+          <Button href="#!">Anchor button</Button>
+          <Button href="#!" size="small">
+            Anchor button
+          </Button>
+        </div>
+        <div className="" style={{ display: 'flex', gap: '1rem', alignItems: 'center' }}>
+          <Button size="big" disabled>
```
zarahzachz, line 114: For a link to be truly disabled, **a user cannot add an `href`**. We don't call this out in our docs and I don't know if we need to as it's part of the anchor spec. 

The reason this is important to note is that while we can visually style our link buttons to look like buttons, when a user wants a disabled anchor button it won't be a 1:1 match behaviorally to a disabled button button (i.e., AT can still access disabled anchor buttons).

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -13,40 +13,46 @@
   font-weight: var(--fontWeight, $button__font-weight);
   line-height: 1.3;
   text-align: center;
-  text-decoration-color: transparent;
```
zarahzachz, line 16: I changed this CSS rule to target `text-decoration` instead and moved it to target `<a>` specifically.

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -13,40 +13,46 @@
...
   display: inline-block;
   padding: var(--space, $spacer-1 $spacer-3);
 
+  &:is(a) {
```
zarahzachz, line 19: This selector is [supported in all modern browsers](https://caniuse.com/css-matches-pseudo).

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -13,40 +13,46 @@
...
+    text-decoration: none;
+
+    &:focus {
```
zarahzachz, line 22: I needed to override some styles that are being applied directly to the anchor tag, so there's some duplicate CSS happening in this block. It makes me sad, but I had to do it.

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -13,40 +13,46 @@
...
+      background-color: var(--backgroundColor, $button__background-color);
+      text-decoration: none;
+    }
+
+    &.ds-c-button--disabled {
+      pointer-events: none; // Needed to render disabled links correctly
```
zarahzachz, line 28: This change made me happy. 

We were previously applying this rule to all `disabled` buttons just to prevent accidental interaction with anchor buttons. By moving this style to a more specific rule, I was able to apply a `not-allowed` cursor to our disabled button buttons, making for a more user-friendly experience.

packages/design-system/src/components/Button/Button.stories.jsx
zarahzachz: D'oh! I just realized when you exclude the `href` from something that was intended to be an anchor button, our component changes it into a button button - therefore the class we're using to style disabled anchor buttons is actually irrelevant. 

Do I remove `ds-c-button--disabled` from the codebase?

line47:
Checked out the HCgov anchor links and they look good :partying\_face:

packages/design-system/src/components/Button/Button.tsx
```diff
@@ -147,6 +145,12 @@ export const Button = ({
     // and remove <button> specific attributes
     delete attrs.disabled;
     delete attrs.type;
+
+    if (disabled) {
+      attrs.role = 'link';
+      attrs['aria-disabled'] = true;
+      delete attrs.href;
+    }
```
zarahzachz, line 153: [Discussed with @pwolfert and @SparrowSpit in Slack.](https://adhoc.slack.com/archives/C645GLLNN/p1659032212248209)

To summarize, we're adding better support for disabled anchor buttons. 

The way this works is users still pass an `href` prop to the `<Button>` to indicate if they want a `<button>` or an `<a>` element. However, if `disabled` is also passed (`<Button href="#!" disabled />`), it will strip the `href` attr from the anchor element and pass additional a11y-friendly attrs.

The disabled class logic was also removed as we're relying on `aria-disabled` to properly style our disabled link buttons.

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -13,49 +13,55 @@
...
   // Make sure active overrides focus styles
-  &:active:focus,
-  &.ds-c-button--active:focus {
+  &:active:focus {
     box-shadow: none;
   }
 
   &:disabled,
-  &.ds-c-button--disabled {
+  &[aria-disabled] {
```
zarahzachz, line 60: I like the idea of tying visual state changes to aria attrs as it gives devs a better visual queue when a11y isn't working quite right.

packages/design-system-tokens/src/themes/medicare-components.ts
```diff
@@ -120,13 +120,13 @@ export const components: AnyTokenValues = {
...
+    '__background-color--active':                 t.color['primary-light'],
     '__background-color--disabled':               t.color['gray-lighter'],
-    '__background-color--hover':                  t.color['primary-darker'],
+    '__background-color--hover':                  t.color['primary-dark'],
     '__background-color':                         t.color['primary'],
-    '__border-color--active':                     t.color['primary-darkest'],
+    '__border-color--active':                     t.color['primary-light'],
     '__border-color--disabled':                   t.color['gray-lighter'],
-    '__border-color--hover':                      t.color['primary-darker'],
+    '__border-color--hover':                      t.color['primary-dark'],
```
pwolfert, line 129: Why are these changes showing up here?

pwolfert: Okay, I just tracked it down to https://github.com/CMSgov/design-system/pull/2001/commits/ee1d6ca7086b7ad37ca1dec0cba214c5746b8df8. @line47, can we remove this and add it to a separate pull request? It doesn't seem related to this change.

pwolfert (approved):
I think we need to pull out the commit I identified below, but other than that, the change looks good to me!

PR #2002, opened 2022-07-27
[WNMGDS-1806] Remove default marker on code example for Safari

zarahzachz:
[Jira](https://jira.cms.gov/browse/WNMGDS-1806)

Code examples in Safari have the default marker icon. Adding Safari styles to remove.

[Demo.](https://cmsgov.github.io/design-system/branch/WNMGDS-1806/safari-details-bugfix/components/badge/?theme=core)


forrestbaer (approved):
Looked good to me, for some reason the url sent was hanging but don't think that was related to this, just navigated to another page and back to badge and it looked goot.

PR #2010, opened 2022-07-30
[WNMGDS-1803][WNMGDS-1802] Storybook-based examples can show React source

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1803
https://jira.cms.gov/browse/WNMGDS-1802

### Added

- Added `React code` button to Storybook-based examples on the doc site that will show its JSX source

### Changed

- Updated the design of the examples to match the new designs
- Updated how the code snippets are rendered so HTML and JSX snippets can coexist in the same space
- Replaced some `EmbeddedExample` instances with more useful `StorybookExample`s

### Fixed

- Fixed a minor CSS bug in the doc site where a rule was being applied beyond its original intention

## How to test

- `yarn start:gatsby` and take a look at examples on various pages


pwolfert:
Currently trying to figure out why gatsby won't build statically on my branch.

pwolfert:
I've found that if I comment out [this line that uses `useCodeSnippet` in the `EmbeddedExample`](https://github.com/CMSgov/design-system/blob/pwolfert/react-from-docs/packages/docs/src/components/EmbeddedExample.tsx#L18), it builds statically. It's only that one component using that hook that is causing a problem.

pwolfert:
Gave up trying to figure out why Gatsby couldn't statically build with my hooks version when a `ReactDOMServer.renderToString` was in play. See https://github.com/CMSgov/design-system/pull/2010/commits/29b5b27a00d07c80def079ad5a4d58533ef2d424 for debugging notes and a summary of the problem. See also https://adhoc.slack.com/archives/C645GLLNN/p1659375362412679

packages/docs/src/components/ExampleFooter.tsx
```diff
@@ -0,0 +1,95 @@
...
+}
+
+/**
+ * HTML code snippet. This component will prettify an html string and do syntax highlighting
+ */
+const ExampleFooter = ({ html, highlightedJsx, sourceLink }: ExampleFooterProps) => {
+  const [snippetState, setSnippetState] = useState<SnippetState>(SnippetState.CLOSED);
+  const highlightedHtml = html && highlightHtmlSyntax(html);
+
+  const snippetIdPrefix = useRef(uniqueId('example-snippet-')).current;
```
scheul93, line 35: [question] what is the benefit of using a ref here instead of setting a regular `const snippetIdPrefix = uniqueId('example-snippet-'))`

pwolfert: We do that in some other components, and I don't think we can guarantee that it won't change between renders if we don't. `useRef` makes sure we take the value we get on the first call and stick with that one, even if `uniqueId()` returns a different one on subsequent renders.

forrestbaer (commented):
Looks really good! Nice work pulling this off.

I am noticing a few loki differences though doesn't appear to be anything major, not sure why some of these loki differences show up even when they're seemingly the same image.

forrestbaer (approved):
Looking good!

PR #2011, opened 2022-08-02
[WNMGDS-1813] updating markup on usa banner

scheul93:
## Summary

**Problem**: USA Banner had some invalid mobile markup with  `<div>` & `<p>` tags being used inside a `<button>`

**Solutions explored**

1. Update problem tags to be `<span>`s
2. Update markup so that `<div>` & `<p>` are maintained but move the button element, similar to the [USWDS](https://designsystem.digital.gov/components/banner/) or [VA.gov](https://github.com/department-of-veterans-affairs/vets-design-system-documentation/blob/157cffb3911766964719c0607d764e22a918395a/src/_includes/header.html) implementation

Option 1 was low lift & accessibility expert did not see any concern with this 

Option 2 was a larger lift. First, the VA's mobile design is slightly different than the original USWDS one (it keeps the 'Here's how you know' text when banner is expanded) which may complicate implementation. And USWDS uses an absolutely positioned, transparent button that overlays the whole header text which I didn't feel great about.

So I chose option 1. Please list any concerns or alternatives you may have.

## How to test

`yarn start-storybook` to see it in action.
Visually, nothing should have changed.


pwolfert (approved):
Sounds good to me 👍

PR #2014, opened 2022-08-02
[WNMGDS-1731, WNMGDS-1730] Feedback/Discussion link and Feedback page

zarahzachz:
[Jira](https://jira.cms.gov/browse/WNMGDS-1731)

- Created feedback.mdx page
- Added content as specified in ticket
  - Couldn't find SVGs of company logos in color and pulled what I could find from the Font Awesome folder
  - Had to write HTML in mdx page to get the layout correct
- Added link to the feedback page in the ToC components (desktop and mobile)


packages/docs/src/components/Layout.tsx
```diff
@@ -68,12 +68,15 @@ const Layout = ({ children, frontmatter, location, theme, tableOfContentsData }:
...
+                <TableOfContentsMobile
+                  title={frontmatter.title}
+                  items={tableOfContentsData || []}
+                />
               </div>
               {children}
             </div>
             <div className="ds-l-lg-col--3 ds-u-display--none ds-u-lg-display--block">
-              <TableOfContents items={tableOfContentsData || []} />
+              <TableOfContents title={frontmatter.title} items={tableOfContentsData || []} />
```
zarahzachz, line 79: I forgot about this... Ugh... I added a `title` prop to the ToC components to make it easier to create the GH discussions link. 

I ultimately removed the GH discussions link from the ToC components because that work was part of a different ticket and forgot to remove this prop reference in the process.

scheul93:
If the url for this page is supposed to be `/feedback`, what I would recommend is adding a page in the `src/pages` directory called `feedback.tsx`. The contents of this page can be a copy'paste from the 404 page, but with the query variable being updated to something like
```
export const query = graphql`
  query PageFeedbackQuery {
    mdx(frontmatter: { title: { eq: "Contact us" } }) {
      id
      body
      frontmatter {
        title
      }
    }
  }
`;
```
I think that solves your use case, but I don't know if we want to consider a different pattern for other potential pages like this in the future

packages/docs/src/components/TableOfContents.tsx
```diff
@@ -37,14 +41,23 @@ export const TableOfContentsList = ({ items, level, className = '' }: TableOfCon
...
       <TableOfContentsList items={items} level={level} />
+
+      <h2 className="c-table-of-contents__heading ds-u-margin-y--0 ds-u-md-margin-top--6 ds-u-font-size--base">
+        Have ideas?{' '}
+      </h2>
+      <ul role="list" className="ds-c-list--bare ds-u-md-margin-y--2">
+        <li>
+          <a href="/not-in-sidebar/feedback">Propose a change</a>
+        </li>
+      </ul>
```
scheul93, line 60: similar to the `<TableOfContentsList>`, if this is the same between desktop and mobile, could this be pulled out into a reusable function so it only needs to be maintained once?

forrestbaer (commented):
Looks good! Just wondering if the amount of space here could be reduced slightly:
image

packages/docs/src/components/Layout.tsx
forrestbaer: No worries! You can go ahead and remove all these changes in `TableOfContents` & `TableOfContentsMobile` as I've implemented these (along with the feedback page link, `/feedback` in my PR for the discussions link: #2024

packages/docs/src/components/TableOfContents.tsx
forrestbaer: I've done this in #2024, asking Sarah to remove the TOC changes in this PR.

forrestbaer:
I've updated this PR with the changes requested and added the link to the github discussions page.

- updated feedback url to `/feedback`
- moved feedback section into reusable element
- fixed a few small padding issues

We don't currently have an email address linked in the 'contact us by email' section of the feedback page, need to do something about that.

pwolfert (approved):
From a code standpoint, this is good. I'm going to request that Scott looks at this page and posts a follow-up PR for content and design updates as needed.

PR #2025, opened 2022-08-08
[WNMGDS-1824] removing healthcare 'secondary' variation option from type

scheul93:
## Summary

Removing the `secondary` value for Button variation in Healthcare. This aligns with the work we've already completed but may have missed this small piece.

## How to test

I ran `yarn build:healthcare` and confirmed that the hgov-specific value for this prop is no longer exported as part of the `types`


forrestbaer (approved):
👍🏻

pwolfert (approved):

PR #2030, opened 2022-08-10
[NO-TICKET] doc site updates

scheul93:
## Summary

* Updating nav urls to include query param from current URL
* Instead of `pushState`, updated to `replaceState` so that user cannot use the back button to access a URL without the theme param in URL
* Fixing a CSS property so it has a valid value

## How to test

`yarn start:gatsby`

## Still TODO
Sometimes, locally, the site will give a 404 on a page that should exist. I think it has something to do with Gatsby client-side routes. This is not something I have been able to reproduce in our beta deployment


packages/docs/src/styles/components/storybookExample.scss
```diff
@@ -13,7 +13,7 @@
 }
 
 .c-storybook-example-footer {
-  display: relative;
+  position: relative;
```
pwolfert, line 16: Whoops, that was me 😭

pwolfert (approved):
It works a lot better now! Nice fix

pwolfert:
https://cmsgov.github.io/design-system/branch/akamm/doc-site-updates

forrestbaer (approved):
nice!

PR #2032, opened 2022-08-10
[WNMGDS-1829] Hgov Inset component documentation

scheul93:
## Summary

Adding content from the hgov-specific component `Inset`. I had to create an additional component variable for the theme

## How to test

`yarn start:gatsby` and navigate to Components > Inset

Hgov doc page:
Screen Shot 2022-08-10 at 11 41 48 AM

Other brand doc page:
Screen Shot 2022-08-10 at 11 42 00 AM

### DIFF
--- a/packages/design-system-tokens/src/themes/healthcare-components.ts
+++ b/packages/design-system-tokens/src/themes/healthcare-components.ts
@@ -320,6 +320,10 @@ export const components: AnyTokenValues = {
     '__color--warn':                              t.color['warn'],
   },
 
+  'inset': {
+    '__border-width':                             t.spacer['half']
+  },
+
   'link': {
     '__color':                                    t.color['primary'],
     '__color--active':                            t.color['primary-darkest'],
--- a/packages/ds-healthcare-gov/src/styles/components/_Inset.scss
+++ b/packages/ds-healthcare-gov/src/styles/components/_Inset.scss
@@ -1,7 +1,5 @@
-$hc-c-inset-border-width: $spacer-half;
-
 .hc-c-inset {
-  border-left: $hc-c-inset-border-width solid $color-primary;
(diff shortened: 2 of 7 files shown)


packages/ds-healthcare-gov/src/styles/components/_Inset.scss
```diff
@@ -1,17 +1,15 @@
...
-
 .hc-c-inset {
-  border-left: $hc-c-inset-border-width solid $color-primary;
+  border-left: $inset__border-width solid $color-primary;
   margin-bottom: $spacer-2;
   padding: $spacer-2;
 }
 
 // Align the inset's border with the checkbox/radio field when used as part
 // of the Exposed Within pattern
```
scheul93, line 8: [Question] It looks like there are situations where this component is used within choice fields. There is no example of this on the current HcGov doc site but wondering 1) If we should make a Story for this use-case & 2) if that should be done as part of this ticket or for future documentation updates?

pwolfert: I think if we were to add a story for it, it would be safe to leave that for a future update. We probably want to look at that use case critically and determine if it's something we should consider deprecating. @line47, thoughts?

packages/design-system-tokens/src/themes/healthcare-components.ts
```diff
@@ -320,6 +320,10 @@ export const components: AnyTokenValues = {
     '__color--warn':                              t.color['warn'],
   },
 
+  'inset': {
+    '__border-width':                             t.spacer['half']
```
pwolfert, line 324: Nice 👍

pwolfert (approved):
Besides fixing the PR title, this LGTM 👍

PR #2033, opened 2022-08-11
[WNMGDS-1796] Fix border width issue for pickers in error state

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1796

The error-state border width wasn't previously being defined as a design token theme variable. 

### Fixed

- Fixed a problem where the different border width of error-state text fields caused the calendar picker border to look funny in the `SingleInputDateField`

## How to test

Using [the storybook instance from this branch](https://cmsgov.github.io/design-system/branch/pwolfert/fix-error-border-width/storybook/?path=/story/components-singleinputdatefield--with-error), look at the error story for the calendar picker and confirm that it has been fixed

### Before
Screen Shot 2022-07-22 at 10 51 44 AM

### After
Screen Shot 2022-08-10 at 5 11 04 PM

PR #2034, opened 2022-08-11
[WNMGDS-1796] Fix bad border-radius on one side of picker on medicare

pwolfert:
## Summary

When working on https://github.com/CMSgov/design-system/pull/2033 and testing the medicare version of the picker component, I noticed that the border radius on the calendar button didn't match the border radius for the text input. Even though the picker styles used the `$text-input__border-radius` variable for its border radius so it would match text fields, the text fields in medicare weren't actually taking their value from `$text-input__border-radius`. An override rule was replacing the value defined in the token theme with a static value in Sass.

### Fixed

- Fixed Medicare `ds-c-field` styles not using the `$text-input__border-radius` variable defined in the token theme
- Fixed border radius on the right side of the calendar picker not matching the left side

## How to test

https://cmsgov.github.io/design-system/branch/pwolfert/fix-medicare-text-border-radius/storybook/?path=/story/components-singleinputdatefield--with-error

### Before
Screen Shot 2022-08-10 at 5 14 14 PM

### After
Screen Shot 2022-08-10 at 5 23 15 PM

### DIFF
--- a/packages/design-system-tokens/src/themes/core-components.ts
+++ b/packages/design-system-tokens/src/themes/core-components.ts
@@ -395,6 +395,8 @@ export const components: AnyTokenValues = {
     '__line-height':                              '1.3',
     '__background-color--disabled':               t.color['gray-lighter'],
     '__border-width':                             '2px',
+    '__border-width--error':                      '3px',
+    '__border-width--success':                    '3px',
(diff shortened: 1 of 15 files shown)


forrestbaer (approved):
Good find 🎉

PR #2041, opened 2022-08-11
[No-ticket] release script revision to order tagging differently

forrestbaer:
## Summary

- Changes order of tagging so it tags the branch it's in then creates a temporary branch with the version bumps from there in order to merge those bumps into master
- Puts some of the logic back around creating a temporary branch.

## How to test

`yarn release` will create a new temporary branch after tagging the current branch, then return you to your previous branch. `yarn release --undo` will revert the version bump commit in the active branch and delete the tags last written.

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -53,6 +53,10 @@ if [ "$DELETE_LAST" = true ]; then
     echo "${GREEN}Removing last set of tags...${NC}"
     git tag -d $TAGS
     git push origin --delete $TAGS
+    echo "${GREEN}Tags deleted...${NC}"
+    git revert $LAST_COMMIT --no-edit
+    git push origin
+    echo "${GREEN}Version bumps reverted...${NC}"
   fi
   exit 0
 fi
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -50,17 +50,14 @@ if [ "$DELETE_LAST" = true ]; then
   echo
   if [[ $REPLY =~ ^[Yy]$ ]]
   then
-    echo "${GREEN}Undoing last release...${NC}"
+    echo "${GREEN}Removing last set of tags...${NC}"
     git tag -d $TAGS
     git push origin --delete $TAGS
-    echo "${GREEN}Tags deleted...${NC}"
-    git revert $LAST_COMMIT --no-edit
-    git push origin
-    echo "${GREEN}Version bumps reverted...${NC}"
   fi
   exit 0
 fi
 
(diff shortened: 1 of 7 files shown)


scripts/release.sh
```diff
@@ -71,13 +72,26 @@ if [ "$PRE_VERSION_HASH" = "$POST_VERSION_HASH" ]; then
...
 
+# create temporary release branch from latest commit which is tagged
+# add an empty commit so it can be pushed to origin and PR'd then 
+# move back to previous branch so tags can be deleted with --undo
+# if necessary.
+echo "${GREEN}Creating temporary release branch...${NC}"
+BRANCHREF=$(git rev-parse --short HEAD)
+TEMP_BRANCH="release-${BRANCHREF}"
+git checkout -b $TEMP_BRANCH
+git commit --allow-empty -m "DS Release Bump"
```
pwolfert, line 91: Do we need this commit? What will happen when we try merging this branch into master? Wouldn't it end up merging the whole release branch so far? Maybe we actually need to create a new branch off of master and cherry-pick the bump commit onto this one 😵

forrestbaer: Hmm, it would merge this branch into master but this branch should only have commits that were already in master. So it would only merge the version bumps when brought up to date with master no?

forrestbaer: Hmm, it would merge this branch into master but this branch should only have commits that were already in master. So it would only merge the version bumps when brought up to date with master no?

forrestbaer: Hmm, it would merge this branch into master but this branch should only have commits that were already in master. So it would only merge the version bumps when brought up to date with master no?

forrestbaer: The commit is so we can push this branch to origin, since it's empty.

pwolfert: This temp branch is built off of the release branch, so it would have all previous commits in the release that would become merge conflicts that need to be resolved. That was a shortcoming of the previous way of doing it—if we had cherry-picked releases, we'd often get merge conflicts. I guess either way the issue can be resolved in the pull request that is made.

forrestbaer: Ah, yes that's true, but by cherry picking to the release branch can avoid that, good point.. I'll see what I can do to update for that.

forrestbaer: Doing it that way means we would need to merge the version bump before we could release which kinda stinks.

pwolfert (approved):

PR #2048, opened 2022-08-15
[WNMGDS-1830] pt 2 mgov components

scheul93:
## Summary

Adding docs for medicare's `Card`, `Stars`, Header & Footer. This includes adding some storybook stories to support these components.

## How to test

`yarn build-storybook:gatsby` `yarn build:gatsby` `yarn serve`

### DIFF
--- a/packages/ds-healthcare-gov/src/components/Accordion/Accordion.stories.jsx
+++ b/packages/ds-healthcare-gov/src/components/Accordion/Accordion.stories.jsx
@@ -12,13 +12,7 @@ export default {
     openItems: [0],
   },
   subcomponents: { AccordionItem },
-  decorators: [
-    (Story) => (
-      <div data-theme="healthcare">
-        <Story />
-      </div>
-    ),
-  ],
+  decorators: [(Story) => <div data-theme="healthcare">{Story()}</div>],
 };
 
 const Template = (args) => <Accordion {...args} />;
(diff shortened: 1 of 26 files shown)


forrestbaer (commented):
The React Code for Stars looks broken, maybe something to do with how the story is written?



Everything else looking ok to me

scheul93:
@forrestbaer , good catch. I found [this thread](https://github.com/storybookjs/storybook/issues/12596#issuecomment-723440097) with the same issue and was able to implement a fix

packages/ds-healthcare-gov/src/components/Accordion/Accordion.stories.jsx
```diff
@@ -12,13 +12,7 @@ export default {
...
   },
   subcomponents: { AccordionItem },
-  decorators: [
-    (Story) => (
-      <div data-theme="healthcare">
-        <Story />
-      </div>
-    ),
-  ],
+  decorators: [(Story) => <div data-theme="healthcare">{Story()}</div>],
```
forrestbaer, line 15: I don't think the `data-theme` stuff is necessary here, I removed it locally from `stars` and it came up fine. Since users of the react components won't be using the data-theme attribute I think we can just return the story without wrapper.

forrestbaer (approved):
Looks good :star2:

PR #2049, opened 2022-08-16
[WNMGDS-1837] updating tab title with page name

scheul93:
## Summary

Updating the browser tab title with the page name to help with history navigation.

## How to test

`yarn start:gatsby`. Click around the site and confirm that the tab title updates.

### DIFF
--- a/packages/docs/src/components/Layout.tsx
+++ b/packages/docs/src/components/Layout.tsx
@@ -57,7 +57,7 @@ const Layout = ({
   return (
     <div className="ds-base">
       <Helmet
-        title="CMS Design System"
+        title={`${frontmatter?.title || ''} - CMS Design System`}
         htmlAttributes={{
           lang: 'en',
         }}


forrestbaer (approved):
:+1:

packages/docs/src/components/Layout.tsx
```diff
@@ -57,7 +57,7 @@ const Layout = ({
   return (
     <div className="ds-base">
       <Helmet
-        title="CMS Design System"
+        title={`${frontmatter?.title || ''} - CMS Design System`}
```
pwolfert, line 60: Could this be rearranged so that if there's no title it shows `CMS Design System` instead of `- CMS Design System`? Or is it never the case that a page will not have a title and this is just to appease TypeScript?

scheul93: I don't think a page would every be without a title, but I did update this, just in case

pwolfert (approved):

PR #2050, opened 2022-08-17
[NO-TICKET] Update loki reference images

pwolfert:
## Summary

Discovered that there were four reference images that were missing on master. The other changes just came along for the ride with `yarn loki update`


pwolfert:
I've got another PR coming to fix our CI tests so they catch when a ref image is missing

forrestbaer (approved):

pwolfert:
[The last commit's tests](https://ci.backends.cms.gov/wds/blue/organizations/jenkins/design-system%2Ftests/detail/pwolfert%2Fupdate-loki-refs/2/pipeline) had a ton of failures that didn't actually fail the CI job for some reason. Looking into it

PR #2065, opened 2022-08-24
[no-ticket] [bugfix] Missing export on mgov logo caused build to fail 

forrestbaer:
## Summary

`yarn build:medicare` is currently failing in master because of a missing export on a props interface for `MedicaregovLogo`

## How to test

`yarn install && yarn build && yarn build:medicare`

### DIFF
--- a/packages/design-system-scripts/gulp/build.js
+++ b/packages/design-system-scripts/gulp/build.js
@@ -23,6 +23,7 @@ const getSrcGlob = (src, changedPath) =>
     ? [changedPath]
     : [
         `${src}/**/*.{js,jsx,ts,tsx}`,
+        `!${src}/**/*.stories.{js,jsx,ts,tsx}`,
         `!${src}/setupTests.{js,jsx,ts,tsx}`,
         `!${src}/**/*{.test,.spec,.d}.{js,jsx,ts,tsx}`,
         `!${src}/**/{__mocks__,__tests__}/**/*`,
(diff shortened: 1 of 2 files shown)


scheul93 (approved):

pwolfert:
Any idea where this came from? I see the error too, and I'm trying to figure out what changed

pwolfert:
Is the stories module getting compiled by our build system and its doc generator running and failing to get that type?

pwolfert:
It does look like our `.stories.` files are being included in the build, and I think we should probably ignore them.

forrestbaer:
> It does look like our `.stories.` files are being included in the build, and I think we should probably ignore them.

Yeah, updating the tsconfig and testing.

packages/design-system-scripts/gulp/build.js
```diff
@@ -23,6 +23,7 @@ const getSrcGlob = (src, changedPath) =>
     ? [changedPath]
     : [
         `${src}/**/*.{js,jsx,ts,tsx}`,
+        `!${src}/**/*.stories.{js,jsx,ts,tsx}`,
```
pwolfert, line 26: When you don't fix the logo module, you also need to update line 114. You might as well also update 72 to match. I was modifying the existing ignores for test and definition files like ``!${src}/**/*{.test,.spec,.d,.stories}.{js,jsx,ts,tsx}`,`

pwolfert (approved):
One suggestion but otherwise LGTM

packages/design-system-scripts/gulp/build.js
forrestbaer: Ah, yes, good point, updated.

PR #2071, opened 2022-08-26
[NO TICKET] stylelint cleanup

zarahzachz:
## Summary

This is a combination of fixing and ignoring the majority of CSS stylelint errors we've had. I've added PR comments to explain decisions.

I should note I left errors in packages/docs/src/styles/index.scss unaddressed as it might warrant taking a deeper look at it. Errors in this file are:

- `scss/at-import-no-partial-leading-underscore`
- `no-invalid-position-at-import-rule`

## How to test

1. Pull down branch
2. Run `yarn lint:stylelint`

You should only see errors relating to the index.scss file mentioned above.


packages/design-system/src/styles/components/_Alert.scss
```diff
@@ -50,8 +50,8 @@
 .ds-c-alert__a11y-label {
   margin-right: 0.5rem;
 
-  // !important needed to overwrite screen reader utility styles
   @media (-ms-high-contrast: active), (forced-colors: active) {
+    /* stylelint-disable declaration-no-important -- needed to overwrite screen reader utility styles */
```
zarahzachz, line 54: Inlined the comment as part of the [lint exception description](https://stylelint.io/user-guide/usage/options/#reportdescriptionlessdisables).

For a11y reasons, we have a visually hidden screen reader label to let users on AT know what kind of alert they're reading (since our alert icons aren't visible). HCM users also need a visual queue, so we overwrite the CSS rules when HCM is enabled. This cannot be refactored in a way to remove `!important` as that specificity is needed for overwriting the SR utility class.

packages/design-system/src/styles/components/_Button.scss
```diff
@@ -83,7 +83,7 @@
 
     // Disable
     --backgroundColor--d: #{$button-alt__background-color--disabled};
-    --borderColor: #{$button-alt__border-color--disabled};
+    --borderColor--d: #{$button-alt__border-color--disabled};
```
zarahzachz, line 86: Discovered a hidden bug on Buttons

packages/design-system/src/styles/components/_FilterChip.scss
```diff
@@ -81,12 +81,6 @@ $filter-chip-icon-size: 10px;
   .ds-c-icon--close-thin {
     height: $filter-chip-icon-size + 2;
     width: $filter-chip-icon-size + 2;
-
```
zarahzachz, line 84: Found this style was unnecessary as `.ds-c-icon--close` has the same styles applied and the thin icon also received those styles as well. 

tldr; redundant styling.

packages/design-system/src/styles/components/_SingleInputDateField.scss
```diff
@@ -8,6 +8,7 @@
 // the source rules.
 // @import 'react-day-picker/dist/style.css';
 
+/* stylelint-disable declaration-no-important, no-duplicate-selectors -- Ignore inlined 3rd party styles */
```
zarahzachz, line 11: As explained in the comment above, we're inlining 3rd party CSS to guarantee it loads in the correct order. Ignoring a few linter rules for the 3rd party styles only.

packages/design-system/src/styles/components/_SingleInputDateField.scss
```diff
@@ -34,10 +35,8 @@
 /* Hide elements for devices that are not screen readers */
 .rdp-vhidden {
   box-sizing: border-box;
-  padding: 0;
   margin: 0;
   background: transparent;
-  border: 0;
```
zarahzachz, line 40: Removed redundant CSS as it was an easy thing to do.

packages/design-system/src/styles/components/_SvgIcon.scss
```diff
@@ -62,7 +62,7 @@
   fill: $color-white;
 
   @media (-ms-high-contrast: active), (forced-colors: active) {
-    fill: Window !important;
+    fill: Window;
```
zarahzachz, line 65: Tested this locally and the `!important` didn't seem necessary

packages/design-system/src/styles/components/_Tabs.scss
```diff
@@ -16,7 +16,7 @@
   padding: $spacer-3;
 
   &[aria-hidden='true'] {
-    display: none !important;
+    display: none;
```
zarahzachz, line 19: Tested locally and the `!important` didn't seem necessary.

TODO: Perhaps Tabs should be refactored to use `hidden` instead of `aria-hidden`? The `hidden` attr has the same effect as `aria-hidden` in addition to visually hiding the content when applied - which is what's being done with this CSS.

packages/design-system/src/styles/components/_TooltipIcon.scss
```diff
@@ -20,8 +20,7 @@
   fill: $tooltip-icon__color;
   height: $tooltip-icon-size;
   left: 0;
-  // fixes clipping issue in Firefox, overrides vendor style
-  overflow: visible !important;
+  overflow: visible !important; /* stylelint-disable-line declaration-no-important -- fixes clipping issue in Firefox, overrides vendor style */
```
zarahzachz, line 23: This CSS had a very specific comment explaining why the `!important` was needed, but I couldn't find an independent source saying this was still an issue with Firefox (found some indicator it was true with Safari however).

Local testing removing this rule didn't show any obvious breaking issues and I think it _could_ be removed, but want to revisit later with an official ticket to do better browser testing with a refactor.

packages/docs/src/styles/components/embeddedExamples.scss
```diff
@@ -7,7 +7,7 @@
 }
 
 .example-list {
-  & article {
+  article {
```
zarahzachz, line 10: This use of the & isn't doing anything because the styles are nested.

packages/ds-medicare-gov/src/styles/components/_Navbar.scss
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable declaration-no-important -- Temporarily ignored until refactor can happen */
```
zarahzachz, line 1: Ignoring as it's `.m-c-` namespaced styles as they were added by outside teams and it's unknown what they're overwriting without doing a deeper dive.

packages/ds-medicare-gov/src/styles/components/_NavigationMenu.scss
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable declaration-no-important -- Temporarily ignore until mgov overrides can be refactored */
```
zarahzachz, line 1: Ignoring as it's `.m-c-` namespaced styles as they were added by outside teams and it's unknown what they're overwriting without doing a deeper dive.

packages/ds-medicare-gov/src/styles/components/_SimpleFooter.scss
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable declaration-no-important -- Temporarily ignore until mgov can be refactored */
```
zarahzachz, line 1: Ignoring as it's `.m-c-` namespaced styles as they were added by outside teams and it's unknown what they're overwriting without doing a deeper dive.

packages/ds-medicare-gov/src/styles/components/_override.dialog.scss
```diff
@@ -1,3 +1,4 @@
+/* stylelint-disable declaration-no-important -- temporarily ignore until mgov can be refactored */
```
zarahzachz, line 1: It's unknown what mgov is overwriting and why - should be migrated up the chain for tokens? Need a ticket for deeper investigation

packages/ds-medicare-gov/src/styles/settings/variables/_layout.scss
```diff
@@ -1 +1,3 @@
+/* stylelint-disable declaration-no-important -- Temporarily ignore until mgov can be refactored */
 $card-border-radius: 6px !important;
```
zarahzachz, line 2: Maybe this needed to be migrated to tokens? Needs ticket to do deeper investigation.

packages/design-system/src/styles/components/_Button.scss
pwolfert: Nice!

packages/design-system/src/styles/components/_Tabs.scss
pwolfert: Maybe it would be worth adding a "TODO" comment in the code 😆

zarahzachz:
I ran Loki off `master` and updated incorrect refs. Used `git switch` to move updated refs from `master` to this branch, and ran `yarn loki test`. No errors. 

Committed updated ref images as part of this PR.

PR #2063, opened 2022-08-24
[WNMGDS-1245] Update autocomplete button 

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1245)

- Added `ds-u-padding-right--0` to the Autocomplete clear button
  - I attempted updating padding on the `ds-c-autocomplete__clear-btn` class, however the cascade of Buttons loading after Autocomplete caused this style to be overwritten without adding `!important` or overly specific classes to this component
  - Using a utility class in this context poses little to no risk as we don't offer users a prop to add their own CSS class to this button - therefore it won't interfere with user implementation
- Updated Autocomplete stories to comply with sentence casing rules
- Updated snapshot tests to match new markup changes

## How to test

1. Visit [Autocomplete docs](https://cmsgov.github.io/design-system/branch/WNMGDS-1245/bugfix-autocomplete-btn-alignment/components/autocomplete/?theme=core)
2. See that clear button is right aligned with input field

### If this is a change to code:

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[ticket] title`
- [x] Ran `yarn type-check && yarn lint && yarn test` with no errors
- [x] Updated unit test snapshots and loki reference images **if necessary** with `yarn loki` && `yarn update-snapshots`

### If this is a change to documentation:

- [x] Checked for spelling and grammatical errors

PR #2079, opened 2022-08-29
[WNMGDS-1857] MGOV typography to tokens, doc site updates

forrestbaer:
## Summary

- Moved several typography settings into theme configuration options so medicare styles will carry through to gatsby docs site
- Extended gatsby headings with `ds-hX` classes to pick up correct font sizing & family settings
- Updated mgov theme with correct font token overrides
- Updated loki snapshots (from Sarah's recent Autocomplete bugfix #2063)
- Updated several class names to use correct token names instead of older style names
- changed `build-tokens` to `build:tokens` to match our `verb:object` layout

## How to test

1. `yarn install && yarn start` - switch themes, check out typography pages, medicare and other themes


pwolfert:
I was looking through the ref image changes, and saw that the text casing changes in .loki/reference/healthcare_chrome_iphone7_Components_AutoComplete_No_Results.png. I'm trying to figure out where that change comes from. Going to check master

forrestbaer:
> Update: I ran `yarn loki:healthcare update` on master, and this casing change did not show up among the other outdated refs

Oh, shoot, I just assumed it came from the work on autocomplete, I'll take a look.

packages/docs/src/styles/pages/layout.scss
```diff
@@ -21,33 +21,38 @@ body {
...
+
 @media (min-width: $width-lg) {
   .page-content__content {
     // removing margin for first heading on content pages
     & > :nth-child(2) {
       margin-top: 0;
     }
     > h2 {
-      font-size: $font-size-3xl;
       font-weight: normal;
```
pwolfert, line 49: Should this font weight come from a token as well? In core it looks like the h2s on our pages are the only headings besides 4xl that aren't bold. Not sure if that's intentional. Based on `t.font['weight-bold']` in the theme file, I'd think it'd be bold.

packages/design-system-tokens/src/themes/core-components.ts
```diff
@@ -418,6 +418,16 @@ export const components: AnyTokenValues = {
     '-trigger__color':                            t.color['primary'],
   },
 
+  'typography': {
+    '-heading__font-family':                      'inherit',
+    '-heading__font-weight':                      t.font['weight-bold'],
+    '-heading-4xl__font-weight':                  'inherit',
+    '-heading-4xl__line-height':                  'inherit',
+    '-text__color':                               'inherit',
+    '-text__font-family':                         'inherit',
```
line47, line 427: Wondering about naming here and using something like   
  
\- text-heading for headings and text-body for body text?

packages/docs/src/styles/pages/layout.scss
```diff
@@ -21,33 +21,38 @@ body {
   flex: 1 0 auto;
 }
 
+.page-content__content {
+  > h2 {
+    @extend .ds-h2;
```
line47, line 26: Trying to phase out these `ds-hx` classes in favor of `ds-text-heading--5xl` `ds-text-heading--4xl` etc...   
Could we update to use those instead?

zarahzachz:
> I was looking through the ref image changes, and saw that the text casing changes

I updated the Autocomplete casing in our Storybook stories to reflect our guidelines. Buttons and form labels are documented to be written using sentence casing and we were using title casing in the original examples - breaking our own rules.

packages/docs/src/styles/pages/layout.scss
zarahzachz, line 26: I was under the impression we wanted to avoid `@extend` rules?

forrestbaer: Well the other option would be to use a plugin with gatsby to add the heading classes to our markup content, what do you think @pwolfert ?

packages/design-system-tokens/src/themes/core-components.ts
forrestbaer: Adding `-body` here as I think removing the secondary name makes it too generic, gotta leave room for more specificity in the future.

packages/design-system-tokens/src/themes/medicare-components.ts
```diff
@@ -419,6 +419,15 @@ export const components: AnyTokenValues = {
     '-trigger__color':                            t.color['base'],
   },
 
+  'typography': {
+    '-heading__font-family':                      t.font['family-montserrat'],
+    '-heading__font-weight':                      t.font['weight-semibold'],
+    '-heading-4xl__font-weight':                  300,
```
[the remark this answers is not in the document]
forrestbaer: I suppose, we haven't had it in the past. It would be below 'normal', 'semibold', 'bold' .. should it be .. `light` ?

packages/docs/src/styles/pages/layout.scss
forrestbaer: Good call, didn't know about that. :+1:

forrestbaer: I actually looked into this while working on these and it was something of a pain to get going.. I agree using extend isn't great but the other option may need another ticket maybe and could be expanded to setup classes for all our elements to utilize our own styling classes.

packages/docs/src/styles/pages/layout.scss
forrestbaer: Removing the weight setting here to try it out, good call out.

packages/design-system-tokens/src/themes/medicare-components.ts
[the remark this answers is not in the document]
forrestbaer: I added `light`

packages/docs/src/styles/pages/layout.scss
pwolfert: It's not in code that we're releasing to other teams, so I think it's a fine solution for now.

pwolfert (commented):
The changes look good to me, but I'm trying to figure out whether the loki failures I'm getting locally mean anything. It would probably be helpful to run the loki update on the main branch and merge it so we can reduce the amount of loki noise we're seeing on this branch/PR

forrestbaer:
> The changes look good to me, but I'm trying to figure out whether the loki failures I'm getting locally mean anything. It would probably be helpful to run the loki update on the main branch and merge it so we can reduce the amount of loki noise we're seeing on this branch/PR

Good point #2088

babsdenney:
@forrestbaer We reviewed all the font styles and below some changes needed. 

**Header 4xl and 5xl**
- Change font-weight to 600

**Header 3xl**
- Line-height to 1.2
- Font-weight to 600

**Header 2xl**
- Line-height to 1.25

**Header xl**
- Line-height to 1.4

**Header md**
- Font-weight to 600
- Line-height to 1.5

**Header sm**
- Font-weight to 600
- Color to #323A45
- Line-height to 1.4

**Body lg**
- Line-height 1.4
- Color to #323A45

**Body md**
- Color to #323A45

**Body sm**
- Color to #323A45
- Line height to 1.4

forrestbaer:
@pwolfert @zarahzachz one of you will have to finish this one up, I'm not going to have time to finish these other updates today.

forrestbaer:
@pwolfert I was able to resolve most of the loki errors, there is just one left or so regarding the footer.

packages/ds-medicare-gov/src/styles/base/_override.typography.scss
```diff
@@ -28,26 +26,18 @@
...
-  font-weight: 300;
-}
-
 .ds-h1,
 .ds-text-heading--3xl {
-  font-size: $font-size-2xl;
   line-height: $spacer-5;
   margin-bottom: 44px;
   @media screen and (min-width: $width-md) {
-    font-size: $font-size-3xl;
```
pwolfert, line 42: I'm tracking down the sources of the loki differences, and this is the source of one of them. The font-size used to be different in medium screens vs everything smaller than that. This is responsive difference is not something we've tried to capture with tokens before. @babsdenney, is this mobile/desktop difference correct? Or should we normalize it to one size? It's too bad that the size is in pixels.

pwolfert: For my own notes: On the main branch, it's 32px on small screens and 40px on large screens. On this branch, it's 30px on small screens and 40px on large screens.

pwolfert: Theoretically if we wanted to capture different values for different screen sizes in the tokens, we could use this pattern a pattern like `typography-heading-3xl__font-size--width-sm` 🤷‍♂️

packages/design-system/src/styles/base/_typography.scss
```diff
@@ -132,7 +121,9 @@
 .ds-text--lead,
 .ds-text-body--lg,
 .ds-text-body--md {
-  line-height: $base-line-height;
+  color: $typography-body__color;
```
pwolfert, line 124: Loki also picks up a visual difference in the drawer footer content that uses the `ds-text` component. This new `color` declaration changes the color of the text in that component from a color set by a `dialog` rule to `#404040`

packages/design-system-tokens/src/themes/medicare-components.ts
```diff
@@ -419,6 +419,18 @@ export const components: AnyTokenValues = {
...
 
+  'typography': {
+    '-heading__font-family':                      t.font['family-montserrat'],
+    '-heading__font-weight':                      t.font['weight-semibold'],
+    '-heading-4xl__font-weight':                  t.font['weight-light'],
+    '-heading-3xl__font-weight':                  t.font['weight-semibold'],
+    '-heading-lg__font-weight':                   t.font['weight-semibold'],
+    '-heading-md__font-weight':                   t.font['weight-semibold'],
+    '-heading-sm__font-weight':                   t.font['weight-semibold'],
+    '-body__color':                               t.color['base'],
```
pwolfert, line 430: It looks like `t.color['base']` is the same value used for the color in `.ds-base`, so I'm wondering if this can be `inherit` like the other themes given [the change found here](https://github.com/CMSgov/design-system/pull/2079/files#r960109260). The color of the drawer footer content on master might be a bug, though.

pwolfert: The problem I see with setting this to `base `explicitly here is that if we're using _on dark_ style in a parent element, this won't pick up on that automatically. Not sure if that's an actual problem, though.

packages/design-system/src/styles/base/_typography.scss
forrestbaer: Hrm, it looks like there isn't currently a variable for this, so would need to create something like `drawer-footer__color` to add to the drawer component theme and set it in `_Drawer.scss`, sound like the right thing?

pwolfert: I'm wondering if we need to actually remove the color declaration here because these are supposed to inherit their color from `ds-base`. If this came from a mgov override, maybe it actually didn't need to be there. Was it [this one?](https://github.com/CMSgov/design-system/pull/2079/files/63a9a7c002ceeccd7edf387911be24622d94ca2e#diff-6b1d7de20b3bbfaf673aa2826fc6b181293c8de6cb898e66001afc9e499cbbf9L95)
```css
color: $black-500;
```
What if we removed it entirely? Does the `ds-text-body` need a different font color, @babsdenney ?

pwolfert: On the other hand, the styles we're getting from that `dialog` rule are probably wrong to begin with. In that case, the snapshot change we're seeing is correct, and we might even want to remove that `.dialog` color declaration from core.

forrestbaer: It looks like `black-500` doesn't even exist, it was removed at some point with tokens. The medicare references file needs to be updated with granite values that reflect the old black values.

forrestbaer: Scratch my last deleted comment.. `black-500` is mapped to `granite-800` in tokens which is the same as `base` for medicare.  Since that was an override, I can't see how it would have changed other than inheriting from dialog like you noted which is why I suggested a custom token and color setting for drawer footer..  At this point I'm a little confused :)

babsdenney: That sounds good! I like the option of removing it entirely so it inherits the ds-base style. Headers and body styles use the same color ds-base color.

forrestbaer: I've removed this component style so we inherit `ds-base` which is the lighter grey color for body items which passes over into list headings and such. I've updated loki snapshots which you can both look at to verify these changes look correct? 

Also there is an update to the drawer background for the close button in core which I believe is now the standard 'outline' style which does not have a background (was previously white)

pwolfert: Those changes look good to me, but @babsdenney or @line47 should give the final word

line47: Looks good to me as well :+1:

line47: I'll give @babsdenney the final review

babsdenney (approved):

pwolfert (approved):

PR #2086, opened 2022-08-30
[WNMGDS-1877] filter chip, pagination, & spinner test updates

scheul93:
## Summary

- Converting FilterChip, Pagination & Spinner tests to React Testing Library

## How to test


forrestbaer (approved):

packages/design-system/src/components/FilterChip/FilterChip.test.tsx
```diff
@@ -0,0 +1,91 @@
...
+    const props = { ...defaultProps, ...customProps };
+    return render(<FilterChip {...props} />);
+  };
+
+  it('should include children as label', () => {
+    renderFilterChip();
+    const chipEl = screen.getByRole('button');
+    expect(chipEl.textContent).toEqual('Foo. Remove Foo filter .');
+
+    expect(chipEl.innerHTML).toMatchInlineSnapshot(
```
pwolfert, line 21: What's the benefit of this being inline vs in a snapshot file?

scheul93: Not sure if there's a benefit. Is it harder to read if it's inline?

pwolfert: All our other snapshots are in separate files, so I was wondering if this was a special case for a hidden reason. I do think it's a little easier, but I'm mainly interested in consistency

packages/design-system/src/components/FilterChip/FilterChip.test.tsx
```diff
@@ -0,0 +1,87 @@
...
+      renderFilterChip({ onDelete: onDelete });
+
+      const chipEl = screen.getByRole('button');
+      expect(chipEl).toBeDefined();
+
+      userEvent.click(chipEl);
+      expect(onDelete).toHaveBeenCalled();
+    });
+
+    it('should call onDelete when certain keyboard keys are pressed', () => {
```
pwolfert, line 42: 100% test coverage, and they're concise and clear!

pwolfert (approved):
🎉

packages/design-system/src/components/Pagination/__snapshots__/Page.test.tsx.snap
```diff
@@ -1,25 +1,28 @@
...
-  <Button
-    aria-label="page 1"
-    href="/#1"
-    onClick={[MockFunction]}
-    variation="ghost"
-  >
-    1
-  </Button>
-</li>
+<div>
```
zarahzachz, line 4: Something I noticed doing my tests was RTL adds an extra wrapping div to components. I don't know if this is an issue or not with our snapshots, but it's worth calling out that this wrapping div may not be part of the actual component we're testing.

scheul93: Good call. I updated to target the element that `Page` actually renders. This makes it more consistent with your accordion tests.

pwolfert: @zarahzachz, if you use `container` to take the snapshot, that container defaults to a `div` and will therefore wrap your snapshot in a div. If you use `asFragment()`, which we do in some of our other tests, the wrapping element in the snapshots will be a `<DocumentFragment>`. I'm a fan of trying to target the specific element you want to snapshot in general, as Anna suggested; it makes for smaller snapshots with less noise.

PR #2088, opened 2022-08-30
[no-ticket] updating loki snapshots

forrestbaer:
## Summary

- Update of loki snapshots from recent master


pwolfert (approved):
😮 we were behind!

There was a slight change in the healthcare.gov logo, and I'm not sure where that came from. It shouldn't really affect anything, though, so 🤷‍♂️

PR #2092, opened 2022-08-30
[WNMGDS-1895] mgov help drawer

scheul93:
## Summary

- Exporting the Mgov HelpDrawerToggle with icon defaulted
- Exporting the mgov IconCircleIconOutline icon
- Exporting the mgov Dialog with proper close icon defaulted

## How to test

1. `yarn start` to see the Dialog & Icon pages updated
2. `yarn storybook:medicare` to see the mgov-specific Dialog & Help Drawer Toggle

[Demo site](https://cmsgov.github.io/design-system/branch/WNMGDS-1895/mgov-help-drawer) to check out updated docs & storybook

- [X] Updated unit test snapshots and loki reference images **if necessary** with `yarn loki` && `yarn update-snapshots`

### Input

- Is there a better representation of the Dialog story where the dialog is actually open to compare to?


packages/ds-medicare-gov/src/components/HelpDrawer/HelpDrawer.stories.jsx
```diff
@@ -80,7 +79,6 @@ export const HelpDrawerToggleWithDrawer = () => {
       <HelpDrawerToggle
         showDrawer={() => setIsDrawerVisible({ isDrawerVisible: true })}
         helpDrawerOpen={isDrawerVisible || false}
-        icon={<InfoCircleOutlineIcon />}
```
pwolfert, line 83: Nice 👍

packages/ds-medicare-gov/src/components/Dialog/Dialog.stories.jsx
```diff
@@ -0,0 +1,73 @@
+import React, { useState } from 'react';
+import Dialog from './Dialog';
+import { Button } from '@cmsgov/design-system';
+
+export default {
+  title: 'Medicare/Dialog',
```
pwolfert, line 6: What does this new story do?

packages/ds-medicare-gov/src/components/Icons/Checkmark.tsx
```diff
@@ -10,7 +10,7 @@ const Checkmark: FunctionComponent<CheckmarkProps> = ({ className }) => {
     `[Deprecated]: Please use the <CheckIcon /> component with 'ds-c-icon-color--primary' CSS class instead. This component will be removed in a future release.`
   );
 
-  return <CheckIcon className={`ds-c-icon-color--primary className`} />;
```
pwolfert, line 13: Oh, nice find!

zarahzachz (approved):
lgtm!

packages/ds-medicare-gov/src/components/Dialog/Dialog.stories.jsx
scheul93: So the Medicare Dialog has a different icon for the close button than core & healthcare. This story is a direct copy/paste of the original Dialog except it uses the mgov Dialog (with defaults) instead of core.

I wasn't sure of another way to show this subtle difference in the core story.

pwolfert: Gotcha. If we want the visual regression tests to capture that difference, shouldn't we have it show the dialog by default rather than using a triggering button? Otherwise, we should probably configure loki to ignore this one because it's a duplicate of the original (just a screenshot of a solid button)

scheul93: I tried showing just the dialog modal in the story without any Button or interaction needed. However, this caused some issues with loading the React code on the documentation site. It seems to be something about the Dialog on the "docs" tab of Storybook that was creating issues.
I spent some time trying different variations and none were working, so I'm opting to skip for now.

pwolfert (approved):

PR #2094, opened 2022-08-31
[WNMGDS-1873] Update Review unit tests

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1873)

I updated the unit tests for Review and ReviewLink to remove Enzyme and use RTL.

- [ ] Updated unit test snapshots and loki reference images **if necessary** with `yarn loki` && `yarn update-snapshots`


packages/design-system/src/components/Review/ReviewLink.test.tsx
```diff
@@ -0,0 +1,42 @@
...
+    renderLink({ onClick });
+
+    const els = screen.getAllByRole('link');
+    expect(els.length).toBe(1);
+
+    const el = els[0];
+    expect(el.textContent).toBe(text);
+
+    userEvent.click(el);
+    expect(onClick).toHaveBeenCalledTimes(1);
```
scheul93, line 40: Should this test also test that `onClick` was called with specific parameter values?

zarahzachz: Good call. I updated to include parameters. With RTL, I couldn't specify a mock event like before (`userEvent.click` doesn't allow for this parameter type to get passed). I used `expect.anything()` for that param because I couldn't find docs and approached the same frustration levels as this [SO OP](https://stackoverflow.com/questions/63735881/react-jest-testing-mock-a-click-event-object).

scheul93 (approved):

PR #2095, opened 2022-09-01
[WNMGDS-1872] vertical nav test updates to testing library

scheul93:
## Summary

- Updating `VerticalNav`, `VerticalNavItem` & `VerticalNavItemLabel` tests to React Testing Library
- Updating one small thing in `VerticalNavItem` to pass a defined id

## How to test

1. `yarn test:unit`


packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
```diff
@@ -88,7 +88,7 @@ export const VerticalNavItem = (props: VerticalNavItemProps): React.ReactElement
     setCollapsed(!collapsed);
 
     if (props.onSubnavToggle) {
-      props.onSubnavToggle(props.id, collapsed);
+      props.onSubnavToggle(id, collapsed);
```
scheul93, line 91: Updated because `props.id` is an optional prop. If id is not provided as a prop, this parameter value is `undefined`. By updating to use the function variable `id`, there will always be a value passed to the callback

packages/design-system/src/components/VerticalNav/VerticalNavItem.test.tsx
```diff
@@ -121,97 +143,70 @@ describe('VerticalNavItem', () => {
...
+      expect(firstSubNavItemEl).toBeDefined();
     });
 
-    describe('with url', () => {
-      beforeEach(() => {
-        props.url = '/foo';
-      });
+    it('adds top-level link to top of subnav', () => {
+      // if a url is provided to an item that also has a subnav, another subnav item will be created at the top
+      // this is because a subnav label is a button for expanding / collapsing the subnav, so the link must be accessed somehow
```
pwolfert, line 203: Thank you for the note

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
pwolfert: Good catch 👍

pwolfert (commented):
LGTM, and thanks for increasing test coverage 👍

pwolfert (approved):

packages/design-system/src/components/VerticalNav/VerticalNav.test.tsx
```diff
@@ -1,111 +1,113 @@
...
     it('does not have aria-label by default', () => {
-      const data = render();
-      const wrapper = data.wrapper;
-
-      const navEl = wrapper.find('nav');
-      const ariaLabelAttr = navEl.prop('aria-label');
+      const { container } = renderVerticalNav();
+      // not sure how else to query the nav container
+      // eslint-disable-next-line testing-library/no-container, testing-library/no-node-access
+      const navEl = container.querySelector('nav');
```
zarahzachz, line 51: I think you could query by `getByRole('navigation')`? This works too.

packages/design-system/src/components/VerticalNav/VerticalNav.test.tsx
```diff
@@ -137,48 +139,20 @@ describe('VerticalNav', () => {
...
-        .hasClass('ds-c-vertical-nav__label--current')
-    ).toBe(true);
-
-    expect(
-      grandchild1Wrapper
-        .find('.ds-c-vertical-nav__label')
-        .first()
-        .hasClass('ds-c-vertical-nav__label--current')
-    ).toBe(true);
+    expect(parentEl.classList).toContain('ds-c-vertical-nav__label--current');
```
zarahzachz, line 150: Would this be a good test to check for the `aria-current` property? I believe that's a recent addition and might not be covered by original unit tests.

zarahzachz (approved):
Looks good. Left a few comments, but not a blocker.

PR #2099, opened 2022-09-02
[WNMGDS-1903] Fix "cmsds-migrate" script failing in downstream projects

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1903

When testing this in the context of a particular downstream application, we found and addressed the following issues:

1. Not all the script files were included when publishing to npm
    - Fix: Include the `scripts` folder in the `files` list in our `package.json`
2. Their project version of node was not interpretting our ES-Module syntax as ES Modules
    - Fix: Rename `.js` files to `.mjs` (and it works in our project)
3. Their project version of node did not support some of the code in our scripts
    - Fix: Change some lines to use backwards-compatible methods of doing the same thing
4. One of our script dependencies wasn't resolving to the right version in their project
    - Fix: Move our script dependencies from `devDependencies` to `dependencies` so the npm client downstream will care

## How to test

1. `yarn build` (not really important, but it makes sure there's a `dist` folder like we normally have)
2. `cd packages/design-system`
3. `npm pack`
4. Copy the generated `cmsgov-design-system-4.0.0.tgz` file into a downstream project folder and then run `npm install ./cmsgov-design-system-4.0.0.tgz`
5. Run `npx cmsds-migrate` in that project


forrestbaer (approved):
Looks good! 👍🏻

PR #2100, opened 2022-09-02
[WNMGDS-1871] Migrate unit tests for TextField to RTL

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1871

## How to test

`yarn test:unit`


packages/design-system/src/components/TextField/TextField.test.tsx
```diff
@@ -1,27 +1,28 @@
...
 
 describe('TextField', function () {
   it('renders', () => {
-    const data = render();
-    expect(data.wrapper).toMatchSnapshot();
+    expect(renderTextField().asFragment()).toMatchSnapshot();
+  });
+
+  it('renders with a label mask', () => {
+    expect(renderTextField({ labelMask: DATE_MASK }).asFragment()).toMatchSnapshot();
```
pwolfert, line 21: I considered not using snapshots, but any meaningful queries I could make against these two other cases would expose implementation details and be more work to maintain. I figured since the snapshots are fairly small, they're more maintainable than the alternative.

zarahzachz: Seems like a good use-case for snapshots

zarahzachz (approved):

PR #2102, opened 2022-09-06
[no-ticket] [bugfix] Update cmsds-migrate to convert 'transparent' to 'ghost'

forrestbaer:
## Summary

- `transparent` is now converted to `ghost`
- removed a couple escape backslashes that linter marked as un-necessary

## How to test

1. `yarn install && yarn cmsds-migrate` - run it using button && react configs, should update appropriately


pwolfert (approved):
Tested this out, and it still works. If you run the cleanup script in the root of our project, it will remove the comments from the migration script itself 😂

PR #2104, opened 2022-09-07
[WNMGDS-1856] Updating doc site privacy button to match other footer links.

forrestbaer:
## Summary

- Changed `button` to `a` so item matches other doc site footer items
 
image

## How to test

1. `yarn install && yarn start` - evaluate footer, link should work as before

## Checklist


packages/docs/src/components/DocSiteFooter.tsx
```diff
@@ -14,12 +14,9 @@ const DocSiteFooter = () => {
           >
             Plain Writing
           </a>
-          <button
-            className="ds-c-button ds-c-button--transparent c-footer__link"
-            data-privacy-policy="modal-trigger-footer"
-          >
+          <a className="c-footer__link" data-privacy-policy="modal-trigger-footer">
```
pwolfert, line 17: I should have looked at the ticket beforehand, but I think the solution needs to be to make it a button that is styled like a link...although that doesn't seem great. The original reason it was made a button instead of a link is that it opens a dialog instead of navigating to another page, and the a11y guidance we were given at the time is that it needed to be a button. @SparrowSpit, if you could weigh in on this, we'd appreciate it.

zarahzachz: Came to comment this. It's triggering an action and not navigating somewhere, so I think to be semantically correct it needs to stay a button element.

zarahzachz: @pwolfert are we actually tracking any of this stuff? Wondering if it could just be removed from our footer altogether? 

Accounting for theming makes this extra tricky. If we have to keep this button, this could be a good solution:

```css
button.c-footer__link {
    all: inherit; /* removes all DS and browser button styling */
    text-decoration: underline;
    color: var(--link__color);
    cursor: pointer;
}

/* Needs hover color change styles somewhere to be a better match */
```

forrestbaer: Ah, of course, good point.. Can we just remove this link?

zarahzachz: It also doesn't seem to work on Firefox

forrestbaer: Bummer, ok, so we need to style this like a link then, i'll set that up, seems like the lowest cost option.

zarahzachz: I meant the modal doesn't trigger via this button/link. Just more fuel to delete it from the footer.

forrestbaer: Updated.

pwolfert (approved):
It blends in so nicely 🥷

PR #2109, opened 2022-09-07
[WNMGDS-1189] Fixing A11y errors in StepList

forrestbaer:
## Summary

- `StepList` was throwing a11y errors because the `<div>`'s which had `aria-label`'s on them did not have an appropriate `role`, resolved this

## How to test

1. `yarn install && yarn test:a11y` - note StepList no longer throws an error, can evaluate in Axe in chrome as well using a storybook iframe, if you do it this way there will be other warnings/errors that are not related to this component specifically based on [WCAG 2.0 AA](https://www.w3.org/TR/WCAG20/) which is what we attempt to meet.

## Checklist


pwolfert:
Because this solution is different from the "short-term" solution identified in the ticket comments to...

> use aria-describedby instead of aria-label

I'll need to defer to @SparrowSpit to verify this change

forrestbaer:
> use aria-describedby instead of aria-label

I am confused as to what element this element would be `describedby`, interested in understanding the correct approach.

forrestbaer:
@pwolfert @zarahzachz updated with recommendations from Dan [in this slack thread](https://adhoc.slack.com/archives/C645GLLNN/p1662986691868769?thread_ts=1662668151.718039&cid=C645GLLNN).

pwolfert (approved):

PR #2111, opened 2022-09-07
[NO-TICKET] Move test configs out of the `design-system-scripts` package

pwolfert:
## Summary

I wanted to move our test configs into the root and carve out a space for custom visual regression tests to live as well.

## How to test

1. `yarn test:unit`
2. `yarn test:a11y` + variations for CDSs

## Checklist


pwolfert:
The CI build for [be87dd1](https://github.com/CMSgov/design-system/pull/2111/commits/be87dd1fadb4ed27833f799fbda4f853951a868d) actually did fail at first. The only reason it is reporting green is that I re-ran it with a change that built core first.

forrestbaer (commented):
Nice! But I ran into an issue that it opened a local chrome window and spammed through all the urls live in my browser, that wasn't happening before. Or at least I wasn't aware of it because the wrapper backgrounded it somehow? Seems ok but a little strange of an experience, I wonder if it will effect things in CI.

PR #2112, opened 2022-09-07
[NO-TICKET] Put Typography and Layout Grid under Foundations in Storybook

pwolfert:
## Summary

This came out of a conversation we had last week. A change I made resulted in a `Typography` folder just hanging out by itself at the top of the nav panel in Storybook. This reorganization makes it match what we have on the doc site.

Note that I also let loki add its ref images, but it doesn't prune old files unless we use `loki approve`. We can't use `loki approve` on one design system without wiping out the refs for the others. We would be able to fix this by moving to a custom VRT solution.

## How to test

`yarn storybook`

## Checklist


forrestbaer (approved):
Hope ya don't mind I added a unique id to the 'equal' story keys to get rid of a console warning that was showing up there.. This looks good to me! 👍🏻

PR #2113, opened 2022-09-08
[WNMGDS-1870] Refactor MonthPicker and Choice unit tests

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1870)

Refactored Choice and MonthPicker unit tests to remove Enzyme and use React Testing Library.

I also fixed an inaccessibility thing in MonthPicker.

## How to test

1. Pull down this branch
2. Run `yarn test:unit`

## Checklist


packages/design-system/src/components/MonthPicker/MonthPicker.test.tsx
```diff
@@ -0,0 +1,181 @@
...
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('renders inverse style', () => {
+    renderMonthPicker({ inversed: true });
+
+    const buttons = screen.getAllByRole('button');
+    const checkboxes = screen.getAllByRole('checkbox');
+
+    buttons.forEach((button) => expect(button.classList).toContain('ds-c-button--on-dark'));
```
forrestbaer, line 30: The way I've seen this done in most cases has been to use `toHaveClass`, but this seems fine, just something I noticed.

packages/design-system/src/components/MonthPicker/MonthPicker.tsx
```diff
@@ -165,7 +165,7 @@ export const MonthPicker = (props: MonthPickerProps) => {
         </Button>
       </div>
       <div className="ds-c-month-picker__months">
-        <ol className="ds-c-list--bare ds-c-month-picker__months-list">
+        <ol role="list" className="ds-c-list--bare ds-c-month-picker__months-list">
```
forrestbaer, line 168: This role is already defined by `ol`, I think this can be removed?

zarahzachz: Fun HTML bug for Safari - when you style a list to remove the list style Safari removes the list role as well. So the recommendation to support Safari is to include the role when removing list styles.

zarahzachz: In short... Safari's gonna Safari.

packages/design-system/src/components/MonthPicker/MonthPicker.test.tsx
zarahzachz: Oh, I was just following what I thought was convention in our DS. Pushing an update to fix this!

packages/design-system/src/components/MonthPicker/MonthPicker.tsx
zarahzachz: It might actually be a better solution in our DS to tie unstyled list styles to the `role="list"` attribute to ensure a11y gets included, but maybe that's overkill? Gonna post in GH to discuss.

pwolfert: oh wow, that's strange!

PR #2117, opened 2022-09-08
[WNMGDS-1897] Fix color swatch hex values not updating for non-core themes

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1897

- Make the color swatch component dependent on the current `theme` so it will re-render when the theme changes
- Switch from `useLayoutEffect` to `useEffect` because `useLayoutEffect` will always run before the DOM gets updated, which means it will never have a value other than what it gets from the default styles that are applied for the `core` theme before the frontend app hydrates

This is a quick fix and should make the theme dependency clearer in the code, but a better long-term solution would be to figure out a way to get static versions of each page for each theme as explained in https://jira.cms.gov/browse/WNMGDS-1899. I'm going to continue to look into that, but for now I want to get a fix up.

## How to test

`yarn build:docs && yarn serve:docs` and navigate to http://localhost:9000/foundation/color/?theme=core

## Checklist

### DIFF
--- a/packages/docs/content/foundation/color.mdx
+++ b/packages/docs/content/foundation/color.mdx
@@ -4,8 +4,6 @@ order: 20
 intro: The design system provides a flexible color palette that meets accessible color contrast requirements.
 ---
 
-import ColorSwatchList from '../../src/components/ColorSwatchList';
-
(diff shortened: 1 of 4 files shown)


forrestbaer (approved):
Proper! 💯

PR #2120, opened 2022-09-09
[WNMGDS-1913] a11y automated test cleanup

forrestbaer:
## Summary

- Resolved several contrast errors by adding a timeout between when the page is loaded and when it is analyzed based on the test name. Some components (drawer) require longer than others (monthpicker).. I found by adding a 100ms timeout for all tests the number of failures was greatly reduced, may need to modify in CI. 
- Resolved an issue with sticky drawer where it's body region was not `tabindex`'d
- Resolved a contrast issue with disabled elements in pagination, added `aria-disabled=true` to resolve
- Updated tests to fail instead of passing when there is an error

## How to test

1. `yarn install && yarn test:a11y` - Should pass with no errors once #2109 is merged, will only report the steplist errors for now, won't merge this PR until that PR is merged.

## Checklist


pwolfert (approved):
It only showed the one failing test as expected and failed the tests. Nice work 👍

PR #2122, opened 2022-09-09
[WNMGDS-1917] Fix missing bg color for variables that don't have a fill utility class

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1917

- Instead of using the fill utility, which doesn't exist for all colors, apply the background color style directly to the swatch elements

## How to test

1. Visit https://cmsgov.github.io/design-system/branch/pwolfert/fix-component-style-swatches/components/badge?theme=healthcare#styles
2. Look at the row for the `$badge__background-color--info` variable. It was white on master rather than blue. On this  branch it should be blue.

### Before
Screen Shot 2022-09-09 at 3 31 52 PM

## Checklist


forrestbaer (approved):
C'est magnifique

PR #2127, opened 2022-09-13
[WNMGDS-1911] Fix anchor styles overriding anchor-button styles when keyboard focusing

### LINKED ISSUE #2110: Anchor button focus state text color reverts to baseline dark color, does not have sufficient contrast
**Describe the bug**
Color of text on anchor buttons when in focus state (and not previously visited) does not retain the "hover" color value.

**To Reproduce**
1. In a new private browser window, go to: https://design.cms.gov/storybook/iframe.html?id=components-button--all-anchor-buttons&globals=theme:medicare&args=&viewMode=story
(note: use a private/incognito window so that `:visited` state does not apply to the links)
2. Use keyboard only navigation (tab / shift+tab).
3. Navigate to "Anchor button on-dark", and "Anchor button solid" examples
4. Note that button text changes to dark color when `<a>` element has focus (and the URL of the `href` has not been visited)

**Expected behavior**
The text `color` value should retain the same `--color-h` value that's applied when hovering over the element (or the element has focus _and_ is hovered over, or the link has been visited and has focus).

**Screenshots**
image

**Desktop:**
 - OS: iOS
 - Browsers: chrome, FF, safari
 - Version: latest builds

**Smartphone:**
_Not a major issue in touch environments unless keyboard/assistive devices are used._

**Additional context**
Anchor buttons are used quite frequently on the static (Drupal) Mgov site. We have an override in place for now to address this. It looks like the simplest solution would be to add the `color` property/value from [line 40](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/styles/components/_Button.scss#L40) in the `_Button.scss` file to the ruleset on [line 22](https://github.com/CMSgov/design-system/blob/master/packages/design-system/src/styles/components/_Button.scss#L22)? If this sounds good to you all I can submit a PR.

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1911

Other rules for button interaction states were setting the color property, but this particular rule didn't. This meant that while the local CSS variable in charge of the text color had the correct value, that value wasn't being applied by a rule that had higher specificity than the rule that applies focus styles for plain anchor elements. When focusing on the link buttons with a keyboard only, the plain anchor rule was winning out, which resulted in a text color that didn't match.

Fixes #2110 

## How to test

1. Start Storybook and view http://localhost:6006/?path=/story/components-button--all-anchor-buttons

## Checklist

### DIFF
--- a/packages/design-system/src/styles/components/_Button.scss
+++ b/packages/design-system/src/styles/components/_Button.scss
@@ -21,6 +21,7 @@
 
     &:focus {
       background-color: var(--backgroundColor, $button__background-color);
+      color: var(--color, $button__color);
       text-decoration: none;
     }
 


forrestbaer (approved):
Looks good 👍🏻

zarahzachz (approved):

PR #2128, opened 2022-09-13
[WNMGDS-1918] Add z-index fix to Drawer

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1918)

Reimplement 500 z-index on Drawer component. Using the Sass var/token that I accidentally forgot to delete the last time I refactored (thanks Past Sarah!).

## How to test

Run `yarn storybook` on this branch to see the z-index was applied


pwolfert (approved):
Tested it with a modified story where I applied relative positioning to some body elements. Was way broken before 👍

PR #2132, opened 2022-09-14
[NO-TICKET] Adapt the background color of inverse stories to the selected theme in Storybook

pwolfert:
## Summary

The previous way of adding a dark background to Storybook stories had a few issues:
1. The background color had to be determined at the time of starting storybook, based on a `STORYBOOK_DS` environment variable and did not adapt to design-system theme changes done within storybook
    - This could be considered a doc-site bug
3. The background color was hard-coded rather than using the styles we already have
4. Logic was duplicated for every story that used it and was verbose

This change addresses those problems. You no longer need to use `storybook:medicare` or `storybook:healthcare` for a dark-background story to show the right color.

## How to test

1. Run `yarn storybook`
3. Navigate to http://localhost:6006/?path=/story/components-button--variations-on-dark
4. Change the theme between core, healthcare, and medicare, and see the background color change appropriately

### DIFF
--- a/.storybook/preview-body.html
+++ b/.storybook/preview-body.html
@@ -1,3 +1,0 @@
-<script>
-  document.body.className += ' ds-base';
-</script>
--- a/.storybook/preview.js
+++ b/.storybook/preview.js
@@ -24,14 +24,6 @@ export const parameters = {
       date: /Date$/,
     },
   },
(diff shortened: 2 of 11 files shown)


forrestbaer (approved):
This is the way 👍🏻 ⚡

PR #2143, opened 2022-09-16
[no-ticket] Removing tsconfig settings which are no longer needed

forrestbaer:
## Summary

- Removes outdated `.tsconfig` settings which seem to no longer be needed

## How to test

1. `yarn type-check` - used to fail because of these, but it seems they are no longer needed, thoughts @pwolfert ?


pwolfert (approved):
O frabjous day! 🎉

PR #2144, opened 2022-09-16
[NO-TICKET] Refactor HelpDrawer's analytics code into a separate function

pwolfert:
## Summary

Was thinking about recent analytics conversations and looking at the code for it and decided to reduce the noise inside `HelpDrawer` by moving its analytics logic into a separate function.

## How to test

- `yarn test:unit`
- `yarn storybook` and navigate to the help drawer story and turn on analytics logging to console and open and close it

## Checklist


packages/design-system/src/components/HelpDrawer/useHelpDrawerAnalytics.ts
```diff
@@ -0,0 +1,54 @@
+import { EventCategory, sendLinkEvent, useAnalyticsContent } from '../analytics';
+import { HelpDrawerProps } from './HelpDrawer';
+import { useCallback } from 'react';
+import { helpDrawerSendsAnalytics } from '../flags';
+
+export default function useHelpDrawerAnalytics({
+  analytics,
+  analyticsLabelOverride,
+}: HelpDrawerProps) {
+  const sendHelpDrawerEvent = useCallback(
```
pwolfert, line 10: Now that I think about it, I don't need `useCallback`, because the function isn't being applied to any components in the subtree.

forrestbaer (approved):
Nice refactor 👍🏻

PR #2145, opened 2022-09-17
[NO-TICKET] Convert Alert to functional component

pwolfert:
## Summary

- Converted Alert to a functional component
- Created new `mergeRefs` and `useAutoFocus` utilities
    - There are other places in our codebase where we can use `mergeRefs` to avoid some messy and verbose code
    - We will need `useAutoFocus` when we're migrating components away from `FormControl`
- Moved analytics logic into a separate `useAlertAnalytics` hook that makes use of the newish `useAnalyticsContent` hook
    - This would simplify the logic, but I added some glue code that makes it match the output of the old logic that gathered analytics content. I wanted to avoid making any behavioral changes with this unticketed refactor, but we should consider adding this to our upcoming breaking-change release
- Added a few more checks to the unit tests

## How to test

1. `yarn storybook`
2. Navigate to http://localhost:6006/?path=/story/components-alert--error-alert&args=autoFocus:false&globals=analytics:on
3. Open your JavaScript console to make sure analytics still fire
4. Try setting the `autoFocus` prop in the story controls. It should focus on the last one in the story (because we have three alerts per story and you can only focus on one)

## Checklist


packages/design-system/src/components/Alert/Alert.tsx
```diff
@@ -172,69 +143,34 @@ export class Alert extends React.PureComponent<
...
+      ref={mergeRefs([alertRef, focusRef])}
+      tabIndex={alertRef || autoFocus ? -1 : null}
+      role={role}
+      aria-labelledby={heading ? headingId : a11yLabelId}
+      {...alertProps}
+    >
+      {getIcon()}
+      <div className="ds-c-alert__body" id={headingId} ref={bodyRef}>
+        {heading ? (
+          <div className="ds-c-alert__header ds-c-alert__heading">
```
pwolfert, line 158: If we want to lean into analytics starting to show the a11y labels consistently, the `headingRef` should actually go on this element.

forrestbaer:
Looks good, analytics showing, however looks like alerts aren't getting their specific `ds-c-alert--error`, etc classes

image

pwolfert:
@forrestbaer, I'm not able to reproduce the issue in your screenshot above. Can we look at it together tomorrow?

forrestbaer (approved):
Working for me with your recent VR testing branch now, not sure what it was before but looking good, wasn't able to reproduce that issue. 👍🏻

PR #2147, opened 2022-09-19
[WNMGDS-1924] RTL conversion of HC tests & enzyme removal

forrestbaer:
## Summary

- Updates all ds-healthcare-gov tests to use RTL
- Removes enzyme from the cmsds! 🎉 

## How to test

1. `yarn install && yarn test:unit`

## Checklist


packages/ds-healthcare-gov/src/components/Header/ActionMenu.test.tsx
```diff
@@ -89,15 +90,15 @@ describe('ActionMenu', function () {
...
+      expect(mock).toHaveBeenCalled();
     });
 
     it('sends analytics event when logged-out menu opened', () => {
       renderForAnalytics({
         links: [{ label: 'label', href: 'href' }],
       });
       fireEvent.click(screen.getByRole('button'));
-      expect(window.utag.link.mock.calls[0][0]).toMatchSnapshot();
+      expect(mock).toHaveBeenCalled();
```
pwolfert, line 101: Why were these snapshots removed?

pwolfert (commented):
This is exciting!

I didn't look at all the snapshot changes for `packages/ds-healthcare-gov/src/components/Footer/__snapshots__/InlineLinkLists.test.tsx.snap`, which could be an indication that they're too verbose more than that I'm lazy.

One concern below about the analytics tests

packages/ds-healthcare-gov/src/components/Header/ActionMenu.test.tsx
forrestbaer: oh good point I forgot about these, when I was working on them I had trouble finding these items within the rtl way of doing things.

forrestbaer: Ok @pwolfert these have been re-added

pwolfert: It looks like the snapshots that were added back in have all calls for all test instances, so they're really big and gnarly. The mock on `window.utag.link` doesn't reset between tests.

forrestbaer: Very true, took me a minute to figure this out, managed to grab only the needed data, updated!

forrestbaer: Whoops but now it's grabbing the wrong one.. looking into it.

forrestbaer: Ok actually fixed these, sorry for the noise.

pwolfert (approved):
LGTM

PR #2148, opened 2022-09-19
[WNMGDS-1915] Linting fixes

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1915)

Reduce the number of our linting errors/warnings, evaluate linting rules we follow.

**Initial warnings:**
CSS: 20 warnings
JS: 209 problems (0 errors, 209 warnings)

**Remaining warnings:**
CSS: 19 warnings
JS: 74 problems (0 errors, 74 warnings)

CSS issues are with Medicare using `@extend` and `!important` to overwrite styles. I think these overwrites were contributed by outside teams, so I think it'd be a good idea to pair with a designer to confirm these overwrites before folding them into the design system proper.


packages/design-system/src/components/IdleTimeout/IdleTimeout.test.tsx
```diff
@@ -189,15 +189,15 @@ describe('Idle Timeout', () => {
     const { getByRole } = renderIdleTimeout({ formatMessage, timeToWarning: 2 });
     showWarning(MOCK_START_TIME + 2 * 60000);
     const dialogBodyText = getByRole('main');
-    expect(dialogBodyText.firstChild.textContent).toEqual('Your session will end in 3.');
+    expect(dialogBodyText.firstChild.textContent).toEqual('Your session will end in 3.'); // eslint-disable-line
```
forrestbaer, line 192: I think all of these linting disables should removed personally, I would rather have the linter keep us honest and have us resolve the issues the 'correct' way sometime in the future, than brush these under the rug until who-knows-when..

zarahzachz: Reposting here from Slack, but we discovered some of our linting rules (no node queries and containers) weren't absolute no-nos when working with RTL, so I've updated our ESLint config to reflect that. As a result, the linting disable comments have been removed.

forrestbaer (commented):
I pushed a couple small updates, all looks good to me, nice work cleaning up!

forrestbaer (approved):
💯 🧹

packages/design-system/src/components/Table/TableCell.tsx
```diff
@@ -72,7 +72,6 @@ export const TableCell: React.FC<
   id,
   scope,
   stackedTitle,
-  stackedClassName,
```
pwolfert, line 75: Is this a bug fix? Did this need to be applied to the cell element and wasn't until now?

packages/docs/src/components/ColorSwatchList.tsx
```diff
@@ -58,14 +59,15 @@ const ColorSwatchList = ({ backgroundClass, colorNames, preface, theme }: ColorS
 
     setColorList(updatedColorList);
   }, [
+    colorList,
```
pwolfert, line 62: This change is causing an endless render loop. When inspecting this, the DOM is constantly being updated.

pwolfert (changes requested):

packages/docs/src/components/ColorSwatchList.tsx
pwolfert: It seems to actually be the combination of the uniqueId change below and this line

pwolfert: And it makes sense that this dependency would cause an endless loop because it's saying that if `colorList` changes to call the function, and the function always updates `colorList` at the end.

packages/design-system/src/components/Table/TableCell.tsx
zarahzachz: Linting flagged this because it was declared and not used in the component. I didn't see it being used in the component or elsewhere in the repo so I assumed it was safe to remove.

pwolfert: By removing it from this destructuring assignment, we're actually _adding_ it to the `tableCellProps` object, so this will now be applied to the cell element. I think you found a bug if it's not being used. We should probably remove this from this PR, though, because it does change behavior.

zarahzachz: Ohhhh OK

packages/docs/src/components/ColorSwatchList.tsx
```diff
@@ -65,7 +66,7 @@ const ColorSwatchList = ({ backgroundClass, colorNames, preface, theme }: ColorS
   return (
     <div className="c-swatch-list ds-u-border--1 ds-u-padding--2">
       {colorList.map(({ name, hex }, index) => (
-        <article className="ds-u-margin-bottom--1 c-swatch" key={`${name}-${index}`}>
+        <article className="ds-u-margin-bottom--1 c-swatch" key={uniqueId(`${name}-`)}>
```
pwolfert, line 69: Sorry, one more thing. What was the rule that spawned this change? It seems unnecessary to me to use `uniqueId` here because this is a key that is used internally by React. I think there's a rule somewhere about not using index, but we have no perfect guarantees that `name` and `hex` will produce unique results...however 🤔, if the combination of `name` and `hex` are not unique, then there is no reason to render a second instance of one...so maybe there's no harm in using that as a key. All that to say, could we change it to
```suggestion
        <article className="ds-u-margin-bottom--1 c-swatch" key={`${name}-${hex}`}>
```

pwolfert: Found the rule: https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/no-array-index-key.md, which links to some React docs. After reading that, I stand by my suggestion

pwolfert (approved):
One suggestion above, but I don't need to block it anymore because you've fixed the bugs

pwolfert:
Looks like CI tests are failing due to two TypeScript errors

PR #2157, opened 2022-09-21
[WNMGDS-1914] Fix allowing stories to be merged without any VRT refs

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1914

This is a follow-up for the new visual regression tests. After that was merged, I found that https://github.com/CMSgov/design-system/pull/2135 was able to be merged without adding any screenshots for its new story, which means that the full (non-smoke) visual regression tests are failing on `main`. The problem is that the smoke tests only test the core theme, and we also ignore tests for child design systems when we're testing a theme that doesn't match it. It's that combination that allowed the new story to slip through the cracks. This should fix that.

- Include healthcare-only and medicare-only stories in smoke tests
    - Smoke tests included 236 tests before, and now they include 272 tests.
- Add reference images for two new healthcare stories that slipped through the cracks
- Fixed our `test:browser:update` command, which had a typo that broke it

## How to test

1. Run `docker run --rm --network host --env SMOKE=true -v $(pwd):/work/ -w /work/ mcr.microsoft.com/playwright:v1.25.0-focal yarn playwright test --config tests/browser/playwright.config.ts --list` and see that the list includes some healthcare and medicare stories.

## Checklist


forrestbaer (approved):
💯

PR #2159, opened 2022-09-22
[NO-TICKET] Organize doc-site components into folders

pwolfert:
## Summary

Now that we have so many components in our Gatsby site, I was having trouble navigating through them. I've created three folders that most of the components can be sorted into. Is this helpful or not?

I also fixed a bad link on the index page.

## Checklist


packages/docs/content/not-in-sidebar/contact.mdx
```diff
@@ -3,10 +3,10 @@ title: Contact us
...
 ---
 
-import GithubIcon from '../../src/components/GithubIcon';
-import JiraIcon from '../../src/components/JiraIcon';
-import SlackIcon from '../../src/components/SlackIcon';
-import EmailIcon from '../../src/components/EmailIcon';
+import GithubIcon from '../../src/components/Icons/GithubIcon';
+import JiraIcon from '../../src/components/Icons/JiraIcon';
+import SlackIcon from '../../src/components/Icons/SlackIcon';
+import EmailIcon from '../../src/components/Icons/EmailIcon';
```
pwolfert, line 9: I just realized that all my folders are capitalized even though the capitalization is only meaningful for `Layout` (because that folder's index points to a component of the same name). Should I instead lowercase them all and have the `Layout` component be referenced like `components/layout/Layout`?

forrestbaer: Yeah I think that's a good idea 👍🏻

forrestbaer (approved):
Seems fine to me, no dissent 👍🏻

PR #2166, opened 2022-09-23
[NO-TICKET] Test forced-colors mode for visual regressions

pwolfert:
## Summary

I was working on our component maturity checklist and thinking about how we verify that our components work in forced-colors mode and had a realization. With our new visual regression tests, we...
1. Have a pretty thorough set of testing targets that are no longer constrained to needing to finish within a reasonable amount of time for CI tests (because of our nightly builds)
2. Now have a lot of control over those testing targets

So why couldn't we test forced-colors mode as well?

This also removes some story tests that aren't necessary.

## How to test

1. `yarn test:browser --list` if you want to see that it registers all the tests under the new target
2. `yarn test:browser` to run it locally if you want. This takes a long time. We can actually narrow the tests we run with a `--project` argument, and I'm adding documentation for that
3. I know it's a lot of work, but please take a look at the new screenshots and try to spot problems

## Checklist


pwolfert:
@brentschneider, I'm tagging you in case you have time to look at the images that are added and try to spot defects with forced-colors mode

pwolfert:
Looked the first 50 images or so. [This clear button has funny margins](https://github.com/CMSgov/design-system/blob/3dae2942796482ad4d23cbaae1bd005b6f70399d/tests/browser/story-screenshots.test.ts-snapshots/components-autocomplete--custom-markup-core-chromium-forced-colors-linux.png), but it's not terrible; it's to be expected.

pwolfert:
I'm finding it much more reliable to view the files on my computer with a file search and use the arrow keys to cycle through them. Here's [another one that looks kind of odd](https://github.com/CMSgov/design-system/blob/3dae2942796482ad4d23cbaae1bd005b6f70399d/tests/browser/story-screenshots.test.ts-snapshots/healthcare-footer--basic-footer-healthcare-chromium-forced-colors-linux.png) (whitehouse logo is all black).

pwolfert:
Went through all the images

zarahzachz:
I love including HCM in this scan. 

I'm reviewing images and IDing some that might need evaluated in a later ticket:
- Healthcare accordion heading: font-weight is very light, making it hard to to see clearly against dark background
- Alert (all): 
    - No distinction between normal and lightweight styles
    - Mobile examples have odd button positioning (might just be a Storybook implementation issue)
    - Why do alerts with buttons have a constrained width while alerts without buttons appear to span width of page? I think both are desktop images.
- Autocomplete: 
    - Clear search button has border/padding issue
    - Some states might not benefit from vis-reg tests (loading and no results) - the important part for these is the option overlay that pops up after a user has entered something and we aren't capturing that appearance
- Badge: Increasing font-weight will make it more legible
- Buttons: 
    - Need another pass, no differentiator between styles and needs tweaked to support filled/ghost styles
    - Vis-reg tests appear to only take snapshots of visible window instead of whole page, so some button examples are cut off
    - Disabled link buttons and button buttons need a closer look
    - I don't know why mobile Chrome and Safari place their buttons at different locations on the window (Safari is about 100px lower than Chrome example? Maybe accounting for the notch?)
- Dialog modal: No modal images captured because it requires user to trigger modal from button before appearing
- Drawer/HelpDrawer: 
    - Need to have borders used evaluated (there's a thicc boi on the left-border and I don't think it's supposed to look like that)
    - Footer might need something to separate it visually from content when it's sticky
    - I thought we removed the focus on heading since headings aren't focusable?
- Icons: Vis-reg tests cut off window for screenshots so I can't fully see results
- Timeout: 
    - Close button needs border removed
    - Mgov close button keeps "Close" text and border
- MultiInputDateField/SingleInputDateField: Error state might need a look (maybe?)
- VerticalNav: Safari (not HCM) has weird border issue on currentPage (both mobile and desktop)
- Healthcare footer: Whitehouse image not visible
- Medicare Navbar: Shows extra text in HCM 
- Mgov footer: Privacy settings button has border
- Mgov stars: All stars filled when they shouldn't be

zarahzachz (approved):
This looks good. I added a comment for outages I saw.

PR #2175, opened 2022-09-28
[WNMGDS-1792] Dialog to native dialog

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1792

- Changes the `Dialog` component implementation to use the native `<dialog>` element (via our internal `NativeDialog` component) instead of using `react-aria-modal`
- Removes all the props that were specific to `react-aria-modal` that we don't need and generally simplifies the props
- Props that were removed:
    - `applicationNode`
    - `dialogClass`
    - `dialogId`
    - `dialogStyle`
    - `escapeExits`
    - `focusDialog`
    - `focusTrapOptions`
    - `focusTrapPaused`
    - `includeDefaultStyles`
    - `initialFocus`
    - `mounted`
    - `scrollDisabled`
    - `titleId`
    - `titleText`
    - `underlayClass`
    - `underlayClickExits`
    - `underlayColor`
    - `underlayStyle`
    - `verticallyCenter`
- Props that were added:
    - `backdropClickExists`
    - `id`
    - All valid attributes of [the `<dialog>` element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog)
- Removes `react-aria-modal` dependency

## How to test

1. Use Storybook to manually test Dialog and the components that use it (IdleTimeout, and external link pattern story)

https://cmsgov.github.io/design-system/branch/pwolfert/dialog-to-native-dialog

## Checklist


packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -335,20 +178,17 @@ export const Dialog = (props: DialogProps) => {
...
   );
 };
 
 Dialog.defaultProps = {
   closeButtonVariation: 'ghost',
   closeIcon: <CloseIcon />,
-  escapeExits: true,
-  escapeExitDisabled: false,
-  underlayClickExits: false,
+  scrollDisabled: true,
```
zarahzachz, line 191: We should remove this prop and bake the functionality into this component.

If a modal is open, there's no reason an application would want to allow users to continue to scroll the background/body content. 

A true modal should prevent interaction with non-modal things, and the dark backdrop that appears when the modal is open obscures any readable text anyway.

pwolfert: That seems totally reasonable to me 👍

pwolfert:
Right now the error I'm getting on the visual regression tests is due to the fact that initial focus for dialogs is now the close button, which I'll need to fix before I take this PR out of draft state.

packages/design-system/package.json
```diff
@@ -41,7 +41,6 @@
     "lodash": "^4.17.21",
     "ora": "^6.1.2",
     "prop-types": "^15.7.2",
-    "react-aria-modal": "^2.11.1",
```
forrestbaer, line 44: 🎉

packages/design-system/src/components/Dialog/Dialog.stories.jsx
```diff
@@ -40,21 +41,20 @@ export default {
   },
 };
 
-export const DialogExample = ({ ...args }) => {
-  const [shown, setShown] = useState();
-  const showModal = () => setShown(true);
-  const hideModal = () => setShown(false);
+export const DialogExample = ({ isOpen: _isOpen, ...args }) => {
+  const [{ dialogOpen }, updateArgs] = useArgs();
```
forrestbaer, line 45: I got this warning in the console:
```
React does not recognize the `dialogOpen` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `dialogopen` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

pwolfert: Thanks for bringing that to my attention. Getting this args stuff to work right has been a challenge. I've got the fix, though, and I'm going to push it up after my local visual regression tests finish updating.

forrestbaer (approved):
Working great for me now!

SparrowSpit:
Focus issue was my mistake. A11y testing completed and passed.

pwolfert:
Thank you, @SparrowSpit!

pwolfert:
@forrestbaer and @zarahzachz, I realized that I hadn't filled out the description of things I changed, and when I was filling it out to make our job easier at release time, I also realized that the prop types didn't include all the attributes of `<dialog>`, so I updated the types in https://github.com/CMSgov/design-system/pull/2175/commits/17db2d010c7fd805df0c8897dea33542f669f102

PR #2178, opened 2022-09-30
[WNMGDS-312] Update margin and padding examples

line47:
## Summary

- Add pixel values to margin and padding examples
- Update examples design to a more conventional format like found on the colors page (example, class, value) 
- Added control to render padding and margin examples
- Added styles for margin and padding rendering

## How to test

1. run the doc site with `yarn start` 
2. the visit the margin page and padding page to see updated examples with pixel values 

## Checklist


packages/docs/content/utilities/margin.mdx
```diff
@@ -2,378 +2,234 @@
...
+    '0px',
+    '4px',
+    '8px',
+    '16px',
+    '24px',
+    '32px',
+    '40px',
+    '48px',
+    '56px',
+    'auto',
```
pwolfert, line 34: I can see why we'd want to show that this is a possible value, but the actual example doesn't seem as helpful as it could be. This would kind of ruin the whole genericness of this component and its values, but what if this one's container element had some width on it to show that `--auto` will center it horizontally within its container?

pwolfert: This and the fact that we have to duplicate all the spacer information in the mdx content makes me think this component should be specialized to spacing rather than generic. We don't need it to dynamically figure out the pixels, though, since spacing is consistent between themes right now; it can still be simple.

pwolfert: Oh, I didn't realize that `PaddingSpacingExample` existed before. If we move back to something like that but refine the design to show the pixel values and expand it to work for margins as well, I think that would work.

pwolfert (approved):
LGTM, but I made changes, so I'd like a second opinion, @zarahzachz or @forrestbaer

packages/docs/content/utilities/margin.mdx
line47: I hacked together some auto margin ideas last night and am pushing those up to the branch. ATM they are embedded examples.  @pwolfert if you see a way to improve on this or incorporate into the spacing component go for it!

zarahzachz (approved):
LGTM

pwolfert:
Hey, your new example is slick 👍

PR #2190, opened 2022-10-10
[WNMGDS-1935] export css instead of sass to dist, utilize scss internally

forrestbaer:
## Summary

- Updated the `Choice` component to use css vars instead of sass variables
- Updated build tasks and storybook setup code to work in this scenario

## How to test

1. `yarn install && yarn build` - evaluate `packages/design-system/dist/css/components/index.css` to css references, look in `packages/design-system/dist/css/settings/variables` to see variable code. (in production we would want to move this file probably to the root)
2. `yarn storybook` - evaluate the `Choice` stories, using css variables

## Notes on this approach / feedback requested

- We can still use `scss.math` operations on these, but they're done at compile time, so if these variables change afterwards, those transforms/calculations will not be applied
- If we chose this direction, we could cleanup the `@extends` and perhaps a lot of our `scss.math` operations
- It's functional, but in my opinion a little ugly for us as developers. I personally find plain css easier to read and reason about. All the nesting can lead to levels of specificity that we may not have intended.
- Overall css vars are really great for developers downstream as far as flexibility with the codebase goes. And the ability to include them separately (as it's configured here) is a nice option.  We could also fold these variables into the main css after the sass compilation is done if we wanted to bundle a single css file.


packages/ds-healthcare-gov/src/styles/components/_Inset.scss
```diff
@@ -6,10 +6,10 @@
 
 // Align the inset's border with the checkbox/radio field when used as part
 // of the Exposed Within pattern
-.ds-c-choice + label + .hc-c-inset {
-  margin-left: calc(calc(#{$choice-size} * 0.5) - calc(#{$inset__border-width} * 0.5));
-}
+// .ds-c-choice + label + .hc-c-inset {
```
forrestbaer, line 9: Just commented this out so storybook wouldn't report trouble for this example.

PR #2191, opened 2022-10-11
[WNMGDS-925] allowing passthrough of aria props for choice components

forrestbaer:
## Summary

- Was a big discovery on this one, review notes and slack threads [here](https://jira.cms.gov/browse/MCT-5112?focusedCommentId=2648029&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-2648029) and [here](https://jira.cms.gov/browse/WNMGDS-925)
- Basically allowing passthrough of `aria-live`, `aria-relevant` and `aria-atomic` to `Choice` components to allow those aria props to be turned off / changed as needed when there is child content.
- Although the ticket didn't require the `aria-relevant` and `aria-atomic` props to be editable, I figured we might as well?

## How to test

1. `yarn install && yarn build`, modify `packages/design-system/src/components/ChoiceList/ChoiceList.stories.jsx` to add add
```
'aria-live': 'off',
'aria-relevant': 'text',
'aria-atomic': 'true',
```
props to one or all of the `choices` in `ChoiceChildren` example at the bottom.
2. `yarn storybook`, validate aria attributes have been modified
3. `yarn test:unit`, validate all tests pass

## Checklist

### DIFF
--- a/packages/design-system/src/components/ChoiceList/Choice.test.tsx
+++ b/packages/design-system/src/components/ChoiceList/Choice.test.tsx
@@ -50,6 +50,14 @@ describe('Choice', () => {
     expect(el).toBeChecked();
   });
 
+  it('allows for modification of aria attributes', () => {
+    renderChoice({ 'aria-live': 'off', 'aria-relevant': 'none', 'aria-atomic': false });
+    const input = screen.getByRole('checkbox');
+    expect(input).toHaveAttribute('aria-live', 'off');
+    expect(input).toHaveAttribute('aria-relevant', 'none');
+    expect(input).toHaveAttribute('aria-atomic', 'false');
+  });
+
   it('is defaultChecked', () => {
     renderChoice({ defaultChecked: true });
     const el = screen.getByRole('checkbox');
--- a/packages/design-system/src/components/ChoiceList/Choice.test.tsx
+++ b/packages/design-system/src/components/ChoiceList/Choice.test.tsx
@@ -50,14 +50,6 @@ describe('Choice', () => {
     expect(el).toBeChecked();
   });
 
-  it('allows for modification of aria attributes', () => {
-    renderChoice({ 'aria-live': 'off', 'aria-relevant': 'none', 'aria-atomic': false });
-    const input = screen.getByRole('checkbox');
-    expect(input).toHaveAttribute('aria-live', 'off');
-    expect(input).toHaveAttribute('aria-relevant', 'none');
-    expect(input).toHaveAttribute('aria-atomic', 'false');
-  });
-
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/ChoiceList/Choice.test.tsx
```diff
@@ -50,6 +50,14 @@ describe('Choice', () => {
     expect(el).toBeChecked();
   });
 
+  it('allows for modification of aria attributes', () => {
+    renderChoice({ 'aria-live': 'off', 'aria-relevant': 'none', 'aria-atomic': false });
+    const input = screen.getByRole('checkbox');
+    expect(input).toHaveAttribute('aria-live', 'off');
+    expect(input).toHaveAttribute('aria-relevant', 'none');
+    expect(input).toHaveAttribute('aria-atomic', 'false');
```
pwolfert, line 58: I actually don't think we want them on both the input and the wrapper. They don't get stripped out of the `inputProps` object, so these custom aria attributes end up on both the wrapper and the input.

forrestbaer: Sorry confused by this, I see the props on the input only when I pass them, I don't see them on the wrapper.

packages/design-system/src/components/ChoiceList/ChoiceList.test.tsx
```diff
@@ -81,6 +82,21 @@ describe('ChoiceList', () => {
...
+        'aria-atomic': 'true',
+      });
+      choices[0].checked = true;
+      const { container } = renderChoiceList({ choices });
+      const wrapper = container.querySelector(':nth-child(3)');
+      expect(wrapper).toHaveAttribute('aria-live', 'off');
+      expect(wrapper).toHaveAttribute('aria-relevant', 'text');
+      expect(wrapper).toHaveAttribute('aria-atomic', 'true');
+      expect(wrapper).toMatchSnapshot();
+    });
```
pwolfert, line 98: We should probably also test the case where it would have applied them automatically (when `checkedChildren` is defined)

pwolfert (approved):
It works! LGTM, but I'm suggesting one more unit test case

packages/design-system/src/components/ChoiceList/ChoiceList.test.tsx
forrestbaer: Oh, yes good point, I thought that's what checking the first choice did mistakenly, gotta add children!

forrestbaer: Done.

PR #2198, opened 2022-10-17
[WNMGDS-1855] Add box-sizing rule. Remove focus sizing styles - unneeded.

zarahzachz:
## Summary

This work folds in a local style from the Windowshop/Enroll app where small checkbox icons weren't displaying correctly. 

It turned out that app's CSS reseat included a `box-sizing: initial` for all elements, (`*, *::before, *::after { box-sizing: initial; }`) that was breaking the layout of our CSS. They fixed it by adding a local `box-sizing: content-box` to the checkbox styles. I migrated this style into our design system.

While testing this change, I discovered another bug:
Screen Shot 2022-10-17 at 11 05 43 AM

Small checkboxes increase in size on focus. After playing around with the style causing this, I was able to determine we could safely remove the offending CSS. Tested in Chrome, Firefox, and Safari and removing this style had no impact on any checkboxes in our design system.

## How to test

Run Storybook locally and navigate to: http://localhost:6006/?path=/story/components-choicelist--small-option&args=type:checkbox

Clicking new checkboxes (moving focus around) should persist the small sizing.

## Checklist

### DIFF
--- a/packages/design-system/src/styles/components/_Choice.scss
+++ b/packages/design-system/src/styles/components/_Choice.scss
@@ -79,6 +79,11 @@ label,
     width: $choice__size;
   }
 
+  &::before,
+  &::after {
+    box-sizing: content-box;
+  }
+
   // Error
   @if $ds-include-choice-error-highlight {
     &.ds-c-choice--error {
@@ -139,12 +144,6 @@ label,
   &:checked {
     &::after {
       opacity: 1;
-    }
-
-    &:focus::before {
-      // have to do a mix of calc & sass interpolation in order to accommodate values being css vars in internal tools
-      height: calc(#{$choice__size} - (#{$choice__border-width} * 2));
-      width: calc(#{$choice__size} - (#{$choice__border-width} * 2));
     }
   }
 }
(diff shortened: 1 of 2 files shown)


packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -140,12 +145,6 @@ label,
     &::after {
       opacity: 1;
     }
-
```
forrestbaer, line 143: Yay removed a calc! 🎉

forrestbaer (commented):
Looks good, were there any changes in visual regression?

zarahzachz:
> Looks good, were there any changes in visual regression?

I ran the command locally and didn't find any. Obviously, there were visual changes to removing the focus calc sizing, but we don't have vis regression tests for interactive states like focus so it didn't flag it.

forrestbaer (approved):

packages/design-system/src/styles/components/_Choice.scss
zarahzachz: Oh, looks like Mgov has overrides for this style. Pushing up a fix addressing that too.

pwolfert: Hey, it looks nice!

PR #2203, opened 2022-10-18
[WIP] Refactoring the Dropdown to use useFormLabel and be functional

pwolfert:
I also saw some opportunities to have useFormLabel handle aria-describedby and aria-invalid more consistently.


pwolfert:
Didn't mean to taket his out of draft. Was looking at the wrong tab.

PR #2206, opened 2022-10-19
[WNMGDS-1950] Contribute to a CSS delinting branch

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1950

This is to merge some more work into Forrest's delinting branch that resolves errors back to the baseline we have on `main`.

## How to test

`yarn lint:stylelint`

## Checklist

### DIFF
--- a/.stylelintrc.js
+++ b/.stylelintrc.js
@@ -20,7 +20,6 @@ module.exports = {
     // Turning this off in favor of 'at-rule-allowed-list'
     'at-rule-no-unknown': null,
     'at-rule-allowed-list': [
-      // Are we intentionally leaving out 'include'?
       [
         'function',
         'if',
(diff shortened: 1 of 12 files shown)


.stylelintrc.js
```diff
@@ -17,34 +17,37 @@ module.exports = {
...
-    'selector-pseudo-class-case': 'lower',
-    'order/properties-alphabetical-order': true,
-    'no-empty-first-line': true,
-    linebreaks: 'unix',
     'at-rule-allowed-list': [
-      ['function', 'if', 'use', 'media', 'font-face', 'import', 'extend'],
-      {
-        severity: 'warning',
-      },
+      // Are we intentionally leaving out 'include'?
```
forrestbaer, line 23: Yeah warnings are good for includes, gotta move to use eventually

forrestbaer (approved):
🧹

.stylelintrc.js
pwolfert: Do you mean import? I can remove import from this list so we get warnings for those, but maybe we should wait until we're ready to migrate to `@use` because it's just noise until we're ready.

PR #2208, opened 2022-10-20
[WNMGDS-1940] Changing Choice wrapper max-width

forrestbaer:
## Summary

- Updated choice wrapper max width to use `$measure-wide` which is `80ex` (https://design.cms.gov/utilities/text/measure)

## Feedback requested

Should this also apply to `Autocomplete`'s `ds-c-autocomplete` and `TextField`'s `ds-c-field` or do those remain constrained by the `$form__max-width` of `470px`?

## How to test

## Checklist

### DIFF
--- a/packages/design-system/src/styles/components/_Choice.scss
+++ b/packages/design-system/src/styles/components/_Choice.scss
@@ -11,7 +11,7 @@
   display: flex;
   gap: 0.5rem;
   margin: 0.5rem 0;
-  max-width: $form__max-width;
+  max-width: $measure-wide;
 }
 
 // Styles for label


pwolfert:
- I think `Autocomplete` and `TextField` using `form__max-width` are correct. I think Choice is an exception because its children can contain other kinds of content besides the label of the checkbox input (like alerts and special boxes). The form elements themselves, like textboxes, shouldn't expand beyond the form max width.
- `SingleInputDateField` is tuned to have a width that is enough to contain date input but not be overly wide
- `UsaBanner` is special because it's supposed to conform to USWDS
- `FormLabel` being $form__max-width makes sense to me.

CC @line47, do you agree with any of the above? These bullet points are addressing questions in the PR description

line47:
@pwolfert that all makes sense to me and I agree

pwolfert (approved):
LGTM 👍

PR #2209, opened 2022-10-20
[WNMGDS-1952] Adding underline attribute when a ghost button is an href

forrestbaer:
## Summary

- `Ghost` buttons were not previously getting an underline when they were created with an `href` prop.
- Added style override for ghost buttons which are links to ensure they get that underline.

## How to test

1. `yarn install && yarn build && yarn storybook` - go to a button example and change it to a ghost button, add an href, gasp in wonder that it is now underlined

## Checklist

- [x] Full Visual regression ran


zarahzachz (changes requested):
GASP! Anchor ghosts have underlines! A few comments:

1. Disabled button buttons aren't underlined, but anchor buttons are - which should it be?
2. Hovering over an anchor button results in a thinner underline - maybe 1px? It should persist the width from the non-hover state

forrestbaer:
> GASP! Anchor ghosts have underlines! A few comments:
> 
> 1. Disabled button buttons aren't underlined, but anchor buttons are - which should it be?
> 2. Hovering over an anchor button results in a thinner underline - maybe 1px? It should persist the width from the non-hover state

Thanks!I fixed the underline width, regarding #1 Based on the sketch files I looked at they should be underlined.

forrestbaer:
It looks like the snapshots for pagination changed slightly with this update, since this fixes ghost buttons which are `href`'s, pagination has changed to underline the page numbers. @line47 should I overwrite the pagination ghost buttons to remove the underline or should they remain underlined?

line47:
Thanks @forrestbaer - Keep the pagination links without the underline

zarahzachz (changes requested):
The underline width on anchor buttons looks great, but noticing 2 issues still:
1. Disabled ghost button buttons don't have an underline
2. The underline of ghost anchor buttons on dark backgrounds takes on a blue color - it should probably be the same color as the text and it's not

pwolfert:
I'm only now realizing that our visual regression tests don't take full-page screenshots 🤦‍♂️

pwolfert:
The fix looks great!

pwolfert (approved):

forrestbaer:
@zarahzachz it looks like I can't merge until your 'requested changes' are resolved. 🙇🏻

zarahzachz (changes requested):
Sorry for the delay! Just took a look and there's 1 last request - the ghost anchor button active state removes the underline. The underline shouldn't be removed when a user clicks the button.

PR #2210, opened 2022-10-20
[WNMGDS-1246] updating focus-trap-react dep and tooltip component

forrestbaer:
## Summary

- Tooltip is the only component still using `react-focus-trap`
- This upgrades `@cmsgov/design-system` dependency for `react-focus-trap` and it's version of `prop-types` as required to resolve dependency warnings after upgrading to React 17
- Added a `fallbackFocus` option to react focus trap as required by the updated API. The functionality remains the same as far as I can see in my testing.

## How to test

1. `yarn install && yarn build && yarn test:unit`

## Checklist


pwolfert (approved):
I can't break it! 🎉

PR #2211, opened 2022-10-20
[no-ticket] addition of yarn clean npm scripts

forrestbaer:
## Summary

- Adds a simple way to clean up `node_modules` folders recursively and doc site `.cache` folder when you run into a situation where that's necessary

## How to test

1. `yarn clean && yarn install` 🧼 🚿


pwolfert (approved):
Nice idea 👍 handy!

PR #2218, opened 2022-10-25
[WNMGDS-2008] Change analytics event properties to support medicare.gov

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2008

- Changed the names of the event object's properties
    - Medicare.gov analytics expects a common set of event properties that look like this:
        - `event_name: string;`
        - `event_type: string;`
        - `event_category: string;`
        - `event_action: string;`
        - `event_label: string;`
    - whereas Healthcare.gov expected these:
        - `event_name: string;`
        - ~`ga_eventType: string;`~ (but we're removing this after talking to the analytics team)
        - `ga_eventCategory: string;`
        - `ga_eventAction: string;`
        - `ga_eventLabel: string;`
    - When we call our common analytics function, we don't want to have to provide duplicate values in every place that it's called, so we duplicate the properties inside our common analytics function instead. I picked the property names that looked more normalized as the ones our function expects.
- Remove the default `event_type` value of "cmsds", which isn't required by the analytics team anymore
- `event_category` had some cross-over with `event_type` based on the values we were using for them, and I've added some type definitions and comments that starts to detangle them (based on my conversations with the analytics team)
- Move button analytics code into a separate hook
- This includes another change from the same team that requested this update, and that is to allow the analytics function to be overridable or configurable with a custom function.

## How to test

Scrutinize the unit tests. They should be keeping us from regressing but needed to change to reflect the new properties we've added (duplicating data under different property names).

## Checklist


forrestbaer:
I think the analytics override function is useful, makes things more extensible for future use. I like thee changes!

pwolfert:
@forrestbaer, I've been thinking about this a lot too, and I don't think the ship is too far out to sea. I think this change (providing an override option) is a step in the direction of moving it back into apps to implement. If we want to push for that, we now have a release that both continues to provide the built-in implementation but also make space for them to provide their own implementation.

PR #2221, opened 2022-10-25
[NO-TICKET] Clean up examples

pwolfert:
## Summary

- Updated the `create-react-app-typescript` example to match the latest template from the CRA project and to use the latest version of the design system
- Removed the plain `create-react-app` example because there is too much overlap with the `create-react-app-typescript` example to justify maintaining both
- Removed the `webpack-demo` example because it's a lot of code that gets out of date quickly, and the other examples cover the setup well enough. We're not a tutorial on how to set up Webpack.
- Cleaned up the `html-css` example and have it us the CDN
- **Question to @forrestbaer:** What if the `cdn` example referenced the assets straight from the CDN instead of with a local build and copy? (See my updates to the `html-css` example)

## How to test

`cd` into each of the examples and make sure you can use the README to get them each to run

## Checklist


examples/html-css/index.html
```diff
@@ -5,20 +5,17 @@
     <title>Design System</title>
     <!-- In a real project this will contain a href link to your local CSS file, 
       while in this case we are using a CDN as an example -->
-    <link
-      rel="stylesheet"
-      href="https://unpkg.com/@cmsgov/design-system@2.0.0/dist/css/index.css"
-    />
+    <link rel="stylesheet" href="https://design.cms.gov/cdn/design-system/4.0.5/css/index.css" />
```
forrestbaer, line 8: Maybe we should add a `latest` symlink or something during CDN publish, just a thought. To be a future ticket perhaps.

forrestbaer (approved):
Looks great, nice work!

As for your question, yeah I think the CDN example should use the CDN assets for sure, I was looking at that html-css example yesterday and thinking that same thing, shouldn't have those dupes in our repo.

examples/html-css/index.html
pwolfert: I like that idea!

PR #2227, opened 2022-10-27
[NO-TICKET] Add missing image file and alt tag

pwolfert:
## Summary

- Add an image that was no longer in the repo but being used in the doc site
- And add some alt text for this image (Or should I use `alt=""` since this section explains the naming convention, @SparrowSpit?)

## How to test

https://cmsgov.github.io/design-system/branch/pwolfert/add-missing-diagram-image/getting-started/developers/code-conventions/?theme=core


zarahzachz:
Something I noticed using Firefox/VO, each piece of content ends with `clickable` when read out. Why is that?

forrestbaer (approved):

PR #2228, opened 2022-10-27
[WNMGDS-1679] Bugfix: tooltip not going away when focus lost

forrestbaer:
## Summary

- Currently focusing on a tooltip trigger will open the tooltip but shifting focus will not close it automatically.
- Changes logic to check if `event.currentTarget !== event.target` to close when focus is lost.
- Adds unit tests for this functionality

## How to test

1. `yarn install && yarn build && yarn storybook`
2. Open the tooltip examples and use keyboard to navigate them, moving focus away from a trigger should close the tooltip.
3. `yarn test:unit`

## Checklist


packages/design-system/src/components/Tooltip/Tooltip.test.jsx
```diff
@@ -40,12 +41,30 @@ describe('Tooltip', function () {
...
     renderTooltip({ dialog: true });
     const tooltipTrigger = screen.getByLabelText(triggerAriaLabelText);
-    fireEvent.click(tooltipTrigger);
+    userEvent.click(tooltipTrigger);
     const contentEl = screen.queryByRole('dialog');
     expect(contentEl).not.toBeNull();
     expect(contentEl).toMatchSnapshot();
   });
 
+  it('closes tooltip when trigger focus is lost', async () => {
```
pwolfert, line 50: Nice test! I'm glad we're using RTL now. Trying to mock this behavior with Enzyme would have been clunky and fragile

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -136,13 +136,9 @@ export const Tooltip = (props: TooltipProps) => {
...
 
-  const handleBlur = (event: MouseEvent) => {
+  const handleBlur = (event: Event) => {
     setTimeout(() => {
-      const focusedInsideTrigger = triggerElement.current?.contains(event.target);
-      const focusedInsideTooltip = tooltipElement.current?.contains(event.target);
-      if (!focusedInsideTrigger && !focusedInsideTooltip && !isHover) {
-        setActive(false);
-      }
+      if (!isHover && event.currentTarget !== event.target) setActive(false);
```
pwolfert, line 141: What a much more elegant solution!

pwolfert (approved):
Nice work! The only thing I'm hesitant about is whether we've covered a potential scenario that might have neccessitated the whole search-for-any-element-inside logic. I tested this with VoiceOver, but maybe @SparrowSpit could try out this component just to be sure nothing else broke.

SparrowSpit:
@forrestbaer  @pwolfert tested with Safari + VoiceOver. Works as it did previously with no issues. Nice work!

PR #2236, opened 2022-11-03
[DON'T MERGE] [WNMGDS-2031] Tooltip checkbox story

zarahzachz:
## Summary

WNMGDS-2031 

This work is for @SparrowSpit to review the a11y of a Tooltip nested within a Choice (checkbox) component.

**DON'T MERGE! PATTERN NOT APPROVED BY DESIGN SYSTEM**

URL: https://cmsgov.github.io/design-system/branch/WNMGDS-2031/tooltip-checkbox-story/storybook/?path=/story/components-choicelist--choice-tooltip&globals=theme:core


SparrowSpit:
For documentation purposes, I'm pasting these thoughts of mine from our Slack convo. 

- Medicare already uses this pattern, fwiw. See screenshot. I wasn't excited when that happened but it came from CMS and was the lesser of the evils they wanted.
- We should point to [Understanding Success Criterion 1.4.13: Content on Hover or Focus](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html)  which basically says to be very careful with things like tooltips covering up content (as you'd mentioned to her in the other Slack) though it provides guidance on how you can do it.
- Screen reader testing doesn't raise issues
- and there's the whole styling/z-index/overlap issue. 

Our feedback to the team (per collaborating with Sarah): 

> We recommend displaying the content without a tooltip if at all possible, especially within a form, so that you're not covering up content that is useful and needs to be acted upon. If the content is useful enough to include then try to make it always visible.
> That being said, if your team still needs to utilize the tooltip then it's allowed to include a tooltip trigger within the hint text but not on the label element directly.
> Lastly, there is the issue of the current styling overlapping as Sarah showed in her recording above but Sarah will work to get the issue in our backlog and see about getting it prioritized. In the meantime, you can probably get around this problem by explicitly declaring the Tooltip placement="top"

PR #2237, opened 2022-11-03
[WNMGDS-2009] Document pass-through props

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2009

If the component we're documenting passes extra props through to an HTML element, we want to include documentation about that. Rather than listing out all the additional attributes for those HTML elements in our props table, I figured we could add some explanatory text with MDN documentation links to the bottom of our props table [like Primer does](https://primer.style/react/Button#props).

- Adds a `PropsTableHtmlElementRow` component
- Modifies `PropsTable` to accept `children`, which it renders at the bottom

## How to test

`yarn start` and visit one of the updated pages

## Checklist

## Requested feedback

1. I figured putting that information in the table would be helpful to people who use screen readers, but is there a better way?
2. Ideas for a better way to explain this?

### DIFF
--- a/packages/docs/content/components/button.mdx
+++ b/packages/docs/content/components/button.mdx
@@ -242,7 +242,7 @@ In addition to the supported props listed, you can also pass in additional
 props, which will be passed to the rendered root component. For example,
 you could pass in a `target` prop to pass to the rendered anchor element.
 
-<PropTable componentName="Button" />
+<PropTable componentName="Button" htmlElementName="button" />
 
 ### Styles
 
--- a/packages/docs/content/components/button.mdx
+++ b/packages/docs/content/components/button.mdx
@@ -243,12 +243,7 @@ props, which will be passed to the rendered root component. For example,
 you could pass in a `target` prop to pass to the rendered anchor element.
 
 <PropTable componentName="Button">
-  <PropTableHtmlElementRow
-    elements={[
-      { name: 'button', link: 'https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button' },
-      { name: 'a', link: 'https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a' },
-    ]}
-  />
+  <PropTableHtmlElementRow elements={['button', 'a']} />
 </PropTable>
 
 ### Styles
--- a/packages/docs/content/components/button.mdx
+++ b/packages/docs/content/components/button.mdx
@@ -242,7 +242,14 @@ In addition to the supported props listed, you can also pass in additional
 props, which will be passed to the rendered root component. For example,
 you could pass in a `target` prop to pass to the rendered anchor element.
 
-<PropTable componentName="Button" htmlElementName="button" />
+<PropTable componentName="Button">
+  <PropTableHtmlElementRow
+    elements={[
+      { name: 'button', link: 'https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button' },
+      { name: 'a', link: 'https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a' },
+    ]}
+  />
+</PropTable>
 
 ### Styles
 
--- a/packages/docs/src/components/content/PropTableHtmlElementRow.tsx
+++ b/packages/docs/src/components/content/PropTableHtmlElementRow.tsx
@@ -16,19 +16,21 @@ export interface PropTableHtmlElementRowProps {
  * table.
  */
 const PropsTableHtmlElementRow = ({ elements }: PropTableHtmlElementRowProps) => {
+  const elementNames = elements.map((name) => <code key={name}>{`<${name}>`}</code>);
   const elementLinks = elements.map((name) => (
     <a href={`https://developer.mozilla.org/en-US/docs/Web/HTML/Element/${name}`} key={name}>
-      {`<${name}>`}
+      {name}
     </a>
   ));
-  const formattedElementLinks = humanizeList(elementLinks, { conjunction: 'or' });
+  const formattedElementNames = humanizeList(elementNames, { conjunction: 'or' });
+  const formattedElementLinks = humanizeList(elementLinks, { conjunction: 'and' });
 
   return (
     <TableRow>
       <TableCell colSpan={4}>
-        This component passes any additional props to its underlying {formattedElementLinks} element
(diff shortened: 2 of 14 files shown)


packages/docs/src/components/content/PropTableHtmlElementRow.tsx
```diff
@@ -0,0 +1,37 @@
...
+    <a href={`https://developer.mozilla.org/en-US/docs/Web/HTML/Element/${name}`} key={name}>
+      {`<${name}>`}
+    </a>
+  ));
+  const formattedElementLinks = humanizeList(elementLinks, { conjunction: 'or' });
+
+  return (
+    <TableRow>
+      <TableCell colSpan={4}>
+        This component passes any additional props to its underlying {formattedElementLinks} element
```
line47, line 29: Can we unlink this and show as an inline code block?

packages/docs/src/components/content/PropTableHtmlElementRow.tsx
```diff
@@ -0,0 +1,37 @@
...
+        as attributes. It will accept any props that are valid attributes of {formattedElementLinks}
```
line47, line 30:
```suggestion
        as attributes. See MDN for a list of valid attributes for {formattedElementLinks}.
```

Unlink the code block and show as inline and link up MDN text instead. I think is more understandable about where the link will be taking the user.

pwolfert: I had it like that in my first iteration, but 2 of the 6 components have a choice between two different underlying elements. How would I link to each one in that case and make it clear which link is which? That's why this version uses an array for the element(s) prop.

line47: Ah gotcha, maybe then adding some text about the MDN would be helpful. Right now it's unclear where the link is going or if it's external

pwolfert: Here's what it looks like now: 

Screen Shot 2022-11-07 at 4 11 40 PM

forrestbaer (approved):
Looks good 👍🏻 Interesting addition, nice implementation.

packages/docs/src/components/content/PropTableHtmlElementRow.tsx
```diff
@@ -0,0 +1,39 @@
...
+ * we want to include documentation about that. Rather than listing out all the
+ * additional attributes for those HTML elements in our props table, we will add
+ * some explanatory text with MDN documentation links to the bottom of our props
+ * table.
+ */
+const PropsTableHtmlElementRow = ({ elements }: PropTableHtmlElementRowProps) => {
+  const elementNames = elements.map((name) => <code key={name}>{`<${name}>`}</code>);
+  const elementLinks = elements.map((name) => (
+    <a href={`https://developer.mozilla.org/en-US/docs/Web/HTML/Element/${name}`} key={name}>
+      {name}
```
zarahzachz, line 22: I think the names should be wrapped in brackets to denote the HTML element referenced. 

MDN page titles wrap the names of these elements in them and it looks a little odd reading the word "div" without brackets or formatted as code in our docs.

packages/docs/src/components/content/PropTableHtmlElementRow.tsx
```diff
@@ -0,0 +1,39 @@
+import humanizeList from 'humanize-react';
```
zarahzachz, line 1: Could you use the [`Intl` API](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/ListFormat) to handle this instead of importing a new library?

packages/docs/content/components/button.mdx
```diff
@@ -242,7 +242,9 @@ In addition to the supported props listed, you can also pass in additional
 props, which will be passed to the rendered root component. For example,
 you could pass in a `target` prop to pass to the rendered anchor element.
 
-<PropTable componentName="Button" />
+<PropTable componentName="Button">
+  <PropTableHtmlElementRow elements={['button', 'a']} />
```
zarahzachz, line 246: `a` looks weird here without being wrapped in brackets or <code> tags

packages/docs/src/components/content/PropTableHtmlElementRow.tsx
pwolfert: 😍

packages/docs/src/components/content/PropTableHtmlElementRow.tsx
pwolfert: I've gone back and forth on this. https://github.com/CMSgov/design-system/pull/2239 has a change (underlined code blocks in anchor tags) that will make this look better. As it stands, if you wrap a code tag in a link, you can't tell it's a link on our site.

packages/docs/src/components/content/PropTableHtmlElementRow.tsx
pwolfert: Oh wait, the reason I used a library is because these are actually arrays of React elements and not strings. Each element in an array of React elements needs a key, so this library will take an array of React elements and return one where more React elements are inserted for the commas and conjunctions with keys without having to convert the links and code blocks into strings and then use `dangerouslySetInnerHTML` later.

packages/docs/src/components/content/PropTableHtmlElementRow.tsx
pwolfert: I've changed the styles for code blocks in links so they look like links (kind of like [`<github>` does](#)). I've also taken out the vertical padding of code blocks everywhere to make them render more consistently in different contexts. During the blog post work, @babsdenney suggested we take out the vertical padding when they're in paragraphs so they don't alter the line heights, which makes a paragraph look wonky. That change didn't affect the table content, so instead of adding another exception, I decided to make it consistent for all contexts. I left some horizontal padding to expand the background a bit for legibility because it doesn't affect line height. What do you think, @line47 and @babsdenney?

Here's what it looks like now:
Screen Shot 2022-11-08 at 9 56 38 AM

packages/docs/src/components/content/PropTableHtmlElementRow.tsx
zarahzachz: Oh fiddlesticks! Didn't know that! I'm good with the library then.

PR #2245, opened 2022-11-11
[WNMGDS-2003] Fix monthpicker error readout

zarahzachz:
## Summary

WNMGDS-2003

Initial SR readout skipped error text when reporting MonthPicker

Discovered `role=alert` basically hid this text from SR's (probably because the expected use was for the error to appear after the user submits, changing the DOM). Changing the alert role to its equivalent `aria-live=assertive aria-atomic=true` fixes this issue. See research [here](https://cdpn.io/pen/debug/yLEJjZL).

## How to test

1. View [MonthPicker component](https://cmsgov.github.io/design-system/branch/WNMGDS-2003/fix-monthpicker-error/storybook/?path=/story/components-month-picker--default-month-picker&globals=theme:core) via SR


pwolfert (approved):
Great work, @zarahzachz! Code LGTM. This is well researched. Is this something that we should consider adopting for all our field components?

zarahzachz:
> Is this something that we should consider adopting for all our field components?

With this change happening on the InlineError component, all field components should inherit this fix 🎉

forrestbaer (approved):
Looks good!

pwolfert:
> With this change happening on the InlineError component...

Doh! I didn't catch that. 🎉

PR #2246, opened 2022-11-14
[WNMGDS-2020] Add a11y guidance for HCgov alert

line47:
## Summary

- Added minor guidance the alert component that had not been moved into the doc site

## How to test

- run the doc site with `yarn start` and see that the alert page has addition a11y content for the HCgov theme. 

## Checklist


zarahzachz (changes requested):
This looks good. Could you tie this PR to a Jira ticket?
Last I heard, we were going to end the NO-TICKET practice to make our release processes more clear.

packages/docs/content/components/alert.mdx
```diff
@@ -250,13 +250,94 @@ A custom heading value can be sent for an analytics event by using the prop `ana
 
 ### Accessibility
 
+<ThemeContent onlyThemes={['healthcare']}>
+
+If dynamically loaded, be mindful of when the alert displays on the page. For example, an alert appearing on a page due to an error or progressive disclosure. Where and how the content appears visually should align with how and when a screen reader user hears the content. Errors need immediate attention, but content through progressive disclosure may not.
```
pwolfert, line 255:
```suggestion
If dynamically loaded, be mindful of when the alert displays on the page. For example, an alert could appear on a page due to an error or progressive disclosure. Where and how the content appears visually should align with how and when a screen reader user hears the content. Errors need immediate attention, but content through progressive disclosure may not.
```

to make a full sentence

packages/docs/content/components/alert.mdx
```diff
@@ -250,13 +250,94 @@ A custom heading value can be sent for an analytics event by using the prop `ana
...
+
+- Users who are viewing the content need it to be placed in an effective location and to draw attention even when CSS is turned off. Content can help here too, using word choices that make the importance of the message clear.
+- Users who hear this content through assistive technology, such as a screen readers, will need to be alerted through the backend code effectively at the right time.
```
pwolfert, line 258:
```suggestion
- Users who hear this content through assistive technology, such as a screen readers, will need to be alerted at the right time, and that can be managed with code.
```

Trying to help this make more sense. It doesn't necessarily take extra "backend code" to do this, but it may require some code tweaking if the defaults won't cut it.

packages/docs/content/components/alert.mdx
```diff
@@ -250,13 +250,94 @@ A custom heading value can be sent for an analytics event by using the prop `ana
...
+
+- Consider the amount of text placed in an alert. Aim for the shortest possible amount of text to convey the concept or issue.
+- Write the message in concise, plain language; avoid jargon and computer code.
+- Alerts support various types and lengths of content from ~45 to ~80 characters wide, including lists and navigational links.
+
+#### Heading content
+
+- Is bolded and acts as the title of the alert
+- Helps to group or frame the information in the alert.
+- Shouldn't end in a period.
```
pwolfert, line 282: These probably shouldn't end in periods for consistency and because none of them are full sentences without "heading content" as the subject, which suggests that this list is all part of the same sentence.

packages/docs/content/components/alert.mdx
```diff
@@ -250,13 +250,94 @@ A custom heading value can be sent for an analytics event by using the prop `ana
...
+
+#### Error content 
+
+- Describe what went wrong and what the the user needs to do.
+- Use the least amount of words possible (i.e., don't use "please").
```
pwolfert, line 287:
```suggestion
- Use the fewest words possible (i.e., don't use "please").
```

pwolfert (approved):
Have suggestions for improved clarity and consistency, but LGTM otherwise

zarahzachz (approved):

PR #2256, opened 2022-11-15
[WNMGDS-2052] temporary 5.0 rel solution for analytics in propTable

forrestbaer:
## Summary

- Updates the `PropTable` doc site with an additional optional `hasAnalytics` prop that displays the analytics props currently stored in a separate file.
- Updates component doc mdx where these props are utilized
- **NOTE**: This is a bandaid fix that will need to be resolved before v6.0 is released in order to have these props available in the documentation due to the limitations of `react-docgen` and the `gatsby-transformer-react-docgen` plugin.

### How to test:

- `yarn clean && yarn install && yarn build && yarn build:storybook:docs && yarn start` - view changed pages, confirm analytics props appear at top of prop table
- `yarn test`


packages/docs/src/helpers/analyticsPropTable.tsx
```diff
@@ -0,0 +1,36 @@
+export const analyticsPropTable = [
+  {
+    name: 'analytics',
+    type: 'boolean',
+    defaultValue: 'true',
+    text: 'Analytics events tracking is enabled by default. Set this value to `false` to disable tracking for this component instance.',
```
forrestbaer, line 6: Using `description` doesn't work as that is MDX data that gets compiled, can't compile a string.. Couldn't figure out a way to convert a string to MDX easily at runtime, if it's possible at all, so adding another field here.

pwolfert (commented):
Nice work, @forrestbaer! This will work for now. We'll want to make it clear in this PR description that this fix will only work for this release but that the `useFormLabel` changes are going to have the same issue when it's time for `6.0`. We'll need to create a follow-up ticket to address it with a longer-term fix.

Also, I noticed that some of these descriptions have markdown annotation like monospace and bold that the string rendering doesn't support

packages/docs/content/components/button.mdx
```diff
@@ -242,7 +242,7 @@ In addition to the supported props listed, you can also pass in additional
 props, which will be passed to the rendered root component. For example,
 you could pass in a `target` prop to pass to the rendered anchor element.
 
-<PropTable componentName="Button">
+<PropTable componentName="Button" hasAnalytics >
```
pwolfert, line 245: I found one more annoying exception: Button uses `AnalyticsParentDataProps` as well

forrestbaer: Ah, I wasn't sure about those, should they go on all that `hasAnalytics` or just button?

forrestbaer: Nevermind, I see, just for buttons.. updating

forrestbaer: Updated

pwolfert (approved):

PR #1854, opened 2022-05-20
[WNMGDS-1402] Remove unnecessary nav els

zarahzachz:
## Summary

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-1402)

For vertical nav: Added conditional logic to wrap link list with a `<nav>` only if it's not nested (using already existing `nested` prop).

For header: WIP

## How to test

For vertical nav: Run `yarn storybook` locally, navigate to http://localhost:6006/?path=/story/components-vertical-nav--default-vertical-nav and open developer tools. You should only find one instance of `<nav>` being used.

For header: Run `yarn storybook:healthcare` and navigate to Header component. Note only one `<nav>` in use.

PR #2258, opened 2022-11-17
[WNMGDS-965] Add aria-label to header nav elements.

zarahzachz:
## Summary

WNMGDS-965

- Add `aria-label` to `<nav>` elements in HCgov header
- Replace `<div>` with semantic `<nav>` element in HCgov header

## How to test

Review a11y tree of Healthcare header (building)

### DIFF
--- a/packages/ds-healthcare-gov/src/components/Header/Menu.tsx
+++ b/packages/ds-healthcare-gov/src/components/Header/Menu.tsx
@@ -24,12 +24,17 @@ const Menu = (props: MenuProps) => {
   });
 
   return (
-    <div id="hc-c-menu" hidden={!props.open} className={classes}>
+    <nav
+      aria-label="Profile and application"
+      id="hc-c-menu"
+      hidden={!props.open}
+      className={classes}
+    >
       {props.submenuTop}
       {props.beforeLinks}
       {props.links && <MenuLinks links={props.links} />}
       {props.submenuBottom}
-    </div>
+    </nav>
   );
 };
 
--- a/packages/ds-healthcare-gov/src/components/Header/Header.tsx
+++ b/packages/ds-healthcare-gov/src/components/Header/Header.tsx
@@ -228,6 +228,7 @@ export const Header = (props: HeaderProps) => {
           </a>
 
           <nav
+            aria-label="action"
             id="hc-c-header__actions"
             className="hc-c-header__actions ds-l-col ds-l-col--auto ds-u-margin-left--auto ds-u-font-weight--bold"
           >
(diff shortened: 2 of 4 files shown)


packages/ds-healthcare-gov/src/components/Header/Menu.tsx
```diff
@@ -24,12 +24,17 @@ const Menu = (props: MenuProps) => {
   });
 
   return (
-    <div id="hc-c-menu" hidden={!props.open} className={classes}>
+    <nav
+      aria-label="Profile and application"
+      id="hc-c-menu"
+      hidden={!props.open}
+      className={classes}
+    >
```
pwolfert, line 32: This actually seems like we're reversing what you did [in the previous nav cleanup changes](packages/ds-healthcare-gov/src/components/Header/Menu.tsx). Is there nuance that I'm missing here?

zarahzachz: DK specified this should be a `<nav>` element in the Jira ticket, so it seems the `<div>` was an incorrect change previously.

pwolfert: Okay, maybe we should ask @SparrowSpit to take a look at this version and give a final ruling

SparrowSpit: I can't seem to view the link Patrick pasted for the old nav cleanup changes. Was there a reason to use the div ? It would seem the nav element would semantically be correct but I just want to make sure there isn't some historical context I'm missing.

pwolfert: Updated the link to point to the previous change commit. Here's [the pull request](https://github.com/CMSgov/design-system/pull/1854), which also has a link to [its Jira ticket](https://jira.cms.gov/browse/WNMGDS-1402)

SparrowSpit: That's helpful. Thank you. So, it would seem to me that we are going back to having the issue DK described where we have nested nav elements.

zarahzachz: Yes. Is that an issue if the nested navs have labels? Or is it best to revert back to the div solution and omit the aria-label?

SparrowSpit: Ok. Sarah and I chatted about this to help me understand the historical context. Let's do this:

- stick with the div on what is now the nested nav element
- keep the outer nav element that contains both the user name and the menu+actions
- use the aria-label="Profile, applications, and coverage" on the outer nav

pwolfert (approved):

PR #2264, opened 2022-11-18
[WNMGDS-2024] Update LabelMask with all Mask Options

forrestbaer:
## Summary

- Updated `LabelMask` with all the available masks for `Mask`
- Updated tests and snapshots, storybook stories

## How to test

1. `yarn clean && yarn install && yarn build && yarn test`
2. `yarn storybook` - inspect new mask stories

## Checklist


packages/design-system/src/components/TextField/Mask.tsx
```diff
@@ -36,6 +36,10 @@ export class Mask extends React.PureComponent<MaskProps, any> {
     const field = this.field();
     const initialValue = field.props.value || field.props.defaultValue;
 
+    console.warn(
+      '[Deprecated]: Please utilize the LabelMask prop in place of the Mask prop. Mask will be deprecated for use in the next major release'
```
zarahzachz, line 40: Should we have a ticket in our backlog to track this?

packages/design-system/src/components/TextField/LabelMask.test.tsx
```diff
@@ -1,79 +1,127 @@
...
-  });
-  describe('updates label mask to reflect', () => {
-    it('complete input value set', () => {
-      const { container } = render(<TestLabelMask />);
-      const data = '12345678';
-      const maskText = '12/34/5678';
-
+allMasks.forEach((currentMask) => {
+  describe(`${currentMask.name} Label mask`, function () {
+    it(`renders default mask, ${currentMask.default}., when no input value set`, () => {
```
zarahzachz, line 74: There's a period here after the interpolated string I don't think should be there - it's showing up in snapshots too

packages/design-system/src/components/TextField/useLabelMask.test.tsx
```diff
@@ -1,8 +1,116 @@
...
+});
+
+describe('CURRENCY_MASK', () => {
+  it('returns just the mask when given no input', () => {
+    expect(CURRENCY_MASK('')).toEqual('$0.00');
+  });
+
+  it('returns correctly formatted dollar amounts', () => {
+    expect(CURRENCY_MASK('1')).toEqual('$1.00');
+    expect(CURRENCY_MASK('12.479')).toEqual('$12.48');
```
zarahzachz, line 68: Just curious - do we expect the `.479` here to round up, or should we trim any chars after 2 decimal places? `.48` vs `.47`?

I also think this is the plot to Office Space.

zarahzachz (commented):
This looks awesome! Love how consistent all these masks are. One question, I noticed for the currency mask, the result says `NaN` when users input any non-numeric character (makes sense), but our other masks seem to ignore that input. 

Do we want users to see `NaN` as a result, or should currency mask behave like other masks and simply ignore non-numeric inputs?

packages/design-system/src/components/TextField/useLabelMask.test.tsx
pwolfert: I would probably vote for trimming. What does the old one do?

packages/design-system/src/components/TextField/Mask.tsx
pwolfert: We can't quite deprecate the old one. I think we should do this in a separate PR under a separate ticket. (And for that reason we should also keep the old all-masks story around.)

packages/design-system/src/components/TextField/LabelMask.test.tsx
```diff
@@ -1,79 +1,127 @@
...
+        expect(mask.textContent).toContain(currentMask.partialResult);
+      });
     });
-  });
-  it('formats input value onBlur', () => {
-    const { container } = render(<TestLabelMask />);
-    const data = '12345678';
-    const formattedData = '12/34/5678';
+    // currency mask doesn't currenty reflect formatting in input, only label
+    if (currentMask.mask !== CURRENCY_MASK) {
```
pwolfert, line 112: Oh, I see this relates to my comment above

pwolfert: Nice tests, though!

pwolfert (commented):
Good organization and reuse!

packages/design-system/src/components/TextField/useLabelMask.tsx
```diff
@@ -5,38 +5,102 @@ import { TextInputProps } from './TextInput';
...
+    .filter((s) => s)
+    .join('/');
+
+  return formattedDate;
+});
+
+/**
+ * Formatting for US phone numbers
+ */
+export const PHONE_MASK: MaskFunction = makeMask(RE_PHONE, 'XXX-XXX-XXXX', (match) => {
```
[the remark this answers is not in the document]
forrestbaer: I think it might be a bit of a pain to get this one to work with the mask being like that, since it uses a different pattern than the others. Personally I think `$#.##` or `$X.XX` both look a little odd compared to `$0.00`

packages/design-system/src/components/TextField/useLabelMask.tsx
```diff
@@ -5,38 +5,102 @@ import { TextInputProps } from './TextInput';
...
+});
+
+/**
+ * Currency mask is a little different, we need to modify the incoming content to strip
+ * out any commas or dollar signs before evaluating it via the Intl.NumberFormat function.
+ */
+export const CURRENCY_MASK = (rawInput = '', valueOnly = false) => {
+  const asNumber = Number(rawInput.replace(/[$,]/g, ''));
+  const USDollar = new Intl.NumberFormat('en-US', { style: 'currency', currency: 'USD' });
+  const formatted = USDollar.format(asNumber);
```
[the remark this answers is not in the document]
forrestbaer: Yeah this currency one needs a little updating, @zarahzachz 's point about it returning NaN with non-numerical numbers is another issue, looking into it.

packages/design-system/src/components/TextField/useLabelMask.tsx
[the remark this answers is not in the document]
forrestbaer: I agree `#` makes sense for the digits

forrestbaer:
@pwolfert @zarahzachz I believe I've addressed all comments and resolved issues, thanks for the feedback!

packages/design-system/src/components/TextField/useLabelMask.tsx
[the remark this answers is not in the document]
pwolfert: I feel like it would be better to show nothing since users would understand that it's looking for numbers. The problem is 1) consistency and 2) avoiding content shift if we show and hide a mask.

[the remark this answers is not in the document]
pwolfert: One thing the old one did, too, was only insert the decimals if they were present in the raw input. It's nice to be able to use the Intl module, but some complexity might be necessary here for more nuanced behaviors.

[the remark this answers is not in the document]
pwolfert: This works as is, though, so maybe we should continue to refine this one in a follow-up pull request.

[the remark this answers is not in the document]
forrestbaer: It seems like we need a discussion as a team to determine the requirements for this one since it's a modified pattern. Since we won't really be meeting again all together till after Thanksgiving I would agree maybe a followup to change anything that comes up after that consensus would be a good idea.

packages/design-system/src/components/TextField/TextField.stories.jsx
```diff
@@ -89,59 +82,50 @@ export const LabelMaskedField = (args) => {
...
-};
+
 LabelMaskedField.argTypes = {
   // Hide irrelevant props
   mask: { table: { disable: true } },
   type: { table: { disable: true } },
   rows: { table: { disable: true } },
 };
 
-export const AllMaskedFields = () => {
```
pwolfert, line 105: Can we put this back for now?

pwolfert (approved):
LGTM except for restoring an old story. Nice work!

packages/design-system/src/components/TextField/TextField.stories.jsx
forrestbaer: Ah yes, forgot I removed it :thum

PR #2268, opened 2022-11-22
[WNMGDS-2094] Fix centered containers shifting left when dialog opens

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2094

- The fix is setting a `left: 0` and `right: 0` on the body to make it fill the whole width of the screen when `position: fixed`, which makes an assumption about how the application wants its body element to be laid out
- Because that assumption could be wrong in some applications, we wanted to make it easy to override the behavior in applications, so we the body styles with a new CSS class

## How to test

1. Start storybook and navigate to http://localhost:6006/?path=/story/components-dialog--prevent-scroll-example
2. Find the `id="root"` element in the story iframe, and apply a `class="ds-l-container"` to mimic the layout of a standard application.
3. Click the button to open the dialog. In this branch, there should be no shift of the content. On the main branch, it will shift all the content to the left.

## Checklist


forrestbaer (approved):
Looks good!

PR #2269, opened 2022-11-22
[WNMGDS-2095] Fix disabled checked choices

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2095

- Fix disabled checked choices not looking disabled
- Improve support for them in forced-colors mode
- Created two new stories for disabled choices so we can capture this scenario in visual regression tests

## How to test

1. Check the new stories in Storybook (and in forced-colors mode)
2. Look at the new screenshots

## Checklist


pwolfert:
I might follow up on this with a pull request that separates the checkbox and radio styles more. Right now they're just really intertwined and complex

packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -165,6 +173,23 @@ label,
...
+      &::before {
+        box-shadow: inset 0 0 4em 1em $choice__border-color--disabled;
+      }
+
+      @media (-ms-high-contrast: active), (forced-colors: active) {
+        border-color: GrayText;
+
+        &::before {
+          background-color: GrayText;
+          box-shadow: inset 0 0 4em 1em GrayText;
```
zarahzachz, line 189: `box-shadow` isn't respected by HCM, so I don't think this is doing anything here.

zarahzachz (approved):
Looks good! Had one comment, but works pretty well overall.

forrestbaer (approved):
Looks good!

packages/design-system/src/styles/components/_Choice.scss
pwolfert: Good call!

PR #2275, opened 2022-11-23
[WNMGDS-2074] Release script updates

forrestbaer:
## Summary

1. [WNMGDS-2074](https://jira.cms.gov/browse/WNMGDS-2074)
2. Split our release script into 2 new scripts `bump-and-tag-release.sh` and `bump-main.sh` as these two processes are separate from each other. By doing our version bumps twice instead of cherry-picking a commit from one place to another, we avoid any possible errors with branch collisions or conflict resolution. Overall I believe this is a faster process as a result.
3. The `bump-and-tag-release.sh` script gets run as `yarn release` by default, to be run in a branch to tag and bump, then push tags. That is all that's needed to kick off a release with jenkins.
4. The `bump-main.sh` script gets run with `yarn release:bump` (maybe just `yarn bump`?) and creates a new branch off main, bumps versions without tagging and automagically creates a bump PR from that branch into main with `gh`

- I'm assuming we all have gh installed on our machines and that is a part of our tooling for the team. If not then it should be!
- [ ] Need to update [confluence documentation](https://confluence.cms.gov/display/HCDSG/Creating+a+code+release)

## How to test

1. `yarn release` in this branch, bump versions to whatever, see that it runs then run `yarn release -u` to undo those changes.
2. `yarn release:bump` in this branch, bump versions to whatever, see that PR is created, close PR and delete branch.

## Checklist


pwolfert:
Explanation for [my change](https://github.com/CMSgov/design-system/pull/2275/commits/306f99276167dd40e37b645014a13e32041cbc55): When I was looking at the new script in my editor, I noticed that this variable we've had in there for a while wasn't being used. I think the intention was originally to output it in the instructions. I went ahead and did that because it's handy.

Screen Shot 2022-11-23 at 9 11 59 AM

pwolfert (approved):
LGTM 👍 Can't wait to use it 🚀

PR #2277, opened 2022-11-28
[WNMGDS-2097] useLabelMask refinements

forrestbaer:
## Summary

[WNMGDS-2097](https://jira.cms.gov/browse/WNMGDS-2097)

* useFormLabel / TextField can be used controlled or uncontrolled
* updated `DATE_MASK` to return to default state when blurring away and return on focus
* updated `CURRENCY_MASK` to use default `makeMask` helper like the others
* `CURRENCY_MASK` will return nothing if `0`, alphanumeric characters / invalid data are entered, will no longer return $0.00
* `CURRENCY_MASK` uses a simpler 'mask' which is just a $ now, as it doesn't really make sense to have 0's in places where they may not be, hopefully less confusing
* added new stories to test uncontrolled components
* updated unit tests and snapshots to reflect changes
* Updated `CURRENCY_MASK` to strip out `.00` when no cents are given, @pwolfert it seems like adding another (basically duplicate) `CURRENCY_MASK` with this behavior is a bit of a waste/cruft.  Give the functionality of this a try and let me know what you think. It seems like the expected behavior when I type `1000` into a currency box would be `$1,000` at the output, `$1,000.00` seems more confusing.

## How to test

1. `yarn clean && yarn install && yarn build && yarn build:storybook`
2. `yarn test:unit && yarn storybook` - validate stories work as intended

- [ ] need to test controlled components outside of storybook or stop using `useArgs` as that may be causing the backspace issue reported in [WNMGDS-2105](https://jira.cms.gov/browse/WNMGDS-2105)

## Checklist


packages/design-system/src/components/TextField/LabelMask.test.tsx
```diff
@@ -52,11 +52,11 @@ const allMasks = [
   {
     name: 'US Currency',
     mask: CURRENCY_MASK,
-    default: '$0.00',
+    default: '$',
     defaultData: '12345.67',
     defaultResult: '$12,345.67',
     partialData: '123',
-    partialResult: '$123.00',
+    partialResult: '$123',
```
pwolfert, line 59: I like that this test tests two different formatting strategies in one

packages/design-system/src/components/TextField/TextField.stories.jsx
```diff
@@ -175,3 +176,19 @@ LabelMaskedCurrency.args = {
   hint: 'Please enter a dollar amount',
   labelMask: CURRENCY_MASK,
 };
+
+export const UncontrolledLabelMaskedDate = UncontrolledLabelMaskedField.bind({});
```
pwolfert, line 180: Really handy to have these 🙌

packages/design-system/src/components/TextField/__snapshots__/useLabelMask.test.tsx.snap
```diff
@@ -125,7 +125,7 @@ exports[`useLabelMask shows masked value when focused 1`] = `
     <span
       class="ds-u-visibility--screen-reader"
     >
-      MM/DD/YYYY
+      12/DD/YYYY
```
pwolfert, line 128: 👍

packages/design-system/src/components/TextField/useLabelMask.tsx
```diff
@@ -93,37 +93,58 @@ export const SSN_MASK: MaskFunction = makeMask(RE_SSN, '###-##-####', (match) =>
...
-export const CURRENCY_MASK = (rawInput = '', valueOnly = false) => {
-  const signed = rawInput.includes('-');
-  const stripped = rawInput.replace(/[^0-9.]/g, '');
+export const CURRENCY_MASK = makeMask(RE_CURRENCY, '$', (match) => {
+  const signed = match[0].includes('-');
+  const stripped = match[0].replace(/[^0-9.]/g, '');
   const clipped = stripped.includes('.') ? stripped.slice(0, stripped.indexOf('.') + 3) : stripped;
   const USDollar = new Intl.NumberFormat('en-US', { style: 'currency', currency: 'USD' });
-  const formatted = USDollar.format(Number(clipped));
+  const formatted = USDollar.format(Number(clipped)).replace(/\.00/, '');
```
pwolfert, line 101: These names are funny 😆. So succinct and descriptive and rhythmic. I feel like it's part of a song until I get to `USDollar`. You know...the US Dollar formatter could go outside the function so we don't break up the song 🎼

packages/design-system/src/components/TextField/useLabelMask.tsx
```diff
@@ -93,37 +93,58 @@ export const SSN_MASK: MaskFunction = makeMask(RE_SSN, '###-##-####', (match) =>
...
 
-  return signed ? '-' + formatted : formatted;
-};
+  console.log(Number(clipped));
```
pwolfert, line 103: Debug log 🗑️

packages/design-system/src/components/TextField/useLabelMask.tsx
```diff
@@ -93,37 +93,58 @@ export const SSN_MASK: MaskFunction = makeMask(RE_SSN, '###-##-####', (match) =>
...
   // TODO: Once we're on React 18, we can use the `useId` hook
   const labelMaskId = useRef(uniqueId('labelmask_')).current;
   const [focused, setFocused] = useState(false);
   const { onFocus, onBlur, onChange } = originalInputProps;
   const value = originalInputProps.value?.toString() ?? '';
+  const [currentValue, setCurrentValue] = useState(value);
+
+  useEffect(() => {
+    setCurrentValue(value);
+  }, [value, setCurrentValue]);
```
pwolfert, line 122: I haven't been able to break it! This really seems to work. I'm still really puzzled why we get the cursor jump in the controlled version. When I put a console log in here like

```ts
console.log(`value: ${value}, currentValue: ${currentValue}`)
```

these values are always the same in the controlled component story, which is what I would expect. If the value isn't actually changing, then `setCurrentValue` should be a no-op, which shouldn't cause a re-render, which shouldn't cause a jump in the cursor. I guess I should also try to figure out if it's re-rendering anyway.

pwolfert (approved):
Great work! LGTM 👍

packages/design-system/src/components/TextField/useLabelMask.tsx
pwolfert: Because this is happening in a hook, I guess it could be more complicated than that and we need to look at the actual components' render cycles. There's a `TextField` rendering a `TextInput` inside a `LabelMask` with a `useLabelMask` hook. This is a task for another day!

packages/design-system/src/components/TextField/useLabelMask.tsx
forrestbaer: Lol :) it's like 'signed, sealed delivered' by Stevie Wonder... 'Signed, stripped, form-matted 🎵'

packages/design-system/src/components/TextField/useLabelMask.tsx
forrestbaer: Phew, yeah that is a complex one but I'll be happy when I understand why it's happening and how to best solve it 💪🏻

PR #2279, opened 2022-11-29
[WNMGDS-2042] Refactor token exporter

zarahzachz:
## Summary

WNMGDS-2042

- Refactor `css-vars` command to export all tokens
- `css-vars` generates Sass map for critical Sass vars
- Remove `sass` from export option, delete Sass exporter file


packages/design-system/src/styles/settings/variables/_core-scss-to-css.map.scss
```diff
@@ -0,0 +1,15 @@
+$grid-columns: 12 !default;
```
forrestbaer, line 1: maybe these filenames could be something more like `layout-tokens.scss` ? The map isn't really relevant anymore with these.

packages/design-system-tokens/src/lib/file.ts
```diff
@@ -27,7 +27,7 @@ export const getAllFiles = (rootPath: string, arrayOfFiles: string[] = []): stri
 };
 
 /**
- * Search out all availible modules under a path and return an
+ * Search out all available modules under a path and return an
```
forrestbaer, line 30: 🙌🏻

forrestbaer (commented):
Looks pretty good outside of a few small comments!

pwolfert (approved):
I just noticed that our `yarn build:tokens` command at the root needs to be updated, but other than that LGTM 🎉

PR #2283, opened 2022-12-01
[WNMGDS-2047] Sass calculation/extra variable cleanup

forrestbaer:
## Summary

- This ticket was supposed to be only for calc cleanup, I found a bunch of calcs that I could clean up and left the ones I thought would remain useful.
- Another thing I did was clean up the **intermediary** sass vars so we can more easily replace them with plain css vars moving forward in this branch
- Updated snapshots which don't look any different to me

## How to test

1. `yarn build && yarn build:healthcare && yarn build:medicare`

## Checklist


packages/design-system/src/styles/components/_Choice.scss
```diff
@@ -243,7 +243,7 @@ label,
   border-left: $spacer-half solid $choice__border-color--left;
   flex: 1 1 100%;
   margin-bottom: $spacer-2;
-  margin-left: calc(calc(#{$choice__size} * 0.5) - calc(#{$spacer-half} * 0.5));
```
pwolfert, line 246: When trying to test this, I realized that we don't use the `ds-c-choice__checkedChild` in our stories. It wasn't until I read the guidance docs that learned it was something we ask our users to add to their `checkedChildren` elements. While I don't love that from a usability perspective, I think we should probably start using it in our stories for now. Right now the visual regression tests aren't going to help us validate this change.

pwolfert: The reason I was testing was that I wanted to visually see why we specifically need a 2px shift. Looking at our example...

Screen Shot 2022-12-01 at 1 21 23 PM

...I think it's actually because of the border on alerts and [healthcare.gov insets](https://design.cms.gov/components/inset?theme=healthcare)! This seems like a weird implementation detail and makes me think that in the future we actually want to work on making this part of the inset class and work on migrating it into the core design system. In other words, this generic class that we tell people to apply to all checked children seems to be more specifically meant for checked children that have a `4px`-wide border that we need to center vertically with the middle of the checkbox or radio button!

CC @line47, @brentschneider

pwolfert: For now, @forrestbaer, could you add a comment above this line about where the `2px` comes from? (You can also convey that we're not 100% certain on the origin. I'm pretty sure it has nothing to do with `$spacer-half` though 😭)

pwolfert (commented):
Only just started reviewing, but I have some comments worth sharing out

packages/design-system/src/styles/components/_StepList.scss
```diff
@@ -30,34 +16,34 @@ $step-button-height: 38px !default;
...
+    border: 2px solid $steplist-step__border-color--default;
     border-radius: $radius-circle;
     box-sizing: border-box;
     color: $steplist-step__color;
     content: counter(step-counter);
     display: block;
-    font-size: $step-list-number-font-size-mobile;
-    height: $step-list-number-size-mobile;
-    left: calc(-1 * (#{$step-list-number-margin} + #{$step-list-number-size-mobile}));
-    line-height: $step-list-number-size-mobile - $step-list-number-border-size * 2;
```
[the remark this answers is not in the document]
pwolfert: But maybe it's not important as long as "it looks good"

packages/design-system/src/styles/components/_Tooltip.scss
```diff
@@ -138,9 +126,9 @@ $tooltip-close-min-width: $spacer-4 !default;
 .ds-c-tooltip {
   background: $tooltip__background-color;
   border-color: $tooltip__border-color;
-  border-radius: $tooltip-border-radius;
+  border-radius: $radius-large;
   border-style: solid;
-  border-width: $tooltip-border-width;
```
pwolfert, line 143: This one seems like it should be a theme variable just because it is repeated so often and seems a shame to turn into a magic number.

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -159,9 +153,9 @@ $usa-banner-font-family: $font-sans;
 }
 
 .ds-c-usa-banner__lock-image {
-  height: 1.5ex;
+  height: 0.75em;
   vertical-align: inherit;
-  width: calc(1.5ex * 52 / 64);
```
pwolfert, line 164: Hey, nice!

packages/design-system/src/styles/components/_FilterChip.scss
```diff
@@ -16,14 +11,14 @@ $filter-chip-icon-size: 10px;
   margin-top: $spacer-1;
   min-height: 0;
   padding: $spacer-half $spacer-1;
-  padding-right: ($spacer-2 + $filter-chip-icon-container-size);
```
pwolfert, line 19: It appears that this should be `34px`, and this discrepency is the cause for the visual regression reference changes for filter chip.

packages/design-system/src/styles/components/_StepList.scss
```diff
@@ -30,34 +16,34 @@ $step-button-height: 38px !default;
...
+    font-size: $font-size-sm;
+    height: $spacer-3;
+    left: -30px;
```
pwolfert, line 32: Not sure if this is the line, but looks like the number is not positioned in the same place it was

step-list-change

forrestbaer:
Hey @pwolfert thanks for the review! I will definitely clean up those px mismatches, thanks for picking them up. 

Regarding breaking out theme variables, I think that would be better handled in another ticket once we have the variables set to css vars to avoid duplicate work, does that sound ok?  Seems like maybe it could even be a meeting with designers to determine what parts of each component we should set those theme variables for.  What do you think @zarahzachz ?

pwolfert:
@forrestbaer, it could certainly be another pull request, but I also wouldn't want to lose any information in the process.

forrestbaer:
> @forrestbaer, it could certainly be another pull request, but I also wouldn't want to lose any information in the process.

Oh sure, fair point, I'll make all the changes.

packages/design-system/src/styles/components/_DateField.scss
```diff
@@ -21,10 +21,7 @@
 
 .ds-c-datefield__separator {
   font-size: $font-size-2xl;
-  // have to do a mix of calc & sass interpolation in order to accommodate values being css vars in internal tools
-  line-height: calc(
-    #{$text-input__line-height} * #{$font-size-base}
-  ); // vertical dimensions should match adjacent inputs
+  line-height: $text-input__line-height * $font-size-base;
```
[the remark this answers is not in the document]
zarahzachz: yeah, when we replace our sass vars with css vars, this formula won't compile correctly so `calc` is needed

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -55,12 +49,12 @@ $usa-banner-font-family: $font-sans;
   align-items: center;
   background-color: $usa-banner-close__background-color;
   display: flex;
-  height: $usa-banner-mobile-close-size;
+  height: 48px;
   justify-content: center;
   margin-left: auto;
   margin-right: -$spacer-2;
```
zarahzachz, line 55: the negatives used in our sass files will need to be evaluated as math i think. when we replace these sass vars with css vars, this won't compile correctly without a `calc`

packages/design-system/src/styles/components/_DateField.scss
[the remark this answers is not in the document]
forrestbaer: Going to swap out spacer half here and add the calc back, at least for the interim it'll work.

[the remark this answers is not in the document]
forrestbaer: Not sure best way to approach this yet.

packages/design-system/src/styles/components/_StepList.scss
```diff
@@ -1,24 +1,10 @@
...
-$step-border-width: 1px !default;
-$step-button-height: 38px !default;
-
 .ds-c-step-list {
   list-style-type: none;
   margin: $spacer-4 0 0;
-  padding: 0 $step-list-padding-h;
-  padding-left: $step-list-padding-left;
+  padding: 0 0;
+  padding-left: 42px;
```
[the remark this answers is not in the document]
forrestbaer: I think we should be more specific and only open up our theme api when we want to allow those things to be modified.. The more of these calculations we do the more chance for side effects I think.

packages/design-system/src/styles/components/_StepList.scss
```diff
@@ -30,34 +16,34 @@ $step-button-height: 38px !default;
...
-    @media (min-width: $step-list-breakpoint) {
-      font-size: $step-list-number-font-size;
-      height: $step-list-number-size;
-      left: calc(-1 * (#{$step-list-number-margin} + #{$step-list-number-size}));
-      line-height: $step-list-number-size - $step-list-number-border-size * 2;
+    top: 7px;
+    width: $spacer-3;
+
+    @media (min-width: $media-width-md) {
+      font-size: 20px;
```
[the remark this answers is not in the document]
forrestbaer: No, 21px is `size-xl`.. maybe we should just make it that..

packages/design-system/src/styles/components/_StepList.scss
```diff
@@ -69,20 +55,20 @@ $step-button-height: 38px !default;
...
+    border-bottom: 1px solid $steplist-step__border-color;
     bottom: 0;
     content: '';
     display: block;
     height: 1px;
-    left: -$step-list-padding-left;
+    left: -42px;
     position: absolute;
-    right: -$step-list-padding-h;
+    right: -0;
```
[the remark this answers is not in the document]
forrestbaer: You've never dealt with -0? It's the grumpy version of 0! 🙄

packages/design-system/src/styles/components/_StepList.scss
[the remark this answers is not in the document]
forrestbaer: I just see it as trouble :)

packages/design-system/src/styles/components/_StepList.scss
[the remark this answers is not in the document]
zarahzachz: doit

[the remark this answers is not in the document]
zarahzachz: imo we should get away from pixel-perfect sizing like this and move towards responsive units.

packages/design-system/src/styles/components/_StepList.scss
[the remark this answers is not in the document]
pwolfert: Good point about side effects. At least now we have regression tests in place to catch when things move around, so if this hard-coded value gets out of sync with something else, we'll be able to see it.

packages/design-system/src/styles/components/_Choice.scss
line47: Could we set up some time to do a demo on this. It would help me understand what's going on here and give proper feedback.

pwolfert: @line47, I moved this particular issue into https://github.com/CMSgov/design-system/pull/2285 and its associated ticket

pwolfert (approved):
The step numbers look better than they did before! I've been looking closely at them 🔬. Nice work!

@zarahzachz have all your items been addressed?

forrestbaer:
> The step numbers look better than they did before! I've been looking closely at them 🔬. Nice work!

Several more changes coming today, more calc removals, working through visual regression results atm.

forrestbaer:
The snapshot updates are due to the changing of the `line-height` in `ds-c-datefield__separator` which gets rid of a multiplication that we won't be able to do in the future.  Sarah and I talked about it and it seems like a good middle-way.

packages/design-system/src/styles/components/_Accordion.scss
```diff
@@ -58,7 +58,7 @@
       forced-color-adjust: none;
       -ms-high-contrast-adjust: none;
       outline: $spacer-half solid LinkText;
-      outline-offset: -$spacer-half;
```
pwolfert, line 64: I think this would have to be converted to `calc(var(--spacer-half) * -1)` once we shift to CSS

packages/design-system/src/styles/components/_Alert.scss
```diff
@@ -6,7 +6,7 @@
   box-sizing: border-box;
   color: $alert__font-color;
   display: flex;
-  min-height: $alert__icon-size + $spacer-2;
```
pwolfert, line 9: I actually don't think we even need this `min-height` property. It seems to be for the case where no content is in the alert, which is not a valid use of our component. The other thing is that when I remove it, it seems to stand up just fine without conent, probably because of structural changes that have happened since this was first written.

Screen Shot 2022-12-06 at 11 22 20 AM

packages/design-system/src/styles/components/_Autocomplete.scss
```diff
@@ -66,6 +66,7 @@
 
 // Need a custom class for bottom error message to make space for the clear search button
 .ds-c-autocomplete__error-message--clear-btn {
+  // @TODO: seems like this could be done differently without calc, perhaps using ch
```
pwolfert, line 69: Nice 👍

packages/design-system/src/styles/components/_DateField.scss
```diff
@@ -21,8 +21,8 @@
 
 .ds-c-datefield__separator {
   font-size: $font-size-2xl;
-  line-height: $text-input__line-height * $font-size-base;
-  margin: calc(#{$text-input__border-width} + #{$spacer-half}) 0;
+  line-height: 1;
```
pwolfert, line 24: Ha! Brilliant

packages/design-system/src/styles/components/_StepList.scss
```diff
@@ -179,7 +179,7 @@
 
   &::before {
     content: '\2022';
-    left: -$spacer-3;
+    left: -#{$spacer-3};
```
pwolfert, line 182: Here's another one. Anyone know of a way to turn a positive number into a negative when it's a CSS prop?

pwolfert (approved):

packages/design-system/src/styles/components/_StepList.scss
forrestbaer: Yeah I did this a few times, I'll run this through the paces to see how it works

forrestbaer: Yep have to do `calc(-1 * var(--cssvar))` or `calc(-1 * #{$sass_var})`

PR #2286, opened 2022-12-02
[WNMGDS-2099] Obfuscate SSNs in label masked TextField

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2099

- Obfuscates social security numbers when the input doesn't have focus by replacing the first five digits with asterisks
- As a side effect of implementing this, we no longer have the cursor-jump problem for label-masked TextFields, but the problem still exists for other TextFields

## How to test

1. Start Storybook
2. Navigate to http://localhost:6006/?path=/story/components-text-field--label-masked-ssn
3. The input should display the full social security number that you input while you're typing, but it will hide the first five digits whenever the input doesn't have focus

## Checklist

### DIFF
--- a/packages/design-system/src/components/TextField/useLabelMask.test.tsx
+++ b/packages/design-system/src/components/TextField/useLabelMask.test.tsx
@@ -2,6 +2,7 @@ import React from 'react';
 import useLabelMask, {
   DATE_MASK,
   SSN_MASK,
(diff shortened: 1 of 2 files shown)


forrestbaer (approved):
Sweet! Looks good, and backspace is fixed, yea! 🎉

PR #2289, opened 2022-12-06
[WNMGDS-1883] Remove font files

zarahzachz:
## Summary

WNMGDS-1883

- Woff2 is supported by all modern browsers ([source](https://caniuse.com/woff2)), so deleting all non-Woff2 font file formats from Core and Mgov dirs
- Added `font-display: swap;` to Core/HCgov font styles to expedite content loading 
- Modified Mgov font weight tokens to support expected weights (only 400 and 600 weights are supported in this DS)
- Added Rubik 600 and Rubik 600 Italic to Mgov dir (only available as TIFF and had to compress as Woff2 using [Google libary](https://github.com/google/woff2))

TODO
- Created WNMGDS-2119 to assign better fallback fonts for browsers >= IE11 + mobile (needs to be prioritized)
- Identified several instances where `font-weight` was hard-coded to `bold` or a specific weight number (**need guidance/input here**)
    - These should be extracted to tokens as our DS themes don't have a 1:1 parity of font weights (Mgov's fattest font is 600, HCgov's is 700)
    - The CSS value `bold` === 700, which is only available in HCgov
    - Alternatively, maybe we should evaluate the weight discrepancy between our themes? HCgov supports both 600 and 700 weights


pwolfert:
> Added `font-display: swap;` to Core/HCgov font styles to expedite content loading

@zarahzachz, could you explain this change some more? My hunch is that we might want to separate it into another pull request with possibly another ticket. Will it affect how the apps look as they load? (Do we need to give stakeholders a heads up that we're making a decision to trade looks for performance?)

forrestbaer (approved):
🧹

pwolfert (approved):
Screenshots look better now with the semibold 👍

PR #2292, opened 2022-12-08
[WNMGDS-2046] Consolidate Styles & Switch to @use over import

forrestbaer:
## Summary

You won't be able to test this PR as this leaves things in a partially broken state until we update all variables to use css vars instead of sass variables. A list of items that need to be completed after this PR as I see it:

1. Update Storybook to utilize appropriate theme files (either load all of them or load them conditionally during build)
2. Ensure appropriate files are `@use`'d where necessary (I've setup these in most places and should be close)

- Updated gulp script to remove `scss` copying to dist and add a method to copy `css` theme files to dist
- Simplified `/styles` structure for core and all child systems
- Instead of importing styles from core `dist` folder for building child system, utilize current core `styles` folder
- Updated a lot of variables which were not using the correct token values
- Updated token `copy_themes.sh` to be more succint and to copy themes to a `styles/themes` folder in each system
- Removed extraneous `index` files where able, moved some variables into theme globals and got rid of `references` files.

## Testing

- Pull branch and poke around in the `src/styles` folders

## Checklist


packages/design-system-tokens/src/copy_themes.sh
```diff
@@ -1,29 +1,20 @@
...
-  done
-  for file in healthcare*.css; do
-    cp -v "$file" "../../ds-healthcare-gov/src/styles/settings/_${file}"
-  done
-  for file in medicare*.scss; do
-    cp -v "$file" "../../ds-medicare-gov/src/styles/settings/variables/_${file}"
-  done
-  for file in medicare*.css; do
-    cp -v "$file" "../../ds-medicare-gov/src/styles/settings/variables/_${file}"
+  for i in "${!PATHS[@]}"; do
```
pwolfert, line 14: Nice 👍

packages/design-system-tokens/src/themes/core.ts
```diff
@@ -135,6 +135,10 @@ export const global: AnyTokenValues = {
   'site-margins-mobile':        '1.5rem',
   'site-max-width':             '1040px',
   'text-max-width':             '53rem',
+  'image-path':                 '"../images"',
+// add an error highlight to choice components inside a ChoiceList with an errorMessage.
+// TODO: remove this, we should not be setting variables like this in sass
+  'ds-include-choice-error-highlight': 'false',
```
pwolfert, line 141: Could you explain a little more why this needed to be added to the theme variables? I don't understand

packages/design-system/src/styles/_base.scss
```diff
@@ -0,0 +1,299 @@
+@use 'mixins' as *;
+
+/**
+ * Global
+ */
+
+// 44 x 44 pixels hit target following Apple iOS Human Interface Guidelines
+$hit-area: 4.4rem !default;
```
pwolfert, line 8: I can't find any references to this in our code. We might be able to remove it

packages/design-system/src/styles/_layout.scss
```diff
@@ -1,4 +1,6 @@
 $grid-columns: 12 !default;
```
pwolfert, line 1: This is out of scope of this ticket, but I think it might be a good idea to generate a doc note at the top of these files to explain that it was generated from the tokens package.

pwolfert: I started working on porting CMS.gov and am now more accutely aware of the stuff that is just magic in our code

pwolfert (commented):
So fresh! I have one question about a variable that was added, but otherwise it LGTM

packages/design-system-tokens/src/themes/core.ts
forrestbaer: I just needed a place to stick it since it didnt seem to fit anywhere.  Maybe a flag, maybe just an override in healthcare.. Open to suggestions.

packages/design-system/src/styles/_layout.scss
forrestbaer: This would be easy to add, i'll work it in here.

packages/design-system-tokens/src/themes/core.ts
pwolfert: Maybe it would be simpler for it to stay [in the child Sass](https://github.com/CMSgov/design-system/blob/acc9e0f67940870462d0e0edd21022bae24921ea/packages/ds-healthcare-gov/src/styles/settings/_override.build.scss#L2) until we can implement that logic solely with CSS variables. As it stands now, we turn it into a token, which then gets converted into Sass only to be used in just Sass. Is the long-term solution to create a variable like `choice__border-color--error` and set it to the same thing as the normal border color in themes where we don't want to show an error highlight?

pwolfert: Oh wait, we currently have that exact variable 😆 . So maybe we don't even need this variable if we update that `choice__border-color--error` in the relevant theme? And we could have a comment about why it's the same color as the normal one.

forrestbaer: Oh dang, perfecto

packages/ds-healthcare-gov/src/styles/components/_ChoiceList.scss
```diff
@@ -0,0 +1,6 @@
+.ds-c-choice {
+  // healthcare is the only place we currently implement a border color for a choice error
+  &.ds-c-choice--error {
+    border-color: $choice__border-color--error;
+  }
```
pwolfert, line 5: I was actually thinking that this would remain in core but without the condition, and then the value of `choice__border-color--error` would be set to the same value as `choice__border-color` for the themes where the feature would have been "off". I think we should minimize the amount of custom css in each theme.

forrestbaer: Ah ok sure I see.

packages/ds-healthcare-gov/src/styles/components/_Footer.scss
```diff
@@ -1,4 +1,6 @@
-@import './Logo';
+@use '../layout' as *;
+@use './Logo' as *;
+@use '@cmsgov/design-system/src/styles/mixins' as *;
```
zarahzachz, line 3: did we want to do the use like this or with a relative import?

zarahzachz, line 3: did we want to do the use like this or with a relative import?

packages/ds-medicare-gov/src/styles/index.scss
```diff
@@ -1,26 +1,23 @@
...
  * a scss (`dist/scss/`) and css (`dist/css/`) dist files.
  */
 
-// Settings: Globally-available variables, mixins, functions and config (Sass code that doesn't output CSS)
-@import '@cmsgov/design-system/dist/scss/settings/index';
-@import 'settings/index';
+@use 'layout';
+@use 'base';
+@use 'fonts';
+@use '@cmsgov/design-system/src/styles/index';
```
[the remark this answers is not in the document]
forrestbaer: Good call, I am not sure how these are going to work once we get to compilation, but it makes sense to set these to static paths for first run.

packages/ds-healthcare-gov/src/styles/components/_Footer.scss
forrestbaer: I'd like to see if we could make an even simpler alias like `@core/mixins` or something.. gonna leave this one in as a test for when we get to compilation.

pwolfert: What's wrong with an import using the package name?

packages/design-system-tokens/src/themes/core-components.ts
```diff
@@ -248,7 +248,7 @@ export const components: AnyTokenValues = {
     '__border-color':                             t.color['base'],
     '__border-color--checked':                    t.color['primary'],
     '__border-color--disabled':                   t.color['gray-light'],
-    '__border-color--error':                      t.color['error'],
+    '__border-color--error':                      t.color['base'],
```
pwolfert, line 251: Would some kind of comment like this be helpful? Or is a theme just full of decisions, and a comment would be noise?

```suggestion
    '__border-color--error':                      t.color['base'], // Don't use error border for choice in this theme
```

pwolfert (approved):
🎉 Nice work!

packages/design-system-tokens/src/themes/core-components.ts
forrestbaer: Sure this works.

PR #2293, opened 2022-12-08
[WNMGDS-2122] Revert use of sass:math to restore support for LibSass in 5.0

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2122

Amid some delinting changes, we accidentally introduced [some Sass code](https://github.com/CMSgov/design-system/commit/f209b1611e94205e0bf83bcd2bf0930b1f373f83#diff-b18f80412afbfbae4bd03e27ba4b4df438eaa1a8874560b8d0d59855bfde619c) that [isn’t supported by LibSass](https://sass-lang.com/documentation/modules/math). Reverts use of `sass:math` only in the `release/5.0` branch.

### DIFF
--- a/packages/design-system/src/styles/settings/mixins/_layout.scss
+++ b/packages/design-system/src/styles/settings/mixins/_layout.scss
@@ -1,5 +1,3 @@
-@use 'sass:math';
-
 @mixin clearfix {
   &::before {
     content: '';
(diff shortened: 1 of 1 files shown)


forrestbaer (approved):

pwolfert:
Looks like the eslint errors we're getting are because the test modules don't import JSX, but that doesn't affect what we ship. I ran the visual regression smoke tests locally, and they passed.

PR #2296, opened 2022-12-13
[WNMGDS-2088] Do our own prop-documentation pre-processing

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2088

The `react-docgen` library through the `gatsby-transformer-react-docgen` plugin doesn't pick up on props that are defined outside of the component's module. It's not really built for TypeScript, and nobody has fixed this issue in that package. There is an alternative library called `react-docgen-typescript` that is a completely separate implementation, and it picks up on all our component props, no matter where they're defined. Rather than create our own Gatsby plugin in the pattern of `gatsby-transformer-react-docgen` that uses `react-docgen-typescript`—and I tried, but it was difficult—I instead decided to extract the prop data as a prerequisite step to building the doc site. It's really easy to use `react-docgen-typescript` and get exactly what we want out of it, so I used it in a new part of the `yarn build` command that writes the prop data to JSON files in our docs package.

- Removes `gatsby-transformer-react-docgen` and code specific to that workflow 
- Adds a step to the build command to extract prop type data from our source code, transform it, and save it to JSON files in the `packages/docs/data` directory
- Changes the `PropTable` component to get its data from those local JSON files instead of a GraphQL static query

## How to test

2. `yarn build && yarn build:healthcare && yarn build:medicare`
3. `yarn start` and open up the local doc site
4. Navigate to component pages and take a look at the prop tables

## Checklist


zarahzachz (commented):
I tried running the build command and it errored out:
> TypeError: Cannot read properties of undefined (reading 'R_OK')

packages/docs/src/components/content/PropTable.tsx
```diff
@@ -45,73 +26,12 @@ interface PropTableProps {
...
-              required
-              tsType
-            }
-          }
-        }
-      }
-    }
-  `);
-
+const PropTable = ({ children, componentName, theme }: PropTableProps) => {
```
forrestbaer, line 29: Ahh, feelsgood 🥇

scripts/extractPropData.js
```diff
@@ -0,0 +1,85 @@
...
+
+function getOutputFilename(rootPath) {
+  const parts = rootPath.split('/');
+  const packageName = parts[parts.length - 1];
+  return `props-${packageName}.json`;
+}
+
+async function extractData(fileName) {
+  // Will throw if this file doesn't exist
+  await access(fileName, constants.R_OK);
```
forrestbaer, line 31: Probably just want to remove this, throwing errors for Sarah and I.

scripts/extractPropData.js
```diff
@@ -0,0 +1,85 @@
...
+      if (prop.description) {
+        prop.description = marked.parseInline(prop.description);
+      }
+    }
+  }
+}
+
+const corePackage = path.join('packages', 'design-system');
+module.exports = async function (rootPath = corePackage) {
+  const outputPath = path.join('packages', 'docs', 'data', getOutputFilename(rootPath));
```
forrestbaer, line 64: Might want to check for existence / create packages/docs/data directory here if it doesn't exist.

forrestbaer (commented):
Works great, just a few comments!

pwolfert:
I guess now the CI tests require that the child design systems be built before they can run `yarn type-check`, which will add about two minutes to the build. This is because `yarn type-check` expects those imported json files to all be there.

Alternatives:
1. Do all the type-checking when Gatsby starts up [and pass the data down from the page component from the page context object](https://www.gatsbyjs.com/docs/how-to/querying-data/using-gatsby-without-graphql/)
2. Separate this data extraction into its own command separate from build and add that to the list of prerequisite steps to building gatsby along with `yarn build:storybook:docs`

pwolfert:
(The last commit only passed the CI tests because I re-ran it with a modification to build all design systems)

forrestbaer:
> 2\. Separate this data extraction into its own command separate from build and add that to the list of prerequisite steps to building gatsby along with `yarn build:storybook:docs`

I would go with this option.

pwolfert:
I didn't want to have to change our CI job, so I separated the command but made it run after the build script so we don't have to remember to run it separately before typechecking with `yarn type-check`. If it doesn't build that json file prior to that, TypeScript will complain that the file doesn't exist when we try to import it in the `PropTable` module.

zarahzachz (approved):
I was able to run the code and saw all props available - LGTM

forrestbaer (approved):
Looks good!

PR #2299, opened 2022-12-14
[WNMGDS-2049] Create and document focus ring utility classes

zarahzachz:
## Summary

WNMGDS-2049

I made this PR from `main` instead of the CSS only feature branch so we could test visuals (currently the CSS only branch build isn't compiling correctly, which will be resolved after Sass vars are reassigned). Since I'm not deleting mixins, I thought this was OK to do, but can change that if you feel otherwise.

- Create utility classes for:
    - `.ds-u-focus`: Applied when `:focus` state active
    - `.ds-u-focus-within`: Applied when `:focus-within` condition met
    - `.ds-u-focus-visible`: Applied when `:focus-visible` condition met
- Added documentation to show usage

Tried deploying [a demo page](https://cmsgov.github.io/design-system/branch/WNMGDS-2049/focus-class/utilities/focus/?theme=core) for testing, but getting a 404 😢


packages/docs/content/utilities/ring.mdx
```diff
@@ -0,0 +1,56 @@
...
+### Focus rings
+
+**Focus**
+
+Use the `.focus:ring` utility to conditionally apply focus styles to your component when that component has focus. Focus will be applied regardless of keyboard or mouse interaction.
+
+Click or tab into the following example to see this utility in action.
+
+<EmbeddedExample>
+  <p tabIndex="0" class="focus:ring ds-u-border--1 ds-u-padding--2">Focus state example</p>
```
zarahzachz, line 20: I struggled to create real world examples to showcase these utils. All the focussable elements in the DS already have focus styles baked in, so it was difficult showcasing the different behaviors of each utility.

packages/design-system/src/styles/utilities/_ring.scss
```diff
@@ -0,0 +1,14 @@
+@use 'sass:math';
+
+.ring,
+.focus\:ring:focus,
+.focus-within\:ring:focus-within,
+.focus-visible\:ring:focus-visible {
```
pwolfert, line 6: While this method makes it flexible enough to use anywhere, I don't actually think we have a use case at the moment for applying it to unfocused elements, and I don't think we want to encourage adding and removing the classes programmatically when the `:focus-within` state should cover pretty much everything. Even though having the extra class doesn't harm anything, it does make it tricky to name. I would actually think this class is more discoverable if it's named "focus" instead of "ring", but the word "ring" is needed if "focus" is taken by ":focus". What about the following names?
- `ds-u-focus` (corresponds to `:focus`)
- `ds-u-focus-witin` (corresponds to `:focus-within`)
- `ds-u-focus-visible` (corresponds to `:focus-visible`) (and adding `:focus-visible` was a good idea 👍)

pwolfert: @line47, do you have an opinion on the utility class naming?

packages/docs/content/utilities/ring.mdx
pwolfert: I'll try to think of a custom element that has been produced in downstream products

packages/docs/content/utilities/ring.mdx
```diff
@@ -0,0 +1,56 @@
...
+
+**Focus-visible**
+
+Use the `.focus-visible:ring` utility to conditionally apply focus styles to your component when that component has keyboard focus.
+
+Tab into the following example to see this utility in action.
+
+<EmbeddedExample>
+  <p tabIndex="0" class="focus-visible:ring ds-u-border--1 ds-u-padding--2">Focus-visible state example</p>
+</EmbeddedExample>
```
pwolfert, line 31: I can't seem to tab into this one. Maybe it's because it's a paragraph? Thinking of another possible example with an interactive element

pwolfert: Never mind. User error

packages/docs/content/utilities/ring.mdx
pwolfert: Still not a real-world example, but I was looking through possible form elements, and I thought this might be cool:
```html
<input type="color" name="color-focus" value="#ff0000" class="focus:ring ds-u-radius" />
```

packages/design-system/src/styles/utilities/_ring.scss
zarahzachz: The only instance I could think of where a generic `.ring`/`.focus` class would be useful was in instances we were applying focus on non-interactive elements (I think we did this with heading text in Drawer and Dialog), but that pattern seems to be getting deprecated, so I agree removing it's a good idea.

forrestbaer (approved):
Looks good!

PR #2305, opened 2022-12-20
[WNMGDS-2115] Sass conversion to use CSS Custom Properties

forrestbaer:
## Summary

- Updated all core/hc/mc/docsite/storybook's scss to utilize css vars instead of sass variables, only sass variables to remain are in the `_layout.scss` files in each system
- Updated build system & tokens in many places to accomidate this change.
- Now copying theme files to `.storybook/static` and `docs/static` for use in storybook/doc site
- Updated doc site method for changing theme variables
- Updated storybook method for changing theme variables
- Updated snapshots
    - Some snapshots I was unable to fullly understand why they were different, the differences in some cases seem to have resolved some visual things that I personally think looked wrong to begin with, such as the distance between header and content in the modals and the size of the spinner icon in the spinner icon buttons. In those cases, the appropriate selector appears to be being used where it wasn't previously.. Anyway will need to go over these a bit more to confirm with designers.
- Several small visual bugfixes along the way
- Adds a new script for building storybook `scripts/build-storybook.sh` which moves static assets appropriately when building storybook for the docs/gatsby site.

## How to test

1. `yarn clean && yarn install && yarn build && yarn build:healthcare && yarn build:medicare`
2. `yarn build:storybook && yarn build:storybook:docs && yarn build:docs && yarn serve:docs`
3. `yarn storybook` `npx http-server ./storybook-static`
4. Evaluate snapshot updates.
5. Evaluate css selector sequencing in dev tools for snapshot updated components next. to `design.cms.gov/storybook` versions.
6. `yarn test`

## Checklist


packages/design-system-tokens/src/copy_themes.sh
```diff
@@ -13,16 +13,15 @@ copyThemes()
...
-    cp -v "$file" "../../docs/src/styles/theme-variables/_${file}"
+    rm "${PATHS[$i]}/${i}-theme.css"
+    echo ":root, ::before, ::after, ::backdrop {" >> "${PATHS[$i]}/${i}-theme.css"
+    cat "${i}-theme.css" >> "${PATHS[$i]}/${i}-theme.css"
+    cat "${i}-components-theme.css" >> "${PATHS[$i]}/${i}-theme.css"
+    echo "}" >> "${PATHS[$i]}/${i}-theme.css"
+    # copy to storybook static for storybook building
+    cp -v "${PATHS[$i]}/${i}-theme.css" ../../../.storybook/static
+    mkdir -p ../../docs/static/themes
+    cp -v "${PATHS[$i]}/${i}-theme.css" ../../docs/static/themes
```
pwolfert, line 24: It might be easier to read if `"${PATHS[$i]}/${i}-theme.css"` were turned into a local variable. It looks like it's being used six times, so it could be justified.

packages/design-system-tokens/src/themes/core-components.ts
```diff
@@ -249,6 +249,7 @@ export const components: AnyTokenValues = {
     '__border-color--checked':                    t.color['primary'],
     '__border-color--disabled':                   t.color['gray-light'],
     '__border-color--error':                      t.color['base'], // Don't use error border for choice in this theme
+    '__border-color--error--inverse':             t.color['white'],
```
pwolfert, line 252: 👍

packages/design-system-tokens/src/themes/medicare-components.ts
```diff
@@ -348,7 +349,7 @@ export const components: AnyTokenValues = {
   },
   
   'review': {
-    'review__border-color':                       t.color['border'],
+    '__border-color':                       t.color['border'],
```
pwolfert, line 352: Nice catch

packages/docs/src/components/layout/Layout.tsx
```diff
@@ -68,6 +68,12 @@ const Layout = ({
       >
         <script>{`window.tealiumEnvironment = "${env}";`}</script>
         <script src="//tags.tiqcdn.com/utag/cmsgov/cms-design/prod/utag.sync.js"></script>
+        <link
+          rel="stylesheet"
+          type="text/css"
+          title="docThemeCss"
+          href={`/themes/${theme}-theme.css`}
```
pwolfert, line 75: Oh hey, the fact that this is declarative makes this nice and simple

packages/docs/src/styles/index.scss
```diff
@@ -1,29 +1,30 @@
...
+// @import 'theme-variables/_core-components-theme.css';
+// // brand themes will use core variables if the theme does not explicitly define that variable
+// @import 'theme-variables/_healthcare-scss-to-css.map.scss';
+// @import 'theme-variables/_healthcare-theme.css';
+// @import 'theme-variables/_healthcare-components-scss-to-css.map.scss';
+// @import 'theme-variables/_healthcare-components-theme.css';
+// @import 'theme-variables/_medicare-scss-to-css.map.scss';
+// @import 'theme-variables/_medicare-theme.css';
+// @import 'theme-variables/_medicare-components-scss-to-css.map.scss';
+// @import 'theme-variables/_medicare-components-theme.css';
```
pwolfert, line 17: Can these lines be removed now?

packages/docs/src/styles/index.scss
```diff
@@ -1,29 +1,30 @@
...
-[data-theme='healthcare'] {
-  // Have to set the image path so that the error label's svg can be found
-  $image-path: '@cmsgov/ds-healthcare-gov/src/images';
-  @import '@cmsgov/ds-healthcare-gov/src/styles/components/index';
-}
+// [data-theme='healthcare'] {
+//   // Have to set the image path so that the error label's svg can be found
+//   $image-path: '@cmsgov/ds-healthcare-gov/src/images';
+//   @import '@cmsgov/ds-healthcare-gov/src/styles/index';
+// }
```
pwolfert, line 27: Checking my understanding: We don't need to worry about including healthcare- or medicare-specific CSS in the doc site CSS because we don't directly embed any of those components and instead rely on the storybook styles being complete, right?

pwolfert:
I can see how the spinner buttons were not following the intention of 

```scss
.ds-c-button > & {
    @include inline-icon;
    @include spinner-size(1em, 1em);
  }
```

because of the old namespaced theme rules having greater specificity! I'm glad this is fixed!

Screen Shot 2022-12-20 at 11 37 33 AM

pwolfert (commented):
I'm not finished going through this, but here are my notes so far. This was a big project!

pwolfert:
Screen Shot 2022-12-20 at 11 52 28 AM

I haven't dug into why, but the little arrow icons aren't showing for me on the dropdowns. I just stumbled upon that randomly. Might have to do with the image path variable.

packages/docs/src/styles/index.scss
forrestbaer: Just the fonts, I will look closer at fonts in this regard tomorrow.

packages/design-system-tokens/src/copy_themes.sh
forrestbaer: > I haven't dug into why, but the little arrow icons aren't showing for me on the dropdowns. I just stumbled upon that randomly. Might have to do with the image path variable.

Hmm that's odd, they're showing up for me.

image

forrestbaer:
I added a new script `scripts/build-storybook.sh` and set up the npm scripts to use it. This script moves the `images` and `fonts` folders when building storybook for the docs site into the `docs/static/` folder so they can be referenced by storybook stories and gatsby in a central location.

I've also updated the aliases for gatsby `images` and `fonts` to use these static assets for building. This seems to resolve an issue I was running into where storybook stories were trying to reference one place and gatsby itself was looking elsewhere, I believe this is the most straightforward way to manage this issue with static assets without needing to have multiple copies of them all over the place.

packages/design-system/src/styles/components/_TooltipIcon.scss
```diff
@@ -40,13 +42,13 @@
 .ds-c-tooltip__trigger:hover,
 .ds-c-tooltip__trigger:focus {
   .ds-c-tooltip-icon__container {
-    border-color: $tooltip-icon__color;
-    border-color: rgba($tooltip-icon__color, 0.25);
+    border-color: var(--tooltip-icon__color);
+    border-color: rgba(var(--tooltip-icon__color), 0.25);
```
zarahzachz, line 46: this color function shouldn't work with css vars - there's goofiness in how the compiled var value fits in with this function. 

sass allows for these color functions to work with sass vars, css props don't.

zarahzachz: [SO explains it better than me](https://stackoverflow.com/a/41265350)

packages/design-system/src/styles/components/_TooltipIcon.scss
```diff
@@ -56,8 +58,8 @@
   &.ds-c-tooltip__trigger:hover,
   &.ds-c-tooltip__trigger:focus {
     .ds-c-tooltip-icon__container {
-      border-color: $tooltip-icon__color--inverse;
-      border-color: rgba($tooltip-icon__color--inverse, 0.25);
+      border-color: var(--tooltip-icon__color--inverse);
+      border-color: rgba(var(--tooltip-icon__color--inverse), 0.25);
```
zarahzachz, line 62: see "issues with css vars and color function" comment above

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -14,7 +15,7 @@
   display: flex;
   flex-direction: row;
   font-size: 12px;
```
zarahzachz, line 17: weird this wasn't a var

packages/design-system/src/styles/utilities/_grid.scss
```diff
@@ -9,13 +10,13 @@
 .ds-l-container {
   margin-left: auto;
   margin-right: auto;
-  max-width: $site-max-width;
-  padding-left: calc(#{$grid-gutter-width} * 0.5);
-  padding-right: calc(#{$grid-gutter-width} * 0.5);
+  max-width: var(--site-max-width);
+  padding-left: calc(var(--grid-gutter-width) * 0.5);
```
zarahzachz, line 14: the unitless division here seems to work! css is a mystery sometimes

zarahzachz (commented):
Screen Shot 2022-12-22 at 5 13 59 PM
the border seems to have disappeared for the review pattern but i can't pinpoint why 🤔

zarahzachz:
i don't know if this is helpful or not, but for some of the css/scss stuff i reviewed, i ran the css solution through [codepen](https://codepen.io/your-work) (and used the computed panel in chrome dev tools on the demo) and the original sass implementation through [sassmeister](https://www.sassmeister.com/) just to compare what the outcomes were. this was how i spotted some discrepancies in `calc()` and color functions

packages/design-system/src/styles/components/_UsaBanner.scss
forrestbaer: We don't have a font value for 12px in tokens.

packages/design-system/src/styles/components/_TooltipIcon.scss
forrestbaer: Interesting, and too bad! I've updated these variables to use new token variables which are computed during compilation.

forrestbaer:
I believe all those issues are resolved @zarahzachz , thanks for the review, good callouts!

packages/design-system-tokens/src/copy_themes.sh
pwolfert: I cleaned and built again, and it wasn't a problem anymore. False alaram

scripts/build-storybook.sh
```diff
@@ -0,0 +1,21 @@
...
+
+if [ "$OUT" = "docs" ]; then
+  build-storybook -o packages/docs/static/storybook && \
+  echo "${GREEN}info${NC}: Moving static assets for use with gatsby."
+  mkdir -p packages/docs/static/fonts && \
+  mkdir -p packages/docs/static/images && \
+  cp -R packages/docs/static/storybook/fonts/* packages/docs/static/fonts/ && \
+  cp -R packages/docs/static/storybook/images/* packages/docs/static/images/ && \
+  rm -rf packages/docs/static/storybook/fonts && \
+  rm -rf packages/docs/static/storybook/images
```
pwolfert, line 18: Why do we remove these directories? Or why don't we need them?

scripts/build-storybook.sh
```diff
@@ -0,0 +1,21 @@
...
+
+OUT="$1"
+
+if [ "$OUT" = "docs" ]; then
+  build-storybook -o packages/docs/static/storybook && \
+  echo "${GREEN}info${NC}: Moving static assets for use with gatsby."
+  mkdir -p packages/docs/static/fonts && \
+  mkdir -p packages/docs/static/images && \
+  cp -R packages/docs/static/storybook/fonts/* packages/docs/static/fonts/ && \
+  cp -R packages/docs/static/storybook/images/* packages/docs/static/images/ && \
```
pwolfert, line 16: Should we consider ignoring these folders in git so we don't get so many duplicate files in version control?

pwolfert: We ignore the `static/storybook` folder already

pwolfert (commented):
I'm noticing some fairly substantial spacing changes in the medicare footer. Do you know what's causing that?

scripts/build-storybook.sh
forrestbaer: I suppose, will just need to make sure to not ignore them if any updates are made there.

scripts/build-storybook.sh
forrestbaer: I wanted to make sure the content in them was only what we want to copy over.

forrestbaer:
@pwolfert looks like an `@extend` was causing some trouble there, I'm replacing that extend with the appropriate vars now and it's working. Updating snapshots.

pwolfert (approved):
Now that they've been committed, I think you now need to explicitly remove those image and font files from git even though they're being ignored for future changes. Everything else LGTM 🎉

PR #2306, opened 2022-12-20
[NO-TICKET] getting rid of defaultProps in stories

forrestbaer:
## Summary

- Was getting a bunch of warnings in storybook about `defaultProps` deprecation so figured I'd clean it up
- Removes `defaultProps` in any stories which use them, instead setting them manually with `args`

**Information on the deprecation here:**
[https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#no-longer-inferring-default-values-of-args](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#no-longer-inferring-default-values-of-args
)

## How to test

1. `yarn clean && yarn install && yarn storybook`
2. Shouldn't see any more warnings about this deprecation

### DIFF
--- a/packages/design-system/src/components/DateField/SingleInputDateField.stories.jsx
+++ b/packages/design-system/src/components/DateField/SingleInputDateField.stories.jsx
@@ -11,7 +11,6 @@ export default {
       control: { type: 'text' },
     },
     errorPlacement: {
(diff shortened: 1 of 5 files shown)


pwolfert (approved):
🧹

PR #2308, opened 2022-12-27
[WNMGDS-1343] Type clamp discovery

zarahzachz:
## Summary

WNMGDS-1343

Removing redundant font styles from DS.

Styles from `.ds-base` should not be duplicated in child elements _unless_ needed to overwrite user agent styles (like `<button>`, `<h1>`, etc.). For the exceptions, I added comments to the style saying why it was needed.


pwolfert (approved):
Nice, this is helpful. Maybe we should merge it into the CSS feature branch, though, and use CSS vars so we don't forget to do that after that feature branch gets merged.

packages/design-system/src/components/DateField/SingleInputDateField.stories.jsx
```diff
@@ -28,8 +27,9 @@ export default {
     },
   },
   args: {
-    label: 'Birthday',
+    errorPlacement: 'top',
```
forrestbaer, line 30: Nice 👍🏻

forrestbaer (approved):
Nice, the comments are helpful for places where we need the font setting.

PR #2310, opened 2022-12-28
[WNMGDS-2135] Fix language detection bug

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2135

Fixes bug where language is detected and stored with its full [Unicode BCP 47 locale identifier](https://www.unicode.org/reports/tr35/#Canonical_Unicode_Locale_Identifiers) instead of only the language tag, which our `i18n` module expects elsewhere.

## How to test

I modified the `examples/cdn` example to test this. I changed the HTML `lang` attribute to `en-US` and to `es-US`. I added the healthcare.gov header and witnessed the bug using the code from `main` and then saw it fixed when switching to a bundle built from this branch.

## Checklist

### DIFF
--- a/packages/design-system/src/components/i18n.test.ts
+++ b/packages/design-system/src/components/i18n.test.ts
@@ -1,4 +1,5 @@
 import {
+  _detectDocumentLanguage,
   getLanguage,
   setLanguage,
   getTranslations,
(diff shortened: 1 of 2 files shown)


forrestbaer (approved):
Looks good!

PR #2312, opened 2022-12-29
[LHDM-1387] Expose Privacy Settings Modal for standalone usage

seesleestak:
## Summary

- Exporting the `PrivacySettingsLink` component so it can be consumed alone, outside of its usage in the footer.
- https://jira.cms.gov/browse/LHDM-1387

## How to test

For testing, I'd mostly done so in the consuming application to ensure the changes would work for my use case. However, the `PrivacySettingsLink` component should be able to be added in any other storybook component instance and work on its own.

## Checklist

### DIFF
--- a/packages/ds-healthcare-gov/src/components/Footer/__snapshots__/InlineLinkLists.test.tsx.snap
+++ b/packages/ds-healthcare-gov/src/components/Footer/__snapshots__/InlineLinkLists.test.tsx.snap
@@ -76,7 +76,7 @@ exports[`InlineLinkLists renders lists of links 1`] = `
         class="hc-c-footer__inline-item ds-u-margin-y--0 ds-u-display--inline-block"
       >
         <button>
-          footer.privacySettings
+          Privacy Settings
         </button>
       </li>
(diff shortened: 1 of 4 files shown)


packages/ds-healthcare-gov/src/components/Footer/__snapshots__/InlineLinkLists.test.tsx.snap
```diff
@@ -76,7 +76,7 @@ exports[`InlineLinkLists renders lists of links 1`] = `
         class="hc-c-footer__inline-item ds-u-margin-y--0 ds-u-display--inline-block"
       >
         <button>
-          footer.privacySettings
+          Privacy Settings
```
pwolfert, line 79: Note to self: We may want to go in and remove `t` props in favor of just calling our translation hook inside these components. It would make the tests simpler and be less data to pass around. All the other labels in this snapshot are produced by a dumb mock translation function.

pwolfert (approved):
LGTM 👍

PR #2313, opened 2022-12-30
[LHDM-1387] Privacy Settings Custom Trigger

seesleestak:
## Summary

- Getting some work I missed in #2312. Adding the option to swap out the text rendered for the PrivacySettings modal trigger.
- https://jira.cms.gov/browse/LHDM-1387

## How to test

1. I've created a storybook page for `PrivacySettingsLink` to demonstrate the functionality

## Checklist


packages/ds-healthcare-gov/src/components/Footer/PrivacySettingsLink.test.tsx
```diff
@@ -0,0 +1,17 @@
...
+    const { container } = render(<PrivacySettingsLink />);
+    expect(container).toMatchSnapshot();
+  });
+
+  it('should render with custom class and children', () => {
+    const { container } = render(
+      <PrivacySettingsLink className="ds-t-test-class">Custom modal trigger</PrivacySettingsLink>
+    );
+    expect(container).toMatchSnapshot();
+  });
```
seesleestak, line 16: The default case is maybe a bit redundant since it's covered in the `Footer` snapshots, but since the custom class and children is specific to usage outside the footer I figured the component needed it's own test file.

pwolfert (approved):
Nice job getting the visual regression tests running! However, I don't think we really need visual regression tests for this, so I made a commit that ignores them. The way we do that right now isn't very intuitive, I'm afraid.

pwolfert:
CI tests might fail because I've modified them for https://github.com/CMSgov/design-system/pull/2314

seesleestak:
> CI tests might fail because I've modified them for #2314

Ah I see, understood.

PR #2320, opened 2023-01-04
[WNMGDS-2139] Sass token distribution & Build Updates

forrestbaer:
## Summary

[WNMGDS-2139](https://jira.cms.gov/browse/WNMGDS-2139) 

- Updates to token system to build Sass token variables and pass them on to `dist` folders during build
- A couple small naming changes in token scripts
- Removes datestamp for token-system-generated files, too much noise updating these every time. I will have another bit of work to re-add this functionality in the future as a custom git commit hook instead so the datestamp is applied only when those files are different than the current checked in versions.

## How to test

2. Check /dist/scss folders for token scss
3. Any other standard tests

## Checklist


pwolfert (commented):
The code looks good, but I'm getting a Gatsby build error now that I don't get on the base feature branch:

```
Module not found: Error: Can't resolve './core-components' in
'/Users/patrickwolfert/Projects/design-system/packages/design-system-tokens/src/themes'
```

forrestbaer:
> The code looks good, but I'm getting a Gatsby build error now that I don't get on the base feature branch:

Should be fixed, the theme index was pointing to the wrong filename.

pwolfert (approved):
Yep, that fixed it! LGTM 👍

PR #2330, opened 2023-01-10
[WNMGDS-2156] Use SVGs for color swatches in docs

pwolfert:
## Summary

- Change the style of the color visualization for the system tokens page to be more appropriate for the color tokens, where we want to show a range of shades
- Use SVGs instead of background color, because the colors will still show up in forced-colors mode
- Apply SVG technique to existing color swatches

## How to test

[Demo page](https://cmsgov.github.io/design-system/branch/pwolfert/svg-swatches/foundation/system-color-tokens/)

## Checklist


line47:
This looks great! Thanks @pwolfert !

line47 (approved):
LGTM

packages/docs/src/styles/components/colorSwatchList.scss
```diff
@@ -12,3 +12,44 @@
   vertical-align: middle;
   width: $spacer-4;
 }
+
+.c-color-ramp__wrapper {
+  flex-shrink: 0;
+  margin-right: $grid-gutter-width;
+  width: 256px;
```
zarahzachz, line 19: **Alt solution**
If you applied `gap: $grid-gutter-width` to this element's parent `div`, you could get away with the following:
```css
.c-color-ramp__wrapper {
  flex-basis: 256px;
}
```

zarahzachz (commented):
I noticed some strange clipping in Firefox (small gap between swatches with transparency visible). Fiddled around in CSS to see if there was an obvious fix and couldn't immediately see one

packages/docs/src/components/content/ColorRamps.tsx
```diff
@@ -0,0 +1,66 @@
...
+/**
+ *
+ */
+const ColorRamps = () => (
+  <div className="ds-u-display--flex ds-u-flex-direction--row ds-u-flex-wrap--wrap">
+    {Object.entries(swatches).map(([swatchName, swatchColors], i) => {
+      const patternId = `pattern-checkers-${i}`;
+      return (
+        <div className="c-color-ramp__wrapper" key={swatchName}>
+          <h2 className="ds-u-text-transform--capitalize" id={`hello-${i}`}>
```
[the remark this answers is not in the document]
pwolfert: That was left over from something I was trying out. Will remove

packages/docs/src/components/content/ColorRamps.tsx
```diff
@@ -0,0 +1,66 @@
...
+  value: string;
+}
+
+const swatches: Record<string, SwatchColor[]> = Object.keys(colorTokens).reduce((swatches, key) => {
+  const colorName = key.split('-')[0];
+  swatches[colorName] = [...(swatches[colorName] ?? []), { name: key, value: colorTokens[key] }];
+  return swatches;
+}, {});
+
+const renderTransparencyPattern = (id: string) => (
```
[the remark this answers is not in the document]
pwolfert: Unfortunately the bleed through still occurs in Firefox when I use the background gradient. And it has this little artifact:
Screen Shot 2023-01-10 at 9 37 15 AM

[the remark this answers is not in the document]
pwolfert: I did just find that if I kind of pad the height of the svg rectangles, though, it fixes it. I had tried that long ago but still had issues due to the `clip-path` on individual items.

packages/docs/src/styles/components/colorSwatchList.scss
pwolfert: Nice! Thanks

pwolfert:
I just realized that something is causing the page to scroll horizontally. I think it's my wide SVGs. Even though I clip them off visually, they're still causing the page to widen.

pwolfert:
Fixed in https://github.com/CMSgov/design-system/pull/2330/commits/72806b37cfa5a24af1f3fe4c75ecb628ca1941e2 by making assumptions, which makes the hacks more fragile.

line47:
I'm noticing a lot of space at the bottom of the page under the footer - not sure where that's coming from

pwolfert:
@line47, I haven't redeployed the demo since my last commit that fixed it. After we decide what to do about the alternative text rendering strategy, I'll redeploy

packages/design-system-tokens/src/lib/utility.ts
```diff
@@ -25,17 +25,62 @@ export const hexOpacity = (hexVal: HexValue | any, opacity: number): HexValue =>
  * @returns An array in the format [R: number, G: number, B: number] or null if there was an error
  */
 export const hexToRgbArray = (hex: HexValue): number[] | null => {
-  const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})$/i.exec(hex);
+  const result = /^#?([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})([a-f\d]{2})?$/i.exec(hex);
```
forrestbaer, line 28: Nice 🥇

packages/docs/src/helpers/themeTokens.ts
```diff
@@ -0,0 +1,34 @@
+import colorTokens from 'design-system-tokens/src/tokens/color';
```
forrestbaer, line 1: Nice utilities!

forrestbaer (approved):
Looks great all around. Tested locally.

packages/docs/src/styles/components/colorSwatchList.scss
```diff
@@ -12,3 +12,61 @@
...
+
+  svg {
+    left: 0;
+    position: absolute;
+    top: 0;
+    width: 100%;
+  }
+
+  code {
+    font-family: monospace;
```
zarahzachz, line 69: Might wanna use the same monospace font-family rule we're using for the [label mask](https://github.com/CMSgov/design-system/blob/7c6466da14cf730d030d036a4fe442bdbcef3cbc/packages/design-system/src/styles/components/_LabelMask.scss#L3)?

packages/docs/src/helpers/themeTokens.ts
pwolfert: I forgot to give you credit for most of it!

packages/docs/src/styles/components/colorSwatchList.scss
pwolfert: Thanks for pointing it out. I don't need it anymore! I was previously selecting `span` and changed it to `code`, but `code` has its own font-family setting.

PR #2339, opened 2023-01-16
[NO-TICKET] Adding child system scss back to doc site build

forrestbaer:
## Summary

- Doc site wasn't adding the child systems styles where they needed to be.
- This was removed in the epic css branch and shouldn't have been

## How to test

1. `yarn clean && yarn install && yarn start`

### DIFF
--- a/packages/docs/src/styles/index.scss
+++ b/packages/docs/src/styles/index.scss
@@ -25,3 +25,11 @@
 @import 'components/tableOfContents.scss';
 @import 'components/spacingUtilityExampleList.scss';
 @import 'components/video.scss';
+
+[data-theme='healthcare'] {
+  @import '@cmsgov/ds-healthcare-gov/src/styles/index';
+}
+
+[data-theme='medicare'] {
+  @import '@cmsgov/ds-medicare-gov/src/styles/index';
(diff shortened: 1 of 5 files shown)


packages/docs/src/styles/index.scss
```diff
@@ -25,3 +25,11 @@
 @import 'components/tableOfContents.scss';
 @import 'components/spacingUtilityExampleList.scss';
 @import 'components/video.scss';
+
+[data-theme='healthcare'] {
+  @import '@cmsgov/ds-healthcare-gov/src/styles/index';
```
pwolfert, line 30: Previously it only imported the components index, which was [the equivalent of importing the seven healthcare style modules](https://github.com/CMSgov/design-system/blob/main/packages/ds-healthcare-gov/src/styles/index.scss#L13-L19). Right now this is including the whole set of DS styles again under the namespace. If we can be more surgical here, we can save on a lot of bytes.

forrestbaer: You're right, this was a lazy addition, updating

pwolfert (approved):
Looks great!

PR #2341, opened 2023-01-17
[NO-TICKET] Don't delete dist/scss folder during build

forrestbaer:
## Summary

This will be cherry picked into the 6.0 release for another beta.

- Adds some logic to gulp build to leave the dist/scss folder alone so it doesn't clean it out during build.
- Removes extra `build:tokens` script from package.json
- Removing subdirectories as well in token dist folder in cases where subfolders may be there

## How to test

1. `yarn build` - check for existence of `dist/scss` for each child system, should be there now.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -37,12 +37,14 @@ const fontsCorePath = path.join(corePackageFiles, 'fonts');
  * clean up dist folder if it exists
  */
 const cleanDist = (cb) => {
-  fs.readdirSync(distPath, (err, files) => {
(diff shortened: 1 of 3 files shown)


pwolfert (approved):
It works 👍 

I was thinking about whether we should concatenate the `*-tokens.scss` and `*-component-tokens.scss` file for ease of use, but maybe it'll aid in their transition when they find they reach a stage where they can stop importing `*-component-tokens.scss` and only import the base tokens.

PR #2343, opened 2023-01-19
[NO-TICKET] Show an example of importing our CSS from the npm package

pwolfert:
## Summary

There is currently a problem with the strategy of importing our new CSS files from a downstream WebPack project's Sass files in that WebPack won't correctly locate the font and image files. That is why teams used to have to set the `$image-path` and `$font-path` variables. This example now shows that a create-react-app project can import the CSS files from a JavaScript module and will correctly find and copy the font and image files to the build directory automatically.

I'll be updating our dev documentation next in another branch.

## How to test

1. `cd examples/create-react-app-typescript`
2. `yarn install && yarn build && http-server build`

### DIFF
--- a/examples/create-react-app-typescript/package.json
+++ b/examples/create-react-app-typescript/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.0",
   "private": true,
   "dependencies": {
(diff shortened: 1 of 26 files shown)


forrestbaer (approved):
Nice 🧹

PR #2347, opened 2023-01-19
[NO-TICKET] Add some notes about importing css to dev docs

pwolfert:
## Summary

See the added note for details. This isn't really an option for some teams, but for teams that don't need it to be interpreted as Sass, it could help.


forrestbaer (approved):
📖

zarahzachz (approved):
Looks good, but could it be more helpful if it was presented as an Alert (info or lightweight type)?

pwolfert:
Good idea, @zarahzachz. Made the change

PR #2348, opened 2023-01-19
[WNMGDS-2162] Add `ds-content` CSS class

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2162

Some teams rely on the ability to `@extend` our heading classes to apply those styles to the base heading elements through their local project SCSS. While this is still possible for teams importing our CSS through SCSS, some bundler configs have problems finding our font and image assets when interpretting our CSS as SCSS. This change introduces a `ds-content` class in the style of [the USWDS's "usa-prose" class](https://designsystem.digital.gov/components/prose/) that applies our heading styles to the appropriate heading levels based on what we and the app teams consider sane defaults (since we divorced our heading styles from the semantic heading numbers).

## How to test

1. View the doc-site changes on the `/foundation/typography/headings/` page
2. Try out the Storybook story

## Checklist


forrestbaer (approved):
Nice fix!

packages/design-system/src/components/utilities/typographyHeadings.stories.tsx
```diff
@@ -33,6 +33,17 @@ AllHeadingsOnDark.parameters = {
   baseInverse: true,
 };
 
+export const ContentClass = () => (
+  <div className="ds-content">
+    <h1>Heading level 1</h1>
```
line47, line 38: Could we add some paragraphs between the headers to give it more of a content block feel?

pwolfert: 👍 I had some in my first iteration but didn't know if it would be distracting

pwolfert:
Not sure what the new build errors are about on Jenkins. I'm getting them locally too. No error from any of the tools—just failing to complete

PR #2358, opened 2023-01-25
[WNMGDS-2019] Remove text color override from mgov, update image diffs

zarahzachz:
## Summary

WNMGDS-2019

Mgov created overrides for typography classes and redeclared the `color` rule. If teams are using our CSS as they should (by applying `.ds-base` or `.ds-base--inverse` to a container) this rule is redundant. So I removed this style.

## How to test

View the Typography styles in Storybook using the mgov theme (build locally)


pwolfert (approved):
A very precise and effective fix 🙂

forrestbaer (approved):

PR #2359, opened 2023-01-25
[WNMGDS-2162] Use :where in ds-content to reduce specificity

zarahzachz:
## Summary

WNMGDS-2162


pwolfert (approved):
Ya'll okay with this temporary measure to fix `@extends` weirdness?

zarahzachz (commented):
i can't approve my own PR, but this looks fine.

PR #2363, opened 2023-01-27
[WNMGDS-2126] Single source of truth for list of themes

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2126

- Create a `themes.json` in the root with useful info about themes
- Use it in Storybook
- Use it in the doc site
- TO DO in a follow-up ticket: Use it in tests. I've got the code ready, but we may want to wait to take the initial visual regression reference images until the new CMS.gov theme has stabilized. Otherwise we'll be running those a lot.


forrestbaer (approved):
Looks great

PR #2364, opened 2023-01-27
[NO-TICKET] Adding more context to our system color tokens.

brentschneider:
## Summary

- Adding more textual context to our system tokens page.
- non ticked change

- [x] Checked for spelling and grammatical errors & reviewed with Scott

### DIFF
--- a/packages/docs/content/foundation/system-color-tokens.mdx
+++ b/packages/docs/content/foundation/system-color-tokens.mdx
@@ -4,6 +4,6 @@ order: 40
 intro: System color tokens are the entire set of color tokens from which our themes are built.
 ---
 
-However, when building your product, it's essential to use the theme color tokens that have been defined for your brand's experience. Think of System color tokens like an entire box of crayons; our themes use specific colors from the box, not all of them.
+While this page shows the full range of colors, it's essential to use the theme color tokens identified for your brand's experience when building your product. Think of system color tokens like an entire box of crayons; our themes use specific colors from the box, not all of them.
 
 <ColorRamps />
--- a/packages/docs/content/foundation/system-color-tokens.mdx
+++ b/packages/docs/content/foundation/system-color-tokens.mdx
@@ -1,7 +1,9 @@
 ---
 title: System color tokens
 order: 40
-intro: System color tokens are the complete set of color tokens from which themes are built.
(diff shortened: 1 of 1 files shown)


packages/docs/content/foundation/system-color-tokens.mdx
```diff
@@ -1,7 +1,9 @@
 ---
 title: System color tokens
 order: 40
-intro: System color tokens are the complete set of color tokens from which themes are built.
+intro: System color tokens are the entire set of color tokens from which our themes are built.
```
pwolfert, line 4: 👍

packages/docs/content/foundation/system-color-tokens.mdx
```diff
@@ -1,7 +1,9 @@
...
 ---
 
+However, when building your product, it's essential to use the theme color tokens that have been defined for your Brand's experience. Think of System color tokens like an entire box of crayons, our themes use specific colors from the box, not all of them.
```
pwolfert, line 7:
```suggestion
However, when building your product, it's essential to use the theme color tokens that have been defined for your brand's experience. Think of System color tokens like an entire box of crayons; our themes use specific colors from the box, not all of them.
```

This second sentence is technically two; otherwise, it's considered a comma splice. I replaced the comma with a semicolon.

The other edit I'm unsure about. Did you mean to capitalize _Brand_?

pwolfert, line 7: Because the text above exists in a `<header>` and the text below exists in an `<article>` that is outside of that `<header>`, I don't know if it's a guarantee that our users will be reading both sections together. I may be wrong. In light of that, though, maybe some additional context at the start of this paragraph might be more appropriate than starting with _However_. Like what if it started with "While this page shows the full range of colors, when building your product...

pwolfert (approved):
This explanation is helpful! I have a couple suggestions for clarity and punctuation.

packages/docs/content/foundation/system-color-tokens.mdx
brentschneider: Ah, gotcha. And now that I'm looking at the code that's shedding more light on our spacing conversation too— thank you.

PR #2365, opened 2023-01-30
[WNMGDS-2206] Theme color page updates

line47:
## Summary

https://jira.cms.gov/browse/WNMGDS-2206

### Visual changes
- Removed the additional colors section that contained 2 colors one of which was transparent 
- Updated styling of colors to be presented more like the system colors page 
- Updated styling of utility > text-color page 
- Updated styling of utility > background-color page

### Code changes
- Reorganized existing color-presentation components into the following components:
    - `ColorSwatch`
        - It's just a svg circle with the color
        - Used in the text-color page and the component pages
    - `ColorExampleRow`
        - It's a row with a background color of the specified color plus two text/code elements with the name of the color and the value of the color
        - Used in both the `ColorExampleList` and `ColorRamps` component
    - `ColorExampleList`
        - It's more generic than `ColorRamps`
        - Used on both the theme-colors page and the background-colors page
    - `ColorRamps`
        - A more specialized form of color example that is specific to the system colors page

## How to test

1. Run `yarn start` and visit the theme colors page and see that the additional colors section is no longer on the page at the bottom and the color samples are now rectangular that stretch the page width.

## Checklist


line47:
@pwolfert curious if we could use the same type of light and dark text color functionality that is on the system color page on this theme page as well?

pwolfert:
@line47, this page looks nice! Were you aware that it also changed the following pages?
- http://localhost:8000/utilities/text/text-color/
- http://localhost:8000/utilities/background-color/

Yes, it would be possible to apply the light/dark text logic to these swatches. We also might want to constrain their width to `measure--base`.

line47:
Pushed some updates and am unsure what a good path forward would be for the text-color page. I was thinking we might want to have a different control to display that as the need is a bit different. thgouths @pwolfert ?

line47:
Came up with a solution that works but feels a bit gross... :)

forrestbaer (approved):
Looks great! One comment on the design, I like the `code.c-swatch__name` with 8px of padding on the left/right  a bit more than 4px.

forrestbaer:
Yikes, I didn't mean to do that.

PR #2371, opened 2023-02-02
[NO-TICKET] Removing faulty husky hook for date stamping

forrestbaer:
## Summary

- Now that token files are generated every `yarn build`, when a date-stamped theme file was re-built it would not have the datestamp, showing it as a change to a file that needs a datestamp which would then date stamp it again. 🐢 
- Right now I can't come up with a great solution, will need to consider this as a future 'nice to have', removing for now.

- Also adds the cmsgov theme into it's new folder!


pwolfert (approved):
🧹

PR #2372, opened 2023-02-02
[WNMGDS-2183] CSS reset

zarahzachz:
## Summary

WNMGDS-2183

I created a normalizing reset for the design system. You'll find annotations in the SCSS around what decisions have been made, and questions about adjustments that might need to be considered.

I haven't added reset styles for images or `dialog` - images is tough to have an opinion on since we don't really have styles or guidance for them. `dialog` is something I'm actively adding but shouldn't block feedback.

## How to test

1. View [demo link](https://cmsgov.github.io/design-system/branch/WNMGDS-2183/css-reset/storybook/?path=/story/reset-overview--overview&globals=theme:core) or build Storybook locally (`yarn build && yarn storybook`) and visit the reset pages


.storybook/static/core-theme.css
```diff
@@ -103,6 +103,8 @@
 --color-warn-lighter: #fce28f;
 --color-warn-lightest: #fef9e9;
 --color-visited: #4c2c92;
+--font-body: 'Open Sans', Helvetica, sans-serif;
+--font-heading: 'Open Sans', Helvetica, sans-serif;
```
zarahzachz, line 107: I created these `--font-*` vars to make it easier to switch between themes on the reset. After I made them, I noticed we also have `--typography-` variables defined and I was wondering if we should standardize behind one? 

Personally, seeing all the "font" vars up top made me think we had outages because I didn't realize there were additional vars with the "typography" prefix towards the bottom.

.storybook/storybookStyles.scss
```diff
@@ -16,3 +16,9 @@
 [data-theme='medicare'] {
   @import '@cmsgov/ds-medicare-gov/src/styles/index';
 }
+
+article {
+  .preview {
+    border: 1px dashed orchid;
```
zarahzachz, line 22: I added this silly style to make the actual reset examples standout - we don't use dashed or purple borders so it made sense to me. This is temporary and will be removed upon approval of the reset.

packages/design-system/src/components/reset/resetDisclosureEls.stories.tsx
```diff
@@ -0,0 +1,42 @@
+import React from 'react';
+
+export default {
+  title: 'Reset / Disclosure Elements',
+};
+
+export const DisclosureElements = () => {
```
zarahzachz, line 7: I plan on adding a `dialog` example here, but with it being a more complex component I didn't want to hold up the reset review

packages/design-system/src/components/reset/resetFormEls.stories.tsx
```diff
@@ -0,0 +1,149 @@
...
+            <fieldset>
+              <legend>Legend: Text input examples</legend>
+              <label htmlFor="text-input">Default text input field</label>
+              <input type="text" name="text-inputs" id="text-input" />
+              <br />
+              <label htmlFor="disabled-text-input">Disabled text input field</label>
+              <input type="text" name="text-inputs" id="disabled-text-input" disabled />
+              {/* <br />
+              <label htmlFor="readonly-text-input">Readonly text input field</label>
+              <input type="text" name="text-inputs" id="readonly-text-input" readOnly /> */}
```
zarahzachz, line 30: `readonly` are common text inputs types, but we don't use them in our DS, so I can't infer style. Is this something we need to support?

packages/design-system/src/components/reset/resetLists.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+          </ol>
+
+          <hr />
+
+          <ol>
+            <li>List Item 1</li>
+            <li>List Item 2</li>
+            <li>
+              List Item 3
+              <ol>
```
zarahzachz, line 33: I need design 👀 on nested lists. We don't have examples and the list marker is the browser default currently. Also need to figure out how spacing works for nested `ul`/`ol` since they have `margin-top` applied

packages/design-system/src/styles/_base.scss
```diff
@@ -120,7 +120,8 @@ $breakpoints: (
   }
 }
 
-a,
+// ❓ if we style the <a> tag, what's the point of this link class?
```
forrestbaer, line 123: Only used in a couple components which overwrite colors (Alert), or extend it (UsaBanner)..

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+// the CSS variable to adjust bold based on theme
+b,
+strong {
+  font-weight: var(--font-weight-bold);
+}
+
+a {
+  color: var(--link__color);
+  // 📢 We may want to revisit tokens for `text-decoration-style`
+  // ❓Do we have an instance where the underline doesn't match the text color?
```
forrestbaer, line 99: @line47 or @brentschneider know? I can't find any examples like that.

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+  margin-block: 1.5em 0;
+  padding: 0;
+}
+
+label {
+  display: block;
+  font-weight: var(--font-weight-bold);
+  margin-block: 1.5em 0;
+
+  // I see max-width defined on label and I wonder if we should consider removing an opinionated width on individual elements?
```
forrestbaer, line 194: Agreed, rely on the grid for layout.

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -83,6 +83,7 @@
   text-align: left;
 }
 
+// ❓ Do we need these text-align classes when we have utils that do the same thing?
```
forrestbaer, line 86: I vote to use the more generic alignment classes and to deprecate these.

forrestbaer (commented):
Nice work! Not sure on several of the questions, but had a few comments.

.storybook/static/core-theme.css
pwolfert: I do think they refer to different contexts, but it's not super clear. The `typography` prefixed ones are used in the typography-related classes like headings and `ds-text-body`, whereas if we're applying this in a reset that includes other styles not related to typography, it wouldn't necessarily fit to have the `typography` prefix. Just trying to make sense of what's there already; I'm not prescribing anything. It also looks like `ds-text` is a duplicate of `ds-text-body`. I think `ds-text` is the older one and out of fashion since the `ds-text-heading*` classes came along.

.storybook/storybookStyles.scss
pwolfert: Regarding togglable styles, could this built-in feature do the same thing, Forrest?

Screen Shot 2023-02-02 at 1 48 04 PM

Or is the utility in the utility that is highlights which part is relevant?

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+  box-sizing: border-box;
+}
+
+html:focus-within {
+  scroll-behavior: smooth;
+}
+
+// Remove all animations, transitions and smooth scroll
+// for people that prefer not to see them
+@media (prefers-reduced-motion: reduce) {
```
pwolfert, line 14: Ooh, nice

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+  font-family: var(--font-body);
+  font-size: var(--font-size-base);
+
+  // 👀 CMS-WDS/scout uses these reset styles
+  -webkit-font-smoothing: antialiased;
+  -moz-osx-font-smoothing: grayscale;
+  line-height: var(--font-line-height-base);
+  margin: 0;
+  min-height: 100vh;
+  text-rendering: optimizeSpeed;
```
pwolfert, line 43: I can see the benefit in having a min-height of 100vh so you don't get weird footer placement

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+
+// 👀 CMS-WDS/sep-screener uses this reset style
+// I've seen similar styles in other repos and I'm curious what it's solving
+#main[tabindex='-1']:focus,
+h1[tabindex='-1']:focus {
+  outline: 0;
+}
+*[tabindex='-1']:focus {
+  outline: 0;
+}
```
pwolfert, line 54: I imagine this is getting rid of the focus outline for non-focusable elements that have explicitly added the `tabindex="-1"` in order to make them focusable. I'd love @SparrowSpit's thoughts on this. Should we by default hide focus rings for such elements?

It also looks like the `#main` and `h1` are redundant with the `*`, and I'm curious if we even need `*` or if the selector could be just `[tabindex='-1']:focus`

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+
+/* Typography */
+// ❓ Do we want to enforce max-width rules on type els?
+// I personally feel like doing this would result in more support work for us
+// but it could result in better enforcement of this rule.
```
pwolfert, line 59: That's a good question. @line47, @brentschneider, and @babsdenney, do you want to weigh in here?

pwolfert: The current strategy in this reset is to automatically apply some standard styles to `h1` through `h6` and `p`, including standardized margin. The question posed here is whether we also want to put a width restriction like `measure--wide` on them as well, or if our users should always opt in to width restrictions.

pwolfert: I'm inclined to vote _no_ since we don't currently have a nice way of opting out through a utility class. Maybe if we had a `ds-u-measure--none` that could remove it 🤔

packages/design-system/src/components/reset/resetFormEls.stories.tsx
pwolfert: That's a good idea. I can't think of one of our products that uses them, but that doesn't mean they aren't or won't be used. Seems like we could do something similar to or the same as a disabled style? @line47, what do you think? [Here's an MDN page on readonly fields](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/readonly).

packages/design-system/src/styles/_reset.scss
pwolfert: I also agree, or at least we make it consistent with the _measure_ utilities. Devs can also apply those _measure_ utilities themselves.

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+}
+
+fieldset {
+  border: 0;
+  margin: 0;
+  padding: 0;
+}
+
+// ❓ Not seeing legends used in DS to draw style conclusions from;
+// assuming they should be styled like label els, but not sure
```
pwolfert, line 181: They aren't used by themselves without a class name on them, but we do use them in the `ChoiceList` in place of a `label`, and we apply `ds-c-label` to it. I'm not sure if we really need to reset it outside of the `FormLabel` usage since the design system doesn't use them in plain form. What does everyone think?

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+}
+
+/* Form elements */
+
+input,
+button,
+textarea,
+select {
+  font: inherit;
+}
```
pwolfert, line 172: I think this font declaration is good to include in the reset code, but I kind of think it's unnecessary to reset all the form elements since we have classes specifically for those things. Or we could possibly restrict the declarations to things like margin. Otherwise, what's the point of those other classes?

pwolfert: The bounds of the design system styles are shifting a lot by this reset, and we may have to move the line several times before we strike the right balance.

pwolfert (commented):
The new stories are a good idea, and this is good work. The comments are really helpful! I hope we can keep some version of these comments in the final version too to give context about decisions.

packages/design-system/src/styles/_reset.scss
SparrowSpit: Yeah, it's common to not display focus rings for non-interactive elements like when activating a skip link will typically send focus to a <main> or a heading element. The caveat to this, of course, would be if the heading element also happens to be a link.

packages/design-system/src/components/reset/resetFormEls.stories.tsx
line47: I think it's worth having a conversation with @mliupaeter about it before implementing. ALso I'm unsure if teams are using readonly fields. :shrug:

packages/design-system/src/components/reset/resetLists.stories.tsx
line47: I'd be happy to work on this with you @zarahzachz

.storybook/static/core-theme.css
line47: **A typeface/typography** is a set of design features for letters and other characters, like the presence or lack of a serif, the letters’ weight and balance, spacing and the height difference between upper and lowercase letters. So think Times New Roman  
  
**A font** is the variation in weight and size of a typeface. So when a typeface is, bold, italic, condensed, size or any other variable, that’s a font.  
  
So there is a difference and it is also definitely nuanced

line47: I'd recommend something like `--typography-body` or `--typeface-body` and using `--font` for the items that deal with sizing, weight, etc...

packages/design-system/src/styles/_reset.scss
line47: I'm also inclined to vote no on this unless we can talk to teams and find out if they want this change or not. Would teams be undoing it often?

packages/design-system/src/styles/_reset.scss
line47: I don't think we do have any examples we could I'll ask and try and get some intel on it

packages/design-system/src/styles/components/_Table.scss
line47: Oh yeah, that's a great idea!

packages/design-system/src/styles/core-theme.css
```diff
@@ -103,6 +103,8 @@
 --color-warn-lighter: #fce28f;
 --color-warn-lightest: #fef9e9;
 --color-visited: #4c2c92;
+--font-body: 'Open Sans', Helvetica, sans-serif;
```
line47, line 106:
```suggestion
--typography-body: 'Open Sans', Helvetica, sans-serif;
```

packages/design-system/src/styles/core-theme.css
```diff
@@ -103,6 +103,8 @@
...
+--font-heading: 'Open Sans', Helvetica, sans-serif;
```
line47, line 107:
```suggestion
--typography-heading: 'Open Sans', Helvetica, sans-serif;
```

packages/docs/static/themes/core-theme.css
```diff
@@ -103,6 +103,8 @@
 --color-warn-lighter: #fce28f;
 --color-warn-lightest: #fef9e9;
 --color-visited: #4c2c92;
+--font-body: 'Open Sans', Helvetica, sans-serif;
```
line47, line 106:
```suggestion
--typography-body: 'Open Sans', Helvetica, sans-serif;
```

packages/docs/static/themes/core-theme.css
```diff
@@ -103,6 +103,8 @@
...
+--font-heading: 'Open Sans', Helvetica, sans-serif;
```
line47, line 107:
```suggestion
--typography-heading: 'Open Sans', Helvetica, sans-serif;
```

packages/docs/static/themes/healthcare-theme.css
```diff
@@ -103,6 +103,8 @@
 --color-warn-lighter: #fce28f;
 --color-warn-lightest: #fef9e9;
 --color-visited: #4c2c92;
+--font-body: 'Open Sans', Helvetica, sans-serif;
```
line47, line 106:
```suggestion
--typography-body: 'Open Sans', Helvetica, sans-serif;
```

packages/docs/static/themes/healthcare-theme.css
```diff
@@ -103,6 +103,8 @@
...
+--font-heading: 'Open Sans', Helvetica, sans-serif;
```
line47, line 107:
```suggestion
--typography-heading: 'Open Sans', Helvetica, sans-serif;
```

packages/docs/static/themes/medicare-theme.css
```diff
@@ -156,6 +156,8 @@
 --font-weight-normal: 400;
 --font-weight-semibold: 600;
 --font-weight-bold: 600;
+--font-body: 'Rubik', sans-serif;
```
line47, line 159:
```suggestion
--typography-body: 'Rubik', sans-serif;
```

packages/docs/static/themes/medicare-theme.css
```diff
@@ -156,6 +156,8 @@
...
+--font-heading: 'Montserrat', sans-serif;
```
line47, line 160:
```suggestion
--typography-heading: 'Montserrat', sans-serif;
```

packages/ds-healthcare-gov/src/styles/healthcare-theme.css
```diff
@@ -103,6 +103,8 @@
 --color-warn-lighter: #fce28f;
 --color-warn-lightest: #fef9e9;
 --color-visited: #4c2c92;
+--font-body: 'Open Sans', Helvetica, sans-serif;
```
line47, line 106:
```suggestion
--typography-body: 'Open Sans', Helvetica, sans-serif;
```

packages/ds-healthcare-gov/src/styles/healthcare-theme.css
```diff
@@ -103,6 +103,8 @@
...
+--font-heading: 'Open Sans', Helvetica, sans-serif;
```
line47, line 107:
```suggestion
--typography-heading: 'Open Sans', Helvetica, sans-serif;
```

packages/ds-medicare-gov/src/styles/medicare-theme.css
```diff
@@ -156,6 +156,8 @@
 --font-weight-normal: 400;
 --font-weight-semibold: 600;
 --font-weight-bold: 600;
+--font-body: 'Rubik', sans-serif;
```
line47, line 159:
```suggestion
--typography-body: 'Rubik', sans-serif;
```

packages/ds-medicare-gov/src/styles/medicare-theme.css
```diff
@@ -156,6 +156,8 @@
...
+--font-heading: 'Montserrat', sans-serif;
```
line47, line 160:
```suggestion
--typography-heading: 'Montserrat', sans-serif;
```

.storybook/static/core-theme.css
```diff
@@ -103,6 +103,8 @@
 --color-warn-lighter: #fce28f;
 --color-warn-lightest: #fef9e9;
 --color-visited: #4c2c92;
+--font-body: 'Open Sans', Helvetica, sans-serif;
```
line47, line 106:
```suggestion
--typography-body: 'Open Sans', Helvetica, sans-serif;
```

.storybook/static/core-theme.css
line47, line 107:
```suggestion
--typography-heading: 'Open Sans', Helvetica, sans-serif;
```

.storybook/static/healthcare-theme.css
```diff
@@ -103,6 +103,8 @@
 --color-warn-lighter: #fce28f;
 --color-warn-lightest: #fef9e9;
 --color-visited: #4c2c92;
+--font-body: 'Open Sans', Helvetica, sans-serif;
```
line47, line 106:
```suggestion
--typography-body: 'Open Sans', Helvetica, sans-serif;
```

.storybook/static/healthcare-theme.css
```diff
@@ -103,6 +103,8 @@
...
+--font-heading: 'Open Sans', Helvetica, sans-serif;
```
line47, line 107:
```suggestion
--typography-heading: 'Open Sans', Helvetica, sans-serif;
```

.storybook/static/medicare-theme.css
```diff
@@ -156,6 +156,8 @@
 --font-weight-normal: 400;
 --font-weight-semibold: 600;
 --font-weight-bold: 600;
+--font-body: 'Rubik', sans-serif;
```
line47, line 159:
```suggestion
--typography-body: 'Rubik', sans-serif;
```

.storybook/static/medicare-theme.css
```diff
@@ -156,6 +156,8 @@
...
+--font-heading: 'Montserrat', sans-serif;
```
line47, line 160:
```suggestion
--typography-heading: 'Montserrat', sans-serif;
```

.storybook/storybookStyles.scss
zarahzachz: I didn't realize Storybook had this setting. It's a lot more granular, but appears to accomplish the same thing.

packages/design-system/src/components/reset/resetFormEls.stories.tsx
zarahzachz: I can omit the `readonly` example for now. We can revisit once there's a known need later on.

packages/design-system/src/styles/_reset.scss
zarahzachz: These elements have crazy user agent styles without being reset, and within the DS use same font styles as the page itself (family, size, line-height, etc.), so generically setting the font styles for these elements like this enables us to remove unnecessary CSS from the classes redeclaring the same things over and over. 

For instance, [these lines of CSS](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/styles/components/_Button.scss#L13-L19) could be removed from our Button styles and [these](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/styles/components/_TextField.scss#L51-L53) from our TextFields.

pwolfert: I'm on board with this line that I highlighted and other font stuff and margins, but I just wasn't sure if we needed stuff like `border-radius` and `background-color` because they are maybe redundant with the classes that we use 🤷‍♂️

zarahzachz: Oh, yeah maybe for the inputs themselves I can remove a lot of the non-font, layout styles. I'm good with that.

packages/design-system/src/components/reset/resetDisclosureEls.stories.tsx
zarahzachz: @pwolfert I'm actually reconsidering adding reset styles for `details` and `dialog` - mainly for the same thing you pointed out with the form elements. We have components/classes that accomplish all the styles needed for these tags and it doesn't really feel like a reset (`background-color`, `padding`, `border`, etc.). Do you think there's a need for resets on these two tags?

.storybook/static/cmsgov-theme.css
```diff
@@ -511,7 +511,7 @@
 --typography-heading-lg__font-weight: 700;
 --typography-heading-md__font-weight: 700;
 --typography-heading-sm__font-weight: 400;
---typography-body__font-family: inherit;
+--typography-body__font-family: 'Open Sans', Helvetica, sans-serif;
```
zarahzachz, line 514: I updated these pre-existing `--typography-` vars to explicitly point to the theme's body font. As it was before, this var was relying on the `--font-sans` being set on `.ds-base`. Now the proper body fonts are being assigned to the `body` tag

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,248 @@
...
+  margin: 0;
+  min-height: 100vh;
+  text-rendering: optimizeSpeed;
+}
+
+// This removes focus rings for non-interactive elements,
+// i.e. when activating a skip link will then send focus
+// to a main or a heading element.
+[tabindex='-1']:focus {
+  outline: 0;
```
zarahzachz, line 46: There was discussion between @pwolfert and @SparrowSpit about whether this rule would be adequate for what teams used it for. Dan pointed out a caveat might be if a heading were a link, but I still think this would work - mainly because the link nested inside the heading would still receive an outline as it is interactive:

```html
<h2 tabindex="-1">
    <a href="">Click this heading to go nowhere</a>
</h2>
```

A quick codepen test showed the link still receives a focus ring despite the heading with a -1 tabindex having this style.

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,248 @@
...
+  &[role='list'] {
+    list-style: none;
+    padding: 0;
+  }
+}
+
+li {
+  > ul,
+  > ol {
+    margin-block: 0.5em 0;
```
zarahzachz, line 145: I previously commented out the flex solution in favor of adding margin to the `li`, however I've reconsidered that position after figuring out how to create consistent spacing for nested lists.

`flex` is a superior solution as it prevents weird spacing problems when the direction of the list changes (see our Footers to see this issue in action).

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,247 @@
...
+
+/* Images */
+
+figure {
+  margin: 0;
+}
+
+/* Disclosure */
+
+details {
```
zarahzachz, line 205: As mentioned above, I don't know if it's worth including styles for `details` or `dialog` as a reset as the styles veer more into "strong DS opinion territory". I could also be persuaded to include the table reset styles here. Should I just delete everything from 203+?

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+    padding: 0;
+  }
+}
+
+// 📢 We don't have examples of nested lists in the design system
+// ❓ How do we handle li `::mark` and nested padding for this pattern?
+// If we choose to support nesting lists, need to remove `gap` solution in favor of li `margin-block-start`
+// li elements are `display: list-item` and cannot receive `flex` or `grid` display without removing `::mark`
+li {
+  margin-block: 0.5em 0;
```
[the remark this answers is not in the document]
zarahzachz: I eventually went back to the `flex` solution. The original problem I had with it was when nesting lists, there was inconsistent spacing between the `li` and `ul`/`ol` nested within it. That's because `li` has a `display: list-item` rule that can't really be changed to flex without losing the bullets or numbers (important for a list!).

I found a way to target `ul`/`ol` children of `li` els so they could be given consistent padding, meaning the flex solution now works perfectly.

I like the `flex` solution because it results in cleaner UI when teams render lists as rows or inline. Currently, our `li` els have this margin-top baked in, so when users create inline lists, there's this weird top spacing they have to contend with. Now that we have this gap solution, the spacing goes away. Teams can also create inline lists by styling them to `flex-direction: row` or adjust spacing between `li` with `gap: <WHATEVER UNIT>`.

packages/design-system/src/components/reset/resetDisclosureEls.stories.tsx
pwolfert: No, I'd agree that it's in the same boat and that our other styles cover it. I doubt teams are using bare dialogs that they're resetting themselves. If the goal is for them to be able to trade their resets for ours so theirs doesn't conflict with ours, I think this is an area where they probably don't need a reset of their own.

packages/design-system/src/styles/_reset.scss
```diff
@@ -0,0 +1,344 @@
...
+
+// Relying on `flex` to set `gap` spacing size on li elements
+// `gap` ensures even spacing and doesn't interfere with layout if list given a `row` direction
+ol,
+ul {
+  // display: flex;
+  // flex-direction: column;
+  // gap: .5em;
+  margin-block: 1em 0;
+  padding-inline-start: 2em;
```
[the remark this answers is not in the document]
zarahzachz: Unfortunately, styling for `li::marker` is extremely limited. I think you can just change its color? [People have complained](https://twitter.com/simevidas/status/1556386048279805954) about this limitation before but I don't think I've seen any browsers/W3 acknowledge they need to change anything here :(

packages/design-system/src/components/reset/resetDisclosureEls.stories.tsx
zarahzachz: Removed disclosure styles and Story as it overstepped bounds of reset

packages/design-system/src/components/reset/resetLists.stories.tsx
zarahzachz: I worked with Scott last week and we came up with an ideal pattern for ordered lists. Updated the Story to reflect this decision.

packages/design-system/src/styles/_reset.scss
pwolfert: Agree 👍

pwolfert (approved):
This looks great! I know it's not final, but great work on this, @zarahzachz!

forrestbaer (approved):
This is a big step in the right direction, thank you!

forrestbaer (approved):

pwolfert:
Looks like both sets of recent VRT updates were changes to the spacing on the clear button of the autocomplete. Was that intentional?

PR #2375, opened 2023-02-07
[NO-TICKET] Add link variables to link page

line47:
## Summary

- Add link variables to link page

## How to test

1. Run `yarn start` and visit the link page and see that the variables are listed under the styles section

## Checklist


line47:
One odd thing that I noticed was that the names/values didn't match the values that I saw in the theme code?

pwolfert:
Ah, the color names thing is a limitation stemming from how we store the tokens. Only the final hex value is what's stored in the variables (not the named colors), which means we have to look through the named variables and find the first match for that given hex value. In this case `coolblue-light` is a duplicate of `secondary` but comes first in the list. @line47

line47:
Given our discussion earlier today about color and `coolblue` colors. I'm hesitant to push this update until that is resolved unless there is some sort of workaround. :thinking:

pwolfert:
The only workaround I can think of is to remove `cooblue` 😆

line47:
That's what I thought, maybe we could stange this work somewhere then as part of the breaking change to remove `coolblue` ?

forrestbaer (approved):

PR #2377, opened 2023-02-08
[NO-TICKET] Remove unused SVG files in Medicare

pwolfert:
## Summary

Removed SVG files from medicare components dir that aren't used. They're not imported into components or anything. It's possible that teams are importing directly from `src/components/*`, but that should be discouraged anyway. These files aren't copied into `dist`.


zarahzachz (approved):
Tests pass locally. Approving!

PR #2380, opened 2023-02-10
[NO-TICKET] Updating snapshot locations

forrestbaer:
## Summary

The 'new/updated' snapshots in this PR are actually correct, it looks like this needed updating anyway. I checked through many of them against current main.

- Updated location of snapshots to `tests/browser/snapshots` for the browser suite of tests.
- Updated naming convention to match interaction snapshots
- Fixed a bug I introduced around logic for when to take certain shots.

This should resolve the nightly failures.


tests/browser/story-screenshots.test.ts
```diff
@@ -27,8 +27,15 @@ Object.values(stories).forEach((story) => {
...
+      if (theme !== 'core' && story.importPath.includes(themes[theme].packageName)) return;
+
       // Don't take screenshots of theme-specific components outside of their themes
-      if (isSmokeTest && !story.importPath.includes(themes[theme].packageName)) return;
+      // during smoke tests.
+      if (
+        isSmokeTest &&
+        !(theme === 'core' || !story.importPath.includes(themes[theme].packageName))
+      )
+        return;
```
pwolfert, line 38: [Prior to the introduction of interaction tests](https://github.com/CMSgov/design-system/blob/753fddf3ef69c55b66a605a3b58faecb316db528/tests/browser/story-screenshots.test.ts), this if-statement had the following comment, which more accurately describes what the logic is there for:

> // For smoke tests, only capture core components in the core theme and theme-specific
> // components in their native theme

The comment that is currently above it goes above the line you're adding now, which is

> Don't take screenshots of theme-specific components outside of their themes

pwolfert: And thanks for simplifying the different checks down to `story.importPath.includes(themes[theme].packageName))`!

pwolfert (approved):
Ran each suite of browser tests with and without `--env SMOKE=true` in the docker command along with `--list` in the playwright command, and I now see a difference in the number of tests run, as expected 👍

tests/browser/story-screenshots.test.ts
forrestbaer: Good call 👍🏻 Updated

tests/browser/story-screenshots.test.ts
```diff
@@ -27,8 +27,15 @@ Object.values(stories).forEach((story) => {
     Object.keys(themes).forEach((theme) => {
       if (themes[theme].incomplete) return;
 
-      // Don't take screenshots of theme-specific components outside of their themes
```
pwolfert, line 30: This comment here should go back in. One comment for each scenario

zarahzachz (approved):

PR #2381, opened 2023-02-10
[WNMGDS-2186] Remove non-semantic `coolblue`, `red`, and `gold` theme colors

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2186

- Started out removing `coolblue`, `gold`, `green`, and `red` colors, which don't have semantic meaning. Unfortunately `green` is used for either `main` or `alternate` buttons for all but the medicare theme. We'll need to work more carefully to get rid of `green`.
- Removed the _fill_ utility classes for the non-semantic colors (including `green`)

## How to test

VRTs should catch problems.


line47 (approved):
LGTM

PR #2382, opened 2023-02-13
[WNMGDS-2186] Normalize theme-level gray palettes

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2186

This is a follow-up to https://github.com/CMSgov/design-system/pull/2381 that tackles the normalization of the grays. It changes the set of variables represented by the grays to be more consistent with our other palettes like `secondary` or `success`. 
- The following theme variable names are now standard across the four themes to be consistent with our other color variables:
    - `gray-lightest`
    - `gray-lighter`
    - `gray-light`
    - `gray`
    - `gray-dark`
    - `gray-darker`
    - `gray-darkest`
- The following variables were removed:
    - `gray-medium`
    - `gray-warm-dark`
    - `gray-warm-light`
    - `gray-medium-light`
    - `gray-darker`
    - `gray-cool-light`
- I had to do some shifting of component variables up or down the gray scale in order to keep the final values the same as they were before. Because the medicare theme used unique values for some of them and because we wanted the darkness/lightness stepping consistent (just made it exactly the same as the other themes), I had to match some of the component variables to the nearest gray value that was represented in the new theme variables.
- Because we removed `gray-cool-light`, we had to remap some day picker assignments. Created the new `--day-picker-button__background-color--hover` component variable for that and assigned it to a new `primary-lightest` theme variable. We'll be getting `primary-lightest` in the next pull request anyway, but I wanted to be able to separate these pull requests as much as possible.
    - This will result in a color shift for the hover background of those day picker buttons, because previously the `gray-cool-light` was mapped to `ocean-50` across the board instead of separate values for each theme. Now that it is in the `primary` family of colors, healthcare will use `sapphire-50` instead of `ocean-50`, medicare will use `teal-50`, and cmsgov will use `
      Screen Shot 2023-02-13 at 4 44 22 PM
    - The shift isn't covered by the visual regression tests at the moment
- We may change the name of the grays to `neutral-*` in a future PR, but that hasn't been totally settled yet

## How to test
- To test the hover color of day picker buttons, use the _with calendar picker_ story for SingleInputDateField


pwolfert:
I know the VRTs will fail, but I'd like to see which stories will be flagged by the CI version, because it seems to eager to update snapshots when I run it locally.

line47:
@pwolfert This looks really good! 

We still need to update the doc site theme colors page
- break out `color-base` into its own section called "Text colors" on the doc site theme colors page. 
- Update the gray swatch set under neutrals heading - it's currently no showing gray-darker or gray-darkest

packages/design-system-tokens/src/themes/medicare-component.ts
```diff
@@ -288,7 +292,10 @@ export const components: AnyTokenValues = {
     '-toggle__background-color--hover':           t.color['secondary'],
     '-toggle__color--hover':                      t.color['white'],
     '-toggle__background-color--hover--inverse':  t.color['white'],
-    '-toggle__color--hover--inverse':             t.color['gray-darker'],
+    // I wonder if we can get rid of this variable and just use link-inverse__color--hover
+    // This variable is only used in a medicare style override, and it's weird. Why doesn't
+    // it just behave like a normal inverse link?
+    '-toggle__color--hover--inverse':             t.color['gray-dark'],
```
pwolfert, line 298: See the comment as well, but I think we should remove this and the CSS override that it represents. This is what the override behavior looks like, which is weird:

medicare-drawer-toggle

In my opinion, the colors should just behave like that of a link.

pwolfert: @line47, what do you think?

line47 (approved):
:chef-kiss: nice!

packages/design-system-tokens/src/themes/medicare-component.ts
pwolfert: Update: We talked about it, and we're going to address it in a separate ticket

PR #2385, opened 2023-02-14
[WNMGDS-2192] Adding 'extension' property to DS analytics object

forrestbaer:
[WNMGDS-2192](https://jira.cms.gov/browse/WNMGDS-2192)

## Summary

- Added `extension` property to our analytics object
- Updated tests & existing analytics to send default DS value

## How to test

1. Evaluate test snapshots, code

## Checklist


packages/design-system/src/components/Dialog/useDialogAnalytics.ts
```diff
@@ -31,6 +31,7 @@ export function useDialogAnalytics({
       event_type: EventType.UI_INTERACTION,
       event_category: EventCategory.UI_COMPONENTS,
       event_label: eventHeadingText,
+      event_extension: 'Design system integration',
```
pwolfert, line 34: This string is repeated in several places. How would you feel about making this a constant?

pwolfert (approved):
Ran through all of them in Storybook, and they work 👍

packages/design-system/src/components/Dialog/useDialogAnalytics.ts
forrestbaer: Yeah not a bad idea since it's something we want to remain the same across all events by default.  I've just pushed those changes.

forrestbaer: whoops not ready yet

forrestbaer: Ok good to go 👍🏻

PR #2386, opened 2023-02-16
[WNMGDS-2191] Update the heading embedded examples and stories to use h4 and below

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2191

Update the heading embedded examples and stories to use h4 and below so as to not mess with the heading levels on the doc page itself.

## How to test

1. Turn VoiceOver on
2. Open http://localhost:8000/foundation/typography/headings/ in Safari
3. Press <kbd>Ctrl + Opt + U</kbd> to open the rotor


pwolfert:
@SparrowSpit, I haven't looked at the doc-site headings in VoiceOver until now, and I'm realizing that the headings are littered with noisy permalink links

Screen shot of the VoiceOver rotor on a doc site page with a list of all the headings, where the headings have text like 'heading 5xl permalink' in front of the actual heading text

This isn't ideal, is it? It's coming from a plugin that we use to generate those permalink buttons.

pwolfert:
The visual regression tests caught a problem with changing the story/example for the `ds-content` class. This class automatically applies default styles to `h1`...`h6` appropriate for each level. The example therefore includes those heading levels and expects the styling to match. I'm not sure what to do about this kind of example.

SparrowSpit:
I'm trying to understand exactly what we're trying to solve here and am a little confused so forgive me. I get the original issue of the ticket but the secondary questions are confusing. 

Which of these are we trying to achieve?
1. get rid of the word "permalink"
2. remove the example ("heading example 4xl permalink ...") from the a11y tree
3. something else
4. all of these

pwolfert:
Closing in favor of https://github.com/CMSgov/design-system/pull/2395

PR #2388, opened 2023-02-22
[NO-TICKET] Add local `test:browser:smoke` command

pwolfert:
## Summary

- Add `test:browser:smoke` command to quickly run a subset of tests only in Chrome
- Switch `test:browser:all` and `test:browser` (we thought this would require a Jenkins script update, but we actually don't 'use these commands directly on Jenkins)

## How to test


forrestbaer (approved):
Looks good!

pwolfert:
Huh, kind of seems like a coincidence that it failed on this branch. Looks like we might need to upgrade playwright in our project because I never locked in the docker image version in Jenkins. (Or I could lock the version in Jenkins)

PR #2390, opened 2023-02-24
[WNMGDS-2130] Implemented automation of release notes for DS

forrestbaer:
## Summary

- [x] Need to add Documentation

[WNMGDS-2130](https://jira.cms.gov/browse/WNMGDS-2130)

- Added a new npm script `release:notes`
- Updated `themes.json` with additional data
- Updated PR summary defaults to add items for labels

**Q:** do we want to keep the `6.0.0-release-notes.md` file? Might be nice to have in the future I dunno..

## How to test

2. `yarn release:notes`
3. Mess around in our [milestone](https://github.com/CMSgov/design-system/milestone/2?closed=1)
4. Add/remove items.. Try removing a label and running.. there has to be at least one `Impacts` and only one `Type`

## Checklist


scripts/notes.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+
+  let successUrl;
+  const draftPre = versions.core.includes('beta') ? '--draft --prerelease' : '--draft';
+
+  try {
+    successUrl = execSync(
+      `gh release create ${getLatestCoreTag()} ${draftPre} --title ${
+        versions.core
+      }-test --notes-file ./${fn}`,
```
forrestbaer, line 210: Will remove this `-test`, just here for your testing use.

scripts/notes.ts
```diff
@@ -0,0 +1,239 @@
...
+
+const rl = readline.createInterface({
+  input: process.stdin,
+  output: process.stdout,
+});
+
+/**
+ * Current milestone reference
+ */
+const cm = JSON.parse(execSync('gh api repos/CMSgov/design-system/milestones').toString())[0];
```
pwolfert, line 40: This would be a little clearer named something like `milestone`. You could still be pretty succinct accessing the data from it if you destructured it like

```suggestion
const milestone = JSON.parse(execSync('gh api repos/CMSgov/design-system/milestones').toString())[0];
const { title: milestoneTitle, open_issues, closed_issues } = milestone;
```

scripts/notes.ts
```diff
@@ -0,0 +1,239 @@
...
+
+/**
+ * Organizes Note(s) by the category/section they should belong in.
+ * In order to keep this code as simple as possible, this just breaks up
+ * pr's by their impacts. Right now if an item impacts multiple items, it
+ * will be presented in each of those sections. An item will also be placed
+ * in multiple type categories if it belongs to multiple.
+ */
+const organizeNotes = (data: PRDetails[]) => {
+  const notes: any = [];
```
pwolfert, line 93: It looks like the notes type is always the same and could be defined. Might help in the next function that consumes it. Looks like `[i, prType[0], pr.title, pr.ghpr]` is just a tuple of `[number, string, string, string]`

```ts
type PRNote = [number, string, string, string];
...
  const notes: PRNote = [];
```

scripts/notes.ts
```diff
@@ -0,0 +1,239 @@
...
+  const notes = data
+    .filter((pr) => {
+      return !pr.ticket?.toLowerCase().includes('ticket');
+    })
+    .map((pr) => {
+      return { ticket: `https://jira.cms.gov/browse/${pr.ticket}`, title: pr.title };
+    });
+  const unticketed = data.filter((pr) => {
+    return pr.ticket?.toLowerCase().includes('ticket');
+  });
```
pwolfert, line 177: Could get rid of some return statements

```suggestion
  const notes = data
    .filter((pr) => !pr.ticket?.toLowerCase().includes('ticket'))
    .map((pr) => ({
      ticket: `https://jira.cms.gov/browse/${pr.ticket}`,
      title: pr.title,
    }));
  const unticketed = data.filter((pr) => pr.ticket?.toLowerCase().includes('ticket'));
```

scripts/notes.ts
```diff
@@ -0,0 +1,239 @@
...
+ */
+const publishNotes = (notes: string) => {
+  const fn = `${versions.core}-release-notes.md`;
+
+  try {
+    writeFileSync(fn, notes, { encoding: 'utf8' });
+  } catch (err) {
+    console.error(err);
+    process.exit(1);
+  }
```
pwolfert, line 194: What's the difference between try-catching and exiting vs just letting it throw its error and exit node itself?

pwolfert (commented):
I love the console output of this command! It's really nice.

When I tried creating a draft, it seemed to have created a 6.0.1 tag and then complained about it. Not sure if that's supposed to work without having actually tagged a release or not:
```
Do these notes look OK to publish as a Draft? (Y/n): Y
tag @cmsgov/cms-design-system-docs@6.0.1 exists locally but has not been pushed to CMSgov/design-system, please push it before continuing or specify the `--target` flag to create a new tag
Error: Command failed: gh release create @cmsgov/cms-design-system-docs@6.0.1 --draft --title 6.0.1-test --notes-file ./6.0.1-release-notes.md
tag @cmsgov/cms-design-system-docs@6.0.1 exists locally but has not been pushed to CMSgov/design-system, please push it before continuing or specify the `--target` flag to create a new tag
```

scripts/notes.ts
```diff
@@ -37,7 +37,8 @@ const rl = readline.createInterface({
 /**
  * Current milestone reference
  */
-const cm = JSON.parse(execSync('gh api repos/CMSgov/design-system/milestones').toString())[0];
+const milestone = JSON.parse(execSync('gh api repos/CMSgov/design-system/milestones').toString())[0];
+const { title: milestoneTitle, open_issues, closed_issues } = milestone;
```
pwolfert, line 41: This changed but references to it didn't, so the script throws errors now

forrestbaer: Yes still working on it! didn't push my changes yet.

scripts/notes.ts
forrestbaer: Making this variable a const requires you initialize it with data of its type, as `any[]` that type can be determined by the loop.  I'm not sure if there is a better way to do this.

scripts/notes.ts
forrestbaer: Oh nothing, this is just a little messy, writeFileSync errors should be fine..

.storybook/static/healthcare-theme.css
```diff
@@ -62,7 +62,7 @@
 --color-secondary-dark: #107b0d;
 --color-secondary-darker: #0d600a;
 --color-secondary-darkest: #094507;
---color-secondary-light: #0d600a;
+--color-secondary-light: #59ac56;
```
forrestbaer, line 65: @pwolfert I think these came from your last PR but I'm not sure why they're showing up here. Maybe this change was just missed in a push?

forrestbaer:
Give it another try when you have a chance @pwolfert, I think using `git ls-remote --tags origin` seems to get better results, sorting by `taggerdate` using `git tag` was giving the wrong result.

.storybook/static/healthcare-theme.css
forrestbaer: Nevermind I see you put in a PR for this :)

scripts/notes.ts
pwolfert: So `const notes: PRNote = [];` results in an error? If that were to be combined with the change below, would it result in an error? Just troubleshooting

```ts
notes.push([i, prType[0], pr.title, pr.ghpr] as any);
```

pwolfert (approved):
This works great! Really nice work

scripts/notes.ts
forrestbaer: Yeah `const declarations must be initialized`. Can't get around it.

PR #2391, opened 2023-02-27
[WNMGDS-2170] remove defensive css

zarahzachz:
## Summary

WNMGDS-2170

As part of the CSS reset work, I pulled out a lot of redundant CSS (and in some cases added new CSS) to ensure the components worked well with the new reset rules. I also removed instances I could find where `ds-base` was being referenced.

Bonus update: I went ahead and implemented [logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties) since I was touching all the CSS anyways. These work exactly as the styles before, only now our CSS supports non-left-to-right readers. It also reduces the lines of CSS we're writing.

You can see the changes in this [demo site](https://cmsgov.github.io/design-system/branch/WNMGDS-2170/remove-defensive-css).


.storybook/static/healthcare-theme.css
```diff
@@ -62,7 +62,7 @@
 --color-secondary-dark: #107b0d;
 --color-secondary-darker: #0d600a;
 --color-secondary-darkest: #094507;
---color-secondary-light: #0d600a;
+--color-secondary-light: #59ac56;
```
zarahzachz, line 65: This was an artifact of merging epic/css-reset into this branch

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -117,12 +117,7 @@ export const Dialog = (props: DialogProps) => {
     ...modalProps
   } = props;
 
-  const dialogClassNames = classNames(
-    'ds-c-dialog',
-    'ds-base',
-    className,
-    size && `ds-c-dialog--${size}`
-  );
+  const dialogClassNames = classNames('ds-c-dialog', className, size && `ds-c-dialog--${size}`);
```
zarahzachz, line 120: `ds-base` no longer needed since its styles are being applied via reset.

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -220,7 +220,7 @@ export const Tooltip = (props: TooltipProps) => {
     } = props;
 
     const TriggerComponent = component;
-    const triggerClasses = classNames('ds-base', 'ds-c-tooltip__trigger', className, {
+    const triggerClasses = classNames('ds-c-tooltip__trigger', className, {
```
zarahzachz, line 223: `ds-base` no longer needed since its styles are being applied via reset.

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -300,7 +300,7 @@ export const Tooltip = (props: TooltipProps) => {
         {...eventHandlers}
       >
         <span className="ds-c-tooltip__arrow" data-popper-arrow />
-        <div className="ds-c-tooltip__content ds-base">
+        <div className="ds-c-tooltip__content">
```
zarahzachz, line 303: `ds-base` no longer needed since its styles are being applied via reset.

packages/design-system/src/components/VerticalNav/VerticalNav.tsx
```diff
@@ -59,7 +59,7 @@ export const VerticalNav = (props: VerticalNavProps): React.ReactElement => {
   const navProps = props.ariaNavLabel ? { 'aria-label': props.ariaNavLabel } : {};
 
   const navList = (
-    <ul className={classes} id={props.id}>
+    <ul role="list" className={classes} id={props.id}>
```
zarahzachz, line 62: Spotted this a11y issue while cleaning up. Adding this attr helped remove some unnecessary CSS along the way.

packages/design-system/src/components/patterns/documentationPage.stories.tsx
```diff
@@ -42,7 +42,7 @@ const navItems = [
 
 export const DocumentationPage = () => {
   return (
-    <main className="ds-base ds-l-container">
+    <main className="ds-l-container">
```
zarahzachz, line 45: `ds-base` no longer needed since its styles are being applied via reset.

packages/design-system/src/components/reset/resetAll.stories.tsx
```diff
@@ -19,7 +18,6 @@ export const Overview = () => {
       <Typography />
       <Lists />
       <FormElements />
-      <Tables />
```
zarahzachz, line 22: In removing redundant CSS, I found our table reset styles overreached into Table (they were identical). I removed the table reset story and styles as a result.

packages/design-system/src/styles/_base.scss
```diff
@@ -29,19 +29,7 @@ $breakpoints: (
  * Body
  */
 
-%ds-base {
-  font-family: var(--font-sans);
-  font-size: var(--font-size-base);
-  line-height: var(--font-line-height-base);
-}
-
-.ds-base {
```
zarahzachz, line 38: Styles within this class now exist on `<body>` in the reset CSS file.

packages/design-system/src/styles/_base.scss
```diff
@@ -155,15 +143,14 @@ $breakpoints: (
 }
 
 @mixin heading-margins {
-  margin-bottom: 0.5em;
-  margin-top: 1.5em;
+  margin-block: 1.5em 0.5em;
```
zarahzachz, line 146: Since I was cleaning up CSS, I updated many of our layout rules to use [CSS logical properties](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Logical_Properties). In a nutshell, it offers the same functionality as before but now with added support for non-left-to-right language preferences.

packages/design-system/src/styles/components/_Drawer.scss
```diff
@@ -148,5 +145,6 @@
 }
 
 .ds-c-drawer__footer-title {
+  font-weight: var(--font-weight-bold);
```
zarahzachz, line 148: Needed to specify as the original footer title was relying on the default user agent browser styles before.

packages/design-system/src/styles/components/_TooltipIcon.scss
```diff
@@ -42,8 +41,8 @@
 .ds-c-tooltip__trigger:hover,
 .ds-c-tooltip__trigger:focus {
   .ds-c-tooltip-icon__container {
-    border-color: var(--tooltip-icon__color);
```
zarahzachz, line 45: This CSS is sus. The way it's written, the first declaration of `border-color` does nothing so I made an assumption here based on the variable name (`__color`) and changed the CSS rule to follow. I don't know if I'm right, but didn't see any screaming from our VRT.

.storybook/static/healthcare-theme.css
pwolfert: You could merge main directly into the epic branch to get rid of this

pwolfert:
I went through the visual regression test reference image changes, and here are my notes:

1. Link color in the alerts got almost imperceptibly darker
2. Clear search button in Autocomplete shifted down
    - I think this fixes [a regression noted here](https://github.com/CMSgov/design-system/pull/2372#issuecomment-1437725574)
    Screenshot 2023-02-28 at 12 52 34 PM
4. Unchecked choices in Medicare have a darker border
    - Seems like a positive change
    Screenshot 2023-02-28 at 12 53 55 PM
5. Drawer footer title wasn't bold and now is 
    - Expected because of [this change](https://github.com/CMSgov/design-system/pull/2391#discussion_r1120655525)
6. Headers in the icon table are no longer left-aligned
7. Text color (heading and body) is darker in the IdleTimeout dialog now 
    Screenshot 2023-02-28 at 12 57 46 PM
8. Medicare multi-input date field has darker input borders
9. Imperceptible change to healthcare single-input date field, but in medicare the border between the day picker and the text field got lighter, so both were probably affected in the same way
    Screenshot 2023-02-28 at 1 02 41 PM
10. Medicare icons table has same alignment issue as the other icon table above
11. StepList screenshots were messed up, and now they're not
12. Reset overview story content changed
    - Expected

packages/design-system/src/styles/_base.scss
pwolfert: Do you think we'll need any documentation for this shift for our users?

packages/design-system/src/styles/components/_Table.scss
```diff
@@ -4,7 +4,7 @@
   border-collapse: collapse;
   border-spacing: 0;
   margin: 0;
-  text-align: left;
+  text-align: start;
```
pwolfert, line 7: The missing left alignment in the VRT diffs appears to be coming from `th` user agent styles taking precedence over this assignment. My browser has the following user agent styles:

```css
th {
    display: table-cell;
    vertical-align: inherit;
    font-weight: bold;
    text-align: -internal-center;
}
```

packages/design-system/src/styles/components/_Dialog.scss
```diff
@@ -25,19 +25,13 @@ dialog:not([open]) {
 
 dialog + .backdrop {
```
pwolfert, line 26: It looks like the darker text color [pointed out in my VRT diff notes](https://github.com/CMSgov/design-system/pull/2391#issuecomment-1448977629) is coming from the `.dialog` polyfill rule. Because the polyfill is for no-longer supported browsers and for server-side rendering frameworks, maybe we should consider getting rid of the polyfill styles entirely.

packages/design-system/src/styles/_base.scss
zarahzachz: Our users shouldn't see a difference, so I don't think anything outside of our normal release notes would be warranted. Now, if we wanted to go a step further and update the name of our utility classes to match the logical property syntax, that would need documentation but I don't see a need to update util class names just yet.

packages/design-system/src/styles/components/_Dialog.scss
zarahzachz: I'm down for that. Should removing it be part of this epic or a larger "clean up IE-only CSS" ticket?

zarahzachz:
> 1. Link color in the alerts got almost imperceptibly darker

Wow! Your eyesight's better than mine. Upon closer inspection, it looks like the reset might've fixed a visual bug. 

The way it's currently coded, we added a special `color` overrides for Alerts but didn't include the other overrides for the underline (`text-decoration-color`). As a result, the underline for links in Alerts were hard-coded to the original darker color, which is why the VRT flagged it.

https://user-images.githubusercontent.com/5159392/222174580-a7391bf9-b8d3-42af-8d76-b0b15fdc8894.mov

In the new reset, I updated the `text-decoration-color` to receive `currentColor` (the defined `color` of the link itself). Talking with Forrest and Scott, they said we don't really have cases where the underline could be a different color from the link text so I'm assuming this Alert style was just an accident before.

Screenshot 2023-03-01 at 9 47 55 AM

zarahzachz:
> 2\. Clear search button in Autocomplete shifted down

Yes, the original reset removed bottom margins from a lot of elements (including inputs) and it threw off some layouts. This shift was to correct that issue.

zarahzachz:
> Unchecked choices in Medicare have a darker border
> * Seems like a positive change

I believe this is a correction from the last reset merge. I compared the border colors between this and prod and they both use #262626.

packages/design-system/src/styles/components/_Dialog.scss
zarahzachz: Looking deeper, the dialog polyfill is also applying its own reset to `<dialog>`, which is overriding the styles set on `<body>`. I can remove the polyfill or update it to exclude our set styles.

zarahzachz:
> 8\. Imperceptible change to healthcare single-input date field, but in medicare the border between the day picker and the text field got lighter, so both were probably affected in the same way

This seems similar to the above issue. Looks like when the color shifting work happened, this part of the component wasn't properly variable-ized (hardcoded to the color var itself and not the component token) so it didn't update with the rest of the component. I assume we want this border to match the rest of the input border?

Screenshot 2023-03-01 at 10 35 27 AM

packages/design-system/src/styles/components/_SingleInputDateField.scss
```diff
@@ -519,7 +519,7 @@
     text-decoration: none;
 
     &:after {
-      background: var(--color-gray-light);
+      background: var(--text-input__border-color);
```
pwolfert, line 522: I went down this path when troubleshooting and concluded that this is not the correct color. Because of the subtle blue backgroundn of the button, I think our designers opted for a lighter color. The image update must have happened on account of the gray shift that happened in an earlier PR (but for some reason the image didn't get flagged as changed). I think we should put this in its own component variable and assign it to the gray-light

packages/design-system/src/styles/components/_Dialog.scss
pwolfert: Ah okay, good call

pwolfert (approved):
Besides the border color item below, this LGTM. Great work!

PR #2396, opened 2023-03-02
[WNMGDS-2202] Components to only have top spacing

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2202

Some of these components didn't technically have bottom _margin_, but they did have bottom _spacing_. In order to be more consistent with spacing, we want components to only have built-in spacing at the top and not the bottom so spacing doesn't stack between components.

## How to test

Check the affected storybook stories. Leaning on visual regression tests

## Checklist


packages/design-system/src/styles/components/_ChoiceList.scss
```diff
@@ -237,7 +237,8 @@ label,
 .ds-c-choice__checkedChild {
   border-left: $spacer-half solid var(--choice__border-color--left);
   flex: 1 1 100%;
-  margin-bottom: $spacer-2;
+  margin-block-end: $spacer-2;
```
zarahzachz, line 240: Could be shorthanded to: `margin block: $spacer-1 $spacer-2`

pwolfert:
@zarahzachz, I just discovered that the visual changes to the month picker exist on `epic/css-reset` as well and that it's stemming from [this reset that adds margin to checkbox labels](https://github.com/CMSgov/design-system/blob/epic/css-reset/packages/design-system/src/styles/_reset.scss#L173-L177).

What's the best path to move forward here? It wouldn't be very nice to set that in the reset and then undo it in the component styles. Should we get rid of it? Should we find a way to make something work for both raw checkboxes + labels and the choice component? In the choice component, the spacing comes from the `gap` property.

The particular reason the month picker is changing is that the added spacing is causing only 3 columns to fit in the space allotted instead of 4.

forrestbaer (approved):
Looks good in my tests

zarahzachz:
[Is this VRT correct](https://github.com/CMSgov/design-system/pull/2396/files#diff-967bbc7b1b94958f08bff1179277f1e362d5f2937c28cb133ff262f66941b1ef)? Don't think the 2nd radio btn should have a red outline and I feel like I've seen this bug before...

zarahzachz:
[Another weird looking error state](https://github.com/CMSgov/design-system/pull/2396/files#diff-48c84bd163664e7548f958572b52da19a8cbadd0f8fe5014a70f38017772f027). I'd give it a 10 on the Patrick-Wonky scale.

zarahzachz (approved):
Your work looks good. Saw some weird error states on radios and checkboxes and I'm not sure how they got there.

pwolfert:
> [Is this VRT correct](https://github.com/CMSgov/design-system/pull/2396/files#diff-967bbc7b1b94958f08bff1179277f1e362d5f2937c28cb133ff262f66941b1ef)? Don't think the 2nd radio btn should have a red outline and I feel like I've seen this bug before...

Oh that does seem weird, but it existed before. In both these cases it's the healthcare theme. [In the CSS](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/styles/components/_ChoiceList.scss#L100) it's governed by two theme variables [that have a different value in core](https://github.com/CMSgov/design-system/blob/main/packages/design-system-tokens/src/themes/core-component.ts#L257) and in healthcare, and it was done on purpose. I don't think it was meant for the case where you have selected one; I think it was meant for attestations where you check or don't check a lone checkbox or radio button. Might be worth revisiting, though, and making sure we have sufficient guidance on the use case and documentation for that variation in the design systems.

PR #2398, opened 2023-03-02
[WNMGDS-2214] Pagination bugfix for Safari 13, layout logic moved to styles.

forrestbaer:
## Summary

- Removed logic in `Pagination` for switching the view from moblie/desktop.
- Added that logic into styles.
- Spacing looks changed by snapshot updates however if you compare to the current live pagination, it's the same, or what it should be. A bit confused by this but it seems correct.

## How to test

1. Validate against current pagination component: https://design.cms.gov/storybook/?path=/story/components-pagination--default

## Checklist

ated unit tests to cover any new or modified code


packages/design-system/src/styles/components/_Pagination.scss
```diff
@@ -15,6 +15,9 @@
       display: inline-flex;
       gap: 1ch;
     }
+    @media (min-width: $media-width-md) {
```
pwolfert, line 18: The previous logic actually corresponded to the `sm` breakpoint

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -285,19 +262,19 @@ function Pagination({
...
-              number: `<strong>${currentPage}</strong>`,
-              total: `<strong>${totalPages}</strong>`,
-            }),
-          }}
-        />
-      ) : (
-        <ul role="list">{pages}</ul>
-      )}
+      <span
+        className={`ds-c-pagination__page-count${!compact ? ' ds-u-display--none' : ''}`}
```
pwolfert, line 266: Right now there's a case where both are shown at once, whereas previously the two views were in a mututally exclusive ternary.

Screenshot 2023-03-02 at 11 32 15 AM

I think the logic could be further pushed to CSS by removing the ternaries on these individual elements and instead placing a `ds-c-pagination--compact` conditionally on the top level component. That would allow the CSS to take advantage of that class to say "if that class exists, hide the list and show the page count; if it doesn't exist, use the media queries to govern which one is shown". If we don't want to have to use `:not()` selectors, we could have an opposite class to `--compact` like `--wide` or something. Moving it all to CSS also grants applications the opportunity to override the behavior if needed.

pwolfert (commented):
I'm also confused why the VRT refs changed, but it looks like it might have fixed a bug in the forced-colors mode. The screenshots are all for forced-colors mode, and when I compare forced-colors mode of this branch against non-forced-colors mode of the main branch, the spacing is now the same.

packages/design-system/src/components/Pagination/Pagination.tsx
forrestbaer: Good points, updated, thanks.

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -144,7 +144,7 @@ function Pagination({
   totalPages,
   ...rest
 }: PaginationProps): React.ReactElement {
-  const classes = classNames('ds-c-pagination', className);
+  const classes = classNames('ds-c-pagination', { 'ds-c-pagination__compact': compact }, className);
```
pwolfert, line 147: To follow BEM syntax, this should be `ds-c-pagination--compact`. Variations or states are signified by `--`, whereas `__` is for child elements.

pwolfert (approved):
It works great! One minor point below

packages/design-system/src/components/Pagination/Pagination.tsx
forrestbaer: oh, right! been a while since I've actually needed to create any classes.. Actually I think this is the first time I've touched CSS outside of token stuff since I started here believe it or not 😆

PR #2399, opened 2023-03-03
[WNMGDS-2171] CSS reset documentation

zarahzachz:
## Summary

WNMGDS-2171

- Adds documentation showing how to use the new CSS reset and what major changes it introduces
- Removes language around needing `.ds-base` to use the design system
- Tweaked some docs site CSS to better work with the reset

## How to test

Review [demo link](https://cmsgov.github.io/design-system/branch/WNMGDS-2171/css-reset-docs/migration-guides/introducing-a-css-reset/?theme=core) for changes.


packages/docs/content/migration-guides/introducing-a-css-reset.mdx
```diff
@@ -0,0 +1,57 @@
...
+## Global defaults
+
+The CSS reset introduces globally-defined defaults that will impact the entire application once the CSS. These defaults include (but aren't limited to):
+
+- `box-sizing: border-box` is set on every element — including `*::before` and `*::after`. This ensures that the declared width of element is never exceeded due to padding or border.
+- The `<html>` element has `scroll-behavior: smooth` set on `:focus-within`. This is a CSS-only solution that provides smooth anchor scrolling.
+- Animations, transitions, and smooth scroll are removed on all elements - including `*::before` and `*::after` for users who have `prefers-reduced-motion` enabled to respect their preferences.
+- The `<body>` element effectively replaces the `.ds-base` class, defining `font-family`, `font-size`, `line-height`, and `color`.
+  - In addition, `<body>` now sets `background-color` to a theme's specified background.
+- The `outline` is removed on elements with the `tabindex="-1"` declared. This is removes focus rings for non-interactive elements, e.g., when activating a skip link will then send focus to a `<main>` or heading element.
```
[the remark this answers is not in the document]
pwolfert:
```suggestion
- The `outline` is removed on elements with the `tabindex="-1"` declared. This is removes focus rings for non-interactive elements. For example, when activating a skip link it will usually send focus to a `<main>` or heading element.
```
This might make it easier to read and simplifies the sentence structure

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
```diff
@@ -0,0 +1,57 @@
...
+- To avoid [margin collapse](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Box_Model/Mastering_margin_collapsing), all elements will only rely on their respective top margins to better control layout. All bottom margins have been removed.
+
+## Typography
+
+### Headings
+
+All heading elements will receive the font styles defined by their parent element using the `font: inherit` rule. 
+
+[`font`](https://developer.mozilla.org/en-US/docs/Web/CSS/font) is a shorthand property that encompasses several font styles, but the key thing to note is a heading will now inherit the `font-family` and `font-size` of its parent or `<body>` element. This rule ensures the semantics of a heading element are distinct from its appearance. If you need to style a heading, refer to our [headings](/foundation/typography/headings/) or [utility class](/utilities/overview/) documentation.
```
pwolfert, line 30: > This rule ensures the semantics of a heading element are distinct from its appearance. If you need to style a heading, refer to our [headings](/foundation/typography/headings/) or [utility class](/utilities/overview/) documentation.

💯

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+Lastly, headings have been reset to have their bottom margins removed. All headings have a top margin set to `1.5em`.
```
pwolfert, line 32: Should we explain why it's `1.5` instead of on the 8px grid? Because it's relative to the heading's font size?

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+### Paragraphs
+
+Like headings, paragraphs have had their bottom margins removed and a top margin of `1em` defined. However, if a paragraph is the `:first-child` its margin will be entirely removed.
```
pwolfert, line 36: > However, if a paragraph is the `:first-child` its margin will be entirely removed.

Should we give the same treatment to headings? Has that been talked about?

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+### Links
+
+Links behave as they previously did, with one exception - the `text-decoration-color` is now set to `currentColor`. This change was made under the assumption that all link underlines will have the same color as its text, regardless of that text's state.
```
pwolfert, line 40:
```suggestion
Links behave as they previously did, with one exception—the `text-decoration-color` is now set to `currentColor`. This change was made under the assumption that all link underlines will have the same color as its text, regardless of that text's state.
```

Emdash with no spaces is what you want here, I believe.

I'm also inclined to use a colon for this sort of thing. I've gotten more liberal with the `:` over the years. Some may say that the semi-colon is more appropriate, but I don't think most readers expect it in this sort of situation.

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+## Lists
+
+The `<ol>` and `<ul>` elements have been styled to match their [`.ds-c-list`](/foundation/typography/list/) counterpart, with one key difference - they use `flex` to better control list item spacing and assist with commonly used inline list patterns.
+
+If you find yourself in need of an inline list, you can add `flex-direction: row` locally to your project. Likewise, if the spacing between list elements isn't to your liking, `gap` can be changed to the value that fits your project best.
+
+One last change worth mentioning for lists is that unstyled lists are defined under the `role="list"` attribute. This makes something like `<ul role="list">` visually the same as applying the `.ds-c-list--bare` class. Because unstyled lists have [accessibility concerns](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type#accessibility_concerns), we opted to include this style rule to ensure proper markup is used.
```
pwolfert, line 48: Love it

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
```diff
@@ -0,0 +1,57 @@
...
+
+## Form elements
+
+Various form elements have been rebooted for simpler base styles. Here are some of the most notable changes:
+
+- The `<input>`, `<button>`, `<textarea>`, and `<select>` elements receive the font styles defined by their parent element using the `font: inherit` rule. 
+- All `:disabled` fields have `cursor: not-allowed` to prevent user confusion.
```
pwolfert, line 55: Did you talk to Dan about this? Or did this stem from some other source?

packages/docs/src/styles/components/navigation.scss
```diff
@@ -25,6 +25,7 @@
 
   .c-navigation__link-list {
     border-block: 1px solid var(--vertical-nav-item__border-color);
+    gap: 0;
```
pwolfert, line 28: Fixing the nav panel might need its own pull request. Even with this in place, I'm seeing way more spacing between items and a narrower panel overall.

Screenshot of the navigation panel being much narrower than before and having more vertical spacing between nav links

packages/docs/content/foundation/base.mdx
```diff
@@ -1,27 +1,12 @@
 ---
-title: Base
+title: Inverse Base
```
pwolfert, line 2: 👍 good call keeping the original file

pwolfert (commented):
This documentation is wonderful! I really appreciate how you've explained the why behind every step. It's well organized too.

zarahzachz:
I was wondering if it would be helpful linking to our RFC for this somewhere? I initially went down the route of explaining our reason why we were adding a reset but realized the RFC gives so much of that context already.

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
zarahzachz: That's a good thought. Our heading style classes currently handle that behavior, but we could easily lift that to the HTML elements

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
zarahzachz: This was something Eric Bailey and I discussed during the HCM work. Mainly, disabled states typically rely on color alone to indicate what it is and adding the cursor was just an extra indicator for users. But we can get confirmation from Dan just in case.

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
zarahzachz: I'm not sure if we need to explain why it's `1.5em`... part of me wonders if this line should be removed entirely. I do specify on line 22 that we're pivoting to using top margins only and maybe that's enough? By hard-coding a value here, does this run the risk of falling out of date with the codebase as we go? And does this margin apply to all themes? I think including this info raises more questions than it solves.

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
zarahzachz: Should I update the reset to include the `:first-child`/`:last-child` rules from our [headings CSS](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/styles/_base.scss#L158)?

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
pwolfert: Yeah, that's a good point. I think it's nice to at least call out that it has top margins and not bottom margins

packages/docs/content/migration-guides/introducing-a-css-reset.mdx
pwolfert: Oh, I didn't know we were doing that! Yeah, I think moving that into the reset is a good idea, especially if the values don't change per heading level.

pwolfert:
Just noticed this VRT change, and it makes me wonder if we should update the button story to stick that default button in the flex div that contains all the other variations so we can avoid the trimming of the top margin and to simplify the story and decrease the code indendation

Screenshot of unexpected change to the all-buttons stories

pwolfert (approved):
Great work! One comment above about an unexpected VRT reference image change and how we could fix it, but the docs LGTM

PR #2403, opened 2023-03-07
[WNMGDS-2222] Upgrade to TS 4.9 and stop referencing dist directly in medicare

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2222

Separating a TypeScript upgrade and type-import cleanup in medicare from the main set of changes for Preact in order to make that upcoming pull request easier to review.

## How to test

Run things like storybook and `yarn type-check`

## Checklist


packages/design-system/src/components/Tabs/Tabs.tsx
```diff
@@ -1,6 +1,6 @@
 import React, { useState, useRef } from 'react';
 import Tab from './Tab';
-import TabPanel from './TabPanel';
+import TabPanel, { TabPanelProps } from './TabPanel';
```
forrestbaer, line 3: Curious why `TabPanel` wasn't also made into a named import here? Because it's also the default export?

pwolfert: Ah, just because it didn't affect what I was doing. This module changed because of an error that came up in the TS upgrade, but all the icon changes were so that the types in those modules would make it into the design system exports so they could be accessed from `ds-medicare-gov`

pwolfert: I didn't make a conscious decision about the exports here

forrestbaer (approved):
All my tests have passed, looks good! 🧹

PR #2409, opened 2023-03-08
[NO-TICKET] VRT logic fix and snapshot update

forrestbaer:
## Summary

The logic that was previously working for us was broken during the interaction tests updates.  This logic updates the tests to work as before. The new snapshots are a result of the updated snapshot naming presented in the interaction test updates.

- updated our npm scripts so they're all more useful when locally testing with docker.
- bumped playwright to version 1.31 from 1.30

## How to test

1. `yarn test:smoke` `yarn test:browser:all` `yarn test:a11y`

## Checklist


tests/browser/story-screenshots.test.ts
```diff
@@ -25,18 +25,16 @@ Object.values(stories).forEach((story) => {
...
 
-      // For smoke tests, only capture core components in the core theme and theme-specific
-      // components in their native theme
-      if (
-        isSmokeTest &&
-        !(theme === 'core' || !story.importPath.includes(themes[theme].packageName))
-      )
-        return;
+      // During smoke tests, only take screenshots in core of core components
+      if (isSmokeTest && theme !== 'core') return;
```
forrestbaer, line 37: I guess I should have added `&& !storyNotInCore` to this, updating now.

forrestbaer:
How would we feel about wiping out the current `snapshots` folder with these changes to the logic and getting a fresh set of shots to compare against?  As a result of our iterating on this process we have a lot of snapshots which aren't being used with the latest changes.

forrestbaer:
It appears `tests/browser/snapshots/foundations-typography-headings--heading-2-xl-healthcare--firefox.png` is still blank, continuing to investigate.

packages/design-system/src/components/Autocomplete/Autocomplete.test.interaction.ts
```diff
@@ -9,8 +9,8 @@ Object.keys(themes).forEach((theme) => {
 
   test(`Autocomplete select  hover: ${theme}`, async ({ page }) => {
     await page.goto(`${storyUrl}&globals=theme:${theme}`);
-    const elem = page.getByRole('listbox');
-    await elem.type('c', { timeout: 200 });
+    const elem = page.getByRole('combobox');
+    await elem.type('c');
```
pwolfert, line 13: Nice that we could get rid of the timeout too!

forrestbaer:
Also I just noticed the date is going to change and break interaction every time we open that `MonthPicker` going to have to set the date to something static.

tests/browser/playwright.config.ts
```diff
@@ -84,7 +84,7 @@ const config: PlaywrightTestConfig = {
   /* Opt out of parallel tests on CI. */
   workers: isCI ? 1 : undefined,
   /* Reporter to use. See https://playwright.dev/docs/test-reporters */
-  reporter: [['html', { outputFolder: 'playwright-report' }]],
+  reporter: [['html', { open: 'never' }]],
```
pwolfert, line 87: Why the change here? My understanding is that this would have outputted a browseable html page to see the failures

pwolfert (commented):
With the `datefield--picker-interaction` test, it looks like we should actually change that story to always start on the same date; otherwise, we're going to get a new result every day. Could be done with args (`&args=fromYear:2024;toDate:1678221003513`), but we don't have a mechanism for providing per-story args in our tests 😞

tests/browser/playwright.config.ts
forrestbaer: It still generates the html it just doesn't try to open it automatically which I found to be annoying. The notes it gives after running are `npx view-report` to view the report if you want to.

forrestbaer: It also may someday cause problems in CI

pwolfert: If we don't specify the `outputFolder`, will it default to some folder name that is relative to the tests?

forrestbaer: It goes to the same place, the output folder we had specified was same as the default.

pwolfert (approved):
Approving so we can get this lined up with the current CI job, and we can come back and fix the calendar picker story later

PR #2422, opened 2023-03-14
[WNMGDS-1611] Move privacy settings modal to core

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1611

- Moves the privacy settings modal into core and makes it flexible enough to work for multiple domains
- Reworks the cookie-setting module to get rid of the `js-cookie` dependency
- Adds a Storybook story for the privacy settings modal on its own
- Simplifies the settings table styles by using what we already have in the design system
    - The mobile view is where all the extra code is coming from, but I don't see why we wouldn't just use our existing stacked table that creates an accessible mobile experience
- Fixes an issue in our stacked table story that I discovered through unit-testing the privacy settings table
    - It was missing props and was printing console warnings

## How to test

1. Try out the new story
2. Try out the old healthcare footer story

## Checklist


forrestbaer (commented):
This looks great! Nice to have it in core. Since this contains an API change should this be `Type: Breaking?`

pwolfert:
> Since this contains an API change should this be `Type: Breaking?`

@forrestbaer, as far as I can tell, we never actually exposed this component by itself either in JavaScript or with TypeScript types, so its API was hidden.

forrestbaer (approved):
I like it!

PR #2425, opened 2023-03-14
[WNMGDS-2238] Healthcare privacy settings modal button css cascading incorrectly.

forrestbaer:
## Summary

- updates hover color for privacy settings button in healthcare footer to match other links
- updates cascade to only style the first child button in this case so it doesn't override button in privacy settings modal

## How to test

1. `yarn clean && yarn install && yarn build && yarn storybook`
2. Open [http://localhost:6006/?path=/story/healthcare-footer--basic-footer&globals=theme:healthcare](http://localhost:6006/?path=/story/healthcare-footer--basic-footer&globals=theme:healthcare)
3. Inspect privacy settings modal button, confirm it does not include reset styles from `.hc-c-footer__inline-item > button`

## Checklist


packages/ds-healthcare-gov/src/styles/components/_Footer.scss
```diff
@@ -43,8 +43,9 @@
     padding: 0;
     text-decoration: none;
 
+    &:hover,
     &:focus {
-      @include mixins.focus-styles;
+      color: var(--color-primary-darker);
```
pwolfert, line 48: Why was this change made to focus? What was there before wasn't perfect either, but this is what it looks like now:

focus

pwolfert (approved):
Nice and concise fix 👍
Question below about the other change

packages/ds-healthcare-gov/src/styles/components/_Footer.scss
forrestbaer: Oh shoot didn't notice the focus state, I'll make that only on hover. The hover state previously is grey and should be blue.

forrestbaer: Updated a couple things here to resolve this
image

pwolfert: Nice, thank you 👍

PR #2431, opened 2023-03-16
[WNMGDS-2241] Make external links screen reader accessible

zarahzachz:
## Summary

WNMGDS-2241

External links should inform all users that interacting with them will take the user to a new website.

Check the [demo url](https://cmsgov.github.io/design-system/branch/WNMGDS-2241/BUGFIX-external-link-a11y/patterns/third-party-external-link/?theme=core). When using a screen reader, you should be informed the link takes you to a new website.


packages/design-system/src/components/Icons/ExternalLinkIcon.tsx
```diff
@@ -5,6 +5,7 @@ import { SvgIcon, IconCommonProps } from './SvgIcon';
 const defaultProps = {
   className: '',
   viewBox: '0 0 512 512',
+  ariaHidden: false,
```
zarahzachz, line 8: Icons have `aria-hidden` set to true by default. When setting this prop to false, it enables certain a11y features (`aria-labelledby` and `role="img"` specifically). 

I'm making an assumption ExternalLinkIcon will always be used in context where screen readers will need to understand its presence means there's an external link on the page and **setting this component to always reveal a11y**. This change will roll out to all apps using this icon component.

Alternatively, I could set the `ariaHidden` prop on the Story itself to illustrate to others how to make this icon accessible, but that requires teams go back and update their code manually. 

I figured an opt-out approach might be warranted for this specific icon - if apps don't this icon to be visible accessibly for some unknown reason, they can overwrite this default prop with `true`. Thoughts?

packages/design-system/src/components/patterns/thirdPartyExternalLinks.stories.tsx
```diff
@@ -13,11 +13,14 @@ export const ThirdPartyExternalLink = () => {
...
     <>
       <Button
-        className="ds-c-button ds-c-button--ghost ds-u-font-weight--normal"
+        className="ds-c-button--ghost ds-u-font-weight--normal"
         onClick={() => setShowModal({ showModal: true })}
       >
         Link to external site
-        <ExternalLinkIcon className="ds-u-margin-left--05" />
+        <ExternalLinkIcon
+          title="This link goes to an external site"
```
zarahzachz, line 21: The ticket had specific verbiage on what the `title` text should say, so I've used that exactly here, overwriting the default title set in the locale files. 

The locale files just say "External link" (in English and Spanish) and I've left that intact. Is that default title text enough context to inform users of that icon's meaning? Or should I update our locale files with this new content? If an update is recommended, I need a Spanish speaking resource to confirm a Google translate version is correct.

zarahzachz:
Something I was thinking about while updating the ExternalLinkIcon component: Why isn't some of the pattern logic embedded within the Button component? 

Our [docs](https://cmsgov.github.io/design-system/branch/WNMGDS-2241/BUGFIX-external-link-a11y/patterns/third-party-external-link/?theme=core) explicitly say **Any non .gov and .mil links will be treated as external whether they belong to a government agency or not, no exceptions**. To me, this line raises two points:

1. The inclusion of the word "will" implies if an `href` value doesn't end with ".gov" or ".mil" it will receive the external link icon (perhaps this confusion was the cause of this [Slack question](https://adhoc.slack.com/archives/C645GLLNN/p1677638178171039)?)
    2. Replacing "will" with "must" might better indicate this is on teams to update themselves.
3. Why doesn't our Button component enforce the requirement that certain domains require this icon?
    4. Maybe this extends our Button component too far, but it seems like we could have logic in place to determine if the `href` prop has a value ending with ".com", ".gov", ".mil", etc. If a url is detected that doesn't end with ".mil" or ".gov", it should include the icon with its required utility classes to take the burden off our users.
    5. Enforcing this rule at the component-level might catch instances where teams aren't following this rule.
    6. I think it still makes sense to keep the ThirdPartyExternalLink pattern to illustrate how the modal should work (...or should this whole pattern be its own component?)

pwolfert:
@zarahzachz, these are good questions.
1. This makes sense to me, and maybe it would also be a good idea to include references to the external link documentation in the guidance for [links](https://design.cms.gov/foundation/typography/links/?theme=core) and [buttons](https://design.cms.gov/components/button/)
2. There might be enough edge cases to make it difficult to accurately detect automatically when an external link should be used. Variant of that idea: If we added that logic to the button component but enabled it through a boolean prop, we could console log warnings in development when we detect external links that don't have that external flag.
3. The other thing we could consider is promoting this pattern to its own component

packages/design-system/src/components/Icons/ExternalLinkIcon.tsx
pwolfert: I agree. Good thinking 👍

zarahzachz:
> @zarahzachz, these are good questions.
> 
> 1. This makes sense to me, and maybe it would also be a good idea to include references to the external link documentation in the guidance for [links](https://design.cms.gov/foundation/typography/links/?theme=core) and [buttons](https://design.cms.gov/components/button/)
> 2. There might be enough edge cases to make it difficult to accurately detect automatically when an external link should be used. Variant of that idea: If we added that logic to the button component but enabled it through a boolean prop, we could console log warnings in development when we detect external links that don't have that external flag.
> 3. The other thing we could consider is promoting this pattern to its own component

Would it be worth putting this topic on next week's co-working agenda? I was going to suggest making a Jira ticket, but it seems like we're torn between enhancing the Button or creating a new ExternalLink component and I'm sure the team would like to discuss.

pwolfert (approved):
Tested it with VoiceOver, and it sounds good to me 👍 

In the pattern story the messaging is a bit redundant, but this ticket is about making the icon meaningful on its own.

PR #2433, opened 2023-03-16
[WNMGDS-2253] Updates to release note generation script

forrestbaer:
## Summary

- Now skips over `.incomplete` themes
- Added cases for `internal` and `documentation`
- Updated prompt tooling, `node:readline` is a nightmare, `readline-sync` is perfect.
- Added list of milestones to choose from
- Fixed some internal logic
- Added edge case for when a ticket is labeled `[WNMGDS-2222][WNMGDS-3333]` which probably shouldn't be a thing but I ran into two of them.
- Added an `Impacts: Internal` label
- Fixed milestones not working when they had spaces in the name

## How to test

1. `yarn clean && yarn install && yarn build && yarn release:notes`

## Checklist

### DIFF
--- a/scripts/notes.ts
+++ b/scripts/notes.ts
@@ -1,8 +1,9 @@
 import { execSync } from 'node:child_process';
 import chalk from 'chalk';
-import readline from 'readline';
 import themes from '../themes.json';
 import { writeFileSync } from 'node:fs';
+
+const prompt = require('readline-sync');
 
 interface PRDetails {
   author: string;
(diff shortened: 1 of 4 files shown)


scripts/notes.ts
```diff
@@ -161,19 +178,35 @@ const makeNotesMD = (notes: any[]): string => {
...
+      if (!sameSystem || (!sameSystem && !sameType)) {
+        // the first item from this group, let's print a title and the first type
+        if (theme[sys]) {
+          md += `## [${theme[sys].longName}](${theme[sys].urlNpm}) [${versions[sys]}]\n`;
+        } else {
+          // we only have two outliers that do not exist in our theme file
+          // internal, and documentation, which each need their place in the
+          // release notes.
+          const sysUrl =
+            sys === 'internal'
```
pwolfert, line 198: One question: What's the difference practically between `Impacts: Internal` and `Type: Internal`? I can see [an example of a pull request](https://github.com/CMSgov/design-system/pull/2409) with `Impacts: Internal` and `Type: Fixed`, but would it be inappropriate to either 1) label it as `Impacts: Core` and `Type: Internal` or 2) Leave it out of the notes entirely by not tagging it? Yesterday when I walked through all the merged PRs, I figured it didn't really affect anyone and might not need to be included in the release notes

forrestbaer: I would be happy to remove all internal notes from our release notes from here on out, just give me the thumbs up.  This was just something I added because it's been used.

pwolfert: Some internal things might be worth mentioning. For example, other people might be interested to know that we "Added a yarn `check:maturity` script to gather component maturity statistics". However, other things like a `[NO-TICKET]` fix to something that was already covered by another PR represented in the notes probably doesn't need to be tagged, and therefore I don't think we need to have two places for internal things to exist. I'd say the method you had before for putting it under an internal subsection of one of the packages was enough.

forrestbaer: I think having an internal section under a child system could be confusing. I would vote we either keep it in a separate section or not include those items in our milestones.

pwolfert: Internal sections have been how we've done it for a long time, where the internal items pertain to the thing (like internal changes to the doc site), but they don't always correspond to a particular product. Okay, let's remove the `Type: Internal` and use the `Impacts: Internal` instead and use it only for things that are significant, worth mentioning, or have tickets associated with them

pwolfert (approved):

PR #2437, opened 2023-03-20
[WNMGDS-2231] Update Button docs

zarahzachz:
## Summary

WNMGDS-2231

Update Button docs to include verbiage defined in ticket. 
- I included this text within the Usage section because it felt most appropriate. 
- I also removed some previous guidance as it somewhat duplicated the new guidance.
- I removed guidance about styling the disabled buttons as I believe it was to ensure the disabled class was being applied to disabled button links. Because we're styling `aria-disabled` as a disabled button, we're using this attribute to enforce styling; this attribute is documented as required for a disabled button link in the Accessibility section. 
- I added more context around disabled link buttons and how to make them properly disabled.

To test, visit [preview link](https://cmsgov.github.io/design-system/branch/WNMGDS-2231/update-disabled-button-guidelines/components/button/#guidance) (building).

### DIFF
--- a/packages/docs/content/components/button.mdx
+++ b/packages/docs/content/components/button.mdx
@@ -346,7 +346,7 @@ Consider when to use links vs. when to use buttons. Links and buttons are read d
 - When styling links to look like buttons, remember that screen readers handle links slightly differently than they do buttons. Pressing the `Space` key triggers a button, but pressing the `Enter` key triggers a link.
 - Dimmed or unavailable buttons using the `<a>` tag must have the `aria-disabled` attribute set to true and its `href` attribute removed. 
     - Unlike the `<button>`, `disabled` is not a valid attribute for `<a>`. To describe the disabled state to screen reader users, use the `aria-disabled` attribute instead.
-    - The `href` attribute is not required on `<a>` and when it's absent it doesn't create a hyperlink.
+    - The `href` attribute is not required on `<a>`, and when it's absent it doesn't create a hyperlink.
 - Dimmed or unavailable buttons using the `<button>` tag should have the `disabled` attribute applied. This removes native click and keypress events from the button. It also prevents automated scanners from logging a low contrast error. Finally, it announces the button as "dimmed" or "disabled" to screen readers, offering users additional context.
 
 
--- a/packages/docs/content/components/button.mdx
+++ b/packages/docs/content/components/button.mdx
@@ -309,10 +309,16 @@ When pairing 2 CTAs:
 - Confirm the user meant to trigger a destructive action before following through with the action.
 - Provide a method for a user to undo a destructive action.
 
-**Disabled buttons**
+**Avoid using disabled buttons**
 
-- Don’t disable buttons, unless there’s a good reason to.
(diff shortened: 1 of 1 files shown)


packages/docs/content/components/button.mdx
```diff
@@ -336,9 +342,12 @@ Consider when to use links vs. when to use buttons. Links and buttons are read d
...
-- Dimmed or unavailable buttons should have the `disabled` attribute applied. This removes native click and keypress events from the button. It also prevents automated scanners from logging a low contrast error. Finally, it announces the button as "dimmed" or "disabled" to screen readers, offering users additional context.
+- Dimmed or unavailable buttons using the `<a>` tag must have the `aria-disabled` attribute set to true and its `href` attribute removed. 
+    - Unlike the `<button>`, `disabled` is not a valid attribute for `<a>`. To describe the disabled state to screen reader users, use the `aria-disabled` attribute instead.
+    - The `href` attribute is not required on `<a>` and when it's absent it doesn't create a hyperlink.
+- Dimmed or unavailable buttons using the `<button>` tag should have the `disabled` attribute applied. This removes native click and keypress events from the button. It also prevents automated scanners from logging a low contrast error. Finally, it announces the button as "dimmed" or "disabled" to screen readers, offering users additional context.
```
pwolfert, line 350: I wonder if this item needs to be updated to recommend `aria-disabled` as well. Our all-buttons story hasn't been switched to use `aria-disabled`, so I'm unsure if that bit about automated accessibility tests still applies to `aria-disabled`. I'm going to test that.

pwolfert (commented):
Good edits 👍 I'm going to do a little test to see what we should do for the bullet point that still references `disabled`

packages/docs/content/components/button.mdx
pwolfert: Looks like axe-core is totally fine with these being `aria-disabled` instead of `disabled`. Tried it with `yarn test:a11y` as well as in my own browser.

pwolfert: I have the local changes for updating that story to use `aria-disabled`. How would you feel about adding that to this PR or at least including it under this ticket? I think that change fell through the cracks between tickets, but it would make sense to me to update that story

zarahzachz: Do we want to use `aria-disabled` instead of `disabled` on buttons? [It seems like `disabled` provides some features `aria-disabled` doesn't](https://css-tricks.com/making-disabled-buttons-more-inclusive/#aa-the-difference-between-disabled-and-aria-disabled).

pwolfert: Are you concerned that if we start recommending developers use `aria-disabled` they won't necessarily account for clicks on those disabled buttons? And that we need to explain how to handle an `aria-disabled` button a little more? My understanding was that we wanted to move away from `disabled` full stop.

This conversation doesn't need to hold up the other changes you've made, so I'm approving.

pwolfert (approved):

packages/docs/content/components/button.mdx
zarahzachz: That's kinda what I'm getting at. It's kind of a combo of things:
- I don't think our Button component has this new logic baked in
    - We're only applying `aria-disabled` on links
    - We don't have logic to remove click events from elements with this attr
- It would require additional guidance to explain how to remove the click event, I worry teams will just start using `aria-disabled` without fully understanding the difference

PR #2438, opened 2023-03-20
[WNMGDS-2040] CDN Index build script

forrestbaer:
## Summary

- Adds a script to build `index.html` files in the `dist` folders during build utilizing the latest version data from each child systems `package.json`
- Since this will only be updated live when we publish with jenkins, the versions will be up to date and correct/different.
- This is 1/3rd of this total ticket, the other 2/3rd's have been pushed live at [design.cms.gov/cdn](https://design.cms.gov/cdn) for browsing.
- @pwolfert I think the jenkins work needs another ticket, to keep it organized. I can create it. The only jenkins work that needs done is ensuring the `version/index.html` file is copied and the main `cdn/index.html` is updated with latest versions after a successful publish. Or maybe not another ticket, I'll be working on it, shouldn't bee too difficult, I've tested the main index with the append method and it works.

## How to test

1. visit [design.cms.gov/cdn](https://design.cms.gov/cdn), you can `wget https://design.cms.gov/cdn/` to see the actual source, the browser automatically fills in some fields that don't exist in the actual file.
2. `yarn clean && yarn install && yarn build`, inspect dist folders new index.html files

## Checklist

### DIFF
--- a/scripts/build-cdn-indexes.ts
+++ b/scripts/build-cdn-indexes.ts
@@ -58,6 +58,11 @@ getSystems().forEach((sysinfo) => {
             <li>DS CSS:<br> <code class="ds-u-fill--gray-lightest ds-u-padding--1 ds-u-margin--1 ds-u-display--inline-block"><a href="https://design.cms.gov/cdn/${system}/${version}/css/index.css">https://design.cms.gov/cdn/${system}/${version}/css/index.css</a></code></li>
             <li>Theme:<br> <code class="ds-u-fill--gray-lightest ds-u-padding--1 ds-u-margin--1 ds-u-display--inline-block"><a href="https://design.cms.gov/cdn/${themeCssPath}">https://design.cms.gov/cdn/${themeCssPath}</a></code></li>
           </ul>
+        <h3 class="ds-text-heading--md">The following assets are also available for this version:</h3> 
+          <ul>
+            <li>Preact JS Bundle:<br> <code class="ds-u-fill--gray-lightest ds-u-padding--1 ds-u-margin--1 ds-u-display--inline-block"><a href="https://design.cms.gov/cdn/${system}/${version}/js/preact-components.js">https://design.cms.gov/cdn/${system}/${version}/js/preact-components.js</a></code></li>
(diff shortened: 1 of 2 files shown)


zarahzachz (approved):

scripts/build-cdn-indexes.ts
```diff
@@ -0,0 +1,74 @@
...
+import fs from 'node:fs';
+import c from 'chalk';
+
+const getSystems = () => {
+  const result: string[][] = [];
+  Object.entries(themes).forEach((theme) => {
+    const config = theme[1];
+    if ('incomplete' in config) return;
+
+    const shortname =
```
pwolfert, line 12: Looks like the only place this is used is to get the theme file name, which should actually match the key in this object. The `config.packageName.split('-')[1]` won't work for `cmsgov-theme.css` as far as I can tell. Instead of having any logic here at all, could you change the parameters part to something like

```
Object.entries(themes).forEach(([themeName, themeInfo]) => {
...
result.push([themeInfo.packageName, version, themeName]);
```

scripts/build-cdn-indexes.ts
```diff
@@ -0,0 +1,74 @@
...
+    <script type="text/javascript" src="https://design.cms.gov/cdn/${system}/${version}/js/react-dom.production.min.js"></script>
+    <script type="text/javascript" src="https://design.cms.gov/cdn/${system}/${version}/js/bundle.js"></script>
+  </head>
+  <body class="ds-base" style="margin: 0">
+    <div id="usa-banner"></div>
+    <script>
+      ReactDOM.render(
+        React.createElement(DesignSystem.UsaBanner),
+        document.getElementById('usa-banner')
+      );
```
pwolfert, line 45: Out of scope now, but it would be kind of cool to use the web component version here. I guess it depends on how we end up using these for monitoring.

pwolfert (approved):
It's working great! One suggestion below to remove some code

scripts/build-cdn-indexes.ts
forrestbaer: Hmm I was thinking it would be called `ds-cms-gov`, ok, yeah that's easier in that case.

scripts/build-cdn-indexes.ts
forrestbaer: Yeah sure, could always pivot to that, didn't want to include anything 'experimental' :)

PR #2441, opened 2023-03-23
[NO-TICKET] Made small updates to release note script after first live run

forrestbaer:
## Summary

- Updated sorting, was using a method that didn't really work in all cases
- Fixed a couple eslint errors that were new
- Removed -test from end of note title

### DIFF
--- a/scripts/notes.ts
+++ b/scripts/notes.ts
@@ -1,8 +1,8 @@
 import { execSync } from 'node:child_process';
-import chalk from 'chalk';
-import themes from '../themes.json';
+import * as themes from '../themes.json';
 import { writeFileSync } from 'node:fs';
 
+const chalk = require('chalk');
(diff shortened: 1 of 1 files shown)


scripts/notes.ts
```diff
@@ -150,12 +150,14 @@ const organizeNotes = (data: PRDetails[]) => {
   // initial sort
   notes.sort();
 
-  // move cmsgov to end
-  notes.forEach((note: PRNote) => {
-    if (note[0] === 'cmsgov') {
-      notes.push(note);
-      notes.shift();
-    }
+  // move cmsgov to end of list
```
pwolfert, line 153: Regarding lodash, I don't care if we use lodash for tooling. It's not like it'd be adding more dependencies to the component library

pwolfert (approved):
Thanks!

PR #2443, opened 2023-03-27
[NO-TICKET] Fix SSR support for new privacy settings modules

pwolfert:
## Summary

Make sure the new privacy settings modules are compatible with SSR by avoiding accessing the `document` if it isn't defined. Found this while trying out the Astro static site generator


forrestbaer (approved):
Seems reasonable!

PR #2449, opened 2023-04-05
[WNMGDS-2112] Add accent colors to themes

line47:
## Summary

- Added tokens for --color-accent-primary for all themes
- Updated doc site page to add color swatches for  --color-accent-primary
- Added styles for using --color-accent-primary for background colors
- **Did not** add accent color for text-color utilities at this time b/c I don't think it would be needed 

## How to test

1. Run doc site with `yarn start` and visit the theme file page to see accent color content on the page
2. Then visit the background color page to see the bg color examples 

## Checklist

### DIFF
--- a/.storybook/static/cmsgov-theme.css
+++ b/.storybook/static/cmsgov-theme.css
@@ -53,6 +53,13 @@
 --color-secondary-dark: #3bcb7a;
 --color-secondary-darker: #2e9e5f;
 --color-secondary-darkest: #217144;
+--color-accent-primary-lightest: #fffce6;
+--color-accent-primary-lighter: #fff7b5;
+--color-accent-primary-light: #ffec52;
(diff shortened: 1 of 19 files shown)


pwolfert (approved):
LGTM, but one question: Why is it called `accent-primary` instead of just `accent`?

forrestbaer (approved):

line47:
@pwolfert I was envisioning a future state where more than one accent color exists - open to other ways to handle this?

pwolfert:
@line47, I would just hate to make it longer than needed if we don't plan on adding more, but it'd also be a breaking change to change it after we've released it. If there are others, what if the primary one was assumed primary so it's like `color-accent`, `color-accent-secondary`, etc? It's your call.

line47:
I'd like to stick with the more explicit and future proof `color-accent-primary`

PR #2454, opened 2023-04-07
[WNMGDS-2082] Steplist aria-label Primary/Secondary sections changed to use aria-describedby

forrestbaer:
## Summary

- Several items in Steplist had `aria-label`s which have been changed to `aria-describedby` in order to utilize i18n translation on them.
- There are two sections in a steplist item with actions. The Secondary actions (which come before) and the Primary action, which follows the the secondary actions and can only be a link or a **Completed** message. The Secondary actions wrapping div had previously had an `aria-label` describing that section as being Secondary which was changed to an `aria-describedby` utilizing a screenreader-hidden `span` for the secondary message. The primary section had the same but since there is only one primary action possible (a link), that text was simply moved into the links screenreader-hidden text.
- 'Primary actions' changed to 'Primary action' as, like the highlander, there can only be one. ⚔️ 

## How to test

1. Current live example [here](https://design.cms.gov/storybook/iframe.html?args=&id=patterns-step-list--step-list-example&viewMode=story).
2. `yarn clean && yarn install && yarn build && yarn storybook`, then go [here](http://localhost:6006/iframe.html?args=&id=patterns-step-list--step-list-example&viewMode=story#step-2).

## Checklist


forrestbaer:
@SparrowSpit Do you have time to evaluate this change before our sprint ends EOD April 19th?

SparrowSpit:
I should be able to look at it tomorrow morning. @forrestbaer

SparrowSpit:
@forrestbaer  I still haven't been successful in getting a local version of the DS up and running. Is it possible that you'd either provide the output html for me to test or put this in an env. I can review?

forrestbaer:
> I still haven't been successful in getting a local version of the DS up and running. Is it possible that you'd either provide the output html for me to test or put this in an env. I can review?

Oh shoot, ok well I can see about putting a demo site up, but sometimes those don't actually come up for a day or .. never.. I can help you get the DS running locally, it should be painless if you're on a mac. A few steps if you want to try:

If you don't have brew installed
```
/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```

Then
```
1. brew install bash
2. brew install git
3. curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.39.3/install.sh | bash
4. git clone https://github.com/CMSgov/design-system.git
5. cd design-system
6. nvm use
7. npm install --global yarn
```

Then to test, from that same folder:
```
1. git checkout WNMGDS-2082-steplist-aria-labeledby
2. yarn clean && yarn install && yarn build
3. yarn storybook
```

forrestbaer:
Managed to get a demo up: https://cmsgov.github.io/design-system/branch/WNMGDS-2082-steplist-aria-labeledby/storybook/iframe.html?globals=theme:core&id=patterns-step-list--step-list-example&viewMode=story

SparrowSpit:
A few things while testing VO + Safari:

- the "Primary" and "Secondary" text itself doesn't make sense to me in this example. Or maybe that wording doesn't really make sense in general for this component? What's being called out as "Secondary" doesn't seem so and would actually seem to be questions that are part of, or contained within, that step's section. So, if this concern is relevant, I'm worried that those terms could be used by teams and be misleading (for teams and end users) given the UX of the component. 
- the steps that aren't visible (beyond the section/step heading) have a span with "ds-u-visibility--screen-reader" that is being read out when nothing is visible. Can we hide that element if that step's description and list aren't being shown visibly?
- The aria-describedby text (within the list item) isn't being read (with VoiceOver) in conjunction with the step heading. If I change it to aria-labelledby then it gets read. Either way, I'm not sure it's crucial to connect this paragraph text with the heading in this way. It seems to me that if someone is navigating line by line (with assistive tech) then they'll encounter both elements together/in succession. So, I'd recommend removing the aria attribute on this step description element. 

Maybe these are not relevant to this work being done and should be another ticket but noting either way.

forrestbaer:
Thanks @SparrowSpit 

The reason for switching to `aria-describedby` was that @zarahzachz had mentioned we shouldn't be using `aria-labeledby` on div elements. 

The **Primary** and **Secondary** items do not make sense I agree, the idea I believe is that the **Primary** action is the one that gets you to the next step, the **Secondary** items are items within a step that can also be completed.  I would prefer to just remove those designations all together.

Good point about the screen reader items that are being read when not visible, will see about removing those.

@pwolfert @zarahzachz any additional thoughts, guidance?

SparrowSpit:
Oh ok. I thought Sarah'd advocated getting rid of 'aria-label' not 'aria-labelledby'.

forrestbaer:
> Oh ok. I thought Sarah'd advocated getting rid of 'aria-label' not 'aria-labelledby'.

Both! But I may have misunderstood her on the labeledby part.

pwolfert:
1. I think we're all in agreement then on removing the "primary" and "secondary" language.
2. Based on what Dan said when he said, "I'd recommend removing the aria attribute on this step description element," we should just remove that `aria-describedby={descriptionHeadingID}`
3. If I'm understanding the discussion about the discussion about the "ds-u-visibility--screen-reader", it sounds like we don't need that at all anymore if we get rid of the "primary" and "secondary" language. In fact, we'd be able to get rid of the `actionsLabelText` and `substepsLabelText` props altogether. I did a search among our applications for use of those props, and nobody is overriding them.

pwolfert:
> we'd be able to get rid of the `actionsLabelText` and `substepsLabelText`

Or maybe we're just getting rid of `substepsLabelText`, since we're now applying `actionsLabelText` inside the resume step link.

@SparrowSpit, in this branch, the Resume button will read out "Resume primary actions for enter household details". Would you prefer it to just say "Resume enter household details"? That is what it says on the main branch except that the div around it has that "Primary actions for Enter household details" aria label

forrestbaer:
> Or maybe we're just getting rid of `substepsLabelText`, since we're now applying `actionsLabelText` inside the resume step link.

That text is actually just a formatted version of `step.heading` so we can actually strip out a lot of unnecessary code here. I've just pushed updates to this and will provide a demo url shortly.

forrestbaer:
The link above should work now if you open it in an incognito window or clear your cache, I've tested it and the text now reads `Resume Enter household Details` and the aria-decribedby sections and extraneous hidden div are gone.

pwolfert (approved):
Codewise LGTM 👍

PR #2455, opened 2023-04-07
[RELEASE] 04/07 Version bump main

pwolfert:


pwolfert:
This was a test!

versions.json
```diff
@@ -0,0 +1,123 @@
+{
+  "design-system": [
+    "6.0.2",
```
pwolfert, line 3: Note that it did what I expected here. After I bumped everything to `x.0.2`, it added them to the tops of these lists

PR #2456, opened 2023-04-10
[WNMGDS-1828] Add version switcher to doc site

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-1828

- Created a version-controlled running list of versions of our packages to be ingested by the doc site in the version switcher
- Added a version switcher dropdown to the doc site's navigation panel

## How to test

https://cmsgov.github.io/design-system/branch/pwolfert/version-switcher

## Checklist


packages/docs/src/components/layout/VersionSwitcher.tsx
```diff
@@ -0,0 +1,61 @@
...
+import { Dropdown } from '@cmsgov/design-system';
+
+function getVersionOptions(versions: string[]) {
+  return versions.map((version) => ({
+    label: version,
+    value: version,
+  }));
+}
+
+function getPackageData(theme: string) {
```
forrestbaer, line 17: Wish we could do this without hard coding our packages and using themes.json instead.  Maybe with an additional field in themes.json for the package.json path, then a dynamic import of the json for that theme?

versions.json
```diff
@@ -0,0 +1,119 @@
+{
```
forrestbaer, line 1: Not a bad idea, I guess we can update the current scripts to utilize this going forward.

forrestbaer (commented):
I like how it looks! Trying this out theoretically with the main design.cms.gov site, many of the url's won't work as they don't exist remotely. i.e. 

- https://design.cms.gov/v/4.0.3
- https://design.cms.gov/v/3.5.4

And others.

packages/docs/src/components/layout/VersionSwitcher.tsx
pwolfert: When I was writing this, I was thinking that I'd be able to do it with Astro if I could keep the part that feeds this component completely server-side. It's not so straightforward to do this with Gatsby. In fact, I'm not sure how to do it without adding it to each page's _context_ through `packages/docs/gatsby-node.js` or creating a plugin that makes that data available through GraphQL 😵

pwolfert: If you have ideas, I'd love to remove the hardcoding!

forrestbaer: Oof yeah I was hoping it would be as easy as I mentioned, that's too bad. I mean we have these kinds of references in several places, probably just best to do a full CMS integration and update the code around the codebase where needed when that happens. It's not like we add child systems too often.

pwolfert:
@forrestbaer, the branch demo has been updated. Now if you are on healthcare and select version `8.0.0`, it should navigate to `/v/4.0.0` (core version)

pwolfert:
Oh also, this was generated by the release script changes in this branch: https://github.com/CMSgov/design-system/pull/2455

packages/docs/src/components/layout/VersionSwitcher.tsx
```diff
@@ -0,0 +1,63 @@
...
+      return medicarePackage;
+  }
+  return { version: '' };
+}
+
+/**
+ * Theme Switcher
+ * The dropdown component to switch the theme of the documentation site
+ */
+const ThemeSwitcher = () => {
```
zarahzachz, line 33: We have another ThemeSwitcher component. Should this also be called ThemeSwitcher? That and the above comment make me a little confused on what the difference between the components is.

pwolfert: Doh! Good catch

pwolfert: Copypasta 🍝

PR #2459, opened 2023-04-12
[WNMGDS-2212] Many storybook stories updated for easier control via args.

forrestbaer:
## Summary

- Updated a number of stories to use the `text` control format so they can be more easily modified in storybook
- Changed a few `select` types to `radio` to be consistent
- Changed a few `radio` types to `boolean` to be consistent
- Looks like one snapshot update which I don't see any changes in, not sure how that happened but it seems OK.

## How to test

1. `yarn clean && yarn install && yarn build && yarn storybook`
2. evaluate changed stories if you like!

## Checklist


packages/design-system/src/components/DateField/MultiInputDateField.stories.jsx
```diff
@@ -7,11 +7,17 @@ export default {
...
   argTypes: {
-    label: { control: false },
     errorMessage: {
       control: { type: 'text' },
       type: { name: 'string', required: true },
     },
+    label: { control: 'text' },
+    monthDefaultValue: { control: { type: 'number', min: 1, max: 12, steps: 1 } },
+    dayDefaultValue: { control: { type: 'number', min: 1, max: 31, steps: 1 } },
+    yearDefaultValue: { control: 'number' },
```
pwolfert, line 17: I think in order for setting these to do anything, we need to update the story itself to no longer `useState` but `useArgs`. I've had mixed success with `useArgs`.

packages/design-system/src/components/Spinner/Spinner.stories.jsx
```diff
@@ -9,14 +9,8 @@ export default {
...
-    inversed: {
-      control: 'radio',
-      options: [true, false],
-    },
-    filled: {
-      control: 'radio',
-      options: [true, false],
-    },
+    inversed: { control: 'boolean' },
+    filled: { control: 'boolean' },
```
pwolfert, line 13: 👍

pwolfert (approved):
Thanks for cleaning these up. There is a set of controls that aren't functional right now noted below, though.

The visual diff is a really subtle letter-spacing change (Saturday label), and I don't know what caused it. I'm guessing it happened a while ago but the pixel diff was so small that it never changed.

letter-spacing-difference

packages/design-system/src/components/DateField/MultiInputDateField.stories.jsx
forrestbaer: Oh interesting I didn't realize that didn't update in that case.. I'm just going to remove these then, I don't think they add a lot of value, mostly just wanted to play with the min/max value stepper.

PR #2461, opened 2023-04-13
[NO-TICKET] Update css reset branch with main

pwolfert:
## Summary

Merges `main` and updates most of the snapshots. The VRTs might have found some actual visual regressions.


pwolfert:
The visual regression tests on Jenkins didn't catch some visual regressions that I see in the visual diff when I run them locally, and I think it's because the bad reference images have already been accepted. I'll list them out so we can work through them. I'm thinking about handling that in a different pull request though.

PR #2463, opened 2023-04-14
[WNMGDS-2232] Address visual regressions

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2232

- Fixes visual regressions found by looking through the visual diffs
- Based on the number of issues here stemming from list reset styles, perhaps we should consider changing or removing the list reset styles. Thoughts, @zarahzachz? Would moving those styles under `.ds-c-list--bare` or something like that result in a bunch of code needing to be added back into other component styles?

## How to test

Look through the commit messages and run through the affected stories

## Checklist

Purposely left out the milestone and labels so it wouldn't show up in the release notes


zarahzachz:
It shouldn't, but I'd need to see it to be 100% sure. The way I addressed these styles in the CSS was to target `ul/ol[role="list"]` and anything using that selector would probably also have the `ds-c-list--bare` class as well.

forrestbaer:
Noticed a few differences, not sure how important, just ones that caught my eye.

- Accordion (tiny differences in spacing)
- MultiInputDateField (looks better in new version, centered correctly)
- FilterChip (slightly less spacing)
- Pagination (slightly more spacing between items, looks better to me)
- UsaBanner (slightly less spacing between letters of 'how you know')

I don't know how important these are!

pwolfert:
I've been thinking about the effects of changing all `ol` and `ul` elements to use `display: flex`. There might be more instances of lists used for non-standard things than we may have previously thought. It could be disruptive to apply styles to all of them, but I think that's just something we'll have to find out when testing in the beta. I don't think we should turn back too early. I've been debating that in my head for the past hour though and [made a branch here](https://github.com/CMSgov/design-system/compare/pwolfert/address-visual-regressions...pwolfert/list-styles-relocation?expand=1).

pwolfert:
@forrestbaer, regarding MultiInputDateField, I think the opposite might be true, and you caught a regression.

Main:

main

This branch:
branch

Looks like the problem is that we weren't applying the box-sizing rule before and now we are.

pwolfert:
I tracked down the `Accordion` spacing differences, and I found out that the line-height used to come from user agent styles. In Chrome, the button element that makes up each accordion header has a `line-height: normal`. In the reset branch, the line height comes from the button reset. While applying that declaration to our accordion buttons would get us back to the way it was before, the new way of relying on the reset is better.

pwolfert:
Same thing is happening with `FilterChip`. It's a button that now inherits its line height instead of having `line-height: normal` from user agent styles. That inherited value is `var(--font-line-height-base)`.

pwolfert:
Pagination's spacing difference is due to the box-sizing reset. I agree, @forrestbaer, that it looks better.

pwolfert:
And finally, the `UsaBanner` link difference comes from that link's font family being `Arial` before 😆

tests/browser/snapshots/components-multiinputdatefield--inverted-multi-input-date-field-healthcare--Mobile-Chrome.png
zarahzachz, line 1: Is the error text color correct here? It looks darker than the original.

zarahzachz (approved):
Approving but left 2 comments.

zarahzachz:
> And finally, the `UsaBanner` link difference comes from that link's font family being `Arial` before 😆

I've had a question about `UsaBanner` for awhile - how similar should it look across gov't sites? 

Is the point of it to have a consistent look across gov't sites regardless of theme so users can easily see it and go, "Yeah, this looks safe!" or are they open to modification? If so, how extensive can that modification be? 

Our banners seem to use the font families for each theme but otherwise look the same, but [when you visit USWDS](https://designsystem.digital.gov/components/banner/) it has a black background and different padding (some of the time).

pwolfert:
@zarahzachz, according to the USWDS guidance, they say you should choose a background color that fits the theme, but I do feel like most gov't sites use the standard gray. I think we can open it up for customization when we get a request to do so. Right now our sites use white backgrounds for the headers, so the grey fits.

> Avoid distraction. The banner should appear on every page of your site. Choose background colors that fit with your site theme, and avoid color combinations that draw excessive attention to the banner.

tests/browser/snapshots/components-multiinputdatefield--inverted-multi-input-date-field-healthcare--Mobile-Chrome.png
pwolfert: Looks like this color change is coming from `main` and the color changes we did there. We'll need to address it there

PR #2464, opened 2023-04-17
[NO-TICKET] Implemented stricter tsc checking with noUnusedLocals & noUnusedParameters

forrestbaer:
## Summary

- Updated `yarn type-check` with more flags: `--alwaysStrict --diagnostics --noUnusedLocals --noUnusedParameters`
- After updating these, found several places where we have unused locals and unused parameters which I then removed.
- Validated these changes and found no difference in effect between local storybook & design.cms.gov/storybook

## How to test

1. `yarn clean && yarn install && yarn build && yarn type-check && yarn test`

## Checklist


forrestbaer:
I don't know if we even need this check so much anymore, in any event I think this might be better handled via the linter if it threw errors instead of just warnings in some cases, will circle back on it.

PR #2468, opened 2023-04-20
[WNMGDS-2290] Upgrade `downshift` and convert `Autocomplete` to functional component

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2290

This is the 1st of 2 pull requests. The `downshift` team has kept its older API around, so I did the upgrade without really changing our `downshift` implementation, and then I converted it to a functional component in https://github.com/CMSgov/design-system/commit/befe75162f67dcc26f29ceec78da208d58d6d82c to prepare for the next step of using their new Hooks API.

## How to test

Test out the Autocomplete. Should be no bugs or functional differences.

## Checklist


packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -161,42 +166,42 @@ function isTextField(child: React.ReactElement): boolean {
...
+    ariaClearLabel,
+    autoCompleteLabel,
+    autoFocus,
+    children,
+    className,
+    clearInputOnBlur,
+    clearInputText,
+    clearSearchButton,
+    focusTrigger,
+    id: _id,
```
forrestbaer, line 186: I'm a bit confused by how this destructuring assignment works. `_id` isn't used anywhere else and I don't see it defined. Is this some react magic that I'm not grasping? The only mention of using underscore prefixes was in a `don't do` document relating to using underscores to faux mock private methods as a design choice (https://github.com/airbnb/javascript/tree/master/react#methods). 

The outcome of this is that it seems like you're storing a copy of id in `_id` which isn't used anywhere else., same for `_labelId`, curious how this works.

forrestbaer (approved):
This looks good, analytics working exactly the same, no functional differences that I can spot, just one question regarding destructuring of a few items.

forrestbaer:
Also is this the correct milestone, this is going with 7? I've set up another milestone for 8.0 if that's where it should go instead?

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
pwolfert: The reason I'm including it in the destructuring assignment is to pull it out of the remainder (`...autocompleteProps`), and I'm underscoring it to signal that we don't use it and so it doesn't conflict with the `id` and `labelId` defined at the top of this function.

PR #2471, opened 2023-04-21
[NO-TICKET] Force theme in theme-specific stories

pwolfert:
## Summary

Force stories for healthcare and medicare to be themed in their theme, regardless of what the global storybook setting is. This is just a quality of life improvement.


pwolfert:
It might be confusing, but I wasn't able to reliably change what's shown in the Storybook settings based on this value. Right now it's only a read operation (derriving a value and passing it down to the story as it renders), but if I were to change it in the settings, it'd also be a write operation, which then requires a rerender of the story.

PR #2475, opened 2023-04-24
[WNMGDS-2050] Pagination page list fix

zarahzachz:
## Summary

WNMGDS-2050

Pagination's list of pages gives incorrect context to screen reader users. E.g., if a page set has 15 pages the readout will still say listitem X of 7 because there are 7 &lt;li> tags on the screen.

To fix this, the solution was to remove the list markup in favor of something more generic that won't confuse SR users.

To test, visit [demo url](https://cmsgov.github.io/design-system/branch/WNMGDS-2050/pagination-page-list-fix/components/pagination/?theme=core) and confirm no list markup exists. Test with a screen reader to confirm the readout is as expected (no confusion around number of items in a list vs actual number of pages).

### DIFF
--- a/packages/design-system/src/components/Pagination/Pagination.test.tsx
+++ b/packages/design-system/src/components/Pagination/Pagination.test.tsx
@@ -31,8 +31,7 @@ function queryNextLink() {
 }
 
 function hasEllipsis() {
-  const items = screen.getAllByRole('listitem');
-  return items.some((item) => item.querySelector('.ds-c-pagination__overflow'));
+  return document.querySelector('.ds-c-pagination__overflow');
 }
 
 describe('Pagination', () => {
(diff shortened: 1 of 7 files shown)


packages/design-system/src/components/Pagination/Pagination.test.tsx
```diff
@@ -31,8 +31,7 @@ function queryNextLink() {
 }
 
 function hasEllipsis() {
-  const items = screen.getAllByRole('listitem');
-  return items.some((item) => item.querySelector('.ds-c-pagination__overflow'));
+  return document.querySelector('.ds-c-pagination__overflow');
```
zarahzachz, line 34: I know testing library prefers you use their API to query component structure, but the changes I made removed the list role items in favor of generic role items and there's no easy way to query that part of the markup without adding something like data-testId attributes everywhere.

pwolfert (approved):
The readout is less confusing 👍 

This is probably an unrelated bug, but I'm finding that on VoiceOver if I activate the "Next" button it will continue to read out "Pagination results: Page 2 of 15" for every page as I cycle through them.

PR #2477, opened 2023-04-24
[NO-TICKET] Fixing notes generator with proper limits and organization

forrestbaer:
## Summary

- Impacts and Types properly organized now
- Limit of 30 items was a problem, set it to 200
- Added underscore to dev dependencies

## How to test

1. [See generated notes from 7.0 milestone](https://github.com/CMSgov/design-system/releases/tag/untagged-ebabf927aa34b1460e06)

## Checklist


package.json
```diff
@@ -147,6 +148,7 @@
     "stylelint-scss": "^4.2.0",
     "ts-node": "^10.8.2",
     "typescript": "^4.9.5",
+    "underscore": "^1.13.6",
```
pwolfert, line 151: In our doc site package, we're actually using `lodash` instead of `underscore`. Should we use the same package?

scripts/notes.ts
```diff
@@ -92,8 +98,9 @@ const latestCoreTag = getLatestCoreTag();
  */
 const getPRs = () => {
   const prData = JSON.parse(
+    // limit to 200 results so we get more than the 30 default
```
pwolfert, line 101: Good comment 👍

pwolfert (approved):
LGTM 👍

package.json
forrestbaer: Oh good call! The script actually works exactly the same using lodash with no changes 🥳 So I'm going to remove the underscore dependency.

PR #2480, opened 2023-04-25
[WIP] [WNMGDS-2275] Tokenize mgov form elements

zarahzachz:
## Summary

WNMGDS-2275

[Demo url for review](https://cmsgov.github.io/design-system/branch/WNMGDS-2275/tokenize-mgov-form-elements) (Medicare ChoiceList and Form elements should look relatively unchanged)

The goal of this ticket was to take the SCSS overrides for ChoiceList and Forms in Medicare and either remove unnecessary styles or tokenize values when appropriate. All of ChoiceList was able to be removed from Medicare, but there are some overrides remaining in the Forms override. 

The remaining styles in Forms proved to be more difficult to tokenize as Medicare has different rules applied to different form element labels (e.g., it's difficult to theme forms when checkboxes and radios have different labels than text field labels). Medicare's treatment of form styles should probably be evaluated to be more streamlined and would be a breaking change.

This PR also introduces a new coding pattern for how to address one-off token creation: creating tokens for an individual theme and relying on CSS vars' default values for the other themes (`var(--example-token, default)`). These tokens have added comments explaining their uniqueness. There are corresponding comments in the CSS files where these tokens are used to explain what's happening.


packages/design-system-tokens/src/themes/medicare-component.ts
```diff
@@ -412,18 +421,21 @@ export const components: AnyTokenValues = {
...
     '__border-width':                             '2px',
+    '__border-width--disabled':                   '1px',
     '__border-width--error':                      '3px',
     '__border-width--success':                    '3px',
     '__border-color':                             t.color['gray'],
     '__border-color--disabled':                   t.color['gray-darker'],
     '__border-color--error':                      t.color['error'],
     '__border-color--error--inverse':             t.color['error-lighter'],
-    '__border-color--inverse':                    t.color['black'],
+    '__border-color--inverse':                    t.color['gray'],
```
zarahzachz, line 431: I didn't mean to change this, but I kept getting visual regressions on datepicker inverse without setting this. Since prod doesn't show a black border for mgov text-input I figured this was a fix - thoughts?

packages/ds-medicare-gov/src/styles/components/_ChoiceList.scss
```diff
@@ -1,24 +0,0 @@
...
-  gap: 1rem;
-
-  &:has(.ds-c-choice--small) {
-    gap: 0.75rem;
-  }
-}
-
-.ds-c-choice:not(.ds-c-choice--small) + label,
-.ds-c-choice:not(.ds-c-choice--small) + .ds-c-label {
-  transform: translateY(0.125em);
```
zarahzachz, line 11: The translateY token in `medicare.ts`, line 274 come from this section.

packages/ds-medicare-gov/src/styles/components/_ChoiceList.scss
```diff
@@ -1,24 +0,0 @@
...
-}
-
-.ds-c-choice[type='checkbox'] {
-  &:checked {
-    &::before {
-      border-radius: var(--choice__border-radius);
```
zarahzachz, line 17: This was fixed using the `inherit` value - meaning the border-radius of this pseudo selector will inherit the border-radius defined on its parent - a checked checkbox in this case.

packages/ds-medicare-gov/src/styles/components/_ChoiceList.scss
```diff
@@ -1,24 +0,0 @@
...
-    }
-  }
-}
-
-.ds-c-choice:not(:disabled):hover {
-  border-color: var(--choice__background-color--checked);
```
zarahzachz, line 23: The border-color token in `medicare.ts`, line 260 comes from this section.

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -1,23 +1,9 @@
 @use '../layout' as *;
 
 .ds-c-field {
-  border-color: var(--text-input__border-color);
-
-  color: var(--text-input__color);
-  font-family: var(--font-family-rubik);
-  font-size: var(--font-size-md);
-  padding: $spacer-2;
```
zarahzachz, line 9: The update to the `text-input__padding` token came from this override - `$spacer-2` equates to `t.spacer['2']`

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -1,23 +1,9 @@
 @use '../layout' as *;
 
 .ds-c-field {
-  border-color: var(--text-input__border-color);
```
zarahzachz, line 4: Many of these overrides were redundant and removed.

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -1,23 +1,9 @@
...
-
-  color: var(--text-input__color);
-  font-family: var(--font-family-rubik);
-  font-size: var(--font-size-md);
-  padding: $spacer-2;
-  &:disabled {
-    border: solid 1px var(--text-input__border-color--disabled);
```
zarahzachz, line 11: The disabled border-width token in `medicare.ts`, line 430 comes from this section.

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -1,23 +1,9 @@
...
-  }
-
   &::placeholder {
     color: var(--color-gray-dark);
   }
-
-  &--error {
-    border-color: var(--text-input__border-color--error);
```
zarahzachz, line 19: Redundant

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -28,9 +14,6 @@ select.ds-c-field {
 }
 
 .ds-c-label {
-  font-family: var(--font-family-rubik);
-  font-size: var(--font-size-md);
```
zarahzachz, line 32: Redundant

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -68,26 +51,4 @@ legend[id^='datefield_label'] {
       text-transform: none;
     }
   }
-
-  > .ds-c-field__hint:not(.ds-u-color--error) {
-    color: var(--form-label__color--not-error);
```
zarahzachz, line 73: Redundant

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -68,26 +51,4 @@ legend[id^='datefield_label'] {
...
-  }
-}
-
-.ds-c-field__hint {
-  font-family: var(--font-family-rubik);
-  font-size: var(--font-size-sm);
```
zarahzachz, line 79: The small font-size for hints token in `medicare.ts`, line 340 comes from this

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -68,26 +51,4 @@ legend[id^='datefield_label'] {
...
-}
-
-.ds-c-datefield__separator {
-  display: none;
```
zarahzachz, line 83: The datefield tokens in `medicare.ts`, lines 282-285 comes from this

packages/ds-medicare-gov/src/styles/components/_Forms.scss
```diff
@@ -68,26 +51,4 @@ legend[id^='datefield_label'] {
...
-}
-
-// Alignment of $ icon for the currency field
-.ds-c-field--currency {
-  padding-left: $spacer-3;
```
zarahzachz, line 88: The currency tokens in `medicare.ts`, lines 443-445 come from these styles. The spacer value had to be adjusted from the override so the "$" looked more correct with the masked field.

zarahzachz:
Requesting reviews from @line47 and/or @brentschneider due to some adjustments made to Medicare's tokens - namely the form padding and the inverse border color.

packages/ds-medicare-gov/src/styles/components/_Forms.scss
pwolfert: I agree. @line47 or @brentschneider, please [take a look at the VRT reference image change](https://github.com/CMSgov/design-system/pull/2480/files#diff-b6ce95ded0ccb2b7c3464d1c64b3713a440a9b19913776ba5a2e49fca0d78a21) to see how it changed

packages/design-system-tokens/src/themes/medicare-component.ts
pwolfert: This sounds like a good fix. I looked at the inverse dropdown on main and this branch, and they're the same gray color. Really I don't think the gray gives very good contrast on a dark background, but I don't think it should be changed right now.

pwolfert (approved):
Looked through the VRT changes and didn't see any noticeable changes besides the currency thing. Nice work!

brentschneider (approved):
@zarahzachz   
I saw the convo in #cms-design-system-team slack.  
The focus style migration looks good to me 👍🏼

PR #2481, opened 2023-04-27
[WNMGDS-1902] Refactor reviewlink

zarahzachz:
## Summary

WNMGDS-1902

[Demo link](https://cmsgov.github.io/design-system/branch/WNMGDS-1902/refactor-reviewlink/storybook/?path=/story/components-review--multiple-reviews&globals=theme:core)

This work removes the need for the ReviewLink component (only used inside Review). I've combined the logic from ReviewLink into Review and added unit tests to cover what ReviewLink did individually.

Additionally, there Review has moved out of the Patterns and into Components and this resulted in VRT updates. Running VRTs locally prior to the location change resulted in no changes from Review. All VRTs being flagged are just from that file change.


packages/design-system/src/components/Review/Review.stories.jsx
```diff
@@ -2,7 +2,7 @@ import React from 'react';
 import Review from './Review';
 
 export default {
-  title: 'Patterns/Review',
+  title: 'Components/Review',
```
zarahzachz, line 5: This line's the culprit in all those renamed VRT files

packages/design-system/src/components/Review/Review.test.tsx
```diff
@@ -78,4 +79,18 @@ describe('Review', function () {
 
     expect(wrapper.classList).toContain('my-class');
   });
+
+  it('calls onClick function when clicked', () => {
```
zarahzachz, line 83: This was the only unique test ReviewLink.test.tsx had, so I ported it over here.

packages/design-system/src/components/Review/Review.tsx
```diff
@@ -1,9 +1,8 @@
 import React from 'react';
-import ReviewLink from './ReviewLink';
 import classNames from 'classnames';
 import { t } from '../i18n';
 
-export type ReviewHeadingLevel = '1' | '2' | '3' | '4' | '5';
+export type ReviewHeadingLevel = '1' | '2' | '3' | '4' | '5' | '6';
```
zarahzachz, line 5: I added an option for the Review to support `<h6>` tags. Our Stories reflect this as an option and it seemed odd to exclude it as an optional level. I don't think this breaks the API as it's just adding another option instead of taking anything away.

packages/design-system/src/components/Review/Review.tsx
```diff
@@ -12,7 +11,7 @@ export interface ReviewProps {
   children: React.ReactNode;
   className?: string;
   /**
-   * Optional label to give screenreaders longer, more descriptive text to
+   * Optional label to give screen readers longer, more descriptive text to
    * explain the context of an edit link.
    */
   editAriaLabel?: string;
```
zarahzachz, line 17: I didn't change this prop as it might break the API and I don't wanna open a can of worms for this small ticket, but do have questions about how this component might pass a11y testing.

I'm thinking, if there are several Review components in a row (as we have in [one example](https://design.cms.gov/storybook/?path=/story/patterns-review--multiple-reviews&globals)) and all Reviews have the default "Edit" link, **do we run into the a11y issue of having multiple links with the same label?**

Should we make an aria-label required? Or provide a better fallback? I was initially thinking a good fallback could be setting an `aria-label=${editText} ${heading}`, but `heading` is also optional. Anyways, this is a longwinded way to say we might need a follow-up ticket to address this issue (if it is one).

cc: @SparrowSpit

SparrowSpit: In a real use-case scenario this is used at the end of the healthcare application, right? If so, I'm pretty sure what we have going on there is `aria-labelledy={heading}` where it points to the heading's ID but I'd have to triple check to be 100% on that. 

That accomplishes what you're getting at where we're "connecting" the two elements and providing contextual info for assistive tech.

packages/design-system/src/components/Review/__snapshots__/Review.test.tsx.snap
```diff
@@ -18,12 +18,10 @@ exports[`Review renders review 1`] = `
       review text
     </div>
   </div>
-  <div>
```
pwolfert, line 21: I would kind of expect the removal of this div to have an effect on how it renders in the flex parent because `a` isn't a block element, but it doesn't seem to change anything 🤷. One less div!

pwolfert (approved):
LGTM 👍

PR #2483, opened 2023-04-28
[WNMGDS-2290] Switch from Downshift's old API to new and no longer clear Autocomplete input on blur

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2290

- Switch from using Downshift's old render-function API to their hooks API in the `Autocomplete` component
- Some of the default behaviors for Downshift have changed:
    - When a user has typed into the text field but hasn't selected an option, their input will no longer be cleared when the component loses focus
    - When a user presses <kbd>ESC</kbd> before selecting an option, their input will no longer be cleared
    - Related to this, we've removed the `clearInputOnBlur` prop, because we no longer support that unhelpful behavior. (We used to add custom code that disables that behavior when a user passed `clearInputOnBlur={false}`.)
- To satisfy the new API, the menu always stays in the DOM but is visually hidden with `display: none` when it's not open
- Note that the `onChange` is only called when a selection is made or cleared, and that behavior has not changed from what it was previously. **If a team is using this as an uncontrolled component, however, the fact that raw user input can exist in the text field could lead to false-positives in validation, so this is a breaking change.** I do not know of any team using this as an uncontrolled component. All examples that I have found at CMS use the `onChange` function to handle selection changes.
- Because of the change in the underlying API, the pseudo-event objects passed to the `onChange` and `onInputValueChange` functions have changed slightly.
    - Before: Was defined by [ControllerStateAndHelpers](https://github.com/downshift-js/downshift/blob/master/typings/index.d.ts#L203C1-L205)
      ```ts
      onChange?: (
        selectedItem: AutocompleteItem,
        stateAndHelpers: ControllerStateAndHelpers<any>
      ) => void;
      onInputValueChange?: (
        inputValue: string,
        stateAndHelpers: ControllerStateAndHelpers<any>
      ) => void;
      ```
    - After: Is defined by [UseComboboxStateChangeOptions](https://github.com/downshift-js/downshift/blob/master/typings/index.d.ts#L463C1-L466)
      ```ts
      onChange?: (
        selectedItem: AutocompleteItem,
        stateAndHelpers: UseComboboxStateChangeOptions<any>
      ) => void;
      onInputValueChange?: (
        inputValue: string,
        stateAndHelpers: UseComboboxStateChangeOptions<any>
      ) => void;
      ```

## How to test

Try out the Autocomplete stories

## Checklist


dmethvin-gov:
FWIW the Material UI component seems to have the same clear-it-out behavior:
https://mui.com/material-ui/react-autocomplete/

From a correctness standpoint I feel like the component should _never_ call `onChange` with a result that isn't in the expected set, because the parent shouldn't see the incomplete and incorrect text. The component could save any unmatched user text in a local `useState` and only fire `onChange` when there is one possible match and/or the user makes a selection from the filtered list. On blur that text can be hidden and on (re)focus the text can be shown. As long as the component is mounted the `useState` value is preserved so an accidental tab or click elsewhere on the page still shows the value. So the user doesn't lose what they typed, if that was the concern.

pwolfert:
That's a really good point, @dmethvin-gov, about when `onChange` should be called. [The production version of our Autocomplete component](https://design.cms.gov/storybook/?path=/story/components-autocomplete--default&globals=theme:core) only calls `onChange` when a selection is made, and the first parameter is the _item_. For the sake of consistency, we should try to maintain that pattern going forward, so I've got updates to do here.

If we were to allow the input to retain text that the user has typed but still maintain the same `onChange` behavior, then it would still be a breaking change for apps that use this as an uncontrolled component, but I don't currently know of any teams that are doing that, and it may be sufficient to warn those hypothetical application teams of the change in the breaking-change notes. Does that sound like a good compromise? I could still end up going back and re-creating the `clearInputOnBlur` functionality if some teams want that old behavior.

pwolfert:
I've confirmed that in this latest version `onChange` is only called when a selection is made or when the selection is cleared. [Here's a demo](https://cmsgov.github.io/design-system/branch/pwolfert/downshift-hooks/storybook/?path=/story/components-autocomplete--default&globals=theme:core).

forrestbaer (approved):
Wow the whole a11y status message thing is.. interesting, good that they provide that ability but a shame that it needs to be included. All looks good in my run through, I can't find any differences, nice 👍🏻

PR #2486, opened 2023-05-01
[NO-TICKET] Clean out an unused form token after color changes made it obsolete

pwolfert:
## Summary

Cleans out an unused form token. After our color changes a while back, this variable became unnecessary and unused. 

Group this with other color changes in the release notes.


forrestbaer (approved):
🧹

PR #2487, opened 2023-05-02
[WNMGDS-2292] Storybook to honor healthcare's error placement default + pad the bottom errors

pwolfert:
## Summary

WNMGDS-2292

- Add logic in the Storybook theme switcher to update the default error placement for healthcare
- [In our component spacing update](https://github.com/CMSgov/design-system/pull/2396), we got rid of the bottom margin on all form fields. This resulted in a regression for bottom-placed errors where there was no longer any margin in between the field and the bottom error. That has been fixed by adding a class that sets a half-spacer of margin and can be overridden by specific components if we find a need for it

## How to test

Check all the form field stories and see that they place the error on the bottom by default in the healthcare theme

## Checklist


forrestbaer (approved):
Nice work! Good point about flag settings.

PR #2484, opened 2023-04-29
[WNMGDS-2282] Fix `labelId` leaking through `TextField` into the DOM

pwolfert:
## Summary

https://jira.cms.gov/browse/WNMGDS-2282

Fixes the following error coming from the use of `labelId` on `TextField`:
```
console.error
    Warning: React does not recognize the `labelId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `labelid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

## How to test

Run the unit tests. There should be no console errors related to `labelId`.

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.

### If this is a change to code:

- [x] Created or updated unit tests to cover any new or modified code
- [x] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:update`)

PR #2482, opened 2023-04-28
[WNMGDS-1250] Added `DrawerManager` context provider and `useDrawerManager` hook to manage state of multiple drawers on a single page.

forrestbaer:
## Summary

- [RFC](https://github.com/CMSgov/design-system/discussions/1951)
- Adds a `<DrawerManager>` context provider and `useDrawerManager()` hook for managing drawer state so that multiple drawers don't overlap one another.
- Added documentation with example
- Added unit test
- Updated VRT snapshots

## How to test

1. `yarn clean && yarn install && yarn build`
2. documentation update: `yarn start` then -> http://localhost:8000/components/drawer/#managing-multiple-drawers
3. `yarn test:unit`
4. `yarn storybook` then see drawer manager example

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.

### If this is a change to code:

- [x] Created or updated unit tests to cover any new or modified code
- [x] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:update`)

PR #2488, opened 2023-05-02
[NO-TICKET] Found some VRT diffs between Apr 30th and May 2nd

forrestbaer:
## Summary

- PR's between the timeline when they were found:
    - #2484 
    - #2483 
    - #2482 
    - #2485


pwolfert (approved):
Hopefully these intermittent failures go away after this is merged. I don't think it's worth the time to investigate why they're different at the moment.

PR #2495, opened 2023-05-09
[WNMGDS-2334] Improve the Autocomplete stories

pwolfert:
## Summary

WNMGDS-2334

My goals are that the stories would exhibit more realistic behaviors, content be more plausible, and instructions be clearer.

- So we can test that assistive tech responds to changing results, we want the results to be dynamic and based on user input
- I relegated the county example to one story because, while it is a realistic scenario that we might want to showcase, the results themselves are harder to distinguish between when testing with a screen reader. I replaced the results in the default story instead with drug names that start with different letters of the alphabet.

## How to test

[Try the new Autocomplete stories](https://cmsgov.github.io/design-system/branch/pwolfert/better-autocomplete-stories/storybook/?path=/story/components-autocomplete--default), and [compare with the current version](https://design.cms.gov/storybook/?path=/story/components-autocomplete--default&globals=theme:core).

## Checklist


packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -306,17 +306,20 @@ export const Autocomplete = (props: AutocompleteProps) => {
...
+
   return (
     <div className={rootClassName}>
       {renderChildren()}
 
       <div
-        className={classNames('ds-c-autocomplete__list', !isOpen && 'ds-u-display--none')}
+        className="ds-c-autocomplete__list"
         id={menuContainerId}
+        hidden={!(isOpen && menuContent)}
```
pwolfert, line 318: Oh, this change came up during a pair programming session for the new Dropdown component. We can use a built-in browser attribute here rather than relying on our more verbose utility class

forrestbaer (approved):
Nice update, all looks good!

PR #2496, opened 2023-05-10
[WNMGDS-2322] Doc Site Search

forrestbaer:
## Summary

Added Doc site searching!

- Could probably use some feedback on what to do with it on mobile.
- `yarn start` and mess around with it! I'll try to get a demo url up.


packages/docs/src/components/icons/CmsLogo.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+/**
+ * SVG for CMS Logo
+ */
+const CMSLogoIcon = () => {
+  // "#323a45" was what Brent originally sent
+  const fill = 'var(--color-base)';
+
+  return (
+    <SvgIcon title="CMS logo" viewBox="-2 0 88 34" className="c-cms-logo-icon ds-u-font-size--5xl">
```
pwolfert, line 12: I don't remember if the SvgIcon component always makes them square, but the logo is ending up square, which makes the header really tall. We probably don't need the SvgIcon component and could do something similar to what we do for the HealthCare.gov logo.

packages/docs/src/components/page-templates/SearchPage.tsx
```diff
@@ -0,0 +1,63 @@
+import React from 'react';
+import { useEffect, useState } from 'react';
+import { TextField } from '@cmsgov/design-system';
+import { useFlexSearch } from 'react-use-flexsearch';
+import { useStaticQuery, graphql } from 'gatsby';
+import { SearchDataStore, SearchQuery } from '../../helpers/graphQLTypes';
+import avoidRefresh from '../../helpers/avoidRefresh';
+
+const SearchPage = () => {
```
pwolfert, line 9: Can all this be done in the `pages/search.tsx` file? It seems kind of weird that we'd have to define it in `pages`, `content` and `page-templates`

pwolfert (commented):
I know you're not done, but I did look through it and had some comments as I went along. This is so cool!

packages/docs/src/components/page-templates/SearchPage.tsx
forrestbaer: Yeah probably, I had set it up this way as was the way things have been wired, at some point I tried moving it all into one file and ran into some issue so I didn't look at it again, will see about making that happen.   The `SearchForm` in content is just a search form component so that will remain separate for use anywhere.

forrestbaer:
Tweaking the results configuration a bit but otherwise ready for review.

packages/docs/src/components/icons/CmsLogo.tsx
pwolfert: Looks much better now!

packages/docs/src/components/content/SearchForm.tsx
```diff
@@ -0,0 +1,43 @@
...
+import { getQueryParamValue } from '../../helpers/urlUtils';
+
+const SearchForm = (props) => {
+  const [inputText, setInputText] = useState('');
+
+  const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
+    event.preventDefault();
+
+    if (inputText) {
+      window.open(`${location.origin}/search?query=${inputText}`, '_self');
```
pwolfert, line 13: What's the benefit of calling `window.open` vs setting the `location.href`? Was curious about this and did [some research](https://stackoverflow.com/a/23394403) that shows one might be more reliable than the other. We also seem to be doing it a third way [elsewhere in our doc site](https://github.com/CMSgov/design-system/blob/forrest-docsite-search/packages/docs/src/helpers/urlUtils.ts#L74-L76).

pwolfert: A simpler option, too, might be to have no state or effects and let the HTML form behavior run wild without preventing default. But maybe there's a reason we can't do that?

packages/docs/src/components/layout/HeaderFullWidth.tsx
```diff
@@ -0,0 +1,17 @@
...
+import SearchForm from '../content/SearchForm';
+
+const HeaderFullWidth = () => {
+  return (
+    <div className="ds-l-row ds-u-margin--0 ds-u-padding--2 ds-u-padding-left--0 ds-u-border-bottom--2">
+      <div className="ds-u-display--flex ds-l-col--6 ds-u-align-items--center ds-u-padding-left--0">
+        <CMSLogoIcon />
+        <h1 className="ds-u-padding-left--1 ds-text-heading--2xl ds-u-margin--0">Design System</h1>
+      </div>
+      <SearchForm className="ds-u-display--flex ds-l-col--6 ds-u-align-items--center ds-u-justify-content--end" />
```
pwolfert, line 12: This layout doesn't really work for mobile. Should we hide the search form on mobile and stick another one in the mobile menu that is hidden on desktop?

packages/docs/src/components/layout/HeaderFullWidth.tsx
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+import CMSLogoIcon from '../icons/CMSLogo';
+import SearchForm from '../content/SearchForm';
+
+const HeaderFullWidth = () => {
+  return (
+    <div className="ds-l-row ds-u-margin--0 ds-u-padding--2 ds-u-padding-left--0 ds-u-border-bottom--2">
+      <div className="ds-u-display--flex ds-l-col--6 ds-u-align-items--center ds-u-padding-left--0">
+        <CMSLogoIcon />
+        <h1 className="ds-u-padding-left--1 ds-text-heading--2xl ds-u-margin--0">Design System</h1>
```
pwolfert, line 10: I'm not sure what modern best practice is here, but I'm used to making these site logos/headings links to the root of the site. Should we do that?

packages/docs/src/pages/search.tsx
```diff
@@ -0,0 +1,94 @@
...
+          )}
+          <ul>
+            {results.map((result: SearchDataStore) => {
+              let body = stripHTML(result.body);
+              const strLoc = body.indexOf(query);
+              if (strLoc <= 0) {
+                return;
+              } else {
+                body = body.slice(strLoc - 100, strLoc + 100);
+                body = body.replace(query, `<strong>${query}</strong>`) + '...';
```
pwolfert, line 78: These results are looking so good! What would you or @brentschneider or @line47 think about putting a highlight background color behind the query string? Like a yellow or something?

pwolfert (commented):
This is looking really good! Just a few comments on details

packages/docs/src/components/layout/HeaderFullWidth.tsx
line47: Yeah we should show the search link on mobile in the header maybe?

forrestbaer:
Trying to get a demo site up now, has most of the features asked for to review.

forrestbaer:
> I did run into something that I wanted to bring up...
> When I search for any of the colors found on the system colors page like "Orchid" or "Goldenrod" I don't get any results back?

@line47 this is because those items are in a component and compute their value at render time, the way this gatsby search works it looks at the data in the GraphQL database and not the rendered page content. Not sure this can be resolved unfortunately unless we want to just put that content directly into those MDX files instead of using components.

packages/docs/src/components/layout/HeaderFullWidth.tsx
pwolfert: Everything besides the mobile view of the header LGTM. Mobile view used to look like this:

Screenshot 2023-05-16 at 9 44 29 AM

and now it looks like this:

Screenshot 2023-05-16 at 9 44 38 AM

The spacing on the page content looks better, but we're missing the site heading, and the padding is off.

pwolfert: @line47 , how would you feel about putting the menu button the right side like we do on healthcare.gov? That would allow the site title to left-align with the text on the page if we got those margins right.

pwolfert: @forrestbaer, I see it moved into the mobile menu, but the header is still blank on mobile when it should probably say "CMS Design System" like it did before

brentschneider:
Hi @forrestbaer   
  
I'm seeing in search some differences between number of results "X" found and number of results displayed on the page.  
  
Terms used: CORE and CMS

packages/docs/src/pages/search.tsx
```diff
@@ -0,0 +1,94 @@
...
+      frontmatter={{
+        title: 'Search this site',
+      }}
+      location={location}
+      slug="Search"
+      theme={theme}
+    >
+      <div id="search-results">
+        <div className="search-form ds-u-margin-bottom--4">
+          <TextField
```
zarahzachz, line 52: This might need to have the `search` type on it since it's acting as a search field.

packages/docs/src/pages/search.tsx
zarahzachz, line 78: You might want to use the `<mark>` tag instead of `<strong>` ([source](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/mark)). I _think_ this is what Google uses when you perform CTRL+F searches on webpages.

packages/docs/src/pages/search.tsx
```diff
@@ -0,0 +1,94 @@
...
+              if (strLoc <= 0) {
+                return;
+              } else {
+                body = body.slice(strLoc - 100, strLoc + 100);
+                body = body.replace(query, `<strong>${query}</strong>`) + '...';
+              }
+              return (
+                <li key={result.id}>
+                  <a href={location.origin + '/' + result.path + location.search}>{result.title}</a>
+                  <p dangerouslySetInnerHTML={{ __html: body }} />
```
zarahzachz, line 83: It'd be great if the markdown were parsed - but I don't know how difficult that would be. I think in some results I got when searching "button" the markdown syntax for links wasn't compiled correctly (you could see all the parens and brackets).

packages/docs/src/styles/components/pageHeader.scss
```diff
@@ -17,3 +17,10 @@
   margin-right: $spacer-half;
   vertical-align: middle;
 }
+
+#header-full-width {
```
zarahzachz, line 21: I thought we avoided styling `id`s - because an `id` is usually reserved for JS interactivity and not CSS. Not sure if that's a hard rule though.

packages/docs/src/styles/pages/search.scss
```diff
@@ -0,0 +1,27 @@
+#search-results {
```
zarahzachz, line 1: Same `id` comment from above.

zarahzachz (commented):
This is really awesome!

zarahzachz (commented):
This is really awesome!

packages/docs/src/pages/search.tsx
forrestbaer: I did some parsing already to take out most of the HTML & Markdown that I could but what we get from graphQL is limited and so are the transforms we can do live.  We could probably write some kind of graphQL transform thing that gives us what we want during build in graphql.

packages/docs/src/styles/components/pageHeader.scss
forrestbaer: Oh didn't realize, I'll just make these classes.

forrestbaer:
> I'm seeing in search some differences between number of results "X" found and number of results displayed on the page.

Yeah unfortunately I haven't been able to get better results from the search, it sometimes returns items that don't include the term entered so I just hid those, as to why you aren't seeing the results. And since updating of that search data is coming from a hook I wasn't able to update that result like I had hoped. Something we can look at this afternoon at our dev meeting perhaps.

pwolfert (approved):
LGTM. It's functional, and any further improvements we want to make can be done in another PR

PR #2498, opened 2023-05-11
[WNMGDS-2317] Re-implement `Dropdown` as combobox using `Downshift`

pwolfert:
## Summary

WNMGDS-2317

- Switch from using native `select` to implementing our own ARIA _combobox_ using the Downshift library. For reasons why, see the ticket
- Implement a basic style to be refined later when new designs are approved. Note that...
    - Some CSS vars are borrowed from `Autocomplete` for now
    - Disabled icon isn't grayed. The icon might be changing, so I'm not going to mess with refactoring icon usage yet
- We're still supporting two different ways of supplying options and option groups in our API

## How to test

[Branch demo](https://cmsgov.github.io/design-system/branch/pwolfert/dropdown-downshift/storybook/?path=/story/components-dropdown--default&globals=theme:core)
Try the Storybook stories

## Checklist


zarahzachz (commented):
It's weird, but the label on the default dropdown in SB docs is white

pwolfert:
@zarahzachz good catch. That's weird. The docs page body has a base--inverse class 🤔

pwolfert:
Fixed the inverse story issue here: https://github.com/CMSgov/design-system/pull/2502

tests/browser/snapshots/components-dropdown--disabled-cmsgov--Mobile-Chrome.png
zarahzachz, line 1: I think you said styles would be further finessed in a later ticket?

tests/browser/snapshots/components-dropdown--inverse-option-core--chromium.png
zarahzachz, line 1: Again, noticed the error color is off but that might be something addressed later.

tests/browser/snapshots/dropdown--open--cmsgov--Mobile-Chrome.png
zarahzachz, line 1: I'm surprised this passed color contrast rules.

tests/browser/snapshots/dropdown--open--cmsgov--chromium-forced-colors.png
zarahzachz, line 1: HCM will need to have special styling to indicate selected items. I'd think `Highlight` and `HighlightText` would be what we'd want ([source](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color#syntax))

packages/design-system/src/styles/components/_Dropdown.scss
```diff
@@ -18,4 +25,68 @@ select.ds-c-field {
     color: transparent;
     text-shadow: 0 0 0 #000;
   }
+
+  .ds-c-dropdown--open & {
+    border-end-end-radius: 0;
```
zarahzachz, line 30: This made me do a double-take 😆  I never knew that's how this radius was targeted!

packages/design-system/src/components/Dropdown/Dropdown.stories.jsx
```diff
@@ -23,25 +26,38 @@ const dropdownOptions = [
...
-      <option value="2-3">Option 6</option>
+    <optgroup label="Group B">
+      <option value="b-1">Option B-1</option>
+      <option value="b-2">Option B-2</option>
+      <option value="b-3">Option B-3</option>
     </optgroup>
   </>
 );
 
+const htmlOptions = (
```
zarahzachz, line 44: Are these "html" prefixed stories supposed to have native HTML inside this select? Running it locally and I'm not seeing this markup - it's all lists.

packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -1,49 +1,63 @@
...
+import React, { useMemo, useRef } from 'react';
 import classNames from 'classnames';
 import useAutofocus from '../utilities/useAutoFocus';
 import { FormFieldProps, FormLabel, useFormLabel } from '../FormLabel';
+import { useSelect, UseSelectStateChangeOptions } from 'downshift';
+import { isOptGroupArray, parseChildren, validateProps } from './utils';
+import { uniqueId } from 'lodash';
 
-export type DropdownDefaultValue = number | string;
-export interface DropdownOptions {
```
zarahzachz, line 7: Reminder: Removing this exported interface will make this a major change

packages/design-system/src/components/Dropdown/__snapshots__/Dropdown.test.tsx.snap
```diff
@@ -1,66 +1,199 @@
...
+  </li>
+</ul>
+`;
+
 exports[`Dropdown dropdown matches snapshot 1`] = `
 <div
-  class=""
+  class="ds-c-dropdown"
 >
   <label
```
zarahzachz, line 77: ~Just dawned on me that it seems weird having a `<label>` associated with a `<button>`. Does the `for` on label and `id` on button work?~

Nevermind... Just read the [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/label) and buttons can be associated with labels. I learned something!

> Elements that can be associated with a <label> element include [<button>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button), [<input>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input) (except for type="hidden"), [<meter>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/meter), [<output>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/output), [<progress>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/progress), [<select>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select) and [<textarea>](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/textarea).

packages/design-system/src/components/Dropdown/utils.tsx
```diff
@@ -0,0 +1,96 @@
...
+    optionsOrGroups?.length > 0 && (optionsOrGroups as DropdownOptGroup[])[0].options !== undefined
+  );
+}
+
+function findElementsOfType<T extends keyof JSX.IntrinsicElements>(
+  type: T,
+  node: React.ReactNode
+): React.ReactElement<any, T>[] {
+  if (!node || !(React.isValidElement(node) || Array.isArray(node))) {
+    // There's nothing to recurse on, and this is not the droid we're looking for
```
zarahzachz, line 39: lol

packages/design-system/src/components/Dropdown/utils.tsx
```diff
@@ -0,0 +1,96 @@
...
+    return [];
+  }
+
+  if (React.isValidElement(node) && type === node.type) {
+    // We found it! Return an array because it will be flattened
```
zarahzachz, line 44: I feel like I'm seeing the point where Patrick finally cracked.

zarahzachz (commented):
Few style comments but everything looks good

packages/design-system/src/components/Dropdown/utils.tsx
pwolfert: lol 😂

packages/design-system/src/styles/components/_Dropdown.scss
pwolfert: I started with "bottom-right" and then was like, "we're now using logical properties...how i do this?" It's pretty confusing, tbh. I wish I could say `border-end-radius` and do both at once.

forrestbaer (approved):
All my tests ran without any hitches, it looks good, good tests 💯

packages/design-system/src/components/Dropdown/Dropdown.stories.jsx
pwolfert: It accepts HTML but outputs the same thing as the non-HTML examples. It parses child elements and converts it into options that can be fed to our Downshift-based listbox

pwolfert:
Dan has feedback in WNMGDS-2338 that may or may not result in significant changes. I am going to merge this and make those changes a separate pull request, especially since that feedback has its own ticket.

PR #2499, opened 2023-05-12
[WNMGDS-2309] Fix tooltips closing when moving cursor between trigger and tooltip

pwolfert:
## Summary

WNMGDS-2309

- Restores the `box-sizing: content-box` to an element that extends the reach of the tooltip hitbox so it receives hover events around the edges
- Fixes visual regression with the tooltip arrow
- Changes the margins used in the tooltip stories so there is room for top-positioned tooltips

## How to test

In the default storybook story, using a mouse or touchpad, hover over the tooltip trigger and then slowly move the pointer up towards the tooltip. On the main branch, the tooltip will disappear when you get to a certain invisible in-between space. On this branch, it should be fixed.

## Checklist

~- [ ] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`~
~- [ ] Selected appropriate `Impacts`, multiple can be selected.~
- Because this is a fix to the beta version and not a production version, I don't think it needs to be put in the release notes


forrestbaer (approved):
Looks great!

PR #2505, opened 2023-05-18
[WNGMDS-2175] Convert abs font sizes to rel

zarahzachz:
## Summary

WNMGDS-2175

You can check the [demo url](https://cmsgov.github.io/design-system/branch/WNMGDS-2175/convert-abs-font-sizes-to-rel) to see if you spot any visual changes - but this should end up being a 1:1 switcheroo (no visible changes _unless_ you change the font size in your browser settings - then the font sizes should be exactly how you've defined them).


tests/browser/snapshots/mask--text-entry--cmsgov--webkit.png
zarahzachz, line 1: No true visual change - test just input an extra '2' in the text field.

tests/browser/snapshots/components-privacysettingsdialog--default-healthcare--Mobile-Chrome.png
zarahzachz, line 1: This is the only VRT with noticeable changes - possibly due to the strange scrolling bug we've seen in the past. Since other themes/browsers don't show this issue, I think this is related to the scrolling bug we've seen.

packages/ds-medicare-gov/src/styles/components/_Dialog.scss
```diff
@@ -9,7 +9,6 @@
 
     .ds-c-icon {
       color: var(--color-gray-dark);
-      font-size: 22px;
```
zarahzachz, line 12: This style can be removed as dialogs now have a token controlling close icon sizing in each theme.

pwolfert:
Should we put this in the 7.0 milestone?

pwolfert (approved):
LGTM!

PR #2506, opened 2023-05-22
[WNMGDS-2293] UsaBanner now appears on mobile/tablet versions of the doc site.

forrestbaer:
## Summary

- Making `UsaBanner` visible at all breakpoints on the doc site.


packages/docs/src/components/layout/Layout.tsx
```diff
@@ -83,7 +83,7 @@ const Layout = ({
       </Helmet>
       <SkipNav href="#main" />
 
-      <UsaBanner className="ds-u-display--none ds-u-md-display--block" />
+      <UsaBanner />
```
pwolfert, line 86: This adds it, but the mobile menu interaction doesn't feel as polished now. There's a jump. Maybe what we could do is have one banner here for desktop and one in the mobile container that gets a fixed position when open, something that will allow the header to take into account the height of the banner without having to manually specify a height (which is a moving target if it can open and close and text can wrap)

forrestbaer: Very good point, updated the fix.

pwolfert (approved):
Looks great!

pwolfert (approved):
Looks great!

PR #2508, opened 2023-05-22
[WNMGDS-2338] Improve VoiceOver support for Dropdown

pwolfert:
## Summary

WNMGDS-2338

- Manually creates a message in an aria-live region (through Downshift's API) that reads out options being highlighted to get around a [known VoiceOver issue](https://a11ysupport.io/tests/apg__aria-1-2-combobox-with-list-autocomplete-example#assertion-aria-aria-activedescendant_attribute-convey_value-aria-combobox_role)
- Removes `ariaLabel` as an explicit optional prop on `Dropdown` because it complicates things, and users can still add an `aria-label` attribute if they want because of the automatic pass-through
- Removes the `aria-labelledby` from the `listbox` because it was overly verbose and redundant
- Works around an issue where [buttons cannot expose both label and values at the same time](https://www.24a11y.com/2019/select-your-poison/)

## How to test

For the VoiceOver fix,

1. Open the default story in Safari
2. Turn on VoiceOver
3. Focus on the select
4. Use the down and up arrows to move through items. When it's done with its initial readout of the label, it should start to announce the item being highlighted

### To do
- [ ] Run this updated version by an accessibility expert

[Demo url](https://cmsgov.github.io/design-system/branch/pwolfert/dropdown-a11y-fixes/storybook/?path=/story/components-dropdown--default&globals=theme:healthcare)

## Checklist


packages/design-system/src/components/Dropdown/useHighlightStatusMessageFn.ts
```diff
@@ -0,0 +1,34 @@
...
+const defaultGetA11yStatusMessage: GetA11yStatusMessageFn = (
+  (Downshift as any).defaultProps as DownshiftProps<any>
+).getA11yStatusMessage;
+
+/**
+ * Fake the output we want to see in VoiceOver, since VoiceOver currently has
+ * a known bug where it doesn't read changes to the `aria-activedescendant`.
+ * In this function, we're telling Downshift to create an "a11y message" to
+ * read out the currently highlighted item when the highlighted index changes.
+ * More more info on the VoiceOver issue, see:
```
zarahzachz, line 15: Typo. Should be "For more..."

packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -222,7 +225,7 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
       props.inversed && 'ds-c-field--inverse',
       fieldClassName
     ),
-    'aria-label': ariaLabel,
+    'aria-labelledby': `${buttonContentId} ${labelId}`,
```
zarahzachz, line 228: There's something up with this... I'm investigating. Something I noticed is the button seems to read out the word "semicolon" before the button label. I've tried modifying this line to see if it's the culprit (changed it to `FOO ${buttonContentId} ${labelId}`) and my changes aren't announced anywhere in the element. 

I'll push something up once I get to the bottom of it.

zarahzachz: "semicolon" appears to be read out when the dropdown first loads before a selection is made.

zarahzachz:
Created a [test url](https://cmsgov.github.io/design-system/branch/pwolfert/dropdown-a11y-fixes) to review with Angela

forrestbaer (approved):
Works well for me! I wasn't able to hear the semicolon callout that Sarah mentioned.

PR #2511, opened 2023-05-25
[WNMGDS-2348] Remove react props table from docs site

zarahzachz:
WNMGDS-2348

- Update Storybook link in component header to point to the Docs story in Storybook
- Update code examples in docs site
    - Remove React/HTML dropdown
    - Replace stuff under examples with link to corresponding Storybook
    - Remove React props tables from docs site*

*Can't really remove all React props tables until components have been updated with new Storybook syntax. Seeing some components with props tables for multiple components (e.g., Accordion, Table, etc.). Removing props tables where I can.

### DIFF
--- a/packages/docs/content/components/accordion.mdx
+++ b/packages/docs/content/components/accordion.mdx
@@ -4,7 +4,7 @@ intro: An accordion is a list of headers that hide or reveal additional content 
 core:
   githubLink: design-system/src/components/Accordion
   sketchLink: ygDKPRA
-  storybookLink: components-accordion--default
+  storybookLink: components-accordion--docs
 healthcare:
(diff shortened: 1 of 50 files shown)


forrestbaer (approved):
Hey this is nice! It feels a little plain but maybe that's fine? Would a little storybook icon or even an arrow or something be worthwhile? Anyway looks good to me! And nice fix for the iframe size updates..

zarahzachz:
> Hey this is nice! It feels a little plain but maybe that's fine? Would a little storybook icon or even an arrow or something be worthwhile? Anyway looks good to me! And nice fix for the iframe size updates..

Yeah, I think updating the visuals is something Brent's working on. We're interviewing folks to get their thoughts on how they use the docs site right now to see what we should do about that.

PR #2512, opened 2023-05-30
[WNMGDS-2312] Updating default args for several component stories

forrestbaer:
## Summary

WNMGDS-2312

- I believe I've made the updates requested, @line47 I will need your feedback if this was done correctly.
- The naming of the components is a bit verbose, I kinda like having shorter story names, downside is we have to update snapshots across the board.

## How to test

1. `yarn start` or `yarn storybook`

## Checklist


pwolfert (approved):
Will the doc site changes for the TextField component mentioned in the ticket come in a separate PR?

forrestbaer:
> Will the doc site changes for the TextField component mentioned in the ticket come in a separate PR?

Ah, I did not understand the ticket that way, so we should also be removing those examples, I'll add that here.

PR #2513, opened 2023-05-31
[LHDM-1281] Modify a11y label rendering for Alert

seesleestak:
## Summary

For some time, the Learn team has struggled to determine the root cause of some React hydration console errors in our app. These errors can be seen in production on pages like https://www.healthcare.gov/retirees/. After lots of testing and troubleshooting, we've found that those errors are caused by the way the Alert a11y label is rendered (and for reasons that continue to elude us). Modifying the rendering approach for the children of the a11y label `span` is the only solution we've found that appears to address our errors.
 
- Learn ticket for researching console errors: https://jira.cms.gov/browse/LHDM-1281

## How to test

The testing of this change within the context of the Learn app requires linking and building the DS. However, the effects of the change can be seen by comparing the pre-change html output to the post-change html output in storybook.

| Before | After |
| ----- | ----- |
| Screenshot 2023-05-31 at 3 42 27 PM | Screenshot 2023-05-31 at 3 42 49 PM | 

## Checklist


forrestbaer (approved):
Now I'm curious how you found the root of this problem, must have been difficult!

Fix seems good to me. 👍🏻

seesleestak:
It was certainly an adventure :)
Building the DS with it linked to our application and removing things until I stopped seeing errors. A lot of trial and error.

pwolfert:
This is really strange. I'm curious to see how it affected our compiled library code. Do you know, @seesleestak, if your project is building with the ES Modules (`/dist/esnext` or `/dist/react-components/esm`) version?

pwolfert:
Yep, exactly. Thanks!

pwolfert (approved):
Before:
```js
var a11yLabel = /*#__PURE__*/ React.createElement(
  "span",
  {
    className: "ds-c-alert__a11y-label ds-u-visibility--screen-reader",
    id: a11yLabelId,
  },
  t(
    "alert.".concat(
      variation !== null && variation !== void 0 ? variation : "defaultLabel"
    )
  ),
  ":",
  " "
);
```

After:
```js
var a11yLabelText = t(
  "alert.".concat(
    variation !== null && variation !== void 0 ? variation : "defaultLabel"
  )
);
var a11yLabel = /*#__PURE__*/ React.createElement(
  "span",
  {
    className: "ds-c-alert__a11y-label ds-u-visibility--screen-reader",
    id: a11yLabelId,
  },
  "".concat(a11yLabelText, ": ")
);
```

The biggest difference I can see that is reflected in your screenshots is that in the _after_ version there is only a single content string parameter being sent to `createElement`.

I wish we knew what is actually causing the issue so we could avoid it in the future or figure out a longer-term fix, but the code changes you've made make it more readable anyway, so this LGTM 👍

PR #2516, opened 2023-06-02
[WNMGDS-2353] Icon error color token variable name fixed

forrestbaer:
## Summary

- `icon__color__error` -> `icon__color--error` 🥇 
- WNMGDS-2353

image

## Checklist


forrestbaer:
> There are some unrelated VRT changes in here. Not sure why the button ones are changing. Do you? The others are the known flaky ones.

Maybe I should take all those out.

I'm not sure, just more vrt strangeness.. I wonder if we should consider at some point doing these vrt updates via some kind of interface that uses the same environment every time, like a jenkins job that sends screenshot diffs to confirm/block the passing status via slack.

Very out of scope for this work but might be helpful to free up time and energy in the future.

PR #2519, opened 2023-06-05
[WIP] [WNMGDS-2337] Apply dropdown styles

zarahzachz:
## Summary

WNMGDS-2337

Apply styles to dropdown as seen in [Sketch doc](https://www.sketch.com/s/c09342ba-8b6f-4121-86cb-e447c1990cbd/v/jJdZLD/a/JA8aMw5#Inspect).

Includes:
- Updating dropdown icon background colors, font-weight an icon visibility depending on state (default, highlighted, selected)
- Add border radius to bottom of menu container when open
- Add chevron, and position it depending on menu being opened or closed
- Add animation to dropdown items and menu opening
- Add box-shadow to menu container
- Add truncation style to lengthy options once selected


packages/design-system/src/styles/components/_Dropdown.scss
```diff
@@ -58,16 +66,48 @@
...
+  &:not([hidden]) {
+    // We don't have a good way to create alpha versions of colors using our current hex value tokens
+    // Sass has color functions, but they can't be used in CSS custom properties because CSS custom properties are evaluated at runtime
+    // Hex values can't be used in native CSS color functions like `hsla()` or `rgba()`
+    // Creating a temporary variable until we discuss a better solution
+    --shadow-color: 0deg 0% 50%;
+    animation: fade-in var(--animation-speed-1) var(--animation-ease-in-out-expo);
+    border-radius: 0 0 var(--dropdown-menu__border-radius--open)
+      var(--dropdown-menu__border-radius--open);
+    box-shadow: 0px 0.2px 0.2px hsl(var(--shadow-color) / 0.43),
```
zarahzachz, line 80: We don't have a drop-shadow token that matches the Sketch. I tried repurposing the same drop-shadow from dialog, but it was too dark/severe compared to the design. This was the closest I could get it.

packages/design-system/src/styles/components/_Dropdown.scss
```diff
@@ -58,16 +66,48 @@
...
+  100% {
+    opacity: 1;
+  }
+}
+
+// Base styles for grouped and individual dropdown elements
 .ds-c-dropdown__item,
 .ds-c-dropdown__item-group {
   border-block-end: var(--dropdown-item__border-width) solid var(--dropdown-item__border-color);
+  cursor: default;
```
zarahzachz, line 110: I don't know if we wanted the cursor to look like an arrow when hovering over this element, but I figured we'd want it to behave as close to a native select as possible, and select options use this cursor style.

packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -2,6 +2,7 @@ import React, { useMemo, useRef } from 'react';
 import classNames from 'classnames';
 import useAutofocus from '../utilities/useAutoFocus';
 import { FormFieldProps, FormLabel, useFormLabel } from '../FormLabel';
+import { ArrowIcon, CheckIcon } from '../Icons';
```
zarahzachz, line 5: These icons might be placeholders for now... We don't have the icons from the designs yet.

packages/design-system-tokens/src/themes/core-component.ts
```diff
@@ -298,9 +298,10 @@ export const components: AnyTokenValues = {
 
   'dropdown': {
     '__background-color':                         t.color['white'],
-    '__icon-size':                                '10px',
+    '__icon-size':                                '24px',
     '-item__border-width':                        '1px',
     '-item__border-color':                        t.color['gray-lighter'],
+    '-menu__border-radius--open':                 '3px',
```
pwolfert, line 304: Since this should always match the border radius of the ds-c-field button, I don't actually think we should track this as a separate variable.

packages/design-system/src/styles/components/_Dropdown.scss
pwolfert: I think we should try to stick with the shadows we have in the design system rather than trying to match Sketch exactly. I can't imagine there was pixel-perfect intentionality with the drop-shadow design. Was there, @brentschneider?

pwolfert: Like is there a way simpler version of this that would work? Could we just match the calendar picker drop shadow or the tooltip drop shadow?

packages/design-system/src/styles/typography-body-text.stories.tsx
```diff
@@ -28,36 +28,26 @@ export const AllBodyText = () => {
     </>
   );
 };
+
 export const AllBodyTextOnDark = AllBodyText.bind({});
```
pwolfert, line 32: These changes aren't related. Can this and the other typography stories be removed from the diff?

packages/design-system/src/styles/components/_Dropdown.scss
```diff
@@ -58,16 +66,48 @@
...
   }
+
+  // When dropdown is open (not `hidden`), add border radius to bottom of menu
+  &:not([hidden]) {
+    // We don't have a good way to create alpha versions of colors using our current hex value tokens
+    // Sass has color functions, but they can't be used in CSS custom properties because CSS custom properties are evaluated at runtime
+    // Hex values can't be used in native CSS color functions like `hsla()` or `rgba()`
+    // Creating a temporary variable until we discuss a better solution
+    --shadow-color: 0deg 0% 50%;
+    animation: fade-in var(--animation-speed-1) var(--animation-ease-in-out-expo);
```
pwolfert, line 77: When @brentschneider and I talked last, I think we decided that a new bezier curve and animation speed needed to be created for this. I think @brentschneider was going to define what that curve looked like, but you may need to work together on it.

packages/design-system/src/styles/typography-body-text.stories.tsx
brentschneider: Earlier I mentioned the calendar picker drop shadow.

packages/design-system/src/styles/components/_Dropdown.scss
```diff
@@ -1,37 +1,45 @@
...
-    border-end-start-radius: 0;
+  [aria-expanded='false'] & {
+    transform: rotate(0deg);
   }
 }
 
+// Styles for the dropdown menu
 .ds-c-dropdown__menu-container {
   background-color: var(--autocomplete__background-color);
   border: var(--text-input__border-width) solid var(--text-input__border-color);
```
zarahzachz, line 45: Something I noticed: Our error border (`--text-input__border-widthh--error`) is 3px wide and the field border (`--text-input__border-width`) is 2px wide. So when an errored dropdown opens, the menu's border is 2px wide (matching the default field's width). 

Is there a reason our border's are off by 1px? I'd argue there's not enough of a difference between these 2 values to be noticeable by our users so perhaps we should have the same width for all field states.

Applies to:
- [CMS theme](https://github.com/CMSgov/design-system/blob/main/packages/design-system-tokens/src/themes/cmsgov-component.ts#L413-L414)
- [Core theme](https://github.com/CMSgov/design-system/blob/main/packages/design-system-tokens/src/themes/core-component.ts#L413-L414)
- [Hgov theme](https://github.com/CMSgov/design-system/blob/main/packages/design-system-tokens/src/themes/healthcare-component.ts#L423-L424)
- [Mgov theme](https://github.com/CMSgov/design-system/blob/main/packages/design-system-tokens/src/themes/medicare-component.ts#L430-L433) (they also set border-width for disabled fields to 1px?)

packages/design-system/src/styles/components/_Dropdown.scss
zarahzachz: There's an issue using our date picker shadows.

The shadow spreads around the HTML element its attached to and I can't find a way to "turn off" one side of the shadow like you would with a border. So what we get is a shadow that encircles the entire element. It doesn't look bad on the date picker because the calendar opens a distance away from the field, but because the dropdown menu is attached to the field, a gray haze fills the field (see attached images to see this issue). 

The few ways I could think to fix this resulted in unwanted side effects:
1. I could place the dropdown menu behind the field (adjust the z-index), but this results in the field's focus ring appearing on top of the menu. 
2. I could adjust the offset-y of the shadow so it moves the shadow down far enough it doesn't overlap the field, but because the blur-radius is so large (17px) it makes the menu appear as if it's very far off the page. 
3. The previous way I had shadows _seemed_ to fix this overlap issue, but it results in more one-off code and might be harder to maintain in the long run (no corresponding shadow color token, no tokens for the many overlapping alphas, etc.).

Screenshot 2023-06-06 at 2 41 22 PM
Screenshot 2023-06-06 at 2 41 30 PM

This is the way I had the shadows (the overly complex way). I changed the color of the shadow to red so you could more easily see if it was overlapping and I can't tell if it is:
Screenshot 2023-06-06 at 2 56 19 PM

pwolfert: Would it be possible to place put the shadow on a pseudo-element child of the wrapper element of the whole component? Maybe the wrapper could have a z-index of 1 and the shadow element 0?

zarahzachz: I went with an off-menu option - used `clip-path` to cut the shadow off from overlapping the field. The only downside here is it results in:
- The shadow partially overlapping the focus ring
- A harsh line between the shadow and no-shadow (the light grey color on white background makes this less obvious imo)

I shared with Brent and this appears to be the exact same solution (and problems) he faced when designing in Sketch and got a 👍 on it.

image

brentschneider (approved):

PR #2520, opened 2023-06-06
[WNMGDS-2351] Fixed page scroll jump on dialog close.

forrestbaer:
## Summary

WNMGDS-2351

- Temporarily sets the `scroll-behavior` to `auto` when opening a dialog to eliminate animation.
- Moves padding from dialog container to div with document role beneath it to resolve display of text at top of dialog

## How to test

1. `yarn storybook` -> [http://localhost:6006/iframe.html?args=&id=components-dialog--prevent-scroll-example&viewMode=story](http://localhost:6006/iframe.html?args=&id=components-dialog--prevent-scroll-example&viewMode=story) 

## Checklist


packages/design-system/src/styles/components/_Dialog.scss
```diff
@@ -44,14 +44,18 @@ dialog.fixed {
...
   text-align: start;
   width: 95%; // provide space for the background layer to peek through
 
   @media (-ms-high-contrast: active), (forced-colors: active) {
     outline: $spacer-2 solid WindowText;
   }
 
+  > [role='document'] {
+    padding: var(--dialog__padding);
+  }
```
pwolfert, line 57: Could you further explain the problem this is fixing? I don't see a problem with text inside the dialog being called out in the ticket, and I don't see it myself.

pwolfert (commented):
It works! I have a question about the document role fix.

I also realized that the solution to the content-shift problem mentioned in one of the ticket comments is actually a fix that we can just do in our story; I don't think it's actually a fix we need to make in our component. [This line that optionally renders the dialog](https://github.com/CMSgov/design-system/blob/WNMGDS-2351-page-scroll-fix/packages/design-system/src/components/Dialog/Dialog.stories.jsx#L86) should actually be moved down so that it never becomes the first element of its parent, which bumps the h1 out of its rightful place. Could move down to be near the button or something.

packages/design-system/src/styles/components/_Dialog.scss
forrestbaer: This solves the problem of dialogs opening with their text at the top of the window. If you open the privacy settings dialog for example you get: 

image

With this fix it moves the container padding to the content making it appear 'correctly'.

zarahzachz: It might be something meant to solve our PrivacyModal VRT issue. 

I'm assuming Forrest discovered the `tabindex` on the `role="document"` meant it would sometimes scroll into focus. And since the padding was applied to the dialog container itself instead of this  element, the top padding was missing resulting in our VRTs showing differences. That's a guess though.

pwolfert: Oh sweet! Nice!

forrestbaer:
@pwolfert ok jeeze, had to switch the PR target branch to something else then back to main for it to pick up the last 2 commits, thought it might be a setting due to the PR having been approved or something.

PR #2523, opened 2023-06-07
[WNMGDS-2355] Add inputRef and unit test to Dropdown

zarahzachz:
## Summary

WNMGDS-2355

Wire `inputRef` prop up to work with refs in component and add unit test for it.


pwolfert (approved):
LGTM 👍

PR #2525, opened 2023-06-07
[WNMGDS-2373] Update ThirdPartyExternalLink to use anchor instead of button

pwolfert:
## Summary

WNMGDS-2373

We talked it over with accessibility experts, and we've decided to make this a link. Our argument for using a link is that even though it doesn't perform the navigation directly, the intent of the element is to perform navigation, and links are best suited for navigation. This also allows desktop users to hover over the link and see the address, and it allows some other standard browser behaviors related to links. It is also way easier to style to look like a link, because it is a link. We realized we were going to have to fix the styles anyway because its primary use case is to be within blocks of text.

## How to test

`yarn build && yarn storybook` and go to http://localhost:6006/?path=/story/components-thirdpartyexternallink--default

## Checklist

- [] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:update`)


tests/browser/snapshots/components-thirdpartyexternallink--default-cmsgov--Mobile-Chrome.png
zarahzachz, line 1: I think visually this looks much better than the button link - but is it supposed to be visually different? Not sure if we need Melissa's sign-off for changes in underline offset and font-weight.

zarahzachz (approved):
Approving pending if we're supposed to visually change this component

tests/browser/snapshots/components-thirdpartyexternallink--default-cmsgov--Mobile-Chrome.png
pwolfert: Because we want it to behave like other links in a particular design system, I think whatever the link styling is should be how this is styled (which is what this update changes it to). I don't think we ever got Melissa's direct approval on the first version of this component for the cmsgov theme anyway. Looking at CMS.gov, we may need to refine the link styles in general to use bold text, but that can be made in a separate update, since this just inherits those styles.

PR #2533, opened 2023-06-12
[NO-TICKET] Insert the 6.1 versions into our version lists for doc site

pwolfert:
## Summary

When we build the next beta version, we want the 6.1 versions to be listed in our version switcher.


versions.json
```diff
@@ -1,8 +1,8 @@
 {
   "design-system": [
     "7.0.0-beta.1",
+    "6.1.2",
     "6.0.1",
-    "6.0.0",
```
pwolfert, line 5: This version seems redundant to list. Thought I'd remove it while making manual changes.

PR #2539, opened 2023-06-15
[WNMGDS-2368] Update autocomplete styles

zarahzachz:
## Summary

WNMGDS-2368

- Update class names in Autocomplete component to match naming structure from Dropdown
- Update "item" class in Dropdown to "menu-item" to better indicate relationship with "menu" and "menu-container" classes
- Remove incorrect dashes from some Autocomplete classes for correct BEM syntax (e.g., `ds-c-autocomplete__menu-item--message` is incorrect because `--message` isn't really a modifier)
- Fix tests


packages/design-system/src/styles/components/_Autocomplete.scss
```diff
@@ -1,60 +1,37 @@
...
+  @extend .ds-c-dropdown__menu;
 }
 
-.ds-c-autocomplete__list-item--active {
-  background-color: var(--autocomplete-item__background-color--active);
-  color: var(--autocomplete-item__font-color--active);
+// Experienced an error using `@extend .ds-c-dropdown__menu-item` where it wouldn't compile styles for this class.
+// I'm assuming the cause is due to the extended class rules are also applied to `.ds-c-dropdown__menu-item-group` (.ds-c-dropdown__menu-item, .ds-c-dropdown__menu-item-group {}). When styles are split into individual rules, it works, but is less DRY. Attempted using the % placeholder selector, but that didn't work either.
+.ds-c-autocomplete__menu-item {
+  padding: var(--text-input__padding);
```
zarahzachz, line 27: I don't think `@extend` not working here is a blocker since it's using a token all themes have. The purpose of this class is to ensure the list items have the same padding as the text fields in a given theme and this token solves that.

forrestbaer (approved):
Looks good!

PR #2542, opened 2023-06-16
[NO-TICKET] Bugfix: adding cmsgov to list of packages for dropdown to fix error

ghost:
## Summary

- Current version switcher @ design.cms.gov breaks when switching to CMS theme
- Added cmsgov to list of themes to pull package data correctly in docs

## How to test

1. `yarn start`, switch to cms theme

## Checklist


zarahzachz (approved):
That explains why it would break! Good find!

PR #2544, opened 2023-06-20
[WNMGDS-2384] Revert heading spacing changes

pwolfert:
## Summary

- Reverts https://github.com/CMSgov/design-system/pull/2534
- Removes `:first-child` rules on plain heading and paragraph elements
  - We've gotten feedback that they interfere in places they shouldn't. Those rules assumed they would always be used in content areas, but really that should be left up to the `ds-content` class or individual typography classes
- Removes headings from list of components with removed bottom spacing in docs
- Refine the examples and diagrams in the new spacing migration guide

## Checklist


packages/design-system/src/styles/_base.scss
```diff
@@ -139,12 +139,24 @@ $breakpoints: (
   font-family: var(--typography-heading__font-family);
   font-weight: var(--typography-heading__font-weight);
   line-height: var(--font-line-height-heading);
+  margin: 0;
+}
+
+@mixin heading-margins {
+  margin-block: 1.5em 0.5em;
+
+  &:first-child {
```
zarahzachz, line 148: Suggestion: Remove the styles within `:first-child` and `:last-child`.

I've had the most questions around this decision and it does seem to introduce issues for teams when using singular headings on a page. This was written under the assumption our headings would be used with other blocks of text.

pwolfert: Thanks for the reminder. I think it makes sense to remove it. I'll do it in a follow-up PR

pwolfert: It actually looks like [this line was the same in v6](https://github.com/CMSgov/design-system/blob/release/6.0/packages/design-system/src/styles/_base.scss#L162-L168). Did you mean [this line in the reset that applies to plain headings](https://github.com/CMSgov/design-system/blame/pwolfert/revert-heading-spacing-changes/packages/design-system/src/styles/_reset.scss#L65)?

zarahzachz: Oh! Yeah, I think I did get those mixed up!

pwolfert: Updated in b4dd0b90c733f69d6e9df04e3b58a2325dbca8f1

packages/docs/content/migration-guides/v7-spacing-updates.mdx
```diff
@@ -34,11 +33,11 @@ While margin collapse is an understandable concept, it is prone to producing une
...
   </figure>
 </div>
 
-In the above screenshot, notice the inconsistency in how spacing between elements is determined. Between a heading and a paragraph of text, the margins stack, causing there to be a total of 34 pixels betwen them. Between a paragraph of text and a field label, the margins collapse, causing it to take the higher value of the two, which is 24 pixels. Compare this with the following "after" screenshot:
+In the above screenshot, notice the inconsistency in how spacing between elements is determined. Between a choice list and a paragraph of text, the margins stack, causing there to be a total of 24 pixels betwen them. Between a paragraph of text and a field label, the margins collapse, causing it to take the higher value of the two, which is 24 pixels. There are also 4 pixels of padding below the last field. Compare this with the following "after" screenshot where all spacing comes from top margins:
```
pwolfert, line 40: While we've already sunk a lot of time into these diagrams, I actually think we might need to change the example scenario again. We serendipitously get 24px between each item, which actually appears to be more consistent than the "after" state, which might undermine what we're trying to communicate here 😬

forrestbaer (approved):
Looks good! 👍🏻

PR #2549, opened 2023-06-22
[WNMGDS-2354] Fix toolbar in Storybook

zarahzachz:
## Summary

- Refactor `preview.tsx` file to use Storybook 7 syntax for `globalTypes`, `parameters`, and `decorators`
- Add `@storybook/addon-viewport` package to Storybook to wire up viewport toolbar correctly

To test...

i18n: http://localhost:6006/?path=/story/components-month-picker--default
analytics: http://localhost:6006/?path=/story/components-dialog--dialog-example (view console and/or actions panel)


.storybook/preview.tsx
```diff
@@ -16,93 +17,46 @@ import type { UtagContainer } from '@cmsgov/design-system';
 // Rewire analytics events to log to the console
 (window as UtagContainer).utag = { link: console.log };
 
-const customViewports = {
+const breakpointViewportSizes = {
   extraSmall: {
-    name: 'Extra Small - 320px',
+    name: '$media-width-xs',
     styles: {
       width: '320px',
```
zarahzachz, line 24: Worth pointing out, [Storybook documentation shows height/width values as numbers](https://storybook.js.org/docs/react/essentials/viewport#configuration) and not strings. When you follow through to [the code](https://github.com/storybookjs/storybook/blob/next/code/addons/viewport/src/defaults.ts#L167), it shows they are in fact using strings.

.storybook/preview.tsx
```diff
@@ -16,93 +17,46 @@ import type { UtagContainer } from '@cmsgov/design-system';
...
       height: '800px',
     },
+    type: 'mobile',
```
zarahzachz, line 27: [Storybook docs say `type` is needed](https://storybook.js.org/docs/react/essentials/viewport#configuration) for the viewport objects, but they lie. Kept this in because it's not hurting anything.

.storybook/preview.tsx
```diff
@@ -165,9 +119,65 @@ const analyticsSettingsDecorator = (Story, context) => {
...
+    backgrounds: { disable: true },
+    controls: {
+      expanded: true,
+      matchers: {
+        color: /(background|color)$/i,
+        date: /Date$/,
+      },
+    },
+    viewport: {
+      viewports: breakpointViewportSizes,
```
zarahzachz, line 172: When testing, you'll notice a (L) or a (P) next to the selected viewport. This is to denote "landscape" or "portrait" orientation. I can't find a way to change this indicator - it'd be great if I could have it say the thing it's actually doing instead of abbreviating.

.storybook/main.ts
```diff
@@ -1,7 +1,12 @@
 import type { StorybookConfig } from '@storybook/react-webpack5';
 
 const config: StorybookConfig = {
-  addons: ['@storybook/addon-links', '@storybook/addon-essentials', '@storybook/addon-mdx-gfm'],
+  addons: [
+    '@storybook/addon-links',
+    '@storybook/addon-essentials',
+    '@storybook/addon-mdx-gfm',
+    '@storybook/addon-viewport',
```
zarahzachz, line 8: Storybook has no documentation showing this addon is needed to get the viewport toolbar working. However, when you strip out all logic from the preview.tsx and main.ts files (e.g., emulating a bare bones project), even the default viewports don't work. Only way I could get it working was adding this addon.

.storybook/preview.tsx
```diff
@@ -16,93 +17,46 @@ import type { UtagContainer } from '@cmsgov/design-system';
 // Rewire analytics events to log to the console
 (window as UtagContainer).utag = { link: console.log };
 
-const customViewports = {
+const breakpointViewportSizes = {
   extraSmall: {
-    name: 'Extra Small - 320px',
+    name: '$media-width-xs',
```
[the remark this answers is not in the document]
zarahzachz: Also, removed the number associated with the name since Storybook shows the width next to all viewports anyway (it's redundant).

.storybook/preview.tsx
```diff
@@ -165,9 +119,65 @@ const analyticsSettingsDecorator = (Story, context) => {
...
+    },
+    analytics: {
+      name: 'Analytics',
+      description: 'Analytics settings',
+      defaultValue: 'off',
+      toolbar: {
+        icon: 'graphline',
+        items: [
+          { value: 'on', left: 'Analytics', title: 'Log to Actions' },
+          { value: 'off', left: 'Analytics', title: 'Off' },
```
pwolfert, line 144: The "Analytics" prefixes are rendering weird. We could just remove them. Hopefully the menu title gives enough context.

pwolfert (approved):
LGTM except for one weird thing in the analytics menu

PR #2552, opened 2023-06-26
[NO-TICKET] Add v6.1.3 to doc site versions lists

pwolfert:
Add v6.1.3 to doc site versions lists


versions.json
```diff
@@ -2,7 +2,7 @@
   "design-system": [
     "7.0.0-beta.2",
     "7.0.0-beta.1",
-    "6.1.2",
+    "6.1.3",
```
zarahzachz, line 5: I'm curious why we're replacing 6.1.2?

I was thinking this was because it's a patch and we only provide the latest patch+ version of docs per release, but I'm seeing 5.0.0, 5.0.1, and 5.0.2 below so I don't fully understand the reasoning.

zarahzachz (commented):
I'm curious why we're replacing

versions.json
pwolfert: We could remove `5.00` and `5.0.1` from the list too. My thought is that the lists are getting really long, and the docs don't change much with the patches, and the unpatched versions probably aren't that valuable on their own. Since we do have them on the doc site, though, maybe we want to keep them around 🤷. Maybe the next iteration of the version switcher won't use a dropdown.

zarahzachz: Yeah I was thinking it was to prevent too long a list/consolidating small updates. But the existence of the older patches threw a wrench in this pattern for me.

Personally, I think the latest patches of a version are the only versions we'd need to show.

pwolfert: Alright, then I'll trim the list of the other old patches in this PR

pwolfert: Hmm, in this case, do we need the betas? Maybe not

zarahzachz: Just to have a paper trail on this, we discussed in UX sync that preserving old betas aren't needed but we should continue to include the most recent beta. (Only one beta - recent one - needs to have a docs site)

versions.json
```diff
@@ -2,85 +2,37 @@
...
-    "3.7.0-beta.1",
     "3.6.0",
-    "3.6.0-beta.3",
-    "3.6.0-beta.2",
-    "3.6.0-beta.1",
     "3.4.0"
   ],
   "ds-medicare-gov": [
     "9.0.0-beta.2",
     "9.0.0-beta.1",
```
zarahzachz, line 16: Do we need both betas?

pwolfert: I'll remove both since we're about to do a release

zarahzachz (approved):

PR #2553, opened 2023-06-27
[WNMGDS-2294] Updated Storybook stories to CSF3 format, MDX to MDX2 format

forrestbaer:
## Summary

WNMGDS-2294

- Updated all storybook stories to Typescript
- Updated all storybook stories to CSF3 format
- Updated MDX that needed updating to MDX2 format
- Small updates to storybook config for updates
- Removed `argTypes` for stories

## How to test

1. `yarn test:browser`
2. `yarn storybook` evaluate

## Checklist


forrestbaer:
Looks like I messed something up in the merge conflict with buttons, looking into it.

zarahzachz (changes requested):

packages/design-system/src/components/FilterChip/FilterChip.tsx
```diff
@@ -26,15 +26,15 @@ export interface FilterChipProps {
...
+  onDelete: () => any;
   /**
    *  Use alternate thinner close icon in place of standard.
    */
   useAlternateIcon?: boolean;
   /**
    * Sets the size of the chip to larger version.
    */
-  size?: 'big';
+  size?: string;
```
zarahzachz, line 37: Does changing these types cause a breaking change?

packages/design-system/src/components/VerticalNav/VerticalNav.stories.tsx
```diff
@@ -0,0 +1,76 @@
...
+    collapsed: false,
+    nested: false,
+  },
+};
+export default meta;
+
+type Story = StoryObj<typeof VerticalNav>;
+
+const Link = ({ className, ...props }) => (
+  // <Link to={props.href} {...props}>{props.children}</Link>
```
zarahzachz, line 19: Does this commented line need to be here?

packages/design-system/src/components/Accordion/Accordion.stories.tsx
```diff
@@ -0,0 +1,86 @@
...
+import { Accordion } from './Accordion';
+import { AccordionItem } from './AccordionItem';
+import { useArgs } from '@storybook/preview-api';
+
+const meta: Meta<typeof Accordion> = {
+  title: 'Components/Accordion',
+  component: Accordion,
+  args: {
+    bordered: true,
+    openItems: [0],
```
pwolfert, line 12: I think this should be moved to the controlled story, because it only works there, and it's not a prop of the component.

packages/design-system/src/components/FilterChip/FilterChip.tsx
forrestbaer: I don't believe so, it's just making the type looser. There were some typing issues with having this as a string, I could revisit what exactly it was but this was the fix that worked for me to get it going, I could probably resolve the issue without changing this file though, but this is a simple way to do it.

.storybook/preview.tsx
```diff
@@ -56,6 +56,45 @@ const breakpointViewportSizes = {
       width: '1280px',
       height: '800px',
     },
+  },
+};
+
+export const globalTypes = {
```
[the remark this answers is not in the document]
forrestbaer: Thanks, I must have duplicated them during conflict resolution.

packages/design-system/src/components/VerticalNav/VerticalNav.stories.tsx
forrestbaer: No, good 👁️ !

packages/design-system/src/components/Button/Button.stories.tsx
```diff
@@ -2,102 +2,90 @@ import React from 'react';
...
-export const OnDark = Template.bind({});
-OnDark.args = {
-  onDark: true,
-};
-OnDark.parameters = {
-  // Must supply `layout: 'fullscreen'` when we use `onDark: true`
-  onDark: true,
-  layout: 'fullscreen',
-};
+export const Default: Story = {};
```
pwolfert, line 19: Oh whoa, so we don't even need a render function when it's just the plain component?

packages/design-system/src/components/ChoiceList/Choice.stories.tsx
```diff
@@ -13,17 +14,9 @@ export default {
...
   },
-  argTypes: {
-    checkedChildren: { control: 'text' },
-    errorMessage: { control: 'text' },
-    hint: { control: 'text' },
-    label: { control: 'text' },
-    requirementLabel: { control: 'text' },
-    uncheckedChildren: { control: 'text' },
-    value: { control: 'text' },
-  },
```
pwolfert, line 24: Some of the controls are wonky on the docs page, but overall this did a nice job of inferring the types. And the `ReactNode` type is way more usable in this version!

pwolfert: No action needed here. Just commenting.

packages/design-system/src/components/DateField/SingleInputDateField.stories.tsx
```diff
@@ -0,0 +1,59 @@
...
+};
+
+export const WithPicker: Story = {
+  ...Default,
+  args: {
+    label: 'What day did you move?',
+    hint: 'This date should be within the past 60 days in order to qualify',
+    fromYear: new Date().getFullYear(),
+    // TODO: Due to some unknown issue with this story that causes us to lose args
+    // defined with query parameters, we can't supply a specific date in the
```
pwolfert, line 44: Oh, I think this was the problem I was talking about when we were troubleshooting `useArgs`

packages/design-system/src/components/FilterChip/FilterChip.tsx
pwolfert: The `onDelete` change is harmless, but this one with the `size` causes devs to lose the hint in their code editor as to what one string value is allowed here.

pwolfert (commented):
I added to Sarah's comment about avoiding changing the `FilterChip` API for a storybook fix, but other than that, this looks good. Going through all the stories gave me the opportunity to find some unrelated issues that we'll want to try to fix before releasing that I've shared with the team.

packages/design-system/src/components/Button/Button.stories.tsx
forrestbaer: Yeah in cases where there are no args to pass in it's not needed, note to you and @zarahzachz if you don't have a render function with args the args will not show up as editable in SB. So if you run into a case where you aren't seeing args being configurable it's likely a missing render function with args passed..

```
render: function Component(args) { },
```

.storybook/preview.tsx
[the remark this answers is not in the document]
forrestbaer: Could you mark this change as ok when you review?

zarahzachz (approved):

PR #2555, opened 2023-06-27
[NO-TICKET] Fix inverse stories being too tall

pwolfert:
## Summary

A follow-up to https://github.com/CMSgov/design-system/pull/2502, this fixes a problem where the inverse stories were really tall when viewed in the component docs pages in storybook, where they aren't in iframes.

This is what it looked like before:

Inverse stories taking up way more vertical space than they should on a docs page in storybook

## How to test

Go to the button docs page in storybook and scroll down to the inverse story. Also try the inverse story itself and try it out on the doc site as well.

### DIFF
--- a/.storybook/storybookStyles.scss
+++ b/.storybook/storybookStyles.scss
@@ -14,11 +14,14 @@
 // Storybook's styles change, we'll need to update these rules. It is used for
 // the decorator of on-dark stories.
 .on-dark-story {
-  height: 100vh;
   padding: 1rem;
 
   #storybook-docs & {
     padding: 30px 20px;
+  }
+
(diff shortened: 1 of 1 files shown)


.storybook/storybookStyles.scss
```diff
@@ -14,12 +14,15 @@
...
 .on-dark-story {
-  height: 100vh;
   padding: 1rem;
 
   #storybook-docs & {
     padding: 30px 20px;
   }
+
+  #storybook-root & {
+    height: 100vh;
```
zarahzachz, line 24: Checked locally and this works.

zarahzachz (approved):

PR #2561, opened 2023-06-30
[NO-TICKET] Update docs for `Autocomplete` `items` prop

pwolfert:
## Summary

- Be more clear in the docs about the relationship between `items` and the messaging coming from the component


packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -105,7 +105,9 @@ export interface AutocompleteProps
    */
   itemToString?: UseComboboxProps<any>['itemToString'];
   /**
-   * Array of objects used to populate the suggestion list that appears below the input as users type. This array of objects is intended for an async data callback, and should conform to the prescribed shape to avoid errors.
+   * Array of objects used to populate the suggestion list that appears below the input as users type.
+   * Passing an empty array will show a "No results" message. If you do not yet want to show results,
+   * pass in `undefined` or `null`.
```
kaipokent, line 110: Right now passing `null` gives a TS error since `items?:` would equate to `AutocompleteItem[] | undefined`

pwolfert: Oh good catch! Thanks

kaipokent (approved):
🎉 Thanks for updating the documentation!

PR #2567, opened 2023-07-03
[WNMGDS-2415] Refactor AccordionItem

zarahzachz:
## Summary

WNMGDS-2415

- Refactor the AccordionItem from class to function component.
- Delete Hgov Accordion story file, redundant as the core story is themed correctly for hgov


packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -49,98 +49,86 @@ export interface AccordionItemProps {
...
+  const contentClasses = classNames('ds-c-accordion__content', contentClassName);
+  const buttonClasses = classNames('ds-c-accordion__button', buttonClassName);
+  const HeadingTag = `h${headingLevel}` as const;
+  const CloseIcon = closeIcon as React.ReactNode;
+  const OpenIcon = openIcon as React.ReactNode;
+  const isControlled = !!onChange;
+  const contentId = id || uniqueId('accordionItem_');
+  const buttonId = `${contentId}-button`;
+  const [isOpen, setIsOpen] = useState(() => {
+    return isControlled ? isControlledOpen : defaultOpen;
```
pwolfert, line 91: This doesn't need to be a function as far as I can tell

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -49,98 +49,86 @@ export interface AccordionItemProps {
...
-  }
+export const AccordionItem: React.FC<AccordionItemProps> = ({
+  buttonClassName,
+  children,
+  contentClassName,
+  defaultOpen = false,
+  heading,
+  headingLevel = '2',
+  id,
+  isControlledOpen,
```
pwolfert, line 65: For a future update, maybe we should make a note to change the name of this to just `isOpen`. It doesn't conflict with anything that I can tell. We could add that prop and deprecate the old one for a while.

packages/design-system/src/components/Accordion/AccordionItem.tsx
```diff
@@ -49,98 +49,86 @@ export interface AccordionItemProps {
...
-    } = this.props;
+  };
 
-    const contentClasses = classNames('ds-c-accordion__content', contentClassName);
-    const buttonClasses = classNames('ds-c-accordion__button', buttonClassName);
-    const HeadingTag = `h${headingLevel}` as const;
-    const isItemOpen = this.isControlled ? isControlledOpen : this.state.isOpen;
-    const CloseIcon = closeIcon as React.ReactNode;
-    const OpenIcon = openIcon as React.ReactNode;
+  const isItemOpen = isControlled ? isControlledOpen : isOpen;
```
pwolfert, line 103: To test this logic, I wrote a unit test:

```jsx
it('uses isControlledOpen as source of truth', () => {
    const onChange = jest.fn();
    const baseProps = { ...defaultProps, children: defaultChildren, onChange };
    const {rerender} = renderAccordionItem({ ...baseProps, isControlledOpen: false });

    const buttonEl = screen.getByRole('button');
    expect(buttonEl).toHaveAttribute('aria-expanded', 'false');

    rerender(<AccordionItem {...baseProps} isControlledOpen={true}>{defaultChildren}</AccordionItem>);
    expect(buttonEl).toHaveAttribute('aria-expanded', 'true');

    rerender(<AccordionItem {...baseProps} isControlledOpen={false}>{defaultChildren}</AccordionItem>);
    expect(buttonEl).toHaveAttribute('aria-expanded', 'false');

    // Even clicking the button shouldn't expand it if it's controlled
    userEvent.click(buttonEl);
    expect(buttonEl).toHaveAttribute('aria-expanded', 'false');
  });
```

pwolfert (commented):
It's working! I'm going to be doing a similar thing to the Tabs component soon

pwolfert (approved):
LGTM!

PR #2570, opened 2023-07-05
[WNMGDS-2419] Make Dropdown option checks less strict

pwolfert:
## Summary

WNMGDS-2419

In the last update, we moved the Dropdown implementation away from using the native HTML select element to implementing a combobox pattern with the Downshift library. The native select element doesn't care if you give it no options; it'll just look weird. Our component actually made checks to validate the options given to it and would throw errors if they were malformed. However, we discovered in the latest patch update that our doc site build has transient states where our version dropdown list will have malformed options until the version list loads from disk, which caused build errors.

- Logs warnings to the console when options are malformed instead of throwing JavaScript errors. We want to alert developers when something's not quite right, but we don't necessarily want to blow up if the HTML `select` wouldn't
- Adds tests and a story for controlled Dropdowns

## How to test

1. Modify one of the stories so that `options` is an empty array
2. Start storybook and view the story
3. Observe that there are console warnings instead of it crashing
4. Also try a `value` that doesn't match any of the option values

## Checklist


forrestbaer (approved):
Working well for me.

PR #2573, opened 2023-07-06
[WNMGDS-2420] Left align wrapping text in Accordion; side effect of reset CSS

zarahzachz:
No ticket for this work.

If teams have extra extra extra long text in the Accordion headers, the text center-aligns. Text should always be left-aligned.

This behavior came about because of the button CSS reset. Button text is center aligned. Accordions use buttons. Therefore, Accordion text ends up center aligned unless that style is overwritten.


pwolfert (approved):
Verified that the fix works 👍 

Wait to merge until we get a ticket

PR #2578, opened 2023-07-10
[WNMGDS-2379] Update USA banner code

zarahzachz:
## Summary

WNMGDS-2379

Update USA banner code:
- Refer to [official USWDS banner examples](https://designsystem.digital.gov/components/banner/)
- [Lofi mockups (Confluence)](https://confluence.cms.gov/display/HCDSG/The+official+website+of+the+United+States+banner+component)
- [Theme alignment (Confluence)](https://confluence.cms.gov/display/HCDSG/Component+alignment+accross+themes)


packages/design-system-tokens/src/themes/cmsgov-component.ts
```diff
@@ -449,11 +449,20 @@ export const components: AnyTokenValues = {
 
   'usa-banner': {
     '__background-color':                         t.color['gray-lightest'],
-    '__color':                                    t.color['black'],
-    '-close__background-color':                   t.color['gray-light'],
-    '-action__color':                             t.color['primary-darker'],
-    '-lock-icon__color':                          t.color['success-light'],
-    '-heading__font-size':                        '0.75rem',
+    '__color':                                    t.color['base'],
+    '__max-width':                                t.global['site-max-width'],
```
zarahzachz, line 453: Adding a custom prop to allow themes to determine width of USA banner (it's a token called out on USWDS docs as well). This prop uses the value defined in `site-max-width` to determine how the banner will line up with other global components like a header.

Experiencing issues getting `t.global[]` to build. Error says the value is undefined, but the theme file does have a `site-max-width` value within the `global` const, so I'm not sure if there's another way to import this value?

packages/design-system-tokens/src/themes/cmsgov-component.ts
```diff
@@ -449,11 +449,20 @@ export const components: AnyTokenValues = {
...
+    '-heading__font-size':                        t.font['size-sm'],
+    '-panel__font-family':                        t.font['sans'],
+    '-panel__line-height':                        t.font['line-height-base'],
+    '-panel__font-size':                          t.font['size-base'],
+    '-link__color':                               t.color['primary'],
+    '-domain-icon__color':                        t.color['primary'],
+    '-security-icon__color':                      t.color['success-light'],
+    '-close-button__background-color--mobile':    t.color['gray-lighter'],
+    '-close-button__color--mobile':               t.color['primary'],
   },
```
zarahzachz, line 466: Added more custom props than USWDS called for - mainly because I can foresee future themes wanting more customization.

USWDS calls out the following variables:
- `$theme-banner-background-color`
- `$theme-banner-font-family`
- `$theme-banner-link-color`
- `$theme-banner-max-width`

packages/design-system-tokens/src/themes/cmsgov-component.ts
```diff
@@ -449,11 +449,20 @@ export const components: AnyTokenValues = {
...
-    '-close__background-color':                   t.color['gray-light'],
-    '-action__color':                             t.color['primary-darker'],
-    '-lock-icon__color':                          t.color['success-light'],
-    '-heading__font-size':                        '0.75rem',
+    '__color':                                    t.color['base'],
+    '__max-width':                                t.global['site-max-width'],
+    '-gutter-width':                              t.global['grid-gutter-width'],
+    '-heading__font-family':                      t.font['sans'],
+    '-heading__line-height':                      '1.1',
+    '-heading__font-size':                        t.font['size-sm'],
```
zarahzachz, line 457: Our original implementation used `.75rem` (12px) as the font-size for the banner heading text. USWDS is slightly larger at `.8rem` (12.8px). The smallest font size in our DS was `.875rem` (14px) and I used that for consistency - can easily change though.

packages/design-system-tokens/src/themes/cmsgov-component.ts
```diff
@@ -449,11 +449,20 @@ export const components: AnyTokenValues = {
...
+    '-panel__font-family':                        t.font['sans'],
+    '-panel__line-height':                        t.font['line-height-base'],
+    '-panel__font-size':                          t.font['size-base'],
+    '-link__color':                               t.color['primary'],
```
zarahzachz, line 461: This token name comes from USWDS. We originally used `-action__color`. I figured it'd be best to align with USWDS tokens, but can change this back.

packages/design-system-tokens/src/themes/cmsgov-component.ts
```diff
@@ -449,11 +449,20 @@ export const components: AnyTokenValues = {
 
   'usa-banner': {
     '__background-color':                         t.color['gray-lightest'],
-    '__color':                                    t.color['black'],
-    '-close__background-color':                   t.color['gray-light'],
-    '-action__color':                             t.color['primary-darker'],
-    '-lock-icon__color':                          t.color['success-light'],
```
zarahzachz, line 455: Changed this from `-lock-icon__color` to `-security-icon__color` - figured it'd be easier to maintain if we decided to change icons in the future. What if the icon stopped being a lock at some point? Might be over-engineering here...

packages/design-system/src/components/UsaBanner/UsaBanner.tsx
```diff
@@ -29,123 +29,75 @@ export interface UsaBannerProps {
  */
 export const UsaBanner: React.FunctionComponent<UsaBannerProps> = (props: UsaBannerProps) => {
   const [isBannerOpen, setBannerOpen] = useState<boolean>(false);
-  const [shouldRenderMobileView, setShouldRenderMobileView] = useState<boolean>(false);
   const classes = classNames('ds-c-usa-banner', props.className);
   const id = props.id || uniqueId('gov-banner_');
 
-  useEffect(() => {
```
zarahzachz, line 36: Found a way, using USWDS as a reference, to make this component without the extra JS

packages/design-system/src/components/UsaBanner/UsaBanner.tsx
```diff
@@ -29,123 +29,75 @@ export interface UsaBannerProps {
...
-
   const toggleBanner = () => {
     setBannerOpen(!isBannerOpen);
   };
 
-  const flagIcon = (
-    <UsaFlagIcon
-      className="ds-c-usa-banner__header-flag"
-      title={t('usaBanner.flagIconTitle')}
-      ariaHidden={false}
```
zarahzachz, line 63: USWDS sets this as `aria-hidden` - I think because this is a decorative-only icon. So setting `ariaHidden={false}` isn't needed here.

packages/design-system/src/components/locale/en.json
```diff
@@ -135,19 +135,19 @@
     "collapse": "Collapse sub-navigation"
   },
   "usaBanner": {
-    "bannerLabel": "Official government website",
+    "flagIconTitle": "U.S. Flag",
```
zarahzachz, line 138: I think this can be deleted. I didn't because it's being called in the flag icon component itself and I don't know if there's a reason that is - it still has `aria-hidden` applied, so doesn't look like it's doing anything.

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -1,189 +1,208 @@
...
+  color: var(--usa-banner-color);
 }
 
 .ds-c-usa-banner__header {
-  align-items: flex-start;
-  display: flex;
+  // margin-inline: auto;
+  // max-width: var(--usa-banner__max-width);
+  // padding-inline: calc(var(--usa-banner-gutter-width) * 0.5);
+  @extend .ds-l-container;
```
zarahzachz, line 16: I would like to remove these `@extend` rules if possible. I tried creating custom props where teams could overwrite the banner's max-width if needed, but they aren't generating correctly 😢

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -1,189 +1,208 @@
...
+    position: relative;
+    text-align: start;
+    width: fit-content;
+  }
 }
 
 .ds-c-usa-banner__button-text {
-  @extend .ds-c-link;
+  // This element is screen-reader only on mobile/tablet
+  @extend .ds-u-visibility--screen-reader;
```
zarahzachz, line 113: I tried putting this in a media query (`@media (max-width: $media-width-md) {}`) but discovered you can't `@extend` in a `@media` - o how the limitations of SCSS vex me so!

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -1,189 +1,208 @@
...
+
+  @media (min-width: $media-width-md) {
+    // Need to overwrite specific screen reader only styles so the element becomes visible again on larger viewports
+    // Would be nice to have some more flexibility with our utils:
```
zarahzachz, line 117: I can remove this comment later. I was venting.

zarahzachz:
I need to walk through this PR with @brentschneider

packages/design-system-tokens/src/themes/cmsgov-component.ts
zarahzachz: Something to comment on here... if teams can customize their font family, it makes sense to also allow them to customize their line-heights and font-sizes. Not all fonts are built the same, so expecting a heading to have the same visual weight regardless of font-size is a little short-sighted imho.

packages/design-system-tokens/src/lib/types.ts
```diff
@@ -43,7 +43,7 @@ export interface ThemeTokens {
   animation: AnimationTokens;
   color: ColorTokens;
   font: FontTokens;
-  global?: AnyTokenValues;
+  global: AnyTokenValues;
```
zarahzachz, line 46: This change was needed so I could reference `t.global[]` tokens from theme tokens

packages/design-system/src/components/locale/en.json
pwolfert: USWDS hides the icon from screen readers?

packages/design-system/src/components/__snapshots__/i18n.test.ts.snap
```diff
@@ -1,29 +1,29 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`i18n getTranslations and addTranslations getTranslation returns core English translations 1`] = `"Official government website"`;
+exports[`i18n getTranslations and addTranslations getTranslation returns core English translations 1`] = `"Official website of the United States government"`;
```
pwolfert, line 3: Are these all updates from the latest version of USWDS or something?

packages/design-system/src/styles/components/_UsaBanner.scss
pwolfert: Hmm, maybe it's better to use the utility class in this case. It would be easier to debug in the future. I know we want to limit use of utility classes in our own component code, but I feel like the screen reader utility is an exception

packages/design-system/src/styles/components/_UsaBanner.scss
pwolfert: Hmm, yeah, I see what you mean. This is tricky. I wish it were easy to rename our utilities to give us room for more variations, but at this point it would be a wide-sweeping breaking change. Like `ds-u-min-md-*` vs `ds-u-md-*`. It's also more code in our css bundle

packages/design-system-tokens/src/themes/cmsgov-component.ts
pwolfert: This change seems fine to me, but we should probably run it by @brentschneider when he's back. Not sure if any stakeholders will want to know about the height difference ahead of time.

packages/design-system/src/styles/components/_UsaBanner.scss
pwolfert: Now that I've seen your other comments about overriding this later, I understand why you did this

pwolfert: Should we actually make a mixin for the screen reader styles so we can use them exactly where and how we want? Would we then be able to put them inside a media query with max width like you suggested?

packages/design-system/src/components/locale/en.json
zarahzachz: Might be tough to see... but in USWDS's banner, they're using an `` instead of an `<svg>`, so attributes are obviously different, but on the image they have `aria-hidden="true"` and an empty `alt` attr, which tells me this is presentational only.

Screenshot 2023-07-12 at 10 14 11 AM

I don't think including a title or aria-label for this icon is actually helpful for AT users because it doesn't provide additional context or information. "U.S. Flag" doesn't really add to an a11y experience imo, so I think USWDS's approach is the correct one.

packages/design-system/src/styles/components/_UsaBanner.scss
zarahzachz: A mixin is worth exploring... looking into that now!

packages/design-system/src/components/__snapshots__/i18n.test.ts.snap
zarahzachz: Ah yes, USWDS had one change with the English `aria-label` that I updated.

zarahzachz: Oh - and we were using the wrong apostrophes in our JSON file. I think it was straight vs curly apostrophes? Not sure, but it was flagged as an incorrect symbol in VS Code

packages/design-system/src/components/locale/en.json
pwolfert: That makes sense. The "An official website of the United States government" seems like a better opening line than "U.S. Flag". While we make it possible for people to use the icon outside of it, in that rare occurance when someone wants to use it, they can add their own title. Removing it cuts the bundle size a bit.

pwolfert (approved):
I love how much more succinct the React code is 😍 

Code side LGTM. We'll probably want to alert stakeholders that content has been updated

brentschneider:
Hi @zarahzachz 

I got a chance to pull and run the latest changes locally.
From the colors you shared in slack lest go with Ocean for the building and Spring for the lock.

USA-banner-icon-colors

brentschneider (approved):
Absolutely flawless, awesome work @zarahzachz

zarahzachz:
This PR is a breaking change because we renamed several tokens for USA banner - specifically around the `-lock-icon__color` (now named "domain" instead of "lock").

Renaming or removing a token creates a major version change.

pwolfert:
@zarahzachz, can you list what tokens were added and removed? We could include that granular info in the release notes directly or at least have it here in the PR description if people follow it in the release notes. We've gotten feedback a few times about having more granular info about changes like this. `-close__background-color` I think was removed (renamed), etc.

PR #2585, opened 2023-07-13
[WNMGDS-2432] Convert the FormLabel component to a functional component

pwolfert:
## Summary

[WNMGDS-2432](https://jira.cms.gov/browse/WNMGDS-2432)

Converted the FormLabel component to a functional component before making other changes in a follow-up PR


packages/design-system/src/components/FormLabel/FormLabel.tsx
```diff
@@ -54,99 +54,88 @@ export interface FormLabelProps {
...
     );
   }
-}
+
+  const ComponentType = component;
+  const classes = classNames('ds-c-label', className, inversed && 'ds-c-label--inverse');
+
+  return (
+    <ComponentType className={classes} htmlFor={fieldId} id={id} {...labelProps}>
+      <span className={classNames(textClassName)}>{children}</span>
```
zarahzachz, line 132: wondering if there's extra optimization we can do here to remove unused tags...

if a label doesn't have a hint or error, the following will render:

```html
<label><span class="optional-class">LABEL TEXT!</span></label>
```

i know the [dom size can impact performance](https://web.dev/dom-size-and-interactivity/#how-can-i-reduce-dom-size) and i wonder if there's an opportunity here to logically say, if no error or hint prop passed, render `<label class="optional-class">LABEL TEXT!</label>` instead?

zarahzachz (approved):
this looks good. 

added comment for additional optimization if we wanna go that route. could be a different ticket if we need it to be.

packages/design-system/src/components/FormLabel/FormLabel.tsx
pwolfert: Yeah, we could actually remove the `textClassName` prop as well. I'll do this in the next PR

pwolfert (commented):
GitHub is acting weird and making me post a review with words in the comment in order to reply to your above thread

PR #2593, opened 2023-07-26
[WNMGDS-2451] Add ability for keyboard scroll in dropdown list

zarahzachz:
WNMGDS-2451

Adding ability for keyboard scrolling in Dropdown


pwolfert (approved):
Nice find, and confirmed it does fix it!

pwolfert:
Looks like the CSS changes affected how long the autocomplete list can be according to the VRTs

PR #2597, opened 2023-07-26
[WNMGDS-2432] Move hint text out of label element

pwolfert:
## Summary

WNMGDS-2432

- Moved both hint text and error messages out of the `label` element and instead linked them to the input via the `aria-describedby` attribute
  - Bottom-placed error messages had previously been linked that way, so there are no changes to bottom-placed error messages
- Added a `FieldHint` component because of the new independent nature of hints
- Renamed the `InlineError` to `FieldError` to match the new `FieldHint` component. `InlineError` is now an alias for `FieldError`
  - I think we should also rename `FormLabel` to `FieldLabel`. It's not labeling a form; it's labeling a field.
- Changed the margins of medicare labels as a side-effect. See c3e2d096fed4d80c333bf818d54786ce5e1cda7e for more details
- Updated visual regression test references to account for changes to medicare label spacing as well as slight changes to spacing between label and hint text
- Marked the FormLabel's `textClassName` prop as deprecated. It was used to distinguish between classes applied to the label text and the whole label element, which would also contain the hint and error before. There is no such distinction anymore because the label only contains the label
- Created a new internal utility function called [`mergeIds`](f83e6950e4191af3910ce94f325976e5609788ab), which is an alias of the `classNames` function

## How to test

## Checklist


packages/design-system/src/components/ChoiceList/Choice.test.tsx
```diff
@@ -229,5 +229,26 @@ describe('Choice', () => {
       expect(screen.getByTestId('checked').textContent).toBe('I am checked');
       expect(screen.queryByTestId('unchecked')).toBeNull();
     });
+
+    it('applies correct aria attributes when checkedChildren is set', () => {
```
pwolfert, line 233: Moved these tests from `ChoiceList` tests

packages/design-system/src/components/ChoiceList/ChoiceList.test.tsx
```diff
@@ -81,40 +81,11 @@ describe('ChoiceList', () => {
       expect(fieldsetEl).toBeDefined();
     });
 
-    it('allows for modification of aria attributes', () => {
```
pwolfert, line 84: These were moved to the `Choice` tests

packages/design-system/src/components/ChoiceList/Choice.tsx
```diff
@@ -219,10 +226,22 @@ export class Choice extends React.PureComponent<
...
+        </FieldError>
+      );
+    }
+
     // Remove props we have our own implementations for
     if (inputProps.id) delete inputProps.id;
     if (inputProps.onChange) delete inputProps.onChange;
 
+    inputProps['aria-describedby'] =
+      mergeIds(this.errorId, this.hintId, inputProps['aria-describedby']) || undefined;
```
pwolfert, line 243: All the fields that have multiple inputs (ChoiceList, MonthPicker, MultiInputDateField) need to pass down the root `aria-describedby` attributes for the root hint and error messages down to the individual inputs. Dan has tested this

zarahzachz: This comment might be worth putting into the code as an FYI

packages/design-system/src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -71,24 +71,39 @@ describe('Dropdown', () => {
...
     const button = screen.getByRole('combobox');
     expect(button).toHaveAttribute('aria-invalid', 'true');
-    expect(button).toHaveAttribute('aria-describedby', '1_error');
+    expect(button).toHaveAttribute('aria-describedby', errorId);
     expect(button).toHaveClass('ds-c-field--error');
-    expect(container).toMatchSnapshot();
+
+    const dropdown = container.querySelector('.ds-c-dropdown');
+    const error = container.querySelector(`#${errorId}`);
+    expect(dropdown.lastChild).toEqual(error);
```
zarahzachz, line 106: clever way to check for bottom placed error!

tests/browser/snapshots/autocomplete--type--medicare--Mobile-Chrome.png
zarahzachz, line 1: i assume this change is from the mgov css overrides mucking with the label text?

tests/browser/snapshots/autocomplete--type--medicare--chromium-forced-colors.png
zarahzachz, line 1: interesting forced colors went from black to blue 🤔

packages/design-system/src/components/utilities/mergeIds.ts
zarahzachz, line 1: clever!

packages/design-system/src/styles/components/_ChoiceList.scss
```diff
@@ -9,46 +9,44 @@
...
+  column-gap: var(--choice-wrapper__gap, 0.5rem);
+  display: grid;
+  grid-template-areas:
+    'check label'
+    'check hint'
+    'check error-message';
+  grid-template-columns: min-content 1fr;
   margin-block-start: $spacer-1;
   max-width: var(--measure-wide);
+  row-gap: 0;
```
zarahzachz, line 21: i don't think this line's needed

packages/design-system/src/components/FieldError/FieldError.tsx
```diff
@@ -4,25 +4,25 @@ import { AlertCircleIcon } from '../Icons/AlertCircleIcon';
...
+export function FieldError({
   children,
   className,
   id,
   inversed,
-}: InlineErrorProps): React.ReactElement {
+}: FieldErrorProps): React.ReactElement {
   const classes = classNames(
-    'ds-c-inline-error',
+    'ds-c-inline-error', // For backwards compatibility
```
zarahzachz, line 25: is it worth adding the same deprecation comment here as well?

packages/design-system/src/components/FieldHint/FieldHint.tsx
```diff
@@ -0,0 +1,49 @@
...
+      // Add punctuation after the requirementLabel so it doesn't run into the hint
+      requirementLabel = requirementLabel + '.';
+    }
+
+    // Add space between hint and preceding requirementLabel
+    hintPadding = ' ';
+  }
+
+  return (
+    <div id={id} className={hintClasses}>
```
zarahzachz, line 43: would a `<p>` tag be more semantic than a `<div>`?

packages/design-system/src/components/FormLabel/FormLabel.tsx
```diff
@@ -69,70 +75,47 @@ export const FormLabel = (props: ComponentProps) => {
...
-    <ComponentType className={classes} htmlFor={fieldId} id={id} {...labelProps}>
-      <span className={classNames(textClassName)}>{children}</span>
+    <>
+      <ComponentType className={classes} htmlFor={fieldId} id={id} {...labelProps}>
+        {children}
+      </ComponentType>
       {hintElement}
-      {errorMessageElement}
-    </ComponentType>
+      {errorMessage}
```
zarahzachz, line 117: 🤌

packages/design-system/src/components/FormLabel/useFormLabel.tsx
```diff
@@ -78,7 +91,7 @@ export interface UseFormLabelProps extends FormFieldProps {
  *   - labelProps:   Props to be applied to a `FormLabel` (where we get the name)
  *   - fieldProps:   Props to be applied to the field (a.k.a., the input)
  *   - wrapperProps: Props to be applied to the wrapping element
- *   - bottomError:  A rendered React element representing a bottom-placed error
+ *   - errorElement:  A rendered React element representing a bottom-placed error
```
zarahzachz, line 94: the logic for `topError` might mean updating this description to say maybe:
> A rendered React element representing error placement

tests/browser/snapshots/components-choice--default-cmsgov--chromium-forced-colors.png
zarahzachz, line 1: is there a change in max-width?

zarahzachz (commented):
did a partial review - will complete tomorrow

tests/browser/snapshots/components-textfield--success-field-medicare--Mobile-Chrome.png
zarahzachz, line 1: i thought scott said something awhile back about getting rid of "success" color fields? this might need a 2nd look

tests/browser/snapshots/autocomplete--type--medicare--Mobile-Chrome.png
pwolfert: This is where the top margins of all medicare labels became slightly shorter (24px to 21px) because 1.5em of 14px instead of 16px

tests/browser/snapshots/autocomplete--type--medicare--chromium-forced-colors.png
pwolfert: I really don't know why this happened. Probably need to dig into that more.

packages/design-system/src/components/FieldError/FieldError.tsx
pwolfert: Just that we don't use that class to style anything anymore?

packages/design-system/src/components/FieldHint/FieldHint.tsx
pwolfert: Possibly, but I'm hesitant because it's not really a standalone paragraph of text. It's linked to the field, which is what gets the tab focus. Maybe we should ask Dan or Angela.

packages/design-system/src/components/FieldError/FieldError.tsx
zarahzachz: yeah. we comment in the inline error file that this class will be deprecated, but don't have any sort of TODO here to actually deprecate it.

packages/design-system/src/components/FormLabel/useFormLabel.tsx
pwolfert: Good catch

tests/browser/snapshots/components-choice--default-cmsgov--chromium-forced-colors.png
pwolfert: Ah, I put a max-width on the hint text since it no longer gets the max-width of the label, but this choice change was unintentional :/

tests/browser/snapshots/components-textfield--success-field-medicare--Mobile-Chrome.png
pwolfert: Oh, I didn't realize we still had these. I guess this is "fixed" since the medicare overrides no longer clobber this style...but maybe in another PR we should just remove these stories.

packages/design-system/src/styles/components/_ChoiceList.scss
pwolfert: You're right. This was left-over from when I had the wrapper class on the root element

pwolfert:
@zarahzachz, I just changed the target branch to `feature/label-hint-error` to take some of the pressure off for getting it perfect. I think some follow-ups are in order for
1. `p` vs `div` question
2. Investigating forced-colors change
3. Removing _success_ field stories
4. Further work on removing the `useFormLabel` abstraction and possibly renaming `FormLabel` to `FieldLabel`

zarahzachz (approved):

PR #2599, opened 2023-07-27
[WNMGDS-2427] Filter-dialog prototype for doc site theme and version switching

pwolfert:
## Summary

WNMGDS-2427

- Adds a prototype of a `FilterDialog` component that uses `NativeDialog` under the hood
- Uses the `FilterDialog` to create a new theme and version switcher for the doc site
- Fixes `useTheme` causing two renders each time it's used
- Fixes a bug where dialogs have white text when their parent has the inverse colors applied

## How to test

[Demo](https://cmsgov.github.io/design-system/branch/pwolfert/filter-dialog)

Try it out with VoiceOver in Safari

## Checklist


pwolfert:
Testing with a screen reader revealed that the automatic update that is made to the versions dropdown when the value of the theme dropdown is changed is actually writing over the announcement for the theme selection event. Downshift has one single `#a11y-status-message` element, and it renders the latest message there. So the theme updates, then immediately after, the version updates, and the version a11y message wins. Thus, all you hear when you select a theme is "11.0.3 has been selected" or whatever the version is.

brentschneider (approved):
Functionality wise this gets us there! 
(Separating the two dialogs from each other satisfies the previous a11y concerns)

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -76,12 +76,10 @@ export interface BaseDialogProps extends AnalyticsOverrideProps {
...
   /**
-   * This function needs to handles the state change of exiting (or deactivating) the modal.
-   * Maybe it's just a wrapper around `setState()`; or maybe you use some more involved
-   * Flux-inspired state management — whatever the case, this module leaves the state
-   * management up to you instead of making assumptions.
-   * That also makes it easier to create your own "close modal" buttons; because you
-   * have the function that closes the modal right there, written by you, at your disposal.
+   * Called when the user triggers an exit event, like by clicking the close
+   * button or pressing the ESC key. The parent of this component is
+   * responsible for showing or not showing the dialog, so this callback should
```
zarahzachz, line 81: might wanna remove "should" so we sound more certain of what this prop will do

packages/design-system/src/components/Drawer/Drawer.tsx
```diff
@@ -47,6 +47,12 @@ export interface DrawerProps {
    * Enables "sticky" position of Drawer footer element.
    */
   isFooterSticky?: boolean;
+  /**
+   * Called when the user activates the close button or presses the ESC key if
+   * focus trapping is enabled. The parent of this component is responsible for
+   * showing or not showing the drawer, so this callback should make that
```
zarahzachz, line 53: same comment about "should"

packages/design-system/src/components/NativeDialog/NativeDialog.tsx
```diff
@@ -26,7 +26,7 @@ interface NativeDialogProps extends Omit<DialogHTMLAttributes<HTMLElement>, 'chi
   showModal?: boolean;
 }
 
-const NativeDialog = ({
+export const NativeDialog = ({
```
zarahzachz, line 29: i didn't think we wanted to export this to users?

packages/docs/src/components/layout/FilterDialog/CloseButton.tsx
```diff
@@ -0,0 +1,31 @@
...
+  React.ComponentPropsWithRef<'button'>,
+  keyof BaseCloseButtonProps
+> &
+  BaseCloseButtonProps;
+
+/**
+ *
+ */
+export const CloseButton = ({ className, ...buttonAttributes }: CloseButtonProps) => (
+  <button className={classNames('ds-c-close-button', className)} {...buttonAttributes}>
```
zarahzachz, line 26: should have `type="button"` maybe?

packages/docs/src/components/layout/FilterDialog/FilterDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+   */
+  heading: string | React.ReactNode;
+  /**
+   * A unique `id` to be used on heading element to label multiple instances of FilterDialog.
+   */
+  headingId?: string;
+  /**
+   * Heading type to override default `<h3>`
+   */
+  headingLevel?: '1' | '2' | '3' | '4' | '5';
```
zarahzachz, line 34: why don't we ever include level 6 in our heading levels?

packages/docs/src/components/layout/FilterDialog/FilterDialog.tsx
```diff
@@ -0,0 +1,88 @@
...
+      id={id}
+    >
+      <div className="ds-c-filter-dialog__window" tabIndex={-1} aria-labelledby={headingId}>
+        <div className="ds-c-filter-dialog__header">
+          <Heading id={headingId} className="ds-c-filter-dialog__heading" ref={headingRef}>
+            {props.heading}
+          </Heading>
+          <CloseButton className="ds-c-filter-dialog__close" onClick={props.onExit} />
+        </div>
+        <div className={classNames('ds-c-filter-dialog__body')} tabIndex={0}>
```
zarahzachz, line 75: curious about the `tabIndex` here 🤔

zarahzachz, line 75: should the body of this filter be wrapped with a [form element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog#usage_notes)? there's a `method="dialog"` attr that could be used in this case

packages/docs/src/components/layout/SideNav/ThemeVersionSection.tsx
```diff
@@ -0,0 +1,64 @@
...
+  const [isThemeDialogOpen, setIsThemeDialogOpen] = useState(false);
+  const [isVersionDialogOpen, setIsVersionDialogOpen] = useState(false);
+  const themeButtonRef = useRef<HTMLButtonElement>();
+  const versionButtonRef = useRef<HTMLButtonElement>();
+
+  return (
+    <>
+      <div>
+        Theme:{' '}
+        <button
```
zarahzachz, line 21: `type="button"`

packages/docs/src/components/layout/FilterDialog/FilterDialog.tsx
zarahzachz: i don't think `tabindex` is needed in our modals - but might wanna ask dan

zarahzachz (approved):
approving. my comments are more clean-uppy than anything. overall functionality/appearance is good

zarahzachz:
darn it i approved it, but noticed it behaves weird when opening both switcher dialogs. the most recent dialog should appear above the older one, but i'm not seeing that.

packages/design-system/src/components/Dialog/Dialog.tsx
pwolfert: I'm prescribing what _they_ should use this prop to do. Maybe I need to make that more clear

pwolfert: > so you need to use this callback to make that happen

How about that?

packages/design-system/src/components/NativeDialog/NativeDialog.tsx
pwolfert: If you take a look at the change in `packages/design-system/src/components/index.ts`, I'm removing the export from the package but adding it to this individual module just so I can import it in the doc site package.

packages/docs/src/components/layout/FilterDialog/CloseButton.tsx
pwolfert: good idea

packages/docs/src/components/layout/FilterDialog/FilterDialog.tsx
pwolfert: heh, this is just copypasta from the dialog. Maybe because that seems excessive?

packages/docs/src/components/layout/FilterDialog/FilterDialog.tsx
pwolfert: That may be something that would be valuable for the web component version of this, but right now whoever creates the body and buttons has the power to pull whatever info they need from the dialog. I'm going to leave that out of this prototype version, but it's something to revisit later.

pwolfert: Hmm, that was copied and pasted from our Drawer component. We should probably work on making that consistent between drawers and dialogs. Removing it from here 👍

packages/docs/src/components/layout/SideNav/ThemeVersionSection.tsx
```diff
@@ -0,0 +1,67 @@
...
+        >
+          {themeDisplayName} <span className="ds-u-visibility--screen-reader">(Change theme)</span>{' '}
+          <ArrowIcon direction="down" />
+        </button>
+        {themeDialogProps.isOpen && (
+          <ThemeDialog
+            theme={theme}
+            onExit={() => {
+              themeDialogProps.closeClick();
+              themeButtonRef.current?.focus();
```
pwolfert, line 36: @zarahzachz, for some reason, I have to manually focus back on this element after closing the dialog, and I haven't gotten to the bottom of it. This could also be affecting the drawer component, but we just need the right scenario to reproduce it. I can't seem to find documentation on the html `<dialog>` element's behavior on restoring focus after closing.

PR #2602, opened 2023-07-28
[WNMGDS-2455] Reimplement hgov accordion story, fix default props for accordionitem

zarahzachz:
WNMGDS-2455

- Refactor AccordionItem to rely on defaultProps for default props - this allows theme overrides to work
- Reimplement Accordion Story for hgov theme so we can display the correct icon overrides in our docs site

### DIFF
--- a/packages/ds-healthcare-gov/src/styles/components/_Accordion.scss
+++ b/packages/ds-healthcare-gov/src/styles/components/_Accordion.scss
@@ -1,10 +1,8 @@
 @use '../layout' as *;
 
-:root {
-  --accordion-icon-color: #0071bb;
-}
-
 .ds-c-accordion {
+  --accordion-icon-color: #0071bb;
+
   .ds-c-accordion__button {
     background-color: var(--accordion__background-color);
     color: var(--accordion-button__color);
@@ -18,6 +16,7 @@
       background-color: transparent;
       color: var(--color-primary-darker);
       text-decoration: underline;
+
       .ds-c-icon--minus-circle__circle,
       .ds-c-icon--plus-circle__circle {
         fill: var(--accordion-icon-color);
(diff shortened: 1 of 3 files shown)


pwolfert (commented):
I see we still have the VRT refs ✅ 

This mostly LGTM, except I noticed that in this version and v6 we've had a bug where the icon disappears on hover. Maybe it's a quick CSS fix?

packages/ds-healthcare-gov/src/styles/components/_Accordion.scss
```diff
@@ -1,10 +1,8 @@
 @use '../layout' as *;
 
-:root {
+.ds-c-accordion {
   --accordion-icon-color: #0071bb;
```
zarahzachz, line 4: browsers didn't recognize this custom prop as having a value. moving it within the accordion class fixed that.

zarahzachz: this has been broken for some time. i don't know if this is the correct fill color hgov wants, but it seems unusual to me that they have a one-off hex for something like this (not seeing this in [available system colors](https://design.cms.gov/foundation/system-color-tokens/?theme=healthcare)).

pwolfert: Weird, yeah, given that this has been broken for a while, I don't see any harm in setting it to `--accordion__background-color` instead when we fix it.

brentschneider: Selected component interactions and links I believe use `--color-primary` the accordion should follow the same convention.  Nice find @zarahzachz 👍🏼

pwolfert (approved):
Nice work!

PR #2603, opened 2023-07-28
[WNMGDS-2286] Add the usabanner component to hgov global header

zarahzachz:
WNMGDS-2286

- Add UsaBanner component to the Header component in the Healthcare theme
    - The banner should push the page's content down
    - The flag within the banner should align with the Healthcare.gov logo in the header
- Add CSS to prevent the header's box-shadow from overlapping the banner

[Testing URL](https://cmsgov.github.io/design-system/branch/WNMGDS-2286/add-usa-banner-to-hgov-header/storybook/?path=/story/healthcare-header--logged-in-controlled-header&globals=theme:core)

## A11y questions for @SparrowSpit 

These are the ones that currently jump out at me, so there's probably more to be looked at.

1. The implementation on CMS.gov (our reference) shows banner and header as siblings, so I'm assuming that's how we should organize our components as well. **Does it make sense to keep them separate, or should the banner live within the global `<header>` tag?**
2. The banner markup was derived from USWDS, where the `<section>` tag is given an `aria-label` and not the `<header>` within. However, [the accessibility guidance](https://designsystem.digital.gov/components/banner/#accessibility) alludes to the label being applied to the `<header>` to differentiate it from any other `<header>` tags on the page. **Should the `aria-label` be moved from the `<section>`, do we need an additional `aria-label` for the nested `<header>`, or is the current implementation OK?**
3. **Does the layout shift (banner pushing all page content down when expanded) pose any issues for those with vestibular disorders?**


pwolfert (approved):
This looks great!

brentschneider:
@zarahzachz overall this is looking really awesome! Loving the responsive alignment too.  
  
I was comparing our 320 mobile version to the [USWDS version of the banner](https://designsystem.digital.gov/components/banner/) and noticed they added the \[x\] close button at the smallest breakpoint \``$media-width-xs` and hid the Here's how you know link, but not at the larger breakpoints.  
  
Should we follow that same pattern?

SparrowSpit (commented):
To your questions
1. I believe I'd keep the usa banner and global <header> separate as you have them now. 
2. They've added the aria-label on the <section> to provide an accessible name to that region. If you were to add that to the <header> instead then it would lose that benefit. And no, I don't think the <header> within the usa banner would benefit from an aria-label for clarification.
4. I don't think it's enough movement to cause issues though I would think a slight transition could help since currently it's slightly jarring of a display (all or none).

zarahzachz:
> First question. The skip link is being covered by the usa banner in the test link. Is this a separate issue and so I should ignore?

not a separate issue - i'll dig in!

zarahzachz:
> I was comparing our 320 mobile version to the [USWDS version of the banner](https://designsystem.digital.gov/components/banner/) and noticed they added the [x] close button at the smallest breakpoint ``$media-width-xs` and hid the Here's how you know link, but not at the larger breakpoints.

@brentschneider yup - this is how USWDS does it too. i think this exists to increase the tap area for mobile devices? the accordion-style opener is kinda small and the X is much larger so a finger can close it easier, but that's a guess and i don't really know why this design decision was made

brentschneider:
> _this is how USWDS does it too_

Here's a screen shot of the USWDS (blue focus) and ours (red focus) highlighting the difference I'm seeing between the two.   
  
I'm advocating to align to USWDSs interactions as close as we can get.

zarahzachz:
@SparrowSpit i updated the demo link with the fix for the skip nav issue.

@brentschneider i updated [the usa banner code](https://cmsgov.github.io/design-system/branch/WNMGDS-2286/add-usa-banner-to-hgov-header/storybook/?path=/story/components-usabanner--default&globals=theme:core) to hide the action text on mobile when the banner is open - this change is only reflected in the usa banner story and **not** the hgov header as hgov is pulling in `"@cmsgov/design-system": "7.0.2"`, which doesn't have these design changes. hgov will get these changes after its released in core.

SparrowSpit:
@zarahzachz skip link issue resolved! Only other issue I experience now is that the icons don't scale as I zoom but I don't know where that fits in to this work.

zarahzachz:
> Only other issue I experience now is that the icons don't scale as I zoom but I don't know where that fits in to this work.

this appears to be a safari-only issue 🤔 i'm going to dig into this, but it's probably an issue with icon zooming overall and i don't think it's specific to this ticket (icons in this work are `1em` and should scale with the font-size declared on the parent)

PR #2604, opened 2023-07-28
[NO-TICKET] Fix passthrough of `getA11yStatusMessage` on dropdowns

pwolfert:
## Summary

Because we have the `getA11yStatusMessage` prop defined, our users expect to be able to define their own function. This prop came from `Autocomplete`, but the way we use it there is slightly different. It would also be reasonable to remove this prop, but that would be a breaking change. The discovery of this missing prop passthrough came from my work on https://github.com/CMSgov/design-system/pull/2599

- Fixes Dropdown's `getA11yStatusMessage` prop not being passed down to Downshift
- Adds prop type definition for `getA11ySelectionMessage` and passes it down to Downshift

## How to test

1. Check out 405fc0704fc49350ce235230d63a66b4d1f5a677
2. Go to the default dropdown story
3. Check the console as you interact with the dropdown

## Checklist


zarahzachz (approved):
this looks good

PR #2610, opened 2023-08-02
[NO-TICKET] Restrict list normalization rules

pwolfert:
## Summary

See the notes in the CSS files changed, but this is mostly about reining back the `ul`/`ol` rules.

- Only apply default list styles to `ul`/`ol` elements that don't have a custom `role` attribute on them
- Update some component styles that either had to undo these default styles or relied on those base styles when they weren't being styled as a list

## Checklist


packages/ds-healthcare-gov/src/styles/components/_Footer.scss
```diff
@@ -15,9 +15,9 @@
   padding: var(--spacer-2);
 
   ul[role='list'] {
+    display: flex;
     flex-flow: row wrap;
-    margin: 0;
-    padding: 0;
+    flex-direction: column;
```
zarahzachz, line 20: you can remove the `flex-flow` rule if you're specifying a column direction (the column undoes the flex-flow)

packages/design-system/src/styles/components/_Dropdown.scss
```diff
@@ -72,7 +72,8 @@
 
 .ds-c-dropdown__menu {
   @include mixins.ds-unstyled-list;
-  gap: 0;
+  display: flex;
```
zarahzachz, line 75: these flex rules don't appear to be doing anything

tests/browser/snapshots/foundations-typography-list--all-lists-cmsgov--Mobile-Chrome.png
zarahzachz, line 1: we have the bare list class applied on the unstyled example, which specifies no padding, however because this example also doesn't have a role applied, it gets the `padding-inline-start: 2em;` rule from the reset (`ul:not([role])`). 

i wonder if specifying `ul:not([class])` or `ul:not([class="ds-u-list--bare"])` would help in this particular example?

tests/browser/snapshots/reset-lists--lists-cmsgov--Mobile-Chrome.png
zarahzachz, line 1: the nested `li` in these examples appear to have wider top margins than before. makes me think they aren't using the `gap` solution.

also, unstyled example doesn't have any gap between `li`. not sure if this is the desired effect.

packages/ds-healthcare-gov/src/styles/components/_Footer.scss
pwolfert: Huh, I wonder why I added that. I thought I had tested and fixed the footer. Removing

pwolfert: It appears that I still need `flex-flow` since I removed the flex direction. Not sure how that works, but removing it broke it

tests/browser/snapshots/reset-lists--lists-cmsgov--Mobile-Chrome.png
pwolfert: Regarding the missing gap between unstyled items, I'm leaving a difference between a `ds-c-list--bare` list and a list that doesn't get the reset styles applied. I don't want to mess with users' custom lists if they haven't opted into using one of our classes (or if it's not a custom list).

Regarding the nested list items, it looks like the `ol:not([role]), ul:not([role])` selector is taking precedence over `li > ul, li > ol`

pwolfert: Fixing the specificity ⏳

pwolfert:
According to the VRT results I got locally, I might need to track down a regression on the healthcare header

zarahzachz (approved):

PR #2617, opened 2023-08-09
[NO-TICKET] Fix the USA banner content width on doc site

pwolfert:
## Summary

Use the new design token to fix the width of the USA banner on the doc site. Its content was getting constrained to the "site max width", but our doc site is full screen.

## How to test

`yarn start`

## Checklist

### DIFF
--- a/packages/docs/src/styles/pages/layout.scss
+++ b/packages/docs/src/styles/pages/layout.scss
@@ -2,6 +2,10 @@ body {
   margin: 0;
 }
 
+:root {
+  --usa-banner__max-width: unset;
+}
+
 .full-height {
   // viewport height minus the USA banner at the top
   height: calc(100vh - 22px);


packages/docs/src/styles/pages/layout.scss
```diff
@@ -2,6 +2,10 @@ body {
   margin: 0;
 }
 
+:root {
```
zarahzachz, line 5: i think you need to include `:root, ::before, ::after, ::backdrop`. tried building this locally and couldn't see this change reflected. adding these additional pseudos helped (but why do we define them on all these different pseudos to begin with?).

i think because you didn't include these, your overwrite was being overwritten by them - if that makes sense

pwolfert: ahh, weird. I didn't think the other pseudos applied to the scenario in which I was using the variable, so I didn't think it would be affected

pwolfert: It works for me 😆 . I'll update it though

pwolfert: Hmm, it doesn't work at all for me now. It must be the order in which stylesheets are loaded.

zarahzachz: i wondered about that as well.

zarahzachz: i'd think our docs site should load after any theme stuff though...

zarahzachz (approved):

PR #2619, opened 2023-08-09
[WNMGDS-2476] icon performance evaluation

zarahzachz:
WNMGDS-2476

Preparing the experiment required commenting out the exports of any component using an icon. This sets the size of /dist in core at 7.9MB.

## Icons as React components

Compared 30 React icon components (+ SvgIcon wrapper component) vs 10 React icon components (+ SvgIcon wrapper component). 

30 icons - 8.4MB
10 icons - 7.9MB

## Icons as SVG Spritesheet

Compared one spritesheet with 30 icons vs another spritesheet with 10 icons.

30 icon spritesheet -
10 icon spritesheet - 

## Icons as CSS

Compared 30 CSS icon classes vs 10 CSS icon classes.

30 CSS classes - 
10 CSS classes - 

## Icons via FontAwesome

TBD


zarahzachz:
Some stuff I forgot to mention!

[On loading data urls](https://css-tricks.com/data-uris/)...
> …when measuring the performance of hundreds of thousands of mobile page views, that loading images using a data URI is on average 6x slower than using a binary source link such as an img tag with an src attribute!

The CSS solution isn't just stuck in CSS-land. Being a CSS var/custom property, it's modifiable by JS. So teams could query a specific icon token within a given framework's component then use that token in, say an image `src` attr (something like `<img src={queriedIcon}`). This could be a great way to apply accessibility (or render them decorative only) with the `alt` attr an image provides.

PR #2630, opened 2023-08-14
[NO-TICKET] add list role to hgov header lists

zarahzachz:
`main` nightly build was failing due to broken VRT for logged out hgov header. I updated the VRT in this branch and merging this should result in a successful `main` build.

**update:**
vrt issues stemmed from reset rollback work and adding `role="list"` to unstyled lists fixed the broken VRT. not sure how that PR got merged with broken VRT but here we are lol


zarahzachz:
your hunch was right. the reset work is what resulted in these changes - there's a margin-top added to the list of links in the logged out header.
Screenshot 2023-08-15 at 9 57 52 AM

i'm noticing another spacing change uncaught by VRT as well. the open menu on the header has a lot more spacing added to it. 
Screenshot 2023-08-15 at 10 01 01 AM
Screenshot 2023-08-15 at 10 01 05 AM

but... i think these are good bugs because we should have the `role="list"` applied and don't, so these visual bugs have alerted us to a11y issued 🎉

pwolfert (approved):
Thanks for fixing this!

PR #2632, opened 2023-08-15
[WNMGDS-2479] Add support for both independent and grouped options at the same time in Dropdowns

pwolfert:
## Summary

WNMGDS-2479

- Adds support for both independent and grouped options at the same time in Dropdowns
- Attempts to improve screen-reader experience for grouped items (unverified). It does not work in VoiceOver, but I have yet to find a solution out there that does. It doesn't read the group names.
- Updates one of the storybook stories to be an example of a grouped dropdown with an independent default option

## How to test

## Checklist


zarahzachz (approved):
lgtm!

SparrowSpit:
- default option is counted as an option. Not sure if anything can be done about this but noting it's a little confusing that you hear "X of 7" when it's really X of 6. I realize that this is because the default is a list item with a role and, again, may not be able to be changed. 
- JAWS & NVDA (with Edge and Chrome) does read the Groups, i.e. “Group A grouping”, when you enter a group navigating with the arrow keys
- JAWS & NVDA, if listbox is open, ESC key has to be pressed twice to close. Not sure if anything can be done about this but just noting. 
- In JAWS there is a lot of verbosity. Maybe it's just a setting I have in my instance of JAWS. [See a screen grab of testing with JAWS to see the repeated options](https://drive.google.com/file/d/1DqrUgvbrtDAi3YSRKzjiNVtQqjHVFGDb/view?usp=sharing). 

Tested with 
- Andi, Axe, WAVE
- VoiceOver+Safari, JAWS & NVDA+Chrome and Edge

I haven't had a chance to test on my phone but hope to at some point.

pwolfert:
Thank you for testing this, @SparrowSpit! Do you know if the double-ESC problem existed [before this branch](https://design.cms.gov/storybook/?path=/docs/components-dropdown--docs&globals=theme:core)? I would like to know if my changes introduced that.

pwolfert:
If the grouping and the JAWS verbosity was better in the old version, I can revert all my markup changes. What I changed was putting grouped items _inside_ a group item, rather than just having a group label be another sibling element of all the other items.

Before (simplified):
```html
<ul role="listbox">
  <li role="group" disabled="true">Fruit</li>
  <li role="option">Apple</li>
  <li role="option">Banana</li>
</ul>
```

After (simplified): 
```html
<ul role="listbox">
  <li role="group" aria-labelledby="group-1">
    <div id="group-1" role="presentation">Fruit</div>
    <ul role="presentational">
      <li role="option">Apple</li>
      <li role="option">Banana</li>
    </ul>
  </li>
</ul>
```

SparrowSpit:
- Apparently the Esc thing was on the previous demo too. Certainly something you could revisit after the patch since it's still very usable.
- I just retested [this older version](https://design.cms.gov/storybook/?path=/story/components-dropdown--option-groups&globals=theme:core) versus this newer version and I'm not sure what we'd gain by reverting. Having the grouping read out is a plus which we weren't getting before. Thoughts?

pwolfert:
I know VoiceOver doesn't read the groups in either version, so I wasn't sure if that's something we gained between versions in other screen readers. Sounds like the answer is that we did gain that functionality, so I'm going to merge as is, and we'll revisit improving it later. Thank you, @SparrowSpit!

PR #2633, opened 2023-08-16
[WNMGDS-2369] Derive all generated ids from a standard `id` prop

pwolfert:
## Summary

WNMGDS-2369

In order to give product teams control over the generated ids, we should standardize all of our components that generate element ids to have at least a single, optional root id prop that can be defined by our users that will stand as the basis for all internal generated ids, instead of relying on our id-generating code, which may not be practically deterministic in all environments (like multi-threaded static-site generators whose content comes from fetched data).

- All components that generate ids now have at least an optional `id` prop
- Components that had props for specific element ids now derive those ids from the `id` prop or generated root id when not provided by prefixing. For example, a generated alert heading would look like `${id}__heading`, where `id` is either the value of the `id` prop or a generated id.
- Unit tests have been updated to use a static id where possible to avoid unintentional snapshot changes, which just become noise in our pull requests

Several smaller PRs will be merged into this one as we address the 25 component changes in chunks.

## How to test

Take a look at the unit tests. Ids can also be added in storybook.

## Checklist


packages/design-system/src/components/Autocomplete/__snapshots__/Autocomplete.test.tsx.snap
```diff
@@ -70,15 +70,15 @@ Array [
   <li
     aria-selected="false"
     class="ds-c-autocomplete__menu-item"
-    id="downshift-3-item-0"
+    id="downshift-4-item-0"
```
zarahzachz, line 73: this is interesting... looks like old id assigning behavior?

zarahzachz (approved):
this looks good aside from one snapshot i had a question on.

packages/design-system/src/components/Autocomplete/__snapshots__/Autocomplete.test.tsx.snap
pwolfert: Oh, hmm, at first I thought these ones were fine because they seemed to be based on the item index, but now I see that the root of it ("downshift-X") changed and is not deterministic.

pwolfert: I'll try to fix this

PR #2638, opened 2023-08-17
[WNMGDS-2473] add color contrast info for all colory things

zarahzachz:
WNMGDS-2473

- create a color contrast text snippet to be reused in several main docs pages
- add reusable text component to the bottom of the following pages:
    - /foundation/theme-colors
    - /foundation/system-color-tokens
    - /utilities/background-color
    - /utilities/text/text-color
- remove inaccurate a11y info from /text-color

### DIFF
--- a/packages/docs/content/utilities/text/text-color.mdx
+++ b/packages/docs/content/utilities/text/text-color.mdx
@@ -2,9 +2,9 @@
 title: Text Color
 ---
 
-Use the text color utility to change the default text color of an element.
+import ColorContrastGuidelines from '../../../src/reusable-text-snippets/ColorContrastGuidelines.mdx';
 
-The options below offer color palette combinations that fall within the range of Section 508 compliant and WCAG 2.0 level AA standards which is a minimum color contrast ratio of 4.5:1.
+Use the text color utility to change the default text color of an element.
 
 **Format**: `.ds-u-color--[COLOR NAME]-[SHADE]`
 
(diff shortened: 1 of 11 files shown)


packages/docs/src/reusable-text-snippets/ColorContrastGuidelines.mdx
```diff
@@ -0,0 +1,21 @@
+## Accessibility considerations
```
zarahzachz, line 1: i'm not sure where this text component should live, so i'm open to ideas on location and component/dir name as i _think_ this is the first time we're reusing mdx in this way.

packages/docs/content/utilities/text/text-color.mdx
```diff
@@ -2,9 +2,9 @@
 title: Text Color
 ---
 
-Use the text color utility to change the default text color of an element.
+import ColorContrastGuidelines from '../../../src/reusable-text-snippets/ColorContrastGuidelines.mdx';
 
-The options below offer color palette combinations that fall within the range of Section 508 compliant and WCAG 2.0 level AA standards which is a minimum color contrast ratio of 4.5:1.
+Use the text color utility to change the default text color of an element.
```
zarahzachz, line 7: removed the a11y paragraph here because its wrong. tried correcting the color pairing examples, but theming makes that difficult (each theme as a different color assigned to each utility 😵 ). 

i can roll back this change and wait until we have a ticket for fixing this page in its entirety, but it seemed like keeping bad content around due to process (waiting for a ticket/prioritization) might result in teams accidentally assuming all these pairings are gtg.

packages/docs/src/reusable-text-snippets/ColorContrastGuidelines.mdx
```diff
@@ -0,0 +1,21 @@
...
+
+Text over gradients and background images still need to meet contrast requirements.
+
+### Non-text contrast
+
+Elements that are not text, but still important, including buttons, icons that convey information, data visualizations (charts and graphs), and form inputs need a contrast ratio of at least 3:1. Also included in this are states of elements such as the selected state of an element, expanded vs. collapsed, active vs. inactive, elements with keyboard focus, etc.
+
+### Easily test your color combinations
+
+Note that the colors listed below only display colors that are available but when you are pairing colors together, make sure the color contrast ratio is sufficient. Here are three tools we suggest for testing color contrast:
```
pwolfert, line 15: > the colors listed below

They aren't below

packages/docs/content/utilities/text/text-color.mdx
pwolfert: Yeah, good call. It's a wish-list item to come up with approved background and foreground combos for each theme

pwolfert (approved):
Nice improvement 👍

PR #2640, opened 2023-08-17
[WNMGDS-2369] Id update set 2

pwolfert:
## Summary

One of several PRs into https://github.com/CMSgov/design-system/pull/2633, this PR updates the following components:

- All components that use form labels

### DIFF
--- a/packages/design-system/src/components/MonthPicker/MonthPicker.tsx
+++ b/packages/design-system/src/components/MonthPicker/MonthPicker.tsx
@@ -8,6 +8,7 @@ import classNames from 'classnames';
 import { NUM_MONTHS, getMonthNames } from './getMonthNames';
 import { fallbackLocale, getLanguage, t } from '../i18n';
 import { FormFieldProps, FormLabel, useFormLabel } from '../FormLabel';
+import useId from '../utilities/useId';
 
 const monthNumbers = (() => {
   const months = [];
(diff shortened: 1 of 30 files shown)


packages/design-system/src/components/MonthPicker/MonthPicker.tsx
```diff
@@ -169,6 +172,7 @@ export const MonthPicker = (props: MonthPickerProps) => {
                 type="checkbox"
                 value={i + 1}
                 label={month}
+                id={`${id}__choice--${i + 1}`}
```
zarahzachz, line 175: like that the id kinda aligns with the number month it is

packages/design-system/src/components/PrivacySettingsDialog/__snapshots__/PrivacySettingsTable.test.jsx.snap
```diff
@@ -398,7 +398,7 @@ Object {
             >
               <legend
                 class="ds-c-label"
-                id="field_2-label"
+                id="choice-list--2__label"
               >
                 <span
```
zarahzachz, line 403: 🤔 this empty span doing nothing looks sus (unrelated to your work). should probably be removed in another ticket?

zarahzachz (approved):

packages/design-system/src/components/PrivacySettingsDialog/__snapshots__/PrivacySettingsTable.test.jsx.snap
pwolfert: I actually think this might be fixed in the hint text updates that are currently in a feature branch

PR #2644, opened 2023-08-21
[WNMGDS-2369] Standardize id prop for components S-V

zarahzachz:
WNMGDS-2369

Update ids used on:
- SvgIcon
- Table
- Tooltip
- UsaBanner
- VerticalNavItem

PR #2648, opened 2023-08-22
[WNMGDS-2369] Remove icon ids that are now optional

pwolfert:
## Summary

Remove icon ids that are now optional as of https://github.com/CMSgov/design-system/pull/2644


zarahzachz (approved):
🤌 deleted code is my favorite!

PR #2649, opened 2023-08-23
[WNMGDS-2486] Fix high specificity of list normalization rules and fix bare list styles

pwolfert:
## Summary

WNMGDS-2486

- Use `:where` in our `ol:not([role])` selector so our reset styles don't have a greater specificity than application-applied classes!
- Modified the `.ds-c-list--bare` rule so "bare" lists get the base normalization styles. We actually want to let teams apply a `role="list"` to `ds-c-list--bare` lists, which are a little bit different historically from completely unstyled lists. Historically, we've applied the standard item spacing to bare lists vs completely unstyled lists.
- Revert the look of `ds-c-list--bare` styles to what they were prior to [this change introduced in 7.0.5](https://github.com/CMSgov/design-system/pull/2610/) where bare lists have no additional left padding
- Update our stories and documentation, including removal of redundant stories

## How to test

1. Lists with both `.ds-c-list--bare` and `role="list"` should behave as bare lists did in previous major versions
2. You should be able to apply custom classes to elements that override our list normalization rules with only the specificity that comes from that single class

## Checklist


pwolfert:
This fix was verified in one of the apps!

PR #2655, opened 2023-08-24
[RELEASE] 7/24 Version bump main

pwolfert:
Please review version updates and compare against latest release branch

PR #2656, opened 2023-08-24
[NO-TICKET] New, interactive release script

pwolfert:
## Summary

- Rewrites our release script in TypeScript
- It now interactively guides us through the release process and includes bumping the main branch and drafting release notes
- Upgrade Lerna from v3.20 to v5.6

## How to test

1. Check out the `test/release-branch` branch, which includes these changes plus some fake changes to the core package so that lerna will allow a version bump
2. Run `yarn release`. Go ahead and do everything it says. You can choose to draft the release notes but not publish the draft in the `Do these notes look OK to publish as a Draft? (Y/n):` step.
3. After that, undo the release with `yarn release --undo`

Note that [it automatically created this pr when I ran it](https://github.com/CMSgov/design-system/pull/2655)


scripts/release.ts
```diff
@@ -0,0 +1,192 @@
...
+    message: `Would you like to create a pull request to bump versions on ${c.cyan('main')}?`,
+  });
+  if (!yes) {
+    console.log(c.green('Skipping version-bump pull request.'));
+    return;
+  }
+  console.log(c.green(`Creating a version-bump branch to merge into ${c.cyan('main')}...`));
+
+  const d = new Date().getDate();
+  const m = new Date().getMonth();
```
zarahzachz, line 112: do months start at 0? my PR says it was created last month

zarahzachz: [yup!](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/getMonth)

> zero-based value (where zero indicates the first month of the year)

pwolfert: doh! I also don't like how verbose it is. The original script used a shell command to format the date, and I'm tempted to do that here since it's one line and doesn't require a dependency like `moment`. Why is date formatting hard in JavaScript?

pwolfert: Good catch

PR #2663, opened 2023-08-30
[WNMGDS-2488][WNMGDS-2489] Swap `Downshift` for `react-aria`

pwolfert:
## Summary

Converts the Dropdown component from a [Downshift](https://www.downshift-js.com/)-based implementation to a [react-aria](https://react-spectrum.adobe.com/react-aria/useSelect.html)-based implementation, which has several important accessibility advantages. The react-aria implementation of this ARIA pattern uses focus as the mechanism for highlighting dropdown options rather than `aria-activedescendent`, which notably isn't supported by VoiceOver. This implementation should give us broader screen-reader support while also giving us native HTML form support for free.

- Creates a new, internal `DropdownMenu` component along with `DropdownMenuSection` and `DropdownMenuOption`, which can be used in the `Autocomplete` component as well
    - These components are responsible for rendering [react-stately](https://react-spectrum.adobe.com/react-stately/useSelectState.html) items and sections into actual DOM elements. In order to use _Adobe Spectrum_'s suite of functions to create our dropdown, we have to define our options in terms of their state-management library
- Adds support for HTML forms. Our current Dropdown implementation will not send the selected option through a form submission. The new one has a hidden form input that keeps track of the selection in a way that a browser will know how to read at the time of form submission

Note that once this pull request is approved, I will merge subsequent pull requests into it. At least a second PR will need to be created for the conversion of `Autocomplete`.

## How to test

Try out all the Storybook stories

## Checklist


packages/design-system/src/components/Dropdown/DropdownMenu.tsx
```diff
@@ -0,0 +1,56 @@
...
+  });
+
+  return (
+    <div className={className} ref={containerRef}>
+      <DismissButton onDismiss={state.close} />
+      <ul
+        {...listBoxProps}
+        id={menuId}
+        aria-labelledby={labelId}
+        className="ds-c-dropdown__menu"
```
pwolfert, line 40: Note that when I move on to Autocomplete, I'll need to build in a method of changing the root of the class name

zarahzachz (commented):
The disabled example is still interactive (opens when clicked) and probably shouldn't. Also noticed there wasn't a "placeholder" option in the grouped options examples - don't know if this wasn't added for a reason (do we support mixed single/grouped options with this implementation) but I know there was conversation around a team needing this support.

pwolfert:
I've updated this with some feedback I got from accessibility testing. I made it so tabbing away will select the option you currently have highlighted. This matches the behavior of the old one and is the expected behavior for comboboxes. However, right now it's firing double blur events when we do this. I still need to figure out how to fix that in a way that doesn't break other things.

zarahzachz (approved):
looks good (aside from the double blur issue you're tackling)

PR #2664, opened 2023-09-05
[WNMGDS-2484] Preact testing infrastructure

zarahzachz:
WNMGDS-2484

Several unit tests fail when running Preact. Seems to be related to our controlled/uncontrolled components; lacking a default value results in wonky snaps. There are other test failures, but this seems to be the big offender.

Removed snapshots from offending tests (controlled fields) as there is a difference in how Preact/React render said components (Preact strips empty `value` attributes from DOM). Replaced snap tests with more direct querying tests and added missing value check tests where needed.


packages/design-system/src/components/DateField/DateInput.test.tsx
```diff
@@ -10,8 +10,8 @@ const defaultProps: DateInputProps = {
   dayLabel: 'Day',
   monthName: 'month',
   monthLabel: 'Month',
-  yearName: 'Year',
-  yearLabel: 'year',
```
zarahzachz, line 14: changed this to match other date field formatting

packages/design-system/src/components/TextField/useLabelMask.test.tsx
```diff
@@ -10,6 +10,7 @@ import useLabelMask, {
 } from './useLabelMask';
 import { render, screen } from '@testing-library/react';
 import { act, renderHook } from '@testing-library/react-hooks';
+// import { act, renderHook } from '@testing-library/preact-hooks';
```
zarahzachz, line 13: i'm not sure how to dynamically switch between imports. 

preact-hooks are needed to test when in preact mode and react-hooks needed for react mode.

packages/design-system/src/components/Icons/SvgIcon.tsx
```diff
@@ -74,7 +74,6 @@ export function SvgIcon({
     <svg
       aria-hidden={ariaHidden}
       className={svgClasses}
-      focusable={false}
```
zarahzachz, line 77: only needed for ie support - since we don't support ie anymore i removed it

package.json
```diff
@@ -91,6 +91,8 @@
     "@storybook/react-webpack5": "^7.1.0-alpha.1",
     "@storybook/theming": "^7.1.0-alpha.1",
     "@testing-library/jest-dom": "^5.10.1",
+    "@testing-library/preact": "^3.2.3",
```
pwolfert, line 94: It looks like we don't use this. Is that true?

packages/design-system/src/components/Autocomplete/Autocomplete.test.tsx
```diff
@@ -19,8 +19,34 @@ function makeAutocomplete(customProps = {}) {
...
+    expect(input).toHaveAttribute('aria-expanded', 'false');
+    expect(input).toHaveAttribute('role', 'combobox');
+    expect(input).toHaveAttribute('type', 'text');
+
+    const menuEl = container.querySelector('.ds-c-autocomplete__menu-container');
+    expect(menuEl).toBeInTheDocument();
+    expect(menuEl).toHaveAttribute('hidden');
+
+    const list = screen.queryByRole('listbox');
+    expect(list).not.toBeInTheDocument();
```
pwolfert, line 40: These are really thorough! One step further for `aria-controls` would be `toHaveAttribute('aria-controls', list.id)`

packages/design-system/src/components/Icons/SvgIcon.tsx
pwolfert: Nice!

packages/design-system/src/components/TextField/TextField.test.tsx
```diff
@@ -20,11 +20,46 @@ describe('TextField', function () {
   });
 
   it('renders with a label mask', () => {
-    expect(renderTextField({ labelMask: DATE_MASK }).asFragment()).toMatchSnapshot();
+    const { container } = renderTextField({ labelMask: DATE_MASK });
+
+    const label = container.querySelector('.ds-c-label');
+    expect(label).toBeInTheDocument();
```
pwolfert, line 26: For some of these labels, we could also check that their `for` attribute matches the `input.id` so we know they're properly linked. That's not something that our VRTs tell us. It's possible that that's covered by our a11y tests though.

pwolfert (approved):
This works great! A couple suggestions for other things to cover, but it's mergeable as is. Great work!

package.json
zarahzachz: i tried running `yarn remove @testing-library/preact` on it but it ended up breaking functionality for the preact-hooks library. not sure if i did the removal wrong.

PR #2665, opened 2023-09-06
[WNMGDS-2488] Autocomplete to react-aria

pwolfert:
## Summary

Part of https://github.com/CMSgov/design-system/pull/2663, this migrates the Autocomplete from Downshift to react-aria.

- In some places, I've intentionally kept behaviors the same as the old version, even when the react-aria defaults might make a bit more sense. We can revisit these in the next release.
- Deprecated several things that we don't need anymore
- Not including a state object in the callbacks might be too much of a change. However, we'd have to come up with what we think teams might need in that object and make it ourselves. I started off passing `state` because it has some similar properties to the old Downshift object, but the `state` object available in the render function is actually stale. It represents the point in time before the handler is called.
- **Note that we discovered that the keyboard interactions do not work in Preact.** I've identified a fix that can be made in `react-aria`, but exactly how to get past the problem in our project is still an issue to be worked out. Options:
    1.  Preact support could be fixed upstream in `react-aria`. The problem with this is that my fix breaks other things in their library. They do not wish to take on official support of Preact because of a long list of compatability issues.
    2. We copy their code to our project and patch it there. While this is ugly, it would be achievable.
    3. We distribute a different version of Autocomplete/Dropdown for Preact that goes back to using Downshift

## How to test


packages/design-system/src/components/Autocomplete/__snapshots__/Autocomplete.test.tsx.snap
```diff
@@ -8,6 +8,7 @@ exports[`Autocomplete renders Autocomplete component 1`] = `
     class="ds-u-clearfix"
   >
     <label
+      aria-hidden="true"
```
pwolfert, line 11: I don't know where this is coming from. It doesn't seem to be happening in Storybook. This is something I'll need to fix before merging.

packages/design-system/src/components/Autocomplete/Autocomplete.test.tsx
```diff
@@ -16,15 +16,27 @@ function makeAutocomplete(customProps = {}) {
...
 
+function open() {
+  const autocompleteField = screen.getByRole('combobox');
+  userEvent.click(autocompleteField);
+  userEvent.type(autocompleteField, 'c');
+}
+
 describe('Autocomplete', () => {
-  it('renders Autocomplete component', () => {
+  it('renders a closed Autocomplete component', () => {
```
zarahzachz, line 26: nice!

PR #2669, opened 2023-09-11
[WNMGDS-2495] Fix dropdown menus getting cut off at dialog window boundaries

pwolfert:
## Summary

WNMGDS-2495

- Make the inner wrapping element of the dialog the new "dialog window" via CSS. The whole screen is now technically taken up by the `dialog` element so that content that renders outside of it doesn't get cut off by the browser.

## How to test

1. Check out 6b9361a3ce3959a1c307faffdc15dae07101fa80 and start storybook
2. Navigate to the default dialog story and open the dropdown. Nothing should be cut off.

## Checklist


zarahzachz (approved):
this looks good - i like the "glow" the open select has on the backdrop.

you might want a pair of a11y eyes to look this over, just because you've repurposed the `dialog` element/role and i don't know if this change impacts AT users at all (i don't think it does but i'm not 100% on that).

pwolfert:
We can get some manual a11y testing on the beta

PR #2675, opened 2023-09-13
[NO-TICKET] Fix dropdown errors

pwolfert:
## Summary

- Fixes using a possibly null callback function, which honestly TypeScript should have caught 😕 
- Inlines a few small functions from react-aria's dependencies. The version of the package that it was resolving to for one of our downstream projects wasn't actually compatible. Since we're including the react-aria source anyway, I figured we can just simplify by inlining and avoid the issue. None of this is a long-term fix.

### DIFF
--- a/packages/design-system/src/components/react-aria/index.js
+++ b/packages/design-system/src/components/react-aria/index.js
@@ -1,12 +1,58 @@
 import $73SJx$react, { useState, useRef, useEffect, useCallback, useMemo, useContext } from 'react';
 import $12uGp$clsx from 'clsx';
-import { _ as _$1 } from '@swc/helpers/_/_class_private_field_get';
-import { _ as _$2 } from '@swc/helpers/_/_class_private_field_init';
-import { _ } from '@swc/helpers/_/_class_private_field_set';
 import { LocalizedStringFormatter, LocalizedStringDictionary } from '@internationalized/string';
 import '@internationalized/date';
 import '@internationalized/number';
 import 'react-dom';
+
+// Inlining the following:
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/react-aria/index.js
```diff
@@ -1,13 +1,56 @@
 import $73SJx$react, { useState, useRef, useEffect, useCallback, useMemo, useContext } from 'react';
 import $12uGp$clsx from 'clsx';
-import { _ as _$1 } from '@swc/helpers/_/_class_private_field_get';
-import { _ as _$2 } from '@swc/helpers/_/_class_private_field_init';
-import { _ } from '@swc/helpers/_/_class_private_field_set';
```
seesleestak, line 5: If this is the only usage, should the dependency be removed rather than commented out? Or is the intention to reinstate them at some point?

seesleestak: Ah, sorry, I think I'm turned around from looking through various package.json's yesterday and assumed you guys had `@swc/helpers` as a direct dependency rather than a child dependency of `react-aria`.
Although, if the code does directly leverage the package _should_ it be a direct dependency? Perhaps the fact that it was being consumed as a dependency of a dependency was a cause for npm's difficulty to resolve the correct package version for us 🤷‍♂️

pwolfert: That's possible. I copied the source of `react-aria` into here so I could make a fix for Preact that they couldn't accept. I made that decision because of the short timeline we're facing for the release, but I want to keep the direct dependencies lean.

zarahzachz (approved):

PR #2682, opened 2023-09-18
[WNMGDS-2497] Web components unit tests

zarahzachz:
WNMGDS-2497

Initialize unit tests for web components.

### Testing

To run unit tests: `yarn test:unit`
To run Preact tests: `yarn test:unit:preact`
To run web component tests: `yarn test:unit:wc`

Web component tests should only cover like 12 tests and be excluded from the other testing scripts and vice versa


packages/design-system/src/components/Alert/Alert.test.tsx
```diff
@@ -4,19 +4,31 @@ import { UtagContainer } from '../analytics';
...
+import register from 'preact-custom-element';
+register(Alert, 'ds-alert');
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      'ds-alert': AlertProps;
+    }
+  }
+}
+
```
zarahzachz, line 16: we only need this import,`register()`, and TS interface for testing web components. tried modifying `testMatch` in jest.config and couldn't get this kind of dynamic import working...

packages/design-system/src/components/Alert/Alert.test.tsx
```diff
@@ -4,19 +4,31 @@ import { UtagContainer } from '../analytics';
...
 const defaultText = 'Ruhroh';
 
 function renderAlert(props: AlertProps = {}) {
   // eslint-disable-next-line react/no-children-prop
-  return render(<Alert children={defaultText} {...props} />);
+  // return render(<Alert children={defaultText} {...props} />);
+  return render(<ds-alert {...props}>{defaultText}</ds-alert>);
```
zarahzachz, line 22: to test the web component, we need to use the web component and not the Preact/React one.

packages/design-system/src/components/Alert/Alert.test.tsx
```diff
@@ -4,19 +4,31 @@ import { UtagContainer } from '../analytics';
...
 }
 
 function expectHasClass(className: string) {
   expect(screen.getByRole('region').className).toContain(className);
 }
 
 describe('Alert', function () {
-  it('renders alert', () => {
+  // WC renders <slot> in snaps
+  it.skip('renders alert', () => {
```
zarahzachz, line 31: i added `skip` to all snapshot tests just to clean up the reported errors. snapshots all render the `<slot>` tag and some research seems to indicate this is a JSDOM implementation of custom elements (and i'm finding mentions that JSDOM doesn't fully support web components. recommendation was to use [web-test-runner](https://modern-web.dev/docs/test-runner/overview/) w/ playwright or puppeteer since it uses a real browser which natively supports web components)

packages/design-system/src/components/Alert/Alert.test.tsx
```diff
@@ -41,8 +53,8 @@ describe('Alert', function () {
   it('renders additional className and role prop', () => {
     const className = 'ds-u-test';
     const role = 'alert';
-    renderAlert({ className, role });
-    const alert = screen.getByRole(role);
+    renderAlert({ 'class-name': className, role });
+    const alert = document.querySelector('.ds-c-alert');
```
zarahzachz, line 57: had to change the `getByRole` query because we're adding a native HTML attr (`role`) to the web component itself; `role` appears on both the root element and internally so this test couldn't find the right element to test

packages/design-system/src/components/Alert/Alert.test.tsx
```diff
@@ -41,8 +53,8 @@ describe('Alert', function () {
   it('renders additional className and role prop', () => {
     const className = 'ds-u-test';
     const role = 'alert';
-    renderAlert({ className, role });
-    const alert = screen.getByRole(role);
+    renderAlert({ 'class-name': className, role });
```
zarahzachz, line 56: web component props/attrs can only have lowercase and hyphenated names. our camelCase props don't work.

packages/design-system/src/components/Alert/Alert.test.tsx
```diff
@@ -63,17 +75,21 @@ describe('Alert', function () {
     expect(alert.tabIndex).toBe(-1);
   });
 
+  // experiencing similar issues to analytics hook; is it hooks WC has issues with?
   it('sets tabIndex when alertRef is passed', () => {
     const alertRef = jest.fn();
-    renderAlert({ alertRef });
+    renderAlert({ 'alert-ref': alertRef });
```
zarahzachz, line 81: i can't figure out why our fn mocks aren't working with web components 😢

packages/design-system/src/components/Alert/Alert.test.tsx
```diff
@@ -63,17 +75,21 @@ describe('Alert', function () {
...
     const alert = screen.getByRole('region');
     expect(alert.tabIndex).toBe(-1);
     expect(alertRef).toHaveBeenCalled();
   });
 
+  // adding native attrs to WC results in multiples being added throughout component
+  // seems like we need to apply native attrs using setAttribute
   it('renders additional attributes', () => {
     const ariaLabel = 'additional aria alert';
-    renderAlert({ 'aria-label': ariaLabel });
```
zarahzachz, line 76: adding additional HTML attrs resulted in strange output; saw multiples of `aria-label` and `arialabel` applied to several nodes, which made it hard for the test to query the correct element to apply the expectation. i _think_ the way to apply native attrs is to use `setAttribute` since this is a native HTML element but i'm not sure. either way, doing that seems to satisfy this test

packages/design-system/src/components/Badge/Badge.test.tsx
```diff
@@ -1,42 +1,50 @@
 import React from 'react';
-import Badge from './Badge';
+import Badge, { BadgeProps } from './Badge';
```
zarahzachz, line 2: of all the components, i had the most trouble with badge. for some reason it's rendering multiple children within and i can't figure out why this component is so different compared to the others.

"Foo" appears as a direct child of `<ds-badge>` **and** as a child of the `<span class="ds-badge">` within. i don't know what it's behaving this way.

packages/design-system/src/components/Badge/Badge.test.tsx
```diff
@@ -1,42 +1,50 @@
...
   it('should render a big badge', () => {
-    render(<Badge size="big">Foo</Badge>);
+    renderBadge({ size: 'big' });
     const badge = screen.getByText('Foo');
     expect(badge.className).toContain('ds-c-badge--big');
   });
 
-  it('should render custom classNames', () => {
-    render(<Badge className="super-special-badge">Foo</Badge>);
+  // adding class-name multiple times?
```
zarahzachz, line 43: same multiples problem. `class-name` and `classname` appear on several nodes within component and i can't figure out why.

packages/design-system/src/components/Badge/Badge.test.tsx
```diff
@@ -1,42 +1,50 @@
...
+import Badge, { BadgeProps } from './Badge';
 import { render, screen } from '@testing-library/react';
 
-describe('Badge', () => {
-  it('should render a default badge', () => {
-    render(<Badge>Foo</Badge>);
-    expect(screen.getByText('Foo')).toMatchSnapshot();
-  });
-
-  it('should render an info badge', () => {
```
zarahzachz, line 11: i removed many snaps for badge because our VRT should catch all the variations of badge. we just need to test that changing the `variation` prop will do its job once.

packages/design-system/src/components/Button/Button.test.tsx
```diff
@@ -67,17 +85,22 @@ describe('Button', () => {
     expect(button.classList.contains('ds-c-button--small')).toBe(true);
   });
 
-  it('applies disabled, inverse, and variation classes together', () => {
+  // I think there's a naming issue with the `onDark` prop.
+  // Because it starts with `on`, it's not rendering the class like you'd expect. I think it assumes its an event handler?
+  // Other props like `isAlternate` are work as expected.
+  it.skip('applies disabled, inverse, alternate, and variation classes together', () => {
```
zarahzachz, line 91: i don't know if this is caused by the nature of web components or JSDOM testing.

packages/design-system/src/components/Button/Button.test.tsx
```diff
@@ -67,17 +85,22 @@ describe('Button', () => {
...
     renderButton({
       href: '#!',
       disabled: true,
-      onDark: true,
+      'on-dark': true,
+      'is-alternate': true,
```
zarahzachz, line 96: added `is-alternate` just to confirm we can add hyphenated props to buttons

packages/design-system/src/components/Alert/Alert.test.tsx
zarahzachz, line 56: alert was the only component to have red squiggles when hyphenating props. all other components seems to recognize hyphenating a prop meant the same thing, but alert doesn't.

packages/design-system/src/components/web-components/ds-alert.test.tsx
```diff
@@ -0,0 +1,180 @@
+import React from 'react';
+import Alert, { AlertProps } from '../Alert/Alert';
+import { UtagContainer } from '../analytics';
+import { setAlertSendsAnalytics } from '../flags';
+import { render, screen } from '@testing-library/react';
+import register from 'preact-custom-element';
+
+register(Alert, 'ds-alert');
```
zarahzachz, line 8: i tried using the `define()` from preactement and it resulted in no passing tests compared to using `register()` from preact-custom-element. not sure what's going on here, but perhaps this is something to look into with the next phase?

packages/design-system/src/components/web-components/ds-alert.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+function expectHasClass(className: string) {
+  expect(screen.getByRole('region').className).toContain(className);
+}
+
+describe('Alert', function () {
+  // WC renders <slot> in snaps
+  it('renders alert', () => {
+    renderAlert({ id: 'static-id' });
+    const alert = screen.getByRole('region');
+    expect(alert).toMatchSnapshot();
```
zarahzachz, line 34: i figured out why i didn't notice the root web component elements in our original snap tests - we're querying child elements within the root component before taking the snap. as a result, we only get the guts of a web component captured in the snaps. it's the same with our react components. i wanted to maintain some parity between the two tests, so i kept the results as is, but this does seem like something to look into for both web component and react tests imo.

tests/unit/jest.config.js
```diff
@@ -23,6 +34,7 @@ module.exports = {
     'docs/public/',
     'docs/static',
   ],
+  ...ignoreWebComponentTests,
```
zarahzachz, line 37: i tried adding this rule inside the original `testPathIgnorePatterns` array but the terminal would error out. this was the only configuration i could come up with that gave me the results i wanted.

packages/design-system/src/components/Badge/Badge.test.tsx
pwolfert: 💯

packages/design-system/src/components/Button/Button.test.tsx
```diff
@@ -67,17 +67,19 @@ describe('Button', () => {
     expect(button.classList.contains('ds-c-button--small')).toBe(true);
   });
 
-  it('applies disabled, inverse, and variation classes together', () => {
+  it('applies disabled, inverse, alternate, and variation classes together', () => {
     renderButton({
       href: '#!',
       disabled: true,
       onDark: true,
+      isAlternate: true,
```
pwolfert, line 75: 👍

packages/design-system/src/components/web-components/ds-alert.test.tsx
```diff
@@ -0,0 +1,180 @@
+import React from 'react';
+import Alert, { AlertProps } from '../Alert/Alert';
```
pwolfert, line 2: This doesn't actually import the `ds-alert.ts` module, so I'm not sure what it's doing when it renders. I tried swapping this out for `import './ds-alert';`, removing the `register` call (because `ds-alert.ts` does it), and temporarily simplifying the types that complain to `any`, and when it takes the snapshots, it has different results. I think these four modules need to switch to importing from their respective web components modules. Upside is that I don't think we'll need `preact-custom-element`.

pwolfert (changes requested):

packages/design-system/src/components/web-components/ds-alert.test.tsx
```diff
@@ -0,0 +1,180 @@
...
+  });
+
+  it('sets tabIndex when autoFocus is passed', () => {
+    renderAlert({ autoFocus: true });
+    const alert = screen.getByRole('region');
+    expect(alert.tabIndex).toBe(-1);
+  });
+
+  // experiencing similar issues to analytics hook; is it hooks WC has issues with?
+  it.skip('sets tabIndex when alertRef is passed', () => {
```
pwolfert, line 82: I think this is a great use for `skip`. It's not that it was working, and now it's not, and now we're ignoring it; it's that we haven't gotten to it yet, so the "skipped" totals in the test output serve as a good indicator of our progress.

tests/unit/jest.config.js
```diff
@@ -1,20 +1,31 @@
...
+      }
+    : {};
+
+const matchWebComponentTests = useWebComponents && {
+  testMatch: ['<rootDir>/design-system/src/components/web-components/**/*.test.[jt]s(x)?'],
+};
+
+const ignoreWebComponentTests = !useWebComponents && {
+  testPathIgnorePatterns: ['<rootDir>/design-system/src/components/web-components'],
+};
```
pwolfert, line 20: These could possibly just be combined into one object. The names are possibly confusing when you just look at the big config object below. Like does it mean that we're both "matching web component tests" and "ignoring web component tests"?

```suggestion
const conditionalWebComponentsConfig = useWebComponents ? {
  testMatch: ['<rootDir>/design-system/src/components/web-components/**/*.test.[jt]s(x)?'],
} : {
  testPathIgnorePatterns: ['<rootDir>/design-system/src/components/web-components'],
};
```

or something like this?

pwolfert (approved):
Looks great to me! Good work. One suggestion below about the config

tests/unit/jest.config.js
zarahzachz: works like a dream!

PR #2684, opened 2023-09-18
[RELEASE] 9/18 Version bump main

pwolfert:


versions.json
```diff
@@ -1,5 +1,6 @@
 {
   "design-system": [
+    "8.0.0",
     "8.0.0-beta.3",
```
zarahzachz, line 4: do we remove all these betas?

pwolfert: I'll make an update to the release script to do this automatically in the future, and I'll remove them in the release branch for now

PR #2688, opened 2023-09-20
[NO-TICKET] Release script to cull old beta versions from lists

pwolfert:
## Summary

- When we're appending new versions to the version lists for the doc site, we now also cull all old beta versions automatically

## How to test

I started writing some unit tests for this, but testing it was turning out to be more trouble than it's worth. Just release a bogus version and then roll it back with `--undo`.


scripts/release.ts
```diff
@@ -1,9 +1,9 @@
-import appendVersions from './append-versions';
 import c from 'chalk';
 import yargs from 'yargs';
-import { sh, shI, verifyGhInstalled } from './utils';
-import { hideBin } from 'yargs/helpers';
+import { updateVersions, cullBetaVersions } from './versions';
 import { confirm } from '@inquirer/prompts';
+import { hideBin } from 'yargs/helpers';
+import { sh, shI, verifyGhInstalled } from './utils';
```
zarahzachz, line 6: nice!

zarahzachz (approved):

PR #2693, opened 2023-09-25
[ NO-TICKET] show/hide guidance heading only if guidance text exists for a theme

zarahzachz:


packages/docs/content/components/icon.mdx
```diff
@@ -102,8 +102,6 @@ The following CSS variables can be overridden to customize Icon components:
 
 <ThemeContent onlyThemes={['medicare']}>
 
-## Guidance
```
zarahzachz, line 105: duplicate

packages/docs/content/foundation/typography/headings.mdx
```diff
@@ -197,8 +199,12 @@ The following CSS variables can be overridden to change the type sizes:
 - `--font-size-md`, uses `--font-size-base`
 - `--font-size-sm`
 
+<ThemeContent onlyThemes={['medicare', 'healthcare']}>
```
zarahzachz, line 202: guidance only exists for these themes, so hide this heading unless these themes are active

PR #2694, opened 2023-09-26
[WNMGDS-2514] Fix delayed receipt of items failing to open Autocomplete results list

pwolfert:
## Summary

WNMGDS-2514

The most common way our applications use Autocomplete involves fetching asynchronous data to show in the results. After the results come back, they are passed to the Autocomplete component in an array. However, in the new implementation of Autocomplete, the internal heuristics (in `react-stately`) for determining when to open the menu determine that it shouldn't be open. This overcomes that problem by manually detecting when new items come in and that the results list should be open.

## How to test

1. Check out 8e13dca0d7a90d70db7f8c39219b058b07846041 to see the problem in the default Autocomplete story
2. Check out 3abfa13115a2071f9469767d6758c82bce41c227 to see it fixed
3. To see the new unit test work, comment out the new functionality in Autocomplete and watch it fail

## Checklist

### DIFF
--- a/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from 'react';
+import React, { useRef } from 'react';
 import Button from '../Button/Button';
 import DropdownMenu from '../Dropdown/DropdownMenu';
 import classNames from 'classnames';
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -288,6 +289,16 @@ export const Autocomplete = (props: AutocompleteProps) => {
     },
   };
 
+  const oldItems = usePrevious(items);
+  useEffect(() => {
+    // If the items come in significantly later than when the user started typing,
+    // react-stately will not realize that it should be showing those results. There
+    // might be items, but `isOpen` will be false 🤦‍♂️.
+    if (state.isFocused && items && items !== oldItems && items.length !== oldItems?.length) {
```
pwolfert, line 297: Realizing that `items !== oldItems && items.length !== oldItems?.length` is redundant, and what I meant was `(items !== oldItems || items.length !== oldItems?.length)` for cases where they might be passing the same array instance...in which case I guess it would also fail the length check. I'm just going to remove the length check. Users can't expect to pass in the same instance of an array to any React component and have it cause an update.

zarahzachz (approved):

PR #2703, opened 2023-09-27
[WNMGDS-2510] Fix skipped WC unit tests

zarahzachz:
WNMGDS-2510

- Update snap tests to capture fragment; ensures the wrapping WC is included in snaps
- Alias `onDark` prop for Button to `isOnDark` to avoid event handler confusion
- Remove `ref` prop for Alert as it's not needed on WC


packages/design-system/src/components/web-components/__snapshots__/ds-usa-banner.test.tsx.snap
```diff
@@ -1,105 +1,218 @@
...
+                stroke="currentColor"
+                stroke-linecap="round"
+                stroke-width="2"
+              />
+            </svg>
+          </span>
+        </button>
+      </header>
+      <div
+        class="ds-c-usa-banner__guidance [ ds-u-leading--base ds-u-font-size--base ]"
```
pwolfert, line 114: Huh, this class name is a little weird 🤔

packages/design-system/src/components/web-components/ds-alert.ts
```diff
@@ -15,7 +15,6 @@ const attributes = [
   'analytics-parent-heading',
   'analytics-parent-type',
   'id',
-  'alert-ref',
```
pwolfert, line 18: You know, maybe we could create some special element object interface in the future to expose certain elements without having the user query for them. It'd be more work for us and might not be worth it, but the benefit to the user is they wouldn't have to know our markup, and if the markup changed, we'd still have a consistent interface. I'm thinking like...
```js
alert.alertElement.focus()
```
or
```js
textField.inputElement.focus()
```
(using focus as an arbitrary example of something you'd want to do with that element)

pwolfert (approved):
There's something weird going on in one of the class lists, but other than that, this change set LGTM

packages/design-system/src/components/web-components/__snapshots__/ds-usa-banner.test.tsx.snap
zarahzachz: oh, i'd grouped those specifically in usa banner using brackets (the brackets don't impact css like this). there's a comment in the component saying these specific classes are used to hard code font treatment across themes. this is the only component like this in our DS so i was trying to call out those 2 utility classes were together and important for that reason.

PR #2714, opened 2023-10-02
[NO-TICKET] Add default values to css custom properties to prevent FOUC

zarahzachz:
we've had this problem for awhile where elements of our docs site will appear incorrectly when switching between themes (or on first load if cache was cleared). i discovered this is caused by the way we generate theme tokens and fixed this problem.

our tokens are generated via js. when our css loads, we have some rules in place that point to a themed token (e.g., `width: var(--font-size-sm)`). when the browser loads the information, it has nothing to go on for what this example's width should be so it appears undefined until the js loads, at which point you see a flash of the incorrectly sized element now receiving its assigned theme value.

the fix for this is to add a default value to our tokens so the browser has some kinda placeholder until the js fully loads in to replace it. 

for this work, i've modified styles pertaining to the vertical nav and our css reset. why? the vertical nav and navigation stylesheets each have styles targeting the width/height of the svg used within (causing jumbo chevrons to appear). by assigning `aspect-ratio: 1` i ensure the icon remains squared in size and by adding a default css value of `.875rem` and `1rem` (values derived from the core theme), the icon size is constrained.

for the css reset adjustments, i similarly added default equivalent values from core to ensure minimal layout shifting on the site. this should positively impact all consumers now that css rules with only custom property values now have a value regardless if the js has loaded or not. so no undefined reset rules!

[demo url](https://cmsgov.github.io/design-system/branch/NO-TICKET/fix-fouc-for-vertnav-and-docs-site/?theme=core)


brentschneider (approved):
I tested the **proposed FOUT fix** across various browsers, including desktop Safari, Firefox, and mobile Safari. It is a remarkably smoother experience, great work!

Looking forward to seeing this enhancement live. 👍🏼

packages/design-system/src/styles/components/_VerticalNav.scss
```diff
@@ -98,9 +98,9 @@
   }
 
   .ds-c-icon {
+    aspect-ratio: 1;
     color: var(--vertical-nav-label-icon__color);
-    height: var(--font-size-md);
-    width: var(--font-size-md);
+    width: var(--font-size-md, 1em);
```
pwolfert, line 103: Good working getting to the bottom of what causes the flash. What I don't get is why it wouldn't have this value. The theme file with this variable is in the `<head>` of the document, and all stylesheets loaded in the head are supposed to block rendering.

zarahzachz: good point - could this be how gatsby's wired up?

when viewing the sources tab in chrome dev tools, i see 2 potentials sources for css: one inlined and tied to the url `?theme=X`, which gets loaded as the html is loaded, and another pointing to a linked stylesheet listing all css custom prop values for the same theme, which loads after the html/inline styles (attaching images below). 

i'm not sure where this logic comes into play, but our docs site seems to split our token files apart from our styles.

css custom property values for cmsgov

inlined css for cmsgov theme

pwolfert: The styles that are inlined for `?theme=cmsgov` are the universal styles imported by `docs/src/styles/index.scss`, and they don't contain any theme data. It's the equivalent of loading our `index.css`. I took a look at one of the statically built html files in our docs `public` folder after running `yarn build:docs`. Almost right after the inline styles in the `<head>` comes a `<link>` to the core theme stylesheet (which gets replaced when JavaScript loads). Since it has the core stylesheet loaded, it should have values for those variables. Could there possibly be a moment when `Helmet` _removes_ the core stylesheet and _adds_ the selected theme's stylesheet and in that moment there are no theme variables?

PR #2715, opened 2023-10-03
[WNMGDS-2498] Add web-component support to our Storybook

pwolfert:
## Summary

WNMGDS-2498

- Adds two web-component stories
- Adds the web-components folder as a side effect for the design system `package.json` file so that Storybook understands that those modules have side-effects that should not be trimmed in the tree-shaking process
- Switches to using a local copy of the `preactement` library that adds functionality for being able to set and get properties on the JS representation of the elements

## How to test

1. Run Storybook in Preact mode with `yarn storybook:preact`
2. Navigate to the web component stories and try using them

## Checklist


packages/design-system/src/components/web-components/ds-alert.stories.tsx
```diff
@@ -0,0 +1,39 @@
...
+    },
+  },
+  args: {
+    variation: 'success',
+    heading: 'Yay!',
+  },
+};
+
+const Template = (args) => (
+  <ds-alert {...args} key={JSON.stringify(args)}>
```
zarahzachz, line 29: is the JSON still needed to prevent duplicate children/slots?

packages/design-system/src/components/web-components/index.ts
```diff
@@ -12,7 +12,7 @@ export * from '../analytics';
 export * from '../flags';
 export * from '../i18n';
 
-export * from './ds-alert';
-export * from './ds-badge';
-export * from './ds-button';
-export * from './ds-usa-banner';
+import './ds-alert';
```
zarahzachz, line 15: curious why we have to import here instead of exporting

packages/design-system/src/components/web-components/ds-alert.stories.tsx
pwolfert: yeah, without keying on changed props, it won't re-render the whole thing, which means it'll add the "new" children to the previously rendered children. At least that's my theory for what's happening

packages/design-system/src/components/web-components/index.ts
pwolfert: We aren't actually exporting anything from those modules, and we just want to import them to get their side effects (registering the custom elements)

PR #2726, opened 2023-10-05
[WNMGDS-2516] add presentational role to hgov logo

zarahzachz:
WNMGDS-2516

added `role="img"` to both english and spanish hgov logos. adding this presentational role to the svg prevents the paths within from being read out individually. mgov has the same role applied for the same purpose.


packages/ds-healthcare-gov/src/components/Footer/__snapshots__/Footer.test.tsx.snap
```diff
@@ -273,6 +273,7 @@ exports[`Footer renders basic footer 1`] = `
             <svg
               aria-labelledby="hc-c-footer__logo-title"
               height="25"
+              role="img"
```
pwolfert, line 276: I'm also noticing now that this was a difference between our icon svgs and these logos. Our icons have this role 👍

pwolfert (approved):
Sounds good to me 👍

PR #2727, opened 2023-10-05
[WNMGDS-2525] Add a `ds-usa-banner` story

pwolfert:
## Summary

WNMGDS-2525

- Adds a `ds-usa-banner` story for the web component
- Maps a `root-id` attribute on the outer custom element to the internal `id` prop of the Preact component so that when the component is rendered, there won't be duplicate ids on the page

## How to test

Open the new story (`yarn storybook:preact`) and fiddle with the controls

## Checklist


packages/design-system/src/components/web-components/ds-usa-banner.tsx
```diff
@@ -0,0 +1,15 @@
+import React from 'react';
+import { define } from './preactement/define';
+import { UsaBanner, UsaBannerProps } from '../UsaBanner/index';
+
+const attributes = ['class-name', 'root-id'];
```
zarahzachz, line 5: i don't think `class-name` is actually needed. looking at the react component, this only seems to exist to give teams a place to style the wrapper element. with web components, they can achieve the same thing by applying `class` to the root element. 

`root-id` makes more sense since the needed `id` is buried deeper within the component.

pwolfert: I guess I've been thinking of this as a drop-in replacement for the react components, but this is actually a new API altogether. The fact that the `className` mapped to the `.ds-c-alert` element in the React component doesn't mean it needs to in this version. All of the special `--[state]` classes we have on `ds-c-alert*` are covered by the other attributes (like `variation` and `hide-icon`), so it's not like they need to be able to apply those themselves. I now realize we're not talking about Alert, but I guess the same principle applies.

pwolfert: Anyway, yeah, I like the idea of keeping it simpler, especially if this component doesn't require a lot of customization usually

pwolfert: It'll mean getting rid of a couple of your unit tests too, btw

zarahzachz (approved):

PR #2734, opened 2023-10-09
[NO-TICKET] Ensure steplist bullet point receives same font across browsers

zarahzachz:
- Add font inheritance style to StepList bullet to ensure consistent rendering across browsers


zarahzachz:
> @zarahzachz, should there be some VRT ref updates resulting from this change?

there is now!

pwolfert (approved):
🕵️‍♀️

PR #2729, opened 2023-10-06
[NO-TICKET] Expand token support in the sketch plugin to component colors

pwolfert:
## Summary

Updates the tokens Sketch plugin with some little improvements

- Adds support for importing colors from component variables
- Instead of replacing _all color swatches_ in a document, it merges updates with the existing colors in the document
    - That allows us to import from two different kinds of files (theme colors and theme component variables) without destroying what was already imported

## How to test

1. Uninstall the old "CMSDS Color Token Importer" plugin from Sketch if it's installed
2. Run `yarn install && yarn build` to create a new cmsds-token-importer.sketchplugin file.
3. Double-click the file generated to install it in Sketch
4. Run `yarn build themes json` from the `packages/design-system-tokens/` folder to get json files to import. The JSON files will be placed in the `dist` folder under the `packages/design-system-tokens/` folder.
5. Run the plugin from the plugins menu in Sketch and open one of the theme json files

PR #2736, opened 2023-10-09
[WNMGDS-2530] Sketch plugin: Update swatches in place

pwolfert:
## Summary

WNMGDS-2530

Building on https://github.com/CMSgov/design-system/pull/2729, I've figured out how to update all the swatches and preserve all the assignments to the old ones in the specific layer styles of the document. For instance, if we have a designed component that is referencing the `alert__background-color` for its fill style, that fill style will get updated when we switch themes with the plugin. Previously I believed the only way we could do that was searching through all the styles in the document and finding matches, but the code I have in this branch seems to work where we're just kind of updating the swatches in place and calling some API functions to update references.

Note that my previous method of replacing the document's whole `swatches` array seemed to be part of the problem.

- Theme tokens are now built into the plugin. Whatever tokens exist when the plugin is built, those tokens will be used in the various themes. Instead of selecting json files from your filesystem when you run the plugin, you select from a list of themes.
- Token values are updated "in place" so that references from layer styles are maintained
- Updated readme

## How to test

1. Pull down this branch and switch to it
2. Open a sketch document
3. Use the plugin and choose a theme
4. Create a thing that uses one of the token colors
5. Use the plugin again to change the theme
6. It should update the colors on the thing

## Checklist


zarahzachz (approved):

brentschneider:
@pwolfert could we use the Experience Dev huddle to workshop the token changes?

brentschneider (commented):
@pwolfert thanks for walking through the token changes, and where we currently are at.  
this is good from my perspective.

PR #2738, opened 2023-10-10
[WNMGDS-2521] Add custom event handling to web components

zarahzachz:
WNMGDS-2521

- Skipping analytics unit tests until solution can be agreed upon 
    - Web component define function overwrites default analytics function; need to figure out how to fix this
    - Alert analytics fire before events can be added
- Update USA banner web component snapshot (changes due to Learn's whitespace solution)
- Add functionality to define custom events on a web component


packages/design-system/src/components/web-components/preactement/define.ts
```diff
@@ -97,20 +98,26 @@ function setupElement<T>(component: ComponentFunction<T>, options: IOptions = {}
...
         },
       });
     });
 
+    // This works, but it clobbers pre-existing event handlers made via `addEventListener`
+    // events.forEach((name) => {
+    //   Object.defineProperty(CustomElement.prototype, name, {
+    //     set(v) {
+    //       if (this.__mounted) {
+    //         this.attributeChangedCallback(name, null, v);
```
pwolfert, line 115: Just had an idea that we could explore later: What if instead of trying to define the callback that they add as one of the props (which clobbers any listeners we made before) we instead added an event listener in this function? Like `bob.onChange = myCallback` results in `bob.addEventListener('ds-change', myCallback)`?

pwolfert: We could pull the `kebabCaseIt(name.replace('on', 'ds'))` logic out into its own function and then use it here in this `set` function and in the `proxyEvents` function

packages/design-system/src/components/web-components/preactement/define.ts
```diff
@@ -138,6 +145,40 @@ function setupElement<T>(component: ComponentFunction<T>, options: IOptions = {}
   };
 }
 
+function proxyEvents(props, eventNames, CustomElement) {
```
pwolfert, line 148: I think this function could use some comments now that it's working. It could use a high-level explanation of what it sets out to do. We should probably give attribution to the inspiration for this function too.

packages/design-system/src/components/web-components/preactement/define.ts
```diff
@@ -138,6 +145,40 @@ function setupElement<T>(component: ComponentFunction<T>, options: IOptions = {}
...
+    const customName = kebabCaseIt(name.replace('on', 'ds'));
+    let existingCallback = () => null;
+
+    // Don't know why `existingCallback` is being defined
+    // Why would a callback already be defined in props?
+    if (props && props[name]) {
+      existingCallback = props[name].bind({});
+    }
+
+    const customCb = (event) => {
```
pwolfert, line 161: This could also use a comment along the lines of "this is the callback function that will be passed to the preact component and will dispatch a custom event when called."

packages/design-system/src/components/web-components/preactement/define.ts
```diff
@@ -150,6 +191,8 @@ function onConnected(this: CustomElement) {
   const json = this.querySelector('[type="application/json"]');
   const data = parseJson.call(this, props || json?.innerHTML || '{}');
 
+  const eventCallbacks = proxyEvents(this.__properties, this.__options.events, this);
```
pwolfert, line 194: I also think we could name this a little better to make it clear that these are the names of the event-handler props. Maybe even just calling it `eventHandlers` would help, because that's what they're called in React.

pwolfert:
Since this is mostly about adding the event-handling code, could the PR title reflect that?

packages/design-system/src/components/web-components/__snapshots__/ds-alert.test.tsx.snap
```diff
@@ -63,9 +63,8 @@ exports[`Alert renders alert 1`] = `
     id="static-id"
   >
     <div
-      aria-labelledby="static-id__a11y-label"
+      aria-labelledby="alert--1__a11y-label"
```
pwolfert, line 66: Do you know why this changed? I don't see code that actually defines the `root-id`, but I also don't see that it was removed in this branch. Like I get why the static id isn't present, but why was it ever static before?

pwolfert (approved):
Good work! This is an exciting feature

packages/design-system/src/components/web-components/__snapshots__/ds-alert.test.tsx.snap
zarahzachz: i noticed the static id wasn't set and figured this was a fix but no clue why it changed.

PR #2743, opened 2023-10-10
[WNMGDS-2531] Fix dropdown tap-through

pwolfert:
## Summary

WNMGDS-2531

- Fixes a behavior on mobile devices where tapping an option in the dropdown menu will accidentally click any elements behind the menu

## How to test

Here's a video showing the red button's click event being triggered accidentally. It shouldn't do this in this branch. To test, check out one of the commits before 0a7354024cdf450a8a8540ad58b4f477732ac694.

https://github.com/CMSgov/design-system/assets/7595652/1da8b4a0-c384-4b49-a2f8-e8f2899d9aa5

See also: [Demo story with fix](https://cmsgov.github.io/design-system/branch/pwolfert/fix-dropdown-tap-through/storybook/?path=/story/components-dropdown--tap-through)

If you take a look at the commit messages and the Jira ticket comments, I mention that the workaround this is based on does not work well with scrolling. The first version of the workaround on the linked issue thread does not allow for scrolling at all, and the second version of the workaround will fail to catch the tap-through bug if you've recently scrolled the menu. Once we've verified that this works, I'll contribute this updated workaround back to that issue thread.

## Checklist

I think this is enough of an edge case not to warrant special tests. I think it would be hard to reproduce in jsdom, and I don't want to add a browser test for it. I'm open to reconsidering that last point though.

### DIFF
--- a/packages/design-system/src/components/Dropdown/Dropdown.stories.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.stories.tsx
@@ -8,6 +8,9 @@ import { Dialog } from '../index';
 const meta: Meta<typeof Dropdown> = {
   title: 'Components/Dropdown',
   component: Dropdown,
+  args: {
+    name: 'dropdown-field',
+  },
 };
 export default meta;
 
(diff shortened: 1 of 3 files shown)


zarahzachz (approved):

pwolfert:
Just did some additional manual testing on Autocomplete to make sure the fix worked there too. Passed 👍

PR #2752, opened 2023-10-13
[WNMGDS-2536] Fix css-variable tables being rendered inside storybook wrapper elements

pwolfert:
## Summary

WNMGDS-2536

My hypothesis was that this was a weird rehydration issue where the `section` element that wrapped the table for a `ComponentThemeOptions` component was similar enough to the `section` element that wraps storybook stories that Gatsby was mixing them up during rehydration. The table was getting placed in a `section.c-storybook-example__iframe-wrapper` element, which has a height of 200px, so everything that overflowed that was rendering over everything else further down on the page. All the table rows were being printed over all the guidance documentation! I think what triggered this weird behavior is that we have a bunch of conditional content inside `<ThemeContent>` blocks which remove certain elements (choose not to render them) when we're viewing certain themes. Since all the static html that needs to be rehydrated is rendered with only the core theme, this makes it difficult for Gatsby to match the two versions of the HTML—the one rendered at build time and the one rendered by the live React components.

It seems like that hypothesis was correct, because when I _hide_ the content but keep it in the DOM, the problem goes away. Note that you can only see the problem in a statically built site (`yarn build:docs && yarn serve:docs`); the problem doesn't exist in Gatsby's dev server.

- The `<ThemeContent>` doc-site components will now always render their children, but they will put a `hidden` attribute on content that doesn't pass the condition.

**Update: The problem doesn't actually go away. (See @zarahzachz's comment below.). I've changed this to a temporary workaround that she suggested, which is to change the order of the sections on the page.**

## How to test

1. yarn build:docs && yarn serve:docs
2. Go to http://localhost:9000/components/alert/?theme=medicare
3. On main, you'll see this: The css-variable table is being printed over all the guidance documentation!
    But on this branch everything should look normal.


zarahzachz (commented):
this seems to remove some guidance 🤔 

there's a `<div hidden>` in the dev tools. when i expand it, there's content reading

> Having icons is the preferred use for alerts. Some alerts might have a large amount of content and may need the extra space and in those cases the icon can be removed. Removing the icon should be considered a exception. Validate the design and spacing on mobile to help make this decision.

(there's other content as well...). when i search for the term "icons" in the prod docs site, I see "Display & behavior" has 4 bullets while the local version only has 3.

pwolfert:
@zarahzachz, I've reverted all other changes and re-ordered the page sections instead.

PR #2771, opened 2023-10-25
[WNMGDS-2530] Sketch plugin: Generate text styles from tokens

pwolfert:
## Summary

Building on https://github.com/CMSgov/design-system/pull/2736, this adds support for text-related tokens. The new code...
- Goes through each component's set of tokens
    - finding text tokens
    - and grouping them by the state that they represent (like "success" or "disabled")
- Converts those grouped text properties into Sketch's _shared text styles_
    - with a name that represents that component and state
    - and filling in any missing values with those from a "base" text style that [matches what our html elements would inherit](https://github.com/CMSgov/design-system/blob/pwolfert/sketch-tokens-update-text-styles/packages/design-system/src/styles/_reset.scss#L35-L38)
 - Resulting in a set of named groups of named text styles that can be used in individual components
    - And one of those groups is `typography`, which [corresponds to this set of typography tokens](https://github.com/CMSgov/design-system/blob/pwolfert/sketch-tokens-update-text-styles/packages/design-system-tokens/src/themes/core-component.ts#L437-L446)

## How to test

See previous pull request for instructions on running the plugin. New text styles will be in the text styles section of Sketch.


packages/design-system-tokens/sketch/cmsds-token-importer/src/json-token-importer.js
```diff
@@ -9,66 +9,16 @@ import medicareTheme from '../../../dist/medicare.tokens.json';
...
-  healthcare: { themeColors: healthcareTheme.color, components: healthcareComponents },
-  medicare: { themeColors: medicareTheme.color, components: medicareComponents },
-  cmsgov: { themeColors: cmsgovTheme.color, components: cmsgovComponents },
+  core: { ...coreTheme, components: coreComponents },
+  healthcare: { ...healthcareTheme, components: healthcareComponents },
+  medicare: { ...medicareTheme, components: medicareComponents },
+  cmsgov: { ...cmsgovTheme, components: cmsgovComponents },
 };
 
-function updateSwatchesFromTheme(doc, themeTokens) {
```
pwolfert, line 20: Moved all this color-related code to a new `color.js` module

pwolfert:
One problem I just noticed with the `typography` text styles is that the `typography` component tokens only contain font weights and font families. The sizes come from the theme-level `font` tokens. We need to figure out a nice way of marrying the two.

brentschneider:
Great convo in Slack today @pwolfert, I think you and @zarahzachz are on to some solid ideas for moving the needle forward on this.

pwolfert:
Created a separate ticket for this and will reopen with that one

PR #2772, opened 2023-10-27
[WNMGDS-2546] Add `ds-dropdown` web component

pwolfert:
## Summary

WNMGDS-2546

- Adds a new `<ds-dropdown>` web compoent that wraps our `Dropdown` Preact component
- Adds documentation for `ds-dropdown` and for its custom events
- Adds unit tests for `ds-dropdown`
- There's currently only one way to specify the options—through the options array—because I can't get the children to show up correctly in Preact. You can either...
    - Declare your options array as a stringified attribute or
    - Assign your options array to a property on the JavaScript element like `myElement.options = []`

## How to test

Use storybook

## Checklist


packages/design-system/src/components/web-components/ds-dropdown.stories.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+export default {
+  title: 'Web Components/Dropdown',
+  argTypes: {
+    autofocus: {
+      description: 'Sets the focus on the dropdown when it is first added to the document.',
+      control: 'boolean',
+    },
+    children: { control: false },
+    'class-name': {
```
zarahzachz, line 20: is this needed as a prop? users could add `class` to a root element instead of needing a special prop

packages/design-system/src/components/web-components/ds-dropdown.stories.tsx
```diff
@@ -0,0 +1,148 @@
...
+        },
+        'ds-blur': {
+          description: 'Dispatched whenever the dropdown loses focus.',
+        },
+      },
+    },
+  },
+};
+
+/*
```
zarahzachz, line 112: why is this commented out?

packages/design-system/src/components/web-components/ds-dropdown.test.tsx
```diff
@@ -0,0 +1,100 @@
...
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('renders disabled dropdown', () => {
+    renderDropdown({ disabled: 'true' });
+    const button = screen.getByRole('button');
+    expect(button).toHaveAttribute('disabled');
+  });
+
+  it('renders disabled', () => {
```
zarahzachz, line 31: what's this test doing that the previous one didn't?

packages/design-system/src/components/web-components/ds-dropdown.stories.tsx
pwolfert: I'm just trying to keep it consistent with what we have so far. It might not be totally necessary, but it puts the classes on the same element as the `ds-c-dropdown`. We do the same with `ds-c-alert`. Maybe we can revist this when we pilot them.

packages/design-system/src/components/web-components/ds-dropdown.stories.tsx
pwolfert: I left it in there for a future state where we can use `<option>` children to define the options. I can add a TODO comment to it.

packages/design-system/src/components/web-components/ds-dropdown.test.tsx
pwolfert: nothing to see here!
🧹💨

pwolfert: thanks for catching that

zarahzachz (approved):

PR #2775, opened 2023-10-31
[WNMGDS-2547] Fix id/aria-describedby syncing issue

zarahzachz:
WNMGDS-2547

Issue:

If user doesn't define `id` prop value, the autogenerated unique id isn't used and `aria-describedby` is removed from the tooltip trigger button.

If user does define `id` prop value, `aria-describedby` appears as expected on the tooltip trigger button.

We want `aria-describedby` to appear consistently, regardless if a user has defined the `id` prop value or not (for non-dialog tooltips).

### DIFF
--- a/packages/design-system/src/components/Tooltip/Tooltip.stories.tsx
+++ b/packages/design-system/src/components/Tooltip/Tooltip.stories.tsx
@@ -59,6 +59,24 @@ export const IconTrigger: Story = {
   },
 };
 
+export const idTest: Story = {
+  render: function Component(args) {
+    return (
+      <>
+        <p className="ds-u-margin--0 ds-u-display--inline">Tooltip with icon trigger</p>
+        <Tooltip {...(args as any)} />
+      </>
+    );
+  },
+  args: {
+    ariaLabel: 'Label describing the subject of the tooltip',
+    id: 'test-id-foo',
+    className: 'ds-c-tooltip__trigger-icon ds-u-display--inline',
+    title: 'Tooltip trigger uses <TooltipIcon> for the trigger content',
+    children: <TooltipIcon />,
+  },
+};
+
 export const InlineTrigger: Story = {
   render: function Component(args) {
     return (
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/Tooltip/Tooltip.stories.tsx
```diff
@@ -59,6 +59,24 @@ export const IconTrigger: Story = {
...
+    return (
+      <>
+        <p className="ds-u-margin--0 ds-u-display--inline">Tooltip with icon trigger</p>
+        <Tooltip {...(args as any)} />
+      </>
+    );
+  },
+  args: {
+    ariaLabel: 'Label describing the subject of the tooltip',
+    id: 'test-id-foo',
```
zarahzachz, line 73: created this story to quickly test user-created `id` values. storybook was occasionally breaking, citing some sort of async leak?

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -212,7 +212,7 @@ export const Tooltip = (props: TooltipProps) => {
       className,
       component,
       dialog,
-      id,
+      // id,
```
zarahzachz, line 215: here's where it gets interesting. 

including `id` in the list of trigger props resulted in `aria-describedby` appearing **only when the user defined an `id`** (`id` returned `undefined` if user didn't define a value, which resulted in `aria-describedby` being removed).

removing `id` from the prop list resulted in the unique `id` and user-created `id` being correctly assigned to `aria-describedby` **but resulted in adding an extra `id` attribute to the tooltip trigger itself** (obviously breaking the rule that every `id` should have a unique value - both button and tooltip have same `id` here).

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -107,7 +107,7 @@ export interface TooltipProps {
  */
 export const Tooltip = (props: TooltipProps) => {
   const popper = useRef(null);
-  const id = useId('tooltip-trigger--', props.id);
+  const contentId = useId('tooltip-trigger--', props.id);
```
zarahzachz, line 110: renaming from `id` to `contentId` fixed some confusion happening in the trigger. the trigger was trying to apply the `props.id` value to `aria-describedby` instead of the value saved in this `id` const.

pwolfert (approved):
LGTM 👍

PR #2776, opened 2023-10-31
[WNMGDS-2545] Reorganize the font tokens

pwolfert:
## Summary

WNMGDS-2545

1. Move line height and font weight settings to the theme level - Medicare already modifies the weights, and these seem pretty specific to the fonts being used. Line height values are already semantic. We want semantic on the theme level and non-semantic at the system (crayon) level.
2. Define non-semantic font size tokens at the system level, which we now use to assign to named sizes at the theme level
3. Create component variables for heading sizes
4. Create a `label` bucket in the component variables - We don't call it "form-label" anymore. This also allows us to capture some of the medicare overrides in tokens

### CSS variable changes

- The following css variables were renamed (BREAKING):
    - `--form-label__color--inverse` renamed to `--label__color--inverse`
- The following css variables were removed (BREAKING):
    - `--font-family-open-sans`
    - `--font-family-rubik`
    - `--font-family-montserrat`
    - `--font-family-bitter`
- The following css variables were added:
    - `--font-size-2xl--mobile`
    - `--font-size-3xl--mobile`
    - `--font-size-4xl--mobile`
    - `--font-size-5xl--mobile`
    - `--font-size-5xl--tablet`
    - `--typography-heading-base__font-size`
    - `--typography-heading-sm__font-size`
    - `--typography-heading-md__font-size`
    - `--typography-heading-lg__font-size`
    - `--typography-heading-xl__font-size`
    - `--typography-heading-2xl__font-size--mobile`
    - `--typography-heading-2xl__font-size`
    - `--typography-heading-3xl__font-size--mobile`
    - `--typography-heading-3xl__font-size`
    - `--typography-heading-4xl__font-size--mobile`
    - `--typography-heading-4xl__font-size`
    - `--typography-heading-5xl__font-size--mobile`
    - `--typography-heading-5xl__font-size--tablet`
    - `--typography-heading-5xl__font-size`
    - `--label__color--inverse`
    - `--label__font-family`
    - `--label__font-size`
    - `--label__font-weight`
    - `--label__letter-spacing`
    - `--label__text-transform`

## How to test

## Checklist


.storybook/static/medicare-theme.css
```diff
@@ -82,30 +82,31 @@
...
---font-family-rubik: Rubik, Seravek, 'Gill Sans Nova', Ubuntu, Calibri, 'DejaVu Sans', source-sans-pro, sans-serif;
---font-family-montserrat: Montserrat, Avenir, Corbel, 'URW Gothic', source-sans-pro, sans-serif;
---font-family-bitter: Bitter, Superclarendon, 'Bookman Old Style', 'URW Bookman', 'URW Bookman L', 'Georgia Pro', Georgia, serif;
---font-line-height-reset: 1;
---font-line-height-heading: 1.3;
---font-line-height-base: 1.5;
---font-line-height-lead: 1.7;
+--font-sans: Rubik, Seravek, 'Gill Sans Nova', Ubuntu, Calibri, 'DejaVu Sans', source-sans-pro, sans-serif;
+--font-montserrat: Montserrat, Avenir, Corbel, 'URW Gothic', source-sans-pro, sans-serif;
+--font-rubik: Rubik, Seravek, 'Gill Sans Nova', Ubuntu, Calibri, 'DejaVu Sans', source-sans-pro, sans-serif;
```
pwolfert, line 87: Note that these just moved positions in the file. They existed before

packages/design-system-tokens/src/tokens/zIndex.ts
```diff
@@ -3,13 +3,11 @@
  */
 import { to, zIndexTokens } from '../lib/types';
 
-const z = to<zIndexTokens>()({
+export const z = to<zIndexTokens>()({
   deep: -99999,
   default: 1,
   drawer: 500,
   dialog: 1000,
```
zarahzachz, line 10: just dawned on me these tokens existed. do we use `dialog` and `drawer` still or was the need for these removed by using the `<dialog>` element?

packages/design-system-tokens/src/themes/medicare.ts
```diff
@@ -104,29 +104,43 @@ export const shadow: ShadowTokens = {
...
-  'weight-bold': 600,
+const font: FontTokens = {
+  'sans':                       fontFamily['family-rubik'],
+  'montserrat':                 fontFamily['family-montserrat'],
+  'rubik':                      fontFamily['family-rubik'],
+  'size-base':                  fontSize['20'],
+  'size-sm':                    fontSize['10'],
+  'size-md':                    fontSize['20'],
+  'size-lg':                    '1.25rem',
+  'size-xl':                    '1.5rem',
```
zarahzachz, line 123: should these sizes be called out as "--desktop" since we're making a distinction between mobile and tablet? i'll point out what i mean further down with a 🍕

packages/design-system/src/styles/_base.scss
```diff
@@ -156,73 +156,74 @@ $breakpoints: (
 
 @mixin heading--5xl {
   @include heading;
-  font-size: var(--font-size-3xl);
+  font-size: var(--typography-heading-5xl__font-size--mobile);
```
zarahzachz, line 159: 🍕 in this rule, we're saying the font size starting point is --mobile, with our md media query not having a size suffix

packages/design-system/src/styles/_base.scss
```diff
@@ -156,73 +156,74 @@ $breakpoints: (
...
+  @media (min-width: $media-width-md) {
+    font-size: var(--typography-heading-2xl__font-size);
+  }
 }
 
 @mixin heading--xl {
   @include heading;
   @include heading-margins;
-  font-size: var(--font-size-xl);
+  font-size: var(--typography-heading-xl__font-size);
```
zarahzachz, line 205: 🍕 and here we're saying the starting point is the un-suffixed font size. i guess what confuses me is 

1. by excluding a suffix on some tokens, are we making an assumption that token is the "base" or "default"?
2. are un-suffixed tokens the "default" size or the desktop size (or both)?
3. the combination of our new size tokens and media query sizes is a little confusing to me (mixing "--tablet" and `$media-width-sm`, "default" and `$media-width-md`)
    1. this might be a personal thing, but when i think of medium viewports, i think tablet
    2. i wonder if this will cause issues in the future, e.g., needing size X that happens to have a token name of --tablet as the starting point font size 

tl;dr the code gets hard to reason about when combining non-suffixed and viewport-suffixed tokens with media queries using a different sizing unit (sm, md, lg, etc)

packages/design-system-tokens/src/tokens/zIndex.ts
pwolfert: It looks like _only_ `--z-drawer` and `--z-dialog` are used, but they might not even be necessary like you said. We have hard-coded z-index values in our CSS, and we're not even using these values like `--z-default` and `--z-deep`. Probably worth following up on soon.

packages/design-system/src/styles/_base.scss
pwolfert: Yeah, and naming them can't be based on the screen size name like `md` because then we'd have `size-md--md`. Naming is hard. I'm trying to capture what the CSS looks like now in tokens without modifying the tokens we already have for `--font-size-*`. My hope is that at some point in the future we're getting rid of the `--mobile` and `--tablet` variables and using consistent sizing

zarahzachz (approved):

PR #2781, opened 2023-11-02
[WNMGDS-2550] Work around React/Google-Translate incompatibility in healthcare header

pwolfert:
## Summary

WNMGDS-2550

There is a [known issue with React and Google Translate](https://github.com/facebook/react/issues/11538#issuecomment-390386520) where content formatted in a certain way that is then manipulated by Google Translate causes React to crash. This problem exists in our Healthcare Header menu button, and this implements the workaround for it.

If you look at the ticket, you'll see instructions for reproducing the errors in Storybook. Unfortunately I can't get it to 1) produce the error locally in the uncontrolled story, only the controlled story and 2) be fixed in Storybook at all, even in the statically built copy. We can reproduce it on design.cms.gov/storybook but not locally. The only environment where it's fixed is in our create-react-app project. It could be triggered by something in Storybook itself.

## How to test

1. Check out 3fda915997e0a67058a99847dc69689f025f9ac6
2. `yarn build`
3. Go to `examples/create-react-app-typescript` and make sure there is no `node_modules`. Delete it if there is
4. `yarn install` in that directory
6. In the browser, you should get a test page that has the healthcare.gov header
7. Enable Google Translate as outlined in the ticket description
8. Click on the header menu button
9. It shouldn't crash on this commit
10. Check out 58768e76fe6a1129d80a5652c625163fe159e6e3 and perform the steps above to see it broken, making sure to destroy `node_modules` in the example project

## Checklist


zarahzachz (approved):

pwolfert:
We discovered that the reason we couldn't reproduce the uncontrolled story locally in Storybook is that our Storybook on `main` uses Preact now, not React. Preact doesn't crash when faced with this issue, but unfortunately it does duplicate all the content, as shown in the screenshot below.

Partial screenshot of healthcare header showing both English and Spanish content being rendered side-by-side in the menu button and other places

PR #2777, opened 2023-10-31
[WNMGDS-2545] Sketch plugin: Generate text styles from tokens

pwolfert:
## Summary

WNMGDS-2545

This is an updated version of https://github.com/CMSgov/design-system/pull/2771

Building on https://github.com/CMSgov/design-system/pull/2736, this adds support for text-related tokens. The new code...
- Goes through each component's set of tokens
    - finding text tokens
    - and grouping them by the state that they represent (like "success" or "disabled")
- Converts those grouped text properties into Sketch's _shared text styles_
    - with a name that represents that component and state
    - and filling in any missing values with those from a "base" text style that [matches what our html elements would inherit](https://github.com/CMSgov/design-system/blob/pwolfert/sketch-tokens-update-text-styles/packages/design-system/src/styles/_reset.scss#L35-L38)
 - Resulting in a set of named groups of named text styles that can be used in individual components
    - And one of those groups is `typography`, which [corresponds to this set of typography tokens](https://github.com/CMSgov/design-system/blob/pwolfert/sketch-tokens-update-text-styles/packages/design-system-tokens/src/themes/core-component.ts#L437-L446)

## How to test

See previous pull request for instructions on running the plugin. New text styles will be in the text styles section of Sketch.

PR #2783, opened 2023-11-03
[WNMGDS-2553] Sketch plugin: Add inheritance logic for text styles

pwolfert:
## Summary

WNMGDS-2553

Adds inheritance logic for text styles based on the naming conventions of the keys. Consider the following(incomplete) heading tokens for the core theme:
```json
"typography-heading": {
    "__font-family": "inherit",
    "__font-weight": 700,
    "-sm__font-weight": 400,
    "-md__font-weight": 700,
    "-lg__font-weight": 700,
    "-sm__font-size": ".875rem",
    "-md__font-size": "1rem",
    "-lg__font-size": "1.125rem",
}
```

The code that previously generated the text styles would have created the following text styles (sticking with JSON format for consistency, even though these would be converted to Sketch's internal representation of text styles):

```json
"typography-heading": {
    "font-family": "inherit",
    "font-weight": 700,
},

"typography-heading-sm": {
    "font-weight": 400,
    "font-size": ".875rem",
},

"typography-heading-md": {
    "font-weight": 700,
    "font-size": "1rem",
},

"typography-heading-lg": {
    "font-weight": 700,
    "font-size": "1.125rem",
}
```

The problem with this—alluded to in [#2777](https://github.com/CMSgov/design-system/pull/2777)—is that assigning text styles in Sketch is _all or nothing_. You can't apply both the "typography-heading" and "typography-heading-sm" to text in Sketch, so those styles need to contain all the information needed to render that text. In this case, we want the `-sm`, `-md`, and `-lg` styles to inherit from the base styles. That's where this new logic comes in.

The new code uses the naming conventions in a group of component tokens to determine a parent style name, and the text styles can therefore inherit properties from the parent. (And technically everything inherits from a "base" text style called `defaultTextStyle`, which is equivalent to the text-related styles we apply to the `body` element in the CSS normalization rules.) The generated text styles with the new logic would look more like this:

```json
"typography-heading": {
    "font-family": "Open Sans",
    "font-weight": 700,
},

"typography-heading-sm": {
    "font-family": "Open Sans",
    "font-weight": 400,
    "font-size": ".875rem",
},

"typography-heading-md": {
    "font-family": "Open Sans",
    "font-weight": 700,
    "font-size": "1rem",
},

"typography-heading-lg": {
    "font-family": "Open Sans",
    "font-weight": 700,
    "font-size": "1.125rem",
}
```

## How to test

1. Check out this branch
2. Run `yarn install && yarn build` in your terminal, where the current directory is the root of the repository
3. Run `cd packages/design-system-tokens/sketch/cmsds-token-importer`
4.  Run `yarn install && yarn build`
5. Open a Sketch document
6. Click on the "Plugins" menu and select "CMSDS Color Token Importer" (or double-click on the `packages/design-system-tokens/sketch/cmsds-token-importer/cmsds-token-importer.sketchplugin` file to install the plugin if it's not in the menu)
7. Select a theme
8. Click on the "Show components" button and then the "Show text styles" button
9. The plugin will have created "components" and "typography" folders (lowercase). You can browse the styles there.

**This change is most noticeable if you select the "Medicare" theme and look at the "components" > "label" styles. Previously the inverse styles did not have the right font family or weight.**

## Checklist


packages/design-system-tokens/sketch/cmsds-token-importer/src/text.js
```diff
@@ -47,6 +61,10 @@ function parseFontWeight(tokenValue) {
 }
 
 function parseFontFamily(tokenValue) {
+  if (tokenValue === 'inherit') {
+    return null;
+  }
+
   // Only return the first one, and remove all quotes
   return tokenValue.split(',')[0].replaceAll('"', '').replaceAll("'", '').trim();
```
zarahzachz, line 69: does this need 2 `replaceAll`?

pwolfert: One finds double quotes around font families, and one finds single quotes. I could use one regex, I guess

zarahzachz: oh i didn't see the double vs single there 😅 disregard!

zarahzachz (approved):

PR #2784, opened 2023-11-04
[NO-TICKET] Fix up text tokens and usage

pwolfert:
## Summary

Somewhat related to https://github.com/CMSgov/design-system/pull/2776

- Removed misnamed `typography-heading-base__font-size` token, which really describes body text
- Use the theme-level font-size variables in the body-text CSS instead of the heading variables
  - I could have made variables for these under `typography-body`, but they don't differ between themes...but then again, maybe we need these for Sketch?


zarahzachz:
i actually prefer moving away from "heading" and "body" token names - at least from a theming perspective when it comes to font-size. i believe every system i've encountered does this sorta thing, from [tailwind](https://tailwindcss.com/docs/font-size) to [open props](https://open-props.style/#typography) to [utopia](https://utopia.fyi/type/calculator?c=320,18,1.2,1240,20,1.25,5,2,&s=0.75%7C0.5%7C0.25,1.5%7C2%7C3%7C4%7C6,s-l&g=s,l,xl,12). 

it's like having a color palette to choose from, only you now have a font-size palette. more flexible, more maintainable, and i think more intuitive (you don't have to remember edge cases and rules to use one size with another). this approach might also lend itself to icon sizing since its more agnostic.

pwolfert:
@zarahzachz, given the constraints of [this problem](https://github.com/CMSgov/design-system/pull/2777) where we want to be able to define the properties of heading components/elements/classes in the theme definitions, how would you structure and define this data? Some pieces of information that would need to live somewhere (CSS, theme files, and/or Sketch):
- Core has a different font weight for "small" headings
- Medicare uses the same weight for all headings
- Medicare uses a different font family for headings and body text
- Core uses the same font family for headings and body text

zarahzachz:
ok i have a clearer idea of what you're asking now.

i think using the higher up tokens (more abstract ones) to define font-sizing for body text makes sense rn because all themes share the same body font sizing. i could see the eventual need for more granular size tokens after the layout/type scale work is done, but in the meantime this solution works fine imo

zarahzachz (approved):

PR #2786, opened 2023-11-07
[WNMGDS-2551] CSS deprecation

zarahzachz:
WNMGDS-2551

- Create a SCSS file for marking CSS classes as deprecated
    - Assigns `--debugger` to `<html>` element for all themes, sets default value to `false`
    - Uses style `@container` queries to apply styles to classes if a given style parameter (in this case, the `--debugger` custom property) is true
    - When a CSS class is marked for deprecation, add it both to the deprecation SCSS file outside AND inside the `@container`
        - Adding the classes outside the query make it easy to delete from the project when the class is deemed deprecated
        - Adding the classes inside the query applies the special styles to the class when `--debugger` is `true`
    - Special style is `2px solid magenta` so it stands out across themes
- Create a React helper component to automatically reassign the `--debugger` custom property to true if application not in PROD
    - Adding this helper component to the top-level of your application will result in all deprecated classes in use being highlighted when the application is not in PROD
    - This was the only way I could think of shipping JS that queried every page load AND had knowledge of PROD mode

The React component is an optional helper if teams don't feel comfortable flipping the variable in the browser's dev tools. For those who are comfortable, they can open Chrome's dev tools (`@container` is only supported in Ch-based browsers at this time) and change the value of `--debugger` to `true`.


pwolfert (commented):
Nice system 👍, and we don't have to worry about support for newer CSS features because it's only supposed to show up on devs' machines anyway, huh?

packages/design-system/src/components/CssDebugger/CssDebugger.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+    return () => {
+      window.removeEventListener('load', handleRefresh);
+    };
+  }, []);
+
+  const handleRefresh = () => {
+    if (process.env.NODE_ENV !== 'production') {
+      document.documentElement.style.setProperty('--debugger', 'true');
+    }
```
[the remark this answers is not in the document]
zarahzachz: what do you mean by the root scope of the module?

packages/design-system/src/components/UsaBanner/UsaBanner.tsx
```diff
@@ -28,6 +28,25 @@ export interface UsaBannerProps {
  * [refer to its full documentation page](https://design.cms.gov/components/usa-banner/).
  */
 export const UsaBanner: React.FunctionComponent<UsaBannerProps> = (props: UsaBannerProps) => {
+  // Logic to see deprecated CSS in use
+  useEffect(() => {
+    window.addEventListener('load', handleRefresh);
+
+    return () => {
+      window.removeEventListener('load', handleRefresh);
+    };
```
pwolfert, line 37: I'm testing this out in storybook, and I can't get the `handleRefresh` function to execute. I think the window `load` event is firing long before we get here. I don't think we need these at all. I think you should be able to put the contents of `handleRefresh` directly in this useEffect

pwolfert: Also, in order for us not to lose it, and so we can better test it, we should probably put this into its own hook in the utilities folder like
```js
function useCssDeprecationDebugger() {
  useEffect(...)
}
```

pwolfert: If we do that, we might be able to write a unit test with jsdom where we define the `--debugger` variable and verify that it behaves differently when it's pre-defined

packages/design-system/src/components/CssDebugger/CssDebugger.tsx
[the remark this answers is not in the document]
pwolfert: Update: We talked through this and found that my idea of putting it at the root scope of the module would cause the module to have side-effects, and our core package currently doesn't have side-effects. We want to keep it that way.

pwolfert (approved):
Love it! Works like a charm. I tested it in storybook for the ability to set it to true in dev envs and ignore it if the variable was already defined in css, and I also tested it by statically building the doc site to make sure it didn't show by default in production (but I can set it in the browser!).

PR #2798, opened 2023-11-09
[WNMGDS-1922] Update prop type definitions to support passthrough attributes

pwolfert:
## Summary

The following components' prop type definitions were updated so that they allow additional attributes that will be passed through to an underlying HTML element:
- All the icon components
- `Badge`
- `SingleInputDateField`
- `Tooltip`

Note that this only satisfies the typing portion of this ticket. A follow-up PR will be made later.

## How to test

1. Icons
    - To test this, add `const addIcon = <AddIcon bbox="hello" />` to `SvgIcon.test.tsx` and run `yarn type-check` with and without the update
2. Badge, SingleInputDateField, Tootip
    - In `Badge.test.tsx`, try to add an attribute to `<Badge />` that is a valid HTML attribute but isn't in the props
    - Same kind of thing for the others

## Checklist


zarahzachz (approved):
looks good, but isn't this technically a major change? i think in the past scout imported some component interfaces and changing the names might cause problems if that's still the case

pwolfert:
That's a good call

pwolfert:
@zarahzachz, I looked at this again, and I think all the changes that have been made are backwards compatible. We haven't dropped any exported types even though we build them with intermediate `Base*` interfaces internally. That is, even though we changed what used to be `BadgeProps` to `BaseBadgeProps`, we still export a `BadgeProps` type that is composed from `BaseBadgeProps` and native HTML attributes.

PR #2802, opened 2023-11-13
[WNMGDS-2523] Break some `useFormLabel` functionality into a `useInlineError` hook

pwolfert:
## Summary

WNMGDS-2523 is about lessening the negative effects of the useFormLabel abstraction. I'm working on a feature branch that breaks `useFormLabel` up into some smaller, more focused hooks that can be mixed and matched. This is one of several pull requests to go into this feature branch.

- Creates a new `useInlineError` hook that returns...
    - `topError` and `bottomError`, only one of which will have a rendered element in it based on the error positioning
    - `errorId` for referencing by other elements
    - `invalid` boolean for applying to whatever element needs an `aria-invalid` property

## How to test

These components should be manually tested a bit, but automated tests should also cover this work.

### DIFF
--- a/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from 'react';
+import React, { useRef } from 'react';
 import Button from '../Button/Button';
 import DropdownMenu from '../Dropdown/DropdownMenu';
 import classNames from 'classnames';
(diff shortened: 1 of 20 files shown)


zarahzachz (approved):
this looks good. tested the components in storybook locally and it runs well. did notice a typeerror when setting the error message on the label story itself - this might be something addressed in another PR though

> Uncaught TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property '__' -> object with constructor 'Object'
    |     property 'props' -> object with constructor 'Object'
    --- property 'errorMessage' closes the circle

i found this by visiting the default label story and trying to change the `errorMessage` prop

pwolfert:
Good catch, Sarah! I'm not exactly sure what the root cause is, but I do know that it's not very helpful to have those props use _object_ controls. [Other people have seen it too](https://stackoverflow.com/a/59996477). I switched to strings, and that made the errors go away.

PR #2805, opened 2023-11-14
[NO-TICKET] Adding a name property to the DropdownChangeObject

joefarhaven:
## Summary

- Add a "name" property to the DropdownChangeObject in order to preserve functionality that was previously provided by the generic event object. Passing the input name (if a param is provided) as part of the onChange event allows consumers of the event to identify which input triggered the event. This allows the user of patterns that centralize component management.

## How to test

1. Validate that onChange events are still fired successfully when a value is selected.
2. Inspect the target parameter passed to onChange in order to verify that both name and value are being passed.

## Checklist


pwolfert (approved):
Looks great, @joefarhaven 👍 

We'll chat about publishing timelines.

PR #2806, opened 2023-11-15
[WNMGDS-2571] Combine browser tests for faster CI

pwolfert:
## Summary

Combines a11y tests into other browser tests to speed them up. Right now it takes CI 6m37s to run our a11y tests. Combining them with the smoke VRT tests means we can not only run fewer of them, but we can load each story once and run two different tests against it. It takes CI way too long to load these pages, so this should hopefully speed things up. It takes our CI 3m47s to run the VRT smoke tests. So between a11y and vrt, that's 10m24s. That's the number to compare to.

- Combining them means that we load each story page only once
- We're also running a11y tests in a mobile-sized browser window now, whereas before we only ran them in a desktop-sized browser window
- More thorough checks for each theme permutation of each component now happen in our nightly builds, just like our visual regression tests

### Results

The [combined tests](https://ci.backends.cms.gov/wds/blue/organizations/jenkins/design-system%2Ftests/detail/pwolfert%2Fstreamline-browser-tests/4/pipeline/31) for [3a6ae04](https://github.com/CMSgov/design-system/pull/2806/commits/3a6ae0466e737924a682141645b30c558d0997fa) ran in 5m18s. That is 5 minutes faster in total than what we had before. 

_Note that the CI test script has is still running separate a11y tests in that test run. Once the command has been removed from the CI script by merging my companion pull request, I will update this branch to remove the old a11y tests._

## How to test

Take a look at the logs, and also note that I've got a pull request for updating the CI script used in Jenkins.

## Checklist


zarahzachz (approved):
noice!

PR #2804, opened 2023-11-14
[WNMGDS-2523] Replace `useFormLabel` with `useLabelProps` and `cleanFieldProps`

pwolfert:
## Summary

WNMGDS-2523 is about lessening the negative effects of the useFormLabel abstraction. I'm working on a feature branch that breaks `useFormLabel` up into some smaller, more focused hooks that can be mixed and matched. This is one of several pull requests to go into this feature branch.

- Replaces the remaining functionality of `useFormLabel` with a more focused `useLabelProps` function and a `cleanFieldProps` function

## How to test

New automated tests are coming in https://github.com/CMSgov/design-system/pull/2807

PR #2803, opened 2023-11-14
[WNMGDS-2523] Break some `useFormLabel` functionality into a `useHint` hook

pwolfert:
## Summary

WNMGDS-2523 is about lessening the negative effects of the useFormLabel abstraction. I'm working on a feature branch that breaks `useFormLabel` up into some smaller, more focused hooks that can be mixed and matched. This is one of several pull requests to go into this feature branch.

- Adds a `useHint` hook that will conditionally render a hint and return a `hintId` if the component props contain hint data.

## How to test

These components should be manually tested a bit, but automated tests should also cover this work.

PR #2807, opened 2023-11-16
[WNMGDS-2571] Add tests for Storybook docs

pwolfert:
## Summary

In order to better validate that the work in #2802, #2803, and #2804 does not result in prop-documentation regressions, we want to have tests that will take a snapshot of the data contained in the prop tables in the storybook docs for each component. This branch adds to the browser test suite a series of tests that visit the doc pages in Storybook and take a text snapshot of the data contained in the args tables. Note that we only need to run the tests in a single browser, since the data will always be the same across browsers. Note that some doc pages have no args tables, and some (the web components docs) have a separate event args table as well.

## How to test

Run `yarn test:browser:storybook-docs`

[Here's an example of it running in CI](https://ci.backends.cms.gov/wds/blue/organizations/jenkins/design-system%2Ftests/detail/PR-2807/4/pipeline/) (48s to run these tests by themselves).


dvogel:
hi

zarahzachz (approved):
this looks good but i was curious about the test file names - why do most end with "snapshot-1" and one end in "webkit"? seems like it'd be cleaner to omit the snapshot part of the file name since its in a dir called "snapshots"? not sure if that's possible though.

pwolfert:
@zarahzachz, I'm not seeing the file that ends with `webkit`. What's the full name of the file?

Regarding the `-snapshot-1`, it's automatically added by Playwright based on the test name. It's actually `-prop-table-matches-snapshot` + `-1`. Right now it's redundant, but if we were ever to add a second test to that module, it would keep it unique. It does seem really overly verbose though 🤔

PR #2808, opened 2023-11-17
[WIP] [WNMGDS-2437] Import tokens

zarahzachz:
WNMGDS-2437


gulpfile.js
```diff
@@ -57,31 +71,61 @@ const cleanDist = (cb) => {
...
+/**
+ * Copy Sass layout files from design-system-tokens to theme styles folder
+ */
+
+const copySassLayoutTokens = (cb) => {
+  const sassLayoutTokensFiles = `${tokensPackageFiles}/css-vars/${theme()}-layout-tokens.scss`;
+
+  gulp
+    .src(sassLayoutTokensFiles)
+    .pipe(rename('_layout.scss'))
```
pwolfert, line 83: I found it! What will allow us to avoid copying this file is [the at-forward rule](https://sass-lang.com/documentation/at-rules/forward/). If we replace our `src/styles/_layout.scss` file contents with the following, we'll be able to remove this gulp task:

```scss
@forward "../../../design-system-tokens/dist/css-vars/[THEME_NAME]-layout-tokens.scss";
```

gulpfile.js
```diff
@@ -57,31 +71,61 @@ const cleanDist = (cb) => {
...
 /**
  * compile sass assets to css, copy to /dist/css folder
  */
 const compileSass = (cb) => {
   const envDev = process.env.NODE_ENV === 'development';
 
-  const sassSourcePaths = isCore
-    ? `${srcPath}/styles/**/*.scss`
-    : [`${sassCorePath}/**/*.scss`, `${srcPath}/styles/**/*.scss`];
+  const sassSourcePaths = `${tokensPackageFiles}/scss/${theme()}-*.scss`;
```
pwolfert, line 124: This is no longer compiling the sass of each design system package. It is only compiling the sass variables files in the tokens. That is, we no longer get a `dist/css/index.css` in each of the design systems.

pwolfert: I don't think we actually need any of the changes in this function, but I could be wrong.

packages/design-system-tokens/src/index.ts
```diff
@@ -53,7 +65,12 @@ const tokenExporter = (inputType: string, exportType: string): number => {
   if (!INPUT_TYPES.includes(inputType)) help(`valid import types are: ${INPUT_TYPES}`);
   if (!EXPORT_TYPES.includes(exportType)) help(`valid export types are: ${EXPORT_TYPES}`);
 
-  const res = tokenExporter(inputType, exportType);
+  let res;
+  if (exportType === 'css-vars') {
+    res = cssVarExporter(inputType);
+  } else {
+    res = tokenExporter(inputType, exportType);
+  }
```
pwolfert, line 73: Since there is already some conditional logic about the export type in the `tokenExporter` function, is there a way to marry the two functions again? It actually looks like the logic inside them is exactly the same? Same `outputPath` being calculated

packages/design-system-tokens/src/lib/exportCssVars.ts
```diff
@@ -133,15 +132,29 @@ const writeCssFile = (filename: string, importedModule: any, sep: string) => {
...
+      acc[theme] = [];
+    }
+    acc[theme].push(file);
+    return acc;
+  }, {});
+
+  Object.entries(groupByTheme).forEach(([theme, files]) => {
+    let output = '';
+
+    files.reverse().forEach((file) => {
```
pwolfert, line 147: Rather than relying on knowing the order and reversing it, could we use some kind of string matching to make sure the one we want to come first comes first?

pwolfert: I like the structure of this `exportCssVars` function!

gulpfile.js
```diff
@@ -57,31 +71,61 @@ const cleanDist = (cb) => {
 };
 cleanDist.displayName = '🧹 cleaning up dist path';
 
+/**
+ * Copy Sass layout files from design-system-tokens to theme styles folder
+ */
+
+const copySassLayoutTokens = (cb) => {
+  const sassLayoutTokensFiles = `${tokensPackageFiles}/css-vars/${theme()}-layout-tokens.scss`;
```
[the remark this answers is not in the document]
pwolfert: Okay, I see that it's because it's related to the generation of the css-vars. Keeping it here is a good reminder that we haven't completely become independent of Sass variables.

packages/design-system-tokens/src/lib/exportCssVars.ts
```diff
@@ -133,15 +132,29 @@ const writeCssFile = (filename: string, importedModule: any, sep: string) => {
...
+      const importedModule = require(`${file.moduleImportName}`);
+      // component files do not need a separator
+      const sep = file.baseName.includes('component') ? '' : '-';
+      const layoutFilename = `${outPath}/${theme}-layout-tokens.scss`;
+      writeSassLayout(layoutFilename, file, importedModule, sep);
```
pwolfert, line 152: It looks like this part can move down to line 156, because I think it's actually writing the file twice! This would also be a good opportunity to add some more comments about why we need to write this file in the first place.

packages/docs/static/themes/cmsgov-theme.css
```diff
@@ -493,7 +493,6 @@
 --typography-heading-lg__font-weight: 700;
 --typography-heading-md__font-weight: 700;
 --typography-heading-sm__font-weight: 400;
---typography-heading-base__font-size: 1rem;
```
pwolfert, line 496: Huh, is `main` not up-to-date?

pwolfert (commented):
Good work so far! I've made some realizations while testing this and walking through the code. I made a comment about building the doc site css that I don't yet have a good solution for.

gulpfile.js
zarahzachz: i was trying to move all the logic out of the copy_themes.sh file. in there, there was logic to copy the Sass files into dist/sass, while our gulp func seemed to compile Sass files into CSS (dist/css). 

am i good to just omit copying our Sass to dist/sass?

pwolfert: Oh I see, the `tokens/dist/scss/*` files still need to be copied to our `[ds-packages]/dist/scss` folder. This function is compiling them, though, so I think we could just add those scss files to the glob in `copyThemes`, since that is just a file copy.

packages/docs/static/themes/cmsgov-theme.css
pwolfert: I just [put  up a PR to fix it](https://github.com/CMSgov/design-system/pull/2812)

gulpfile.js
zarahzachz: i'm rolling back this func to what it was before and breaking the scss copy logic out into its own func. 

i tried grouping it with the `copyThemes` func, but ran into build errors. mainly because i was trying to take files from 2 different sources (`tokens/**/css-vars/*.css` and `tokens/**/scss/*.scss`) and copy them into different directories (`theme/dist/css/*` and `theme/dist/scss/*`). gulp couldn't handle that logic in a singular function

packages/design-system-tokens/src/lib/exportCssVars.ts
```diff
@@ -133,15 +132,36 @@ const writeCssFile = (filename: string, importedModule: any, sep: string) => {
...
+    let output = '';
+
+    files.forEach((file) => {
+      const importedModule = require(`${file.moduleImportName}`);
+      // Component files do not need a separator
+      let sep = '';
+      output += mapCssVariablesToValues(importedModule, sep);
+
+      // Scss files are only generated from global token files
+      if (!file.baseName.includes('component')) {
```
zarahzachz, line 154: so i tried refactoring this so the scss layout file is written outside the `forEach` loop, but ran into some issues - mainly that the `writeSassLayout()` relies on the `file` parameter to get info to generate the scss vars 😵 

i honestly think that `writeSassLayout()` could probably be simplified since the only scss vars generated come from the theme's global token file and not the component files (right now that func appears to grab ALL tokens, then filters them based on if the token matches the name of a sass var in the `sassFilename` array at the top of the file).

zarahzachz: in this loop, i only run the write sass func if the `file` isn't a component token file. this ensures it only runs once.

zarahzachz: i can't wait to get rid of sass

packages/design-system-tokens/src/lib/exportCssVars.ts
zarahzachz: i actually don't think the `reverse()` matters since it's just assigning the tokens to the :root and we don't have tokens dependent on other tokens in this setup. so i removed it.

gulpfile.js
```diff
@@ -308,7 +351,7 @@ const displayHelp = (cb) => {
 log('🪴 building the cmsds');
 exports.build = gulp.series(
   cleanDist,
-  gulp.parallel(copyThemes, copyImages, copyFonts, copyJSON),
+  gulp.parallel(copyCssThemes, copyScssThemes, copyThemesToDocs, copyImages, copyFonts, copyJSON),
```
pwolfert, line 354: I've been getting an error when I try to build, and I can't figure out why:
```
[16:12:54] The following tasks did not complete: build, <parallel>, 🖍  compiling sass into css
[16:12:54] Did you forget to signal async completion?
```

All the functions look good to me. We're calling the callback functions as expected.

pwolfert (commented):
The code is looking good, but I can't run it for some reason. We'll have to pick this up again on Monday and look at it together.

gulpfile.js
zarahzachz: lol i'm an idiot. it was caused by a typo in the updated exportCssVars - i was exporting a sass file called `{theme}-core-tokens.scss` instead of `${theme}-layout-tokens.scss`. pushing the fix up now

pwolfert: oh wow, ha ha! I could *not* see it no matter how hard I looked

pwolfert (approved):
LGTM except for this one function that I don't think you need anymore. Good work!

---

Also, a note to us in the future: Since I was scrutinizing the dist folders and looking for regressions, I noticed that we have an extra `dist/css/components/index.css` in the child design systems. I think we could probably clean that up pretty easily. We might even be able to do it by simplifying the following lines:

```js
const sassSourcePaths = isCore
  ? `${srcPath}/styles/**/*.scss`
  : [`${sassCorePath}/**/*.scss`, `${srcPath}/styles/**/*.scss`];
```

to something like

```js
const sassSourcePaths = [
  `${sassCorePath}/index.scss`,
  `${srcPath}/styles/index.scss`,
];
```

pwolfert:
Just looked at the failing CI tests, and it looks like we might need to add a `yarn build:tokens` to our CI script

gulpfile.js
```diff
@@ -308,7 +352,14 @@ const displayHelp = (cb) => {
 log('🪴 building the cmsds');
 exports.build = gulp.series(
   cleanDist,
-  gulp.parallel(copyThemes, copyImages, copyFonts, copyJSON),
+  gulp.parallel(
+    copySassLayoutTokens,
+    copyThemes,
+    copyThemesToDocs,
```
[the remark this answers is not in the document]
zarahzachz: i created an npm script in /docs to handle the copying and removed the gulp func.

packages/docs/static/themes/core-theme.css
```diff
@@ -529,4 +375,158 @@
...
+--animationspeed-1: 0.25s;
+--animationspeed-2: 0.3s;
+--animationspeed-3: 0.5s;
+--animationspeed-4: 0.8s;
+--colorwhite: #ffffff;
+--colorblack: #000000;
+--colortransparent: #ffffff00;
+--colortransparent-black-alpha50: #00000080;
+--colortransparent-black-alpha25: #00000040;
+--colortransparent-white-alpha50: #ffffff80;
```
pwolfert, line 389: It took me running the doc site to notice this. There is no delimiter between `color` and the color name! (And same with `animation` and `speed`

zarahzachz: doh! i think it's cuz i wrote the css vars to the output before the !file.baseName.includes('component') logic. checking that theory now.

zarahzachz:
> Just looked at the failing CI tests, and it looks like we might need to add a `yarn build:tokens` to our CI script

looks like i don't have access to the jenkins anymore. i had this happen like a month ago where i wasn't able to see atlassian sites on my work laptop, but could on my personal one. ad hoc and gov't help desk couldn't help and it ended up working on its own within a few days 😦

pwolfert (approved):
Tested it again, and LGTM. We could remove `docs/static/themes` from version control entirely, but maybe keeping one copy around is a good sanity check since we don't have any unit tests for the css vars.

PR #2811, opened 2023-11-18
[WNMGDS-2523] Update form field types to not use `FormFieldProps`

pwolfert:
## Summary

WNMGDS-2523 is about lessening the negative effects of the useFormLabel abstraction. I'm working on a feature branch that breaks `useFormLabel` up into some smaller, more focused hooks that can be mixed and matched. This is one of several pull requests to go into this feature branch.

- Updates prop types to get props from individual label/hint/error hook props
- Deletes the now unused `useFormLabel` module

## How to test

The new `yarn test:browser:storybook-docs` command tests for changes to prop documentation, which is also an indicator of unintentional type changes.


zarahzachz (commented):
I'm getting an error trying to run the `yarn test:browser:storybook-docs` on my work laptop

```bash
browserType.launch: Executable doesn't exist at /Users/sarahsachs/Library/Caches/ms-playwright/chromium-1048/chrome-mac/Chromium.app/Contents/MacOS/Chromium
```

pwolfert:
@zarahzachz, I'll change the npm script so it runs in Docker like the others, even though it's not necessary for the tests. That'll at least make it consistent. Had you ever run [npx playwright install](https://github.com/CMSgov/design-system/blob/main/README.md?plain=1#L99) on that computer?

zarahzachz (approved):

PR #2828, opened 2023-12-01
[WNMGDS-2578] Standardize X for dialog close button

pwolfert:
## Summary

- Dialog uses a new, internal `CloseButton` that is an "X" for sighted users and still gives plenty of context for screen-reader users
    - Button is visually aligned with upper-right corner and with the heading text
    - Heading text is no longer nudged down from the top edge of the dialog's padding by its line height property
    - The whole dialog header has spacing between it and the dialog body, instead of margins coming from individual elements in the header
- Removed props for customizing the dialog button text and style (except for aria labels)
- Removed Medicare overrides for dialog and dialog button
- Removed Medicare overrides for the close (X) icon sizes, which affects VRTs for the tooltip
- Removed Medicare-specific dialog story in Storybook

## How to test

Here's a [Storybook demo](https://cmsgov.github.io/design-system/branch/pwolfert/dialog-close/storybook/?path=/story/components-dialog--dialog-example). Try it in all themes.

## Checklist

This demo will be taken to the UX leads for final approval.


packages/design-system/src/components/CloseButton/CloseButton.tsx
```diff
@@ -0,0 +1,40 @@
...
+ *
+ */
+export const CloseButton = ({ className, id: idProp, ...buttonAttributes }: CloseButtonProps) => {
+  const id = useId('close-button--', idProp);
+  return (
+    <button
+      className={classNames('ds-c-close-button', className)}
+      type="button"
+      id={id}
+      {...buttonAttributes}
```
zarahzachz, line 33: i feel like because this is an icon button, it needs some sort of aria-label describing what it's doing, and that label should support i18n as well

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -177,20 +151,16 @@ export const Dialog = (props: DialogProps) => {
       >
         <header className={headerClassNames}>
           {heading && (
-            <h1 className="ds-h2" id={headingId} ref={headingRef}>
+            <h1 className="ds-c-dialog__heading" id={headingId} ref={headingRef}>
```
zarahzachz, line 154: 🤔 does this h1 need to be evaluated? wondering if this component should be able to change heading levels, or if because it's a dialog it doesn't have to follow the 1 h1 per page rule 

cc: @SparrowSpit

packages/design-system/src/components/CloseButton/CloseButton.tsx
zarahzachz: saw aria-label was added from within other components like dialog, but i feel like there should still be a sensible fallback in case this gets used and someone doesn't realize this prop would be required

zarahzachz: i also don't think `ariaHidden` should be false since the button would have the label describing the action. the icon doesn't need to also provide this information - too much noise

tests/browser/snapshots/components-icons--available-icons-medicare--Mobile-Chrome.png
zarahzachz, line 1: does the size change on the thin icon impact other components?

zarahzachz (commented):
code looks good. needs design approval since VRTs changed.

tests/browser/snapshots/components-icons--available-icons-medicare--Mobile-Chrome.png
zarahzachz: i was actually thinking this might impact filter chip, but it doesn't seem to since that component's using the chunky close. should filters also be updated to align with the new close icon design?

packages/design-system/src/components/Dialog/Dialog.tsx
SparrowSpit: Given flexibility that this component must have combined with the unreliability of devs to choose the right heading, I vote for now to leave as is. We can revisit later if the there is enough steam to do so. Good question though.

packages/design-system/src/components/CloseButton/CloseButton.tsx
pwolfert: The aria label is coming from [this line in the dialog](https://github.com/CMSgov/design-system/pull/2828/files#diff-1d82a8d2bfc6ee2fb3dc5d5eff699b17a50835b8f9bb131ff1f87690f4f60094R159). It's an internal component, but maybe I should set the `aria-label` prop to required so we know it's being defined.

tests/browser/snapshots/components-icons--available-icons-medicare--Mobile-Chrome.png
pwolfert: There are several opportunities to align on a single close button style, but the scope of the changes was specifically restricted to this component in the UX meeting where the work was prioritized. It'll be a good future improvement.

packages/design-system/src/components/CloseButton/CloseButton.tsx
pwolfert: Added in [fde5930](https://github.com/CMSgov/design-system/pull/2828/commits/fde593024df284c24c9afd41ec3a299095a12615)

pwolfert: Does that alleviate your concerns, @zarahzachz?

pwolfert:
I think I'll make a follow-up PR to tokenize the close button colors

zarahzachz (approved):

PR #2834, opened 2023-12-08
[WNMGDS-2582] Add a11y docs to button

zarahzachz:
WNMGDS-2582

I created a subsequent ticket to address inconsistencies in the component guidance. I wonder if it makes sense to move some of the guidance to be closer to the example related to it? E.g., we have guidance around icon buttons, which icons are allowed, how to handle them accessibly, etc. and I think it's far more likely users will have a better understanding of how to use that specific example if its instructions are with it, otherwise they'll have to read all the button guidance and may miss the stuff related to icons.

### DIFF
--- a/packages/docs/content/components/button.mdx
+++ b/packages/docs/content/components/button.mdx
@@ -303,17 +303,6 @@ When pairing 2 CTAs:
 - Confirm the user meant to trigger a destructive action before following through with the action.
 - Provide a method for a user to undo a destructive action.
 
-**Avoid using disabled buttons**
-
-Disabling a "button" to act as a guide to the level of form completion is a common anti-pattern (e.g., you aren't able to proceed in the next step until the form contains enough data). Why it's a problem:
-
-- Buttons tend to have call-to-action text that matches what users want to do, so people try to interact with them.
-- They have terrible color contrast.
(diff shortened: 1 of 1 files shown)


pwolfert:
Regarding the examples thing, that might be a good thing to bring up to Brent, since his designs for component page IA drastically cut down on what's on the page.

packages/docs/content/components/button.mdx
```diff
@@ -328,19 +317,90 @@ For these reasons, disabled buttons can be frustrating for all users but especia
...
+Links and buttons are read differently when using a screen reader. In its most basic form, the rules are:
+
+- If it goes somewhere, use a link in the form of `<a>`.
+- If it does something on the page such as submitting a form, opening a dialog, canceling an action, or deleting an item, then use a button in the form of `<button>`.
+
+##### Navigating with a screen reader
+
+Web browsers and assistive technology aid users by providing lists of common elements like headings, buttons, links, and images, and let you navigate directly to the element of your choosing. This allows screen reader users to speed up their comprehension of the page.
+
+One example of this collection of lists in a screen reader is the [VoiceOver rotor](https://support.apple.com/guide/voiceover/with-the-voiceover-rotor-mchlp2719/mac) on a Mac. To turn on VoiceOver’s rotor, use the keyboard shortcut: **control + alt + U**.
```
pwolfert, line 333: Nice 👍

pwolfert (approved):
LGTM 👍

PR #2849, opened 2023-12-21
[WIP] [WNMGDS-2594] implement a11y docs on dropdown

zarahzachz:
WNMGDS-2594

This ticket was to implement the a11y testing guidelines on the dropdown docs using the [a11y testing guidance doc](https://docs.google.com/document/d/1mjMh77CCBvRihmUj4baQEi1KsWYs_A8R9NBlvLbIFy0/edit#heading=h.jeqxbav7txu2), but dropdown hasn't been completed yet. However, it looks like there are a11y guidelines already in existence from an earlier time (possibly when a custom dropdown was made). It's possible the guidelines written will need re-evaluated once the a11y doc's completed.

Otherwise, I fixed a broken link in our docs and updated the heading structure to include "Accessibility" so it matches the other pages.


packages/docs/content/components/dropdown.mdx
```diff
@@ -59,7 +59,9 @@ This component also makes use of some text input styles, which can be customized
 
 **[View the "Forms" guidelines for additional guidance and best practices.](/patterns/Forms/forms/)**
 
-### Accessibility testing
+### Accessibility
+
+##### Accessibility testing
```
pwolfert, line 64: Looks like it's jumping from h3 to h5. Is the second supposed to be h4?

packages/docs/content/components/dropdown.mdx
```diff
@@ -59,7 +59,9 @@ This component also makes use of some text input styles, which can be customized
...
 
 #### General observations
```
pwolfert, line 66: Looks like this heading level might need to be updated too

pwolfert (approved):

PR #2857, opened 2023-12-29
[WNMGDS-2624] Use `p` tags for hints and errors

pwolfert:
## Summary

This is something our a11y experts would prefer even though it doesn't really affect things. Semantically the `p` is more meaningful for this block of text than a `div`.


zarahzachz (approved):

zarahzachz (commented):
local tests are failing because snapshots need to be updated

zarahzachz (approved):

PR #2858, opened 2023-12-29
[NO-TICKET] Updating masking pattern to apply to valid US phone numbers, which sh…

joefarhaven:
…ould not begin with 0 or 1

## Summary

**Problem:**

https://cmsgov.slack.com/archives/CHH0381RD/p1703794147097219

> We are seeing people include the US country code with their phone number. For example, [1.800.867.5309](tel:18008675309) instead of just [800.867.5309](tel:8008675309). With the way that the component/mask currently works, the number is saved as 1800867530 which is invalid.

**Proposed Solution:**

The first character of a valid 10-digit US phone number cannot be 0 or 1. Update masking patterns to only recognize phone numbers where the first digit is 2-9. The consumer of the component will still be responsible for data validation and error messaging.

## How to test

1. Verify that all automated testing passes.
2. Verify that phone numbers starting with 0 or 1 are not treated as valid US phone numbers.
3. Verify that phone numbers starting with 2-9 are treated as valid US phone numbers.

## Checklist

- [X] Created or updated unit tests to cover any new or modified code


pwolfert:
This prototype was made during a discussion about phone number validation, and it was later concluded that this was an insufficient solution for the general case. I think the preferred path forward is showing explicit error messages when the numbers are invalid.

PR #2863, opened 2024-01-02
[WNMGDS-2621] Enforce mutual exclusivity of the `mask` and `labelMask` props

pwolfert:
## Summary

This updates the logic for `TextField` to make the two different mask props truly mututally exclusive. Previously if you included both `mask` and `labelMask`, you'd see two inputs in a `TextField`.

- Overview of changes
- A link to the Jira ticket for quick reference

## How to test

Use [this local storybook link](http://localhost:6006/?path=/story/components-textfield--label-masked-phone&args=mask:ssn) to have a story with both props. Prior to this fix, you will have seen it render two inputs.

## Checklist


zarahzachz (commented):
i think there's a storybook issue... if you type too quickly, the component seems to lose its styles

https://github.com/CMSgov/design-system/assets/5159392/180851fb-f492-4412-bd38-bbd4ed5225d4

pwolfert:
@zarahzachz, it looks like this has to do with our `ControlledTemplate` story. It happens in prod right now, and if I change the default TextField story to use the controlled template, it has the same problem. I think it has to do with the args being updated in the URL and storybook not rendering everything fast enough.

zarahzachz (approved):

PR #2873, opened 2024-01-05
[WNMGDS-2543] Add LTS policy

zarahzachz:
WNMGDS-2543

Added LTS policy to browser support docs page.


pwolfert (commented):
The text is clear and concise; that LGTM. However, I don't think this is its home. I don't know where to move it, but I don't think putting it in the "Browser Support" page is very discoverable. Maybe we should have a sibling page called "Support Policy'. Maybe it would be appropriate to move browser support into a "Support Policy" page too.

pwolfert (approved):

PR #2876, opened 2024-01-05
[WNMGDS-2614] Remove Firefox focus styles added in #510

zarahzachz:
WNMGDS-2614

Focus styles were added to override the select element in Firefox. Because the dropdown's been refactored to NOT be a select element, these styles are no longer applicable. In fact, they cause a fun little bug that turns the SVG inside white when focus-visible happens. What a journey! [This is the context in which it was originally added](https://github.com/CMSgov/design-system/pull/510/files#diff-be195afbb32878f98661d965a1e5565af48defe54f8a89bd585eab496d1bff7a).


pwolfert (approved):
Woohoo! It's working

PR #2880, opened 2024-01-10
[WNMGDS-2633] Add the missing `ds-u-color--gray-dark` util that we said existed

pwolfert:
## Summary

WNMGDS-2633

Add the missing `ds-u-color--gray-dark` util that we said existed [in our docs](https://design.cms.gov/utilities/text/text-color/).

## How to test

The screenshots below show that the `gray-dark` color wasn't actually being applied to the `ds-u-color--gray-dark` text in the table because that utility class didn't actually exist in our CSS. It worked in the circle swatch because that was pulling directly from the token variable.

### Before
Before screenshot showing that the dark gray text is actually the default black

### After
After screenshow showing that the dark gray text is actually dark gray

## Checklist

### DIFF
--- a/packages/design-system/src/styles/utilities/_text-color.scss
+++ b/packages/design-system/src/styles/utilities/_text-color.scss
@@ -60,6 +60,10 @@
   color: var(--color-gray-light) !important;
 }
 
+.ds-u-color--gray-dark:not(:focus) {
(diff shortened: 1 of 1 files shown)


hi-its-hailey (approved):
@pwolfert Looks great, thank you!

PR #2881, opened 2024-01-10
[NO-TICKET] Clean up lint warnings

dmethvin-gov:
## Summary

Clean up the easy-to-fix lint warnings. They generally fall into a few categories:
* SCSS syntax that stylelint didn't recognize -- Added SCSS directives to the allowed list
* Things `import`ed but never used in the file -- Removed the imports
* Multiple `import`s for the same file -- Combined the imports on one line
* Function arguments never used -- Removed the args
* Functions with empty bodies -- Had them return `null` which is fine
* dangerouslySetInnerHtml -- Disabled the warning if the source string seemed safe
* Other things: Explanation comments in the PR below

   **Before**: 15 stylelint warnings, 113 eslint warnings
   **After**: 5 stylelint warnings, 68 eslint warnings

The remaining warnings fall into a few categories as well. They involve non-cosmetic changes to the code or non-trivial changes to dependencies, so I didn't attempt them. In particular there are a lot of these:

* Missing dependencies in `useEffect` functions - These can cause erratic behavior so it would be good to analyze and fix them. These sort of changes often trade one sort of (perhaps undiscovered) bug for another.
* Things that shouldn't give warnings. For example, eslint is configured to allow unused variables when defined along with `...rest` declarations but it doesn't seem to be working. It might be fixed by an ESLint upgrade, but when I tried doing that I encountered conflicts. So I left all that alone for now.

## How to test

There should be no functional difference. All existing unit and snapshot tests pass but one, and I can't see the difference in it but it might be due to an extra space after removing the unnecessary fragment.

[webkit] stories.test.ts:70:13 › Components/Icons/Available Icons › with medicare theme › matches snapshot

## Checklist


packages/design-system/src/components/Table/Table.stories.tsx
```diff
@@ -98,8 +98,8 @@ const Default: Story = {
           </TableRow>
         </TableHead>
         <TableBody>
-          {tableTemplateData.data.map((dataItem, index) => (
-            <TableRow key={index}>
+          {tableTemplateData.data.map((dataItem) => (
+            <TableRow key={dataItem.documentTitle}>
```
dmethvin-gov, line 102: Eliminates a "using index as a key" lint warning.

packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
```diff
@@ -1,9 +1,6 @@
 import { RefObject } from 'react';
 
-export function getAnalyticsContentFromRefs(
-  refs: RefObject<any>[],
-  componentName?: string
```
dmethvin-gov, line 5: This arg isn't used in the function.

packages/design-system/src/components/Pagination/Pagination.tsx
```diff
@@ -196,7 +195,7 @@ function Pagination({
      * Renders all Page components in range (3 pages) to Pagination component.
      */
 
-    pageRange.map((page) => {
+    pageRange.forEach((page) => {
```
dmethvin-gov, line 198: Since `map` isn't returning anything, `forEach` is the right choice here.

packages/design-system/src/components/Icons/icons.stories.tsx
```diff
@@ -195,32 +195,31 @@ const iconData = [
 ];
 
 export const AvailableIcons = () => (
-  <>
-    <table className="ds-c-table">
```
pwolfert, line 199: 👍

packages/design-system/src/components/Pagination/Pagination.tsx
pwolfert: Nice, yep, that makes sense 👍

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -3,7 +3,6 @@ import Alert from '../Alert/Alert';
 import TableCaption from './TableCaption';
 import TableContext from './TableContext';
 import classNames from 'classnames';
-import uniqueId from 'lodash/uniqueId';
```
pwolfert, line 6: Nice, no more lodash here 🎉

packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
pwolfert: Looks like it isn't used anymore, but there is a lingering instance of it being called with the second parameter that needs to be updated. Right now it fails to build.

packages/ds-healthcare-gov/src/components/ThirdPartyExternalLink/ThirdPartyExternalLink.tsx
```diff
@@ -1,4 +1,3 @@
-import React from 'react';
 import { ThirdPartyExternalLink } from '@cmsgov/design-system';
```
pwolfert, line 1: We held back on removing these React imports because of a problem we found somewhere with backwards compatibility. It doesn't look like there's any problem with removing them from the story files, but there may be an issue with removing it from the ones we export in our package. I need to do a little more digging.

pwolfert (commented):
Hey, thanks for cleaning these up! It looks like most of this will work, but there's a build error to fix, and there's some research I need to do on the React import removals.

pwolfert:
not a wolf

packages/ds-healthcare-gov/src/components/ThirdPartyExternalLink/ThirdPartyExternalLink.tsx
dmethvin-gov: I noticed that as well with React imports in the main code (not Storybook). I think it probably has something to do with the new JSX transform introduced a few years ago. It's possible the current build setup may be using the old way, which required React in scope even if it wasn't being explicitly used. 

https://legacy.reactjs.org/blog/2020/09/22/introducing-the-new-jsx-transform.html

packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
dmethvin-gov: Somehow I missed that! I must have done the build before I saved this file. I really do appreciate that the project has a thorough set of unit and snapshot tests, it makes it much easier for newbs like me to change things and know if we're breaking something.

packages/ds-healthcare-gov/src/components/ThirdPartyExternalLink/ThirdPartyExternalLink.tsx
dmethvin-gov: I just checked and yes, it's using the old JSX transform. Since the min peerDependency for React is 17.0.0 it should be possible to upgrade without a major version bump. Several of the build dependencies are almost 2 years old so it might be a good idea to freshen them up as well. If you want I can take a look after I look into the react-day-picker+date-utils thing.

pwolfert: I actually did not do a good job of making sure that peer dependency version stayed where it needed to. We're technically still supporting React 16 projects, mostly on the medicare side. Hopefully this year we'll be able to drop React 16. For that reason, I don't think we're ready for this. Thank you for the offer to look into it. If you really want to, it might be good to figure out everything that we will need to and get to change when we do finally drop React 16, even if we can't merge a PR right now.

dmethvin-gov: For the ones that I removed, there was no JSX being rendered in the file so React wasn't needed regardless. It looks like the medicare.gov home page is on 17.0.2, according to React devtools? I used to have visibility into the medicare and healthcare repos but they moved them late last year.

dmethvin-gov:
I updated this to remove componentName everywhere. The failure is probably the one previous snapshot failure, but I don't have visibility into the Jenkins jobs to see what failed. If it's something else, let me know.

packages/design-system/src/components/HelpDrawer/useHelpDrawerAnalytics.ts
```diff
@@ -39,7 +39,6 @@ export default function useHelpDrawerAnalytics({
   }
 
   const [headingRef] = useAnalyticsContent({
-    componentName: 'Dialog',
```
pwolfert, line 42: Your changes here made me realize that in [this change](https://github.com/CMSgov/design-system/pull/1944) not only did I fail to remove the unused `componentName` property, but I also mislabeled this component through copypasta. Since this is in a fork, would you mind [updating this string for me](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/HelpDrawer/useHelpDrawerAnalytics.ts#L27) to reflect that it's for the help drawer?

packages/ds-healthcare-gov/src/components/ThirdPartyExternalLink/ThirdPartyExternalLink.tsx
pwolfert: Oh, I understand. I didn't realize that there was no JSX in those files. Carry on!

pwolfert (approved):
I have one request, but LGTM otherwise!

Currently investigating some sort of instability in my storybook-docs tests, which is what is currently failing your CI tests. (You can see it locally if you're curious with `yarn build:storybook && yarn test:browser:storybook-docs`.)

packages/ds-medicare-gov/src/components/HelpDrawer/HelpDrawer.stories.tsx
```diff
@@ -59,7 +59,7 @@ const drawerContent = (
 );
 
 export const HelpDrawerToggleWithDrawer: Story = {
-  render: function Component(args) {
+  render: function Component() {
```
pwolfert, line 62: I went down a rabbit hole of cleaning up these stories trying to get the arg tables to show up. Right now this medicare-specific story shows either a table with bogus args or no table at all. Both the main `HelpDrawer` and this Medicare-specific one need work, but I'm spending too many cycles on it right now and need to move on. Would you mind reverting this module's changes for now so the CI tests pass, and you can merge?

pwolfert: I'll create a follow-up PR at a future date

pwolfert: @dmethvin-gov, the current CI failure is due to these modules changes. If you revert this module to what's on `main`, they should pass. I'll clean these up at some point.

dmethvin-gov: That is super strange, since they aren't used. However, it did indeed fix the problem locally?

pwolfert: [Here's the follow-up](https://github.com/CMSgov/design-system/pull/2902)

PR #2882, opened 2024-01-10
[WNMGDS-2634] Update the visited link color for the cmsgov theme

pwolfert:
## Summary

WNMGDS-2634

Designers on CMS.gov worked with the design system team and UX Leads to pick a color from the design system palette that would work better for visited links than the original one. Visited links are now `orchid-800`.

## How to test

Using dev tools, you can toggle on the `:visited` state to the Typography > Links story in Storybook

### Before

Old windsor colored link

### After

New orchid colored link


hi-its-hailey (approved):
Looks good!

PR #2890, opened 2024-01-18
[WNMGDS-2555] Add `isOpen` to `Drawer` and `Dialog` to fix focus-management issue

pwolfert:
## Summary

What I think has been happening is that React removes the `Dialog` and `Drawer` from the DOM before the browser has a chance to fully execute its close cycle and transfer focus to the element that had focus before the `<dialog>` underlying was opened. By always having the `Dialog` or `Drawer` (internally `NativeDialog`) rendered by its parent element and passing an `isOpen` prop to control its state, the browser is able to successfully transfer focus back to the right place after the dialog is closed.

For temporary backwards compatibility during a deprecation period, you can still conditionally render the dialog, and it will behave as it did before, but it will warn you that you need a `isOpen` prop. This is slated to go out in v9, so we'll fully deprecate the old behavior in v10.

I was curious if other React libraries conditionally rendered or used a boolean prop for toggling state.  What I found was that _Carbon_, _Ant_, _Forma 36_, _Helsinki_, _Polaris_, and _Primer_ use a boolean prop instead of conditionally rendering the modal to show and hide it. _Gestalt_ uses conditional rendering, but their focus doesn't automatically return to the triggering button. The `react-aria` library uses modal trigger components like we were originally thinking we'd have to do (kind of an ugly solution to all this that would require more work for teams). All that to say, _we're not weird for adopting this pattern_.

- Added `isOpen` prop to `Dialog` and `Drawer`

## How to test

Use `yarn storybook:react` to reproduce the original problem, which only happens in React. If you open a dialog or drawer and then close it, focus won't move back to the button that opened it.

## Checklist


packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -115,22 +117,7 @@ export const Dialog = (props: DialogProps) => {
...
-  useLayoutEffect(() => {
-    // https://css-tricks.com/prevent-page-scrolling-when-a-modal-is-open/
-    const y = window.scrollY ?? 0;
-    const bodyClass = 'ds--dialog-open';
-    document.body.classList.add(bodyClass);
-    document.body.style.setProperty('--body_top--dialog-open', `-${y}px`);
-    document.documentElement.style.setProperty('scroll-behavior', 'auto');
-    return () => {
-      document.body.classList.remove(bodyClass);
-      window.scrollTo({ top: y, behavior: 'auto' });
```
pwolfert, line 130: Making the state dependent on the value of `isOpen` instead of following when it's mounted and unmounted, it meant that it was no longer as simple as a _use effect_ and _cleanup_. Because I no longer had direct access to the `y` value created in the top of the _use effect_ in a closure, I had to start tracking that state separately. I didn't love that, so you'll notice that I ended up moving this all into a separate module and used a reducer, which I ended up trying to model as a finite state machine. It might be a little more code than is needed, but I feel more confident in it being a finite state machine, especially since I've been able to leverage TypeScript to help make sure my logic is correct.

packages/design-system/src/components/Dialog/useBodyScrollPrevention.ts
```diff
@@ -0,0 +1,59 @@
...
+interface OpenState {
+  name: 'open';
+  bodyScrollY: number;
+}
+
+interface ClosedState {
+  name: 'closed';
+}
+
+const OPEN = (_state: ClosedState): OpenState => {
```
pwolfert, line 15: This `OPEN` action requires a starting point of `ClosedState` and an ending point of `OpenState`.

packages/design-system/src/components/NativeDialog/NativeDialog.tsx
```diff
@@ -50,8 +61,18 @@ export const NativeDialog = ({
...
+    // open state.
+    if (isOpen) {
+      if (!dialogNode.open) {
+        showModal ? dialogNode.showModal() : dialogNode.show();
+      }
+    } else {
+      if (dialogNode.open) {
+        dialogNode.close();
+      }
+    }
```
pwolfert, line 75: Because some of this state comes from the DOM, it didn't lend itself as well to modeling as a finite state machine. I didn't want to duplicate that state and make it way more verbose. Don't @ me. Or do, because this is a review.

pwolfert:
@zarahzachz, the tests are finally passing. This is ready for real review.

zarahzachz (approved):

PR #2908, opened 2024-01-27
[WNMGDS-2658] Fix crash of privacy settings dialog in node environments

pwolfert:
## Summary

WNMGDS-2658

The reason this is a problem now and wasn't before is that previously the `PrivacySettingsDialog` wouldn't be rendered in build environments like Gatsby's because that component's visibility was controlled by conditionally rendering it. When we switched to `isOpen`, it would render that dialog even if not open, which caused this code to be executed where it wasn't before. While this code had some safety checks in place, they weren't actually effective because the `getPrivacySettings` function didn't have the same short-circuiting behavior.

- Don't set a default cookie in test environments anymore.
    - Instead of always setting a default cookie in the root module when we're in a test environment, we'll just return the default value in `getPrivacySettings()` so the privacy settings dialog can go about its business without us having to touch the cookies. That seems to actually make it a little easier to read and reason about, and we get rid of the module side-effects.
- Return default settings in non-browser environments instead of crashing

## How to test

I wanted to test this better, but I wasn't successful. but it is difficult. I can change the jest environment on a per-module basis with [this directive](https://jestjs.io/docs/configuration#testenvironment-string), but it still runs the `setupTests` module and crashes. It was taking too much time to try reconfiguring Jest, so I tried mocking like this:

```ts
let windowSpy;

function mockNodeEnvironment() {
  windowSpy = jest.spyOn(window, "window", "get");
  windowSpy.mockImplementation(() => undefined)
}

function restoreJsdomEnvironment() {
  windowSpy.mockRestore();
}
```

but even though I could get `window` to be `undefined`, I couldn't get the `typeof window === 'undefined'` to be true.

## Checklist


zarahzachz (approved):
I couldn't recreate the issue, but it doesn't break anything either so approving!

Also, isn't Google phasing out cookies by end of year?

pwolfert:
Oh interesting, I hadn't read about _phasing out cookies_ until now. Yeah, sounds like the team in charge of that might need to move to a new strategy.

PR #2893, opened 2024-01-19
[NO-TICKET] Make Choice a function component and clean up logging

pwolfert:
## Summary

Cleaned out all the console logs and warnings and errors from unit tests, and in the process I converted `Choice` from a legacy _class component_ to a modern _function component_.

## How to test

`yarn test:unit`

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [ ] Selected appropriate release milestone

### If this is a change to code:

- [x] Created or updated unit tests to cover any new or modified code
- [x] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:update`)

PR #2921, opened 2024-02-06
[NO-TICKET] Fix the visual regression in #2893

pwolfert:
## Summary

Fixes the visual regression in #2893 where `Choice` error messages saw a slight vertical shift when viewed in the healthcare theme. The HealthCare.gov Design System sets all field errors to bottom-placement by default. When I started using the `useInlineError` hook in the `Choice` component, it began applying a bottom-placement CSS class that was causing the shift. We separated these hooks out of the one giant `useFormLabel` hook for this very reason—so that we could mix and match which features we want to use in each component. I simply stopped using that hook, because we don't need its placement logic. We can just use the `InlineError` component directly for this simpler case.

## How to test

- I ran the full suite of visual regression tests
- You can use Storybook to make sure there's no vertical shift when compared to our deployed Storybook for v9

## Checklist

### DIFF
--- a/packages/design-system/src/components/ChoiceList/Choice.tsx
+++ b/packages/design-system/src/components/ChoiceList/Choice.tsx
@@ -2,8 +2,9 @@ import React, { useEffect, useState } from 'react';
 import EvEmitter from 'ev-emitter';
 import classNames from 'classnames';
 import useId from '../utilities/useId';
+import { InlineError } from '../InlineError';
 import { Label } from '../Label';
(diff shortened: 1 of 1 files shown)


zarahzachz (approved):
I ran the `yarn test:browser:all` command and got 10 VRT issues for hgov. I'm not sure if this PR is reverting previously updated VRTs or if this is a legit issue though. Since CI is passing, I'm assuming VRTs are matching as expected.

pwolfert:
Can't reproduce, but we'll see what the nightly build says

PR #2922, opened 2024-02-06
[WIP] [WNMGDS-2160] Consolidate docs theme

zarahzachz:
WNMGDS-2160

- Docs site should only use core theme and no longer switch its appearance based on user theme selection
- Accordions used in docs site should be styled as Healthcare's accordion style (in progress)
- Embedded examples should show selected theme
    - How do we reflect theming for our foundation documentation (e.g., [body text](https://design.cms.gov/foundation/typography/body/?theme=healthcare)) 
    - Storybook embeds do reflect chosen theme


zarahzachz:
this branch has the latest of all UI updates in it:
- includes updates from base `feature/info-architecture-update` branch
- includes Accordion from hgov 
- removes ability for docs site to be themed, but keeps themed Storybook embeds

only reason I'm asking for a review in this branch instead of merging it into the info-architecture one is i'm not sure we want to keep the "persisting theme on docs site" work. open to discussing, but i personally found it hard to know what theme i was looking at when everything looked the same. it could be some sorta bias though since i've been working in this site for so long - i just don't think the core theme looks distinct enough from hgov to be our default docs look and it could confuse non-hgov users on our site

pwolfert:
One way that would provide a clue about what theme they're looking at that seems to be missing in this branch is the background color of the theme-switcher section changing to match the primary color for that theme

zarahzachz:
> One way that would provide a clue about what theme they're looking at that seems to be missing in this branch is the background color of the theme-switcher section changing to match the primary color for that theme

Or changing the title name in the header to reflect which theme was active like this:

image

or 

Screenshot 2024-02-06 at 8 03 22 PM

Bonus points for having a stylized logo for each theme

pwolfert:
Well the color thing was in the original designs, or I would have pushed back on the idea. I think we at least need the brand color somewhere.

hi-its-hailey:
Oh yeah, I vote for the brand color on the theme swither itself! The header switch would also be good, but I could see it being a more missable difference. Otherwise, looks good!

PR #2926, opened 2024-02-09
[WNMGDS-2511] Update hgov footer html

zarahzachz:
WNMGDS-2511

- Replace `h4` with `p` tag in hgov footer
    - Why? The current markup uses `h4` and it sometimes causes issues with the heading structure when `h2` or `h3` aren't used. This tag doesn't need to be a heading at all and only exists as a label for the list of language resources, so it's been replaced with a `p`
    - This needs to be an `aria-labelledby` to take advantage of browser translation features
- ~Add `aria-label="global"` to hgov footer~
    - ~This wasn't part of the original ticket, but hgov's footer should _probably_ have an `aria-label` to indicate it's the global footer for the site. I mirrored the same label used on hgov's header~
    - Footer doesn't need a label because it's the only element of its kind on a page ([source](https://adhoc.slack.com/archives/C049H6KSE6Q/p1707495223073149?thread_ts=1707492217.292529&cid=C049H6KSE6Q))

### DIFF
--- a/packages/ds-healthcare-gov/src/components/Footer/Footer.tsx
+++ b/packages/ds-healthcare-gov/src/components/Footer/Footer.tsx
@@ -39,7 +39,7 @@ export const Footer = (props: FooterProps) => {
   const classes = classnames('hc-c-footer', props.className);
 
   return (
-    <footer className={classes} role="contentinfo">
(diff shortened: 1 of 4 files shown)


pwolfert (approved):
LGTM 👍

PR #2936, opened 2024-02-13
[WNMGDS-2650] Remove unnecessary display utils from docs layout

zarahzachz:
WNMGDS-2650

If you visit a page like design.cms.gov/migration-guides/sass-to-css/ and start resizing your browser, you'll find a horizontal scroll will appear (around a tablet-sized viewport). Did some digging and this appears to be caused by a flex display being applied to the page content. These util classes don't seem to be adding anything so I removed them.


packages/docs/src/components/layout/Layout.tsx
```diff
@@ -112,10 +112,10 @@ const Layout = ({
               {children}
             </article>
           ) : (
-            <article className="ds-u-padding-x--3 ds-u-sm-padding-x--6 ds-u-sm-padding-bottom--6 ds-u-padding-top--1 ds-u-padding-bottom--3 ds-u-md-display--flex page-body">
+            <article className="ds-u-padding-x--3 ds-u-sm-padding-x--6 ds-u-sm-padding-bottom--6 ds-u-padding-top--1 ds-u-padding-bottom--3 page-body">
```
zarahzachz, line 115: This `article` has 1 direct child, a `div` with `ds-l-row`. It's unknown what the `ds-u-md-display--flex` class was adding. Removing fixes the overflow issues on the `pre` tag and doesn't appear to have any negative effects.

packages/docs/src/components/layout/Layout.tsx
```diff
@@ -112,10 +112,10 @@ const Layout = ({
...
               <div className="ds-l-row">
                 <div className="ds-l-lg-col--9">
-                  <div className="ds-u-display--block ds-u-lg-display--none ds-u-margin-bottom--3">
+                  <div className="ds-u-lg-display--none ds-u-margin-bottom--3">
```
zarahzachz, line 118: Removed this `ds-u-display--block` class because `div` is a display block by default and no other classes were changing this (until it was viewed at a large display, and that util makes sense)

pwolfert (approved):
Nice fix! It's seamless 🪡

PR #2937, opened 2024-02-14
[WNMGDS-2350] Fix background color of docs vertical nav on hover

zarahzachz:
WNMGDS-2350

I thought we needed to refactor our vertical nav CSS due to how complex the docs site CSS overrides were, but I was wrong and the solution turned out to be much simpler 🎉 

- Removed clever CSS in favor of redefining vertical nav token in docs site
- Removed unnecessary util classes from docs header
- Fixed `:hover:focus` state of hamburger icon

### DIFF
--- a/packages/docs/src/components/layout/SideNav/SideNav.tsx
+++ b/packages/docs/src/components/layout/SideNav/SideNav.tsx
@@ -106,9 +106,9 @@ const SideNav = ({ location }: SideNavProps) => {
       })}
     >
       <UsaBanner className="ds-u-display--block ds-u-md-display--none" />
-      <header className="c-navigation__header ds-u-md-display--block ds-u-md-display--none">
+      <header className="c-navigation__header ds-u-md-display--none">
         <Button
-          className="ds-u-md-display--none ds-u-padding-left--0 ds-u-padding-right--1"
+          className="ds-u-padding-left--0 ds-u-padding-right--1"
           variation="ghost"
           aria-expanded={isMobileNavOpen}
           aria-controls="c-navigation__menu"
--- a/packages/docs/src/styles/components/navigation.scss
+++ b/packages/docs/src/styles/components/navigation.scss
@@ -141,7 +141,6 @@
 .ds-c-vertical-nav__label {
   &:hover,
   :visited:hover {
-    // --vertical-nav-item__background-color--hover: var(--color-primary-lightest);
     --vertical-nav-item__background-color--hover: color-mix(
       in srgb,
       var(--color-primary-lightest),
--- a/packages/docs/src/styles/components/navigation.scss
+++ b/packages/docs/src/styles/components/navigation.scss
@@ -11,8 +11,13 @@
     gap: $spacer-2;
     padding: $spacer-2;
 
-    button {
+    > button {
       color: currentColor;
+
+      &:hover,
+      &:focus:hover {
+        color: currentColor;
+      }
     }
   }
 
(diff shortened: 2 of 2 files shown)


packages/docs/src/components/layout/SideNav/SideNav.tsx
```diff
@@ -106,9 +106,9 @@ const SideNav = ({ location }: SideNavProps) => {
       })}
     >
       <UsaBanner className="ds-u-display--block ds-u-md-display--none" />
-      <header className="c-navigation__header ds-u-md-display--block ds-u-md-display--none">
+      <header className="c-navigation__header ds-u-md-display--none">
         <Button
-          className="ds-u-md-display--none ds-u-padding-left--0 ds-u-padding-right--1"
```
zarahzachz, line 111: removing `ds-u-md-display--none` from `Button` because the parent `header` is handling this logic

packages/docs/src/components/layout/SideNav/SideNav.tsx
```diff
@@ -106,9 +106,9 @@ const SideNav = ({ location }: SideNavProps) => {
       })}
     >
       <UsaBanner className="ds-u-display--block ds-u-md-display--none" />
-      <header className="c-navigation__header ds-u-md-display--block ds-u-md-display--none">
```
zarahzachz, line 109: removing the `ds-u-md-display--block` class as the following `ds-u-md-display--none` cancels it out.

packages/docs/src/styles/components/navigation.scss
```diff
@@ -11,8 +11,13 @@
     gap: $spacer-2;
     padding: $spacer-2;
 
-    button {
+    > button {
```
zarahzachz, line 14: made the button selector more precise here with the `>`

packages/docs/src/styles/components/navigation.scss
```diff
@@ -11,8 +11,13 @@
...
       color: currentColor;
+
+      &:hover,
+      &:focus:hover {
+        color: currentColor;
```
zarahzachz, line 19: noticed the 🍔 icon is illegible when `:focus:hover` is applied and fixed that

packages/docs/src/styles/components/navigation.scss
```diff
@@ -133,42 +138,23 @@
   }
 }
 
+.ds-c-vertical-nav__label {
+  &:hover,
+  :visited:hover {
+    --vertical-nav-item__background-color--hover: color-mix(
```
zarahzachz, line 144: all the code in the `:before` styles seemed to be in an attempt to modify an existing color token to be something we don't have (adding an alpha to whatever the vert nav hover bg was). 

css filters are expensive in terms of overall perf, so removing that was a boon to us. also using [color-mix()](https://caniuse.com/?search=color-mix) to accomplish the transparency needs.

side-note: i don't really think we should be combining/modifying tokens for one-off implementation, and the best solution here was probably to update this to just rely on `--color-primary-lightest`. i tried that out, but the color was noticeably different than our current color and decided to be less aggressive with the change. maybe add this to the pile for things to review if we update the docs site?

packages/docs/src/styles/components/navigation.scss
```diff
@@ -133,42 +138,23 @@
   }
 }
 
+.ds-c-vertical-nav__label {
```
zarahzachz, line 141: moved this rule outside of the desktop media query as the hover state should be available across viewport sizes

packages/docs/src/styles/components/navigation.scss
```diff
@@ -133,42 +138,23 @@
...
+}
+
 // Desktop
 @media (min-width: $media-width-md) {
   .c-navigation {
     // Set it to a fixed width that is wide enough for most links so that it
     // doesn't jump around when we open and close subsections.
     width: 17rem;
-
-    .c-navigation__header {
```
zarahzachz, line 143: because the header doesn't appear on md+ sized viewports, this isn't needed

packages/docs/src/styles/components/navigation.scss
pwolfert: Oh yeah, I never noticed that because I must have always tested it in touch-device mode

packages/docs/src/styles/components/navigation.scss
pwolfert: Great analysis and solution! Very elegant

Sounds good to pile it onto future doc site changes 👍

pwolfert (approved):
Tested, and it looks good!

PR #2938, opened 2024-02-14
[WNMGDS-2653] Examples to use local design system packages

pwolfert:
## Summary

Thanks to Yarn _workspaces_ and Lerna, the example projects now share dependencies and use local versions of the design systems. This will also keep the examples up-to-date. A good follow-up to add would be to test the examples with VRTs so we can make sure that they don't break as we update our other packages.

- Added the `examples` directory to our Yarn _workspace_ and Lerna
    - Ignore the example projects when we're doing version bumps and publishing through Lerna
    - Make sure the example project packages have unique names
    - Update yarn lockfile
- Updated dependencies in the examples
- Standardized the build system and presentation of the `react-app`, `preact-app`, and `preact-react-app` examples
- Retired the `create-react-app-typescript` example since CRA is no longer maintained (and because it is incompatible with _workspaces_) and replaced it with a newer TypeScript-enabled framework
- Upgraded our project Node version because the version we were using was no longer being maintained and because some of our new dependencies were only compatible with Node v18+

## How to test

1. `yarn clean && yarn install` in the root
2. Go through the README instructions for each of the affected examples to build and run them

## Checklist


examples/astro/src/components/react/AccordionExample.tsx
```diff
@@ -3,7 +3,7 @@ import React from 'react';
 
 function AccordionExample() {
   return (
-    <div>
+    <>
```
pwolfert, line 6: All of these were copied from the old example and modified

examples/preact-react-app/package.json
```diff
@@ -1,22 +1,22 @@
...
   "private": true,
   "dependencies": {
-    "@cmsgov/design-system": "file:../../packages/design-system",
-    "preact": "^10.11.3"
+    "@cmsgov/design-system": "9.0.1",
+    "preact": "10.11.3"
   },
   "devDependencies": {
-    "@babel/core": "^7.20.12",
+    "@babel/core": "^7.21.4",
```
pwolfert, line 10: Now that we're sharing all the dependencies, I had to standardize a few of the dependencies so we didn't get errors caused by mismatched versions

examples/astro/src/layouts/Layout.astro
```diff
@@ -0,0 +1,35 @@
...
+	</body>
+</html>
+
+<style is:global>
+	body {
+		margin: 0;
+	}
+
+	h1, h2, h3, h4, h5, h6 {
+		margin-block-start: 1.5em !important;
```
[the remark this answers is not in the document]
pwolfert: Updated it in [4fc6578](https://github.com/CMSgov/design-system/pull/2938/commits/4fc657864e11022a7a48e228a5caa75377132ab5)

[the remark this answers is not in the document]
zarahzachz: i see it now lol

could we add padding-block-start: 1.5em to get around this? or does each component have a different margin size?

examples/astro/src/components/react/AccordionExample.tsx
zarahzachz, line 1: I think Accordion styles are broken for me - there's a gap between the Accordion heading and body panel

examples/astro/src/components/react/ModalDialogExample.tsx
zarahzachz, line 1: Seeing weird behavior with modal - it I open/close it more than once it jumps everything to the top of the page and seems to lock everything up (can't scroll)

examples/astro/src/components/react/PaginationExample.tsx
zarahzachz, line 1: Pagination example doesn't appear to work for me

zarahzachz:
Building cdn-html-css and saw an issue with Inline Error appearing inline with a form label. I think it hasn't been updated with the revised form error code?

zarahzachz:
The adding machine doesn't work in cdn-web-components

examples/react-app/src/scripts/index.js
```diff
@@ -1,33 +1,43 @@
...
+      <UsaBanner />
+      <header className="ds-base--inverse ds-u-padding-y--3">
+        <div className="ds-l-container">
+          <span className="ds-text-heading--xl">ExampleWebsite.gov</span>
+        </div>
+      </header>
+
+      <div className="ds-l-container ds-u-padding-top--2">
+        <div className="ds-u-measure--base">
+          <h1 className="ds-h1">React-app example</h1>
```
zarahzachz, line 21: deprecated css

zarahzachz (commented):
everything built, but saw some funny stuff here and there

examples/astro/src/components/react/AccordionExample.tsx
pwolfert: Oh right, if you use `ds-content`, the accordion heading gets messed up 😬

pwolfert: Addressed in https://github.com/CMSgov/design-system/pull/2942

examples/astro/src/components/react/ModalDialogExample.tsx
pwolfert: Updating to the new API fixed it. I thought we had fixed the bugs with the old APIs, though, so this is a bit troubling.

zarahzachz (approved):

PR #2939, opened 2024-02-14
[WNMGDS-2663] Add defaultProp for spinner aria-valuetext

zarahzachz:
WNMGDS-2663

Removed `aria-valuetext` prop from the component logic and added as a proper `defaultProp` to the component so it would show in our Storybook documentation. 

One issue with this (that I discussed with Patrick) was our `t()` doesn't execute and evaluates as a string in the Storybook prop table. We decided this was OK behavior as it shows a default exists despite it not executing the actual default text.


pwolfert:
Looks like the snapshots just need to be updated with `yarn browser:test:storybook-docs:update`

pwolfert (approved):
LGTM 👍 . We're not going to battle a Storybook deficiency when it comes to showing an accurate value. It's good enough to know that there _is_ a default value.

PR #2940, opened 2024-02-14
[WNMGDS-2669] Replace `currentColor` values in tokens

pwolfert:
## Summary

WNMGDS-2669

Figma doesn't know what to do with `currentColor`, so we would need to write code to calculate what those values are when we go to sync, or we need to replace those values with static color values. The only problem with that is that it isn't always straightforward what the static value should be, because `currentColor` is inherently adaptive and dynamic. Turns out, though, that the only outliar is a token that we're not even using! If we had been using it, we determined that we could get rid of it, but we aren't even using it. Once we removed that token, the others were simple to replace.

- Removed unused `--button-icon__fill` variable
- Replaced all other instances of `currentColor` with static values from our theme colors

## How to test

## Checklist


zarahzachz (approved):
I'm seeing 11 VRT issues when running `yarn test:browser:all` locally. These could be false errors or something (can't see what the difference actually is?) so I'm approving

11 failed VRTs

pwolfert:
I ran these locally myself, but I'll run them again to double check.

pwolfert:
I'm not getting any local changes. I wonder why you are. We'll see what the nightly build says.

PR #2945, opened 2024-02-20
[WNMGDS-2667] Separate color utils into their own module and add to them

pwolfert:
## Summary

WNMGDS-2667

I've separated these changes to make the next PRs less noisy. The color utils are generally useful by many different modules, but the other utils in this file were specific to the token export. I'm separating color utils into their own module and adding some things that will be used by our Figma-syncing code.

## How to test

Should be able to do all the token stuff and build the doc site without any problems.

## Checklist


packages/design-system-tokens/src/lib/colorUtils.ts
```diff
@@ -4,6 +4,55 @@
...
 
+export interface RgbaObject {
+  r: number;
+  g: number;
+  b: number;
+  a?: number;
+}
+
+export function rgbToHex({ r, g, b, a }: RgbaObject) {
+  if (a === undefined) {
```
zarahzachz, line 15: could you remove this conditional by setting a default value in the argument? `export function rgbToHex({ r, g, b, a = 1 }: RgbaObject) {}`

zarahzachz (approved):
looks good. one small comment

packages/design-system-tokens/src/lib/colorUtils.ts
pwolfert: Yeah, that's a good change. This came verbatim from Figma's example project

PR #2947, opened 2024-02-21
[WNMGDS-2681] Fix instance of static id in Dialogs

pwolfert:
## Summary

WNMGDS-2681

Fixed an id in Dialog that was static and caused duplicates now that we can actually render multiple dialog elements to the DOM at once without showing them after the recent API change to `Dialog`.

## Checklist

### DIFF
--- a/packages/design-system/src/components/Dialog/Dialog.tsx
+++ b/packages/design-system/src/components/Dialog/Dialog.tsx
@@ -104,6 +104,7 @@ export const Dialog = (props: DialogProps) => {
   const rootId = useId('dialog--', id);
   const headingRef = useDialogAnalytics(props);
   const headingId = `${rootId}__heading`;
+  const contentId = `${rootId}__content`;
 
   const dialogClassNames = classNames('ds-c-dialog', className, size && `ds-c-dialog--${size}`);
   const headerClassNames = classNames('ds-c-dialog__header', headerClassName);
@@ -152,7 +153,7 @@ export const Dialog = (props: DialogProps) => {
           />
         </header>
         <main role="main" className="ds-c-dialog__body">
-          <div id="dialog-content">{children}</div>
+          <div id={contentId}>{children}</div>
           {actions && <div className={actionsClassNames}>{actions}</div>}
         </main>
       </div>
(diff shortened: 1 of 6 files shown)


packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -152,7 +153,7 @@ export const Dialog = (props: DialogProps) => {
           />
         </header>
         <main role="main" className="ds-c-dialog__body">
-          <div id="dialog-content">{children}</div>
+          <div id={contentId}>{children}</div>
```
zarahzachz, line 156: this looks good but i wonder if we need to do a wider update of the dialog html. i'm mostly curious about the inclusion of a `role="document"` and `<main>` tags here. sites should only have 1 of these per page, and it's said [`main` shouldn't be hidden](https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element) - does that apply to the dialog `open` attr?

zarahzachz: i did some local poking around i i think we can remove the div with `ds-c-dialog__window`, apply the bulk of those styles to the `<dialog>` itself and move most of the props onto `NativeDialog` (`ref` needs some massaging but everything else works fine). 

we can also remove the `main` tag and change the div with the `actionsClassNames` to a `footer` for proper semantics.

i think this markup was to support pre-native dialog times and could use a facelift (which is out of scope here, but a future todo imo)

zarahzachz: another issue i'm seeing (and it might not be an issue) is our current implementation seems to override which element gets default focus. could be the result of the `tabindex=-1` attr, but if you remove a lot of the excess tags and get down to a basic dialog structure, the close button receives default focus ([which seems to be the pattern established by browsers](https://web.dev/learn/html/dialog#:~:text=By%20default%2C%20when%20a%20dialog,when%20the%20dialog%20is%20opened.)).

zarahzachz: if we like these update ideas, it would need to apply to drawer as well

zarahzachz (approved):
looks good, added comments for other future work

packages/design-system/src/components/Dialog/Dialog.tsx
pwolfert: Feel free to create a demo of what you're talking about. I agree that it's out of scope for this fix (and we'd probably want to wait until v10), but we could create a ticket out of the demo. Remember that we also hacked the structure a bit to make sure the dialog element itself took up the whole page so it wouldn't cut off things like dropdowns.

PR #2949, opened 2024-02-22
[WNMGDS-2673] Audit and update default props

zarahzachz:
- Moved inline default prop values to `defaultProps` to ensure props table populates with something
- Updated Storybook props table


packages/design-system/src/components/VerticalNav/VerticalNavItemLabel.tsx
```diff
@@ -55,8 +55,8 @@ export const VerticalNavItemLabel = (props: VerticalNavItemLabelProps): React.Re
 
   let otherProps;
   if (LabelComponent === 'button') {
-    const collapsedLabel = props.ariaCollapsedStateButtonLabel ?? t('verticalNav.expand');
-    const expandedLabel = props.ariaExpandedStateButtonLabel ?? t('verticalNav.collapse');
+    const collapsedLabel = props.ariaCollapsedStateButtonLabel;
+    const expandedLabel = props.ariaExpandedStateButtonLabel;
```
pwolfert, line 59: These could be inlined in the one place they're used now

pwolfert (commented):
All these changes look correct, but I just found a problem that I don't have an immediate solution for: Unfortunately by evaluating these `t` function calls when the module loads, we can no longer update the language dynamically and expect the default content to change languages.

packages/design-system/src/components/VerticalNav/VerticalNavItemLabel.tsx
zarahzachz: i just realized this is unnecessary to include - this sets the `title` attribute on a button and titles aren't accessible so this does nothing and should be removed

packages/design-system/src/components/Spinner/Spinner.tsx
```diff
@@ -47,14 +47,15 @@ export const Spinner: React.FunctionComponent<SpinnerProps> = (props: SpinnerPro
 
   return (
     <span className={className} role={props.role}>
-      <span className="ds-u-visibility--screen-reader">{props['aria-valuetext']}</span>
+      <span className="ds-u-visibility--screen-reader">
+        {props['aria-valuetext'] ?? t('spinner.ariaText')}
```
zarahzachz, line 51: reverting the spinner update - dynamically updating prop isn't supported when defined as defaultProp

packages/design-system/src/components/ThirdPartyExternalLink/ThirdPartyExternalLink.tsx
```diff
@@ -66,8 +66,4 @@ const ThirdPartyExternalLink = ({
   );
 };
 
-ThirdPartyExternalLink.defaultProps = {
-  learnMoreUrl: '',
```
zarahzachz, line 70: i don't know why this was an empty string but it seems like if we were to define a default value, it should be useful

packages/design-system/src/components/ThirdPartyExternalLink/ThirdPartyExternalLink.tsx
```diff
@@ -12,7 +12,7 @@ interface ThirdPartyExternalLinkProps {
   children: string;
   /** Additional classes to be applied to the external link button. */
   className?: string;
-  /** Specify the URL User's should visit to learn more about your application's external link policy. */
+  /** Specify the URL users should visit to learn more about your application's external link policy. */
   learnMoreUrl?: string;
```
zarahzachz, line 16: should this be a required prop?

pwolfert (approved):
LGTM 👍

PR #2950, opened 2024-02-22
[NO-TICKET] Make accordion SVG hidden to AT users

zarahzachz:
Icons used in AccordionItem are decorative and the button uses `aria-expand` to announce if the button's open or not, so removing the SVG title and associated i18n text.


pwolfert (approved):
Tested with VoiceOver, and it sounds good to me 👍

PR #2951, opened 2024-02-23
[WNMGDS-2699] Remove title attr from vert nav label component, unhelpful attr.

zarahzachz:
WNMGDS-2699

Removing the `title` attr and associated props and prop values from VerticalNavLabel.

Why? `title` attr are generally considered inaccessible due to the fact that they are only interactive for pointer users and screen readers cannot access the value. On top of that, these buttons use `aria-expanded` which announce whether the button is open or closed. So aside from the `title` being inaccessible, it's also redundant.


pwolfert:
Does this have a ticket associated with it now?

zarahzachz:
> Does this have a ticket associated with it now?

It does now

pwolfert (approved):
Still works great! Nice deletions!

PR #2952, opened 2024-02-23
[WNMGDS-2746][WNMGDS-2776] Update token build scripts to use JSON source

pwolfert:
## Summary

Along with the Figma migration, we're implementing a new token storage method based on JSON ([see this pull request for background info](https://github.com/CMSgov/design-system/pull/2946)). Our previous scripts for exporting our token data into CSS, SCSS, CSV, and JSON are therefore outdated and need to be rewritten to take the new JSON format as input.

- Removed support for exporting tokens to CSV
    - This was helpful when we didn't have any other tools for viewing token data in table format, but now we have Figma's variable editor tables, and they are actually really handy. We can always re-implement this in the future if we need it.
- Removed support for exporting tokens to JSON
    - Our tokens are now natively in JSON, and we were previously exporting to this format to make it easier to import them into Sketch. We don't need to support that old Sketch code anymore.
- Rewrote the exporters for CSS and SCSS to use the JSON tokens as input
- Updated Figma-syncing code to support peculiarities with translating between CSS-based _dimensional_ units and Figma's generic _number_ type
- Updated the package README and included an explanation of our token structure from a high-level standpoint

This is a (pull) request to merge into a feature branch, as this feature is not completely ready to merge into the main branch.

### CSS changes to address:

Running `yarn test:unit -u packages/design-system-tokens` will show some differences between the new generated CSS-variable files and the ones generated by the previous tokens system. The following are outstanding changes to our CSS variables that either need to be accepted with VRT changes or need to be fixed in some way:

- [ ] Addition (or failure of removal) of `--autocomplete*` variables
- [ ] Change to CMS.gov badge radii
- [ ] Missing `--button__font-weight` in Figma
- [ ] Figure out how to use `--dropdown-option-group__padding` and `--dropdown__background-color` in the CSS
- [ ] Address all font family and font weight variable changes
- [ ] Use new `--global__color*` variables in the reset file
- [ ] Find out why `--label__*` variables disappeared
- [ ] Stop using `--shadow-base` in CSS
- [ ] Investigate missing `--usa-banner` variables

## How to test

1. `yarn build`
2. `yarn test:unit`
3. `yarn storybook`
4. Note that because of the list of CSS changes that still need to be addressed, running the VRTs is probably going to produce a large number of errors
5. Note also that the doc site does not build because it still references old token modules
6. Try adding new variables or deleting them in Figma and seeing if they show up after syncing
   - Note that this branch still references a testing Figma file rather than the true production file. I'll need to change this in the `.env.example` file in a future PR


packages/design-system-tokens/src/css/__snapshots__/css-snapshots.test.ts.snap
```diff
@@ -0,0 +1,2056 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`CSS snapshots variables in cmsgov-theme.css match snapshot 1`] = `
```
pwolfert, line 3: Note that these represent the variables generated by the old token system. I just want to check this into our feature branch so we have something to compare against as we address the CSS changes

packages/design-system-tokens/src/lib/conversions.ts
```diff
@@ -0,0 +1,33 @@
+const PIXELS_PER_REM = 16;
+const PIXELS_PER_EX = 8.5625;
```
zarahzachz, line 2: Curious about where this unit comes from? It's my understanding that `ex` units vary based on the font-size, so it seems an `ex` unit in one theme may be different in another.

zarahzachz (approved):
I tested it locally and changed from Figma populated in code just great! Good job!

Question - should the /design-system-tokens/sketch dir be deleted as part of this work or in a followup ticket?

packages/design-system-tokens/src/css/__snapshots__/css-snapshots.test.ts.snap
pwolfert: There's a comment in the test module that generates it, but it's not marked as a TODO. I just commented that it can be deleted as soon as it's causing more trouble than it's worth

packages/design-system-tokens/src/lib/conversions.ts
pwolfert: Shoot, I think you're right. It really just comes from working backwards from the values we had to have in Figma. The designs use static pixel values for these, so if it has a constant here to convert it back into one `ex` value, it just means that in the browser it might have slightly different widths than the designs depending on what theme you're using. In my mind, it's not that bad, but I should definitely explain this in the comment if nothing else.

packages/design-system-tokens/src/css/translate.ts
```diff
@@ -0,0 +1,230 @@
...
+    //   renderVariableAlias &&
+    //   (aliasedTokenName in themeTokens || isIncludedSystemToken(aliasedTokenName))
+    // ) {
+    //   return renderVariableAlias(aliasedTokenName);
+    // }
+
+    return resolveTokenValue(aliasedToken, resolveConfig);
+  } else if (token.$type === 'number' || token.$type === 'fontWeight') {
+    return token.$value + '';
+  } else if (typeof token.$value === 'string') {
```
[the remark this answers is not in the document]
pwolfert: This block is for if it was a string on line 58 but wasn't an alias. I guess I could have nested this condition as an `else` inside a starting `if (typeof token.$value === 'string')` block, but I wanted to keep all the different cases as shallow as possible.

pwolfert:
Oh yeah, we should go ahead and delete the sketch folder, but I'll do it in a follow-up PR.

PR #2953, opened 2024-02-26
[WNMGDS-2673] Deprecate scrollableNotice prop

zarahzachz:
WNMGDS-2673

The `defaultProps` for Table has an Alert component with specific text and CSS classes applied. Discussed with Patrick and it seems this was an over-optimization situation; Would be better to hard-code the styled Alert and provide a prop to overwrite the Alert text instead of the whole component.

Oh! Also, the default Alert wasn't translated and that needed to be done.

Todos:
- Need official Spanish translation for Alert
- Need guidance for how the Alert should appear on small breakpoints - is this pattern even useful/accessible as it references arrow keys?

### DIFF
--- a/packages/design-system/src/components/Table/Table.tsx
+++ b/packages/design-system/src/components/Table/Table.tsx
@@ -5,6 +5,7 @@ import TableContext from './TableContext';
 import classNames from 'classnames';
 import debounce from '../utilities/debounce';
 import useId from '../utilities/useId';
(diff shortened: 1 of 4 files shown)


pwolfert:
Are we scheduled to discuss this in any meetings?

zarahzachz:
closing because we've decided to refactor tables instead of removing this singular prop

PR #2961, opened 2024-03-02
[WNMGDS-2674] Add VRTs for remaining example

pwolfert:
## Summary

This is a follow-up to https://github.com/CMSgov/design-system/pull/2959 that rounds out the list of examples.

Note: I've had to disable the `SingleInputDateField` and `Tooltip` examples because they both have dependencies that do not support ES Modules. These will need to be addressed by future tech-debt tickets. It will become a blocker for any team using newer build systems that may rely on ES Modules support (like a team migrating to Astro).

## How to test

`yarn test:browser:examples:update` to build and test, and make sure it didn't write any changes

## Checklist


tests/browser/snapshots/examples/astro-react/astro-react--Mobile-Chrome.png
pwolfert, line 1: The major issue I'm trying to troubleshoot right now is that, in the browser window that Playwright loads, the `astro-island > h2` selector doesn't seem to apply. Looks like the rule exists in the stylesheet it loads...

pwolfert: Aha! That version of chromium doesn't support the native css nesting. Just to be safe, I'm going to get rid of that nesting for now. We don't use it anywhere else yet.

zarahzachz (approved):
Code looks good but seeing a modified snapshot (unrelated to Astro):
> tests/browser/snapshots/examples/cdn-web-components/cdn-web-components--chromium.png

But running `npx playwright show-report` shows no failures or anything? Not sure what to make of this result, but it seems unrelated to your Astro work so I'm approving.

pwolfert:
Not sure why that file got updated, but reverting it to what's on `main` seemed to pass twice in a row on Jenkins, and I haven't been able to get it to update again. We can keep an eye on it and see if we have to put some kind of extra time on it so the fonts load.

PR #2970, opened 2024-03-11
[WNMGDS-2696] remove live region attrs from Pagination

zarahzachz:
WNMGDS-2696

Remove live region attributes from Pagination; the component isn't dynamic (would trigger a page reload given it uses links to navigate to different urls) and is a `nav` element (users would understand interacting with it would result in a page change).

I think we added these attrs because Storybook's example _is_ dynamic.

[Link to the Slack discussion here](https://cmsgov.slack.com/archives/CHH0381RD/p1709914079761229).


pwolfert (approved):
It does get rid of the readout

PR #2971, opened 2024-03-11
[WNMGDS-2686] remove aria-labelledby from accordion; not needed

zarahzachz:
WNMGDS-2686

Remove `aria-labelledby` from AccordionItem. This attr is on a `div`, which has the default role of generic and generics cannot accept aria labels as part of the spec. This label is also just not needed. More details in [this Slack thread](https://cmsgov.slack.com/archives/CHH0381RD/p1708710668265149).


pwolfert (approved):
It works a lot better now because it automatically reads the content! Thanks for the testing help

PR #2972, opened 2024-03-12
[NO TICKET] Truncate docs page is crazy wide

zarahzachz:
Fixing page layout for the truncate utility class page: 
Screenshot 2024-03-12 at 2 44 01 PM

Issue seemed to be `.truncate` set `max-width: 100% !important` and `.ds-u-measure--wide` set max-width to about 80ex and both these classes were placed on the same element of text, causing a discrepancy in how the browser resolved the width.

Anyways... seeing `ds-u-measure--wide` placed individually on a code example seemed like a code smell (all our content should be this width, so why the need for the util to begin with?). Ended up removing a lot of single instances where this util was applied and made it so the Layout's page content used it instead (it's only called once). This has the added benefit of fixing the layout shift I discovered for desktop TOC (table of contents shifted vertically depending on the page rendered - now the TOC appears in a more fixed position).


packages/docs/content/migration-guides/migrating-to-new-buttons.mdx
```diff
@@ -6,13 +6,16 @@ intro: We've created a more consistent, scalable, and themable button framework
 
 import oldButtons from '../../src/images/old-buttons.png';
 import newButtons from '../../src/images/new-buttons.png';
-import sketchButtonStrucutre from '../../src/images/Sketch-button-achitecture.png';
+import sketchButtonStructure from '../../src/images/Sketch-button-architecture.png';
```
pwolfert, line 9: Oh wow, two typos in the same line 😅

packages/docs/content/migration-guides/migrating-to-new-buttons.mdx
```diff
@@ -23,19 +26,19 @@ The previous implementation of the design system button architecture varied betw
 There were button types and styles that existed in child design systems only and were not available in the core design system creating a siloed model where updates and improvements to buttons could not be made to the entire system. In addition, the language used for buttons varied throughout the system. Names such as "primary" and "secondary" were confusing and inconsistent because they were not mapped to primary or secondary brand colors.
 
 <img
-  alt="3 columns of buttons that are missmatched across the entire design system. Screenshot."
+  alt="3 columns of buttons that are mismatched across the entire design system. Screenshot."
```
pwolfert, line 29: 👍

packages/docs/src/components/content/ContentRenderer.tsx
```diff
@@ -58,7 +58,7 @@ const PreformattedWithLanguageClass = (props: any) => {
 };
 
 const TextWithMaxWidth = (props: any, Component) => {
-  const className = `ds-u-measure--wide ${props.className || ''}`;
+  const className = props.className ? `${props.className}` : null;
```
pwolfert, line 61: 1) Do we need to put this in a template string now? and 2) Do we actually need this ternary or constant at all now?

pwolfert (approved):
The page layouts seem to work so much better now. I didn't test all the pages, but I did do a spot test on some random pages and on the truncate page.

packages/docs/src/components/content/ContentRenderer.tsx
zarahzachz: i actually don't think we ever used this. what it was doing was applying `class="ds-u-measure--wide"` to EVERY `<p>` in the site and i didn't see a single instance where props we being used to change that.

pwolfert: So could we remove this line entirely is what I was asking

PR #2973, opened 2024-03-12
[WNMGDS-2657] Web components can now read child elements

pwolfert:
## Summary

For a long time we were struggling with the problem of no useful data being present on the React/Preact element passed through the `children` prop. We couldn't use it to traverse because no meaningful `props` property was present, which meant we couldn't inspect `children` and traverse through its virtual dom. I took a closer look at the `parseHtml` and `onConnected` functions of `preactement` and found that the top-level element is always given `{}` for its props! [They've been doing it ever since they introduced parseHtml](https://github.com/jahilldev/component-elements/commit/556058c1b048d33f32dd3a7479abc2e00385c4bd#diff-1ed00a29b9bf1ec50c4ac3b4b786a7833e31f53e16fcb338e5b974d929e59131R169), but I don't know why. I'm sure they had a reason, but our code works now that I've changed it.

This has enabled us to offer HTML as an alternate interface for Dropdown options and will allow us to do similar nesting in the future where we need to analyze the children.

- Fixed the issue in the preactement source
- Added another line of code to the dropdown options parser to account for an idiosyncrasy of the preact/preactement parser
- Added new stories to show off the new capabilities
- Added new unit tests
- Updated the Astro example to use the HTML options

## How to test

1. Test storybook
2. Test the Astro example
3. Run the wc tests

## Checklist


examples/astro/src/pages/web-components.astro
```diff
@@ -32,8 +32,11 @@ import Layout from '../layouts/Layout.astro';
         <ds-dropdown
```
zarahzachz, line 32: dropdown looks weird in my local build

image

zarahzachz (commented):
playing with dropdown WC in storybook and noticed the width of the dropdown was changing based on which tribe was selected.

examples/astro/src/pages/web-components.astro
pwolfert: Did you `yarn build` first?

pwolfert:
Figured out why it was changing widths and will commit a fix

examples/astro/src/pages/web-components.astro
zarahzachz: I ran `yarn build:examples` from root before cd'ing into /astro and running `yarn start`

pwolfert:
Unfortunately `build:examples` assumes you've run `build` on the design
system itself. It only builds the examples statically so that
`serve:examples` will work. You don't need to run `build:examples` if you
are running Astro's `yarn start` because that starts an Astro dev server.

zarahzachz:
Yeah, I ran the build commands first.

2. `yarn build && yarn build:storybook`
3. `yarn storybook` (to test Storybook)
4. `yarn build:examples`
5. `cd examples/astro`
6. `yarn start`

I get the same unstyled WC dropdowns I screenshat. Even cleared cache to confirm it wasn't a browser issue. Is there a different set of commands I need to run?

Not sure if it's related, but the React Astro dropdown example is also not functional (clicking dropdown doesn't open options)

zarahzachz (approved):

PR #2977, opened 2024-03-13
[WNMGDS-1859] Deprecate truncate utility class

zarahzachz:
WNMGDS-1859

* Add `.ds-u-truncate` to deprecation CSS list
* Add comments where class is being used to remove in v11 release
* Remove class from documentation where used
* Move truncate styles to a dropdown-specific class
    * Confirmed with AS truncate is fine for dropdown because users can open it to see what was selected and ARIA was applied to announce the selection to users


pwolfert (approved):
LGTM

zarahzachz:
Met with Tiffany:

- Truncation styles on Dropdown are not an issue if an `aria-label` announces the total text to SR users
- Sighted users will be able to access full length of option text by opening the Dropdown
- Dropdown is limited in how much space it can occupy and it's unreasonable to expect the Dropdown to expand to include multi-lines if the option text warrants it
- Deprecating `.truncate` is OK as long as there is enough of a visual cue to indicate where it's used so teams can talk with UX to find an alternative solution

PR #2982, opened 2024-03-18
[WNMGDS-2700] Deprecate css utils v10

zarahzachz:
WNMGDS-2700

- Remove CSS classes marked for deprecation in v10 (includes font utility classes)
- Update stories, tests and documentation using deprecated classes


pwolfert (approved):
This cut 6KB, which is almost 4% of `index.css`

zarahzachz:
> This cut 6KB, which is almost 4% of `index.css`

Must be all the breakpoint namespaced utils causing that bloat. It didn't seem like I deleted 4% of CSS, but it's gotta be those breakpoint loops causing it for sure!

PR #2983, opened 2024-03-18
[WNMGDS-2704] Remove unused packages

zarahzachz:
- Used `npx depcheck` to identify candidates for removal in repo
- Did a cross search within repo to see if packages were being used internally
- Discovered the following packages were unused and not referenced anywhere, but appear to be dependencies of Storybook (some Stories broke without these packages) so I didn't remove them:
    - "@babel/helper-string-parser": "^7.19.4"
    - "@babel/plugin-proposal-decorators": "^7.21.0"
    - "@babel/plugin-proposal-export-default-from": "^7.18.10"
    - "@babel/register": "^7.8.4"

Ran the following commands without error and spot checked docs and Storybook to see everything worked A-OK

2. `yarn build && yarn build:storybook && yarn build:storybook:docs && yarn build:docs && yarn build:examples`
3. `yarn test`
4. `yarn test:browser:all`

### DIFF
--- a/packages/design-system-tokens/package.json
+++ b/packages/design-system-tokens/package.json
@@ -6,7 +6,6 @@
   "devDependencies": {
     "@types/node": "^17.0.22",
     "ts-node": "^10.4.0",
-    "tslib": "^2.3.1",
     "typescript": "^4.9.5"
   },
   "scripts": {
--- a/packages/ds-healthcare-gov/package.json
+++ b/packages/ds-healthcare-gov/package.json
@@ -46,7 +46,6 @@
   ],
   "dependencies": {
     "@cmsgov/design-system": "9.0.2",
-    "@types/js-cookie": "^3.0.2",
     "@types/react": "^18.2.6",
     "@types/react-dom": "^17.0.10",
     "classnames": "^2.2.5"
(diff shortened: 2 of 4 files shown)


packages/design-system-tokens/package.json
```diff
@@ -6,7 +6,6 @@
   "devDependencies": {
     "@types/node": "^17.0.22",
     "ts-node": "^10.4.0",
-    "tslib": "^2.3.1",
```
pwolfert, line 9: I wasn't sure about this one, but I think you're right that we don't need it anymore. Looks like we don't have [this importHelpers config option](https://github.com/microsoft/tslib?tab=readme-ov-file#usage) anymore, though at one point we might have. It's not like a normal dependency where you can just check for import references to it, and even if we remove it, it's still installed indirectly through a boatload of other dependencies.

packages/ds-healthcare-gov/package.json
```diff
@@ -46,7 +46,6 @@
   ],
   "dependencies": {
     "@cmsgov/design-system": "9.0.2",
-    "@types/js-cookie": "^3.0.2",
```
pwolfert, line 49: 👍

pwolfert (approved):
Sounds like we could possibly clean up those other dependencies you mentioned when we get to the Storybook upgrade/cleanup 👍

PR #2985, opened 2024-03-21
[NO-TICKET] Make new example package private

pwolfert:
## Summary

- Make the new example package private so it doesn't mess up the release
- Make all the example version numbers match core to make things simpler

### DIFF
--- a/examples/astro/package.json
+++ b/examples/astro/package.json
@@ -1,7 +1,8 @@
 {
   "name": "astro-example",
   "type": "module",
-  "version": "0.0.1",
+  "version": "9.0.2",
+  "private": true,
   "scripts": {
     "dev": "astro dev",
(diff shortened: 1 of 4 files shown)


examples/astro/package.json
```diff
@@ -1,7 +1,8 @@
 {
   "name": "astro-example",
   "type": "module",
-  "version": "0.0.1",
```
zarahzachz, line 4: Is the `version` even required if we aren't planning on publishing?

pwolfert: I actually thought this is what we did for `design-system-tokens`, but I guess it isn't. I was going for consistency there, but it's not consistent.

I do think every package file has to have a version number in order to be valid.

zarahzachz (approved):

PR #2996, opened 2024-03-27
[WNMGDS-2717] Reduce dialog polyfill to bare minimum

pwolfert:
## Summary

[WNMGDS-2717](https://jira.cms.gov/browse/WNMGDS-2717)

Shave down the dialog polyfill to only be a shim for SSR or testing environments that don't yet support it. We've dropped support for browsers that don't have built-in support for dialog anyway, and we've only been keeping the polyfill around because it allowed us to continue to run it in older environments that don't have definitions for dialog.

This cuts the `react-components.js` bundle down to `407KB` from `418KB`, which is 2.6%.

### Further exploration to be done later

It would be nice if we could only load the shim module in non-production environments and therefore exclude it from our CDN bundle and from product bundles. It would be nicer if we could do that with TypeScript without involving Babel. How I imagine it would work (what I've done in other projects) is have an intermediary module that says something like

```js
  module.exports = require('the real module');
} else {
  module.exports = require('a fake module that exports an empty function');
}
```

I don't think using the dynamic `import()` function would be right, because it could cause code-splitting in downstream builds.

## How to test

It actually appears that our doc site builds now with or without the shim. It's our unit tests that use [jsdom and therefore don't work](https://github.com/jsdom/jsdom/issues/3294#issuecomment-1118542780).

## Checklist


packages/design-system/src/components/NativeDialog/shim.ts
```diff
@@ -0,0 +1,62 @@
...
+    setOpen(false);
+    const cancelEvent = new CustomEvent('close', {
+      bubbles: false,
+      cancelable: true,
+    });
+    dialog.dispatchEvent(cancelEvent);
+  };
+
+  // Set a11y role. Browsers that support dialog implicitly know this already.
+  if (!dialog.hasAttribute('role')) {
```
zarahzachz, line 41: is this block still needed?

zarahzachz (approved):

packages/design-system/src/components/NativeDialog/shim.ts
pwolfert: I figured if people were trying to query by role in unit tests...actually it's something we do. We currently need it for our unit tests.

PR #3003, opened 2024-04-01
[WNMGDS-2698] Upgrade Storybook

zarahzachz:
WNMGDS-2698

Upgraded Storybook 7 to [Storybook 8](https://storybook.js.org/blog/storybook-8/). Comments will explain what changed and why, as some of our features were deprecated.

I was able to remove the `@storybook/addon-mdx-gfm` package, but ended up adding `@storybook/addon-webpack5-compiler-babel` as part of the migration, so no obvious big perf wins here.


.storybook/preview.tsx
```diff
@@ -156,7 +156,6 @@ const preview: Preview = {
     },
   },
   parameters: {
-    actions: { argTypesRegex: '^on[A-Z].*' },
```
zarahzachz, line 159: `argsTypesRegex` broke the SB migration several times and the only way to continue was in its removal. The [migration guide](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#implicit-actions-can-not-be-used-during-rendering-for-example-in-the-play-function) gets into the thick of it, but it seems [the preferred way to target specific events](https://storybook.js.org/docs/essentials/actions#via-storybooktest-fn-spy-function) in each Story.

packages/docs/content/components/dropdown.mdx
```diff
@@ -89,7 +89,7 @@ The following CSS variables can be overridden to customize Dropdown components:
   - I use the arrow keys to select an option and I hear the selected option is changed.
   - I select an option with the Enter or Space key and hear the selected option.
 
-<!-- [We need to implement this] I use the escape key and I hear it collapse and focus moves to the select -->
+{/* [We need to implement this] I use the escape key and I hear it collapse and focus moves to the select */}
```
zarahzachz, line 92: Original comment syntax broke migrations, so I had to change these files to complete them.

.storybook/main.ts
```diff
@@ -1,17 +1,15 @@
...
-    '@storybook/addon-essentials',
-    '@storybook/addon-mdx-gfm',
-    '@storybook/addon-viewport',
+    getAbsolutePath('@storybook/addon-links'),
+    getAbsolutePath('@storybook/addon-essentials'),
+    getAbsolutePath('@storybook/addon-viewport'),
+    '@storybook/addon-webpack5-compiler-babel',
   ],
-  features: {
-    buildStoriesJson: true,
```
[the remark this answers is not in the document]
pwolfert: Looks like this just became a default option, but they've renamed the `stories.json` file to `index.json` and changed its structure slightly. We rely on it [in storybook-docs.test.ts](https://github.com/CMSgov/design-system/blob/WNMGDS-2698/upgrade-storybook/tests/browser/storybook-docs.test.ts#L3) and [stories.test.ts](https://github.com/CMSgov/design-system/blob/WNMGDS-2698/upgrade-storybook/tests/browser/stories.test.ts#L2). The import file needs to be updated, and the `import { stories as` needs to change to `import { entries as`.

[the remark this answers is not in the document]
pwolfert: Fixed this, but now there are TypeScript errors.

yarn.lock
```diff
@@ -7739,7 +8202,7 @@
   dependencies:
     "@types/react" "*"
 
-"@types/react@*", "@types/react@>=16", "@types/react@^18.2.6":
+"@types/react@*", "@types/react@^18.2.6":
```
pwolfert, line 8205: There are now TypeScript errors, and they seem to be failing on class components (whether in our project or dependencies). It could possibly come from this version change for `@types/react`, but I lack evidence to back that hypothesis. When I ran `yarn why @type/react` in this branch and main, the top result stayed the same. The odd entry present in both of them is this:

```
=> Found "@types/react-dom#@types/react@17.0.59"
```

which is probably due to this strange mismatch in types that we've allowed to persist:

```json
"@types/react": "^18.2.6",
```

pwolfert: Note that changing these entries to the following DOES NOT fix it:

```json
"@types/react": "^17.0.39",
"@types/react-dom": "^17.0.11",
```

zarahzachz:
This upgrade requires more work. Union types change on the args table, which results in flaky VRT results. Waiting on [SB dependency](https://cmsgov.slack.com/archives/CT169N26L/p1720459980488729?thread_ts=1720446538.393959&cid=CT169N26L) to resolve this.

pwolfert (approved):

PR #3010, opened 2024-04-02
[WNMGDS-2698] NPM cleanup

zarahzachz:
WNMGDS-2698

In an effort to clean up NPM packages before switching from Yarn 1 to NPM, I'm cleaning up some packages!

- [x] Style linting
- [ ] ~JS linting~ ESLint and plugins are all minor version bumps, so we're holding off on upgrading.
- [ ] ~Testing~ Starting with v12.1.5, RTL only supports React 18+. We're on RTL 11.2.7, the latest version is 14.2.2. It doesn't make sense to bump testing to v12 if it drops React 17 support a few minor versions in.
- [x] Deps in sub directories


packages/design-system-tokens/package.json
```diff
@@ -3,11 +3,6 @@
   "version": "0.0.1",
   "private": true,
   "description": "CMSDS Design System Tokens",
-  "devDependencies": {
```
zarahzachz, line 6: These devDeps exist in the root package.json. Removing from this file had no obvious impact on token scripts.

packages/design-system/src/styles/_mixins.scss
```diff
@@ -63,8 +64,8 @@
 }
 
 @mixin flexbox-col($size) {
-  flex: 0 0 percentage(math.div($size, $grid-columns));
-  max-width: percentage(math.div($size, $grid-columns)); // IE10+ and Firefox
+  flex: 0 0 math.percentage(math.div($size, $grid-columns));
```
zarahzachz, line 67: Using `math.` prefix was called out by new linter rules

packages/design-system/src/styles/components/_ChoiceList.scss
```diff
@@ -87,9 +86,9 @@ label,
   display: grid;
   grid-area: check;
   height: var(--choice__size);
-  justify-items: center;
   margin: 0;
   place-content: center;
+  place-items: center;
```
zarahzachz, line 91: Linter combined `justify-items` and `align-items` into shorthand property

packages/design-system/src/styles/components/_LabelMask.scss
```diff
@@ -1,4 +1,4 @@
 .ds-c-label-mask {
   color: var(--color-muted);
-  font-family: Menlo, Consolas, Monaco, Liberation Mono, Lucida Console, monospace;
+  font-family: Menlo, Consolas, Monaco, 'Liberation Mono', 'Lucida Console', monospace;
```
zarahzachz, line 3: Can't believe these strings weren't caught before - multi-word font families won't work right unless wrapped in quotes

packages/design-system/src/styles/components/_SingleInputDateField.scss
```diff
@@ -36,8 +36,6 @@
 
 /* Hide elements for devices that are not screen readers */
 .rdp-vhidden {
-  -moz-appearance: none;
```
zarahzachz, line 39: Linter doesn't like prefixes. I believe the autoprefixer package is supposed to handle these in the Gulp file, so this removal makes sense

packages/design-system/src/styles/components/_SingleInputDateField.scss
```diff
@@ -524,7 +526,7 @@
     text-align: center;
     text-decoration: none;
 
-    &:after {
+    &::after {
```
zarahzachz, line 529: I think the linter rule for having double-colons on `::before` and `::after` is just being pedantic. It's technically correct, but the browser also doesn't care. It's easy enough running `yarn lint:stylelint --fix` to clean these sorts of things up though.

packages/design-system/src/styles/components/_SvgIcon.scss
```diff
@@ -90,7 +90,7 @@
 
     path:nth-of-type(1),
     path:nth-of-type(5) {
-      fill: #ffffff;
```
zarahzachz, line 93: A [teeny little super guy](https://www.youtube.com/watch?v=Y8j565lLvJU) of a change - saves a minuscule amount of bytes

package.json
```diff
@@ -26,7 +26,7 @@
     "deploy-demo": "ts-node ./scripts/deployDemo.ts",
     "lint": "yarn lint:stylelint && yarn lint:eslint",
     "lint:eslint": "yarn eslint ./packages/",
-    "lint:stylelint": "yarn stylelint ./packages/**/*.scss --custom-formatter=node_modules/stylelint-formatter-pretty",
+    "lint:stylelint": "yarn stylelint --custom-formatter stylelint-formatter-pretty ./packages/**/*.scss",
```
zarahzachz, line 29: New Stylelint seems to have different syntax for formatter-pretty

package.json
```diff
@@ -101,7 +101,7 @@
     "@storybook/react-webpack5": "^7.1.0-alpha.1",
     "@storybook/theming": "^7.1.0-alpha.1",
     "@testing-library/jest-dom": "^5.10.1",
-    "@testing-library/preact": "^3.2.3",
+    "@testing-library/preact": "^2.0.0",
```
zarahzachz, line 104: Console kept screaming that this package was the wrong version for preact. Downgrading it resolved terminal screams

package.json
```diff
@@ -110,6 +110,7 @@
     "@types/node": "^17.0.22",
     "@typescript-eslint/eslint-plugin": "^5.16.0",
     "@typescript-eslint/parser": "^5.16.0",
+    "autoprefixer": "^10.4.19",
```
zarahzachz, line 113: I'm confused by this - it wasn't part of any package.json file, yet it's called in the Gulp files. Most other packages referenced there were included in this list, but not this one.

After bumping Stylelint and changing the config, `yarn build:core` starting screaming at me about not being able to find this module. Maybe it was part of another package I and it somehow just worked? Either way, adding it here resolved the build issues.

package.json
```diff
@@ -152,11 +152,10 @@
     "sass": "^1.53.0",
     "sass-loader": "10.2.1",
     "storybook": "^7.1.0-alpha.1",
-    "stylelint": "15.10.1",
-    "stylelint-config-recommended": "^7.0.0",
-    "stylelint-formatter-pretty": "^3.1.0",
-    "stylelint-order": "^5.0.0",
-    "stylelint-scss": "^4.2.0",
+    "stylelint": "16.3.1",
+    "stylelint-config-standard-scss": "^13.0.0",
```
zarahzachz, line 156: Using [this package](https://github.com/stylelint-scss/stylelint-config-recommended-scss/blob/master/index.js) instead of the other recommended config helped remove other packages.

This SCSS config included `postcss-scss`, `stylelint-config-recommended`, and `postcssScss` (instead of `postcss-scss`) and I could remove all three of those package in favor of adding this one.

package.json
```diff
@@ -152,11 +152,10 @@
...
+    "stylelint-formatter-pretty": "^4.0.0",
+    "stylelint-order": "^6.0.0",
```
zarahzachz, line 158: I question the need for this package, but kept it since we wanted it at one point so it's gotta have a purpose. I just don't know how valuable it is specifying the order in CSS

packages/design-system/src/styles/components/_Tooltip.scss
```diff
@@ -17,8 +17,9 @@
   text-decoration-style: dotted;
   text-underline-position: under;
 }
+
 // Chrome needs this offset because it handles text-underline-position differently and without this the underline runs into descending letters like g,j,q,p,y.
-@media all and (-webkit-min-device-pixel-ratio: 0) and (min-resolution: 0.001dpcm) {
+@media (resolution >= 0.001dpcm) {
```
zarahzachz, line 22: I don't know why this is targeting webkit when it's a Chrome override

.stylelintrc.js
```diff
@@ -12,42 +11,24 @@ module.exports = {
...
+    'keyframes-name-pattern': null,
     'no-descending-specificity': null,
-    'no-invalid-position-at-import-rule': null,
+    'selector-class-pattern': null,
+    'value-keyword-case': ['lower', { ignoreKeywords: ['currentColor', 'optimizeSpeed'] }],
     'order/order': ['custom-properties', 'declarations', 'rules'],
     'order/properties-alphabetical-order': true,
-    'scss/dollar-variable-no-missing-interpolation': true,
-    'scss/selector-no-redundant-nesting-selector': true,
+    'scss/at-extend-no-missing-placeholder': null,
```
zarahzachz, line 29: https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/at-extend-no-missing-placeholder/README.md

.stylelintrc.js
```diff
@@ -12,42 +11,24 @@ module.exports = {
...
+    'scss/at-mixin-pattern': null,
```
zarahzachz, line 30: https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/at-mixin-pattern/README.md

.stylelintrc.js
```diff
@@ -12,42 +11,24 @@ module.exports = {
...
+    'scss/comment-no-empty': null,
```
zarahzachz, line 31: https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/at-mixin-pattern/README.md

.stylelintrc.js
```diff
@@ -12,42 +11,24 @@ module.exports = {
...
+    'scss/dollar-variable-empty-line-before': null,
```
zarahzachz, line 32: https://github.com/stylelint-scss/stylelint-scss/blob/master/src/rules/dollar-variable-empty-line-before/README.md

.stylelintrc.js
zarahzachz: Ran into this as an issue where we were [`@extend` on heading classes](https://github.com/CMSgov/design-system/blob/c8c19efe72521d2ced80c4773af210d955411390/packages/docs/src/styles/pages/layout.scss#L43). Linter prefers the `%` prefixed classes to reduce amount of CSS generated

.stylelintrc.js
zarahzachz: Defaulted to wanting all mixins written in kebab case, but we don't really have a standard for how those are written.

.stylelintrc.js
zarahzachz: We have block comments that include a line break within, so this is just to accommodate that.

.stylelintrc.js
```diff
@@ -12,42 +11,24 @@ module.exports = {
...
-      ],
-      { severity: 'warning' },
-    ],
-    // Importants should only be used in our utility classes if possible
+    'comment-empty-line-before': null,
+    'custom-property-pattern': null,
+    // `!important` should only be used in our utility classes if possible
     'declaration-no-important': [true, { severity: 'warning' }],
-    'function-no-unknown': [true, { ignoreFunctions: ['percentage', 'math.div'] }],
+    'keyframes-name-pattern': null,
```
zarahzachz, line 23: Linter wanted keyframe styles as kebab case names, but we don't currently have a standard on that. Can revisit later if needed.

.stylelintrc.js
```diff
@@ -12,42 +11,24 @@ module.exports = {
...
     'no-descending-specificity': null,
-    'no-invalid-position-at-import-rule': null,
+    'selector-class-pattern': null,
+    'value-keyword-case': ['lower', { ignoreKeywords: ['currentColor', 'optimizeSpeed'] }],
```
zarahzachz, line 26: Linter wanted to change many CSS values to lowercase, but I don't know if we wanted to do that so I added exceptions

.stylelintrc.js
```diff
@@ -12,42 +11,24 @@ module.exports = {
...
-  plugins: ['stylelint-scss', 'stylelint-order'],
+  plugins: ['stylelint-order'],
   reportDescriptionlessDisables: true,
-  reportNeedlessDisables: true,
   reportInvalidScopeDisables: true,
+  reportNeedlessDisables: true,
   rules: {
-    // Turning this off in favor of 'at-rule-allowed-list'
-    'at-rule-no-unknown': null,
-    'at-rule-allowed-list': [
```
zarahzachz, line 22: Pretty sure the SCSS config handles all these now

.stylelintrc.js
```diff
@@ -12,42 +11,24 @@ module.exports = {
...
-      { severity: 'warning' },
-    ],
-    // Importants should only be used in our utility classes if possible
+    'comment-empty-line-before': null,
+    'custom-property-pattern': null,
+    // `!important` should only be used in our utility classes if possible
     'declaration-no-important': [true, { severity: 'warning' }],
-    'function-no-unknown': [true, { ignoreFunctions: ['percentage', 'math.div'] }],
+    'keyframes-name-pattern': null,
     'no-descending-specificity': null,
```
zarahzachz, line 24: I think this is worth investigating later. We have a lot of CSS written in incorrect order that screws with specificity, e.g., referring to `.element:first-child` before defining `.element` - the `:first-child` selector technically has higher specificity but it's in a higher place in the cascade and could cause issues.

I tried enabling this rule and correcting what I saw broken, but we have too many files with too complex CSS to do this quickly

packages/design-system/src/styles/components/_Tooltip.scss
zarahzachz: This is smelly to me and deserves a deeper look with design. I'm not sure what the intention is because browsers have `text-decoration-skip-ink` to control how the underline flows with font descenders. And why is Chrome singled out? Just smelly...

zarahzachz:
Stylelint changes might've saved 4kb? That seems too good to be true

package.json
pwolfert: Nice find 👍

package.json
pwolfert: Honestly I don't love this lint rule. While it's kind of nice to see them in alphabetical order to find properties, it's actually kind of arbitrary, and there are good reasons to go against this in many cases.

packages/design-system-tokens/package.json
pwolfert: I actually think that if it's used by a script _in this package_ it should stay a dependency of this package even if it's a duplicate. Downside to that is we need to try to keep the version numbers in sync.

packages/design-system/src/styles/components/_ChoiceList.scss
pwolfert: 😮

packages/design-system/src/styles/components/_LabelMask.scss
pwolfert: 😮

packages/design-system/src/styles/components/_SingleInputDateField.scss
pwolfert: I thought at one point single colon had better browser support or something. Agree this is pedantic but neither do I care 😆

packages/design-system/src/styles/components/_TextField.scss
```diff
@@ -7,6 +7,7 @@
   border-radius: var(--text-input__border-radius);
   color: var(--text-input__color);
   display: block;
+  line-height: var(--text-input__line-height); // Needed to overwrite user agent heading tag styles
```
pwolfert, line 10: How did you discover this one?

packages/design-system/src/styles/components/_Tooltip.scss
pwolfert: Chrome used to use webkit, but is that that old? It can't be from that time because I looked it up, and the switch from WebKit happened in 2013 I think.

Curious if the VRTs will detect any change on this.

packages/docs/src/styles/index.scss
```diff
@@ -3,38 +3,38 @@
...
+@import 'components/maturityChecklist';
+@import 'components/navigation';
+@import 'components/pageHeader';
+@import 'components/responsiveExample';
+@import 'components/storybookExample';
+@import 'components/syntaxHighlighting';
+@import 'components/textColorList';
+@import 'components/tableOfContents';
+@import 'components/spacingUtilityExampleList';
+@import 'components/video';
```
pwolfert, line 30: 🤷‍♂️

packages/ds-medicare-gov/src/styles/components/_HelpDrawer.scss
```diff
@@ -48,10 +50,11 @@ $mct-drawer-close-button-size: 26px;
   &.ds-c-help-drawer__toggle {
     &:hover {
       .icon-wrapper {
-        &:before {
-          background-color: rgba(var(--drawer-toggle__background-color--hover--inverse), 0.8);
+        &::before {
+          background-color: rgb(var(--drawer-toggle__background-color--hover--inverse) 0.8);
```
pwolfert, line 54: Good to know this is the current syntax

packages/docs/src/styles/index.scss
zarahzachz: Oh, SCSS doesn't need file extensions when importing

packages/design-system/src/styles/components/_TextField.scss
zarahzachz: This was an artifact me me trying to resolve all the descending specificity lint rules. It was big mad we were defining ds-c-field styles (it was combined with the ds-c-field-mask rule) after `:-moz-ui-invalid` and `:disabled` were defined, so I split the two up and put line height with the rest of the ds-c-field styles. I could revert it, but it seemed nice having all styles associated with a given class in 1 section.

packages/design-system/src/styles/components/_Tooltip.scss
pwolfert: VRTs show no change, but I really wonder if we even need this. Even though they didn't change, the snapshots for the inline trigger in Safari and Firefox look kind of weird. The dotted lines are way too dense, and the line is really far away from the text. It actually looks like in these old snapshots that this media query was applying to those browsers even though I don't see it happening when I try it in those browsers myself. With your change, the query applies to those browsers, but those snapshots were taken when it wasn't supposed to.

zarahzachz: It's a hardware query and not a browser query though, right? If the hardware supports a resolution over that .001dpcm amount (which is a fraction of a dpi) it kicks in. So I don't know if VRT would capture this style change because... what is the dpi of a docker container? Or when it runs a job in Jenkins? Does it use our hardware to determine this stuff? 

That's why this is so darn smelly. It leads me to think this dotted style needs work. It's too brittle. 

If we tackle the Tooltip work, maybe this can be part of that?

pwolfert: I think until we get to the bottom of this, we shouldn't necessarily expand it beyond webkit. This is something to take to our designers.

pwolfert:
> Stylelint changes might've saved 4kb? That seems too good to be true

Are you looking at `dist/css/index.css`? I'm seeing 156KB on both this branch and `main`.

.stylelintrc.js
pwolfert: Follow-up ticket?

.stylelintrc.js
pwolfert: Weird, I didn't even know that lowercase `currentcolor` was legal

.stylelintrc.js
pwolfert: Not such a big deal when it happens in our doc site vs our public packages

.stylelintrc.js
pwolfert: Yeah, this and the one above it seem like picky rules

packages/design-system/src/styles/components/_TextField.scss
pwolfert: Oh I see now 👍

package.json
zarahzachz: I can remove it if we're in agreement. 

Current config has order doing the following:

- alphabetizing rules
- ordering styles by custom properties > declarations > rules 
  - custom properties are CSS vars
  - declarations are like `display: block`
  - rules are nested rules (`a { span {} }`)
  - example of how this impacts css:
```css 
.button {
  // custom props
  --bg: red;
  --c: white;
  
  // declarations
  background-color: var(--bg);
  color: var(--c);

  // new rule
  svg {
    [declarations here]
  }
}
```

I personally don't see a benefit to alphabetizing and already write my CSS the way the order's specified in the 2nd bullet point.

packages/design-system/src/styles/components/_Tooltip.scss
zarahzachz: Should I make a ticket for this? Or add it to our coworking agenda?

.stylelintrc.js
pwolfert: If it's not too much effort to fix the others...I'd be curious to see if any bytes are actually saved

.stylelintrc.js
zarahzachz: WNMGDS-2731

packages/design-system/src/styles/components/_Autocomplete.scss
```diff
@@ -5,7 +5,7 @@
 // Autocomplete styles extend Dropdown styles
 
 .ds-c-autocomplete {
-  @extend .ds-c-dropdown;
+  @extend %dropdown;
```
pwolfert, line 8: For some reason this just doesn't work, and I haven't been able to figure out why. Sass never ends up creating a `.ds-c-autocomplete` rule in Storybook, though it seems to exist in the CSS file generated by the core build. Like you saw locally, it doesn't matter whether Storybook is built statically or served dynamically; the Sass just doesn't match. Until we can figure this out, my recommendation is to revert the changes and allow `@extend` in the config.

pwolfert (approved):

PR #3017, opened 2024-04-04
[RELEASE] 4/4 Version bump main

pwolfert:


versions.json
```diff
@@ -1,5 +1,6 @@
 {
   "design-system": [
+    "10.0.0-beta.3",
```
zarahzachz, line 3: should i make a ticket to clean up the older minor/patch versions we show? e.g., we don't need to show versions 8.0.0-8.0.4 since 8.0.5 has all the changes in it

zarahzachz (approved):

PR #3019, opened 2024-04-04
[NO-TICKET] Organize web components into folders

pwolfert:
## Summary

Just giving the web components the same treatment as we do the react components by putting them in their own folders. There are going to be more of them now, so that root folder is getting big.

## How to test

`yarn storybook`


zarahzachz (approved):
This works well. 

Noticed something that's unrelated to this change: If you visit Stories for Alert, Badge, Button and change the themes, the children are duplicated. I can't recall if this is something we were aware of

pwolfert:
> Noticed something that's unrelated to this change: If you visit Stories for Alert, Badge, Button and change the themes, the children are duplicated. I can't recall if this is something we were aware of

I did not notice that. I might be able to mix the theme into the stringified key, and that should fix it.

pwolfert:
@zarahzachz, I'm going to address the theme-switching issue in a follow-up PR

PR #3028, opened 2024-04-08
[WNMGDS-2707] Refactor tables

zarahzachz:
WNMGDS-2707

To make more accessible tables, we decided to remove the option to allow for a scrollable layout and default tables to use the stacked layout for mobile. 

There are still design questions:
- How does stackable layout work with [multi-header tables](https://design.cms.gov/storybook/?path=/story/components-table--multi-header-table&globals=theme:core)?
- Do we want to allow for a "stackable only" layout? E.g., table only appears as stackable regardless of viewport size.
 

Need to talk with a11y in the next huddle to figure out correct implementation for stackable labels.


packages/design-system/src/components/Table/Table.stories.tsx
```diff
@@ -40,7 +40,7 @@ const tableTemplateData = {
         'Statement adopted by the Continental Congress declaring independence from the British Empire.',
       links: (
         <a href="https://billofrightsinstitute.org/founding-documents/declaration-of-independence/">
-          https://billofrightsinstitute.org/founding-documents/declaration-of-independence/
+          Declaration of Independence
```
zarahzachz, line 43: I thought our table styles weren't responsive but discovered the issue was in the text string didn't break correctly when a naked link was used. Rather than adding styles to fix this example, I fixed the example to work with our styles.

packages/design-system/src/components/Table/Table.stories.tsx
```diff
@@ -140,55 +138,48 @@ export const Default: Story = {
...
-              ))}
-            </TableRow>
-          ))}
-        </TableBody>
-      </Table>
-    );
-  },
-};
+// This UI doesn't seem to work with stackable layout
+// export const MultiHeaderTable: Story = {
```
zarahzachz, line 142: Commented this out because I have questions... the multi-header table seems very different from the default story; it's almost like a native table example that also uses our components? I'm seeing a mix of things like `<col>` and `<TableRow>`... Is this more of a pattern than a component?

The other question is... now that we force the stackable layout for small breakpoints it kinda breaks this entire example. Looking at [PROD](https://design.cms.gov/storybook/?path=/story/components-table--multi-header-table&globals=theme:core) and I'm not entirely sure how this pattern could work with the stacked layout. 

What to do?

packages/design-system/src/components/Table/Table.test.tsx
```diff
@@ -43,34 +43,10 @@ describe('Table', function () {
     expect(screen.getByRole('table')).toHaveClass('ds-c-table--striped');
   });
 
-  it('accepts custom id', () => {
```
zarahzachz, line 46: Custom `id` seems to be used to tie the caption to the scrollableNotice, but since that was removed, the `id` served no purpose

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -65,11 +47,6 @@ interface BaseTableProps {
    * A striped variation of the table.
    */
   striped?: boolean;
-  /**
-   * Disables the warning message on development console when a responsive stackable table cell does not contain an `id` or `headers`.
-   * It's recommended that accessibility with screen readers is tested to ensure the stacked table meets the requirement.
-   */
-  warningDisabled?: boolean;
```
zarahzachz, line 72: I think I have a TS question here... TableCell has optional props (`header` and `id`) that are actually required, but only if another prop is either a `td` or a `th`. I've heard there a way to create a specific type that fits this circumstance (I think it's called a [conditional type](https://www.typescriptlang.org/docs/handbook/2/conditional-types.html)?) and I wonder if it's a better approach than having these warning props?

Something like:

```js
interface BaseTableCellProps {
  component?: 'th' | 'td';
}

interface TDProps {
  header: string;
}

interface THProps {
  id: string;
}

type TableCellProps = BaseTableCellProps extends 'th' ? THProps : TDProps;
```

I'm not sure if this is a better implementation or not, so discussion would be appreciated. It just seems we have a giant bucket of props that only apply if certain other props are set and it might be cleaner to split them up?

packages/design-system/src/components/Table/TableCell.tsx
```diff
@@ -43,10 +41,6 @@ interface BaseTableCellProps {
    * component is `TableHead` to identify the header cell is a header for a column.
    */
   scope?: TableCellScope;
-  /**
-   * Additional classes to be added to the stacked Title element.
-   */
-  stackedClassName?: string;
```
zarahzachz, line 49: This prop wasn't being used?

packages/design-system/src/components/Table/Table.tsx
```diff
@@ -90,111 +67,36 @@ export const Table = ({
...
-    const debounceHandleResize = debounce(handleResize, 500);
-    window.addEventListener('resize', debounceHandleResize);
-    return () => {
-      window.removeEventListener('resize', debounceHandleResize);
-    };
-  }, [setScrollActive, scrollable]);
-
   const classes = classNames(
-    'ds-c-table',
+    `ds-c-table ds-c-${stackableBreakpoint}-table--stacked`,
```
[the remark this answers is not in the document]
zarahzachz: Syke! After sleeping on this, I think we should keep this prop. My argument is: We don't know how many rows of data a table could have. If a table has many columns of data, the breakpoint needed for the stacked layout could be larger than that of a table with fewer cols.

PR #3030, opened 2024-04-09
[WNMGDS-2740] Add local focus styles to USA banner 

zarahzachz:
WNMGDS-2740

Didn't consider how USA banner would be positioned/laid out when doing initial dev work. USA banner will always abut top, left, and right sides of browser viewport, so focus styles on mobile get cut off by the viewport. Similarly, USA banner's also always above the global header of a product, so that header element cuts off the bottom focus styles of banner's mobile layout. 

The quick solution is to apply a `z-index` on the banner to show the focus styles above neighboring elements, but I wonder if a better solution could be creating an inset version of focus styles to be applied only to these scenarios? 

I noticed our skip link has a similar issue regarding it always being positioned near the edges of the browser and I think a better focus style for these kinds of elements might be if the focus style appears within the confines of the element instead of outside of it. This would require design input, developing a proof of concept and stakeholder approval though, so it's not something to be implemented any time soon. Just wanted to document the thought.

-----

Update 4/12/24:

Created a new focus style that combines the focus style links with a negative outline offset. This allows the banner to have a completely visible focus state on mobile.


pwolfert:
I like the inset idea, and I think it's worth exploring. Even when we now have the bottom outline visible, it's not much to go off of, and it's not actually entirely clear what element it belongs to (which element has focus).

Screenshot 2024-04-09 at 9 59 12 AM

zarahzachz:
An update on this - we met with the team to discuss the focus not encircling the banner and decided to create a local focus style to fix the problem

packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -142,9 +146,8 @@
   [aria-expanded='true'] & {
     background-color: var(--usa-banner-close-button__background-color--mobile);
     display: flex;
-    inset: 0;
+    inset: 0.25em 0.25em 0.25em 0;
```
zarahzachz, line 149: Removing min-height and changing `inset` to accommodate the offset value allows the close button, which only appears on mobile when open, reveals the focus ring around the entire component.

pwolfert: A possibly unintended side-effect of this insetting is that the unfocused button looks different now. The 'X' button loses its square shape and gets moved.

Before

After

What if we applied the outline to an absolutely positioned pseudoelement that could render over the top of that 'X' span?

pwolfert: @zarahzachz, it looks to me like [the border strategy](https://github.com/CMSgov/design-system/pull/3030/commits/9883d56ef15a0c804525519ad4fea44676e5a179) has the same side-effect of making the X button rectangular and inset

pwolfert: Never mind! I forgot to pull!

pwolfert: Screenshot 2024-04-15 at 9 22 31 AM

there is still this subtle shift of the X to the left inside the button, but this is a more elegant solution than the one I had proposed

PR #3033, opened 2024-04-10
[WNMGDS-2745] Fix dialogs inheriting font size from ancestors

pwolfert:
## Summary

Protect a dialog's width and font size from inheriting from ancestors. Fix drawers too, which had the same problem.

## How to test

1. Open up a story with a dialog
2. Manually set the font size of the story div (dialog's ancestor) to have a really large or really small font size
3. Without this fix, the dialog will grow really wide or really narrow and the internal font sizes will be affected. With this fix, those two things should remain unaffected.

## Checklist

### DIFF
--- a/packages/design-system/src/styles/components/_Dialog.scss
+++ b/packages/design-system/src/styles/components/_Dialog.scss
@@ -54,6 +54,10 @@ dialog.fixed {
   box-sizing: border-box;
   color: var(--color-base); // Needed to override user-agent styles `canvasText`
   display: block;
+  font-size: var(
+    --font-size-base
+  ); // Reset so font size of ancestor elements don't affect the dialog
+
   margin: auto;
   margin-block-start: $spacer-6;
   max-width: var(--measure-base);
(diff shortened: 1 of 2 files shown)


packages/design-system/src/styles/components/_Dialog.scss
```diff
@@ -54,6 +54,9 @@ dialog.fixed {
   box-sizing: border-box;
   color: var(--color-base); // Needed to override user-agent styles `canvasText`
   display: block;
+
+  // Reset so font size of ancestor elements don't affect the dialog
```
pwolfert, line 58: Looks like our linter is automatically putting an extra line break above this comment. When I had the comment on the same line as the code, it made it even weirder.

zarahzachz (approved):
This looks good to me. 

I'm trying to figure something out though... if you add a crazy large font-size to a parent element, the dialog button shifts downward. I'm poking around on dev tools and can't figure out what would cause that layout shift. It's just something odd I'm noticing that is unrelated to your work.

pwolfert:
Oh, I see what you mean, it's with the trigger button behind, not the dialog

PR #3047, opened 2024-04-17
[WNMGDS-2723] Choice web component

zarahzachz:
WNMGDS-2723

Implement Choice as a web component.

`checkedChildren` and `uncheckedChildren` are being used as slots instead of proper attributes and that's OK!

### DIFF
--- a/packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx
+++ b/packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx
@@ -41,10 +41,10 @@ export default {
         "A unique ID to be used for the input field, as well as the label's `for` attribute. A unique ID will be generated if one isn't provided.",
       control: 'text',
     },
-    // inversed: {
-    //   description: 'Applies the "inverse" UI theme',
-    //   control: 'boolean',
-    // },
+    inversed: {
+      description: 'Applies the "inverse" UI theme',
+      control: 'boolean',
+    },
     size: {
       description: 'Sets the size of the input to `"small"`',
       options: [undefined, 'small'],
@@ -137,7 +137,16 @@ const Template = (args) => {
     };
   });
 
-  return <ds-choice {...args} />;
+  return (
+    <ds-choice {...args}>
+      <div slot="checked-children">
+        <ds-alert variation="error">checked children</ds-alert>
+      </div>
+      <div slot="unchecked-children">
+        <ds-alert>unchecked children</ds-alert>
+      </div>
+    </ds-choice>
+  );
 };
 
 export const Default = Template.bind({});
@@ -146,7 +155,5 @@ export const CheckedChildren = {
   render: Template,
   args: {
     'default-checked': true,
-    'checked-children': <p>checked</p>,
-    'unchecked-children': <p>unchecked</p>,
   },
 };
--- a/packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx
+++ b/packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx
@@ -1,0 +1,152 @@
+import { useEffect } from 'react';
+import WebComponentDocTemplate from '../../../../../../.storybook/docs/WebComponentDocTemplate.mdx';
+import { action } from '@storybook/addon-actions';
+import './ds-choice';
+import '../ds-alert';
+import { webComponentDecorator } from '../storybook';
+
+export default {
+  title: 'Web Components/Choice',
+  argTypes: {
+    'default-checked': {
+      description:
+        'Sets the initial `checked` state. Use this for an uncontrolled component otherwise, use the `checked` property.',
+      control: 'boolean',
+    },
+    checked: {
+      description:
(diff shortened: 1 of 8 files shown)


packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx
```diff
@@ -0,0 +1,154 @@
...
+    return () => {
+      choice.removeEventListener('ds-change', onChange);
+      choice.removeEventListener('ds-blur', onBlur);
+    };
+  });
+
+  return (
+    <ds-choice {...args}>
+      <div slot="checked-children">
+        <div className="ds-c-alert">{args['checked-children']}</div>
```
zarahzachz, line 145: for some reason hooking the story up to the args caused the html to be stripped away. if the value is hard-coded it's fine, so this seems like a storybook problem.

packages/design-system/src/components/web-components/ds-choice/ds-choice.test.tsx
```diff
@@ -0,0 +1,274 @@
...
+    expect(el).toBeChecked();
+  });
+
+  it('is defaultChecked', () => {
+    renderChoice({ 'default-checked': true });
+    const el = screen.getByRole('checkbox');
+    expect(el).toBeChecked();
+  });
+
+  it.skip('is required', () => {
```
zarahzachz, line 59: i'm noticing some strangeness when it comes to spreading props... they aren't being correctly applied to the underlying elements like we see on react components. i'm skipping these types of tests for now.

packages/design-system/src/components/web-components/ds-dropdown/ds-dropdown.test.tsx
```diff
@@ -15,7 +15,7 @@ function renderDropdown(attrs = {}) {
   return render(<ds-dropdown {...defaultAttrs} {...attrs} />);
 }
 
-describe('Button', () => {
+describe('Dropdown', () => {
```
pwolfert, line 18: Need to update the snapshots after this change

packages/design-system/src/components/web-components/ds-choice/ds-choice.test.tsx
```diff
@@ -0,0 +1,274 @@
...
+    it('calls the onBlur handler', () => {
+      renderChoice(props);
+      const el = screen.getByLabelText('George Washington');
+      el.focus();
+      userEvent.tab();
+
+      expect(props.onBlur).toHaveBeenCalledTimes(1);
+      expect(props.onChange).toHaveBeenCalledTimes(0);
+    });
+  });
```
pwolfert, line 187: This needs to be updated to use `addEventListener` and the custom events. I think these are only being passed down successfully by a trick of Preact in the unit tests.

packages/design-system/src/components/web-components/ds-choice/ds-choice.test.tsx
```diff
@@ -0,0 +1,274 @@
...
+          <div slot="checked-children">
+            <strong data-testid="checked" className="checked-child">
+              I am checked
+            </strong>
+          </div>
+          <div slot="unchecked-children">
+            <strong data-testid="unchecked" className="unchecked-child">
+              I am unchecked
+            </strong>
+          </div>
```
pwolfert, line 202: Since we're adding both, maybe we should do a check for internal state management in this unit test where we click on the checkbox and see if the other one becomes visible and the first one gets hidden.

packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx
pwolfert: Update: @zarahzachz wrote a follow-up ticket for adding partial attribute pass-through support.

pwolfert (commented):
Made changes for all my feedback items except the unit tests

packages/design-system/src/components/web-components/ds-choice/ds-choice.test.tsx
pwolfert: Oh, I now see that's what you were doing in the next test

packages/design-system/src/components/web-components/ds-choice/ds-choice.test.tsx
```diff
@@ -0,0 +1,266 @@
...
+    renderChoice();
+
+    const choiceRoot = document.querySelector('ds-choice');
+    const mockHandler = jest.fn();
+    choiceRoot.addEventListener('ds-change', mockHandler);
+
+    const input = screen.getByRole('checkbox');
+    userEvent.click(input);
+
+    expect(mockHandler).toHaveBeenCalledTimes(1);
```
pwolfert, line 169: I wanted to assert here that `mockHandler.mock.lastCall[0].details.target.checked` was true, but the event object is kind of messed up. I'm wondering if it's a limitation with jsdom and custom events. It seems to work right in Storybook.

pwolfert (approved):
Applied the unit test feedback too. LGTM

PR #3050, opened 2024-04-19
[WNMGDS-2742] Create spinner web component.

zarahzachz:
WNMGDS-2742

(Accidentally named the branch after the wrong ticket - the skip link wc)


packages/design-system/src/components/Spinner/index.ts
```diff
@@ -1 +1 @@
-export { default as Spinner } from './Spinner';
+export { default as Spinner, SpinnerProps } from './Spinner';
```
zarahzachz, line 1: We weren't exporting `SpinnerProps` for some reason and I added it here so I could use it in the 🚾

packages/design-system/src/components/web-components/ds-spinner/ds-spinner.stories.tsx
```diff
@@ -0,0 +1,49 @@
...
+        'Additional classes to be added to the spinner element. Useful for adding utility classes.',
+    },
+    inversed: {
+      control: 'boolean',
+      description: 'Set to "true" to apply the "inverse" theme.',
+    },
+    filled: {
+      control: 'boolean',
+      description: 'Adds a background behind the spinner for extra contrast.',
+    },
```
pwolfert, line 24: This has nothing to do with your code, but I just noticed that `inversed` doesn't do anything unless `filled` is also true. Maybe in the absence of better documentation or a better API we should include the same kind of "Inversed filled" example that the React spinner has.

pwolfert (approved):
Looks great! One Storybook suggestion, and it looks like you might need the new VRT refs for this story. Don't forget that we can now do `--grep "Spinner"`!

PR #3051, opened 2024-04-22
[WNMGDS-2712] Update deprecated aria

zarahzachz:
WNMGDS-2712

Made assorted ARIA tweaks. ARIA has deprecated some of its attrs on certain roles, so we needed to update those roles to better work with ARIA. More context is available in #2978.

[Demo url](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria)


TiffanyPender:
For [WNMGDS-2769 A11y QA](https://jira.cms.gov/browse/WNMGDS-2769), I've added my findings for the components updated as part of[ WNMGDS-2712](https://jira.cms.gov/browse/WNMGDS-2712).

- [Autocomplete](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/iframe.html?globals=theme:core&id=components-autocomplete--default&viewMode=story)
  - `aria-selected="false"` is still present on li elements.
  - JAWS doesn't announce options as you move through the option list. It announces "collapsed" after making a selection, but not what you selected.
  - NVDA announces the letter you typed as you move through the option list. It announces what you selected plus "collapsed" after making a selection.

- ChoiceList
  - [Checkbox](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-choicelist--default-checkbox&globals=theme:core)
    - What changed? When I enable the error state by providing an error message, aria-invalid is applied to the fieldset element. 
      - From 2712: "Set `role="listbox"` on `fieldset`, works with `aria-invalid`" —— This is not valid - [ARIA role listbox is not allowed for given element (Deque)](https://dequeuniversity.com/rules/axe/4.8/aria-allowed-role?application=AxeChrome) (fieldset element). This role could possibly be used on a div, but I'm not sure that we'd want to deviate from using the fieldset element.
    - `aria-invalid` is not allowed on the fieldset element (unless it has the radiogroup role).
      - VO, JAWS doesn't announce "invalid" with the attribute. The only way to know is with the error message text present.
      - NVDA announces "grouping invalid entry, check some boxes".

  - [Radio](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-choicelist--with-error&globals=theme:core)
  - `role="radiogroup"` is applied, along with `aria-invalid="true"`
  - VO, JAWS doesn't announce "invalid" with the attribute. The only way to know is with the error message text present.
  - NVDA announces "grouping invalid entry, choose an option".
  - Additional resources:
    - [Support for marking radio buttons required, invalid](https://adrianroselli.com/2022/02/support-for-marking-radio-buttons-required-invalid.html) (Adrian Roselli) - Numbers 3 and 6
    - [Add required-attribute support for fieldsets](https://github.com/whatwg/html/issues/6868) (WHATWG GitHub)
    - [Consider allowing aria-required on fieldset element](https://github.com/w3c/aria/issues/1322) (ARIA GitHub)

- [Dialog](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-dialog--dialog-example&globals=theme:core)
  - What changed? Focus moves to the dialog as expected. 
  - `H2` is present as default.

- [Drawer](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-drawer--drawer-default&globals=theme:core)
  - `aria-labelledby` is on the dialog element and referencing the drawer heading text. The default of `H3` may cause an issue with heading hierarchy.
  - The drawer body is in the tab order, with a visible focus. Users should not be able to tab to the drawer body. Tabs should only go to interactive elements.

- [Dropdown](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-dropdown--with-error&globals=theme:core)
  - What changed? I don't see `role="combobox"` on the button element to test `aria-invalid="true"`.
  - [Button role should allow aria-required](https://github.com/w3c/aria/issues/1253) (ARIA GitHub)

- [MonthPicker](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-monthpicker--with-error&globals=theme:core)
  - What changed? I see `aria-invalid="true"` on the `fieldset`.  ARIA role `listbox` is not allowed for given element (Deque) (fieldset element). This role could possibly be used on a div, but I'm not sure that we'd want to deviate from using the `fieldset` element.
  - VO, JAWS doesn't announce "invalid" with the attribute. The only way to know is with the error message text present.
  - NVDA announces "grouping invalid entry, select available months".
  - As an aside, NVDA reads out the abbreviations, so it would be good to add in screen reader text with the full month name, hiding the abbreviations from assistive technologies.

zarahzachz:
> [Autocomplete](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/iframe.html?globals=theme:core&id=components-autocomplete--default&viewMode=story)
> 
> * `aria-selected="false"` is still present on li elements.
> * JAWS doesn't announce options as you move through the option list. It announces "collapsed" after making a selection, but not what you selected.
> * NVDA announces the letter you typed as you move through the option list. It announces what you selected plus "collapsed" after making a selection.

This appears to be additional markup aria-react adds to the options in the component. Unfortunately, I don't have control over how `aria-selected` appears on those elements.

We have a component called `renderStatusMessage` that's used to show ["no results"](https://design.cms.gov/storybook/?path=/story/components-autocomplete--no-results&globals=theme:core) or ["loading..."](https://design.cms.gov/storybook/?path=/story/components-autocomplete--loading-message) messages in the Autocomplete and for these examples, `aria-selected` is not present.

Because the code I changed impacts only those two examples, it seems the NVDA and JAWS issues are unrelated to my changes. I'll make a follow-up ticket to address them (and investigate if it's possible to modify the aria-react library to remove the redundant attr).

zarahzachz:
> [Checkbox](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-choicelist--default-checkbox&globals=theme:core)
> 
> * What changed? When I enable the error state by providing an error message, aria-invalid is applied to the fieldset element.
>   
>   * From 2712: "Set `role="listbox"` on `fieldset`, works with `aria-invalid`" —— This is not valid - [ARIA role listbox is not allowed for given element (Deque)](https://dequeuniversity.com/rules/axe/4.8/aria-allowed-role?application=AxeChrome) (fieldset element). This role could possibly be used on a div, but I'm not sure that we'd want to deviate from using the fieldset element.
> * `aria-invalid` is not allowed on the fieldset element (unless it has the radiogroup role).
>   
>   * VO, JAWS doesn't announce "invalid" with the attribute. The only way to know is with the error message text present.
>   * NVDA announces "grouping invalid entry, check some boxes".

No changes were made to the checkbox implementation of Choice/ChoiceList because a suitable `role` that worked with `aria-invalid` couldn't be found. Currently, checkboxes are still wrapped in a `<fieldset>` tag, which has an implicit role of group and `aria-invalid` is not allowed on that role. 

Setting `aria-invalid` for checkboxes still requires guidance.

zarahzachz:
> * [Radio](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-choicelist--with-error&globals=theme:core)
> * `role="radiogroup"` is applied, along with `aria-invalid="true"`
> * VO, JAWS doesn't announce "invalid" with the attribute. The only way to know is with the error message text present.
> * NVDA announces "grouping invalid entry, choose an option".

For radio Choice/ChoiceList, I've added `role={props.type === 'radio' ? 'radiogroup' : null}` to the `<fieldset>` element. This will apply the `radiogroup` role to `<fieldset>` only if the Choice type is "radio," and this seems to comply with the guidance present in the list of references. I wish checkboxes had a similar solution, but c'est la vie.

It's worth noting, the logic I added will result in checkbox fieldsets having the implicit "role" group, which is no change to how it functions today. So the only thing changing for Choice/ChoiceList is a slight improvement for radio types.

zarahzachz:
> [Dialog](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-dialog--dialog-example&globals=theme:core)
> 
> * What changed? Focus moves to the dialog as expected.
> * `H2` is present as default.

For Drawer, I moved the `aria-labelledby` off an internal `<div>` and placed it on the `<dialog>` element itself, since `aria-labelledby` can't be placed on an element with a "generic" role. 

I also updated some of the semantic markup used within:
- Replaced `<header>` and `<main>` with `<div>` and removed `role="document"` to match the markup we're using on Dialog (Drawer is just a different layout/style of Dialog, so they should have the same markup).
- Updated the heading tag to not be an `<h1>` by default.

zarahzachz:
> [Dropdown](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-dropdown--with-error&globals=theme:core)
> 
> * What changed? I don't see `role="combobox"` on the button element to test `aria-invalid="true"`.
> * [Button role should allow aria-required](https://github.com/w3c/aria/issues/1253) (ARIA GitHub)

No changes were made to Dropdown (aside from adding a code comment). I tested re-implementing `role: "combobox",` but ran into the same VO issues mentioned in the code comments (VO announcing `<button>` as an input with misspelling errors). 

This is another side effect of using the react-aria library, and library authors appear to be aware of it.

zarahzachz:
> [MonthPicker](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-monthpicker--with-error&globals=theme:core)
> 
> * What changed? I see `aria-invalid="true"` on the `fieldset`.  ARIA role `listbox` is not allowed for given element (Deque) (fieldset element). This role could possibly be used on a div, but I'm not sure that we'd want to deviate from using the `fieldset` element.
> * VO, JAWS doesn't announce "invalid" with the attribute. The only way to know is with the error message text present.
> * NVDA announces "grouping invalid entry, select available months".
> * As an aside, NVDA reads out the abbreviations, so it would be good to add in screen reader text with the full month name, hiding the abbreviations from assistive technologies.

Because MonthPicker is just a more extreme version of checkbox ChoiceList, I didn't implement any changes (due to the same `<fieldset>` issues mentioned in the checkbox ChoiceList comment).

Making [a follow-up ticket](https://jira.cms.gov/browse/WNMGDS-2794) to address the abbreviated month names for NVDA. Will need a11y testing when this one's complete since we don't have NVDA screen readers.

packages/design-system/src/components/Dialog/Dialog.tsx
```diff
@@ -123,31 +123,26 @@ export const Dialog = (props: DialogProps) => {
...
-      <div
-        className="ds-c-dialog__window"
-        role="document"
-        ref={containerRef}
-        tabIndex={-1}
-        aria-labelledby={headingId}
-      >
-        <header className={headerClassNames}>
+      <div className="ds-c-dialog__window" ref={containerRef} tabIndex={-1}>
+        <div className={headerClassNames}>
```
pwolfert, line 129: I just know the original design was tested by accessibility specialists with a variety of screen readers, so will this be tested again by the accessibility team before merging?

packages/design-system/src/styles/components/_Drawer.scss
```diff
@@ -72,6 +72,9 @@
   overflow: auto;
   overscroll-behavior: contain;
 
+  // TODO: Investigate why this is needed.
+  // Block DOM els already have a column layout?
+  // Our supported browsers support flex, so this query not needed.
```
pwolfert, line 77: Huh, it does seem sus. I traced it back to [this commit](https://github.com/CMSgov/design-system/commit/e19b8ec66508ed19fe82dc8ddcad7cc9325cce2f), and according to that message, these `@supports` queries were added to support IE11. Let's remove them!

pwolfert (approved):

packages/design-system/src/components/Dialog/Dialog.tsx
zarahzachz: @TiffanyPender would it be possible to get another review of the [Dialog changes](https://cmsgov.github.io/design-system/branch/WNMGDS-2712/update-deprecated-aria/storybook/?path=/story/components-dialog--dialog-example&globals=theme:core) in this PR? 

The changes made to this component are:
- Moving `aria-labelledby` to the `<dialog>` element
- Replacing the `<header>` and `<main>` tags with `<div>`
    - `<header>` and `<main>` didn't have accessible names
    - Including `<main>` seemed to fly against the rule of only 1 `<main>` element / page
- Removing `role="document"` from an internal `<div>`

These changes would align Dialog with Drawer markup (Drawer does not use these elements) and I have a sneaking suspicion the original Dialog markup was written well before the release of `<dialog>`, so attributes like `role="document"` are unnecessary.

zarahzachz: @pwolfert I got feedback [from Tiffany with feedback for Dialog](https://jira.cms.gov/browse/WNMGDS-2796).

Changes introduced:
1. Moved focus to `<dialog>` element (where the `aria-labelledby` can label the contents of the dialog)
2. Added the option to hide aria content for CloseButton

pwolfert (approved):
Dialog sounds good in VoiceOver 🎉

PR #3055, opened 2024-04-24
[WNMGDS-2714] Fix `Dialog` and `HelpDrawer` analytics firing too early

pwolfert:
## Summary

Fixes `Dialog` and `HelpDrawer` analytics open event firing before it's actually opened. When we changed the `NativeDialog` API to use an `isOpen` prop to control its open state, it meant that rendering a `NativeDialog` was no longer equivalent to opening a dialog. Our analytics events for `Dialog` and `HelpDrawer` _open_ and _close_ were based on mounting and unmounting the components, so when one of those comonents rendered, it would fire the _open_ event even if `isOpen` was `false`.

## How to test

1. Take a look a the unit tests
2. Take a look at the analytics logging in Storybook

## Checklist


zarahzachz (commented):
seeing weird behavior for `onEnter` with Dialog (discussed in office hours)

zarahzachz (approved):

PR #3056, opened 2024-04-25
[WNMGDS-2743] Skip nav web component

zarahzachz:
WNMGDS-2743

Convert SkipNav to web component.


packages/design-system/src/components/web-components/ds-skip-nav/ds-skip-nav.tsx
```diff
@@ -0,0 +1,28 @@
...
+        HTMLElement
+      >;
+    }
+  }
+}
+/* eslint-enable */
+
+const Wrapper = ({ children, ...otherProps }) => (
+  <SkipNav href={otherProps.href}>
+    {children[0] === undefined ? t('skipNav.default') : children}
```
zarahzachz, line 24: OK... I originally had this logic on the React file. Moved it to the 🚾 and it does seem to work 🎉 

I had `console.log(Boolean(children))` and `console.log(children)` on the React component to see if 1) `children` ever returned `false` (it didn't) and 2) what the value of `children` was. Turns out `children` returned `true` as an array with a length of 0 if no overriding text was sent to the component. This meant the default i18n would never appear as the 🚾 always had `children` (even if empty).

If I checked to see if the 0th child of children was undefined, it would show the default i18n text. 

This seems to work? Thoughts?

pwolfert: I actually think we can fix this in `define.ts`, and then it'll work for all future web components and won't require the glue code. The fact that we pass an empty array when it's empty HTML is more of an artifact of our function that converts HTML to Preact VNodes. Replacing the `unwrapTemplateVNode` function with this seems to work:

```ts
function unwrapTemplateVNode(vnode: VNode): VNode {
  const children = vnode.props.children[0].props.children;
  if (Array.isArray(children) && children.length === 0) {
    return undefined;
  } else {
    return children;
  }
}
```

pwolfert: Thanks for figuring out that it'd always return an empty array when there's no inner HTML

pwolfert (approved):
LGTM 🎉 

but we can ignore this story in the `storySkipList` in `tests/browser/stories.test.ts` since it doesn't have any visuals.

PR #3068, opened 2024-05-01
[WNMGDS-2741] Month picker web component

zarahzachz:
WNMGDS-2741

- Refactor React MonthPicker to allow children `option` to be passed
- Implement web component MonthPicker


packages/design-system/src/components/web-components/ds-month-picker/ds-month-picker.tsx
```diff
@@ -0,0 +1,58 @@
...
+}
+/* eslint-enable */
+
+interface WrapperProps extends Omit<MonthPickerProps, 'inversed'> {
+  inversed?: string;
+  rootId?: string;
+}
+
+const Wrapper = ({ children, rootId, ...otherProps }: WrapperProps) => (
+  <MonthPicker {...otherProps} id={rootId} inversed={parseBooleanAttr(otherProps.inversed)}>
```
pwolfert, line 50: I think we should give the three array attributes the same treatment that we do the `options` array attribute in `ds-dropdown` where teams are allowed to directly assign arrays to the properties (`monthPicker['disabled-months'] = [1]`) or assign a JSON string in the actual HTML attribute. There's an angular team who requested that functionality on the `ds-dropdown`, so I think they'd probably want a similar thing here.

```suggestion
  <MonthPicker
    {...otherProps}
    id={rootId}
    inversed={parseBooleanAttr(otherProps.inversed)}
    disabledMonths={typeof disabledMonths === 'string' ? JSON.parse(disabledMonths) : disabledMonths}
 >
```

Could turn this into one of the `wrapperUtils`

packages/design-system/src/components/web-components/ds-month-picker/ds-month-picker.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+  },
+  parameters: {
+    docs: {
+      page: WebComponentDocTemplate,
+      description: {
+        component:
+          'For information about how and when to use this component, [refer to its full documentation page](https://design.cms.gov/components/month-picker/).',
+      },
+      componentEvents: {
+        'ds-change': {
```
pwolfert, line 106: We need entries for the select all and clear all events

packages/design-system/src/components/web-components/ds-month-picker/ds-month-picker.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+    'requirement-label': { control: 'text' },
+    'root-id': {
+      description:
+        "A unique ID for this element. A unique ID will be generated if one isn't provided.",
+      control: 'text',
+    },
+    'select-all-text': {
+      description: 'The text for the "Select all" button for internationalization.',
+      control: 'text',
+    },
```
pwolfert, line 89: We should include docs for the three array attributes as well with special notes that the alternative is to use `<input>` elements.

packages/design-system/src/components/web-components/ds-month-picker/ds-month-picker.stories.tsx
```diff
@@ -0,0 +1,147 @@
...
+      monthPicker.removeEventListener('ds-change', onChange);
+      monthPicker.removeEventListener('ds-clear-all', onClearAll);
+      monthPicker.removeEventListener('ds-select-all', onSelectAll);
+    };
+  });
+
+  return <ds-month-picker {...args} />;
+};
+
+export const Default = Template.bind({});
```
pwolfert, line 147: The default story on at the bottom of the docs page actually fails to render for me unless I make the following change to the decorator component in `web-components/storybook.tsx` to strip out the Preact element children:

```tsx
export function webComponentDecorator(Story, context) {
  // The Preact element `children` can have circular references which trip up Storybook's JSON evaluation
  const {children, ...simpleArgs} = context.args;
  return (
    <Fragment key={JSON.stringify({ ...simpleArgs, ...context.globals })}>
      <Story />
    </Fragment>
  );
}
```

packages/design-system/src/components/web-components/storybook.tsx
```diff
@@ -1,9 +1,12 @@
...
 
 export function webComponentDecorator(Story, context) {
+  // The Preact element `children` can have circular references which trip up Storybook's JSON evaluation
+  const { children, ...simpleArgs } = context.args;
   return (
-    <Fragment key={JSON.stringify({ ...context.args, ...context.globals })}>
-      <Story />
+    <Fragment key={JSON.stringify({ ...simpleArgs, ...context.globals })}>
+      {' '}
+      <Story />{' '}
```
pwolfert, line 9: There are extra spaces here for some reason

packages/design-system/src/components/MonthPicker/MonthPicker.stories.tsx
```diff
@@ -9,27 +13,79 @@ const meta: Meta<typeof MonthPicker> = {
...
+        onChange={onChange}
+        onSelectAll={onSelectAll}
+        onClearAll={onClearAll}
+      />
+    );
+  },
+};
+
 export const Default: Story = {
+  // ...UncontrolledTemplate,
```
pwolfert, line 77: What happened to the uncontrolled template?

packages/design-system/src/components/web-components/ds-month-picker/ds-month-picker.stories.tsx
```diff
@@ -0,0 +1,159 @@
...
+
+export const MonthOverrides = {
```
pwolfert, line 152: Good name 👍

packages/design-system/src/components/MonthPicker/MonthPicker.stories.tsx
zarahzachz: oh... moved it to the render in the meta data like we did for Pagination stories. Forgot to remove this comment.

packages/design-system/src/components/web-components/storybook.tsx
zarahzachz: oh that must've carried over from copying from GH. removing!

packages/design-system/src/components/MonthPicker/MonthPicker.stories.tsx
pwolfert: oh okay, I think the commit diff I was looking at was missing the fact that it moved to the meta section. I thought it was lost.

packages/design-system/src/components/web-components/ds-month-picker/ds-month-picker.stories.tsx
```diff
@@ -0,0 +1,173 @@
...
+        "A unique ID for this element. A unique ID will be generated if one isn't provided.",
+      control: 'text',
+    },
+    'select-all-text': {
+      description: 'The text for the "Select all" button for internationalization.',
+      control: 'text',
+    },
+    'selected-months': {
+      description:
+        'Array of month numbers, where `1` is January, and any month included is selected. This will render a read-only field. If the field should be mutable, use `defaultSelectedMonths`. Alternatively, you can apply `checked` to `<input>` element children.',
```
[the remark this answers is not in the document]
pwolfert: Since it would be weird behavior anyway to not allow a user to change something that isn't disabled, I think we should leave it, and maybe we should actually remove the `selected-months` attribute from the web component entirely.

pwolfert (approved):
Turns out that the "read-only" language in the docs doesn't apply anymore—either to this or the base React component—but everything else looks good to go!

packages/design-system/src/components/web-components/ds-month-picker/ds-month-picker.stories.tsx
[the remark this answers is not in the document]
zarahzachz: removing `selected-months`!

PR #3070, opened 2024-05-02
[WNMGDS-2744] Add bundle generation for individual web components

pwolfert:
## Summary

WNMGDS-2744

See the updated documentation and the comments for details, but there are now two ways of consuming the web components from the CDN: All of them at once or a la carte.

- Changed the webpack config to support two different styles of bundles for web components
- Created alternate imports in the astro example to showcase the other way
- Updated the developer docs on the doc site
- Updated the CDN index documentation
- The **BREAKING** change is that the normal web components bundle is now called `all.js` instead of `web-components.js`. I'm open to feedback on this. I wanted to differentiate from `base.js`, but it'd be nice to not cause a breaking change too...

## How to test

1. Try the example
2. If you want to see them imported from a child design system, check out 154417334

## Checklist


zarahzachz (approved):
OK so testing in different browsers has different outcomes. In Firefox, Spinner doesn't appear, but these changes seem to have fixed whatever my issue with Chrome was and all components (including Spinner) work/appear fine. Safari also works fine. I'm going to approve because this is probably a "Sarah's weird laptop" issue.

PR #3071, opened 2024-05-03
[WNMGDS-2771] Add `ds-accordion` and `ds-accordion-item` web components

pwolfert:
## Summary

- Adds `ds-accordion` and `ds-accordion-item` web components, along with stories, unit tests, and VRT images
    - I left support for the `closeIconComponent` and `openIconComponent` out of the `ds-accordion-item` element since those are really React-specific and we want to move away from that kind of customization anyway.
    - Didn't include `isControlledOpen` either because controlled components are a bit of a foreign concept for web components anyway.
- Changes the logic of the `AccordionItem` React component to only follow _controlled_ behavior if the `isControlledOpen` prop is passed (a pre-existing prop that was used to set its open state if controlled). The previous logic for determining whether it is controlled was incompatible with web components, because we always supply a callback for all events.

## How to test

Try out the stories and unit tests

## Checklist


packages/design-system/src/components/web-components/ds-accordion/ds-accordion-item.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+        component: `For information about how and when to use this component, [refer to its full documentation page](https://design.cms.gov/components/accordion/).`,
+      },
+    },
+  },
+  decorators: [webComponentDecorator],
+};
+
+const Template = (args) => (
+  <ds-accordion-item {...args}>
+    This is some detailed information inside an accordion item.
```
zarahzachz, line 52: Something's up with this Story - not seeing a border on my local Storybook
image

zarahzachz: Does AccordionItem need a Story? React doesn't have one.

zarahzachz: (Also, Accordion does have the correct borders, so my screenshot probably means nothing then)

packages/design-system/src/components/web-components/ds-accordion/ds-accordion-item.stories.tsx
```diff
@@ -0,0 +1,56 @@
...
+  },
+  parameters: {
+    docs: {
+      page: WebComponentDocTemplate,
+      description: {
+        component: `For information about how and when to use this component, [refer to its full documentation page](https://design.cms.gov/components/accordion/).`,
+      },
+    },
+  },
+  decorators: [webComponentDecorator],
```
zarahzachz, line 47: I don't see `onChange` being documented in the Story's meta.

packages/design-system/src/components/web-components/ds-accordion/ds-accordion-item.stories.tsx
pwolfert: [See commit note](https://github.com/CMSgov/design-system/pull/3071/commits/e336ff749dec47434de8740157f17ca708df50c3) - I only included it because otherwise we wouldn't have a docs page for that one

packages/design-system/src/components/web-components/ds-accordion/ds-accordion-item.stories.tsx
pwolfert: Good catch. I forgot there was an onChange

zarahzachz (approved):

PR #3073, opened 2024-05-06
[NO-TICKET] Add live-analytics option to Storybook

pwolfert:
## Summary

In order to test analytics with the analytics team before new analytics features get released, we were thinking we'd enable analytics for specific components on the doc site, which already sends analytics. However, this would require that we make specific test pages for features we're trying to test. If we add it to Storybook, we have all our components already available to test. Therefore I'm adding a new config option to be able to choose live analytics for validating the actual events sent to our analytics provider. This can be run either locally or on a deployed copy.

## How to test

Select the new analytics config option and try out one of the analytics-enabled components.

Screenshot of new 'On (Live)' analytics option in the analytics config menu of our Storybook

Also try turning it off and on again and switching between the different options.

## Checklist

### DIFF
--- a/.storybook/preview.tsx
+++ b/.storybook/preview.tsx
@@ -8,7 +8,14 @@ import type { UtagContainer } from '@cmsgov/design-system';
 import type { Preview } from '@storybook/react';
 
 // Rewire analytics events to log to the console
-(window as UtagContainer).utag = { link: console.log };
+let originalUtag;
+function mockUtag() {
+  originalUtag = (window as UtagContainer).utag;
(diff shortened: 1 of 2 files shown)


zarahzachz (approved):
Works well locally

pwolfert:
> Does `button_style` need to be updated to include info on whether `isAlternate` is set? Noticing it only captures if a certain `variation` exists (sold or ghost) but doesn't capture if it's using the primary or alternate color. Is that an issue? Trying to figure out what purpose this data serves analytics.

That's a really good question. Let's ask the analytics team.

PR #3074, opened 2024-05-06
[WNMGDS-2737] Add ability to spread props on drawer

zarahzachz:
WNMGDS-2737

### DIFF
--- a/packages/design-system/src/components/Drawer/Drawer.tsx
+++ b/packages/design-system/src/components/Drawer/Drawer.tsx
@@ -69,55 +69,77 @@ export interface DrawerProps {
  * [refer to its full documentation page](https://design.cms.gov/components/drawer/).
  */
 export const Drawer = (props: DrawerProps) => {
-  const headingRef = useRef(null);
-  const headingId = useId('drawer--', props.headingId);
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/Drawer/Drawer.tsx
```diff
@@ -69,55 +69,77 @@ export interface DrawerProps {
...
+    backdropClickExits,
+    children,
+    className,
+    closeButtonText,
+    closeButtonVariation,
+    footerBody,
+    footerTitle,
+    hasFocusTrap,
+    heading,
+    headingId,
```
pwolfert, line 83: One pattern we have in a couple components right now to solve competing prop names and variable names is redefining the prop with a prefix of `user*`, like `headingId: userHeadingId`. That might prevent us from accidentally using `headingId` later on when we mean `headingElId`. See Dropdown's `userOnBlur` and Autocomplete's `userInputRef`.

pwolfert (approved):
It's working! One suggestion below, and also can you add unit tests that check for the ability to pass arbitrary attributes from HelpDrawer all the way down to the dialog?

PR #3079, opened 2024-05-09
[NO-TICKET] Add documentation for web component slots

pwolfert:
## Summary

- Adds a new `WebComponentSlotsTable` to Storybook
- Uses it in the `ds-choice` docs to explain slots generally and list the two slots

## How to test

Look at the storybook docs for `ds-choice`.

## Checklist

### DIFF
--- a/.storybook/docs/WebComponentSlotsTable.tsx
+++ b/.storybook/docs/WebComponentSlotsTable.tsx
@@ -1,0 +1,62 @@
+import { useOf } from '@storybook/blocks';
+
+/**
+ * A table documenting a web component's available slots
+ */
+export const WebComponentSlotsTable = ({ of }) => {
+  const resolvedOf = useOf(of || 'story', ['story', 'meta']);
+  if (resolvedOf.type !== 'story') {
+    return null;
+  }
+  const slots = resolvedOf.story.parameters?.docs?.slots;
+  if (!slots) {
+    return null;
+  }
+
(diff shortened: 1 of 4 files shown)


.storybook/docs/WebComponentSlotsTable.tsx
```diff
@@ -0,0 +1,62 @@
...
+        hidden depending on the checked state. To pass HTML as a slot inside your web component, use
+        either{' '}
+        <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/slot">
+          the slot element
+        </a>{' '}
+        with a <code>name</code> attribute that matches one of the slots in the table below or{' '}
+        <a href="https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/slot">
+          the slot attribute
+        </a>{' '}
+        on an element type of your choosing.
```
zarahzachz, line 44: Should we specify this needs to be set on a web component itself? Maybe a better way to say this is "on a web component host of your choosing"

pwolfert: I was actually referring to something like `<div slot="checked-children">` like we're doing in the example

zarahzachz: ooooohhhhh OK

zarahzachz (approved):

PR #3081, opened 2024-05-11
[WNMGDS-2724] ds-choice-list web component

zarahzachz:
WNMGDS-2724

- Create ChoiceList web component, stories and tests
- Add capability for web component to accept `children` as HTML elements (only `<ds-choice>`)


packages/design-system/src/components/web-components/ds-choice-list/ds-choice-list.stories.tsx
```diff
@@ -0,0 +1,188 @@
...
+    <ds-choice type="checkbox" label="Bar" value="bar">
+      <div slot="checked-children">
+        <ds-alert heading="You'll save more with this option">
+          Based on the household information you provided, this option will give you the maximum
+          savings. We are adding some filler text just to show what it looks like when you have a
+          long alert as the checkedChildren of a Choice component.
+        </ds-alert>
+      </div>
+    </ds-choice>
+    <ds-choice type="checkbox" label="Baz" value="baz">
```
pwolfert, line 170: I think it would be helpful for people interacting with this story if the checkbox label or hint text explained what these checked children are doing. It's a little weird to see the unchecked children right from the get-go with the "Foo Bar Baz", even if it's useful to us for visual regression testing.

I think we should also apply the same `ds-c-choice__checkedChild` class that we do in the React ChoiceList story for consistency.

packages/design-system/src/components/web-components/ds-choice-list/ds-choice-list.stories.tsx
```diff
@@ -0,0 +1,188 @@
...
+      },
+    },
+  },
+  decorators: [webComponentDecorator],
+};
+
+const Template = (args) => {
+  useEffect(() => {
+    const onChange = (event) => {
+      action('ds-change')(event);
```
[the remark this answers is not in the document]
pwolfert: Fixed in https://github.com/CMSgov/design-system/pull/3082

packages/design-system/src/components/web-components/ds-choice-list/ds-choice-list.stories.tsx
zarahzachz: Should we remove the unchecked markup?

pwolfert: If you can come up with something that seems plausible, it seems fine to me to include it. But if you can't come up with something plausible, removing it sounds fine too.

tests/browser/snapshots/stories/web-components-choicelist--html-choices-cmsgov--Mobile-Chrome.png
pwolfert, line 1: This one screenshot isn't like the others for some reason

pwolfert: It doesn't have the blue bar on the left

packages/design-system/src/components/web-components/ds-choice-list/ds-choice-list.stories.tsx
```diff
@@ -0,0 +1,206 @@
...
+    docs: {
+      page: WebComponentDocTemplate,
+      description: {
+        component:
+          'For information about how and when to use this component, refer to the [checkbox](https://design.cms.gov/components/checkbox/) and [radio](https://design.cms.gov/components/radio/) documentation pages. Checkboxes and radios can be managed as a group using `<ChoiceList>` or individually using `<Choice>`. Note that each of the items in the `choices` array represents props that will be passed to an individual `<Choice>` component. You can therefore define any of the props listed in the `<Choice>` props table below, including all valid attributes of the [HTML input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input).',
+      },
+      componentEvents: {
+        'ds-change': {
+          description:
+            'A callback function that\'s invoked when a month\'s checked state is changed. Note: This callback is not called when a month is selected or deselected via the "Select all" or "Clear all" buttons – use the `onSelectAll` and `onClearAll` event handlers for those instances.',
```
pwolfert, line 116: Refers to _months_

packages/design-system/src/components/web-components/ds-choice-list/ds-choice-list.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+        /**
+         * Begin logic to define optional nested children of a choice.
+         */
+        let checkedChild,
+          uncheckedChild = undefined;
+
+        if (element.props.children.length > 0) {
+          element.props.children.map((child) => {
+            const { children, slot } = child.props;
```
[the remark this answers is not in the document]
pwolfert: This hasn't been fixed yet. A type of `string | React.ReactElement` will reveal the error, which needs a code solution to be resolved.

packages/design-system/src/components/web-components/ds-choice-list/ds-choice-list.test.tsx
```diff
@@ -0,0 +1,223 @@
...
+    expect(choice.name).toBe('spec-field');
+    expect(choice.value).toBe('1');
+  });
+
+  it('accepts HTML as children', () => {
+    const { asFragment } = render(
+      <ds-choice-list type="checkbox">
+        <ds-choice label="Foo" value="foo"></ds-choice>
+        <ds-choice label="Bar" value="bar"></ds-choice>
+      </ds-choice-list>
```
pwolfert, line 82: 👍

packages/design-system/src/components/web-components/ds-choice-list/ds-choice-list.tsx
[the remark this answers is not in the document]
zarahzachz: I'm sorry, I don't understand what the bug is here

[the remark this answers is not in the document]
pwolfert: If you add it to the Astro example with slots in them, it will crash. Some of the `children` in those `ds-choice` children happen to be whitespace characters, which are just strings. So when line 67 tries to access `child.props`, it will get undefined, and then when it tries to pull `{children, slot}` out of it, it will crash because it's trying to access properties of undefined.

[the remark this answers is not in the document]
pwolfert: Adding a type annotation for the `child` parameter on line 66 will more accurately reflect what possible values are being passed to that function, and TypeScript will know that you can't access `props` from something that is possibly just a string.

packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx
```diff
@@ -1,7 +1,7 @@
 import { useEffect } from 'react';
 import WebComponentDocTemplate from '../../../../../../.storybook/docs/WebComponentDocTemplate.mdx';
 import { action } from '@storybook/addon-actions';
-import './ds-choice';
+import '../ds-choice';
```
pwolfert, line 4: Looks like this might have been accidental

tests/browser/snapshots/stories/web-components-choicelist--html-choices-cmsgov--Mobile-Chrome.png
pwolfert: Can you run this one again and see if it changes, @zarahzachz ?

zarahzachz: it's also got the old labels

pwolfert (approved):
It's working great!

tests/browser/snapshots/stories/web-components-choicelist--html-choices-cmsgov--Mobile-Chrome.png
pwolfert: I see the right file locally too. GitHub must be confused

PR #3084, opened 2024-05-14
[NO-TICKET] Combine `ds-choice` and `ds-choice-list` into a single bundle

pwolfert:
## Summary

Combine `ds-choice` and `ds-choice-list` into a single bundle like we do for `ds-accordion` and `ds-accordion-item`. A little simpler for users and reduces waste in the form of duplicated JavaScript if they end up importing both.

## How to test

1. Delete `packages/design-system/dist`
2. `yarn build:core`
3.  Look at the output bundles in dist
4. Try out the astro example

## Checklist


packages/design-system/src/components/web-components/ds-choice/ds-choice-list.test.tsx
```diff
@@ -1,7 +1,7 @@
 import { render, waitFor, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import './ds-choice-list';
-import '../ds-choice';
+import '.';
```
zarahzachz, line 4: I don't think I've seen this syntax before. Is it shorthand for importing `choice`?

zarahzachz (approved):
Aside from 1 syntax question this looks good to me

packages/design-system/src/components/web-components/ds-choice/ds-choice-list.test.tsx
pwolfert: Whoops, that was vscode's automatic import refactor

PR #3089, opened 2024-05-15
[WNMGDS-2760] Web component pass-thru attributes

zarahzachz:
WNMGDS-2760

Supporting pass-thru props on web components proves to be challenging (because our solution requires you define the attribute at the top and doesn't allow `...randomProps` type syntax like React), so we opted for creating a series of allow lists with attributes we expect ppl to pass through the component. This has the added benefit of consolidating commonly used attributes in their respective files, making it easier to build new components in the future without accidentally omitting something (e.g., form field components will all use the `formAttrs` list, components with analytics will use `analyticsAttrs`, etc.).


packages/design-system/src/components/StepList/StepLink.tsx
```diff
@@ -16,7 +16,7 @@ export interface StepLinkProps {
   screenReaderText?: string;
   className?: string;
   onClick?: (href?: string, stepId?: string) => any;
-  component?: React.ComponentType<StepLinkComponentProps> | keyof JSX.IntrinsicElements;
+  component?: React.ComponentType<StepLinkComponentProps> | 'a';
```
zarahzachz, line 19: This was changed to fix a build error happening with web component work. Web components are defined as intrinsic elements globally (I think this was to fix some issues with `ds-choice` and testing) and as a result, StepLink component types included all web components.

Looking at StepLink, it has an `href` prop so why does it allow for any component type when this prop indicates it should be a link? There's also an `onClick` applied to this link that shouldn't exist (`onClick` is button behavior, links use the History API afaik so this sort of manual controlling of browser history is frowned upon - afaik).

In short, this should fix it and StepLink needs future love.

packages/design-system/src/components/web-components/shared-attributes/accessibility.ts
```diff
@@ -0,0 +1,50 @@
...
+  'aria-busy',
+  'aria-checked',
+  'aria-colcount',
+  'aria-colindex',
+  'aria-colspan',
+  'aria-controls',
+  'aria-current',
+  'aria-describedby',
+  'aria-details',
+  // 'aria-disabled', // This prop resulted in errors in Astro.
```
zarahzachz, line 16: Including `aria-disabled` resulted in rendering issues on Astro. It's not entirely clear to me why.

I tested this pass-thru logic locally on Astro's web components (ds-choice, ds-dropdown) and it placed the attrs where they ought to be. I used something crazy like `aria-checked="mixed"` so it would stand out.

zarahzachz: Oh - I also thought I would've needed to include this on ds-choice-list and ds-month-picker, but neither of those have the same prop spreading as the more atomic elements do.

.storybook/docs/HtmlElementArgs.tsx
```diff
@@ -37,7 +36,7 @@ export const HtmlElementArgs = ({ attributes, of }) => {
 
   return (
     <>
-      <h3 id="additional-props">Additional {attributes ? 'attributes' : 'props'}</h3>
```
zarahzachz, line 40: This block didn't really apply to our web components because spreading any prop doesn't actually filter through the host web component element. As a result, we've created these attr lists with pre-defined attributes one might want to pass along to an internal element. 

So I removed the "attribute" language from this component and created a new one to work for web component docs.

packages/design-system/src/components/web-components/ds-badge/ds-badge.stories.tsx
```diff
@@ -32,7 +32,6 @@ export default {
       description: {
         component: `For information about how and when to use this component, [refer to its full documentation page](https://design.cms.gov/components/badge/).`,
       },
-      underlyingHtmlElements: ['span'],
```
zarahzachz, line 35: Removing these bits from our web components docs because it's inaccurate. I even tested this in our Astro project to confirm, but applying an attribute for `span` onto `ds-badge` doesn't result in that attribute being applied to the correct element.

packages/design-system/src/components/web-components/shared-attributes/accessibility.ts
pwolfert: I'm actually not 100% sure we should be adding these aria ones to the attributes of the web component. What happens with assistive tech when both the web component root and one of its children has an `aria-label` on it that is the same? We might have to pass the aria attributes down using some other alias like how we have `root-id` that passes an `id`.

I think all the other lists are good, but this one might need more research.

.storybook/docs/WebComponentAttrs.tsx
```diff
@@ -0,0 +1,49 @@
...
+ * array in `story.parameters.docs` to determine if it should show this additional
+ * text and what elements' documentation it should link to.
+ */
+export const WebComponentAttrs = ({ of }) => {
+  const resolvedOf = useOf(of || 'story', ['story', 'meta']);
+  if (resolvedOf.type !== 'story') {
+    return null;
+  }
+
+  const elements = resolvedOf.story.parameters?.docs?.underlyingAttrLists;
```
pwolfert, line 22: The "underlying" part has perhaps lost its meaning. Naming suggestion: `sharedAttributesLists`

pwolfert (approved):
I think we might need to do some more research on the effects of the ARIA list, but other than that, this LGTM 👍

packages/design-system/src/components/web-components/shared-attributes/accessibility.ts
zarahzachz: OK I'll keep the aria list and remove the imports from 🚾 so we can experiment in the future if we have time

PR #3090, opened 2024-05-15
[NO-TICKET] Generate the top-level CDN index

pwolfert:
## Summary

In the first version of the top-level CDN index, we manually uploaded a file to our CDN that could be appended to by our deploy script, and it would render new versions with JavaScript. For long-term maintenance of the project, this isn't the most ideal, because we have to SFTP into the CDN in order to make updates to that file. I've added to the script that generates the CDN index for each package version and had it generate a top-level index in using the same layout. What we lose in this process is the top-level index's ability to retain every single version of the design system in this list. It instead uses what's available in the `versions.json` file. What we gain is the ability to generate the index file at any time to deploy to the CDN. Some benefits:

- The top-level index and the package version indices can stay in sync visually
- We can replace the top-level index if we lose it
- We have a straightforward way to make updates to it and deploy it (through the normal asset deploy process)

## How to test

1. `yarn build:cdn-index`
2. Open `dist/index.html` in a browser, and compare it with the package ones and [the production index](https://design.cms.gov/cdn/)

## Checklist


scripts/build-cdn-indexes.ts
```diff
@@ -1,37 +1,99 @@
...
+    const renderItem = (version: string) => `
+      <li>
+        <a href='https://design.cms.gov/cdn/${packageName}/${version}'>${version}</a>
+      </li>
+    `;
+
+    return `
+      <section class="ds-u-margin-top--3 ds-u-measure--wide">
+        <h2>${packageName}</h2>
+        <ul style="column-count: 4; display: block">
```
zarahzachz, line 58: an a11y enhancement for this could be adding an `id` to the `<h2>` and `aria-labelledby` on the lists to give the lists a name. otherwise looks good.

zarahzachz: `column-count` looks a little goofy on mobile but i don't know if anyone cares about mobile when referencing CDN

scripts/build-cdn-indexes.ts
```diff
@@ -1,37 +1,99 @@
...
+  return `<!DOCTYPE html>
+  <html lang="en">
+    <meta charset="utf-8" />
+    <head>
+      <meta name=“viewport” content=“width=device-width, initial-scale=1, shrink-to-fit=no”>
+      <title>${title} - CMSDS</title>
+      <link rel="stylesheet" href="https://design.cms.gov/cdn/${system}/${version}/css/index.css" />
+      <link rel="stylesheet" href="https://design.cms.gov/cdn/${system}/${version}/css/${theme}-theme.css" />
+    </head>
+    <body class="ds-content" style="margin: 0">
```
zarahzachz, line 27: is ds-content needed here?

scripts/build-cdn-indexes.ts
```diff
@@ -1,37 +1,99 @@
...
+      <ds-usa-banner></ds-usa-banner>
+      <header class="ds-base--inverse ds-u-padding-y--3">
+        <div class="ds-l-container">
+          <h1 class="ds-text-heading--2xl">${title}</h1>
+        </div>
+      </header>
+      <div class="ds-l-container ds-content ds-u-padding-y--4">
```
zarahzachz, line 34: when you have ds-content here as well?

zarahzachz (approved):
this looks good. added minor optional feedback.

i know you said a downside for this is not being able to include all our DS versions in this CDN, but if push comes to shove and a team REALLY needs a specific version, they could probably use [unpkg](https://www.unpkg.com/) to get it. another neat enhancement could be indicating which versions aren't supported by the DS

scripts/build-cdn-indexes.ts
pwolfert: oh, good call. I'll remove one of them

scripts/build-cdn-indexes.ts
pwolfert: I'm having trouble figuring out why the page looks weird on mobile. I have the viewport meta tag that is supposed to fix that.

pwolfert: It was that my quotes weren't real quotes on the meta tag 🤦‍♂️

scripts/build-cdn-indexes.ts
```diff
@@ -17,14 +17,14 @@ function renderPageHtml(theme: keyof typeof themes, title: string, mainContent:
   const version = packageVersions[system as keyof typeof packageVersions][0];
   return `<!DOCTYPE html>
   <html lang="en">
-    <meta charset="utf-8" />
     <head>
-      <meta name=“viewport” content=“width=device-width, initial-scale=1, shrink-to-fit=no”>
```
pwolfert: This wasn't using valid quote characters 🤦‍♂️

PR #3095, opened 2024-05-30
[WNMGDS-2784] Change end to flex-end to quiet another teams console warn

zarahzachz:
WNMGDS-2784

UsaBanner had a style using `justify-content: end` and another team was receiving a support warning for it. [This style is widely supported](https://caniuse.com/?search=justify-content%3A%20end), but I think the team supports an older browser or something. Regardless, [because the element this style's applied to has a `display: flex`](https://developer.mozilla.org/en-US/docs/Web/CSS/justify-content#flex-end) also applied, changing this to `flex-end` seemed appropriate.


pwolfert (approved):
LGTM 👍

PR #3099, opened 2024-05-31
[WNMGDS-2779] Replace dropdown label with div

zarahzachz:
WNMGDS-2779

Replacing visible `<label>` on Dropdown with a `<div>` (I know the ticket said to use a `<span>`, but `<div>` is already a block level element and I didn't want to add the extra style to make `<span>` do that).

Aside from some markup changes, there should be no visual/functional change in this component.

### DIFF
--- a/packages/design-system/src/components/Dropdown/Dropdown.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.tsx
@@ -248,7 +248,6 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
     ...useLabelProps({
       ...props,
       labelClassName: classNames('ds-c-label', 'ds-c-dropdown__label', props.labelClassName),
-      fieldId: id,
     }),
   };
 
--- a/packages/design-system/src/components/Dropdown/Dropdown.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.tsx
@@ -9,7 +9,6 @@ import mergeRefs from '../utilities/mergeRefs';
 import useClickOutsideHandler from '../utilities/useClickOutsideHandler';
 import useId from '../utilities/useId';
 import useAutofocus from '../utilities/useAutoFocus';
-import { Label } from '../Label';
 import { SvgIcon } from '../Icons';
 import { getFirstOptionValue, isOptGroup, parseChildren, validateProps } from './utils';
 import { Item, Section, useSelectState } from '../react-aria'; // from react-stately
@@ -246,8 +245,11 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
 
   const labelProps = {
     ...useSelectProps.labelProps,
-    ...useLabelProps({ ...props, id }),
-    fieldId: id,
+    ...useLabelProps({
+      ...props,
+      labelClassName: classNames('ds-c-label', 'ds-c-dropdown__label', props.labelClassName),
+      fieldId: id,
+    }),
   };
 
   const buttonProps = {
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -246,8 +245,10 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
 
   const labelProps = {
     ...useSelectProps.labelProps,
-    ...useLabelProps({ ...props, id }),
-    fieldId: id,
```
zarahzachz, line 250: The inclusion of both `id` and `fieldId` seemed unnecessary. Removing both resulted in no changes in snapshot tests and I was able to use Storybook controls to confirm it still accepts user-defined ids. 

Do you know why these were added?

pwolfert: Normally the `fieldId` would be supplied to the `<Label>` component to use in its `for` attribute to point it to the input field. The `id` is passed to the `useLabelProps` only so that it can be used inside the generated `labelId` [on this line](https://github.com/CMSgov/design-system/blob/WNMGDS-2779/change-dropdown-markup-for-wave/packages/design-system/src/components/Label/useLabelProps.ts#L37). However, in this case, we didn't actually need `fieldLabel` because the dropdown button links itself to the label via its `aria-labelledby` attribute.

I think it's right to remove `fieldId`, but removing `id` from the object passed to `useLabelProps` results in the label having the following `id` in the HTML: `undefined__label`. You can see this on your branch if you open Storybook and inspect the label.

zarahzachz: OK I think I know what you're saying here and wanna make sure we're on the same page then.

In the DOM...
1. `id` on the "label" should default to "dropdown__button--<NUMBER>" + the word "__label"
2. `id` on the `<button>` should default to "dropdown__button--<NUMBER>"
3. `aria-labelledby` on the `<button>` should default to "dropdown__button--<NUMBER>__button-content dropdown__button--<NUMBER>__label"
4. When the user changes the `id` prop using the Storybook control, that string value would replace **all instances** of "dropdown__button--<NUMBER>"
    1. "label": `id="foo_label"`
    2. `<button>`: `id="foo" aria-labelledby="foo__button-content foo__label"`

I'm being pretty granular with my expectations because I think I might've found [a bug in prod](https://design.cms.gov/storybook/?path=/story/components-dropdown--default&args=id:foo&globals=theme:core) - changing the `id` prop doesn't reflect in the DOM and I believe it's because in the JSX file, `id` comes after props are spread? I can fix this bug in this PR to do the above expected behavior but wanted to clarify with you on this first.

zarahzachz: Hmm... maybe disregard. It looks like if you change that value in Storybook it doesn't update the DOM, but if you navigate to that change via the URL query param it does reflect. So I guess this isn't a bug after all!

pwolfert: Hmm, item 1 does look weird. We probably want to use [a pattern like this](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/Alert/Alert.tsx#L74-L76) where we derive both the button id and the label id from the same root id so it becomes `dropdown--X__label` and `dropdown--X__button` instead of `dropdown--X__button--__label`.

pwolfert: I think the reason we made it this way originally was so that if a user specifies an `id`, that id definitely gets put on a real element they can link to.

pwolfert: Maybe actually all we'd want to do is change [this line](https://github.com/CMSgov/design-system/pull/3099/files#diff-c132558c7ba0b4e130880148a65be06ed0cb42a950020183fb88bc11a899a351R141) to

```ts
const id = useId('dropdown--', props.id);
```

pwolfert (approved):
LGTM 👍

PR #3100, opened 2024-05-31
[WNMGDS-2785] Fix SCSS token generation and consumption

pwolfert:
## Summary

This is one of several PRs for WNMGDS-2785

- Fixes error caused by empty `dist` directory
- Fixes usage and generation of the scss layout files
- Fixes a TypeScript error in the token-related docs code

## How to test

Try `yarn build:tokens` and running Storybook or the doc site (note that there will still be CSS issues)


zarahzachz (approved):
Looks good locally. Only issue I see is missing fonts, which I believe is being solved in another PR. Also USA banner and form width tokens are missing, but again, probably handled in a different PR

pwolfert:
Right, it only fixes the three things listed, and there's another series coming that will fix actual CSS issues

PR #3103, opened 2024-06-03
[WNMGDS-2779] Update wc snaps, add component logic to fix wc test

zarahzachz:
WNMGDS-2779 (related to #3099)

- Update snapshots for 🚾 
- Add logic supporting inverse labels

### DIFF
--- a/packages/design-system/src/components/Dropdown/Dropdown.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.tsx
@@ -248,7 +248,12 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
     ...useLabelProps({
       ...props,
       id,
-      labelClassName: classNames('ds-c-label', 'ds-c-dropdown__label', props.labelClassName),
+      labelClassName: classNames(
+        'ds-c-label',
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -248,7 +248,12 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
     ...useLabelProps({
       ...props,
       id,
-      labelClassName: classNames('ds-c-label', 'ds-c-dropdown__label', props.labelClassName),
+      labelClassName: classNames(
+        'ds-c-label',
+        'ds-c-dropdown__label',
+        props.inversed && 'ds-c-label--inverse',
```
pwolfert, line 254: It looks like this was missing prior to https://github.com/CMSgov/design-system/pull/3099, but I see [the inverse story working on prod](https://design.cms.gov/storybook/?path=/story/components-dropdown--inverse-option). What happened?

pwolfert (approved):
LGTM once we see that the CI job is now checking this successfully 🎉

PR #3111, opened 2024-06-07
[WNMGDS-2785] Updated Accordion with Figma changes

pwolfert:
## Summary

Updated Accordion with Figma changes.

### BREAKING changes
- Removed the healthcare-only `--accordion-icon__size` variable
- Split `--accordion__border-radius` into the following variables to support Figma:
    - `--accordion__border-bottom-left-radius`
    - `--accordion__border-bottom-right-radius`
    - `--accordion__border-top-left-radius`
    - `--accordion__border-top-right-radius`

### DIFF
--- a/packages/design-system/src/styles/components/_Accordion.scss
+++ b/packages/design-system/src/styles/components/_Accordion.scss
@@ -18,7 +18,7 @@ ds-accordion-item:not(:first-child) {
 .ds-c-accordion__button {
   background-color: var(--accordion__background-color);
   border: 0;
-  border-radius: var(--accordion__border-radius);
+  border-radius: var(--accordion__border-top-left-radius) var(--accordion__border-top-right-radius) var(--accordion__border-bottom-right-radius) var(--accordion__border-bottom-left-radius);
   color: var(--accordion-button__color);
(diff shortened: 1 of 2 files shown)


packages/design-system/src/styles/components/_Accordion.scss
```diff
@@ -18,7 +18,7 @@ ds-accordion-item:not(:first-child) {
 .ds-c-accordion__button {
   background-color: var(--accordion__background-color);
   border: 0;
-  border-radius: var(--accordion__border-radius);
```
zarahzachz, line 21: I'm trying to determine why this needs to be broken up like this and it probably boils down to a  Figma limitation... `border-radius` is a shorthand property that can receive multiple values and CSS vars are just placeholders for strings (essentially), so could we keep `--accordion__border-radius` as a token and in the theme, assign it the values you want?

Like for mgov, maybe it's `--accordion__border-radius: .25em`, but for hgov it's `--accordion__border-radius: .25em .5em 2px`. Can Figma support that string of ".25em .5em 2px"?

zarahzachz (approved):

PR #3112, opened 2024-06-07
[WNMGDS-2785] Update USA Banner with Figma changes

pwolfert:
## Summary

Fixed visual regressions in USA Banner by addressing token changes.

### BREAKING changes:

- Removed the following CSS variables from the theme files:
    - `--usa-banner-gutter-width`
        - This variable can continue to be used by teams to modify the component, but we will no longer be providing a default value in the theme files because it is redundant information.
    - `--usa-banner__max-width`
        - Similar to the previous variable, teams can still use it to change the max width from the default, but we'll no longer be declaring it.
    - `--usa-banner-heading__font-family`
    - `--usa-banner-heading__font-size`
    - `--usa-banner-heading__line-height`
    - `--usa-banner-panel__font-family`
    - `--usa-banner-panel__font-size`
    - `--usa-banner-panel__line-height`

## How to test

Try setting a `--usa-banner-gutter-width` or `--usa-banner__max-width` in the browser in Storybook


packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -15,18 +15,24 @@
   --security-icon__color: #12890e; // spring-500
 
   background-color: var(--usa-banner__background-color);
-  color: var(--usa-banner-color);
```
zarahzachz, line 18: Here's a question... if USABanner doesn't define its color, it falls back to the color defined on `body` (`--global-color`). I did some playing around in Storybook and it seems like we'd never have a need to change this color to be different from the `body`, but I could be wrong.

If I'm not wrong, then could this rule be written as `color: var(--usa-banner-color, currentColor)` so banner defaults to inheriting the current color of its parent and allows for users to override this if that's not wanted?

zarahzachz (approved):
I have one possible change for a color token, but looks good.

Also, I thought I found another optimization/consolidation in `--usa-banner-link__color` since I figured all banners would default to whatever link color a theme has specified. This is unfortunately not true for Medicare and again, [I am disappointed](https://www.youtube.com/watch?v=cdEQmpVIE4A).

packages/design-system/src/styles/components/_UsaBanner.scss
pwolfert: Are you saying it would allow us to feel secure in removing the `--usa-banner-color` from our tokens?

zarahzachz: Yeah, not defining it would result in it falling back like `currentColor`. Doing it this way would just be a little safer.

pwolfert: Hmm, I actually think that the link color for medicare is wrong. If you look at [the production site](https://www.medicare.gov/), it uses the normal link color, and that seems more correct. Maybe we should get rid of both variables.

zarahzachz (approved):

PR #3114, opened 2024-06-10
[NO-TICKET] Add support for relative `em` units

pwolfert:
## Summary

There really isn't a reason not to support this unit generally, and I may have to use it pretty soon.

### DIFF
--- a/packages/design-system-tokens/src/lib/conversions.ts
+++ b/packages/design-system-tokens/src/lib/conversions.ts
@@ -8,6 +8,11 @@ export function pixelNumberToRem(pixels: number): string {
   return `${pixels / PIXELS_PER_REM}rem`;
 }
 
+export function pixelNumberToEm(pixels: number): string {
+  // For Figma, count rem and em as equal
+  return `${pixels / PIXELS_PER_REM}em`;
+}
+
 export function pixelNumberToEx(pixels: number): string {
   return `${Math.round(pixels / PIXELS_PER_EX)}ex`;
 }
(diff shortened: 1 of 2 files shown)


packages/design-system-tokens/src/lib/conversions.ts
```diff
@@ -8,6 +8,11 @@ export function pixelNumberToRem(pixels: number): string {
   return `${pixels / PIXELS_PER_REM}rem`;
 }
 
+export function pixelNumberToEm(pixels: number): string {
+  // For Figma, count rem and em as equal
+  return `${pixels / PIXELS_PER_REM}em`;
```
zarahzachz, line 13: This might be a trickier unit to support. `em` is based on the parent's font size, not the root. So with `rem`, it's safe-ish to assume there are always 16px/`rem` (unless a user updated their browser settings). With `em` it's different. E.g., if a parent sets a font-size equal to 20px, .5em = 10px, but if the parent has a font size equal to 32px, .5em is 16px.

pwolfert: That's a fair point. I'm not sure any way around it if we want ems in tokens. If there became an issue with using it inside Figma, I think we'd have to not use the variable on the Figma side to avoid issues with translating to JSON.

zarahzachz: I think because of the relative-heavy nature of `em` and `<percent>` (that both units are based on a variable parent measurement and not a constant), and because we're using Figma tokens as source of token values, it might be best to avoid them in this work.

pwolfert: We were able to avoid it in a few different places where they were previously used already. I was hoping to use them in the text-decoration variables, but we would still be able to avoid that if we wanted to stick with the fixed sizes for those things (underline thickness and offset).

pwolfert: I'll leave this PR open until we know what we're doing with link tokens. If we end up avoiding `em`s there, I'll close it.

pwolfert:
Probably don't need this

PR #3120, opened 2024-06-11
[NO-TICKET] Fix broken doc-site build

pwolfert:
## Summary

Don't print any of the boolean values on the doc site. Those boolean tokens are mostly internal and are not robust enough to set to different values. They shouldn't be shown in the docs as configuration options. This fixes build because I made the CSS value function to throw an error if we try to get the value of a boolean token.


zarahzachz (approved):
Site builds great!

Side quest: I was spot checking our docs site and noticed we still include links to Sketch in the docs (like `/foundation/spacing/?theme=core`). I'm not sure if this work is captured in a ticket?

PR #3122, opened 2024-06-12
[NO-TICKET] Remove inversed prop from label-div

zarahzachz:
No ticket.

`useLabelProps` spread the attr `inversed` onto the newly created label-div element (`<div>` styled as a `<label>` to skirt some Axe/Wave issues). Because the `div` doesn't have the concept of an `inversed` attr, and I'm using `classNames` to get the same effect, remove that prop from `labelProps`.

### DIFF
--- a/packages/design-system/src/components/Dropdown/Dropdown.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.tsx
@@ -155,7 +155,6 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
     defaultValue,
     value,
     inputRef,
-    inversed,
     getA11yStatusMessage,
     getA11ySelectionMessage,
     ...extraProps
@@ -257,6 +256,9 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
     }),
   };
 
+  // Excluding `inversed` prop from `<div>` label because it's not a valid attr
+  const { inversed, ...divLabelProps } = labelProps;
+
   const buttonProps = {
     ...useButtonProps.buttonProps,
     ...cleanFieldProps(extraProps),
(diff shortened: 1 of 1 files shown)


packages/design-system/src/components/Dropdown/Dropdown.tsx
```diff
@@ -257,6 +256,9 @@ export const Dropdown: React.FC<DropdownProps> = (props: DropdownProps) => {
     }),
   };
 
+  // Excluding `inversed` prop from `<div>` label because it's not a valid attr
+  const { inversed, ...divLabelProps } = labelProps;
```
pwolfert, line 260: For some reason I'm still getting the same error when I run the unit tests:

```
console.error
      Warning: Received `true` for a non-boolean attribute `inversed`.
      
      If you want to write it to the DOM, pass a string instead: inversed="true" or inversed={value.toString()}.
          at button
          at div
          at Dropdown (/Users/patrickwolfert/Projects/design-system/packages/design-system/src/components/Dropdown/Dropdown.tsx:139:16)
```

Does the error go away for you?

zarahzachz: I ran `yarn test` locally and everything passed. When I built locally, the `inversed` attr was removed so I assumed it was fixed

zarahzachz: Wait! Maybe this is a web component issue? I need to investigate that when I sign on tomorrow.

zarahzachz: J/k, I think I've fixed it for real this time. Wasn't a 🚾 issue

pwolfert: Nice, that fixes it 👍 

Alternatively, it might be clearer what is going on if you leave the first one as `inversed,` and change this second one to something like

```suggestion
  const { inversed: _removedInversed, ...divLabelProps } = labelProps;
```

pwolfert (approved):
Made a clarity suggestion above, but this is fixed ☑️

PR #3123, opened 2024-06-13
[WNMGDS-2797] Update links with Figma changes

pwolfert:
## Summary

- We can no longer use `auto` as a value for the underline offset and still be compatible with Figma. We have to pick a value. I've set values for core and healthcare that are basically equivalent.
- "Fixed" ghost-button underlines to match the thickness and offset of the links for each theme.
- Updated affected VRTs

## How to test

For reviewing, take a look at the individual commits


tests/browser/snapshots/stories/components-autocomplete--custom-markup-cmsgov--chromium-forced-colors.png
zarahzachz, line 1: just noticed this button text treatment in HCM. not part of this ticket, but might need revised in a future ticket

zarahzachz (approved):
just one observation that's unrelated to your work. approving!

tests/browser/snapshots/stories/components-autocomplete--custom-markup-cmsgov--chromium-forced-colors.png
pwolfert: Oh yeah, that doesn't look great

PR #3129, opened 2024-06-20
[WNMGDS-2800] Refactor alert so headingId is on heading node, not alert body

zarahzachz:
WNMGDS-2800

The current Alert assigned `headingId` to the body element, which contained text for the heading, alert type and body text (aka, wall-o-text).

This update assigns `headingId` to the element containing just the heading text and alert type, making it easier for AT users to scan a given page. If no heading or icon are provided, the accessible label will default to announcing the alert type.

You can test these changes by comparing PROD and local builds, either by using VO or by opening your browser's accessibility tools to see what the label is.

Local readout:

Screenshot 2024-06-20 at 10 52 18 AM

PROD:

Screenshot 2024-06-20 at 10 53 15 AM

### DIFF
--- a/packages/design-system/src/components/Alert/Alert.tsx
+++ b/packages/design-system/src/components/Alert/Alert.tsx
@@ -151,9 +151,9 @@ export const Alert: React.FC<AlertProps> = (props: AlertProps) => {
       {...alertProps}
     >
       {getIcon()}
-      <div className="ds-c-alert__body" id={headingId} ref={bodyRef}>
+      <div className="ds-c-alert__body" ref={bodyRef}>
         {heading ? (
-          <div className="ds-c-alert__header ds-c-alert__heading" ref={headingRef}>
+          <div id={headingId} className="ds-c-alert__header ds-c-alert__heading" ref={headingRef}>
             {a11yLabel}
             {headingElement}
           </div>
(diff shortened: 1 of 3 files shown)


packages/design-system/src/components/Alert/Alert.tsx
```diff
@@ -151,9 +151,9 @@ export const Alert: React.FC<AlertProps> = (props: AlertProps) => {
       {...alertProps}
     >
       {getIcon()}
-      <div className="ds-c-alert__body" id={headingId} ref={bodyRef}>
+      <div className="ds-c-alert__body" ref={bodyRef}>
         {heading ? (
-          <div className="ds-c-alert__header ds-c-alert__heading" ref={headingRef}>
+          <div id={headingId} className="ds-c-alert__header ds-c-alert__heading" ref={headingRef}>
```
pwolfert, line 156: What happens if there is no heading?

pwolfert: Oh, I see that the label goes to `heading ? headingId : a11yLabelId`

pwolfert (approved):
I'm not really sure how to get VoiceOver to read this. I'm getting kind of weird results myself. The accessibility tree seems right, though.

packages/design-system/src/components/Alert/Alert.tsx
zarahzachz: That ternary gives me more reason to think the original placement of the `headingId` was incorrect.

PR #3131, opened 2024-06-20
[WNMGDS-2797] Remove `ds-u-sans-serif` and `ds-u-serif` utility classes

pwolfert:
## Summary

WNMGDS-2797

Removes the `ds-u-sans-serif` and `ds-u-serif` classes and related variables. We do not have theme-level fonts that universally fill those categories, and we don't find that these are useful as utility classes. Font families should be inherited from a theme's components or base style and not fiddled with directly.

## Checklist


zarahzachz (approved):

packages/docs/static/themes/medicare-theme.css
```diff
@@ -444,7 +444,6 @@
 --color-warn-darker: #ae8916;
 --color-warn-darkest: #7c6210;
 --color-visited: #590000;
---font-sans: Rubik, Seravek, 'Gill Sans Nova', Ubuntu, Calibri, 'DejaVu Sans', source-sans-pro, sans-serif;
```
zarahzachz, line 447: did a local search for these variables and this one popped up in the `design-system-tokens/src/css/__snapshots__` dir.

pwolfert: That snapshot will get updated pretty soon

PR #3132, opened 2024-06-20
[NO-TICKET] Fix regression in max-width of USA banner on doc site

pwolfert:
## Summary

Fix regression in max-width of USA banner on doc site. For some reason a value of `initial` seems to trigger the `var()` fallback mechanism instead of just removing the max width. I guess that's how custom properties work?

## How to test

Make sure the doc site's USA banner content spans the full width like the production one.

### DIFF
--- a/packages/docs/src/styles/pages/layout.scss
+++ b/packages/docs/src/styles/pages/layout.scss
@@ -3,7 +3,7 @@ body {
 }
 
 .ds-c-usa-banner__header {
-  --usa-banner__max-width: initial;
+  --usa-banner__max-width: none;
 }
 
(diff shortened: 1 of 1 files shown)


zarahzachz (approved):
This seems to work.

Side note: Built the docs site locally and noticed fonts aren't loading. Not sure if there's another regression?

pwolfert:
> Side note: Built the docs site locally and noticed fonts aren't loading. Not sure if there's another regression?

I've had that happen to me, and then I build a bunch of stuff and it works, so there might be a new gap somewhere in the build scripts. Maybe we can debug it together.

PR #3136, opened 2024-06-25
[WNMGDS-2801] `aria-live` and `aria-atomic` regions should always be present in DOM

zarahzachz:
WNMGDS-2801

An element with `aria-live` and `aria-atomic` attrs applied should always be visible in DOM. AT won't read the messaging out if the element is added/removed from a page, just when the contents of the element changes. 

To satisfy this, I've made InlineError always visible in the DOM and added logic around the messaging to show/hide so it's correctly announced.

Note: This change won't take place inside the individual Choice component - getting guidance from Melissa on how that should work/look.


packages/design-system/src/components/ChoiceList/Choice.tsx
```diff
@@ -110,16 +110,12 @@ export const Choice = (props: ChoiceProps) => {
...
-    errorId = props.errorId ?? `${props.id}__error`;
-    errorElement = (
-      <InlineError id={errorId} inversed={props.inversed} className={props.errorMessageClassName}>
-        {props.errorMessage}
-      </InlineError>
-    );
-  }
+  const errorId = props.errorId ?? `${props.id}__error`;
+  const errorElement = (
+    <InlineError id={errorId} inversed={props.inversed} className={props.errorMessageClassName}>
```
zarahzachz, line 115: I have a question around how our Choice/ChoiceList should handle this change.

Choice is used by ChoiceList. Individual Choices have their own InlineError. ChoiceList has its own InlineError (using `useInlineError`).

If I do what's asked (always show the live/atomic regions) there are at least two live/atomic regions shown (one for ChoiceList and one for Choice). I'm 99% sure if a Choice is used in ChoiceList, it shouldn't display a live/atomic region but I'm unsure how a child element would be aware if its being used in ChoiceList to know not to display some code.

I believe this logic is also screwing up some snap tests (seeing undefined IDs in 🚾).

pwolfert: I think if we were to implement any kind of logic that took into account the context, it should go in another PR as an enhancement. The individual choice's inline error is really only supposed to be used for lone checkboxes, but we don't currently restrict it. I don't think it's a valid design to put an error message on a choice inside a choice list, so it's not something we _should_ be seeing in production.

pwolfert: And also I think that enhancement would be to start blocking the use of choice inline error messages in choice list

packages/design-system/src/components/ChoiceList/__snapshots__/Choice.test.tsx.snap
```diff
@@ -25,6 +26,12 @@ exports[`Choice accepts a node as innerHTML 1`] = `
            World
         </p>
       </label>
+      <p
+        aria-atomic="true"
+        aria-live="assertive"
+        class="ds-c-inline-error"
+        id="static-id__error"
+      />
```
pwolfert, line 34: I see what you mean

packages/design-system/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -26,7 +26,7 @@ type Story = StoryObj<typeof Dropdown>;
 
 const dropdownOptions = [
   { label: '- Select an option -', value: '' },
-  { label: 'Confederated Tribes and Bands of the Yakama Nation', value: '1' },
+  { label: 'Confederated Tribes and Bands of the Yakama Nation', value: '1', key: 'foo' },
```
pwolfert, line 29: Why is this needed?

packages/design-system/src/components/Dropdown/Dropdown.stories.tsx
```diff
@@ -38,6 +38,8 @@ const dropdownOptions = [
     label:
       'Nisqually Indian Tribe (formerly the Nisqually Indian Tribe of the Nisqually Reservation)',
     value: '6',
+    isDisabled: true,
+    disabled: true,
```
pwolfert, line 42: Oh, it looks like this might have been from debugging that other issue

packages/design-system/src/components/Dropdown/Dropdown.stories.tsx
zarahzachz: er... i was just testing your PR reviewing skills lol (thanks for catching this)

tests/browser/snapshots/stories/components-autocomplete--custom-markup-healthcare--Mobile-Chrome.png
pwolfert, line 1: Is there a way to avoid the extra 4px gap this creates in the healthcare theme? The healthcare theme puts inline errors below the element, and it has 4px of margin above it.

4px of margin for the empty inline error div between the autocomplete field and 'clear search' button

packages/design-system/src/components/ChoiceList/Choice.tsx
zarahzachz: I think the future solve for this will be the following (in a follow-up ticket):

1. Update the Choice/ChoiceList pattern so individual Choice cannot show error messages when used in ChoiceList
2. ChoiceList passes an internal prop to Choice to hide the InlineError markup for individual Choices
3. ChoiceList throws warning if an error is detected in use of ChoiceList Choices

packages/design-system/src/components/InlineError/useInlineError.tsx
```diff
@@ -51,18 +51,18 @@ export function useInlineError<T extends UseInlineErrorProps>(props: T) {
...
 
-  const errorElement = errorMessage ? (
+  const errorElement = (
     <InlineError
       id={errorId}
       inversed={inversed}
       className={classNames(
         errorMessageClassName,
-        errorPlacement === 'bottom' && 'ds-c-inline-error--bottom'
+        errorPlacement === 'bottom' && errorMessage && 'ds-c-inline-error--bottom'
```
pwolfert, line 60: Oh smart!

pwolfert (approved):
LGTM 👍

PR #3142, opened 2024-06-28
[WNMGDS-2808] Fix Firefox scrolling pages to the top when a dialog opens

pwolfert:
## Summary

Fix Firefox scrolling pages to the top when a dialog opens in race condition.

On one of the apps, opening a dialog in Firefox would scroll the page to the top. We weren't able to reproduce this in our Storybook, but we could see the behavior that caused it by disabling our code that applied the body classes and properties that freeze it. The polyfill styles by themselves were causing the page to scroll to the top in Firefox. In the race condition, the page would scroll to the top before our JavaScript could poll the document to find its scroll position, so the scroll position was always zero.

We don't need these polyfill styles anymore. We have to keep the skeleton of a JavaScript polyfill around to support testing frameworks, but those don't need any CSS. The dialog element has Baseline 2022 support.

## How to test

We tested the fix in the downstream app that experienced the bug.

## Checklist

### DIFF
--- a/packages/design-system/src/styles/components/_Dialog.scss
+++ b/packages/design-system/src/styles/components/_Dialog.scss
@@ -1,40 +1,6 @@
 @use '../layout' as *;
 @use '../base';
 
-// Start polyfill styles
-// dialog {
-//   background-color: var(--dialog__background-color);
-//   height: fit-content;
-//   left: 0;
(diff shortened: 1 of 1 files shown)


zarahzachz (approved):
Looks good locally

PR #3145, opened 2024-07-02
[WNMGDS-2818] Prevent nested errors in choicelist

zarahzachz:
WNMGDS-2818

Prevent errors from being nested in ChoiceList.

- Individual Choice can have an `errorMessage`
- Choices nested within ChoiceList cannot have an `errorMessage`
- If an error exists in ChoiceList, it should only be revealed by the ChoiceList's `errorMessage`, not on its children

Also added TS checks to make sure `error*` props weren't being used and a `console.warn` to gently scold if they weren't TS users and insisted on doing the bad pattern.

### DIFF
--- a/packages/design-system/src/components/ChoiceList/Choice.tsx
+++ b/packages/design-system/src/components/ChoiceList/Choice.tsx
@@ -45,6 +45,10 @@ export interface BaseChoiceProps {
    */
   className?: string;
   /**
+   * @hide-prop Internal prop used to determine if a Choice is the child of a ChoiceList. Used to hide excessive error messages.
+   */
+  _choiceListChild?: boolean;
+  /**
    * Additional classes to be added to the `input` element.
    */
   inputClassName?: string;
@@ -112,7 +116,7 @@ export const Choice = (props: ChoiceProps) => {
 
   let errorId;
   let errorElement;
-  if (props.errorMessage) {
+  if (!props._choiceListChild) {
     errorId = props.errorId ?? `${props.id}__error`;
     errorElement = (
       <InlineError id={errorId} inversed={props.inversed} className={props.errorMessageClassName}>
@@ -196,6 +200,10 @@ export const Choice = (props: ChoiceProps) => {
       {checked ? checkedChildren : uncheckedChildren}
     </div>
   );
+};
+
+Choice.defaultProps = {
+  _choiceListChild: false,
 };
 
 export default Choice;
--- a/packages/design-system/src/components/ChoiceList/Choice.tsx
+++ b/packages/design-system/src/components/ChoiceList/Choice.tsx
@@ -123,6 +123,8 @@ export const Choice = (props: ChoiceProps) => {
         {props.errorMessage}
       </InlineError>
     );
+  } else {
+    delete props._choiceChild;
   }
 
   // Subscribe to changes from other radio buttons in the same group
--- a/packages/design-system/src/components/ChoiceList/Choice.tsx
+++ b/packages/design-system/src/components/ChoiceList/Choice.tsx
@@ -124,7 +124,8 @@ export const Choice = (props: ChoiceProps) => {
       </InlineError>
     );
   } else {
-    delete props._choiceChild;
+    const { _choiceChild, ...rest } = props;
+    props = rest;
   }
 
   // Subscribe to changes from other radio buttons in the same group
--- a/packages/design-system/src/components/ChoiceList/Choice.tsx
+++ b/packages/design-system/src/components/ChoiceList/Choice.tsx
@@ -45,9 +45,9 @@ export interface BaseChoiceProps {
    */
   className?: string;
   /**
-   * @hide-prop Internal prop used to determine if a Choice is the child of a ChoiceList. Used to hide excessive error messages.
+   * @hide-prop Internal prop used to determine if a Choice is the child of a another component (like ChoiceList or MonthPicker). Used to hide excessive error messages.
    */
-  _choiceListChild?: boolean;
+  _choiceChild?: boolean;
   /**
    * Additional classes to be added to the `input` element.
    */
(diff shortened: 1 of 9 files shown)


packages/design-system/src/components/ChoiceList/Choice.tsx
```diff
@@ -112,13 +116,16 @@ export const Choice = (props: ChoiceProps) => {
...
   let errorElement;
-  if (props.errorMessage) {
+  if (!props._choiceChild) {
     errorId = props.errorId ?? `${props.id}__error`;
     errorElement = (
       <InlineError id={errorId} inversed={props.inversed} className={props.errorMessageClassName}>
         {props.errorMessage}
       </InlineError>
     );
+  } else {
```
zarahzachz, line 126: This condition needed to be added to strip `_choiceChild` from rendered markup (failed snapshot tests).

packages/design-system/src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -19,7 +19,10 @@ export interface BaseChoiceListProps {
   /**
    * Array of objects representing the props for each Choice in the ChoiceList
    */
-  choices: Omit<ChoiceProps, 'name' | 'type'>[];
+  choices: Omit<
+    ChoiceProps,
+    'name' | 'type' | 'errorMessage' | 'errorId' | 'errorMessageClassName'
+  >[];
```
pwolfert, line 25: 👍

packages/design-system/src/components/web-components/ds-choice/__snapshots__/ds-choice.test.tsx.snap
```diff
@@ -132,6 +139,12 @@ exports[`Choice is a checkbox 1`] = `
         >
           George Washington
         </label>
+        <p
+          aria-atomic="true"
+          aria-live="assertive"
+          class="ds-c-inline-error"
+          id="undefined__error"
```
pwolfert, line 146: This is going to be duplicated in the DOM pretty quickly, so I don't think we can ship it in this state. Do you know why this is undefined?

pwolfert: Looks like it's actually happening on the React Choice in Storybook too

packages/design-system/src/components/ChoiceList/Choice.tsx
```diff
@@ -112,13 +116,16 @@ export const Choice = (props: ChoiceProps) => {
...
+    const { _choiceChild, ...rest } = props;
+    props = rest;
```
pwolfert, line 128: What's the advantage of this over `delete props._choiceChild`? It's still mutating the pointer `props` even if it doesn't mutate the object. I think delete is a little clearer

packages/design-system/src/components/ChoiceList/Choice.tsx
```diff
@@ -112,13 +116,16 @@ export const Choice = (props: ChoiceProps) => {
 
   let errorId;
   let errorElement;
-  if (props.errorMessage) {
+  if (!props._choiceChild) {
     errorId = props.errorId ?? `${props.id}__error`;
```
pwolfert, line 120: I think this assignment statement being in a different kind of condition now might be one clue about what's happening

packages/design-system/src/components/web-components/ds-choice/__snapshots__/ds-choice.test.tsx.snap
zarahzachz: This is some thing I was seeing in PROD right now. I think you can see it in the web component side of things.

packages/design-system/src/components/ChoiceList/Choice.tsx
zarahzachz: Using delete caused tests for MonthPicker and ChoiceList to fail. I can't remember what the error was, I think it had to do with a type error where you couldn't delete something from an object.

pwolfert: Huh, it's kind of odd that TypeScript would be mad about that but fine with reassigning `rest` to `props` even though it's a slightly different type. If we moved it up to the function signature like

```ts
export const Choice = ({ _choiceChild, ...props }: ChoiceProps) => {
```

then we wouldn't need the `else` block at all.

packages/design-system/src/components/web-components/ds-choice/__snapshots__/ds-choice.test.tsx.snap
pwolfert: Okay, I see it in prod now

pwolfert: [Created a fix for it](https://github.com/CMSgov/design-system/pull/3146)

pwolfert (approved):
LGTM 👍

packages/design-system/src/components/ChoiceList/Choice.tsx
```diff
@@ -44,6 +44,10 @@ export interface BaseChoiceProps {
    * Additional classes to be added to the root `div` element.
    */
   className?: string;
+  /**
+   * @hide-prop Internal prop used to determine if a Choice is the child of a another component (like ChoiceList or MonthPicker). Used to hide excessive error messages.
+   */
+  _choiceChild?: boolean;
```
pwolfert, line 50: Right now the doc snapshots are failing, and I don't think we have a working method of hiding any props from the arg tables. We can do it for web components where we created a custom table component, but I don't know how to hide it from the arg tables in the storybook one. The following will only hide it from the storybook controls but not the doc page:

```js
argTypes: {
  _choiceChild: {
    // Until this pattern has solidified, we're not going to advertize this feature.
    table: {
      disable: true,
    },
  },
}
```

PR #3153, opened 2024-07-10
[WNMGDS-2645] Remove unused storybook styles, add support for font fallbacks

zarahzachz:
WNMGDS-2645

In PROD, discovered the Nunito font was being downloaded for Storybook UI. Discovered the cmsTheme file we created (I believe when Storybook was first added to this project) wasn't hooked up to style the canvas area of Storybook's UI (the area where the interactive examples were). When adding the cmsTheme to Storybook's canvas area, Nunito was removed, but other styles were introduced. Turns out we had things configured that weren't actually displaying.

Discussed this with Patrick and decided that because we weren't seeing the pre-set configuration to begin with, it's best to remove those options and just configure the things we care about (brand color, names, fonts). So I simplified this config. The only change should be the font-face used on the canvas documentation.

### DIFF
--- a/.storybook/cmsTheme.ts
+++ b/.storybook/cmsTheme.ts
@@ -3,33 +3,11 @@ import { create } from '@storybook/theming';
 export default create({
   base: 'light',
 
-  colorPrimary: '#112e51',
   colorSecondary: '#0071bc',
 
-  // UI
-  appBg: 'white',
-  appContentBg: 'white',
-  appBorderColor: 'grey',
-  appBorderRadius: 4,
-
-  // Typography
-  fontBase: '"Open Sans", sans-serif',
-  fontCode: 'monospace',
-
-  // Text colors
-  textColor: '#212121',
-  textInverseColor: '#fff',
-
-  // Toolbar default and active colors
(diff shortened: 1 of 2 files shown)


.storybook/cmsTheme.ts
```diff
@@ -3,33 +3,11 @@ import { create } from '@storybook/theming';
 export default create({
   base: 'light',
 
-  colorPrimary: '#112e51',
   colorSecondary: '#0071bc',
```
pwolfert, line 6: What would happen if you make this string `var(--color-primary)` or something?

pwolfert (approved):
LGTM 👍 

My comment is just something I'm curious about

.storybook/cmsTheme.ts
zarahzachz: Storybook errors: 

> Uncaught Error: Couldn't parse the color string. Please provide the color as a string in hex, rgb, rgba, hsl or hsla notation.

pwolfert: Ah okay

PR #3157, opened 2024-07-12
[WNMGDS-2807] Fix medicare component overrides and make handling of side effects consistent

pwolfert:
## Summary

[WNMGDS-2807](https://jira.cms.gov/browse/WNMGDS-2807)

While we were trying to make component overrides and side-effects handling consistent between `ds-healthcare-gov` and `ds-medicare-gov`, we discovered that the three component overrides for medicare (`HelpDrawer`, `HelpDrawerToggle`, and `ThirdPartyExternalLink`) don't actually work, because they rely on side effects that aren't flagged for bundlers. In `ds-healthcare-gov`, we were marking the whole package as having side effects, but in `ds-medicare-gov`, we were only marking the web components as having side effects. This resulted in lost opportunity for tree-shaking on the healthcare side and a bug on the medicare side. Both packages have been updated to explicitly list the component-override modules (among others) as having side effects.

### Testing `ds-healthcare-gov`
1. Check out 8d70996240a90b24bf79a0a1977e7f8ddeb351a0
2. Run `yarn build:healthcare`
3. Go to the `examples/react-app` project and build and run it.
4. You should be able to see the healthcare-specific Accordion designs and an external link modal with a healthcare-specific "learn more" link. You should also see the text field's error message on the bottom. 

### Testing `ds-medicare-gov`
1. Check out bafba45c6e24436b4e79ae2dac12b7cf73b7ba05
2. Run `yarn build:medicare`
3. Go to the `examples/react-app` project and build and run it.
4. You should be able to see the medicare-specific HelpDrawer designs in the toggle and the drawer itself and an external link modal with a medicare-specific "learn more" link. The text field doesn't matter, but I forgot to remove it.

## Checklist

### DIFF
--- a/packages/ds-healthcare-gov/src/components/Accordion/index.tsx
+++ b/packages/ds-healthcare-gov/src/components/Accordion/index.tsx
@@ -1,1 +1,1 @@
-export * from './AccordionItem';
+import './AccordionItem';
--- a/packages/ds-healthcare-gov/src/components/i18n.ts
+++ b/packages/ds-healthcare-gov/src/components/i18n.ts
@@ -1,11 +1,11 @@
 /**
- * @file Adds translations to core translation module and exports the i18n-
- * related functions. We want components in this child design system that use
- * translation to import their translation functions from this module so we
- * get the side-effects of adding our package-specific translations. It's not
- * the cleanest way to do things, probably, but it's compatible with unit
- * tests and Storybook, which expect each component to get everything it
- * needs from its imports.
+ * @file Adds translations to core translation module and exports the i18n-related
+ * functions. We want components in this child design system that usetranslation to
+ * import their translation functions from this module so weget the side-effects of
+ * adding our package-specific translations. Note that even though the components in this
+ * child design system import their i18n functions directly from this module, we still
+ * need to declare that this module has side effects in our package.json in order to keep
+ * compilers from tree-shaking out the calls to `addTranslations`.
  */
 import en from './locale/en.json';
 import es from './locale/es.json';
--- a/packages/ds-healthcare-gov/src/components/i18n.ts
+++ b/packages/ds-healthcare-gov/src/components/i18n.ts
@@ -1,7 +1,7 @@
 /**
  * @file Adds translations to core translation module and exports the i18n-related
- * functions. We want components in this child design system that usetranslation to
- * import their translation functions from this module so weget the side-effects of
+ * functions. We want components in this child design system that use translation to
+ * import their translation functions from this module so we get the side-effects of
  * adding our package-specific translations. Note that even though the components in this
  * child design system import their i18n functions directly from this module, we still
  * need to declare that this module has side effects in our package.json in order to keep
--- a/packages/ds-healthcare-gov/src/components/index.ts
+++ b/packages/ds-healthcare-gov/src/components/index.ts
@@ -1,5 +1,3 @@
-'use strict';
-
 /**
  * This is the main entry file for a child design system's React components. It should
  * contain all exported JS from the core CMS design system and all additional child
@@ -18,5 +16,5 @@ export * from './Logo';
 // file's `sideEffects` property in order for build systems to know not to tree-shake
 // these modules out of bundles.
 export * from './config';
-import './ThirdPartyExternalLink';
 import './Accordion';
+import './ThirdPartyExternalLink';
--- a/packages/ds-healthcare-gov/src/components/index.ts
+++ b/packages/ds-healthcare-gov/src/components/index.ts
@@ -1,21 +1,22 @@
 'use strict';
 
 /**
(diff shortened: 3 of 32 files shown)


packages/ds-healthcare-gov/src/components/Accordion/index.tsx
```diff
@@ -1 +1 @@
-export * from './AccordionItem';
+import './AccordionItem';
```
pwolfert, line 1: Using "import" without any specific things we're trying to import should make it more clear that we need this for its side effects and that the magic isn't in the exports.

tamaradement: I think I understand now. By using an import for `AccordionItem`, we emphasize that the primary goal of importing this component here is to apply the healthcare modifications globally via side effects, rather than to reuse it elsewhere in the `ds-healthcare-gov` package.

pwolfert: Yeah, exactly

JackRyan1989 (commented):
So testing on your two commit branches works fine, the React App looks as one would expect. However, I noticed that the Storybook components for the Healthcare DS are not rendering. Most have the same error, though there seems to be some variance. This is happening on both of your commit branches and your parent branch (pwolfert/debugging-side-effects). 
healthcare_accordion_storybook

pwolfert:
> However, I noticed that the Storybook components for the Healthcare DS are not rendering

Hmm, I've updated the stories locally, but it looks like the side effects aren't playing well with Storybook. I'll need to play around with this some more.

tamaradement:
> So testing on your two commit branches works fine, the React App looks as one would expect. However, I noticed that the Storybook components for the Healthcare DS are not rendering. Most have the same error, though there seems to be some variance. This is happening on both of your commit branches and your parent branch (pwolfert/debugging-side-effects). 

On my end, I'm not seeing this issue with storybook on the main branch (pwolfert/debugging-side-effects) or on either of the two commit branches

pwolfert:
I'm not sure why it worked for you but not for us, @tamaradement. It was indeed failing for me as well, so I've updated this branch with a fix.

JackRyan1989:
@pwolfert The accordion looks good in the Healthcare section. We're getting errors for a number of translation keys though:
Translation key 'footer.disclaimer' does not resolve to a string.
Translation key 'footer.privacySettings' does not resolve to a string.
Translation key 'header.myApplicationsAndCoverage' does not resolve to a string.

pwolfert:
Ugh, I guess we even have to declare the side effects inside the i18n module where we already explicitly import it before using the translations. I thought that surely WebPack wouldn't tree-shake out those `addTranslation` calls, but it was. It was even happening in `examples/react-app`. You can test with this commit: [997602b](https://github.com/CMSgov/design-system/pull/3157/commits/997602b29d18a801778c4b73e906d0d46d715b50)

JackRyan1989:
> Ugh, I guess we even have to declare the side effects inside the i18n module where we already explicitly import it before using the translations. I thought that surely WebPack wouldn't tree-shake out those `addTranslation` calls, but it was. It was even happening in `examples/react-app`. You can test with this commit: [997602b](https://github.com/CMSgov/design-system/pull/3157/commits/997602b29d18a801778c4b73e906d0d46d715b50)

Looks good now!

packages/ds-healthcare-gov/src/components/index.ts
```diff
@@ -1,21 +1,20 @@
...
+// Export new components that are specific to this child design system
 export * from './Footer';
 export * from './Header';
 export * from './Logo';
-export * from './ThirdPartyExternalLink';
+
+// The following modules have side effects and must be accounted for in the package
+// file's `sideEffects` property in order for build systems to know not to tree-shake
+// these modules out of bundles.
 export * from './config';
```
kniedermaier, line 18: @pwolfert What's your rationale for keeping this an export vs. importing it like the two modules below?

pwolfert: This one is kind of a hybrid where there's a side effect but also new things to export (`headerSendsAnalytics` and `setHeaderSendsAnalytics`, which should actually be deprecated).

zarahzachz (approved):

packages/ds-healthcare-gov/src/components/i18n.ts
```diff
@@ -1,7 +1,7 @@
 /**
  * @file Adds translations to core translation module and exports the i18n-related
- * functions. We want components in this child design system that usetranslation to
- * import their translation functions from this module so weget the side-effects of
+ * functions. We want components in this child design system that use translation to
+ * import their translation functions from this module so we get the side-effects of
```
pwolfert, line 4: Doh!

PR #3158, opened 2024-07-15
[WNMGDS-2825] Fix release tagging

pwolfert:
## Summary

WNMGDS-2825

- The release script no longer uses Lerna. Instead it manually creates the publish commit and tags.
- The release script no longer asks for version changes for individual packages. Instead, you just select the "release type" from a list, and it applies the change to all packages appropriately.
- Fixed the `--undo` sub-command because we only try to delete tags that we created (because we're in charge of that now).

## How to test

1. Check out the branch
2. Create your own test branch off of it like `git branch -b "your-test-branch-name"` so it doesn't push changes to origin and cause the Jenkins to do extra work with CI tests
3. Run the release script. Check the changes it made for correctness
4. Undo the release with `yarn release --undo`
5. Delete your test branch

## Checklist


zarahzachz:
Do we still need Lerna now that the script's been updated?

pwolfert:
About the GPG issues, maybe we can troubleshoot in our engineering meeting tomorrow, since others were also experiencing this error. I'm wondering if there's a setting in `git` that needs to be set, since I know you have GPG.

package.json
```diff
@@ -137,7 +137,6 @@
     "humanize-react": "^1.0.0",
     "husky": "^7.0.0",
     "jest": "^27.5.1",
-    "lerna": "^5.6.2",
```
zarahzachz, line 140: 🤌 that's a nice deletion!

zarahzachz (approved):

kim-cmsds (changes requested):
While testing locally, I forgot to comment out the two `git push` commands in the release script, which means we'll want to remove this commit (4be137c499f8bfa35db5d4e74a29bd68b3d67334) before merging. Other than that, this looks great.

pwolfert:
Hey, @kim-cmsds, you should be able to run `yarn release --undo` to remove it and its branch

kim-cmsds (approved):
Thank you for the assist, @pwolfert! Now that tagging and the undoing of a release command have been thoroughly tested, this looks good to me!

PR #3166, opened 2024-07-22
[NO-TICKET] Use new docgen union sorting

pwolfert:
## Summary

- Uses the new react-docgen-typescript feature [referenced here](https://github.com/CMSgov/design-system/pull/3003#issuecomment-2239465345)
- Is a follow-up to https://github.com/CMSgov/design-system/pull/3080

## How to test

2. `yarn test:browser:storybook-docs` and make sure it doesn't try to change the order of union-type members from what the snapshots show now (string-stort (somewhat alphabetical) order)

### DIFF
--- a/.storybook/main.ts
+++ b/.storybook/main.ts
@@ -1,5 +1,6 @@
 import { dirname, join } from 'path';
 import type { StorybookConfig } from '@storybook/react-webpack5';
+import { typescript as typescriptPresets } from '@storybook/core-server/dist/presets/common-preset';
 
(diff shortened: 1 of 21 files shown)


zarahzachz (approved):
everything passed locally

pwolfert:
Going to just merge this for now since it works. We can follow up with a dependency version bump when the library does a real release and maybe even after the `@storybook/react-docgen-typescript-plugin` library bumps the version that it uses so we don't have to create a custom resolution in our package file.

PR #3169, opened 2024-07-23
[WNMGDS-2709] Implement open graph tags

zarahzachz:
WNMGDS-2709

Added OpenGraph tags to the docs site. If the page is classified as "not in sidebar," I've taken that to mean the page link title and description should default to generic "website" text (like the home or contact pages). If the page is in the sidebar, it's a real page with usable frontmatter (like components and blog posts).

This OG work is meant to be a first draft and may need further adjustments in the future, we're just proving out what this could look like on our docs site. What does this mean? Well the generic copy I use for the "website" pages (and my definition of "website" pages for that matter) could use more input from the team and/or Patty (or whoever does copywriting). Additionally. we don't have an image to use on [the required `og:image`  tag](https://ogp.me/#metadata) (the DS doesn't have a proper logo or any associated imagery, so what would this even be? To be decided by the team and/or Melissa?).

[Demo url](https://cmsgov.github.io/design-system/branch/WNMGDS-2709/implement-open-graph-tags)


packages/docs/src/components/layout/Layout.tsx
```diff
@@ -80,6 +80,23 @@ const Layout = ({
           lang: 'en',
         }}
       >
+        <meta
+          property="og:title"
+          content={slug?.includes('not-in-sidebar') ? baseTitle : frontmatter?.title}
```
pwolfert, line 85: Should the title also include the context of the CMS design system like the actual tab title does? When I go to a page like Alert, the title is just `Alert` for this meta tag, but the tab itself gives some better context with `Alert - CMS Design System`.

pwolfert (approved):
Overall LGTM. Just one suggestion below

packages/docs/src/components/layout/Layout.tsx
zarahzachz: oh that's a good idea!

PR #3053, opened 2024-04-23
[WNMGDS-2759] Fix nesting web components in other web components and re-renders

pwolfert:
## Summary

- Fixes nesting web components in other web components.
- Fixes re-rendering of web components (making changes that cause the underyling Preact component to re-render) causing a nesting effect (button in a button, etc.).
- Adds a new feature where making changes to a web component instance's inner HTML will actually trigger an update and re-render of the Preact component (see updated Astro example).
- Ports some unit tests from `preactement`, modifying them to fit our new way of doing things.

The new rendering method adds some complexity, but it also allows us to remove some complexity elsewhere in the custom element code. Special thanks to @zarahzachz for her idea to use a `<template>` component instead of a `<div>`!

## How to test

1. Try out all the web component stories.
2. Try out the web component astro example. Maybe even make random DOM changes to web components in the console.
3. Run the web-component unit tests.

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] Selected appropriate release milestone

### If this is a change to design:

- [x] If visual regression image references have been changed, design **MUST** be assigned to review. In this instance, **designer approval is a requirement** before the PR can be merged.

### If this is a change to code:

- [x] Created or updated unit tests to cover any new or modified code
- [x] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:update`)

PR #3172, opened 2024-07-25
[WNMGDS-2843] Add `ds-text-field` component

pwolfert:
## Summary

[WNMGDS-2843](https://jira.cms.gov/browse/WNMGDS-2843)

Adds a new `ds-text-field` web component.

## How to test

1. Try it out in storybook
2. If you really want to, you can try adding it to the astro example too, but I did that myself already and removed it because I didn't have a good use case for it in the example, but maybe that's lazy and I should actually just revamp the example page and make a more meaningful form example that includes more components.

## Checklist


packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+    },
+    'label-mask': {
+      description:
+        'Applies date format masking to the input value entered and renders to a text field above the input. Passing `true` to `valueOnly` will return just the formatted value entered.',
+      options: [undefined, 'phone', 'zip', 'ssn', 'currency'],
+      control: { type: 'radio' },
+    },
+    name: {
+      description:
+        'The `name` is applied to a hidden select element that holds the selected value for the purpose of native HTML form support',
```
jack-ryan-nava-pbc, line 71: The name value gets set both on the parent custom element and the child input tag.

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+    },
+    'requirement-label': { control: 'text' },
+    size: {
+      description: 'Sets the max-width of the input either to `"small"` or `"medium"`',
+      options: [undefined, 'medium', 'small'],
+      control: { type: 'radio' },
+    },
+    value: {
+      description:
+        'Sets the selected value. Using this attribute instead of `default-value` means the dropdown will not maintain its own internal selection state.',
```
jack-ryan-nava-pbc, line 82: I don't think this description is correct. Should say input and maybe explain that setting the attribute will also set the content of the input element itself?

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+  },
+  args: {
+    label: 'Enter some text.',
+    name: 'text-field',
+  },
+  parameters: {
+    docs: {
+      page: WebComponentDocTemplate,
+      description: {
+        component: `For information about how and when to use this component, [refer to its full documentation page](https://design.cms.gov/components/alert/).`,
```
jack-ryan-nava-pbc, line 99: The link here needs to be updated to https://design.cms.gov/components/text-field or https://design.cms.gov/components/text-field/text-field/

jack-ryan-nava-pbc (changes requested):
Overall this looks great! Mostly just some discrepancies I noticed in the Story plus some questions about how some props (well, actually just one prop) are/is being passed in.

packages/design-system/src/components/web-components/ds-text-field/__snapshots__/ds-text-field.test.tsx.snap
```diff
@@ -0,0 +1,34 @@
...
+        aria-live="assertive"
+        class="ds-c-inline-error"
+        id="text-field--1__error"
+      />
+      <input
+        aria-disabled="false"
+        aria-invalid="false"
+        class="ds-c-field"
+        id="text-field--1"
+        parent="[object HTMLElement]"
```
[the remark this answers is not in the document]
pwolfert: That's something added by `preactement`. I'm not sure it's really that helpful to have there, but we never removed it

packages/design-system/src/components/web-components/ds-text-field/__snapshots__/ds-text-field.test.tsx.snap
```diff
@@ -0,0 +1,34 @@
...
+        type="text"
+      />
+    </div>
+    <template />
```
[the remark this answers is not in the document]
pwolfert: The template is actually very critical and used by JavaScript. It's something we added to Preactement [in this change](https://github.com/CMSgov/design-system/pull/3053).

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+const Template = (args) => {
+  useEffect(() => {
+    const onChange = (event) => {
+      action('ds-change')(event);
+    };
+    const onBlur = (event) => {
+      action('ds-blur')(event);
+    };
+    const dropdown = document.querySelector('ds-text-field');
```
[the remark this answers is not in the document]
pwolfert: Doh! Good catch

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.tsx
```diff
@@ -0,0 +1,67 @@
...
+  ariaDisabled?: string;
+  disabled?: string;
+  inversed?: string;
+  labelMask?: string;
+  rootId?: string;
+}
+
+const Wrapper = ({ rootId, ...otherProps }: WrapperProps) => (
+  <TextField
+    {...otherProps}
```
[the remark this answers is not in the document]
pwolfert: Are you talking about collisions with the props we defined below on lines 58-63?

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+import { action } from '@storybook/addon-actions';
+import { useEffect } from 'react';
+import { webComponentDecorator } from '../storybook';
+import './ds-text-field';
+
+export default {
+  title: 'Web Components/ds-text-field',
+  argTypes: {
+    'class-name': {
+      description: 'Additional classes to be added to the root button element.',
```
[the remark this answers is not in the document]
pwolfert: Copypasta, thanks

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
```diff
@@ -0,0 +1,126 @@
...
+      control: 'text',
+    },
+    'error-id': {
+      description:
+        'The ID of the error message applied to this field. If none is provided, the id will be derived from the `root-id` attribute.',
+      control: 'text',
+    },
+    'error-placement': {
+      description: 'Location of the error message relative to the field input',
+      options: [undefined, 'top', 'bottom'],
```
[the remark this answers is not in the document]
pwolfert: By "captured here" do you mean explaining that in the description?

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
pwolfert: That's the same as some of our other web components. Do you see that being a problem?

pwolfert: Updating the description here too

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
pwolfert: I'm actually going to get rid of "default-value" and just include `value` as an option. It works better in this component and doesn't need explaining.

pwolfert: That's possibly an improvement we can make to the `ds-dropdown` too because I doubt that both `default-value` and `value` have uniquely useful use cases

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+      description:
+        'Applies date format masking to the input value entered and renders to a text field above the input. Passing `true` to `valueOnly` will return just the formatted value entered.',
+      options: [undefined, 'phone', 'zip', 'ssn', 'currency'],
+      control: { type: 'radio' },
+    },
+    name: {
+      description: "The `input` field's `name` attribute.",
+      control: 'text',
+    },
+    'requirement-label': { control: 'text' },
```
kim-cmsds, line 68: It might be helpful to add a description for this prop

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
```diff
@@ -0,0 +1,119 @@
...
+      control: 'text',
+    },
+    'label-id': {
+      description:
+        "A unique `id` to be used on the field label. If one isn't provided, a unique ID will be generated.",
+      control: 'text',
+    },
+    'label-mask': {
+      description:
+        'Applies date format masking to the input value entered and renders to a text field above the input. Passing `true` to `valueOnly` will return just the formatted value entered.',
```
kim-cmsds, line 60: What is `valueOnly` referring to in this description?

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.test.tsx
kim-cmsds, line 1: Thoughts on adding a test for the `multiline` attribute?

kim-cmsds (approved):
Nice, this looks great to me! I left a few non-blocking questions.

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.tsx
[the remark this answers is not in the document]
jack-ryan-nava-pbc: Yes, sorry I should've used one of those props explicitly. Like if otherprops contains `autofocus` what happens? This should also be considered a nit, and a non-blocking q.

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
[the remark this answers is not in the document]
jack-ryan-nava-pbc: Yep! I do mean explaining that in the description.

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
jack-ryan-nava-pbc: I don't see it being a problem, just more flagging for the description. Pretty sure `name` only gets used on form submission, but also could be wrong.

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.tsx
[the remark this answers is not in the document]
pwolfert: In that case, it would get overwritten by our autofocus prop that we explicitly define, and that's intentional, because we want to pass a parsed boolean to the Preact component. Except in this case there's a difference between `autofocus` and `autoFocus`, isn't there? So maybe this particular prop/attribute has a bug, and we need to destructure `autocomplete` out of `otherProps.

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.test.tsx
zarahzachz: Could be worth doing a snap test at least - especially since this attr swaps a text input for a `textarea`

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.test.tsx
```diff
@@ -0,0 +1,134 @@
...
+    expect(input).toHaveAttribute('disabled');
+  });
+
+  it('does not render disabled when passed false', () => {
+    renderTextField({ disabled: 'false' });
+    const input = screen.getByRole('textbox');
+    expect(input).not.toHaveAttribute('disabled');
+  });
+
+  it('applies additional classes to the wrapper', () => {
```
zarahzachz, line 37: The following is probably more philosophical than the scope of this ticket warrants. Feel free to disregard my stream-of-conscious take here:

This test initially confused me because I associate "wrapper" with the host for 🚾, and having a special prop to apply a class to a host is unnecessary because it's HTML and `class` would work fine. 

Anyways... I had to do a double-take and read what the test was actually doing and realized "wrapper" means something different because it's querying the clearfix style in the test, which is the internal `div` within the host wrapper. Is this `div` even needed? Who clearfixes in the era of flexbox and grid?

Do we need better language when talking about these sorts of tags?

zarahzachz (approved):
Looks good, just some small comments you can disregard or accept

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.tsx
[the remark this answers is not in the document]
pwolfert: Fixed in [54c0cad](https://github.com/CMSgov/design-system/pull/3172/commits/54c0cad7cdb562f0dd88216b76b3d2a88f29b5ef)

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
pwolfert: Fixed this and the other components that were missing a description for this attribute in [5faed7f](https://github.com/CMSgov/design-system/pull/3172/commits/5faed7fc4555710991870a8436f97f7597fb2f54)

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.stories.tsx
pwolfert: Fixed in [7471e35](https://github.com/CMSgov/design-system/pull/3172/commits/7471e35d92614e90ca03b258d5d140dd21b096d7)

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.test.tsx
pwolfert: We might very well be able to remove that "wrapper" div now. I agree that it's beyond the scope here, but it's worth considering for the future (@kim-cmsds).

That's a great point about the wording here being confusing. I think I'll change it to "root element" or something like that.

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.test.tsx
pwolfert: Ah, you knew I missed that. Good catch. It doesn't even work, does it?

packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx
```diff
@@ -95,8 +95,13 @@ export default {
...
       control: 'text',
-      description:
-        'Text showing the requirement (ie. "Optional", or "Required").\nIn most cases, this should be used to indicate which fields are optional.\nSee the [form guidelines](https://design.cms.gov/patterns/Forms/forms/) for more info.',
+      description: (
+        <>
+          Text showing the requirement (ie. &quot;Optional&quot;, or &quot;Required&quot;).\nIn most
+          cases, this should be used to indicate which fields are optional.\nSee the{' '}
+          <a href="https://design.cms.gov/patterns/Forms/forms/">form guidelines</a> for more info.
+        </>
+      ),
```
pwolfert, line 104: Turns out this change fixes the doc pages but breaks the story pages (probably an error when rendering the Storybook controls). I think Storybook must not like rendering non-strings in the control descriptions.

jack-ryan-nava-pbc (approved):
LGTM!

packages/design-system/src/components/web-components/ds-text-field/ds-text-field.test.tsx
pwolfert: Fixed in [723ccf2](https://github.com/CMSgov/design-system/pull/3172/commits/723ccf2cab83e53992a465fd3a8f2e246a6bfdc1)

PR #3176, opened 2024-07-29
[NO-TICKET] Fix Gatsby static image paths for archives and demos

pwolfert:
## Summary

Fix gatsby static images so they include the prefix path.

Previously if we built the doc site with a prefix path (like https://design.cms.gov/v/9.0.0/blog/release-7.0 having a prefix path of `/v/9.0.0`), the img src attribute would omit this prefix path and just assume the image was at `/static/whatever-the-filename-is`. Overriding this element ensures we include the prefix path always.

If you go look at https://design.cms.gov/v/9.0.0/blog/release-7.0, you'll notice that the image is being pulled from `/static`, which is actually from the root website, not this archived website. Unfortunately this bug will exist in all the old archived copies.

This also fixes the demo branches, which is why I updated the `deploy-demo` script in this commit.

## How to test

Take a look at [this demo](https://cmsgov.github.io/design-system/branch/pwolfert/release-11-blog-post/blog/release-11.0/?theme=core), which now has a working image with the correct prefix path. Previously this blog-post image failed to load.

## Checklist


scripts/deploy-demo.ts
```diff
@@ -0,0 +1,20 @@
...
+  execSync(command, { stdio: 'inherit' });
+}
+
+// Get the name of the branch we're on
+const branch = sh('git rev-parse --abbrev-ref HEAD');
+const urlPath = `/design-system/branch/${branch}`;
+
+// Build the demo site
+shVerbose(`yarn --cwd ${join('packages', 'docs')} clean`);
+shVerbose(`PATH_PREFIX='${urlPath}' PREFIX_PATHS=true yarn build:docs`);
```
jack-ryan-nava-pbc, line 15: Just so I understand what's going on here:
1. We create a urlPath var that contains our branch name
2. We manually override our PATH_PREFIX env var before building our docs. 

Is that correct?

jack-ryan-nava-pbc (approved):
LGTM! Just some nits and small, non-blocking qs.

scripts/deploy-demo.ts
pwolfert: Yep, that's correct

PR #3182, opened 2024-07-31
[WNMGDS-2848] Add ds-date-field web component

tamara-corbalt:
## Summary

- Adds a new `ds-date-field` web component. See Jira ticket for details: https://jira.cms.gov/browse/WNMGDS-2848

## How to test

- Check it out on Storybook: `yarn run storybook`

## Checklist

- [X] If visual regression image references have been changed, design **MUST** be assigned to review. In this instance, **designer approval is a requirement** before the PR can be merged.

- [X] Created or updated unit tests to cover any new or modified code
- [X] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:update`)

- [X] Checked for spelling and grammatical errors


packages/design-system/src/components/web-components/index.ts
```diff
@@ -22,3 +22,4 @@ import './ds-month-picker';
 import './ds-skip-nav';
 import './ds-spinner';
 import './ds-usa-banner';
+import './ds-date-field';
```
pwolfert, line 25: Nit: We've been putting these imports in alphabetical order

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
```diff
@@ -0,0 +1,66 @@
...
+    const element = document.querySelector('ds-date-field');
+    if (element) {
+      const handleStoryBookChange = (event: CustomEvent) => {
+        action('ds-change')(event);
+        updateDate(event.detail.value);
+      };
+      element.addEventListener('ds-change', handleStoryBookChange as EventListener);
+      return () => {
+        element.removeEventListener('ds-change', handleStoryBookChange as EventListener);
+      };
```
pwolfert, line 42: Can you also add an action log for the `ds-blur` event?

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.tsx
```diff
@@ -0,0 +1,95 @@
...
+        [K in (typeof attributes)[number]]?: string;
+      };
+    }
+  }
+}
+/* eslint-enable */
+
+define('ds-date-field', () => Wrapper, {
+  attributes,
+  events: ['onChange', 'onBlur'],
```
pwolfert, line 94: Unfortunately there's no real information being passed to the `ds-change` event here because the `onChange` handler for `SingleInputDateField` doesn't actually get an event like many other `onChange` handlers do in the design system. [What actually gets passed](https://github.com/CMSgov/design-system/blob/tamara/add-ds-date-field-web-component/packages/design-system/src/components/DateField/SingleInputDateField.tsx#L42) are two strings. The default event translator in our `define` function expects an event. That means you'll either need to create a custom handler here [like we do for ds-pagination](https://github.com/CMSgov/design-system/blob/tamara/add-ds-date-field-web-component/packages/design-system/src/components/web-components/ds-pagination/ds-pagination.tsx#L75), or the `SingleInputDateField` API needs to change.

This lack of information is what's causing the following buggy behavior in the Storybook story:

Choosing a date from the calendar fails to update the input

pwolfert: Even though most of the other web component event-handler unit tests aren't this thorough, we may want to consider checking what's passed to the event handler in the unit tests because it's so different from the others. It looks like we do check for the actual information on `detail` in the `ds-pagination` unit tests.

pwolfert (changes requested):
It's almost there! Noticed some issues around events

jack-ryan-nava-pbc:
@tamara-corbalt It looks like the uncontrolled version of the React component is not functioning right because it is expecting an `onChange` function to be passed in. 

If you observe the console while trying to make a change, you'll see a big red error blob of text.

tamara-corbalt:
@jack-ryan-nava-pbc and & @pwolfert, I pushed up changes that should resolve the issues you called out. When time allows, let me know what you think!

pwolfert:
> If you observe the console while trying to make a change, you'll see a big red error blob of text.

@jack-ryan-nava-pbc, I realized after reading this that my PR never made `onChange` optional. But I don't actually know where you were seeing these errors. Can you list your reproduction steps?

tamara-corbalt:
Hi @pwolfert! @jack-ryan-nava-pbc is out sick today. Here's how you should be able to reproduce the behavior Jack was referring to [here](https://github.com/CMSgov/design-system/pull/3182#issuecomment-2269913134):
1. Go to this commit: [Adds new storybook-doc snapshots]([Adds missing storybook-doc snapshots](https://github.com/CMSgov/design-system/pull/3182/commits/eccde0b1dbd1de0d62c93aceb551d2552f8e7f73)).
2. Run storybook locally 
3. Navigate to the "Uncontrolled Component" story for the `SingleInputDateField`.
4. Open the browser console.
5. Choose a date on the datepicker.

You should see the type error in the console as described.

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    requirementLabel: { control: 'text' },
+  },
+};
+export default meta;
+
+const Template = (args) => {
+  useEffect(() => {
+    const element = document.querySelector('ds-date-field');
+    if (element) {
+      const handleStoryBookChange = (event: CustomEvent) => {
```
kim-cmsds, line 33: We know what our custom event's `detail` type will be, so we can pass that into the `CustomEvent` interface, like so:

```suggestion
      const handleStoryBookChange = (event: CustomEvent<{ updatedValue: string; formattedValue: string }> ) => {
```

Two additional resources:
- an [example](https://stackoverflow.com/a/62568282)
- link to the [typescript definition](https://github.com/microsoft/TypeScript-DOM-lib-generator/blob/29de1d981d83c2e99f6995c4e5fae934cd7d709a/baselines/audioworklet.generated.d.ts#L306).

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+        action('ds-change')(event);
+      };
+      const handleStoryBookBlur = (event: CustomEvent) => {
```
kim-cmsds, line 36: It doesn't look like `onBlur` is a custom event, in which case `Event` will do
```suggestion
      const handleStoryBookBlur = (event: Event) => {
```

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+  },
+  args: {
+    hint: 'If you were born on a leap day, entering the date will either crash our servers or open a portal to an alternate dimension.',
+    label: 'Enter your date of birth.',
+    name: 'ds-date-field',
+  },
+  argTypes: {
+    errorMessage: { control: 'text' },
+    hint: { control: 'text' },
+    requirementLabel: { control: 'text' },
```
kim-cmsds, line 24: I'm not sure that `requirementLabel` is wired up correctly. I was not able to see the label change after updating this line from `requirementLabel` to `requirement-label` and then in storybook include a value for that prop.

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+      page: WebComponentDocTemplate,
+    },
+  },
+  args: {
+    hint: 'If you were born on a leap day, entering the date will either crash our servers or open a portal to an alternate dimension.',
+    label: 'Enter your date of birth.',
+    name: 'ds-date-field',
+  },
+  argTypes: {
+    errorMessage: { control: 'text' },
```
kim-cmsds, line 22:
```suggestion
    'error-message': { control: 'text' },
```

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
kim-cmsds: Looks like adding `'requirement-label'` to the array of attributes found in `packages/design-system/src/components/web-components/ds-date-field/ds-date-field.tsx` is needed

kim-cmsds (commented):
This is looking great! I left a few more comments, let me know if you have any questions!

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+    docs: {
+      page: WebComponentDocTemplate,
+    },
+  },
+  args: {
+    hint: 'If you were born on a leap day, entering the date will either crash our servers or open a portal to an alternate dimension.',
+    label: 'Enter your date of birth.',
+    name: 'ds-date-field',
+  },
+  argTypes: {
```
pwolfert, line 21: Were there a bunch fo arg types defined that got removed? Or do the arg types still need to be filled out here? Remember that these are basically the documentation for the web component attributes. I also noticed this odd little guy (screenshot), which is probably is just due to missing argy type definitions:

A blank text field for the 'to-date' control

I didn't set this myself; it was just a blank text field when we'd expect a date string.

packages/design-system/src/components/DateField/SingleInputDateField.stories.tsx
```diff
@@ -28,7 +28,14 @@ type Story = StoryObj<typeof SingleInputDateField>;
...
   render: function Component(args) {
-    return <SingleInputDateField {...args} />;
+    const [dateString, updateDate] = useState('');
+
+    const onChange = (...params) => {
+      action('onChange')(...params);
+      updateDate(params[0]);
+    };
+
+    return <SingleInputDateField {...args} onChange={onChange} defaultValue={dateString ?? ''} />;
```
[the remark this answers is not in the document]
tamara-corbalt: Ooh right, got it. Updated the logic in `SingleInputDateField` to make the `onChange` prop optional.

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
tamara-corbalt: Oh! @pwolfert I missed this comment before I re-requested review. I'll take a look at this now.

tamara-corbalt: @pwolfert, just filled out the `argTypes` to reflect the attributes used in `ds-date-field` 👍  So should be ready for another look, when time allows

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.tsx
```diff
@@ -0,0 +1,100 @@
...
+  'from-date',
+  'from-month',
+  'from-year',
+  'to-date',
+  'to-month',
+  'to-year',
+  'hint',
+  'label',
+  'requirement-label',
+  'error-message',
```
pwolfert, line 21: I was reviewing your `argTypes` additions, and I realized that it's missing the standard set of form-field attributes [like we include here](https://github.com/CMSgov/design-system/blob/tamara/add-ds-date-field-web-component/packages/design-system/src/components/web-components/ds-text-field/ds-text-field.tsx#L25). It's a little helper we added to reduce duplication. We don't actually have the same sort of thing for the `argTypes` definitions, though; we've just been copying and pasting those from other web components and changing the descriptions if needed. You'll see in that `formAttrs` array that there are a few attributes missing from this list.

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
pwolfert: These look mostly right. The only problem I'm seeing with the ones you added is the year being a number. Because it's specified as a string in the story `args`, it ends up being blank in the controls even though there's a value. Date strings seem to be translated fine in Storybook, but for some reason Storybook doesn't seem to do the string-to-number conversion very well.

pwolfert: Simplest solution would probably to make the arg type a string, but it may not be the best long term solution if you want your storybook users to have the number controls.

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.tsx
tamara-corbalt: @pwolfert - Thanks for the feedback. I've added formAttrs to the attribute list and copied the argType definitions for the missing form-related attributes from other web components that have them. Is there anything else I need to include in this list? I noticed that the `SingleInputDateField` controls include things like `size` and `rows`, although it's unclear to me what these do. Could you provide some guidance on whether these are necessary?

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
```diff
@@ -0,0 +1,165 @@
...
+    },
+    'default-value': {
+      description: `Sets the initial value for the component. This value will be used until the user changes the input, at which point the new value will be handled by the component's internal state.
+                    The \`onChange\` event will still be triggered with the updated value, but the component will manage the state internally.`,
+      control: 'text',
+    },
+    'default-month': {
+      description:
+        'The initial month to show in the calendar picker. If not provided, defaults to the month of the currently selected date.',
+      control: 'date',
```
kim-cmsds, line 48: For these args that have the "date" control, we'll want to convert them back to "text" in the render template. The simpler solution would be to make the control "text" for now.

tamara-corbalt: Patrick mentioned [above](https://github.com/CMSgov/design-system/pull/3182#discussion_r1706128093) that date strings seem to translate well in storybook. And the controls seem to work as expected with the picker story where we add a `to-date` arg. Is it reasonable to leave the date-related arg-types as is for now?

tamara-corbalt: @pwolfert @kim-cmsds 
I've opted to make number argTypes into strings for the time being. I've left date argTypes alone, since they appear to work in Storybook. Let me know what you think :)

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.tsx
pwolfert: The new controls look good to me. The `size` and `rows` shouldn't really be on there, but the controls on the React stories are generated mostly by Storybook, which is going off of the TypeScript types. We might be able to remove it by explicitly removing those attributes from the prop types of `SingleInputDateField` like by using `Omit` [on this line](https://github.com/CMSgov/design-system/blob/tamara/add-ds-date-field-web-component/packages/design-system/src/components/DateField/SingleInputDateField.tsx#L104)

pwolfert (approved):
LGTM! That was a lot of work, but I think that now it's everything I'd expect from the `ds-date-field` component

packages/design-system/src/components/web-components/ds-date-field/ds-date-field.stories.tsx
kim-cmsds: Gotcha, thank you for pointing out the earlier conversation—I missed that. I pulled this branch down locally and looking at the "[With Picker](http://localhost:6006/?path=/story/web-components-ds-date-field--with-picker)" story, I updated the "to-date" arg to April 15, 2023 and then after clicking on the picker, the date picker defaults to this month. My expectation is that the date picker should default to April 2023.

tamara-corbalt: Thanks for the call out @kim-cmsds. I didn't fiddle around with the control enough. After further experimentation with the to-date control, I also found it's not working correctly. To investigate, I added some logging within the template. Initially, the value of to-date is a string (2023-02-15T21:56:34.272Z), which is great. However, once a user selects a date in the control, to-date is converted into a number (1676584594272 in this case, representing Feb 16, 2023, which is what I selected).

kim-cmsds: @tamara-corbalt what do you think about changing the controls for these args to "text" for now and potentially following up this PR with another ticket to improve the UX around the storybook story?

tamara-corbalt: @kim-cmsds I didn't see this comment before making this commit, but I added a utility that converts date args into strings within the template. I'm happy to change these argTypes to 'text' if you think that's a better temporary option.

From here, I think a good follow-up would be to develop a few general utilities that can handle these conversions for our web components.

Let me know what you think of this change. Happy to switch back to 'text' if preferred.

kim-cmsds: @tamara-corbalt the utility function looks good to me!

zarahzachz (approved):

PR #3183, opened 2024-07-31
[NO-TICKET] Fix the bug caused by the first fix to Gatsby image paths

pwolfert:
## Summary

This is a follow-up to https://github.com/CMSgov/design-system/pull/3176, where I attempted to fix a bug and caused its equal and opposite bug. See added code comments for details.

## How to test

I re-deployed [this demo](https://cmsgov.github.io/design-system/branch/pwolfert/release-11-blog-post/blog/release-11.0?theme=core) with the fix. Loading that page from the link should successfully show the image, and also navigating away through the "See other releases" link and then back using the "Release 11.0" link should show the image too. Without this fix, it would have failed to load the image in that second scenario, which was the opposite of what it was before the first fix.

## Checklist

### DIFF
--- a/packages/docs/src/components/content/ContentRenderer.tsx
+++ b/packages/docs/src/components/content/ContentRenderer.tsx
@@ -63,6 +63,20 @@ const TextWithMaxWidth = (props: any, Component) => {
 };
 
 /**
+ * Hack to fix missing path prefixes on static images imported from src
+ */
+const PrefixedImg = (props) => {
+  // When navigating from another page, the Gatsby client dynamically pulls the new page
+  // information and renders with the prefix correctly, so we only want to apply the
+  // path prefix manually if it isn't already there. When we load fresh with a new HTTP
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/content/ContentRenderer.tsx
```diff
@@ -74,8 +88,7 @@ const customComponents = (theme) => ({
   ColorRamps,
   ComponentThemeOptions: (props) => <ComponentThemeOptions theme={theme} {...props} />,
   EmbeddedExample,
-  // Hack to fix missing path prefixes on static images imported from src
-  img: (props) => <img {...props} src={withPrefix(props.src)} />,
+  img: PrefixedImg,
```
jack-ryan-nava-pbc, line 91: Thank you for separating into a function!

jack-ryan-nava-pbc (approved):
LGTM! Good catch

PR #3185, opened 2024-07-31
[WNMGDS-2855] Markdown and MDX broken link script

jack-ryan-nava-pbc:
## Summary

[Ticket](https://jira.cms.gov/browse/WNMGDS-2855)

The goal of this script is to provide a relatively quick way to check for outdated links and content within the codebase. If a file/url combo is listed in the output file, the idea is that it needs to be checked because the URL did not return a 200 status code.

I've added a new script that can be run via an npm script. The goal of this script is to check all links that begin with `http` or `https` located within our .md or .mdx files throughout the codebase. I've tried to exclude known directories that we don't want to dig through, like node_modules. If there are others that should be excluded, please let me know.

I'm currently writing to a .txt file. I'm meh about this and am not sure what format might be the easiest for folks to parse and work with when reviewing content. I'm also open to discussing the formatting of the content within the file, as that might not be the easiest to parse. 

## How to test

1. Run `yarn lint:links`
2. Confirm that the content `brokenLinkReport.txt` is overwritten with a series of filenames, urls and HTTP response codes.

## Checklist

## Screenshot

Screenshot 2024-08-07 at 7 54 39 PM


brokenLinkReport.html
```diff
@@ -0,0 +1,20 @@
+<div><h4>Filepath: <a href="/Users/jackryan/Desktop/design-system/README.md" target="_blank">design-system/README.md</a></h4><div style="margin-left:.5rem"><a href="http://facebook.github.io/jest/docs/en/snapshot-testing.html" target="_blank">http://facebook.github.io/jest/docs/en/snapshot-testing.html</a> <p>Code: 404</p></div><br/></div>
```
pwolfert, line 1: Should we gitignore this file?

brokenLinkReport.html
```diff
@@ -0,0 +1,20 @@
...
+<div><h4>Filepath: <a href="/Users/jackryan/Desktop/design-system/packages/docs/content/guidelines/accessibility.mdx" target="_blank">design-system/packages/docs/content/guidelines/accessibility.mdx</a></h4><div style="margin-left:.5rem"><a href="http://www.access-board.gov/508.htm" target="_blank">http://www.access-board.gov/508.htm</a> <p>Code: 404</p></div><br/></div>
+<div><h4>Filepath: <a href="/Users/jackryan/Desktop/design-system/packages/docs/content/getting-started/for-developers.mdx" target="_blank">design-system/packages/docs/content/getting-started/for-developers.mdx</a></h4><div style="margin-left:.5rem"><a href="https://design.cms.gov/cdn/<package-name>/<version>/<resource>" target="_blank">https://design.cms.gov/cdn/<package-name>/<version>/<resource></a> <p>Code: 404</p></div><br/></div>
```
pwolfert, line 20: It works in the browser, but when I try to format this file with Prettier, the following URL breaks HTML parsing:

```
https://design.cms.gov/cdn/<package-name>/<version>/<resource>
```

This URL that it found is a special case, because it actually isn't a link at all but [a URL inside an example code block](https://github.com/CMSgov/design-system/blob/main/packages/docs/content/getting-started/for-developers.mdx?plain=1#L71).

pwolfert (commented):
The HTML is useful, and I'm actually going to put up a PR to fix the first broken link it found 🎉

brokenLinkReport.html
jack-ryan-nava-pbc: Yeah, I think for the time being that makes sense. I might add a line to the script that automatically opens this file in the browser.

brokenLinkReport.html
jack-ryan-nava-pbc: I added this file to the `.prettierignore` file in a later commit. I don't think this file should be linted.

pwolfert: Just bringing it to your attention that that URL that isn't a link is getting incorrectly picked up and that it can break HTML parsing 👍

scripts/check-markdown-links.ts
```diff
@@ -0,0 +1,108 @@
...
+};
+
+// Use node-fetch to query urls and return a status code:
+const curl = (file: string, url: string): void => {
+  const spliceFrom = file.split('/').indexOf('design-system');
+  const filePath: string = file.split('/').splice(spliceFrom).join('/');
+  fetch(url).then((res) => {
+    if (res.status !== 200) {
+      const content = `<div><h4>Filepath: <a href="${file}" target="_blank">${filePath}</a></h4><div style="margin-left:.5rem"><a href="${url}" target="_blank">${url}</a> <p>Code: ${res.status}</p></div><br/></div>\n`;
+      fs.writeFile(outputFileName, content, { flag: 'a+' }, (err) => {
```
tamara-corbalt, line 70: I noticed that `fs.writeFile` is used to append results for each detected broken link. Is there a risk that broken links might be found concurrently and lead to race conditions in the write operation?

brokenLinkReport.html
jack-ryan-nava-pbc: Yeah, maybe will fix in the future. Thanks for pointing out!

brokenLinkReport.html
kim-cmsds, line 1: @jack-ryan-nava-pbc I think you'll need to delete this file since it was added before `brokenLinkReport.html` was added to .gitignore, and .prettierignore

jack-ryan-nava-pbc: I tried the ole `git rm --cached brokenLinkReport.html` which should remove it when this is pulled down. It might not have been sufficient though...

scripts/check-markdown-links.ts
```diff
@@ -0,0 +1,189 @@
...
+  const filePath: string = file.split('/').splice(spliceFrom).join('/');
+  const content = '';
+  fetch(url).then((res) => {
+    if (res.status !== 200) {
+      const rowContent = tableBodyContent(file, filePath, res.status, url);
+      tableBodyContainer.push(rowContent);
+    }
+    if (lastItem) {
+      const outputTable = table(tableHeader(), tableBodyContainer.join(''));
+      writeToFile(
```
kim-cmsds, line 152: It might be nice to add a success message like "Broken links found! Run `open brokenLinkReport.html` to view the list."

kim-cmsds (approved):
Left one last comment about whether it's worth adding a success message. Feel free to ignore since this works well locally! I was able to view the broken links in an organized table. 🎉

scripts/check-markdown-links.ts
jack-ryan-nava-pbc: @tamara-corbalt I think changes to the code get rid of this concern. 
1. File writing is now synchronous
2. We only write to the file in two places now: prior to the url requests to initialize the output file and once when we are done looping through the urls. That's the purpose of that 'lastItem' boolean.

tamara-corbalt (approved):

kim-cmsds (approved):

PR #3188, opened 2024-08-01
[NO-TICKET] Add uncontrolled-component support to `SingleInputDateField`

pwolfert:
## Summary

Add uncontrolled component support to `SingleInputDateField` so it works with web components. Web components are uncontrolled by definition, and I discovered that our `SingleInputDateField` (P)react component didn't actually support an uncontrolled component behavior. Even though the underlying `TextField` component supports the uncontrolled behavior,  we need access to that internal value in the `SingleInputDateField`, so I re-implemented the uncontrolled-component internal state tracking at the `SingleInputDateField` level. We have several other components that follow a very similar pattern of optionally tracking internal state based on whether the `value` is provided.

## How to test

Try out the new `Uncontrolled` story on the `SingleInputDateField` and specifically use the calendar picker to select a date. Calendar selection should also now work on the web component. Note that some changes were made to the web component to support it, and those changes will need to be merged with @tamara-corbalt's changes.

### DIFF
--- a/packages/design-system/src/components/DateField/SingleInputDateField.stories.tsx
+++ b/packages/design-system/src/components/DateField/SingleInputDateField.stories.tsx
@@ -26,26 +26,29 @@ export default meta;
 
 type Story = StoryObj<typeof SingleInputDateField>;
 
-export const Default: Story = {
+const UncontrolledTemplate: Story = {
+  render: function Component(args) {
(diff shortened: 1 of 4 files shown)


tamara-corbalt (approved):

tamara-corbalt:
Looks good!

PR #3199, opened 2024-08-12
[WNMGDS-2838] Update all themes to make inline tool tips black

jack-ryan-nava-pbc:
## Summary

[Ticket](https://jira.cms.gov/browse/WNMGDS-2838)
Changed the value pointed to in the tokens for '--tooltip-trigger__color` class. 

## How to test

1. Pull down the branch locally and build
2. Confirm that for each theme the color of the tooltip underline and text matches the color of the surrounding text. 

## Verify Screenshots look ok

cms_tooltip
cms_tooltip_hover
core_tooltip
core_tooltip_hover
healthcare_tooltip
healthcare_tooltip_hover
medicare_tooltip
medicare_tooltip_hover

## Checklist


packages/design-system-tokens/src/tokens/Theme.cmsgov.json
```diff
@@ -4507,7 +4507,7 @@
       },
       "trigger": {
         "color": {
-          "$value": "{theme.color.primary}",
+          "$value": "{theme.color.black}",
```
pwolfert, line 4510: To confirm, did you change these directly in JSON or make the change in Figma and then sync to code?

jack-ryan-nava-pbc: Changed directly in JSON. I didn't sync anything yet.

zarahzachz (commented):
I don't know how possible this is with Figma tokens, but should this value actually be `currentColor` so the tooltip text/underline matches its surrounding text color? 

I don't know if the goal is to make the tooltip match its surrounding text or to explicitly make it black for all scenarios (including the inverse background - if that's used at all, I don't know)

jack-ryan-nava-pbc:
> I don't know how possible this is with Figma tokens, but should this value actually be `currentColor` so the tooltip text/underline matches its surrounding text color?
> 
> I don't know if the goal is to make the tooltip match its surrounding text or to explicitly make it black for all scenarios (including the inverse background - if that's used at all, I don't know)

@zarahzachz I think this is a good callout. Looks like all of the Design Systems are using `color--base` in the text inherited from the `<body>`. I should be able to reference that variable in the themes via `theme.color.base` and give that a shot.

tests/browser/snapshots/stories/components-icons--hhs-logo-medicare--firefox.png
zarahzachz, line 1: Was this icon supposed to change?

tests/browser/snapshots/stories/reset-form-elements--form-elements-medicare--webkit.png
zarahzachz, line 1: Looks like a sizing change on this one too

tests/browser/snapshots/stories/components-icons--hhs-logo-medicare--firefox.png
jack-ryan-nava-pbc: Nope. That's unexpected. Hmm...

tests/browser/snapshots/stories/reset-form-elements--form-elements-medicare--webkit.png
jack-ryan-nava-pbc: also wut

zarahzachz:
> > I don't know how possible this is with Figma tokens, but should this value actually be `currentColor` so the tooltip text/underline matches its surrounding text color?
> > I don't know if the goal is to make the tooltip match its surrounding text or to explicitly make it black for all scenarios (including the inverse background - if that's used at all, I don't know)
> 
> @zarahzachz I think this is a good callout. Looks like all of the Design Systems are using `color--base` in the text inherited from the `<body>`. I should be able to reference that variable in the themes via `theme.color.base` and give that a shot.

Oh, my recommendation wasn't to change it to `color--base` - it was to change it to `currentColor`, which I don't know is possible with the current token setup (Figma only supports a subset of CSS properties). 

`currentColor` takes the color of whatever its parent has defined, meaning if this text were changed using a utility class like `ds-u-color--error`, the entire string (including the tooltip) would render red. I've attached a video explaining what I mean: 

https://github.com/user-attachments/assets/231545e0-b7cd-4fef-9bd8-baf13fc82689

I don't think Figma supports `currentColor`, which may mean removing this token from the project and hardcoding it to `currentColor` (we do this on other components like SvgIcon fill). Since this token is only used in this one instance, I think an argument could be made in favor of deprecating the token. That would make this a breaking change, but since this is meant to go out in v12 I don't think that's an issue. Just things to talk about with your team.

jack-ryan-nava-pbc:
Noting from discussion as of today:
1. We are continuing with these changes
2. We will investigate removing the tokens associated with Tooltips as part of our future Tooltip discovery work. That work will allow us to use `currentColor`

packages/ds-medicare-gov/src/styles/components/_Tooltip.scss
```diff
@@ -6,14 +6,14 @@
...
-  text-decoration-color: var(--color-primary);
+  text-decoration-color: var(--tooltip-trigger__color);
   text-decoration-style: dashed;
   text-decoration-thickness: var(--link__text-decoration-thickness);
   text-underline-offset: var(--link__text-decoration-offset);
 
   &:hover {
     color: var(--color-black);
-    text-decoration-color: var(--color-primary-darker);
+    text-decoration-color: var(--color-black);
```
pwolfert, line 16: Is this a temporary decision? Should it be flagged to revisit later?

pwolfert: Also, I found through testing that you've overridden the hover color for medicare but not for any of the other themes. Is that intentional?

jack-ryan-nava-pbc: No that was a whoopsie. Thank you for finding that. 

I need to document in the ticket the intent behind these changes and that they will likely be reviewed and possibly changed as we embark on the tooltip discovery work. I think at this moment I don't know if these changes are temporary in such a way that I know for a fact they will be removed and when. So given that, I think documentation in the ticket should be sufficient, but also happy to add a comment to the SCSS file.

pwolfert: 👍 It's up to you how you want to document it. I just wanted you to be able to remember it.

malloryiden (approved):
🎉

packages/design-system-tokens/src/tokens/Theme.cmsgov.json
```diff
@@ -4507,7 +4507,7 @@
       },
       "trigger": {
         "color": {
-          "$value": "{theme.color.primary}",
```
kim-cmsds, line 4510: @jack-ryan-nava-pbc have you run the sync to figma script yet?

kim-cmsds (approved):
Sounds like merging first, then running `syncTokensToFigma` is the path forward for making sure these tokens are updated in Figma. Thanks @jack-ryan-nava-pbc!

PR #3200, opened 2024-08-12
[RELEASE] 8/12 Version bump main

tamara-corbalt:


packages/design-system-tokens/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "design-system-tokens",
-  "version": "0.0.1",
+  "version": "1.0.0-beta.0",
```
pwolfert, line 3: Darn it, ignoring this one is another thing we relied on Lerna to do. What do you all think? Should this version match core or stay independent? Maybe it does make sense to mirror it, even if it remains unpublished.

pwolfert: This is the only problem I saw with this bump branch.

kim-cmsds: I don't have a strong opinion about whether this should stay independent. Since it's a change to expected functionality, @tamara-corbalt can you please revert this change?

jack-ryan-nava-pbc: I think Lerna also ignored the doc site directory as well: https://github.com/CMSgov/design-system/commit/d2ad38914c3e99ae9d29fa410cce170b65dff44f#diff-2d72bdead8afa0798d18995311992d684348a694c2d5e214e8e4d2b6153e4821L4

pwolfert (approved):

jack-ryan-nava-pbc (approved):
LGTM

PR #3205, opened 2024-08-19
[WNMGDS-2875] Fix Release script take 2

jack-ryan-nava-pbc:
## Summary

[Ticket](https://jira.cms.gov/projects/WNMGDS/2875)

- Main changes happened to the beginning of the bumpVersions function.  
- Now we manually update our package.json files
- Use npm version to update versions of items in workspaces as before
- Add getPackages function that pulls in our design system and examples dirs and package.json files.

## How to test

1. Run the `yarn release` script.
2. Confirm all of the appropriate json files are updated. One for each child design system, each example folder and the versions.json file.
3. Updated files:
[ ] - versions.json
[ ] - examples/astro/package.json
[ ] - examples/preact-app/package.json
[ ] - examples/preact-react-app/package.json
[ ] - examples/react-app/package.json
[ ] - packages/design-system/package.json
[ ] - packages/ds-cms-gov/package.json
[ ] - packages/ds-healthcare-gov/package.json
[ ] - packages/ds-medicare-gov/package.json

## Checklist


scripts/release.ts
```diff
@@ -86,11 +109,32 @@ async function bumpVersions() {
...
+  });
 
   // Update versions.json
   const currentVersionsByPackage = updateVersions();
+  // Clean up our staged files
+  sh('git reset ./packages/docs/package.json');
+  sh('git checkout ./packages/docs/package.json');
   sh('git add -u');
+  // Delete lingering package-lock.json
+  sh('rm ../package-lock.json');
```
kim-cmsds, line 137: Is this line needed? I got the following error when running locally:
```
rm: ../package-lock.json: No such file or directory
Command failed: rm ../package-lock.json
```

Maybe a try/catch block would suffice or a check to see whether the file exists before removing it?

scripts/release.ts
```diff
@@ -66,6 +67,26 @@ async function undoLastCommit() {
...
 
+const newDesignSystemVersion = (): string => {
+  const newDesignSystemVersionJSON = JSON.parse(sh(`npm version -w @cmsgov/design-system --json`));
+  return newDesignSystemVersionJSON['@cmsgov/design-system'];
+};
+
+const updateDSVersion = (json: JSON | any): JSON => {
+  const dsKey = '@cmsgov/design-system';
+  json.dependencies[dsKey] = newDesignSystemVersion();
+  return json;
```
kim-cmsds, line 78: Would it be possible to add a line break at the end of this json?

The file diff reads `\ No newline at end of file` for each of the package.json files that were updated after running `yarn release`

scripts/release.ts
```diff
@@ -86,11 +109,32 @@ async function bumpVersions() {
...
+    console.log(
+      c.green(`Bumped ${packageName} to @cmsgov/design-system@${newDesignSystemVersion()}.`)
+    );
+  });
 
   // Update versions.json
   const currentVersionsByPackage = updateVersions();
+  // Clean up our staged files
+  sh('git reset ./packages/docs/package.json');
+  sh('git checkout ./packages/docs/package.json');
```
pwolfert, line 134: I'm not following why this is needed. Is there an opportunity for the comment above to explain why?

scripts/release.ts
```diff
@@ -86,11 +109,32 @@ async function bumpVersions() {
     console.log(c.yellow('No version changes made. Exiting...'));
     process.exit(1);
   }
-  console.log(c.green('Bumped package versions.'));
+  console.log(c.green('Bumped package versions. Bumping dependencies next...'));
+
+  getPackages().forEach((pack) => {
+    const [packageName] = Object.keys(pack);
+    const [packageLocation] = Object.values(pack);
```
pwolfert, line 116: There seems to be some unnecessary complexity here in getting these values from the `getPackages()` function to here. Can this be simplified?

Is the purpose of `getPackages` to get all packages that have `@cmsgov/design-system` as a dependency? If that's the case, could the name be more specific? Could more be done in that function? Less? The boundary of the responsibilities between this inlined `forEach` and that function aren't entirely clear to me.

tamara-corbalt: I agree with Patrick about the complexity here. Perhaps moving the `forEach` logic into its own function (something like `updatePackageVersion`) could make the intention more explicit? I also like the idea of combining the logic in `getPackages` and the `forEach` loop into one function to streamline the processing.

scripts/release.ts
jack-ryan-nava-pbc: I think that might have to do with the fact that I'm then requiring in the JSON rather than reading the file directly. I'll make that change and we'll if it fixes it!

scripts/release.ts
jack-ryan-nava-pbc: Good idea. I also got the proper directory setup with the latest push.

scripts/release.ts
jack-ryan-nava-pbc: Yeah I can add another comment. The goal is to not track/update the docs package.json, as happened before with Lerna. So we're explicitly removing the changes related to the docs site.

pwolfert: Oh like you don't want to increment the version for the doc site?

pwolfert: Did you mean for this to be on the tokens package?

jack-ryan-nava-pbc: Tokens is handled on this [line](https://github.com/CMSgov/design-system/pull/3205/files#diff-21f9896491d70e481e031a1c66e9ffed07592f08455bdcd4a3cecc9c63814768R140).

We want to exclude both the docs and the tokens packages, which I think was the way things were working with the original Lerna-based script.

I'm not sure why it happens for docs and not tokens, but bumping the dependencies for the child system edits the package.json and stages it, just for docs though, so we need another explicit removal later in the script.

scripts/release.ts
jack-ryan-nava-pbc: @tamara-corbalt @pwolfert I tried my darnedest. Please let me know what questions and ideas you have after the refactor.

scripts/release.ts
```diff
@@ -66,6 +67,64 @@ async function undoLastCommit() {
   console.log(c.green('Publish commit deleted.'));
 }
 
+const newDesignSystemVersion = (): string => {
```
kim-cmsds, line 70: Because we're invoking this function for each child package, we're calling `npm version -w @cmsgov/design-system --json` more times than we probably need. What do you think about checking for whether a variable called `newDesignSystemVersion` is stored in memory and if not then firing off the npm version command?

scripts/release.ts
```diff
@@ -66,6 +67,64 @@ async function undoLastCommit() {
...
+ * Gets the file location, reads said file, bumps the version, writes the file.
+ */
+const updatePackageJson = (childPackageToUpdate: Array<string>): void => {
+  // Get our package name and package.json location via destructuring
+  const [packageName, packageLocation] = childPackageToUpdate;
+  /*
+   * The package location is stored relative to the root directory when returned by `npm ls`,
+   * so we have to clean it up a bit so `path` knows where to look.
+   */
+  const cleanedPackageLocation = packageLocation.split('file:')[1].replaceAll('../', '');
```
kim-cmsds, line 110: I think this would read better if you handled the package location updates in the `getPackageNamesAndLocations` function. That way the `getPackageNamesAndLocations` function can just focus on the package JSON updates.

scripts/release.ts
```diff
@@ -66,6 +67,64 @@ async function undoLastCommit() {
...
+  const jsonFileLocation = path.join(root, cleanedPackageLocation, 'package.json');
+  const json = readJson(jsonFileLocation);
+  const updatedJson = updateDSVersion(json);
+  writeJson(jsonFileLocation, updatedJson);
+  console.log(
+    c.green(`Bumped ${packageName} to @cmsgov/design-system@${newDesignSystemVersion()}.`)
+  );
+};
+
+// The thingy, wassit, function, what does the updating of said JSON files, guv'.
```
kim-cmsds, line 120: Looks like a typo?

scripts/release.ts
```diff
@@ -86,12 +147,25 @@ async function bumpVersions() {
...
+  sh('git checkout ./packages/docs/package.json');
   sh('git add -u');
+  // Delete lingering package-lock.json if it exists:
+  try {
+    sh(`rm ${root}/package-lock.json`);
+  } catch {
+    console.log('No package-lock.json at the top level. Moving on...');
+  }
   console.log(c.green('Updated versions.json.'));
+  process.exit(0);
```
kim-cmsds, line 168: Before merging, please remove

kim-cmsds (commented):
Making progress! I left a few comments to address.

scripts/release.ts
pwolfert: 😂

pwolfert: This is an American design system, guv'

scripts/release.ts
```diff
@@ -66,6 +67,64 @@ async function undoLastCommit() {
...
+ *  so we have to dig a bit.
+ * We grab the package names by getting all of the keys in the object associated with the 'dependencies' key.
+ * We loop over the names to get the package.json file locations that are in the object associated with the 'resolved' key.
+ */
+const getPackageNamesAndLocations = (): Array<Array<string>> => {
+  const packageJson: any = JSON.parse(sh('npm ls -ws @cmsgov/design-system --json'));
+  const deps = packageJson['dependencies'];
+  const packageNames = Object.keys(deps);
+  return packageNames.map((name: any) => {
+    return [name, deps[name]['resolved']];
```
kim-cmsds, line 95: what do you think about returning an object here?
```
{
  packageName: string,
  packageJsonFileLocation: string
}
```
I think that would improve the readability for destructuring in the `updatePackageJson` function.

Plus, `packageJsonFileLocation` could be based on `path.join(root, cleanedPackageLocation, 'package.json')` if you move the package location cleaup logic into this function

scripts/release.ts
pwolfert: Let's chat about this one all together live

scripts/release.ts
jack-ryan-nava-pbc: Good catch. Yeah I'll put into a var and then pass in as an argument.

scripts/release.ts
jack-ryan-nava-pbc: fiiiine....

scripts/release.ts
jack-ryan-nava-pbc: woopsie! Thanks for catching.

jack-ryan-nava-pbc: Done. I'm wondering if there is a linter rule that would've caught that?

pwolfert: Maybe ya'll should have a linter for British lingo in the comments

scripts/versions.ts
```diff
@@ -2,14 +2,14 @@ import path from 'node:path';
...
+export const root = path.join(__dirname, '..');
 const versionsFileName = path.join(root, 'versions.json');
 
-function readJson(filename: string): { [key: string]: any } {
+export function readJson(filename: string): { [key: string]: any } {
   return JSON.parse(readFileSync(filename, { encoding: 'utf8' }));
 }
 
-function writeJson(filename: string, json: { [key: string]: any }) {
+export function writeJson(filename: string, json: { [key: string]: any }) {
```
kim-cmsds, line 12: Thinking about how to add a new line to just the end of the package.json files if you didn't want to update how `versions.json` is formatted.

You could update the function like so
```suggestion
export function writeJson(filename: string, json: { [key: string]: any }, addNewLine?: boolean) {
  const stringifiedJson = JSON.stringify(json, null, 2);
  const content = addNewLine ? `${stringifiedJson}\n` : stringifiedJson
  writeFileSync(filename, content);
}
```

and then where `writeJson` is invoked in the `bump-child-deps-utils`, you could call it like `writeJson(jsonFileLocation, updatedJson, true);`

kim-cmsds (approved):
Left one note about improving the format of the `package.json` files to avoid more changed lines than necessary when running `yarn release`. I was able to test this out locally up until the publishing of tags step.

scripts/versions.ts
jack-ryan-nava-pbc: Love it. Thank you @kim-cmsds

scripts/versions.ts
```diff
@@ -2,15 +2,17 @@ import path from 'node:path';
...
-function readJson(filename: string): { [key: string]: any } {
+export function readJson(filename: string): { [key: string]: any } {
   return JSON.parse(readFileSync(filename, { encoding: 'utf8' }));
 }
 
-function writeJson(filename: string, json: { [key: string]: any }) {
-  writeFileSync(filename, JSON.stringify(json, null, 2));
+export function writeJson(filename: string, json: { [key: string]: any }, addNewLine?: boolean) {
+  const stringifiedJson = JSON.stringify(json, null, 2);
+  const content = addNewLine ? `${stringifiedJson}\n` : stringifiedJson;
```
pwolfert, line 14: Since the json files that get checked in are prettified by the pre-commit hook, is this logic necessary?

kim-cmsds: The issue is that they're not getting prettified. It looks like the [lint-staged config](https://github.com/CMSgov/design-system/blob/main/package.json#L63-L75) doesn't pay attention to files that end with `.json`.

pwolfert: Ah, darn

kim-cmsds: yay! https://github.com/CMSgov/design-system/pull/3206

jack-ryan-nava-pbc: Thanks for that super fast change @pwolfert - latest changes merge in main and undo the changes to writeJSON

scripts/bump-child-deps-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+     * The package location is stored relative to the root directory when returned by `npm ls`,
+     * so we have to clean it up a bit so `path` knows where to look.
+     */
+    const cleanedPackageLocation = deps[name]['resolved'].split('file:')[1].replaceAll('../', '');
+    return { packageName: name, packageLocation: cleanedPackageLocation };
+  });
+};
+
+/*
+ * Actually does the updateing of the package.json files.
```
tamara-corbalt, line 53: Nit: typo :)

scripts/bump-child-deps-utils.ts
```diff
@@ -0,0 +1,76 @@
...
+  const newDesignSystemVersionJSON = JSON.parse(sh(`npm version -w @cmsgov/design-system --json`));
+  return newDesignSystemVersionJSON['@cmsgov/design-system'];
+};
+
+/*
+ * Set the @cmsgov/design-system dependency to the new version in each child design system and example package
+ */
+const updateDSVersion = (json: JSON | any): JSON => {
+  const dsKey = '@cmsgov/design-system';
+  newVersionNumber = newVersionNumber ?? getNewDesignSystemVersion();
```
tamara-corbalt, line 21: It looks like `newVersionNumber` will only be `undefined` the first time it's called. On subsequent calls, it will already be defined. Would it be more appropriate to assign its value before entering the loop rather than within it?






[12:05](https://corbalt.slack.com/archives/D073JLCFBDZ/p1724256355490179)
On subsequent iterations of the foreEach, newVersionNumber will be defined.
12:06
So, why not define it before the loop, like so:
// This function kicks off everything.
export const updateChildDSAndExamples = (): void => {
  // Fetch the version number once before the loop.
  newVersionNumber = getNewDesignSystemVersion();
  
  const allOurPackages = getPackageNamesAndLocations();
  allOurPackages.forEach((childPackage) => {
    bumpPackageJsonDependencies(childPackage);
  });
};

jack-ryan-nava-pbc: Good call out! Fixing.

tamara-corbalt (approved):

pwolfert (approved):
Everything's working as far as I can tell 🎉

PR #3206, opened 2024-08-20
[NO-TICKET] Automatically format json files

pwolfert:
## Summary

Automatically format json files so you don't have to worry about formatting when writing to these files programmatically


kim-cmsds (approved):
Nice!

PR #3226, opened 2024-09-10
[NO-TICKET] Sort Figma token scope values alphabetically

jack-ryan-nava-pbc:
## Summary

- @pwolfert and I noticed that the array of scopes for certain Figma tokens were not sorted alphabetically. Guess what? Now they are!

## How to test

1. Run `yarn sync:tokens-from-figma` locally
2. Confirm that any `scopes` arrays with more than one value are sorted alphabetically. You should see some in the file diff viewer in VS Code.

## Checklist


pwolfert (approved):
I would maybe change the `Type: Fixed` label to `Type: Internal` label since it's not affecting anything that gets released downstream, but other than that LGTM 👍

PR #3227, opened 2024-09-11
[WNMGDS-2895] Update font and focus color design tokens in CMS.gov theme

malloryiden:
## Summary

- Changed all headers from Open Sans to Lexend
- Changed body copy from Open Sans to Public Sans
- Changed focus state color from Orchid to Copper
- Changed bold (700) headers to semibold (600)

## How to test

1. Check Theme.cmsgov.json file for above changes
2. Check other theme files for adding the "semibold" token

## Checklist


packages/design-system-tokens/src/tokens/Theme.core.json
```diff
@@ -1149,6 +1149,16 @@
               "codeSyntax": {}
             }
           }
+        },
+        "semibold": {
```
zarahzachz, line 1153: just a heads up that this could cause issues. semibold has no meaning on the web. 

1. some fonts don't have a semibold/600 value
2. as far as markup is concerned, there's no "semibold" tag to use in CMS (content management systems) - we only have normal text tags (like `<p>`) and `<strong>`

I wonder if it makes sense to abstract the idea of bold vs not-bold for future theming? Will themes have a need for nuanced weights (probably not since non-retina screens won't be able to differentiate), so maybe having a token representing font weights as "normal" or "default" and "heavy" (or whatever standin word for your bolded type) could prevent the issues presented above?

What I mean is all themes would have a "normal" and "heavy" font weight token, but the value of those tokens could differ across themes. Mgov could have its "heavy" as 700 and CMS  could have its heavy as 600 without the need to give all themes all weights.

zarahzachz:
Oh! And since this is a font family change, I think this is considered a breaking change (major version)

packages/design-system-tokens/.env.example
```diff
@@ -1,2 +0,0 @@
-FILE_KEY=OYkYP4pC9jwS7j2qafwmiv
```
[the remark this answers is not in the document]
malloryiden: Oh whoops, looks like I renamed it and added my personal token instead of copying it first. Thanks!

packages/design-system-tokens/src/tokens/Theme.core.json
malloryiden: Could this introduce other issues in a case where a theme wants to use two "heavy" weights (like both 600 and 700)?

malloryiden:
> Oh! And since this is a font family change, I think this is considered a breaking change (major version)
Added, thank you!

packages/design-system-tokens/src/tokens/System.Value.json
```diff
@@ -2941,6 +2941,27 @@
       }
     },
     "family": {
+      "lexend": {
+        "$value": [
+          "Lexend",
+          "Open Sans",
```
zarahzachz, line 2947: just noticed this... this should be removed as you wouldn't have another web font as a fallback font - should be lexend + system supported fonts for fallbacks in case lexend doesn't load.

packages/design-system-tokens/src/tokens/System.Value.json
```diff
@@ -2961,6 +2982,27 @@
           }
         }
       },
+      "public-sans": {
+        "$value": [
+          "Public Sans",
+          "Open Sans",
```
zarahzachz, line 2988: same fallback comment as above re: removing open sans here

packages/design-system-tokens/src/tokens/Theme.medicare.json
```diff
@@ -1158,6 +1158,16 @@
               "codeSyntax": {}
             }
           }
+        },
+        "semibold": {
```
zarahzachz, line 1162: this is what i'm talking about above - the "bold" token is mapped to the 600 weight, meaning there is no true bold/700 weight font in mgov

packages/design-system-tokens/src/tokens/Theme.core.json
zarahzachz: sure thing!

so first, we used to support bold, semibold and regular font weights (i think also thin at one point?) but we reduced them to the current selection to shrink our bundle size. 

the variable "semibold" was [recently deprecated](https://github.com/CMSgov/design-system/blob/main/packages/design-system/scripts/cmsds-migrate/configs/deprecated-css-v11.mjs#L5) from the DS (as of 3 weeks ago). but the value for it (600 weight) is still active in the system. you can see this on the mgov tokens

variable fonts could solve this problem, but are also large file sizes, so there's a tradeoff as far as asset downloads go. the same asset size issue exists when adding additional font weight files.

malloryiden: Great context, thank you! (Also lol for deprecating semibold and then introducing it again now).

jack-ryan-nava-pbc:
@malloryiden can you update the title of this PR to be WNMGDS-2895?

jack-ryan-nava-pbc:
The number of snapshots changed is absolutely massive. I'm wondering if it makes sense to make a few of these changes in broken up pull requests so we don't get a diff of 800+ files? What do you think @kim-cmsds @tamara-corbalt @malloryiden ?

jack-ryan-nava-pbc (approved):
I'm approving! Nice work @malloryiden 

I do think we should have a larger conversation around an acceptable number of files changed. 846 is a lot, and maybe we can come up with strategies for breaking up big pull requests (or pull requests with lots of visual changes) so we keep the size to something more manageable. 

cc @kim-cmsds @tamara-corbalt I'm really curious about what you two think. Thanks!

kim-cmsds:
I am curious about what would have happened if this work was split out into 4 separate PRs based on the bullet points that describe the changes made.

kim-cmsds (approved):

malloryiden:
Closing this ticket and will be breaking up each item into it's on PR

PR #3230, opened 2024-09-16
[WNMGDS-2935] Make Sass token file changes backwards compatible

pwolfert:
## Summary

Makes our Figma changes backwards compatible with respect to the sass var files. I had failed to adequately document the changes in https://github.com/CMSgov/design-system/pull/3100, which resulted in a failure to report the breaking change of a file-name change in our release notes. Luckily our integration tests caught it in one of the beta releases.

Note that instead of reproducing the same old contents of those files, they're just duplicates of the new file, and that's enough. It won't break anything if a product imports both of these files, because they just re-declare the same variables. It also has no effect on the final CSS bundle sizes because Sass variables are ephemeral.

## How to test

Clear all your `dist` folders and rebuild. You should see the same old file names that we used to output.

@kim-cmsds, could you also test this directly in the downstream app that had the issue?

## Checklist

### DIFF
--- a/packages/design-system-tokens/src/css/translate.test.ts
+++ b/packages/design-system-tokens/src/css/translate.test.ts
@@ -39,7 +39,19 @@ describe('tokenFilesToCssFiles', () => {
 
 describe('tokenFilesToScssFiles', () => {
   it('matches snapshot', () => {
-    expect(tokenFilesToScssFiles(tokensByFile)).toMatchSnapshot();
+    const files = tokenFilesToScssFiles(tokensByFile);
+
+    // Acknowledge but don't snapshot the duplicated files
+    expect(files['core-component-tokens.scss']).toBeTruthy();
+    expect(files['core-tokens.scss']).toBeTruthy();
+    expect(files['cmsgov-component-tokens.scss']).toBeTruthy();
+    expect(files['cmsgov-tokens.scss']).toBeTruthy();
+    delete files['core-component-tokens.scss'];
+    delete files['core-tokens.scss'];
+    delete files['cmsgov-component-tokens.scss'];
+    delete files['cmsgov-tokens.scss'];
+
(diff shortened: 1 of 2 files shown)


tamara-corbalt:
Looks like we need to update a snapshot. 
```
 FAIL  design-system-tokens/src/css/translate.test.ts
  ● tokenFilesToScssFiles › matches snapshot

    expect(received).toMatchSnapshot()

    Snapshot name: `tokenFilesToScssFiles matches snapshot 1`
```

kim-cmsds:
@pwolfert I'm seeing an error after trying this out locally. On this branch, I ran the `yarn clean && yarn build` commands within the `design-system-tokens` package. Then I ran `npm install` and manually copied the backwards compatible healthcare sass files (`healthcare-component-tokens.scss` and `healthcare-tokens.scss`) into the downstream project's `node_modules/@cmsgov/ds-healthcare-gov/dist/scss` folder. The error I saw when running `npm run start` in the downstream project folder is as follows:

```
8:24:05 AM [vite] Internal server error: [sass] Undefined variable.
    ╷
135 │ $font-weight-heading-5xl: $font-weight-bold;
    │                           ^^^^^^^^^^^^^^^^^
    ╵
  node_modules/@cmsgov/ds-healthcare-gov/dist/scss/healthcare-tokens.scss 135:27  @import
  src/index.scss 6:9                                                              root stylesheet
  Plugin: vite:css
  File: hra-tool/node_modules/@cmsgov/ds-healthcare-gov/dist/scss/healthcare-tokens.scss:135:27
  Error: Undefined variable.
```

pwolfert:
@kim-cmsds, you found a new bug! The order doesn't matter for CSS variables, but in Sass a variable can't be referenced before it's defined. That means my naive alphabetical ordering isn't going to work. Good catch!

kim-cmsds (approved):

PR #3234, opened 2024-09-18
[WNMGDS-2892] Adds `ds-tabs` web component.

tamara-corbalt:
## Summary
 - Adds `ds-tabs` web component.

## How to test

- Run `yarn storybook` & select `ds-tabs` listed under Web Components. ([WNMGDS-2892]([https://jira.cms.gov/browse/WNMGDS-2892]))

## Checklist

Screenshot 2024-09-18 at 10 18 28 AM
Screenshot 2024-09-18 at 10 18 45 AM


packages/design-system/src/components/web-components/ds-tabs/ds-tabs.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+const Wrapper = ({ children = [], ...otherProps }: WrapperProps) => {
+  function parseChildren(node) {
+    const elements = findElementsOfType(['ds-tab-panel'], node);
+
+    return elements.map((element) => {
+      const { children, ...attrs } = element.props || {};
+      return createElement(TabPanel, { ...attrs }, children);
+    });
+  }
```
pwolfert, line 24: This pattern is really coming into its own! Your code here is nice and concise 👌

pwolfert (commented):
It works in Storybook, but it needs a `ds-tabs/index.ts` and an entry in the `web-components/index.ts` in order for the `build` command to work.

kim-cmsds (commented):
Looks like you'll want to export `ds-tabs` and `ds-tab-panel` in this [file](https://github.com/CMSgov/design-system/blob/tamara/wnmgds-2892-add-ds-tabs-web-component/packages/design-system/src/components/web-components/index.ts)

tests/browser/expectNoAxeViolations.ts
```diff
@@ -48,12 +48,14 @@ function createErrorMessage(violations: AxeResults['violations']): string {
...
-export default async function expectNoAxeViolations(page: Page) {
+export default async function expectNoAxeViolations(page: Page, shouldUseAxeLegacyMode: boolean) {
   const results = await new AxeBuilder({ page })
+    .setLegacyMode(shouldUseAxeLegacyMode)
     .withTags(RULESET_ALL)
     .disableRules(DISABLED_RULES)
     .analyze();
   // Disable Jest linting rule because it isn't Jest!
   // eslint-disable-next-line jest/valid-expect
+  console.log('Violations', results?.violations);
```
kim-cmsds, line 59: Can you please remove this console log?

tests/browser/stories.test.ts
```diff
@@ -20,6 +20,8 @@ const storySkipList = [
   'components-label--default', // Captured by individual form fields
   'components-skipnav--default-skip-nav',
   'components-skipnav--skip-nav-example',
+  'components-tabs--default',
+  'components-tabs--disabled',
```
kim-cmsds, line 24: This change isn't needed, correct?

jack-ryan-nava-pbc (commented):
@tamara-corbalt great work on this! It looks great so far. I'm wondering if we shouldn't break this out like the ds-accordion and ds-accordion-item components? I know that the tab-panel and tabs components are unlikely to be used separately, but I wonder if it'll be helpful to the tab-panel in detail and get an idea of the attributes that can be tweaked. What do you think?

tamara-corbalt:
> @tamara-corbalt great work on this! It looks great so far. I'm wondering if we shouldn't break this out like the ds-accordion and ds-accordion-item components? I know that the tab-panel and tabs components are unlikely to be used separately, but I wonder if it'll be helpful to the tab-panel in detail and get an idea of the attributes that can be tweaked. What do you think?

@jack-ryan-nava-pbc, I don't see any downside to following the ds-accordion / ds-accordion-item pattern. I'd be happy to write a follow-up ticket for this, where ds-tab-panel would have a dedicated story file so that its props are fully exposed to developers via storybook.

jack-ryan-nava-pbc:
@tamara-corbalt I noticed a bug where the tab keyboard functionality does not match the arrow key functionality: 

https://github.com/user-attachments/assets/9c803a42-4486-4544-ae24-d5e213754019

tamara-corbalt:
@jack-ryan-nava-pbc, it looks like the arrow behavior issue isn't specific to the ds-tabs component—I'm seeing it with the React version (Tabs) as well. Since this affects both versions, I've created a follow-up ticket to address it and added it to the backlog: [WNMGDS-2976](https://jira.cms.gov/browse/WNMGDS-2976)

I've also created a follow-up ticket regarding adding a dedicated story file for ds-tab-panel so that it aligns more closely with the pattern used for Accordion and Accordion-item components. That has also been added to Jira: [WNMGDS-2977](https://jira.cms.gov/browse/WNMGDS-2977)

@malloryiden, since the aria-label issue affects both the React and web component versions of Tabs, I've added a follow-up ticket in Jira to handle this: [WNMGDS-2978](https://jira.cms.gov/browse/WNMGDS-2978)

malloryiden (approved):
Since we'll address the aria-label issue in [WNMGDS-2978](https://jira.cms.gov/browse/WNMGDS-2978), I'm approving the accessibility portion of this! 

(Note that I only reviewed this from an a11y lens and a code review should still be done!)

kim-cmsds (approved):

PR #3249, opened 2024-09-27
[WNMGDS-2893] Implement Tooltip web component

jack-ryan-nava-pbc:
## Summary

- Initial implementation of the ds-tooltip web component
- Stories are written, and seek to emulate the existing React-based component
- Tests are written, and seek to emulate the existing React-based component
- There are a few notable differences between the web component and the React component:
  - No onOpen or onClose custom events - I figured these could be manually attached rather than being inbuilt, especially since these are used somewhat infrequently and for analytics when they are used.
  - No aria-label - see [Slack Thread](https://cmsgov.slack.com/archives/CT169N26L/p1727468064914539) for context. 

## How to test

1. Pull down branch
2. Confirm tests pass locally
3. Confirm controls in stories all change the component in expected ways.
4. Confirm that the usability of the component matches the expectations from the React component.

## Checklist


packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
```diff
@@ -0,0 +1,118 @@
...
+  if (typeof offset === 'string') {
+    const vals = offset.split(' ');
+    const [xOffset, yOffset] = vals;
+    return [parseInt(xOffset), parseInt(yOffset)];
+  }
+  // Else return the default offset value
+  return Tooltip.defaultProps.offset as TooltipProps['offset'];
+};
+
+const isPlacementValue = (location: string): location is Placement => {
```
jack-ryan-nava-pbc, line 72: The goal with this function is to enforce some guardrails on the user input for the placement value attribute. This could be overwrought, and I can probably just remove it and replace the type definition in the WrapperProps as `string | TooltipProps['placement']`.

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
```diff
@@ -0,0 +1,118 @@
...
+  dialog?: string;
+  inversed?: string;
+  showCloseButton?: string;
+  placement?: string;
+  rootId?: string;
+  offset?: string;
+  title: string | TooltipProps['title'];
+}
+
+const parseOffset = (offset: string): TooltipProps['offset'] => {
```
jack-ryan-nava-pbc, line 62: Trying to support an array as a string is maybe a bit wrongheaded. Curious if people think this is likely unnecessary. I have seen this prop used rarely in the wild in the React components as well. So maybe this just gets cut out.

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+    userEvent.click(tooltipTrigger);
+    const contentEl = screen.queryByRole('dialog');
+    expect(contentEl).not.toBeNull();
+    expect(contentEl).toMatchSnapshot();
+  });
+
+  it('closes tooltip when trigger focus is lost', async () => {
+    jest.useFakeTimers();
+    const { container } = renderTooltip();
+    const tooltip = container.querySelector('.ds-c-tooltip');
```
jack-ryan-nava-pbc, line 79: I doubt this class name is going to change anytime soon which is why I didn't go through the trouble of assigning a testID for this query.

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+  it('renders a tooltip', () => {
+    const { asFragment } = renderTooltip();
+    expect(asFragment()).toMatchSnapshot();
+  });
+
+  it('renders title and contentHeading when passed in as slots', () => {
+    renderTooltip(propsWithSlots);
+    const tooltipTrigger = screen.getByText(customTooltipText);
+    expect(tooltipTrigger).toHaveTextContent(customTooltipText);
+    expect(tooltipTrigger).toHaveTextContent(customHeadingText);
```
jack-ryan-nava-pbc, line 44: Since we set both the `title` and the `contentheading` slot values, we expect both sections of text to be contained within the tooltip.

packages/design-system/src/components/Tooltip/index.ts
```diff
@@ -1,2 +1,4 @@
+export * from './Tooltip';
+export * from './TooltipIcon';
```
jack-ryan-nava-pbc, line 2: We export everything mostly to get access to the type definitions. I can tighten this export statement to just reveal the types if folks think that is a better idea?

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
```diff
@@ -0,0 +1,120 @@
...
+  dialog,
+  inversed,
+  offset,
+  rootId,
+  title,
+  ...otherProps
+}: WrapperProps) => (
+  <Tooltip
+    {...otherProps}
+    ariaLabel={otherProps.triggerAriaLabel}
```
jack-ryan-nava-pbc, line 104: OK, so I think the way this is setup actually handles all the cases we might be worried about re duplicative aria labelling. 
1. You can have both `aria-label` and `trigger-aria-label` exist on the component. The `trigger-aria-label` attribute will still apply to the child trigger element, and the `aria-label` value will apply only to the parent.
2. If you apply an `aria-label` and no `trigger-aria-label` attribute, the new `aria-label` attribute only exists on the parent web component container. It is not applied to the child trigger element.

As the code currently is, I think these are sufficiently distinct. Please let me know if you have any concerns @kim-cmsds @tamara-corbalt @pwolfert

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
pwolfert: Yeah, I'd argue that this one isn't necessary right now

jack-ryan-nava-pbc: Sounds good! To the chopping block it goes.

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
```diff
@@ -0,0 +1,106 @@
...
+    | 'ariaLabel'
+    | 'contentHeading'
+    | 'dialog'
+    | 'inversed'
+    | 'showCloseButton'
+    | 'offset'
+    | 'placement'
+    | 'title'
+  > {
+  contentHeading?: string | TooltipProps['contentHeading'];
```
kim-cmsds, line 52: Do you have a story to demonstrate that it's possible to pass in a `ReactNode` (e.g. what `TooltipProps['contentHeading']` implies here)?

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
```diff
@@ -0,0 +1,106 @@
...
+  dialog?: string;
+  inversed?: string;
+  showCloseButton?: string;
+  placement?: string;
+  rootId?: string;
+  title: string | TooltipProps['title'];
```
kim-cmsds, line 58: Same idea as above. Do you have a story to demonstrate that it's possible to pass in a ReactNode (e.g. what TooltipProps['title'] implies here)?

kim-cmsds (commented):
Things are looking good!

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.test.tsx
```diff
@@ -0,0 +1,152 @@
...
+      'show-close-button': 'true',
+      title: customTooltipText,
+    });
+    const tooltipTrigger = screen.getByText(customTooltipText);
+    userEvent.click(tooltipTrigger);
+    const contentEl = screen.queryByRole('dialog');
+    expect(contentEl).toMatchSnapshot();
+  });
+
+  it('should return focus back to trigger when closed', () => {
```
kim-cmsds, line 129: Is it work adding another expectation around having the tooltipContent be closed?

https://github.com/CMSgov/design-system/blob/ba8a9262992e91a3111900dd1909b51e6ee6f374/packages/design-system/src/components/Tooltip/Tooltip.test.tsx#L113

tests/browser/snapshots/stories-interaction/dropdown-optgroups--open--medicare--webkit.png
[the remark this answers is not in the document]
jack-ryan-nava-pbc: No, I'm not sure why it has appeared?

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -85,6 +85,10 @@ export interface BaseTooltipProps {
    * Duration of the `react-transition-group` CSSTransition. See the [`timeout` option](http://reactcommunity.org/react-transition-group/transition#Transition-prop-timeout) for more info.
    */
   transitionDuration?: number;
+  /**
+   * In the case of web components we need to pass in the aria label separately to target the trigger element so we don't apply it to the web component wrapper.
```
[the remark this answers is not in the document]
jack-ryan-nava-pbc: Love it. Great suggestion!

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
jack-ryan-nava-pbc: So this is my attempt at a [type predicate](https://www.typescriptlang.org/docs/handbook/2/narrowing.html#using-type-predicates). 

This check does return a boolean, but also Typescript narrows the type via the `is Placement` statement.

jack-ryan-nava-pbc: This [Stack Overflow thread](https://stackoverflow.com/questions/40081332/what-does-the-is-keyword-do-in-typescript) is helpful

tests/browser/snapshots/stories-interaction/dropdown-optgroups--open--medicare--webkit.png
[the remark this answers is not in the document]
kim-cmsds: this looks "out of date"! feel free to resolve this thread

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
jack-ryan-nava-pbc: So interestingly the React.ReactNode type has the following definition:

```
type ReactNode =
        | ReactElement
        | string
        | number
        | Iterable<ReactNode>
        | ReactPortal
        | boolean
        | null
        | undefined
        | DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES[
            keyof DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES
        ];
```
So actually adding this as an available type is too permissive on the web component. I think the compiler was yelling at me, let me see if I can figure another way to satisfy Typescript.

pwolfert: > DO_NOT_USE_OR_YOU_WILL_BE_FIRED_EXPERIMENTAL_REACT_NODES

😂

packages/design-system/src/components/web-components/ds-tooltip/ds-tooltip.tsx
kim-cmsds: thank you!

kim-cmsds (approved):
Appreciate you looking into the `ariaLabel` conundrum and providing a solid solution.

PR #3253, opened 2024-09-30
[NO-TICKET] Maintain local state in web components when attributes change

pwolfert:
## Summary

Previously when one of our web components experienced an attribute change, it would call the `render` function again for the Preact component. However, this is not a normal lightweight "re-render" that Preact/React components undergo whenever state changes; it's actually a re-initialization of the entire render tree, which means that it obliterates all previous state and context. That's not really what we want. Here's a roughly drawn sketch of what the new architecture looks like below:

Diagram showing how the StateWrapper fits into the architecture of the define.ts module

The only new piece is the `StateWrapper`. Instead of calling `render(h(...))` whenever attributes change (and obliterating state), we only render a top-level `StateWrapper` once and allow attribute changes to update the `StateWrapper`'s internal state so that it re-renders itself within the same render tree that was created by the first `render(h(...))` call.

I really misunderstood the extent of `render` before and conflated it in my head with the normal _render_ cycle of a component. This new setup comes from an improved mental model.

Note that it's still possible to cause the render tree to be re-initialized within this custom element if inner HTML changes. In that case, we're making an assumption for now that our user is okay with clobbering the internal state because it's such a big change.

### Other changes

- Adds `preact` and `@preact/signals` as dependencies of the core package, because they technically are required for the web components (maybe we can discuss this)
- Removes `preactement` as a dependency of the core package, since we have our own copy of it now

## How to test

Create some examples where an attribute is updated mid-flight. (There's a commit here that was undone that has an example)

## Checklist


packages/design-system/package.json
```diff
@@ -69,7 +70,7 @@
     "inquirer": "^9.0.2",
     "lodash": "^4.17.21",
     "ora": "^6.1.2",
-    "preactement": "1.8.5",
```
kim-cmsds, line 72: It sounds like this change (removing `preactment` as a dependency) is unrelated to this particular PR but provides an opportunity to clean up because we're using a modified API based on preactment that can be found in the `define.ts` file.

Technically, we're relying on `preact` as a dev dependency, correct? However, if another team is using our web components, they'll need access to `preact` as during their local build process, hence why it's in the `dependencies` list. This is safer and make sense.

packages/design-system/src/components/web-components/preactement/define.ts
```diff
@@ -353,30 +355,35 @@ function renderPreactComponent(this: CustomElement) {
...
-    children,
-    ...slots,
-  };
+  // If we were to call this function that we're in every time a prop changed, it would
+  // clobber the internal state of our underlying Preact component. To avoid this, we're
+  // going to tie our attributes (props) to Preact Signals that we can use to trigger
+  // re-rendering a `StateWrapper` component.
+  const propsSignal = signal(this.__properties);
+  this.__propsSignal = propsSignal;
+  const StateWrapper = () => h(this.__component, { ...propsSignal.value, ...slots, children });
```
kim-cmsds, line 379: super clever!

kim-cmsds (approved):

PR #3256, opened 2024-10-01
[WMNGDS-2918] Adds ds-drawer web component.

tamara-corbalt:
## Summary

- Adds `ds-drawer` web component.

## How to test

- Run `yarn storybook` & select `ds-drawer` listed under Web Components. [WNMGDS-2892]([WNMGDS-2892](https://jira.cms.gov/browse/WNMGDS-2918))

## Checklist

Screenshot 2024-10-01 at 11 45 52 AM
Screenshot 2024-10-01 at 11 46 05 AM


packages/design-system/src/components/web-components/ds-drawer/ds-drawer.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+      minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo
+      consequat. Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu
+      fugiat nulla pariatur. Excepteur sint occaecat cupidatat non proident, sunt in culpa qui
+      officia deserunt mollit anim id est laborum.
+    </p>
+  </>
+);
+
+const Template = (args) => {
+  const [drawerOpen, setDrawerOpen] = useState(args['is-open'] ?? false);
```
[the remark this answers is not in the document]
jack-ryan-nava-pbc: I pushed up a small change that I think fixes this.

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+
+      drawerElement.addEventListener('ds-close-click', handleDrawerClose as EventListener);
+      toggleButton.addEventListener('click', handleDrawerOpen);
+
+      return () => {
+        drawerElement.removeEventListener('ds-close-click', handleDrawerClose as EventListener);
+        toggleButton.removeEventListener('click', handleDrawerOpen);
+      };
+    }
+  }, [drawerOpen]);
```
jack-ryan-nava-pbc, line 161: Tracking the state on this useEffect seems to resolve the above highlighted issue.

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+      laboris nisi ut aliquip ex ea commodo consequat.
+    </p>
+  </>
+);
+
+function renderDrawer(args, children) {
+  return render(
+    <ds-drawer {...(args as any)}>
+      {children}
+      {args['footer-body'] && (
```
jack-ryan-nava-pbc, line 20: I don't think you need to check for the attribute and then put in the slot. See above comment.

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.test.tsx
```diff
@@ -0,0 +1,122 @@
...
+  });
+
+  it('should render a footer-body when footer-body attribute is provided', () => {
+    renderDrawer(
+      { 'is-open': 'true', heading: 'Test Drawer Heading', 'footer-body': 'Footer Content' },
+      children
+    );
+    const renderedFooterBodyElement = screen.getByText('Footer Content');
+    expect(renderedFooterBodyElement).toBeInTheDocument();
+  });
```
jack-ryan-nava-pbc, line 75: *NIT*: spacing

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.tsx
```diff
@@ -0,0 +1,89 @@
+import { define } from '../preactement/define';
+import { Drawer } from '../../Drawer';
+import { DrawerProps } from '../../Drawer/Drawer';
+import { parseBooleanAttr } from '../wrapperUtils';
+
+const attributes = [
+  'aria-label',
```
jack-ryan-nava-pbc, line 7: `aria-label` here is duplicated in the parent ds-drawer element and in the child button element. I think you need to alias this like we do for the `root-id` attribute.

jack-ryan-nava-pbc (commented):
Looking good! I am having some real trouble getting this component to work in the Astro example playground. Maybe we can pair on seeing how to render this in a non-storybook environment?

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.tsx
tamara-corbalt: Hey @jack-ryan-nava-pbc, thanks for pointing that out! I think I might be seeing a different issue than the duplication you mentioned. When I pass an aria-label value to the ds-drawer parent in storybook, I expect it to be applied to the [close button element](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/Drawer/Drawer.tsx#L123). Instead, the close button element's aria-label is defaulting to "Close Help Drawer" because the `ariaLabel` prop appears to be `undefined`. 

I think the setup might need an adjustment, but I'm curious—could you outline the steps you took to reproduce the parent/child aria-label duplication?

jack-ryan-nava-pbc: Sure! So the easiest way to replicate this is in storybook itself. 
1. Set the `aria-label` value in the controls section to something
2. Check that it is set on the `ds-drawer` parent element and the button element in the `ds-c-drawer__header`

You can also do this by setting the value in the `formattedArgs` hash in the stories file. I set the `aria-label` key to be 'Test label' and saw it duplicated.

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+            "A callback function triggered when the user clicks the close button or presses the ESC key, provided focus trapping is active. The visibility of the drawer is managed by the parent component, which must utilize this callback to toggle the drawer's display. The drawer itself does not automatically hide.",
+        },
+      },
+    },
+  },
+  argTypes: {
+    'aria-label': {
+      description: 'Gives more context to screen readers on the Drawer close button.',
+      control: 'text',
+    },
```
pwolfert, line 37: Is this going to get the same treatment that @jack-ryan-nava-pbc's is getting where it gets a unique name so it doesn't get applied to both the root and child element in the a11y tree?

pwolfert (commented):
The docs look great, and the component works great 👍  Just one question below

P.S., I also tested it in VoiceOver

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.tsx
tamara-corbalt: @jack-ryan-nava-pbc It seems I tested this earlier while I had some unintended local changes applied. Now I'm seeing the duplication. Do you think we should take a similar approach to the one taken in [ds-tooltip](https://github.com/CMSgov/design-system/pull/3249#discussion_r1781963532)?

jack-ryan-nava-pbc: I think that would make sense so we avoid any confusion around the labelling as read by screen readers.

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.stories.tsx
tamara-corbalt: @pwolfert I think it might. We were just discussing that [here](https://github.com/CMSgov/design-system/pull/3256#discussion_r1784886379)

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+  const toggleButtonArgs = {
+    'class-name': 'ds-c-drawer__toggle',
+    variation: 'ghost',
+  };
+
+  return (
+    <div>
+      <ds-drawer {...formattedArgs}>
+        {drawerContent}
+        {args['footer-body'] && (
```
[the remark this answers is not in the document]
tamara-corbalt: @jack-ryan-nava-pbc, I agree that preactment is handling the slots and corresponding attributes in a mutually exclusive way, which results in the slot content being prioritized when both are provided.

For this storybook configuration, I followed a pattern I saw in the story file for [ds-choice](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/web-components/ds-choice/ds-choice.stories.tsx#L171-L175) I thought it might be good to demonstrate both options—using the `footer-body` attribute and providing content through the slot—so that storybook users could interact with the control directly.
But maybe this is too misleading?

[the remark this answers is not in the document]
jack-ryan-nava-pbc: Huh, I'm not sure. It doesn't look like the way the storybook is set up for ds-choice is completely accurate to the way the component logic works. I think `checked` and `unchecked` children are mutually exclusive in the React component? But I might be wrong. 

I think it's a little confusing in that Storybook makes it look like the two things are dependent on each other when in fact they aren't. Maybe @pwolfert can offer some context into the way the `ds-choice` story is wired up?

kim-cmsds (commented):
Would it be possible to provide an example or test coverage that the `backdrop-click-exits` attribute works as expected?

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.stories.tsx
```diff
@@ -0,0 +1,195 @@
...
+    'close-button-aria-label': {
+      description: 'Gives more context to screen readers on the Drawer close button.',
+      control: 'text',
+    },
+    'close-button-text': {
+      description: 'Content for the close button, typically a string or node.',
+      control: 'text',
+    },
+    'close-button-variation': {
+      description: 'Style variation for the close button.',
```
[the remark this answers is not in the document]
tamara-corbalt: I think this is a good suggestion, will update.

kim-cmsds (approved):
LGTM!

PR #3257, opened 2024-10-01
[WNMGDS-2871] Adds `ds-autocomplete` component

kim-cmsds:
## Summary
[WNMGDS-2871](https://jira.cms.gov/browse/WNMGDS-2871)
Adds a new `ds-autocomplete` web component.

## How to test

1. Try it out in storybook
2. The web component has been added to the astro example project. You'll want to uncomment out [L221](https://github.com/CMSgov/design-system/blob/d3ff4c8c1086af6680069e8013b126ddc62fe35a/examples/astro/src/pages/web-components.astro#L221) and comment out [L222:L233](https://github.com/CMSgov/design-system/blob/d3ff4c8c1086af6680069e8013b126ddc62fe35a/examples/astro/src/pages/web-components.astro#L222-L233). From there, run `yarn build` at the project root and then run `yarn dev` from `/examples/astro`
 
## Checklist


packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.stories.tsx
```diff
@@ -0,0 +1,284 @@
...
+    },
+  },
+  args: {
+    'clear-input-text': 'Clear search',
+    'clear-search-button': true,
+    'loading-message': 'Loading...',
+    'no-results-message': 'No results',
+    name: 'ds-autocomplete',
+  },
+  argTypes: {
```
jack-ryan-nava-pbc, line 26: I think you might want to add this property ```table: {
        disable: true,
      },``` to the following items: textFieldLabel, textFieldHint and children so they don't appear in the attributes table.

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.tsx
```diff
@@ -0,0 +1,109 @@
...
+}: WrapperProps) => {
+  return (
+    <Autocomplete
+      {...otherProps}
+      label={menuHeading}
+      labelId={menuHeadingId}
+      autoFocus={parseBooleanAttr(autofocus)}
+      clearSearchButton={parseBooleanAttr(clearSearchButton)}
+      id={rootId}
+      items={JSON.parse(items)}
```
tamara-corbalt, line 73: Since `items` is an optional prop, I'm wondering if there's a chance this could throw an error if items is `undefined`. Perhaps we could add a check or a fallback value to handle that scenario?

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.test.tsx
```diff
@@ -0,0 +1,434 @@
...
+    expectMenuToBeOpen();
+
+    expect(autocompleteField.value).toEqual('c');
+
+    userEvent.type(autocompleteField, '{esc}');
+
+    expectMenuToBeClosed();
+  });
+
+  // it("calls child TextField's event handlers", () => {
```
tamara-corbalt, line 380: Several unit tests are commented out. Is this intentional, or should we consider re-enabling or removing them?

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.tsx
jack-ryan-nava-pbc: Oh yeah it crashes hard if no items are provided. But not providing any items kinda defeats the purpose of a combobox. Is the idea that these items can be loaded in async in some cases so we don't want to provide a fallback @kim-cmsds ?

packages/design-system/src/components/web-components/ds-autocomplete/index.ts
```diff
@@ -0,0 +1 @@
+export * from './ds-autocomplete';
```
pwolfert, line 1: Testing out the astro example, I was expecting the `dist/web-components/bundle/ds-autocomplete.js` import to work as well—not just the main bundle. I found that if I changed this line to the following, that individual component bundle works too:

```suggestion
import './ds-autocomplete';
```

pwolfert (commented):
Everything's working really well! The only thing that is broken is the component bundle, and that's because it needs to use `import` to signal to webpack that it contains side effects.

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.test.tsx
kim-cmsds: The unit tests that are commented out correspond to how the react component's API expects a `<TextField />` component to be passed in as a child. The new, simpler web component API does not currently allow for that functionality. Instead, we're "flattening" the component and passing in a far less customizable text input here: https://github.com/CMSgov/design-system/blob/7735a395f977b258669ab8c6574f16e7669e5402/packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.tsx#L66-L77

I may spin up a follow-up ticket to see if there's a need to adjust the API to "unflatten" it and have it match the React component's API more closely.

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.stories.tsx
kim-cmsds: Thanks!

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.tsx
kim-cmsds: Good lines of thought. I'll take a look at this use case and push up changes.

packages/design-system/src/components/web-components/ds-autocomplete/index.ts
kim-cmsds: Great, thanks for looking into the _why_. I had been using the main bundle when testing locally.

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.test.tsx
kim-cmsds: https://jira.cms.gov/browse/WNMGDS-2992

tamara-corbalt (approved):
I tested it on storybook as well as the astro example, and it looks good to me!

PR #3258, opened 2024-10-02
[NO-TICKET] Adds optional bundle analysis to the build command

pwolfert:
## Summary

- Adds optional bundle analysis to the build command like `yarn build:core --analyze-bundles`

## How to test

`yarn build:core --analyze-bundles` and open the html files it generates


jack-ryan-nava-pbc:
Works like a charm. I haven't looked too deeply into this package, but is it possible for it to give us a size other than the stat/pre-tree shaken size of the packages? All the `parse` and `gzip` values report 0kb, which seems incorrect?

pwolfert:
There's a `defaultSizes` option, and I wondered if changing it to `gzip` would force it to collect that data, but it didn't work. I'm not sure why that data is missing.

pwolfert:
Oh, there's [this troubleshooting note](https://www.npmjs.com/package/webpack-bundle-analyzer#troubleshooting)

kim-cmsds (approved):
This works locally with the various child design system packages! And let's go with `--analyze-bundles`.

jack-ryan-nava-pbc (approved):

PR #3262, opened 2024-10-04
[WNMGDS-2993] Simplify NPM distribution method for web components

pwolfert:
## Summary

WNMGDS-2993

See [this RFC](https://github.com/CMSgov/design-system/discussions/3259) for a complete explanation.

- Changes `package.json` files for each of the design system packages so that the `/web-components` export path points to the bundles and so individual component bundles can be accessed from the same path like `@cmsgov/design-system/web-components/ds-alert`
  - This fixes a problem where downstream projects are unable to compile the web components due to dependencies of components that reference react packages.
- Updates the Astro web components example to import using the above paths
- Adds theme-specific pages to the Astro web component example so we can visually regression test the web components coming from the different theme packages.

## How to test

You can use the Astro example to test that the web components are successfully being exported from the package.

## Checklist


pwolfert:
Note that I disabled the new tests for now until we can add support for them in the nightly tests. They require that we build each of the child design systems, which would add a lot of time to our CI test execution.

pwolfert:
I've just reverted 71e22ae through cee62b9 and 626c951. I don't want to muddy the waters on this change by introducing what is turning out to be a fairly complicated example feature. We can't use child design systems in an example project without the requirement that those child design systems be built before the example project is built, and that's precisely what _isn't_ happening in our CI tests (in order to save time).

kim-cmsds (approved):

packages/design-system/package.json
```diff
@@ -33,9 +33,10 @@
     },
     "./web-components": {
       "types": "./dist/react-components/types/web-components/index.d.ts",
-      "import": "./dist/preact-components/esm/web-components/index.js",
-      "require": "./dist/preact-components/cjs/web-components/index.js"
+      "import": "./dist/web-components/bundle/all.js",
+      "require": "./dist/web-components/bundle/all.js"
```
jack-ryan-nava-pbc, line 37: Non-blocking: Does this change mean that there is no difference between esm and cjs anymore?

jack-ryan-nava-pbc (approved):
LGTM

packages/design-system/package.json
pwolfert: Yeah, that's right—for web components

PR #3264, opened 2024-10-07
[WNGDS-2978] Add customizable `arialabel` prop to `Tabs` component.

tamara-corbalt:
## Summary

- Added `ariaLabel` to the `Tabs` component interface and introduced `tabsAriaLabel` as an alias for `ariaLabel` specifically for use in the web component (`ds-tabs`). If neither `ariaLabel` nor the alias attribute `tabs-aria-label` is provided, a default value of "Tabs Navigation" will be used for the aria-label in the HTML markup.
- Updated the `ds-tabs` wrapper component to correctly map the `tabs-aria-label` attribute to the `ariaLabel` prop.

## How to test

1.  Run `yarn storybook`
2.  Inspect the rendered `Tabs` component in the chrome devTools elements tab to verify that the `aria-label` attribute is present in the HTML markup.
3. Inspect the rendered `ds-tabs` component in the chrome devTools elements tab. Use the controls to add a value to `tabs-aria-label` and verify that it is correctly passed to the `tablist` element as `aria-label` without duplication.

## Checklist


packages/design-system/src/components/Tabs/Tabs.tsx
```diff
@@ -31,6 +35,10 @@ export interface TabsProps {
    * Additional classes to be added to the component wrapping the tabs
    */
   tablistClassName?: string;
+  /**
+   * Helpful description of the tabs for screenreaders. An alias for `ariaLabel` specifically added to improve accessibility for the web component version of this component.
+   */
+  tabsAriaLabel?: string;
```
pwolfert, line 41: Is this prop meant only for the web component?

packages/design-system/src/components/web-components/ds-tabs/ds-tabs.tsx
```diff
@@ -23,7 +30,11 @@ const Wrapper = ({ children = [], ...otherProps }: WrapperProps) => {
     });
   }
 
-  return <Tabs {...otherProps}>{parseChildren(children)}</Tabs>;
+  return (
+    <Tabs {...otherProps} ariaLabel={tabsAriaLabel}>
+      {parseChildren(children)}
```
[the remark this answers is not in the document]
tamara-corbalt: The react API for `Tabs` expects `TabPanel` components as children, which I think means any children passed into the ds-tabs wrapper need to have a structure that aligns with what Tabs expects. If a string of JSON is passed in, it probably wouldn't be rendered correctly, as the parseChildren function specifically looks for valid ds-tab-panel elements to convert them into react components. Only child nodes that match the expected structure (like HTML or react elements that can be mapped to TabPanel) should be rendered.

packages/design-system/src/components/Tabs/Tabs.tsx
tamara-corbalt: Yes, as far as I'm aware, use in the web component is its only intended use case for now, and I can't think of other contexts where it would apply. Are there other possible scenarios where it might come in handy?

pwolfert: In that case, should it be removed from this file?

jack-ryan-nava-pbc: Yeah @tamara-corbalt this should get removed, since the attribute value on the web component is being added to the prop value [here](https://github.com/CMSgov/design-system/pull/3264/files#diff-01521597bdd1f8032256b3e0f2d9e6c48812f08b37906c7f321beaf507013cd8R26)

tamara-corbalt: @pwolfert, @jack-ryan-nava-pbc, thanks for the call out :) Removed it

kim-cmsds (approved):

packages/design-system/src/components/Tabs/Tabs.test.tsx
```diff
@@ -53,6 +53,21 @@ describe('Tabs', function () {
     expect(firstPanel.getAttribute('aria-labelledby')).toEqual(firstTab.id);
   });
 
+  it('should render a default aria-label when none is provided', () => {
```
jack-ryan-nava-pbc, line 56: I don't think we want this to render a default value when nothing is provided, because it is possible that another element could label the tablist. In that case, a user would want to use the `aria-labelledby` attribute and omit the `aria-label`. A good example of this use case is [here](https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-automatic/#ex_label). You can see in the source code that the h3 element is labelling the tablist and that an aria label is unnecessary.

packages/design-system/src/components/web-components/ds-tabs/ds-tabs.tsx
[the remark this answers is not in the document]
jack-ryan-nava-pbc: Ok cool. Thanks that makes sense. I think in that case you'll want to omit the `children` attribute from the `ds-tabs` web component, because we don't want that to be available for people to apply strings to.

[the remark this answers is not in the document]
tamara-corbalt: :white_check_mark:

packages/design-system/src/components/Tabs/Tabs.test.tsx
tamara-corbalt: @jack-ryan-nava-pbc Thanks for pointing this out! I’ve removed the default value, so aria-label will only be applied if explicitly provided via the prop. Do you think our [accessibility guidance for the Tabs component](https://design.cms.gov/components/tabs/?theme=core#accessibility) in the documentation could use a bit more nuance?

jack-ryan-nava-pbc: I think we should update the accessibility guidance for the Tabs component, maybe to highlight that the aria-label is  optional if another element is describing the content via `aria-describedby`. But we can revisit this in another ticket.

packages/design-system/src/components/Tabs/Tabs.tsx
```diff
@@ -96,6 +100,7 @@ export const Tabs = (props: TabsProps) => {
   const [internalSelectedId, setSelectedId] = useState(initialSelectedId);
   const isControlled = props.selectedId !== undefined;
   const selectedId = isControlled ? props.selectedId : internalSelectedId;
+  const ariaProps = props.ariaLabel ? { 'aria-label': props.ariaLabel } : {};
```
jack-ryan-nava-pbc, line 103: Nice!

jack-ryan-nava-pbc (approved):

PR #3267, opened 2024-10-09
[WNMGDS-2999] Add a shared web components example template and new example project

pwolfert:
## Summary

WNMGDS-2999

### Problem

While working on https://github.com/CMSgov/design-system/pull/3262, I wanted to be able to test that web components could be successfully imported from child design system packages, so I tried to add that functionality to one of our example projects. The problem with importing child design systems from example projects, however, was that our CI tests don't actually build those child design systems. When the CI tests then tried to build the example projects, they would fail. We could start building the child design systems in the CI, but that would add two minutes or so to the time it takes to run our CI tests. The most straightforward solution would have been to create a separate example project that builds the child design systems, but that would make yet another copy of the web component examples for the team to maintain long-term.

Thinking more about that problem of maintaining [web component examples like this](https://github.com/CMSgov/design-system/blob/main/examples/astro/src/pages/web-components.astro "Follow link"), we actually already have multiple projects that show web components, and some of them are out of date. When an engineer adds a web component or updates it, they have to add storybook examples, unit tests, VRT references, and create examples in multiple web component example projects. It's hard to keep them in sync. What if we had a single source for web component example markup that could be re-used in different example projects?

### Requirements

A solution should...

1. Maintain the integrity of the integration test so that we continue to test how web components would be used in a real project
    - A counter-example to this would be creating a library with a reusable JSX component that rendered web components, because then we'd be testing how that library bundled the web components and not how the different projects natively handled web components. Even if we imported an HTML file, the projects would then have to import that file somehow and would have an unnecessary layer of abstraction.
2. Keep the example projects useful to new users so they can quickly see how to use various assets and components without unnecessary complexity
3. Lean into automation
4. Fit into our existing testing pipelines

### Solution

We brainstormed multiple solutions, and we chose one based on simplicity and pragmatism (which means it isn't the most attractive). It works as follows:

1. Define a shared _web-components-template.html file
2. Update the examples that use web components to include two HTML comments like
    ```
    <!-- START WEB COMPONENTS -->
    <!-- END WEB COMPONENTS -->
    ```
3. Write a script that will read the template file and copy its contents into those example project files to replace everything between those comments like this:
    ```
    <!-- START WEB COMPONENTS -->
    <ds-skip-nav href="#main"></ds-skip-nav>
    <ds-usa-banner></ds-usa-banner>
    ... and other web component stuff
    <!-- END WEB COMPONENTS -->
    ```
4. Whenever the script runs and updates these files, we then check the updates into version control as we normally would. That means the web component example code will be available to anyone who goes to one of the examples projects to see how to use them without facing the added complexity of imports and having to reference another external file.

We're then free to add more example projects that use this template, including one that will import child design systems but which will not be built by our CI job.

## How to test

1. Make a change to the template file
2. Run `yarn apply-example-templates` to see the relevant example files updated
3. Run `yarn test:browser:examples` and watch the tests fail (maybe opening `playwright-report/index.html` to see the differences)

Note that this will fail on CI until the necessary update has been made (see our Jenkins script repo for the pull request).

## Checklist


pwolfert:
> So I am not seeing any errors

@jack-ryan-nava-pbc, did you make changes to the template file that would result in visual differences?

tamara-corbalt (approved):
It's working as expected for me! I saw the expected changes staged, and the browser test failures were as expected. Pending resolution of the CI failures, approve.

pwolfert:
Here's a question, @tamara-corbalt and @jack-ryan-nava-pbc: To keep things up to date, should I run the new `yarn apply-example-templates` command in `yarn build:examples`? If the template hasn't changed, it doesn't harm anything, and if things did change, you'd see the changes in version control.

scripts/utils.ts
```diff
@@ -24,6 +24,13 @@ export function shI(
   return spawnSync(command, args, { stdio: 'inherit', ...config });
 }
 
+/**
+ * Just like the `sh` function but prints all stdio output (verbose)
+ */
+export function shV(command: string) {
```
jack-ryan-nava-pbc, line 30: Is this change going to break other scripts that are calling shVerbose?

pwolfert: There was only one script that defined and used `shVerbose`, so I moved it to here and renamed it `shV` so it could be shared

jack-ryan-nava-pbc:
> Here's a question, @tamara-corbalt and @jack-ryan-nava-pbc: To keep things up to date, should I run the new `yarn apply-example-templates` command in `yarn build:examples`? If the template hasn't changed, it doesn't harm anything, and if things did change, you'd see the changes in version control.

The only reason I can think not to include it is that maybe it's making the build:examples script do too much now? But applying the example templates does seem like something you'd want to do when you build. 

I guess maybe my other thought is would it make sense then to have the apply-example-templates script broken out on it's own at all if build:templates called it each time? I guess it would, since sometimes you might just want to apply the templates. 

I think its fine to include it. It makes sense to me to add it.

tamara-corbalt:
I'm not seeing a reason not to add it to the build script, but maybe we can provide a configuration flag that allows developers to opt-in or out of automatically applying templates during the build?

> Here's a question, @tamara-corbalt and @jack-ryan-nava-pbc: To keep things up to date, should I run the new `yarn apply-example-templates` command in `yarn build:examples`? If the template hasn't changed, it doesn't harm anything, and if things did change, you'd see the changes in version control.

pwolfert:
> maybe we can provide a configuration flag

@tamara-corbalt, could you talk more about the use cases for that? In such cases, would commenting out the line that applies the template be sufficient? Or do you think it would happen often enough to warrant extra code to handle that case?

kim-cmsds (approved):

jack-ryan-nava-pbc (approved):

PR #3270, opened 2024-10-10
[WNMGDS-2921] Implement Filter Chip Web Component

JackRyan1989:
## Summary

- [Ticket](https://jira.cms.gov/browse/WNMGDS-2921)
- Addition of `ds-filter-chip` web component
- Stories and Unit tests added

## How to test

1. Pull down this branch locally
2. Run `yarn build`
3. Run `yarn test:unit:wc` to confirm that unit tests pass
4. Run `yarn storybook` & confirm that all attributes are editable in the single filter chip example. I've disabled the props table in the multichip example because the controls didn't do anything.
5. Also in Storybook confirm that a `ds-delete` custom event is fired when clicking the chip
6. I've not added an example in Astro, but have in the CDN Web Components. You can confirm the functionality there.

## Checklist


jack-ryan-nava-pbc:
Ugh. I created this while logged into my old Github account. Apparently that one still has write access?

packages/design-system/src/components/web-components/ds-filter-chip/ds-filter-chip.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+interface WrapperProps extends Omit<FilterChipProps, 'useAlternateIcon'> {
+  useAlternateIcon?: string;
+  clearLabel?: string;
+}
+
+const Wrapper = ({ clearLabel, label, useAlternateIcon, ...otherProps }: WrapperProps) => (
+  <FilterChip
+    ariaClearLabel={clearLabel ?? ''}
+    label={label}
```
tamara-corbalt, line 38: I noticed that the `label` prop is explicitly wired up here, but since it's just a string and doesn't seem to require any special handling, would it make sense to pass it through `otherProps` instead? I'm wondering if there's a specific reason for this approach.

packages/design-system/src/components/web-components/ds-filter-chip/ds-filter-chip.tsx
```diff
@@ -0,0 +1,54 @@
...
+/* eslint-enable */
+
+interface WrapperProps extends Omit<FilterChipProps, 'useAlternateIcon'> {
+  useAlternateIcon?: string;
+  clearLabel?: string;
+}
+
+const Wrapper = ({ clearLabel, label, useAlternateIcon, ...otherProps }: WrapperProps) => (
+  <FilterChip
+    ariaClearLabel={clearLabel ?? ''}
```
tamara-corbalt, line 37: I'm curious about the semantics of passing an empty string here. Is there a specific accessibility concern we're addressing?

packages/design-system/src/components/web-components/ds-filter-chip/ds-filter-chip.tsx
jack-ryan-nava-pbc: I think this was a typescript thing? The only required prop for FilterChip is a label, and the compiler wasn't picking up on whether it was present in the otherprops object or not, so I destructured it and explicitly passed it in to appease the typescript gods.

packages/design-system/src/components/web-components/ds-filter-chip/ds-filter-chip.tsx
jack-ryan-nava-pbc: So the `ariaClearLabel` prop/attribute isn't _actually_ an aria attribute. It's a string that gets passed into a span element. Because the attributes for web components have to be kebab case (aria-clear-label instead of ariaClearLabel), our a11y tests correctly point out that `aria-clear-label` isn't a real aria label. So I made up a new attribute for the web component that drops the `aria-` prefix and then maps onto the `ariaClearLabel` prop when we pass it into the Preact component. Since it's an optional prop and not written onto an actual `aria` attribute anywhere, I figured it would be ok to pass an empty string to it.

tamara-corbalt (commented):
Looks great so far! 
Noticing a lot of updated snapshots here. Are they the same ones as in [pr-3271](https://github.com/CMSgov/design-system/pull/3271)? If so, would it be cleaner to revert these changes and merge with main once 3271 is merged?"

packages/design-system/src/components/web-components/ds-filter-chip/ds-filter-chip.stories.tsx
```diff
@@ -0,0 +1,153 @@
...
+    },
+    size: {
+      description:
+        'Sets the size of the chip to larger version. The only acceptable value is "big".',
+      control: 'text',
+    },
+    'use-alternate-icon': {
+      description: 'Use alternate thinner close icon in place of standard.',
+      control: 'boolean',
+    },
```
pwolfert, line 55: Ask me about this in the office hours session if there's down time

jack-ryan-nava-pbc: Thanks for bringing this up. Closing this loop:
We are creating a ticket to do the discovery work necessary to unify close icons across the design system. This work will happen outside of this pull request.

kim-cmsds (approved):
The filter chip work looks good to me! I would spend a little of time verifying that the snapshot changes are all relevant to work done here to @tamara-corbalt's point.

tamara-corbalt (approved):
Aside from merge conflicts that need to be resolved, looks great to me!

PR #3279, opened 2024-10-14
[WNMGDS-2977] Create a dedicated storybook file for `ds-tab-panel`

tamara-corbalt:
## Summary

- Adds a dedicated storybook file for `ds-tab-panel`.
- Adds unit tests for `ds-tab-panel`.

## How to test

- Inspect on storybook: `yarn storybook`

## Checklist


jack-ryan-nava-pbc (approved):
This looks great! My only very minor suggestion might be to say something about how tab panels aren't meant to be used on their own, but as part of the tabs web component? That might be clear from the top level link going to the Tabs documentation though. What are your thoughts?

tamara-corbalt:
> This looks great! My only very minor suggestion might be to say something about how tab panels aren't meant to be used on their own, but as part of the tabs web component? That might be clear from the top level link going to the Tabs documentation though. What are your thoughts?

@jack-ryan-nava-pbc I think you're absolutely right. I'll add something.

PR #3285, opened 2024-10-17
[WNMGDS-2920] Adds `ds-review` web component

tamara-corbalt:
## Summary

- Adds `ds-review` web component: [Jira Ticket](https://jira.cms.gov/browse/WNMGDS-2920).

## How to test

- Inspect `ds-review` in storybook: `yarn storybook`
- Also, in `examples/astro`, run `yarn dev`

## Checklist


examples/astro/src/pages/web-components.astro
```diff
@@ -231,6 +231,38 @@ import Layout from '../layouts/Layout.astro';
             );
         });
       </script>
+      <h2 class="ds-text-heading--2xl">Review example</h2>
+      <ds-review heading="Heading for the Review Component">
```
pwolfert, line 235: This example is working great, but it looks a little squished. Maybe we should rearrange the example page a bit so the review component isn't inside a `measure--base`

pwolfert (approved):
The storybook docs and everything looks great!

examples/astro/src/pages/web-components.astro
tamara-corbalt: @pwolfert Thanks for catching this! Moved the review example outside of the `measure--base`, so it should be un-squished now. :)

packages/design-system/src/components/web-components/ds-review/ds-review.tsx
```diff
@@ -0,0 +1,31 @@
...
+const Wrapper = ({ children, ...otherProps }) => {
+  return <Review {...otherProps}>{children}</Review>;
+};
+
+/* eslint-disable @typescript-eslint/no-namespace */
+declare global {
+  namespace JSX {
+    interface IntrinsicElements {
+      'ds-review': JSX.IntrinsicElements['div'] & {
+        [K in (typeof attributes)[number]]?: string;
```
jack-ryan-nava-pbc, line 22: This is a nit and an understanding check for myself:

Do you need this name spacing here? My understanding of this (which is likely incomplete/wrong) is that we're saying that the `ds-review` element is a div and that any attributes that are number types should be string. Is that correct? And if it is correct, I don't know that you need it since none of the Review component props take numbers?

jack-ryan-nava-pbc (approved):
Looks great! Approving and just noting that a link needs to be updated. I have a question but it's non-blocking.

packages/design-system/src/components/web-components/ds-review/ds-review.tsx
pwolfert: The number is actually referring to a numerical index in the `attributes` array as it uses it as an array of keys and basically iterates over it.

You know, at this point now that there are almost all of the web components that'll ever exist, we know the typing needs for all the component attributes and could probably create a generic type utility. It might not be worth the abstraction, but I was thinking something like this:

```ts
type WebComponent<Tag extends string, Attributes extends readonly string[], T = HTMLElement> = {
  [K in Tag]: React.DetailedHTMLProps<
    React.HTMLAttributes<T> & {
      [K in (Attributes)[number]]?: string;
    },
    T
  >;
};

/* eslint-disable @typescript-eslint/no-namespace */
declare global {
  namespace JSX {
    interface IntrinsicElements extends WebComponent<'ds-filter-chip', typeof attributes> {}
  }
}
/* eslint-enable */
```

Theoretically you could pass `HTMLDivElement` to the third optional parameter of the generic type for components with a div root.

pwolfert: It really doesn't save much boilerplate, though, or make it easier to read.

jack-ryan-nava-pbc: Thanks for the clarification @pwolfert !

kim-cmsds (approved):

PR #3286, opened 2024-10-17
[WNMGDS-3007] Fix arrow behavior for disabled tabs

tamara-corbalt:
## Summary
Updates the arrow key navigation logic in `Tabs` components so that the content of disabled tabs is no longer reachable or visible using arrow keys.

[Jira Ticket](https://jira.cms.gov/browse/WNMGDS-3007) 

## How to test

- Run `yarn storybook`
- Inspect disabled stories for `Tabs` and/or `ds-tabs`.
- Use `tab` key to initialize focus, then use left and right arrow keys for navigation.
- Repeat the above steps while using a screen reader to confirm that disabled tabs are skipped and their content is not announced.

## Checklist


pwolfert (approved):
It does what you say it should 👍 

I'm not really sure why, but when I open up the web rotor for VoiceOver on this example, the tab links are not in the links menu. I was wondering if that might be another way to discover the "disabled" tab if it were only an `aria-disabled` tab. This comment shouldn't block the PR. I'm just making observations.

jack-ryan-nava-pbc (approved):
Excellent work! Works like a charm

tamara-corbalt:
> It does what you say it should 👍
> 
> I'm not really sure why, but when I open up the web rotor for VoiceOver on this example, the tab links are not in the links menu. I was wondering if that might be another way to discover the "disabled" tab if it were only an `aria-disabled` tab. This comment shouldn't block the PR. I'm just making observations.

That's a good call out—I hadn't thought to use the web rotor. Initially, I thought it might be because the `TabPanel` components in the storybook example don’t have a `tabHref` prop assigned. I tried updating that, but the links still didn't appear in the list. Do you think this is something we should explore further?

pwolfert:
> Do you think this is something we should explore further?

I haven't followed the conversations you've had around this, but I think it's good to come to some decision that can be articulated in words about whether screen-reader users should know about disabled tabs and, if yes, what the exact mechanism is that they're expected to discover it. So in my musings above, if the design system's stance was that they should be able to hear about the tab but couldn't tab to it, would that mean they should be able to see the disabled tab in a list of page links in the web rotor? If the answer is "no" to the disabled tabs being discoverable, that's fine too.

To your original question, I think it's out of scope for this PR.

kim-cmsds:
@malloryiden do you have time to weigh in on the points that Patrick brought up [here](https://github.com/CMSgov/design-system/pull/3286#issuecomment-2422898120)?

kim-cmsds (approved):
These changes works as described.

I think we should spin up a follow-up ticket to decide whether a disabled tab ought to be `disabled` or `aria-disabled` based on the discussion in the related [JIRA ticket](https://jira.cms.gov/browse/WNMGDS-2976?focusedId=3799867&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-3799867).

malloryiden:
Hi! This is an interesting topic!

The UX should be consistent for any method in which the user chooses to interact with the page. So that could be visual, tabbing using VoiceOver, using the VoiceOver rotor menu, etc... The rotor _should_ call out disabled tabs (and mention that they are disabled) so that folx who use the rotor still know that it exists. 

In [this documentation](https://a11y-101.com/development/aria-disabled#:~:text=VoiceOver's%20rotor%20will%20list%20all%20buttons%2C%20even%20if%20they%20are,be%20named%20'dimmed%20button'.), it mentions how using aria-disabled will list disabled buttons (which will be named "dimmed"), we should apply the same principle to other disabled interactive elements like tabs, links, etc... 

After reading it, it seems this should call this out already by just using the aria-disabled tag. Perhaps there is a follow up ticket needed for this to figure out why this isn't working for tabs.

PR #3296, opened 2024-10-25
[WNMGDS-3026] Add analytics support to web components

pwolfert:
WNMGDS-3026

## Problem

First, let me explain why analytics wasn't working in web components:

1. In order to provide a way for users to listen to web component events, the `define` function passes event-proxying functions to all event-handler React props (e.g. `onPageChange` or `onSelectAll`).
2. Whether a user does add an event listener for those proxied events or not, the underlying React component always receives a function for all of its event-handler props.
3. All our analytics code in our React components assumes that if you pass an `onAnalyticsEvent` function, you want to use that one in place of the default `config().defaultAnalyticsFunction` that passes events to Tealium through `window.utag.link`.
4. Therefore all web components were using the custom `onAnalyticsEvent` handler, which just happened to be an event-proxying function and wasn't doing anything with the events.

Because of this, the only way to get analytics on a web component was if you handled it yourself with the custom `ds-analytics-event` event. Because we didn't know what our long-term solution for this was, we didn't document this behavior.

So how do we get the web components to have a default analytics behavior that matches the React components?

## (Partial) Solution

My solution right now involves using the familiar concept of an `event.preventDefault()` function. When an analytics event is triggered, if you don't have an event listener for the `ds-analytics-event` event that calls `preventDefault()`, then the component will assume you want the default analytics behavior and will behave exactly as the React components do in that scenario (respecting your analytics settings in the global and component configs).

This solution doesn't apply perfectly in the same way to `ds-alert`, however. The problem with the alert analytics event is that it fires right on first render. If teams want to override a specific component's analytics event handler (which normally comes from `config().defaultAnalyticsFunction`), they have to call `addEventListener` on the element instance and add their event listener, but at that point the component has already done its first render and fired off the analytics event. It's too late. Therefore in order to intercept that event on the `ds-alert`, users will actually have to apply an `analytics="false"` attribute to the particular alert instance and then trigger it themselves. Fortunately I think this need would be very rare. Teams are always capable of supplying their own `config({ defaultAnalyticsFunction })`, and they can opt in or out of analytics at the global level or individual component instance level. They have a lot of options. We just need to clearly describe this behavior in our documentation for `ds-alert`.

## How to test

1. `yarn test:wc`
2. Try it out in Storybook after turning analytics on


packages/design-system/src/components/web-components/analyticsTesting.ts
```diff
@@ -0,0 +1,78 @@
+import { UtagContainer } from '../analytics/index';
+import { config } from '../config';
+
+export function testAnalyticsEvents({ tagName, renderComponent, configField, triggerAction }) {
```
pwolfert, line 4: I don't know if this is actually going to be a good idea, but I was just playing around with different ways of reducing the code duplication when a lot of the same tests need to be performed for all the components that have analytics.

pwolfert: I figured you could call it multiple times for different permutations of options (like how the `ds-button` could be a `<button>` or `<a>`)

pwolfert:
@kim-cmsds, @jack-ryan-nava-pbc, and @tamara-corbalt, since event-listeners can't be added fast enough to intercept the on-render analytics events like Alert has (it's the only one), what if we just advised teams that if they want to implement custom analytics event listeners for the Alert render event, they pass `analytics="false"` and then implement their own entirely. Otherwise, they can let it call the default analytics function, which they can modify be calling `config()`? I think they actually have _plenty_ of customization options without creating 1-1 parity with the React components on this one.

jack-ryan-nava-pbc:
Re [this](https://github.com/CMSgov/design-system/pull/3296#issuecomment-2442013025)

One thing I was wondering was if it is possible for team to hook into the connectedCallback and supply an analytics function there?

pwolfert:
> I was wondering was if it is possible for team to hook into the connectedCallback

@jack-ryan-nava-pbc, that's a good thought, and it'd probably work. What might the API look like for that? My gut reaction is that it would require a new API that is unfamiliar to our users (rather than using `.addEventListener`) and might end up being more work for you to support and maintain for the gain you'd get in feature parity.

tamara-corbalt:
I think it makes sense and sounds reasonable to allow teams consuming Alert to have an extra configuration step of setting `analytics="false"` and calling `config()` directly for custom analytics behavior. Are there any risks of conflicting or duplicate analytics events if teams forget to disable the default analytics? Also want to acknowledge that I might not be fully understanding all aspects here.

> @kim-cmsds, @jack-ryan-nava-pbc, and @tamara-corbalt, since event-listeners can't be added fast enough to intercept the on-render analytics events like Alert has (it's the only one), what if we just advised teams that if they want to implement custom analytics event listeners for the Alert render event, they pass `analytics="false"` and then implement their own entirely. 
>

pwolfert:
@tamara-corbalt, you bring up a good point about the original solution: There _is_ a risk of duplicate analytics events being reported if they add their own listener (or multiple listeners!) and fail to call `event.preventDefault()`.

It sounds to me like you're understanding is complete, and I'm noting that you're in support of the extra step for people wanting to customize the analytics behavior of alerts 👍

packages/design-system/src/components/web-components/ds-button/ds-button.test.tsx
```diff
@@ -123,52 +120,78 @@ describe('Button', () => {
       (window as any as UtagContainer).utag = {
         link: tealiumMock,
       };
+      jest.useFakeTimers();
```
kim-cmsds, line 123: Is the goal here to import the shared functions defined in `analyticsTesting.ts`?

kim-cmsds:
> what if we just advised teams that if they want to implement custom analytics event listeners for the Alert render event, they pass analytics="false" and then implement their own entirely. Otherwise, they can let it call the default analytics function, which they can modify be calling config()?

@pwolfert I'm good with this direction as long as we have examples in addition to the documentation.

jack-ryan-nava-pbc:
@pwolfert In just spending some time trying to replicate what is happening in the `useAlertAnalytics` hook in the `onConnectedCallback` I have some thoughts.

To be clear about what I was trying to do:
1. Read attributes on the custom element, and if analytics = 'true' then fire off config().defaultAnalyticsFunction
2. This is an implementation that would automatically fire off the onAnalyticsEvent on load, which is pretty much what yours is doing, but wouldn't require the user to fire an event.preventDefault, they could just set analytics to false
3.  The main issue I'm seeing is replicating the event content for the analytics call. The `useAlertAnalytic` hook constructs the following object:
```
event_name: 'alert_impression',
        event_extension: eventExtensionText,
        heading: eventHeadingText,
        type: variation,
```
We need to manually do this in the `define` function as far as I can tell, which seems not optimal. 

Reconstructing the event content is the main roadblock I'm running into. Also, let me know if I'm totally missing the point.

packages/design-system/src/components/web-components/ds-button/ds-button.test.tsx
pwolfert: Yeah, I was experimenting with those shared functions, and I think I'll end up changing the way they're structured so they're easier to reuse. Whatever the result of that, I'll test them consistently between components

pwolfert:
@jack-ryan-nava-pbc, the components don't have a problem calling the default function at the moment. The value of the `analytics` attribute correctly gets passed to the React component an honored by calling our special handler function or not. It's the ability for our users to pass in an alternative to it in time that is the issue. I was thinking you meant a user could call something to tap into the connected callback and set the event handler more directly before the first render happens--since we can't pass in functions through attribute strings. The component doesn't actually need an `analytics` attribute value of true to have analytics if the analytics for that component are turned on globally through the `config` function.

pwolfert:
This should all be working now and ready for review

tamara-corbalt (approved):
Ran wc unit tests locally and they pass. Also, appears to work in storybook :raised_hands:

PR #3301, opened 2024-10-29
NO-TICKET: Expose 'informational' as part of the AlertVariation type

Josh68:
## Summary

When using Design System types, directly, in a project, the AlertVariation type is missing the string for the default Alert style, which is "informational," according to the docs. This update just adds that string to the union, but excludes it from the prop typing, since it's the default.

We are passing this detail via Adobe Analytics, so would like to have the "informational" typing exposed.

## How to test

There is no impact to the use of the component. When built, a project that imports `AlertVariation` from Design System types should see "informational" as one of the options (in our case, for analytics purposes).

## Checklist

- [ ] If visual regression image references have been changed, design **MUST** be assigned to review. In this instance, **designer approval is a requirement** before the PR can be merged.


kim-cmsds:
Jenkins, not a wolf

jack-ryan-nava-pbc:
@kim-cmsds @tamara-corbalt This seems like a perfectly fine thing to include. Probably we want to spin up another PR that replicates the changes here and includes the appropriate labels and milestones. What do you too think? 

Thanks for this suggestion @Josh68. If I understand this change correctly it seems that the benefit of this inclusion will be available to Typescript for hints?

Josh68:
> Thanks for this suggestion @Josh68. If I understand this change correctly it seems that the benefit of this inclusion will be available to Typescript for hints?

Correct, @jack-ryan-nava-pbc. Wherever possible (and hopefully always), we want to use types for design system components from your repo. In this case, for sending alert interaction events to analytics, we want to send the `variation`. While `informational` is a valid `variation`, it's the default and isn't exposed publicly, It should be exposed.

jack-ryan-nava-pbc:
I think we're good to merge this in the New Year @kim-cmsds @tamara-corbalt 

Thanks @Josh68

Josh68:
Updated with a merge from `main`. I hope this can be merged and closed, thanks

kim-cmsds (approved):

PR #3302, opened 2024-10-29
[WNMGDS-2922] Implement ds-modal-dialog component

jack-ryan-nava-pbc:
## Summary

- [Ticket](https://jira.cms.gov/browse/WNMGDS-2922)

## How to test

1. Run `yarn test:unit:wc` and confirm there are no failures
2. Checkout the Storybook stories and confirm the functionality matches the React component
3. Checkout the Astro example and make sure it's reasonable/works in the way you expect.
4. Test with MacOS VoiceOver and confirm that it is read the same way as the React component.

## Checklist


packages/design-system/src/components/NativeDialog/NativeDialog.tsx
```diff
@@ -116,7 +116,7 @@ export const NativeDialog = ({
         rect.left <= event.clientX &&
         event.clientX <= rect.left + rect.width;
       if (!isInDialog) {
-        exit();
+        exit(event);
```
jack-ryan-nava-pbc, line 119: backdropClickExits was throwing an error in the define method because the event.target did not exist. Hence passing in the event here solved the issue.

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
```diff
@@ -0,0 +1,268 @@
...
+        },
+      },
+    },
+  },
+  decorators: [webComponentDecorator],
+};
+export default meta;
+
+const LongContent = ({ args }) => {
+  useEffect(() => {
```
jack-ryan-nava-pbc, line 106: Is there a more concise way of doing this? I tried creating the LongContent section as it's own component in this file and then passing in the Template content as children props. This didn't seem to work. I'd prefer to not have to duplicate this useEffect if possible.

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.test.tsx
```diff
@@ -0,0 +1,242 @@
...
+
+    modal.setAttribute('is-open', 'true');
+    rerenderModalDialog({});
+    expect(modal.getAttribute('is-open')).toBe('true');
+    expect((screen.getByRole('dialog') as HTMLDialogElement).open).toBe(true);
+  });
+
+  // Skipping these for the same reason outlined in the ds-alert.test.tsx file.
+  // See lines 123 - 129.
+  describe.skip('Analytics event tracking', () => {
```
jack-ryan-nava-pbc, line 147: Skipping analytics until we figure that one out

examples/astro/src/pages/web-components.astro
```diff
@@ -281,9 +282,10 @@ import Layout from '../layouts/Layout.astro';
 <script src="@cmsgov/design-system/web-components/ds-choice"></script>
 <script src="@cmsgov/design-system/web-components/ds-dropdown"></script>
 <script src="@cmsgov/design-system/web-components/ds-filter-chip"></script>
+<script src="@cmsgov/design-system/web-components/ds-modal-dialog"></script>
```
tamara-corbalt, line 285: Hey! Looks great—I tested it in storybook, and it appears to work there. I noticed, though, that only the ds-modal-dialog script was imported here. To fully test this component in astro, would we need additional markup and event handling to open and close the modal?

jack-ryan-nava-pbc: Thanks for catching @tamara-corbalt 
Should be fixed in latest commit.

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
pwolfert: Since `Template` is just a component function, could we render it like a component inside your `PreventScrollExample` story like this?

```ts
export const PreventScrollExample = {
  render: function Component(args) {
    return (
      <div>
        all the long content before the template
        <Template {...args} />
        all the long content after the button
      </div>
    );
  },
```

kim-cmsds: You could put this `useEffect` logic into its own decorator:

```
const ModalDecorator = (Story: React.FC) => {
  useEffect(() => {
    const button = document.querySelector('ds-button');
    const modal = document.querySelector('ds-modal-dialog');
    const toggleModal = (event): void => {
      const targetType = event.target.nodeName ?? '';
      if (targetType === 'DS-BUTTON') {
        action('ds-click')(event);
      } else {
        action('ds-exit')(event);
      }
      const isOpen = modal.getAttribute('is-open');
      if (isOpen === 'true') {
        modal.setAttribute('is-open', 'false');
      } else {
        modal.setAttribute('is-open', 'true');
      }
    };
    button.addEventListener('ds-click', toggleModal);
    modal.addEventListener('ds-exit', toggleModal);
    return () => {
      button.removeEventListener('ds-click', toggleModal);
      modal.removeEventListener('ds-exit', toggleModal);
    };
  });

  return (
    <Story />
  )
}
```

You would add the following to each story object `decorators: [ModalDecorator]`

kim-cmsds (commented):
This is looking really great. It looks like the dialog component has an interaction test set up, what do you think about adding one for the web component version?

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.tsx
```diff
@@ -0,0 +1,62 @@
...
+  ...otherProps
+}: WrapperProps) => (
+  <Dialog
+    id={rootId}
+    alert={parseBooleanAttr(alert)}
+    ariaCloseLabel={dialogCloseLabel}
+    isOpen={parseBooleanAttr(isOpen)}
+    size={isAcceptableSize(size) ? size : null}
+    {...otherProps}
+    analytics={analytics && Boolean(JSON.parse(analytics))}
```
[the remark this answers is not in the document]
jack-ryan-nava-pbc: I think I just copied from the `ds-alert.tsx` file and didn't think about it?

examples/astro/src/pages/web-components.astro
```diff
@@ -267,6 +268,62 @@ import Layout from '../layouts/Layout.astro';
...
+        <div>
+          <p>
+            Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed accumsan diam vitae metus
+            lacinia, eget tempor purus placerat.
+          </p>
+          <div slot="heading">
+            <h2>This is the modal heading.</h2>
+          </div>
+          <div slot="actions">
+            <ds-button id="submit">Submit (and close the dialog)</ds-button>
```
kim-cmsds, line 284: I tried to close this example using the keyboard and was not able to do so after tabbing to the "Open Modal" button, pressing "enter" and then tabbing this submit button into focus and then pressing "spacebar" or "enter. Do you know why that's the case @jack-ryan-nava-pbc?

tamara-corbalt (approved):
Tested locally in Storybook and Astro, as well as with VoiceOver. It appears to be read in the same way as the React component. :raised_hands:

pwolfert:
Found out that the backdrop-click-exits code [is picking up on the close event click](https://github.com/CMSgov/design-system/commit/92f5e15f3f8cfd39be34f541180722b169ec1d3f) and calling the `onExit` prop again. That's where the second firing of the `ds-exit` event seems to be coming from. Calling `stopPropagation` makes it so the `ds-exit` event only fires once on those two stories that have `backdrop-click-exits` set to `true`.

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
```diff
@@ -124,27 +122,23 @@ const Template = (args) => {
...
 
     return () => {
-      modal?.removeEventListener('ds-exit', (event) => action('ds-exit')(event));
-      toggleButton?.removeEventListener('click', toggleModal);
-      form?.removeEventListener('submit', toggleModal);
-      closeButton?.removeEventListener('click', toggleModal);
+      modal?.removeEventListener('ds-exit', (event) => {
+        action('ds-exit')(event);
+        modal?.setAttribute('is-open', 'false');
+      });
```
pwolfert, line 140: I did this in my troubleshooting branch, but I want to flag it here because these anonymous functions are two different objects, so it won't actually remove the original event listener added on line 130.

pwolfert:
I just noticed that the backdrop click exiting doesn't actually work. That is, you can't click the backdrop to exit.

jack-ryan-nava-pbc:
> I just noticed that the backdrop click exiting doesn't actually work. That is, you can't click the backdrop to exit.

This should be fixed @pwolfert thanks for catching!

examples/_web-components-template.html
```diff
@@ -262,5 +262,57 @@ <h2 class="ds-text-heading--2xl">Review example</h2>
...
+
+        // You can also open and close via the underlying dialog element's close() and showModal()
+        // methods.
+        //   const dialog = document.querySelector('dialog') as HTMLDialogElement;
+        //   if (dialog.open) {
+        //     dialog.close();
+        //   } else {
+        //     dialog.showModal();
+        //   }
+        // };
```
pwolfert, line 305: Is this still something you want to have in here?

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
```diff
@@ -0,0 +1,232 @@
...
+        <div slot="heading">
+          <h2>This is the modal heading.</h2>
+        </div>
+        <div slot="actions">
+          <form method="dialog">
+            <ds-button value="rstBtn">Reset</ds-button>
+            <ds-button type="submit" value="sbmBtn">
+              Confirm
+            </ds-button>
+          </form>
```
pwolfert, line 29: Would it be a good idea to put some kind of long-form explanation on the doc page explaining how to set this up?

tests/browser/snapshots/stories-interaction/ds-modal-dialog--open--healthcare--Mobile-Chrome.png
pwolfert, line 1: Should there be a gap between the buttons like our React example has? And maybe swap out the lorem ipsum text?

Also I'm not sure if people will understand that "Reset" is not supposed to close the dialog without context. If we changed this whole example to resemble the React story where both buttons close it, we could explain in the documentation that if there's any button you want to _not_ close the dialog, you just don't make it a submit button.

pwolfert (approved):
It's looking great! A few suggestions below about documentation

examples/_web-components-template.html
jack-ryan-nava-pbc: It's probably not necessary

tests/browser/snapshots/examples/cdn-web-components/cdn-web-components--chromium.png
pwolfert, line 1: Take a look at the bottom of this screenshot. For some reason the dialog is being inlined in the page!

pwolfert (changes requested):

tests/browser/snapshots/examples/cdn-web-components/cdn-web-components--chromium.png
jack-ryan-nava-pbc: Looks like there's some other weird things going on with the Autocomplete and Filter Chip examples. This seems to only be happening in the CDN.

jack-ryan-nava-pbc: Ok yeah what's going on here is that the index.html of the cdn-web-components is pointing to a script src of `https://design.cms.gov/cdn/design-system/11.0.0/web-components/bundle/web-components.js` which is pointing at version 11 of the DS, so that's why the Autocomplete and Filter Chip are not rendering, and any existing version of the Design System JS on the CDN will not have the Dialog Modal on there because it is not released.

Let me update the version of the DS we're referencing here and see if that fixes the missing Autocomplete and Filter Chip, and then we should expect the Modal Dialog to render this way until it is released.

pwolfert: Oh right, I forgot that it's referencing an old version that doesn't have the new changes, so we can't actually expect it to work right now.

pwolfert (approved):

PR #3303, opened 2024-10-30
[WNMGDS-3025] Add markdown rendering to web component Storybook docs

pwolfert:
## Summary

A known limitation with our documentation tables within Storybook for web component attributes and events is that the descriptions don't render Markdown. And the TypeScript types for Storybook keep us from being able to pass JSX instead. I think there's even some kind of error in Storybook if you do pass in JSX to the argTypes to work around the issue (because the Storybook arg controls use the same descriptions).

- Add markdown rendering capabilities to the attributes and event tables
- Updates the event table descriptions, which _were_ able to have JSX workarounds before, unlike the `argTypes` rendered by the attributes table

## How to test

Open up Storybook and look at the `Description` column for both the attributes and events tables and the `Event object` column for the events tables.

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] Selected appropriate release milestone

PR #3304, opened 2024-10-31
[WMNGDS-3022] Removes `ds-tab-panel` module

tamara-corbalt:
## Summary 

- **Removes `ds-tab-panel` Module**: This component module was effectively dead code when used as intended (i.e., in tandem with `ds-tabs`). In `ds-tabs`, we were previously parsing each child to check if it was a `ds-tab-panel` web component (which, under the hood, was the react component `TabPanel`). Since the underlying react `Tabs` API expects each child to be a react component, we would convert it back into a react component—resulting in an awkward double conversion. Through prior investigative work ([PR-3295](https://github.com/CMSgov/design-system/pull/3295)), we confirmed that unregistered web components are treated as regular `<div>` elements, eliminating the need for `ds-tab-panel`.

- **Updates Parsing Logic in `ds-tabs`**:
    1. **Removes `ds-tab-panel` Dependency**: Simplifies `ds-tabs` by removing the unnecessary `ds-tab-panel` module.
    2. **Attribute Handling**: Updates the parsing function in `ds-tabs` to directly check for and map necessary attributes on each child `<div>` to the appropriate prop names that the react `Tabs` API expects. This refactor also addresses the original parsing issue from [PR-3295](https://github.com/CMSgov/design-system/pull/3295), converting `kebab-cased` attributes to `camelCased` props.

- **Attribute Prefix**: Because some attribute names (like `selected` and `disabled`) are not standard `<div>` attributes, they now use the `data-` prefix (e.g., `data-selected`, `data-disabled`) to avoid conflicts with reserved HTML attributes.

- **Updates Storybook Documentation**: Enhances Storybook documentation for `ds-tabs` to specify the expected attributes for each child `<div>`, ensuring developers have clear guidance on attribute usage and requirements.

Jira ticket: [WNMGDS-3022](https://jira.cms.gov/browse/WNMGDS-3022)

## How to test

- Run `yarn test:unit:wc -t Tabs`
- Check out `ds-tabs` docs in storybook.

## Checklist


packages/design-system/src/components/web-components/ds-tabs/ds-tabs.stories.tsx
```diff
@@ -13,8 +12,27 @@ const meta: Meta = {
     docs: {
       page: WebComponentDocTemplate,
       description: {
-        component:
-          'For information about how and when to use this component, [refer to its full documentation page](https://design.cms.gov/components/tabs/).',
+        component: `
+For information about how and when to use this component, [refer to its full documentation page](https://design.cms.gov/components/tabs/).
+
+The \`ds-tabs\` component accepts regular \`<div>\` elements as children to represent tab panels. Each \`<div>\` should contain specific content and attributes for tab functionality.
```
tamara-corbalt, line 18: Once we have markdown capabilities in storybook `argType` descriptions ([pr-3303](https://github.com/CMSgov/design-system/pull/3303)) we might consider relocating this documentation.

packages/design-system/src/components/web-components/ds-tabs/ds-tabs.stories.tsx
```diff
@@ -13,8 +12,27 @@ const meta: Meta = {
...
+- \`id\`: A unique identifier for the tab panel, linking it to its associated tab.
+
+**Optional Attributes:**
+- \`class-name\`: Custom class for styling the tab panel container.
+- \`data-selected\`: Accepts \`"true"\` or \`"false"\` as strings, marking the tab panel as initially selected.
+- \`data-disabled\`: Accepts \`"true"\` or \`"false"\` as strings, disabling the tab panel and making it unselectable.
+- \`data-tab\`: The label displayed on the associated tab for this panel.
+- \`data-tab-class-name\`: Additional CSS class for styling the associated tab.
+- \`data-tab-href\`: URL or link to navigate to when the tab is clicked.
+- \`data-tab-id\`: The \`id\` of the associated tab, used for \`aria-labelledby\` accessibility.
```
pwolfert, line 33: This is the first time we've had to prefix attributes with `data-` for web components. I guess that was a benefit of it being a custom element that nobody's HTML validators or TypeScript was going to complain about extra attributes.

packages/design-system/src/components/web-components/ds-tabs/ds-tabs.stories.tsx
```diff
@@ -13,8 +12,27 @@ const meta: Meta = {
...
+- **Content** (\`children\`): The inner content within the \`<div>\` representing the tab panel.
+
+**Required Attribute:**
+- \`id\`: A unique identifier for the tab panel, linking it to its associated tab.
+
+**Optional Attributes:**
+- \`class-name\`: Custom class for styling the tab panel container.
+- \`data-selected\`: Accepts \`"true"\` or \`"false"\` as strings, marking the tab panel as initially selected.
+- \`data-disabled\`: Accepts \`"true"\` or \`"false"\` as strings, disabling the tab panel and making it unselectable.
+- \`data-tab\`: The label displayed on the associated tab for this panel.
```
pwolfert, line 30: Is this one actually optional? Doesn't every tab need some tab text?

pwolfert (commented):
It all seems to be working, and you've successfully cut the dead code 💥. Was the `data-` prefixing an unforeseen consequence?

tamara-corbalt:
> Was the `data-` prefixing an unforeseen consequence?

Yes, unforeseen on my end (chalk it up to my naivete!). Using <ds-tab-panel> tag without defining it leads to typeScript complaints about it not being part of `JSX.IntrinsicElements`. And if we remove the `data-` prefix from div attributes, typeScript flags them as not being part of `HTMLDivElement`. So, the prefixing was my workaround to resolve these issues. Open to any other ideas!

packages/design-system/src/components/web-components/ds-tabs/ds-tabs.stories.tsx
tamara-corbalt: Yes, will update this

pwolfert (commented):
Found a typo [here](https://github.com/CMSgov/design-system/pull/3304/files#diff-1dea237bc429ac4ee4faa105c1b9a1c4755670c0d952d7241526a14315790fb1L58) when I was testing things in Storybook

packages/design-system/src/components/web-components/ds-tabs/ds-tabs.stories.tsx
jack-ryan-nava-pbc: So this idea is fine, I think my main concern was just making sure that the props that can be edited by users are visible/documented somewhere. So if that information is captured somewhere, and it looks like it is, then I'm not totally sure we need this custom element, unless @tamara-corbalt you feel different.

jack-ryan-nava-pbc (approved):
LGTM!

packages/design-system/src/components/web-components/ds-tabs/ds-tabs.stories.tsx
tamara-corbalt: Thanks for the suggestion! I wasn’t quite able to get it working, but I experimented a little and found that by extending `JSX.IntrinsicElements` directly, it allows typeScript to recognize `ds-tab-panel` as a valid element with specific attributes—so, maybe no need to define a new class or custom element. It seems to be working well, so I’ll push up the changes shortly.

packages/design-system/src/components/web-components/ds-tabs/ds-tabs.tsx
```diff
@@ -29,6 +60,16 @@ declare global {
...
+      'ds-tab-panel': JSX.IntrinsicElements['div'] & {
+        id: string;
+        'class-name'?: string;
+        selected?: string;
+        disabled?: string;
+        tab: string;
+        'tab-class-name'?: string;
+        'tab-href'?: string;
+        'tab-id'?: string;
+      };
```
pwolfert, line 72: 👍

pwolfert (approved):
Nice work!

PR #3310, opened 2024-11-07
[WNMGDS-2748] Removes deprecated aria-invalid from fieldset container for `MultiInputDateField`

tamara-corbalt:
## Summary

W3C suggests that the [aria-invalid state](https://www.w3.org/TR/wai-aria/#aria-invalid) is deprecated as a global state in ARIA 1.2. 
This change removes `aria-invalid` from the `<fieldset>` in the `MultiInputDateField` component. 
Jira ticket: [WNMGDS-2748](https://jira.cms.gov/browse/WNMGDS-2748)

- `aria-invalid` is intended for actionable elements (such as input fields) where it provides meaningful feedback. Since a `<fieldset>` is a container rather than an interactive element, `aria-invalid` does not effectively convey error information to screen readers when placed on the `<fieldset>` itself.
- `aria-invalid` is now applied only to specific inputs (day, month, or year) that are in an error state.
- Added unit tests to confirm that `aria-invalid` is correctly applied to invalid input fields only, ensuring compliance with ARIA 1.2 standards.

## How to test

- Try out the `MultiInputDateField` component with a screen reader in Storybook to confirm that fields with errors are announced correctly.
- Run unit tests: `yarn unit:test`

## Checklist

### DIFF
--- a/packages/design-system/src/components/DateField/MultiInputDateField.test.tsx
+++ b/packages/design-system/src/components/DateField/MultiInputDateField.test.tsx
@@ -36,6 +36,27 @@ describe('MultiInputDateField', () => {
     });
   });
 
+  it('renders with aria-invalid set to true on specific invalid inputs only', () => {
+    const errorProps = {
+      errorMessage: 'Please enter a valid date',
+      monthDefaultValue: '12',
+      dayDefaultValue: '35',
+      yearDefaultValue: '2050',
+      dayInvalid: true,
+      yearInvalid: true,
+    };
+
+    render(<MultiInputDateField label="A date field" {...errorProps} />);
+
+    const monthInput = screen.getByLabelText(/month/i);
+    const dayInput = screen.getByLabelText(/day/i);
+    const yearInput = screen.getByLabelText(/year/i);
+
+    expect(monthInput).toHaveAttribute('aria-invalid', 'false');
+    expect(dayInput).toHaveAttribute('aria-invalid', 'true');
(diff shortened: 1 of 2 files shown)


pwolfert (commented):
I tried it out with VoiceOver, and it does correctly read out the error message when I'm focused on the input that is market as invalid. Even at the time of originally implementing this, it wasn't a perfect solution, and when we got feedback from a new accessibility-team member, we knew we needed to research this further. This might need to be tested with more screen reader and browser combos.

Note that we use `aria-invalid` on fieldsets in `ChoiceList` and `MonthPicker` as well.

tamara-corbalt:
> I tried it out with VoiceOver, and it does correctly read out the error message when I'm focused on the input that is marked as invalid. Even at the time of originally implementing this, it wasn't a perfect solution, and when we got feedback from a new accessibility-team member, we knew we needed to research this further. This might need to be tested with more screen reader and browser combos.
> 
> Note that we use `aria-invalid` on fieldsets in `ChoiceList` and `MonthPicker` as well.

Thanks for testing with VoiceOver! I agree we should test across more assistive tech and browsers. I can create a follow-up ticket for broader testing with screen readers (JAWS, NVDA, etc.) and browsers. I can also apply similar changes to `ChoiceList` and `MonthPicker` in a branch so we can test everything together. Does this sound like a good plan? 
CC: @kim-cmsds @jack-ryan-nava-pbc

PR #3315, opened 2024-11-12
[WNMGDS-2750] Remove `aria-disabled` from `TabPanel` components

tamara-corbalt:
## Summary

- Removes deprecated `aria-disabled` attribute from `TabPanel` components to align with current W3C guidance, which advises against using `aria-disabled` on elements with `role="tabpanel"`.
- Updates a unit test to ensure that disabled `TabPanel` components are marked with `aria-hidden="true"`.

See the [Jira story and commentary](https://jira.cms.gov/browse/WNMGDS-2750) for context.

## How to test

- Open the Tabs component's disabled story in Storybook.
- Use a screen reader (e.g., VoiceOver) to navigate the tabs.
- Confirm that the screen reader announces disabled tabs as "dimmed," indicating they are non-interactive.

## Checklist


jack-ryan-nava-pbc:
Do we want to remove `disabled` as an available prop [here](https://github.com/CMSgov/design-system/pull/3315/files#diff-6dc00264b1de7edead0f1dc2cabc578086a6487422320737962f3ca95007c863R15)?

Also, just thinking out loud here but I think this is a breaking change because if folks already have `disabled` set to true on a tab, that tab will no longer be disabled/hidden from users once they upgrade? Is that correct?

tamara-corbalt:
> Do we want to remove `disabled` as an available prop [here](https://github.com/CMSgov/design-system/pull/3315/files#diff-6dc00264b1de7edead0f1dc2cabc578086a6487422320737962f3ca95007c863R15)?
> 
> Also, just thinking out loud here but I think this is a breaking change because if folks already have `disabled` set to true on a tab, that tab will no longer be disabled/hidden from users once they upgrade? Is that correct?

Hey @jack-ryan-nava-pbc, thanks for taking a look! The purpose of this change is to remove the `aria-disabled` attribute from `TabPanel` in the disabled state, as `aria-disabled` isn’t recommended on elements with `role="tabpanel"`. We’re keeping the `aria-hidden="true"` attribute to ensure that the disabled TabPanel stays non-interactive, so I don’t think we want to remove the `disabled` prop itself.

jack-ryan-nava-pbc:
> > Do we want to remove `disabled` as an available prop [here](https://github.com/CMSgov/design-system/pull/3315/files#diff-6dc00264b1de7edead0f1dc2cabc578086a6487422320737962f3ca95007c863R15)?
> > Also, just thinking out loud here but I think this is a breaking change because if folks already have `disabled` set to true on a tab, that tab will no longer be disabled/hidden from users once they upgrade? Is that correct?
> 
> Hey @jack-ryan-nava-pbc, thanks for taking a look! The purpose of this change is to remove the `aria-disabled` attribute from `TabPanel` in the disabled state, as `aria-disabled` isn’t recommended on elements with `role="tabpanel"`. We’re keeping the `aria-hidden="true"` attribute to ensure that the disabled TabPanel stays non-interactive, so I don’t think we want to remove the `disabled` prop itself.

Ok cool, thanks for the clarification. Do you think it would be helpful to rename the prop to `hidden` so that it more closely maps onto what the prop gets used for?

tamara-corbalt:
> Ok cool, thanks for the clarification. Do you think it would be helpful to rename the prop to `hidden` so that it more closely maps onto what the prop gets used for?

@jack-ryan-nava-pbc, I see where you're coming from with renaming, but I think the `disabled` prop is still accurately named based on its usage within the `Tabs` component.

If you look here ([link to code](https://github.com/CMSgov/design-system/blob/tamara/WNMGDS-3027/add-wc-sections-to-component-pages/packages/design-system/src/components/Tabs/Tabs.tsx#L199-L206)), each TabPanel child is mapped to a Tab, and the `disabled` prop from `TabPanel`is passed along to the `Tab `component. This `disabled` prop then determines whether `aria-disabled` is applied to `Tab` components (those with `role="tab"`), which is supported for accessibility.

jack-ryan-nava-pbc (approved):
Works! Approve!

kim-cmsds (approved):

PR #3318, opened 2024-11-15
[WNMGDS-3004] Update USA Banner to match Figma

jack-ryan-nava-pbc:
## Summary

- [Ticket](https://jira.cms.gov/projects/WNMGDS/3004)
- USA Banner now uses Open Sans & proper font weights

## How to test

1. Make sure Core and child design systems match the following specs:
- Font family is open-sans, 
- font weights are 400 (normal) and 700 (bold)

## Checklist


packages/design-system/src/styles/components/_UsaBanner.scss
```diff
@@ -5,9 +5,10 @@
 .ds-c-usa-banner {
   // The following local CSS vars aren't meant to be themed; they're meant to be consistent across all CMS properties.
   // System font stack was chosen to avoid having users download unnecessary fonts just for the banner.
```
pwolfert, line 7: With this change, it seems like this comment is no longer valid

PR #3320, opened 2024-11-15
WNMGDS-3037 Adds script to compute percentage of nightly tests that pass

tamara-corbalt:
## Summary

This script analyzes jenkins nightly build data to compute a general pass rate and outputs a list of failed builds for manual inspection.
**Note:** It utilizes the Puppeteer library to simulate logging into Jenkins as a workaround, pending resolution of issues with setting up a proper API token.

```
// Sample output for a range of builds
% npx ts-node get-nightly-build-pass-rate.ts 750 3
Logging in to Jenkins...
Login successful.
Analyzing builds from 748 to 750...
Fetching data for build #748...
Fetching data for build #749...
Fetching data for build #750...
Analysis complete.
Pass Count: 3
Fail Count: 0
Pass-to-Fail Rate: 100.00%
Failed Builds: 
```
[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3037) 

## How to test
- Pull down the branch
- Ensure you are connected to the Zscaler VPN.
- Navigate to scripts directory:`cd scripts`
- Create a .env file in the scripts directory with the following variables:
```
JENKINS_ID=<your-eua-id>
JENKINS_PASSWORD=<your-password>
JENKINS_LOGIN_URL=<Jenkins login URL>
JENKINS_BASE_API_URL=<Jenkins base API URL>
```
- Run `yarn install`
- Run `npx ts-node get-nightly-build-pass-rate.ts <build number> <numberOfDays>`
  - Replace <build number> with the latest jenkins build number you want to inspect.
  - Replace <numberOfDays> with the range of days to analyze (optional; defaults to 1 if omitted).

## Checklist


scripts/get-nightly-build-pass-rate.ts
```diff
@@ -0,0 +1,108 @@
...
+
+        const responseBody = await response.text();
+
+        if (!responseBody) {
+          console.log(`Build #${buildNumber}: Empty response body.`);
+          continue;
+        }
+
+        let data: { result?: string } | null = null;
+        try {
```
tamara-corbalt, line 57: Planning to move the block that handles parsing and processing build data into a separate function in the future. This will make the code more modular and easier to maintain.

pwolfert:
Would the private toolkit repo be a more appropriate place for this?

tamara-corbalt:
> Would the private toolkit repo be a more appropriate place for this?
> 
> If you're using Puppeteer instead of Playwright, there's no gain in shared dependencies by putting it in this repo anyway. That is, you'd have to install a browser-testing tool in the toolkit, but you're installing a new one anyway.

Hey @pwolfert, that's a great call-out, and I completely agree. I think this script would be a better fit for our private toolkit repo. I'll work on moving it over. Thanks for pointing that out! 😊

scripts/get-nightly-build-pass-rate.ts
```diff
@@ -0,0 +1,108 @@
+import puppeteer from 'puppeteer';
+import dotenv from 'dotenv';
+dotenv.config();
+
+const loginAndAnalyzeBuilds = async (
+  latestBuildNumber: number,
+  numberOfDays: number,
```
jack-ryan-nava-pbc, line 7: I think we're more likely to have the start date of a PI on hand rather than the number of days back we want to go. Do you think you could change this to accept a PI start date?

scripts/get-nightly-build-pass-rate.ts
```diff
@@ -0,0 +1,108 @@
...
+  try {
+    console.log('Logging in to Jenkins...');
+    await page.goto(loginUrl, { waitUntil: 'networkidle2' });
+    await page.type('#j_username', username);
+    await page.type('input[name="j_password"]', token);
+    await Promise.all([
+      page.click('button[type="submit"]'),
+      page.waitForNavigation({ waitUntil: 'networkidle2' }),
+    ]);
+    console.log('Login successful.');
```
jack-ryan-nava-pbc, line 30: Just checking but pretty much everything before here will be removed once you get access to the Jenkins REST API?

scripts/get-nightly-build-pass-rate.ts
```diff
@@ -0,0 +1,108 @@
...
+      const url = `${baseApiUrl}/${buildNumber}/api/json`;
+      try {
+        const response = await page.goto(url, { waitUntil: 'networkidle0', timeout: 10000 });
+
+        if (!response) {
+          console.log(`Build #${buildNumber}: No response from server.`);
+          continue;
+        }
+
+        const responseBody = await response.text();
```
jack-ryan-nava-pbc, line 49: Could you call `response.json()` here and remove the later call to JSON.parse? Maybe not since Puppeteer is returning HTML?

scripts/get-nightly-build-pass-rate.ts
jack-ryan-nava-pbc: Nit: This seems pretty straightforward and not super long, I think you could leave it inlined but a separate function is fine too!

scripts/get-nightly-build-pass-rate.ts
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    const failCount = failedBuilds.length;
+    const passToFailRate = (passCount / (passCount + failCount)) * 100;
+
+    console.log(`Analysis complete.`);
+    console.log(`Pass Count: ${passCount}`);
+    console.log(`Fail Count: ${failCount}`);
+    console.log(`Pass-to-Fail Rate: ${passToFailRate.toFixed(2)}%`);
+    console.log(`Failed Builds: ${failedBuilds.join(', ')}`);
```
jack-ryan-nava-pbc, line 88: This is looking great! I would love to have these results written to a file. Do you think you could write these out to a CSV with 4 columns? PI number, pass count, fail count, ratio and maybe a list of the builds that were not accounted for for whatever reason. Does that sound like a good structure/reasonable?

scripts/get-nightly-build-pass-rate.ts
```diff
@@ -0,0 +1,108 @@
...
+        let data: { result?: string } | null = null;
+        try {
+          data = JSON.parse(responseBody);
+        } catch (jsonError) {
+          console.error(`Build #${buildNumber}: Failed to parse JSON.`, jsonError);
+          continue;
+        }
+
+        if (!data || !data.result) {
+          console.log(`Build #${buildNumber}: Data unavailable or missing result.`);
```
jack-ryan-nava-pbc, line 65: Can we track all these weird data unavailable cases? Like if the server returns a 500 or if there are missing data on a build for some other reason?

scripts/get-nightly-build-pass-rate.ts
tamara-corbalt: Yes, that's the plan. Once we sort out the issue with API credentials, we can probably dispense with Puppeteer, which feels like a heavy-handed tool for what this script could be. At that point, we’d likely switch to something simpler, like `node-fetch`, to handle the requests.

scripts/get-nightly-build-pass-rate.ts
tamara-corbalt: Oh yes, good call! I'd like to see better formatting for this output.

scripts/get-nightly-build-pass-rate.ts
tamara-corbalt: So, I didn't think it was guaranteed that the response would always be JSON. Since we're retrieving page content, it might return HTML or plain text, so I felt using response.text() and handling JSON parsing in a try/catch block was a safer approach for now. We can revisit and potentially refactor this once we move away from Puppeteer.

scripts/get-nightly-build-pass-rate.ts
tamara-corbalt: I think accepting a date as input might introduce some challenges. Are you suggesting that the user manually identify the build number for the start of the PI based on that date? I feel that could make things more difficult, especially since we'll likely be running this script after a PI concludes.

It seems easier to locate the build number for the end of the PI—jenkins and our slack channel make it straightforward to find the most recent build. Backtracking to determine the exact build corresponding to a PI's start date would likely require more effort on the user's part. That said, if there’s an automated way to map dates to build numbers, I’d be happy to explore that!

jack-ryan-nava-pbc: I'm also asking you to work with Dates and Time in _Javascript_, so it's ok if that's just a categorical 'no' 😆

tamara-corbalt: @jack-ryan-nava-pbc, Ohhh I misunderstood! I think I see the appeal in using a specific date to calculate the starting build number—it definitely aligns more naturally with how we think about time. However, since our PIs are typically a fixed duration (I should double check this assumption), sticking to a set number of days might keep things simpler and more predictable for our use case?

scripts/get-nightly-build-pass-rate.ts
tamara-corbalt: @jack-ryan-nava-pbc, After giving it some more thought, I think it might be best to hold off on finalizing the output format. I discussed some possible new features with @kim-cmsds earlier, and I foresee that we might be adding some elements down the road that could impact how we structure our data output. So to avoid revisions, I'd like to delay this for now. What do you think?

scripts/get-nightly-build-pass-rate.ts
tamara-corbalt: I'll add more descriptive logging for unavailable cases

scripts/get-nightly-build-pass-rate.ts
jack-ryan-nava-pbc: Well, a fixed number of days wouldn't work if the PI isn't complete yet?

scripts/get-nightly-build-pass-rate.ts
jack-ryan-nava-pbc: Sounds good! I'm totally onboard.

scripts/get-nightly-build-pass-rate.ts
tamara-corbalt: > Well, a fixed number of days wouldn't work if the PI isn't complete yet right?

Ah, I see your point now! I think I might not have been super clear earlier. When I mentioned a set number of days, I was referring to the fact that PIs are typically a fixed duration (e.g., 54 days for PI4). Since this script is mostly being used in retrospect, I thought it might be easier to find the final build number and work backward a set number of builds, rather than trying to pinpoint the exact date for the start of the PI.

If you feel strongly about using dates, let’s move this discussion over to [pr-75](https://github.cms.gov/CMS-WDS/experience-toolkit/pull/75) since I’m planning to close this one out.

tamara-corbalt:
Moving this script into `experience-toolkit` repo.

PR #3329, opened 2024-11-20
[NO-TICKET] Updates PR Template to reflect new script that updates all browser tests

kim-cmsds:
## Summary

- Adds a reference to the new script (`yarn test:browser:all:update`) that will update snapshots in each type of browser test in the PR template file
- One goal is to remind developers that while CI tests may pass because those send along `SMOKE=true` in [tests.groovy](https://github.cms.gov/CMS-WDS/experience-toolkit/blob/main/jenkins/tests.groovy#L116), those skip some projects like 'chromium-forced-colors' due to this [logic](https://github.com/CMSgov/design-system/blob/main/tests/browser/playwright.config.ts#L111), which means nightly build snapshots may not be updated and tests might still pass.

## Checklist


tamara-corbalt (approved):
Awesome!

PR #3330, opened 2024-11-21
[NO-TICKET] Prevent double firing of exit event in Modal Dialog

jack-ryan-nava-pbc:
## Summary

- Problem: when backdropClickExits is enabled, we see the onExit or ds-exit event called twice when exiting the Modal either via the close button or a form submit button
- This is only an issue in Chrome and Firefox and any other browsers that set clientX & clientY to zero on key press (likely Edge though I haven't tested). Safari is smart and doesn't have this issue. Maybe I'll start using Safari for more than just testing, but also probably not.
- The simplest fix that we landed on is to add a simple check for the number of clicks on the event. This information is stored in the event.detail read only property. 
- If it's 0, that means there was no click of a mouse, so a key must've been pressed (or so my logic goes) & return out of the function
- Else, check if we're in the bounding rect of the dialog element, which we won't be if you clicked outside of the dialog element.

## How to test

1. `yarn build`
2. `yarn storybook`
3. Open modal dialog (either React or Web Component version) in Storybook in Browser
4. Make sure backdropClickExits/backdrop-click-exits is true
5. Open the dialog
6. Press `TAB` to bring the close or submit buttons into focus
7. Press `ENTER` or `SPACEBAR`
8. Note that only one onExit or ds-exit event has been fired.

## Checklist

### DIFF
--- a/packages/design-system/src/components/NativeDialog/NativeDialog.tsx
+++ b/packages/design-system/src/components/NativeDialog/NativeDialog.tsx
@@ -108,6 +108,17 @@ export const NativeDialog = ({
 
     const dialogNode = dialogRef.current;
     const handleClick = (event) => {
+      // In Chrome and Firefox Pointer Events triggered by a key press receive a clientX & clientY value of 0 each.
+      // This puts the pointer outside of our dialog element, and so we trigger the exit() event.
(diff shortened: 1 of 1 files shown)


tamara-corbalt (approved):
Works for me!

kim-cmsds (approved):

PR #3335, opened 2024-11-26
[WNMGDS-3044] Adds custom playwright reporter to provide JSON summary of test result data

tamara-corbalt:
## Summary

- Adds a custom Playwright reporter that generates a JSON summary of test results.
- The JSON report includes statistics on the number of passing, failing, and skipped tests, as well as details of failing tests and skipped tests.
- The primary objective is to make test statistics more accessible on the `test-nightly` pipeline in Jenkins.
- This effort lays the groundwork for tracking test coverage of browser tests.

[Jira Ticket](https://jira.cms.gov/browse/WNMGDS-3044)

### Example JSON output
Screenshot 2024-11-26 at 11 01 46 AM

## How to test

- Pull down branch.
- Run `yarn test:browser` or `yarn test:browser:examples` or any other test browser commands.
- Inspect logs: verify that hierarchy paths for failing tests are logged correctly during the run.
- Confirm that the `test-report.json` file is generated in the expected directory: `tests/browser`

## Checklist


tamara-corbalt:
**Note:** Marking this as a draft for now. The latest commit seems to have caused issues with the browser tests, likely due to a conflict between Playwright and Jest snapshot directories. I’ll investigate and resolve this later.

tests/browser/custom-reporter.ts
```diff
@@ -0,0 +1,88 @@
...
+      summary: {
+        passed: this.passCount,
+        failed: this.failCount,
+        skipped: this.skipCount,
+      },
+      failingTests: this.failingTests,
+      skippedTests: this.skippedTests,
+    };
+
+    const reportPath = path.resolve(__dirname, 'test-report.json');
```
kim-cmsds, line 78: What do you think about adding this file to `/test-results/` so that we're not tracking the `test-report.json` file?

tests/unit/jest.config.js
```diff
@@ -13,16 +13,25 @@ const preactModuleMapper =
...
+        '<rootDir>/packages/design-system/src/components/web-components/**/*.test.[jt]s(x)?',
+        '<rootDir>/tests/browser/custom-reporter.test.ts',
+      ],
+      testPathIgnorePatterns: ['<rootDir>/tests/browser/'],
     }
-  : { testPathIgnorePatterns: ['<rootDir>/design-system/src/components/web-components'] };
+  : {
+      testPathIgnorePatterns: [
+        '<rootDir>/packages/design-system/src/components/web-components',
+        '<rootDir>/tests/browser/',
```
jack-ryan-nava-pbc, line 25: Why is /tests/browser ignored here?

tests/browser/custom-reporter.ts
tamara-corbalt: Sounds good!

tests/unit/jest.config.js
tamara-corbalt: Ooh, yes good catch, I think the logic of this configuration could use some clarity. It seems that if I remove <rootDir>/tests/browser/ from the testPathIgnorePatterns list when useWebComponents is set to false, we end up overwriting it in the main configuration spread, [here](https://github.com/CMSgov/design-system/blob/main/tests/unit/jest.config.js#L26-L33). This results in unintentionally including the playwright tests in our jest runs.

tests/browser/custom-reporter.ts
```diff
@@ -0,0 +1,88 @@
...
+import path from 'path';
+
+class MyReporter implements Reporter {
+  private passCount = 0;
+  private failCount = 0;
+  private skipCount = 0;
+  private failingTests: { path: string; name: string }[] = [];
+  private skippedTests: { path: string; name: string }[] = [];
+
+  onBegin(config: FullConfig, suite: Suite) {
```
jack-ryan-nava-pbc, line 19: Nit: Is `config` required here as an argument since it is not used?

tests/browser/custom-reporter.ts
```diff
@@ -0,0 +1,88 @@
...
+    console.log(`Starting the run with ${suite.allTests().length} tests`);
+  }
+
+  onTestEnd(test: TestCase, result: TestResult) {
+    const getHierarchyPath = (test: TestCase): string => {
+      const path: string[] = [];
+      let parent: Suite | undefined = test.parent;
+      while (parent) {
+        path.unshift(parent.title || '');
+        parent = parent.parent;
```
jack-ryan-nava-pbc, line 29: Maybe use optional chaining on parent.title and parent.parent to prevent this blowing up?

tests/browser/playwright.config.ts
```diff
@@ -76,7 +76,12 @@ const smokeTestProjects = [
  */
 const config: PlaywrightTestConfig = {
   testDir: './',
-  testIgnore: ['storybook-docs.test.ts', 'examples.test.ts'],
+  testIgnore: [
+    'storybook-docs.test.ts',
+    'examples.test.ts',
+    '**/__snapshots__/**',
```
jack-ryan-nava-pbc, line 82: Can you explain why there's a new snapshots directory? I'm just curious and wonder why we aren't using the one that already exists? I can see these snapshots of a test reported being like a different kind of snapshot from our UI/storybook snapshots.

tests/browser/custom-reporter.ts
tamara-corbalt: @jack-ryan-nava-pbc, I think `parent.title` and `parent.parent` are safe in this context, because the `while` loop condition ensures these lines will only execute if `parent` is truthy

tests/unit/jest.config.js
jack-ryan-nava-pbc: Ah yeah ok this makes sense I think. Is this happening because of the changes to the rootDir and coverageDirectory changes below?

jack-ryan-nava-pbc (approved):
Super nits and understanding checks. Looks good!

tests/browser/playwright.config.ts
tamara-corbalt: @jack-ryan-nava-pbc, The `__snapshots__` directory contains snapshots generated by unit tests for the custom reporter. It’s separate from the snapshots directory used by browser tests to avoid mixing unrelated test artifacts.

tests/browser/custom-reporter.ts
jack-ryan-nava-pbc: I'm being over anxious and just wondering if there could be situation where `parent` exists but without the fields of title or parent attached?

tests/browser/custom-reporter.ts
tamara-corbalt: @jack-ryan-nava-pbc, Looks like this parameter is required by the Reporter api, but I've renamed it to `_` to indicate that it's not used.

tests/browser/custom-reporter.ts
tamara-corbalt: @jack-ryan-nava-pbc, Not over-anxious at all—it’s a good point! Based on the `Suite` type, `title` is always present, and `parent` is optional. The `while (parent)` condition ensures that `parent` is truthy before we access it, so the current implementation is safe. Let me know if you’d still prefer adding optional chaining for extra clarity, though I don’t think it’s strictly necessary here.

jack-ryan-nava-pbc: Perfect. Thanks for confirming @tamara-corbalt!

kim-cmsds (approved):

PR #3344, opened 2024-12-05
[WNMGDS-3046] Add option groups to autocomplete

tamara-corbalt:
## Summary
This pull request introduces support for grouped items in the Autocomplete component. It leverages the `<Section>` element from react-aria, aligning with the pattern already used in our Dropdown component.

### Key Details:
- Instead of traditional dividers, grouped items are displayed with indentation under headers.
- This approach maintains consistency across components by reusing a similar grouping pattern.
- **Note on Testing**: Unit tests are not yet included, as feedback from the team is needed to confirm the suitability of this pattern.

### Autocomplete with grouped items:
Screenshot 2024-12-06 at 4 44 40 PM

### Autocomplete with a mix of grouped items and standalone items:
Screenshot 2024-12-06 at 4 44 29 PM

## How to test
- `yarn storybook`
- Check out the `Autocomplete` stories:
  - "Option Groups" 
  - "Groups and Standalone Items"

## Checklist


pwolfert (commented):
It works well, and I tried it in VoiceOver too. One question below about the ability to mix and match.

packages/design-system/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -139,6 +163,47 @@ export const LabeledList: Story = {
...
+      makeGroup('Healthcare Specialties', [
+        makeItem('Pediatrics'),
+        makeItem('Gastroenterology'),
+        makeItem('Dermatology'),
+        makeItem('Cardiology'),
+        makeItem('Neurology'),
+        makeItem('Orthopedics'),
+      ]),
+    ],
+  } as any,
```
pwolfert, line 204: Does this have to be `any` because the type here is `Array< AutocompleteItem | AutocompleteItemGroup>` instead of `AutocompleteItem[] | AutocompleteItemGroup[]`?

pwolfert: Would retyping `AutocompleteItems` to be an array of both item types be more accurate?

tamara-corbalt: > Does this have to be `any` because the type here is `Array<AutocompleteItem | AutocompleteItemGroup>` instead of `AutocompleteItem[] | AutocompleteItemGroup[]`?

@pwolfert, I think I might be missing something here, because it looks like the `as any` is applied to the `args` object, not directly to the `items` array. That said, I could probably apply more accurate type casting to the `items` array. Let me know if you had something else in mind!

pwolfert: Oh, I didn't catch that. I thought it was for the array. I'm curious if you were to remove it if it would complain about the `items` array specifically along with any other errors it has.

kim-cmsds: @tamara-corbalt instead of `as any`, you could do something like this `as AutocompleteProps & { textFieldLabel: string; textFieldHint: string }`

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -39,6 +39,14 @@ export interface AutocompleteItem extends Omit<React.HTMLAttributes<'option'>, '
   isResult?: boolean;
 }
 
+export interface AutocompleteItemGroup {
+  label: string;
+  items: AutocompleteItem[];
+  id: string;
+}
+
+export type AutocompleteItems = AutocompleteItem[] | AutocompleteItemGroup[];
```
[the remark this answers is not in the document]
tamara-corbalt: After visiting [care compare](https://www.medicare.gov/care-compare/?providerType=Physician), it looks like  there are both standalone items and items grouped under headers. This suggests that our Autocomplete component should support a mix of individual items and groups to align with this use case.

jack-ryan-nava-pbc (changes requested):
This looks great! The only thing is that the Headings appear instead of the 'No results' message if you provide an invalid input. I think that probably should be fixed before shipping.

tamara-corbalt:
> This looks great! The only thing is that the Headings appear instead of the 'No results' message if you provide an invalid input. I think that probably should be fixed before shipping.

@jack-ryan-nava-pbc Great catch! I've updated the logic to evaluate whether we have valid items, ensuring the 'no results' message is displayed when there are only empty item groups. I also added a unit test to verify this behavior. Let me know what you think!

kim-cmsds (commented):
@tamara-corbalt this is looking good! Can you please add an [interaction test](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/Autocomplete/Autocomplete.test.interaction.ts#L4) for the `ItemGroups` story? That way we'll get slightly more helpful snapshots

kim-cmsds (approved):
Appreciate the extra test coverage ✅

jack-ryan-nava-pbc (approved):
One small non-blocking q and approve!

PR #3345, opened 2024-12-06
[NO-TICKET] Use isPossibleValue util function in more places and add spec

jack-ryan-nava-pbc:
## Summary

- In [3331](https://github.com/CMSgov/design-system/pull/3331) I added a cute little function that serves as a type predicate for some of our attributes on our web components
- Here we use it more broadly, introducing the needed constants and types in the ModalDialog and Tooltip components
- I also added a spec. Which isn't the best spec since it doesn't do the type predicate stuff that we use this function for, but maybe it's sufficient to ensure the basic functionality of the thing. 

## How to test

1. Confirm tests pass locally
2. Fire up Storybook or one of the example apps and confirm that ModalDialog and Tooltip web components operate as expected.

## Checklist

### DIFF
--- a/packages/design-system/src/components/web-components/utils.test.ts
+++ b/packages/design-system/src/components/web-components/utils.test.ts
@@ -1,0 +1,28 @@
+import { isPossibleValue } from './utils';
+
+type Fruit = 'apple' | 'banana' | 'cherry';
+
+const acceptableFruit: Fruit[] = ['apple', 'banana', 'cherry'] as const;
+
+describe('isPossibleValue', () => {
+  it('should return true if the value is in the possibleValues array', () => {
+    const value = 'apple' as Fruit;
+    expect(isPossibleValue(value, acceptableFruit)).toBe(true);
+  });
+
+  it('should return false if the value is not in the possibleValues array', () => {
+    const value = 'orange';
+    expect(isPossibleValue(value, acceptableFruit)).toBe(false);
+  });
+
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/web-components/utils.test.ts
```diff
@@ -0,0 +1,28 @@
+import { isPossibleValue } from './utils';
+
+type Fruit = 'apple' | 'banana' | 'cherry';
+
+const acceptableFruit: Fruit[] = ['apple', 'banana', 'cherry'] as const;
```
jack-ryan-nava-pbc, line 5: I added the Fruit type and accpetableFruit array to get closer to how we use this function in the wild. I'm currently researching whether there are type testing approaches or frameworks that might be useful, since I can't really test the types. I've found this [helpful article](https://www.totaltypescript.com/how-to-test-your-types) which suggests two separate libraries for testing types, and a unique approach using Expect and Equal. The problem with the Expect and Equal approach is that the 'test' is just checking if the code compiles. Which is correct, but kinda outside of how we do tests normally.

kim-cmsds: [TIL](https://stackoverflow.com/questions/62028427/how-to-use-ts-type-as-a-value) that "You can't make a value from a type, but you can infer a type from a value."

I'm personally ok with a test that gets us closer to how it's used in the rest of the codebase.

kim-cmsds (approved):

PR #3346, opened 2024-12-06
[WNMGDS-3057] Add Shadow DOM capabilities to preactement

pwolfert:
## Summary

- Renamed the public `renderPreactComponent` function to `forceRender` to make it clear what that function does from the outside
- Split the render function into two possible branches depending on whether the `shadow` option is turned on:
  1. `renderWithoutShadowDom` - Same behavior as before
  2. `renderWithShadowDom` - Passes `<slot>` elements for the `children` prop of the underlying Preact component as well as named `<slot name={theName}>` elements for each of the props that the user has defined with slots (instead of as an element attribute). Then it renders the Preact component into a `ShadowRoot`.

## How to test

You can run just the `define` tests in isolation with the following:

```sh
yarn test:unit:wc packages/design-system/src/components/web-components/preactement/define.test.tsx
```


packages/design-system/src/components/web-components/preactement/globalStyles.ts
```diff
@@ -0,0 +1,84 @@
+const POLLING_INTERVAL = 10;
```
tamara-corbalt, line 1: Curious to know how this specific value was chosen, given the potential performance impacts of frequent polling.

tamara-corbalt (commented):
Overall, this looks great! Left a question about the chosen polling frequency for detecting changes in global styles. 
Also, curious to know if we have plans in place to monitor the performance impacts of this polling mechanism?

tamara-corbalt:
> Also, curious to know if we have plans in place to monitor the performance impacts of this polling mechanism?

I think I have more context now that I've read the description on [pr-3327](https://github.com/CMSgov/design-system/pull/3327/) From what I understand, the long-term plan involves restructuring our CSS to move toward a solution where such frequent polling won't be necessary. Is that correct?

packages/design-system/src/components/web-components/preactement/globalStyles.ts
pwolfert: I wanted it short enough to not see flashes of _mis_-styled content, and I figured since I was only checking the 2-10 or so stylesheet elements in a document and not accessing all their children that it would be fairly inexpensive. This is not based on any empirical data, though. Do you have a specific number in mind?

pwolfert: Everyone vote on their favorite polling interval? 😄

pwolfert: Also I hadn't thought of monitoring the performance, but do you have ideas?

pwolfert:
> From what I understand, the long-term plan involves restructuring our CSS to move toward a solution where such frequent polling won't be necessary

Yes, that is one path forward, and I do hope the polling isn't the long-term solution. Like ideally we'd only apply the necessary styles for that component in a long-term solution.

packages/design-system/src/components/web-components/preactement/globalStyles.ts
tamara-corbalt: Thanks for clarifying the polling scope; it’s more limited than I realized, which is good to know. Regarding performance impacts—granted I haven't thought about this all that deeply—I was thinking about using browser profiling tools against our example angular application as a possible way to assess performance. I'm not super experienced with profiling tools, but that's the first thing that came to mind.

pwolfert: Okay, I successfully profiled it and found that our code that runs every 10ms executes on average in 50μs (50 millionths of a second)!

packages/design-system/src/components/web-components/preactement/globalStyles.ts
```diff
@@ -0,0 +1,84 @@
...
+  if (globalSheets === null) {
+    globalSheets = copyGlobalStyleSheets();
+    watchForFutureChanges();
+  }
+
+  return globalSheets;
+}
+
+function copyGlobalStyleSheets(): CSSStyleSheet[] {
+  return Array.from(document.styleSheets).map((x) => {
```
jack-ryan-nava-pbc, line 16: Super nit: Can we replace `x` with `sheet` or something slightly more informative?

pwolfert: Yeah, that's a good idea. This original function was [from this helpful blog post](https://eisenbergeffect.medium.com/using-global-styles-in-shadow-dom-5b80e802e89d) for reference

pwolfert: [c2cdf75](https://github.com/CMSgov/design-system/pull/3346/commits/c2cdf75ad1584d9cbbf29fb796b498e45955a449)

jack-ryan-nava-pbc (approved):
This is really nice. Thank you for breaking this out into a separate PR! Great work! ✋ 5️⃣  (high-five)

PR #3347, opened 2024-12-06
[WNMGDS-3057] Update analytics-content-extraction method to support slots

pwolfert:
## Summary

We need to change our React component analytics-content-extracting code to be able to follow `slot` elements across the Shadow DOM barrier. Otherwise, the analytics events will have nothing for things like alert heading when the heading content is defined with a slot.

## How to test

Run the unit tests and make sure no snapshots change.

### DIFF
--- a/packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
+++ b/packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
@@ -1,7 +1,34 @@
 import { RefObject } from 'react';
 
+/**
+ * Gets text content from an element and even follows `slot` elements across Shadow DOM
+ * barriers to get the full text that the user sees.
+ */
+function getTextContent(nodes: Node[]): string {
+  return nodes
+    .map((node) => {
+      if (node.nodeType === Node.TEXT_NODE) {
+        return node.textContent.trim();
+      } else if (node.nodeType === Node.ELEMENT_NODE) {
+        if ((node as HTMLElement).tagName === 'SLOT') {
+          // If it's a slot, it won't actually have any content in itself, so we have to
+          // follow the trail to its "assigned nodes" and look there.
+          const assignedNodes = (node as HTMLSlotElement).assignedNodes({ flatten: true });
+          return getTextContent(assignedNodes);
+        } else {
+          return getTextContent(Array.from((node as HTMLElement).childNodes));
+        }
+      } else {
+        return '';
+      }
+    })
+    .join(' ');
+}
+
 export function getAnalyticsContentFromRefs(refs: RefObject<any>[]): string | undefined {
-  return refs.map((ref) => ref.current?.textContent).find((textContent) => textContent);
+  return refs
+    .map((ref) => (ref.current ? getTextContent(Array.from(ref.current.childNodes)) : undefined))
+    .find((textContent) => textContent);
 }
 
 export default getAnalyticsContentFromRefs;
--- a/packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
+++ b/packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
@@ -8,7 +8,7 @@ function getTextContent(nodes: Node[]): string {
   return nodes
     .map((node) => {
       if (node.nodeType === Node.TEXT_NODE) {
-        return node.textContent.trim();
+        return node.textContent;
       } else if (node.nodeType === Node.ELEMENT_NODE) {
         if ((node as HTMLElement).tagName === 'SLOT') {
           // If it's a slot, it won't actually have any content in itself, so we have to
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
```diff
@@ -1,7 +1,34 @@
...
+/**
+ * Gets text content from an element and even follows `slot` elements across Shadow DOM
+ * barriers to get the full text that the user sees.
+ */
+function getTextContent(nodes: Node[]): string {
+  return nodes
+    .map((node) => {
+      if (node.nodeType === Node.TEXT_NODE) {
+        return node.textContent;
+      } else if (node.nodeType === Node.ELEMENT_NODE) {
```
tamara-corbalt, line 12: @pwolfert  thanks for the accompanying video walking through this—it helped me a lot. I was wondering, for readability, if there might be a way to make the logic a bit more modular since there's a lot going on with the recursion and inline branching?

pwolfert: I'm not really sure what you mean. Can you give an example of how you might refactor it?

tamara-corbalt: Sure! My idea is to break out the logic that handles the element node cases into separate functions. Here's a rough sketch of what I mean—let me know what you think!
```
function getTextContentFromSlot(node: HTMLSlotElement): string {
  const assignedNodes = node.assignedNodes({ flatten: true });
  return getTextContent(assignedNodes);
}

function getTextContentFromElementNode(node: HTMLElement): string {
  if (node.tagName === 'SLOT') {
    // If it's a slot, it won't actually have any content in itself, so we have to
    // follow the trail to its "assigned nodes" and look there.
    return getTextContentFromSlot(node as HTMLSlotElement);
  }
  return getTextContent(Array.from(node.childNodes));
}

function processNode(node: Node): string {
  if (node.nodeType === Node.TEXT_NODE) {
    return node.textContent;
  } else if (node.nodeType === Node.ELEMENT_NODE) {
    return getTextContentFromElementNode(node as HTMLElement);
  } 

  return '';
}

function getTextContent(nodes: Node[]): string {
  return nodes.map(processNode).join('');
}
```

packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
```diff
@@ -1,7 +1,34 @@
...
 
+/**
+ * Gets text content from an element and even follows `slot` elements across Shadow DOM
+ * barriers to get the full text that the user sees.
+ */
+function getTextContent(nodes: Node[]): string {
+  return nodes
+    .map((node) => {
+      if (node.nodeType === Node.TEXT_NODE) {
+        return node.textContent;
```
tamara-corbalt, line 11: I had another quick question—could `node.textContent` ever be `undefined` or is it safe to assume it will always return a string in this context?

jack-ryan-nava-pbc:
@pwolfert since this recursively descends the DOM, is there any problem with having analytics set on child elements like an accordion item and the parent accordion? I'm thinking the biggest risk could be duplicative information coming from the parent, but I'm not sure.

packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
pwolfert: It should be safe to assume it's always at least an empty string for the `TEXT_NODE` type. I'm pretty sure it's part of the spec, and TypeScript thinks it always has a value.

packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
pwolfert: I see what you mean now! Since none of the functions are meant to be used on their own, I personally prefer having all the logic contained in that one recursive function. Do you feel strongly about it?

pwolfert:
@jack-ryan-nava-pbc, it really has to do with the specific requirements of analytics for each of those components. Right now since we don't have any analytics functionality for accordion and accordion items, it's not a problem _now_, but it could potentially require tweaking in the future. But actually, if you think about it for the accordion and item case, this new behavior would have the same end result as just getting the `.textContent` before if slots weren't involved. The traversal only goes up one level if it encounters a slot, and it technically can't descend to the same place because that parent can't have a circular reference to the same slot, though it can have a reference to a new slot provided by _its_ parent, but it will eventually reach an end and decend until it doesn't find any more slots and terminate.

jack-ryan-nava-pbc (approved):
LGTM!

packages/design-system/src/components/analytics/getAnalyticsContentFromRefs.ts
tamara-corbalt: No, not at all. Your reasoning makes sense to me

PR #3348, opened 2024-12-06
[WNMGDS-3057] Introduce the Shadow DOM to components that need it

pwolfert:
## Summary

The whole reason we were selectively introducing the Shadow DOM in our web components was to fix issues with Angular. This pull request fixes Angular event bindings on nested web components for most of the container-like web components by adding the Shadow DOM to the following web components:

- `<ds-accordion>` and `<ds-accordion-item>`
- `<ds-alert>`
- `<ds-drawer>`
- `<ds-modal-dialog>`
- `<ds-review>`

## How to test

1. You can run the web component unit tests.
2. You can test these in Storybook to make sure they still work as expected.

Note that the updates to the example projects will come in another PR.


packages/design-system/src/components/web-components/ds-accordion/ds-accordion.test.tsx
```diff
@@ -1,35 +1,28 @@
...
+    const accordion = shadowRoot.querySelector('.ds-c-accordion');
     expect(accordion.classList).toContain('ds-u-test');
   });
 
   it('renders ds-c-accordion--bordered class when a bordered prop is set', () => {
-    const { container } = renderAccordion({ bordered: 'true' });
-    const accordion = container.querySelector('.ds-c-accordion');
+    const { shadowRoot } = renderAccordion({ bordered: 'true' });
+    const accordion = shadowRoot.querySelector('.ds-c-accordion');
     expect(accordion.classList).toContain('ds-c-accordion--bordered');
```
tamara-corbalt, line 26: Do you think it would be worthwhile to also assert that an `accordion-item` child receives the correct `ds-c-accordion__content--bordered` style when the parent accordion has the bordered prop set?

jack-ryan-nava-pbc (approved):

packages/design-system/src/components/web-components/ds-accordion/ds-accordion.test.tsx
pwolfert: Oh good catch! I never added any unit tests related to that functionality

pwolfert: Oh dang! Adding a unit test caught a bug. The `ds-accordion` component doesn't have a wrapper component that parses the `bordered` attribute into a boolean, so even values of `bordered="false"` give it a border!

pwolfert: I think it's not happening in Storybook because it uses a boolean control in Storybook and ends up providing the component with a real boolean value in the web component that is rendered by Preact (which is not an expected real-world scenario)

tamara-corbalt (approved):
Looks good!

PR #3349, opened 2024-12-07
[WNMGDS-3057] Update example projects for Shadow-DOM changes

pwolfert:
## Summary

- Updated example projects for Shadow-DOM changes
- Updated the Angular example to showcase the fixes to Angular event binding on nested elements (the whole reason for this feature branch)

## How to test

1. Run the example projects (not that the CDN examples won't work because the changes haven't been released to the CDN)


jack-ryan-nava-pbc (approved):
Weird that the modal doesn't seem to be rendering properly in the snapshots but that seems like a separate issue. Approve!

pwolfert:
@jack-ryan-nava-pbc, that's because these snapshots are of the CDN example, and the new code hasn't been released to the CDN yet

PR #3350, opened 2024-12-09
[WNMGDS-3143] Fix `analytics={false}` being passed in `ds-modal-dialog`

pwolfert:
## Summary

- Fixes `analytics={false}` being passed in `ds-modal-dialog`.
  - Note that we should probably change `parseBooleanAttr` to behave in a more expected way and audit our existing use of it.
- Applies our standard `onAnalyticsEvent` handler [so our proxy function doesn't override our default analytics function](https://github.com/CMSgov/design-system/pull/3296)

## How to test

1. Turn on analytics logging in Storybook and open the dialog in the `ds-modal-dialog` stories. It should log those events to the Actions tab and the console.
2. Run the new unit tests

## Checklist


packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.tsx
```diff
@@ -53,14 +54,14 @@ const Wrapper = ({
     isOpen={parseBooleanAttr(isOpen)}
     size={isPossibleValue(size, availableSizes) ? size : undefined}
     {...otherProps}
-    analytics={parseBooleanAttr(analytics)}
+    analytics={analytics === undefined ? undefined : analytics !== 'false'}
```
jack-ryan-nava-pbc, line 57: Seems like we should consider whether we want analytics to be opt-in, or on by default. Seems like we've been doing on by default, which is why this fails, right?

jack-ryan-nava-pbc: Opt in meaning that users have to pass true to turn on analytics, just to be explicit. If opt in were the case, would they need to supply additional configuration info, or could they use the defaults provided by the config function?

jack-ryan-nava-pbc (approved):
Works locally! Some thoughts and approve!

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.tsx
pwolfert: This line change makes it behave like our React components behave, where providing a value for `analytics` (whether `true` or `false`) overrides the application-level default. We do actually require teams to opt in, but they can either opt in at the application level by calling `config()` and changing the per-component settings or by passing `analytics={true}` to a particular instance of a component.

pwolfert: That is, this should now accurately reflect the [analytics setting behavior we describe here](https://design.cms.gov/components/analytics/?theme=core#turning-it-on-and-off)

pwolfert: I just noticed that the code looks different for [ds-alert and ds-button](https://github.com/CMSgov/design-system/blob/pwolfert/fix-ds-modal-dialog-analytics/packages/design-system/src/components/web-components/ds-alert/ds-alert.tsx#L28) but it should have roughly the same result. Maybe I should change it here to be consistent, though.

PR #3351, opened 2024-12-10
[WNMGDS-3047] Adds test coverage when async data fetching with Autocomplete

kim-cmsds:
## Summary

- https://jira.cms.gov/browse/WNMGDS-3047
- Adds a new story and unit test that demonstrates what happens when fetching async data and while supplying a default list of options to `<Autocomplete />`

## How to test

1. Pull down the branch, run storybook, and take a look at the "Async Items" [story](http://localhost:6006/?path=/story/components-autocomplete--async-items)

## Checklist

- [x] Verified that running both `yarn test:unit` and `yarn test:browser:all` were each successful
- [x] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:all:update`)


tests/browser/snapshots/stories/components-autocomplete--async-items-cmsgov--webkit.png
pwolfert, line 1: Do we need these snapshots? Could they be ignored in the test file?

pwolfert (approved):
This is cool! So does that mean the component already supports what was asked of it?

tests/browser/snapshots/stories/components-autocomplete--async-items-cmsgov--webkit.png
kim-cmsds: That's a good question / point. I'm going to do some snapshot cleanup.

packages/design-system/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -269,5 +272,151 @@ export const NoResults: Story = {
...
+          }));
+          setAdditionalItems(additionalItems);
+        });
+    };
+
+    const debouncedSearch = useCallback(
+      debounce(() => {
+        searchArtwork();
+      }, 500),
+      []
```
jack-ryan-nava-pbc, line 382: So we're getting an eslint[react-hooks/exhaustive-deps](https://github.com/facebook/react/issues/14920) warning here. I don't think we want to track any dependencies because we need the timer to stick around, but can we appease ES lint with a sacrificial comment?

packages/design-system/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -269,5 +272,151 @@ export const NoResults: Story = {
...
+      fetch(searchURL)
+        .then((response) => {
+          if (response.status === 200) {
+            return response.json();
+          }
+        })
+        .then(({ data }: { data: MockedDataResponse[] }) => {
+          const additionalItems = data.map(({ id, title, artist_title }) => ({
+            id: id.toString(),
+            name: `${title} by ${artist_title}`,
```
jack-ryan-nava-pbc, line 372: `artist_title` is coming back undefined. :\

jack-ryan-nava-pbc (approved):
Looks good! Small requests :)

packages/design-system/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -1,10 +1,13 @@
...
 import type * as React from 'react';
-import { Autocomplete, AutocompleteProps } from './Autocomplete';
+import { Autocomplete, AutocompleteItem, AutocompleteProps } from './Autocomplete';
 import TextField from '../TextField/TextField';
 import uniqueId from 'lodash/uniqueId';
 import { action } from '@storybook/addon-actions';
 import type { Meta, StoryObj } from '@storybook/react';
+import { debounce } from '../utilities/debounce';
+
+type AutocompleteArgs = AutocompleteProps & { textFieldLabel: string; textFieldHint: string };
```
tamara-corbalt, line 10: :thumbsup:

packages/design-system/src/components/Autocomplete/Autocomplete.stories.tsx
tamara-corbalt: I pulled down the branch and added some logging. It looks like each item in the `data` array has an `artist_display` property instead of `artist_title`. Adjusting this should fix the issue.

tamara-corbalt (commented):
Left one comment about misnamed property in results, otherwise, looks great!

packages/design-system/src/components/Autocomplete/Autocomplete.stories.tsx
kim-cmsds: This shouldn't matter if the mock was working correctly. Did y'all run `yarn install` and double check that the network request tab isn't hitting the endpoint?

tamara-corbalt (approved):

kim-cmsds:
> This is cool! So does that mean the component already supports what was asked of it?

@pwolfert Yes! The component already supported the use case described in the ticket, so the changes included here are specific test coverage for the use case.

PR #3352, opened 2024-12-11
[WNMGDS-3127] Adds stories to document grouped-item functionality in `ds-autocomplete`

tamara-corbalt:
## Summary

- Added two new Storybook stories (`Item Groups` and `Groups and Standalone Items`) to document the grouped-item functionality of the `ds-autocomplete` component.
- Introduced unit tests to validate the new functionality.

[Jira Ticket](https://jira.cms.gov/browse/WNMGDS-3127)

## How to test

- Run `yarn test:unit:wc`
- Run `yarn storybook` and navigate to the `ds-autocomplete` stories ("Item Groups" & "Groups and Standalone Items").

## Checklist


packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.stories.tsx
```diff
@@ -140,9 +140,24 @@ const Template = (args) => {
...
+        if (item.label && item.items) {
+          // Handle grouped items
+          const filteredGroupItems = item.items.filter(
+            (groupItem) =>
+              !groupItem.name || groupItem.name.toLowerCase().includes(input.toLowerCase())
+          );
+          return {
+            ...item,
+            items: filteredGroupItems,
+          };
```
pwolfert, line 154: Is the intention for a group label to always be visible in the list even if it has no matching items?

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.test.tsx
```diff
@@ -1,8 +1,10 @@
-import { render, waitFor, screen } from '@testing-library/react';
+import { render, screen } from '@testing-library/react';
 import { config } from '../../config';
 import userEvent from '@testing-library/user-event';
 import './ds-autocomplete';
 
+jest.mock('lodash/uniqueId', () => jest.fn((prefix) => `${prefix || 'id'}-test`));
```
pwolfert, line 6: Can this be achieved by passing a `root-id` to the component? Having a static id for the snapshots?

tests/browser/snapshots/stories/web-components-ds-autocomplete--groups-and-standalone-items-cmsgov--Mobile-Chrome.png
pwolfert, line 1: These snapshot tests can probably [be ignored](https://github.com/CMSgov/design-system/blob/tamara/WNMGDS-3127/add-grouped-items-to-autocomplete-wc/tests/browser/stories.test.ts#L14) because they're redundant

packages/design-system/src/components/web-components/ds-autocomplete/__snapshots__/ds-autocomplete.test.tsx.snap
```diff
@@ -20,22 +20,22 @@ exports[`Autocomplete can add a custom class name to the error message 1`] = `
       >
         <label
           class="ds-c-label"
-          for="autocomplete---test"
-          id="autocomplete---test__label"
+          for="autocomplete--145"
+          id="autocomplete--145__label"
```
pwolfert, line 24: Looks like the dynamic ids came back though. Did you try using `root-id`?

tamara-corbalt: Thanks for the nudge! I’ve just updated the tests to use `root-id`, so they should now be using a static value for consistency. :)

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.stories.tsx
tamara-corbalt: Good point! I agree—it would make sense to filter out empty groups.

tests/browser/stories.test.ts
```diff
@@ -32,6 +32,7 @@ const storySkipList = [
   'web-components-ds-accordion-item--default', // Redundant
   'web-components-ds-modal-dialog--default', // Doesn't show the open dialog
   'web-components-ds-modal-dialog--prevent-scroll-example', // Doesn't show the open dialog
+  'web-components-ds-autocomplete--groups-and-standalone-items', // Redundant
 ];
```
pwolfert, line 36: I think the other story can be ignored too since the screenshot doesn't show anything. It'd possibly be valuable to take screenshots of the open menu with interaction tests, but these ones of the inert and empty text field don't seem that valuable.

Also the images that were created for the one you started ignoring are still part of the diff for the pull request. Do you need to delete them still?

pwolfert (commented):
Everything is looking great in the story. Possibly some more opportunities for test refinement, and there seem to be some extra images in the PR ☝️

kim-cmsds (approved):
This looks good as is, but I agree with @pwolfert's concerns around how the snapshots stored in this PR aren't particularly helpful and can be ignored in [stories.test.ts](https://github.com/CMSgov/design-system/blob/main/tests/browser/stories.test.ts#L6). An interaction test might be helpful demonstrating what the results look like when the dropdown menu is open.

tests/browser/stories.test.ts
tamara-corbalt: Ahh, yes! I've removed the unnecessary snapshots and updated this list to include both `item-groups` and `groups-and-standalone-items`

pwolfert (approved):
LGTM 🎉

tamara-corbalt:
> An interaction test might be helpful demonstrating what the results look like when the dropdown menu is open.

Hi @kim-cmsds, I recently added an integration test for the grouped-item functionality in [`Autocomplete`](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/Autocomplete/Autocomplete.test.interaction.ts#L26-L41). My assumption is that this should be sufficient given this shared functionality, but do you think dedicated integration tests for `ds-autocomplete` would still be useful?

kim-cmsds:
@tamara-corbalt despite the shared functionality, I would like to see the `ds-autocomplete`-specific snapshots for the grouped items. I could see those being useful to compare style changes.

kim-cmsds:
@tamara-corbalt I'm walking back my earlier comment about needing the interaction tests. It looks like the functionality is covered in `ds-autocomplete.test.tsx`.

PR #3353, opened 2024-12-13
[NO-TICKET] Change the failure threshold for VRT screenshots

pwolfert:
## Summary

We've experienced some false positives in lately in the visual regression tests running in our CI environment. When actually reviewing those expected and actual images, there were no perceptible differences. Even running it through a separate image diffing tool revealed no differences. While we still don't know the root cause of these false positives, we're seeing with this pull request if upping the threshold to allow for one pixel difference (`maxDiffPixels: 1`) will filter out these false positives. One pixel shouldn't be high enough to, on the flip side, miss real visual regressions.

- Cleans up the integration test modules with some common helper functions
- Creates a new `expectScreenshot` helper function that applies our threshold override and common options
- Replaces all bare uses of `expect(page).toHaveScreenshot` with the above helper function

### If we continue to see the same problem in the future...

If we continue to see the same problem in the future, it could mean a couple things:

1. That threshold is not enough.
    - This seems unlikely since the tool is not actually reporting that any pixels are different. If they were, the diff would show red pixels.
2. Something else is going on.
    - It could be that some other failure is happening during the comparison ([this is about where it's happening in Playwright's source](https://github.com/microsoft/playwright/blob/e995ecd9b8c8c2578b7af9eee92997ad8a841a0d/packages/playwright-core/src/utils/comparators.ts#L76-L89) but that Playwright's error handling is not set up to distinguish between a failure to match and another, unknown error. Our CI environment being less reliable than our dev machines could mean that the tool times out or something.

## How to test

1. Run tests

## Checklist


tests/browser/interactionHelpers.ts
```diff
@@ -0,0 +1,20 @@
+import { test } from '@playwright/test';
+import themes from '../../themes.json';
+
+export { expectScreenshot } from './expectScreenshot';
+
+export function sleep(ms) {
```
kim-cmsds, line 6: Nice, glad to see a robust set of shared utility functions for tests.

What do you think about running a search in the codebase for `const sleep` or function sleep` and replacing those inline functions with this shared function?

tests/browser/expectScreenshot.ts
```diff
@@ -0,0 +1,12 @@
+import { expect, Page } from '@playwright/test';
+
+export function expectScreenshot(page: Page, filename: string, options: any = {}) {
+  return expect(page).toHaveScreenshot(filename, {
+    fullPage: true,
+    // We've been getting odd results in our CI tests where the two images are exactly
+    // the same in every way, but Playwright is failing the comparison. We'll try setting
+    // the `maxDiffPixels` to 1 and see if we still get those kinds of false positives.
+    maxDiffPixels: 1,
```
kim-cmsds, line 9: 👍

kim-cmsds (approved):
This looks like a positive step forward for removing some of those flakey VRTs.

tests/browser/interactionHelpers.ts
pwolfert: The other local definitions for sleep functions that I can find are in unit tests. Are you suggesting we share a helper module between browser tests and unit tests?

kim-cmsds: Thanks for pointing that out. No, let's not create a helper module between those types of tests. That work's outside the scope of this PR.

pwolfert:
Looks like we're still having the same problem: https://ci.backends.cms.gov/wds/blue/organizations/jenkins/design-system%2Ftests/detail/PR-3356/2/artifacts

PR #3356, opened 2024-12-17
[WNMGDS-2928] Standardize `inputRef` prop types

pwolfert:
## Summary

WNMGDS-2928

- Standardizes `inputRef` prop types to use the React `Ref` type, which covers both Object refs and ref functions
- Adds unit tests to make sure both kinds of refs work

## How to test

1. Run unit tests
2. Try one of the ChoiceList stories and verify that the `onComponentBlur` event still works

## Checklist


kim-cmsds (approved):
Pulled this down locally and ran through the ChoiceList stories and unit tests ✅ 

Looks like the storybook docs need to be rebuilt for tests to pass.

PR #3357, opened 2024-12-17
[WNMGDS-3077] Fix wrong Storybook iframes loading on doc site

pwolfert:
## Summary

Apply the client-side-only fix to story iframes to fix rehydration issue. This is the same workaround that we used in [#2962](https://github.com/CMSgov/design-system/pull/2962/files). Making it client-side only means there is no mismatched HTML for Gatsby to try to reconcile, so it'll always show the right one. There's really no reason to pre-render that component anyway.

Unfortunately this fix does not retroactively fix earlier versions of the doc site, so if we want to fix older archived versions, we have to go back and apply the fix to those versions and redeploy.

## How to test

To reproduce the original issue on the main branch:

1. `yarn build:docs`
2. `yarn serve:docs`
3. Navigate to http://localhost:9000/components/alert/?theme=medicare
4. Take a look at the first storybook example. Notice that even though this is the Medicare theme, it shows the info alert of the core theme. If you inspect the iframe in the page, you'll notice its url is wrong. It's pointing to the core theme in Storybook even though the client-side JavaScript that loads knows the correct theme and url to load. It's that Gatsby doesn't correctly reconcile the statically generated HTML with the new HTML being rendered in the client.

## Checklist


tamara-corbalt (approved):
LGTM! Approve, pending passing CI.

PR #3361, opened 2024-12-23
[WNMGDS-2928] Follow up to improve the standardized `inputRef` types

pwolfert:
## Summary

The types I added previously introduced in #3356 didn't cover mutable refs, which are what you get when you use `useRef`. I guess the immutable `Ref` types are what you get when you use `createRef`. I thought I had read the type definitions thoroughly before, but I missed that the `MutableRefObject` was missing. And I didn't understand the difference in the returned ref type between those two functions until now.

Also, if you look at the types we were using before [my last change](https://github.com/CMSgov/design-system/pull/3356/files), they used a lot of `any` types. I'm nervous that if we start introducing really specific types like `React.Ref<HTMLInputElement | HTMLTextAreaElement>`, we're going to start to get a lot of complaints from down stream.

- Change all of the standardized `inputRef` type definitions to accept `MutableRefObject` as well
- Replace the really specific HTML element types with `any` to be maximally compatible rather than precisely correct. We don't really know how they've typed their `TextField` `inputRef`s, and I think we should err on the side of flexibility in this case.

## How to test

Try providing some `inputRef`s in stories.

## Checklist

- [x] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:all:update`)

### DIFF
--- a/packages/design-system/src/components/Button/Button.test.tsx
+++ b/packages/design-system/src/components/Button/Button.test.tsx
@@ -2,7 +2,7 @@ import Button from './Button';
 import { UtagContainer } from '../analytics';
 import { config } from '../config';
 import { fireEvent, render, screen } from '@testing-library/react';
-import { createRef } from 'react';
+import { createRef, useEffect, useRef } from 'react';
 
 const defaultProps = {
   children: 'Foo',
@@ -88,6 +88,19 @@ describe('Button', () => {
     renderButton({ inputRef });
     expect(inputRef.current).toBeInTheDocument();
     expect(inputRef.current.tagName).toEqual('BUTTON');
+  });
+
+  it('forwards a mutable object inputRef', () => {
+    const MyComponent = () => {
+      const inputRef = useRef<HTMLButtonElement>();
+      useEffect(() => {
+        expect(inputRef.current).toBeInTheDocument();
+        expect(inputRef.current.tagName).toEqual('BUTTON');
+      }, []);
+      return <Button inputRef={inputRef}>Hello world</Button>;
+    };
+
+    render(<MyComponent />);
   });
 
   it('forwards a function inputRef', () => {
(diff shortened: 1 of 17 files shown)


tamara-corbalt (commented):
Looks good to me! I noticed that some of these components have tests verifying that we forward inputRef objects with `createRef` ([example](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/Button/Button.test.tsx#L86-L91)). Should we consider adding assertions to cover both immutable and mutable object types?

packages/design-system/src/components/Button/Button.test.tsx
```diff
@@ -90,6 +90,19 @@ describe('Button', () => {
     expect(inputRef.current.tagName).toEqual('BUTTON');
   });
 
+  it('forwards a mutable object inputRef', () => {
+    const MyComponent = () => {
+      const inputRef = useRef<HTMLButtonElement>();
+      useEffect(() => {
```
tamara-corbalt, line 96: I'm curious about the usage of `useEffect` in tests, as I haven't encountered it much before. Would using `getByRole` to locate the element and then asserting the `ref.current` value work in this case? Or is there a specific reason we need `useEffect` here instead?

pwolfert: I didn't think I could access that value outside my test component, so that's why I put my assertion in a useEffect. I guess I could `let inputRef` define it outside of `MyComponent`, but I didn't know if that would produce unexpected results. Seemed the most straightforward to put the assertion in the normal render cycle of `MyComponent`

tamara-corbalt: Ok, that make sense!

tamara-corbalt (approved):
Approve, pending CI

PR #3362, opened 2024-12-23
[WNMGDS-3147] Enable `--list` for playwright custom playwright reporter

tamara-corbalt:
## Summary
This PR introduces the following improvements to the custom reporter:
- Adds a `--list` flag to display test hierarchy paths without executing tests.
- Implements logic to track and display the current test index in the logged output (e.g., `[1/160]`).
- Formats browser names with square brackets (`[]`) in the output for improved readability.

[Jira Ticket](https://jira.cms.gov/browse/WNMGDS-3147)

## How to test

1. Run tests using the `--list` flag and a `grep` pattern to filter specific test cases. 
**Example:**`% yarn test:browser --grep="ds-autocomplete" --list `
2. Disable the custom reporter in `playwright.config.ts`, run the tests and compare the logged output with the custom reporter’s output.
3. Run tests without the `--list` flag both with and without the custom reporter enabled. Check that all combinations of browsers and themes are logged as expected, including the test index format (e.g., `[1/160]`).

## Checklist

### DIFF
--- a/tests/browser/custom-reporter.ts
+++ b/tests/browser/custom-reporter.ts
@@ -16,6 +16,8 @@ class MyReporter implements Reporter {
   private failingTests: { path: string; name: string }[] = [];
   private skippedTests: { path: string; name: string }[] = [];
   private isListMode = false;
+  private totalTests = 0;
+  private currentTestIndex = 0;
 
   private getHierarchyPath(test: TestCase): string {
     const path: string[] = [];
@@ -45,8 +47,12 @@ class MyReporter implements Reporter {
       });
       return;
     }
-
+    this.totalTests = suite.allTests().length;
     console.log(`Starting the run with ${suite.allTests().length} tests`);
+  }
+
+  onTestBegin(_: TestCase) {
+    this.currentTestIndex++;
   }
 
   onTestEnd(test: TestCase, result: TestResult) {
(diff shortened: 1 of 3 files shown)


tests/browser/custom-reporter.ts
```diff
@@ -47,10 +74,14 @@ class MyReporter implements Reporter {
...
     }
 
-    console.log(`${hierarchyPath} > ${test.title}: ${result.status}`);
+    console.log(
+      `[${this.currentTestIndex}/${this.totalTests}] ${hierarchyPath} > ${test.title}: ${result.status}`
+    );
   }
 
   onEnd(result: FullResult) {
+    if (this.isListMode) return;
```
pwolfert, line 83: I don't know how much work it would be, but the thing I found most useful in the `--list` feature of the default reporter was telling me how many tests would be run. It looks liket his:

Screenshot 2024-12-23 at 11 11 59 AM

I use the list command to figure out how many tests are run in the smoke tests vs the nightly tests, how many actual test variations are executed for a given change (all the permutations of a story with themes and browsers), etc.

tamara-corbalt: Thanks, that's good to know! I've added the total number of tests and test files to the summary output at the end, matching the default reporter's format. 

Are the permutations of story, theme, and browser readable enough? It looks to me like we're listing everything the same as the default reporter, but let me know if I've missed any nuances.

pwolfert: This is great!

pwolfert (approved):
LGTM

PR #3366, opened 2025-01-03
[WNMGDS-2749] Update `Alert` to dynamically assign `role` based on `variation`

tamara-corbalt:
## Summary

Updated `Alert` to dynamically assign role based on the variation type unless explicitly overridden by the user. Added unit tests to verify the new behavior and updated existing `ds-alert` tests impacted by the change.

## How to Test

1. Run `yarn storybook` and use VoiceOver or your preferred screen reader to verify that the `role` is correctly announced based on the `variation`.
2. Run the unit tests with `yarn test:unit`.
3. Run the web component-specific unit tests with `yarn test:unit:wc`.

## Checklist


packages/design-system/src/components/Alert/Alert.tsx
```diff
@@ -167,7 +166,6 @@ export const Alert = (props: AlertProps) => {
 };
 
 Alert.defaultProps = {
-  role: 'region',
```
kim-cmsds, line 170: It seems like the logic to default to `region` for the `<Alert />` role remains as part of the definition of `dynamicRole`, but is no longer reflected in the storybook docs table. Did you consider keeping this line and changing how the variable is defined on L132?

kim-cmsds (approved):
Pulled this down locally and ran through storybook without issue 👍

packages/design-system/src/components/Alert/Alert.tsx
```diff
@@ -122,31 +122,30 @@ export const Alert = (props: AlertProps) => {
...
-      case 'error':
-        return <AlertCircleIcon className={iconClass} />;
-      case 'success':
-        return <CheckCircleIcon className={iconClass} />;
-      case 'warn':
-        return <WarningIcon className={iconClass} />;
-      default:
-        return <InfoCircleIcon className={iconClass} />;
-    }
+    const IconComponent = variationConfig[variation]?.icon || variationConfig.default.icon;
```
jack-ryan-nava-pbc, line 139: Very succinct! Love this implementation.

packages/design-system/src/components/web-components/ds-alert/ds-alert.test.tsx
```diff
@@ -20,8 +20,8 @@ const renderAlert = createTestRenderer('ds-alert', (attrs = {}) => (
   <ds-alert {...attrs}>{defaultText}</ds-alert>
 ));
 
-function expectHasClass(shadowRoot: ShadowRoot, className: string) {
-  expect(getByRole(shadowRoot as any as HTMLElement, 'region').className).toContain(className);
+function expectHasClass(shadowRoot: ShadowRoot, className: string, role = 'region') {
```
jack-ryan-nava-pbc, line 23: Side quest/feel free to mull over in your future travels:
Is this function worth extracting into a separate module? We're repeating with slight variation, is something like this used elsewhere?

jack-ryan-nava-pbc (approved):
This looks good to me! One question: have you tested this out with VoiceOver? I think it would be good to confirm that the role changes are appropriately read out by assistive tech.

packages/design-system/src/components/web-components/ds-alert/ds-alert.test.tsx
tamara-corbalt: @jack-ryan-nava-pbc - I like that idea! I did a quick search of the codebase, and it seems this function is only used in two places: Alert.test.ts and ds-alert.test.ts. Given the limited usage, it might not yet justify extraction into a separate module, but I’ll keep an eye out for similar patterns elsewhere.

tamara-corbalt:
> This looks good to me! One question: have you tested this out with VoiceOver? I think it would be good to confirm that the role changes are appropriately read out by assistive tech.

@jack-ryan-nava-pbc  - I tested it with VoiceOver, and the behavior was as I expected. That said, I’d love to hear if you’ve had a chance to test it as well or if you notice anything I might have missed!

packages/design-system/src/components/Alert/Alert.tsx
kim-cmsds: I like the idea of additional documentation if nothing else!

Is this too simple? `const dynamicRole = role ?? (variationConfig[variation]?.role || Alert.defaultProps.role);`

tamara-corbalt: I see where you’re coming from with referencing `Alert.defaultProps.role`. The issue I see with this approach is that it defaults the role to 'region' (from `defaultProps`) whenever the user doesn’t provide a `role`, even if a `variation` is present and should determine the `role`.

kim-cmsds: Yep! I'm missing that part of the equation. Ok, perfect, let's go with your initial idea to expand the documentation and ensure that link (how variation might dictate role) is clear.

tamara-corbalt: Cool! I just pushed up a change updating documentation for the `role` prop in storybook. Let me know what you think!

kim-cmsds (approved):

PR #3367, opened 2025-01-06
[NO-TICKET] Skip test combinations that trigger a Playwright bug

pwolfert:
## Summary

These Autocomplete screenshot tests are randomly failing in forced-colors mode because of [an old bug in Playwright](https://github.com/microsoft/playwright/issues/15211) that hasn't been fixed. I'm therefore skipping this test in forced-colors mode so we don't experience the flaky false positives. We've had to do this before in https://github.com/CMSgov/design-system/pull/2856.

## Checklist


kim-cmsds (approved):
Appreciate the context and comments!

pwolfert:
Tests are failing on the Drawer interaction tests because it's taking a screenshot half-way through the opening animation. [Playwright is supposed to be disabling animations by default](https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-2-option-animations). It's never been a problem before, and I wonder if it was the recent Playwright update.

pwolfert:
I can't figure out why the drawer test is behaving that way, so I'm adding a workaround in this PR.

packages/design-system/src/components/Drawer/Drawer.test.interaction.ts
```diff
@@ -11,7 +11,10 @@ describeByTheme((theme) => {
     await page.goto(storyUrl('components-drawer--default', theme));
     const elem = page.getByRole('button');
     await elem.click();
-    await sleep(100);
+    // Work what might be a Playwright bug. Animations are supposed to be disabled by
+    // default when taking a screenshot, but it doesn't seem to be working for this
+    // component. Animation takes 300ms.
+    await sleep(320);
     await expectScreenshot(page, `drawer--open--${theme}.png`);
```
kim-cmsds, line 18: @pwolfert reading through the [docs](https://playwright.dev/docs/api/class-pageassertions#page-assertions-to-have-screenshot-2-option-animations), it looks like animations are supposed to be disabled by default. What do you think about updating the helper `expectScreenshot` method to explicitly pass in `animations: 'disabled'` to the `options` object?

pwolfert: Unfortunately [I tried that](https://github.com/CMSgov/design-system/pull/3367/commits/d3025a450b90e47990ab7c67c661fd6580654b92), and it didn't work.

kim-cmsds: Shoot!

kim-cmsds (approved):

PR #3379, opened 2025-01-09
[NO-TICKET] Update the description of the design system package dist folder

pwolfert:
## Summary

Update the description of the design system package dist folder to match the current state of the build output.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -23577,10 +23577,15 @@ react-dev-utils@^12.0.1:
     strip-ansi "^6.0.1"
     text-table "^0.2.0"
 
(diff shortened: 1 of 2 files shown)


jack-ryan-nava-pbc:
Thank you!

What's going on with the yarn.lock changes? @kim-cmsds and I have been running into these as well. Seems innocuous enough however.

jack-ryan-nava-pbc (approved):

yarn.lock
pwolfert, line 1: Oh whoops, I didn't notice that this changed or that I committed it. Good catch, @jack-ryan-nava-pbc. I'm not sure what the origin of this is. Did you and Kim have the same changes? If so, it might have just been that someone updated something and forgot to commit the lockfile change.

PR #3393, opened 2025-01-15
[NO-TICKET] Fix `hintClassName` being applied to DOM elements

pwolfert:
## Summary

Fixes an error found by the TextField "applies custom hintClassName to the hint element" unit test where React is complaining that the `hintClassName` is being applied directly to a DOM element.

## How to test

Run `yarn test:unit packages/design-system/src/components/TextField/TextField.test.tsx` and make sure no errors are printed to the console.

## Checklist


jack-ryan-nava-pbc (approved):
Warning goes away when this line is added! Nice catch

PR #3394, opened 2025-01-16
[NO-TICKET] Preemptively fix CORS blocking stylesheet access in web component CDN example

pwolfert:
## Summary

In anticipation of the next web components update, add `crossorigin="anonymous"` to the stylesheets in the cdn example.

We found that the code that copies document styles into the Shadow DOM for web components won't work if CORS blocks the stylesheets. Teams will be pulling these styles from a trusted source (`design.cdn.gov`), so we don't need to restrict JavaScript's access to the rules in the stylesheet. Here's the error we get when this is not enabled and we're using the new web component code:

```
Security Error in ds-accordion :
Uncaught (in promise) SecurityError: Failed to read the ‘cssRules’ property from ‘CSSStyleSheet’: Cannot access rules
    at globalStyles.js:14:42
    at Array.map (<anonymous>)
    at Y (globalStyles.js:12:43)
    at r.U (globalStyles.js:6:20)
    at r.connectedCallback (define.js:79:19)
    at z (define.js:27:20)
    at 3595 (ds-accordion-item.js:41:1)
    at n (bootstrap:19:32)
    at make namespace object:6:63
    at make namespace object:6:63
```

## How to test

1. Build the core package and copy the `all.js` web component bundle into the `examples/cdn-web-components` folder
2. Replace the script import in the example to point to `all.js`
3. Run `yarn http-server` from that directory and visit http://localhost:8080
4. Without this change, there will be a JavaScript error. With this change, there won't

## Checklist


jack-ryan-nava-pbc (approved):
LGTM!

PR #3396, opened 2025-01-17
[WNMGDS-3164] Add analytics tracking for documentation searches

pwolfert:
## Summary

WNMGDS-3164

Some of the fields are still being discussed in a ticket with the analytics team, but the functionality is here.

## How to test

1. `yarn start`
2. Open http://localhost:8000/
3. In the JavaScript console, execute `window.utag.link = console.log` to be able to intercept the analytics events and see them printed to the console.
4. In your console settings, enable the preserving of logs between navigation requests. We'll be performing actions that both send analytics and go to a new page, so we don't want the page refresh to clear the console before we can see the event printed
5. Type something into the doc-site search bar and click the search button. You should see an analytics event logged in the console. This will take you to the search page, and you might need to execute the `utag.link` mock again so you can see the next events.
7. On the search page, try changing the search term in the text box and watch it both update the results in the page and trigger an analytics event
8. Try clicking on one of the search result links. This should print a different event to the console.

## Checklist


packages/docs/src/pages/search.tsx
```diff
@@ -57,16 +60,34 @@ const SearchPage = ({ location }: MdxQuery) => {
...
               let body = result.body;
               const strLoc = body.toLowerCase().indexOf(query.toLowerCase());
               body = body.slice(Math.max(strLoc - 160, 0), strLoc + 160);
               const re = new RegExp(query, 'gi');
               body = body.replace(re, `<mark>${query}</mark>`) + '...';
 
+              const sendAnalyticsEvent = () => {
+                sendLinkEvent({
+                  event_name: 'search_result_engaged',
+                  search_result_count: results.length,
```
jack-ryan-nava-pbc, line 73: This needs to be a string

pwolfert: Good catch

pwolfert: Fixed

jack-ryan-nava-pbc:
I see the `search_initiated` event in the console, however I am not seeing the `search_result_engaged` event when I click on a link

jack-ryan-nava-pbc:
Ah, I had to rerun the `window.utag.link = console.log` after executing the search

PR #3401, opened 2025-01-23
[WNMGDS-3201] Add link tracking events to doc site

jack-ryan-nava-pbc:
## Summary

[Ticket](https://jira.cms.gov/browse/WNMGDS-3201)

## How to test

- `yarn start`
- Open http://localhost:8000/
- In the JavaScript console, execute window.utag.link = console.log to be able to intercept the analytics events and see them printed to the console.
- In your console settings, enable the preserving of logs between navigation requests. We'll be performing actions that both send analytics and go to a new page, so we don't want the page refresh to clear the console before we can see the event printed
- Click on some links both internal and external
- You should see the content of the object provided to the link function printed in the console. It should look similar in shape to this:
```
event_name: "internal_link_clicked",
link_url: https:/design.cms.gov/etc,
link_type: 'link_other',
parent_component_heading: "Some text from the heading"
parent_component_type: "P"
text: "The link text",
```

## Checklist

### DIFF
--- a/packages/docs/src/components/layout/SideNav/SideNav.tsx
+++ b/packages/docs/src/components/layout/SideNav/SideNav.tsx
@@ -13,6 +13,7 @@ import {
 } from '../../../helpers/navDataFormatUtils';
 import GithubIcon from '../../icons/GithubIcon';
 import NewsIcon from '../../icons/NewsIcon';
+import linkAnalytics from '../../../helpers/linkAnalytics';
 
 interface SideNavProps {
   location: LocationInterface;
@@ -22,7 +23,7 @@ interface SideNavProps {
  * A wrapper for the gatsby link to handle internal site navigation
  */
 const GatsbyLink = (props: any /* See VerticalNavItemLabel.tsx */) => (
-  <Link to={props.href} {...props}>
+  <Link onClick={linkAnalytics} to={props.href} {...props}>
     {props.children}
   </Link>
 );
@@ -120,9 +121,9 @@ const SideNav = ({ location }: SideNavProps) => {
           )}
         </Button>
         <div>
-          <a className="c-navigation__title" href="/">
+          <Link onClick={linkAnalytics} className="c-navigation__title" to="/">
             CMS Design System
-          </a>
+          </Link>
         </div>
       </header>
 
@@ -143,13 +144,18 @@ const SideNav = ({ location }: SideNavProps) => {
             selectedId={location ? location.pathname : ''}
           />
           <p>
-            <Link to="/blog/" className="c-navigation__bottom-link ds-c-link">
+            <Link
+              onClick={linkAnalytics}
+              to="/blog/"
+              className="c-navigation__bottom-link ds-c-link"
+            >
               <NewsIcon />
               What&apos;s new?
             </Link>
           </p>
           <p>
             <a
+              onClick={linkAnalytics}
               href="https://github.com/CMSgov/design-system"
               className="c-navigation__bottom-link ds-c-link"
             >
--- a/packages/docs/src/helpers/linkAnalytics.ts
+++ b/packages/docs/src/helpers/linkAnalytics.ts
@@ -1,0 +1,27 @@
+import { sendLinkEvent } from '@cmsgov/design-system';
+
+type LinkType = 'internal' | 'external';
+
+function composeAnalyticsEvent(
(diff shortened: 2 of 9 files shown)


packages/docs/src/helpers/linkAnalytics.ts
```diff
@@ -0,0 +1,27 @@
+import { sendLinkEvent } from '@cmsgov/design-system';
+
+type LinkType = 'internal' | 'external';
+
+function composeAnalyticsEvent(
+  event: React.MouseEvent<HTMLAnchorElement, MouseEvent>,
+  type: LinkType
```
pwolfert, line 7: Looks like this isn't actually used anywhere. Will you be using it in the future?

packages/docs/src/helpers/linkAnalytics.ts
```diff
@@ -0,0 +1,27 @@
...
+) {
+  return {
+    event_name: `${type}_link_clicked`,
+    link_url: (event.target as HTMLAnchorElement).baseURI,
+    link_type: 'link_other',
+    parent_component_heading:
+      (event.target as HTMLAnchorElement).parentElement.innerText ?? undefined,
+    parent_component_type: (event.target as HTMLAnchorElement).parentElement.tagName ?? undefined,
```
pwolfert, line 15: Is this really what the analytics team wants? The "heading" part doesn't quite make sense, especially when it's being used in paragraphs. I believe it, but it seems strange. Here's an example of values from one of the events:

```
parent_component_heading:  "The design system is a set of open source design and front-end development resources for creating Se"
parent_component_type:  "P"
```

packages/docs/src/components/layout/SideNav/SideNav.tsx
```diff
@@ -22,7 +23,7 @@ interface SideNavProps {
  * A wrapper for the gatsby link to handle internal site navigation
  */
 const GatsbyLink = (props: any /* See VerticalNavItemLabel.tsx */) => (
-  <Link to={props.href} {...props}>
+  <Link onClick={linkAnalytics} to={props.href} {...props}>
```
pwolfert, line 26: It looks like the `onClick` for the nav items are being clobbered by `VerticalNav` [here](https://github.com/CMSgov/design-system/blob/jryan/WNMGDS-3201-link-tracking-events/packages/design-system/src/components/VerticalNav/VerticalNav.tsx#L81). I think if this property were to be directly applied to the individual `navItems` it might work out of the box. Oh, but it might be easier to just check for `props.onClick` in this `GatsbyLink` component and make sure to call both handlers if one exists.

pwolfert (commented):
Most of the links so far a working. Trying to be thorough in my testing. I found one dead link on our home page that I'll make a pull request for.

packages/docs/src/components/layout/SideNav/SideNav.tsx
pwolfert: I guess technically the clobbering is happening by this component with the spread operator, but it's the `VerticalNav` component that is passing its own `onClick` to the `<VerticalNavItem>` instance, which is coming into this render function above in the props.

packages/docs/src/helpers/linkAnalytics.ts
jack-ryan-nava-pbc: Nope! Checked that all external links are handled via the MDX transform to the ThirdPartyExternal link component.

packages/docs/src/helpers/linkAnalytics.ts
jack-ryan-nava-pbc: I mean, it's in the tech spec. The character limit/truncation is imposed by our code, so I could follow up and make sure that they're ok with it. I'm also doing active fact finding about this as we go, so I can confirm.

packages/docs/src/components/layout/SideNav/SideNav.tsx
jack-ryan-nava-pbc: I think I acknowledged this in this [commit](https://github.com/CMSgov/design-system/pull/3401/commits/28df6f2ba8d8e0c24b008ff2d61360f6aa912913)

pwolfert: It works now!

pwolfert (approved):
Works great!

PR #3405, opened 2025-01-27
[WNMGDS-3166] Add eMedicare footer web-component

tamara-corbalt:
## Summary

- Adds the `ds-simple-footer` module, along with its Storybook file and unit tests.

**Note:** `ds-simple-footer` is the first web component introduced within one of the child design systems. To establish this new pattern, a few unexpected changes were required:

- A deeply nested relative path is used for the `define` function, which resides in the core design system package. To ensure Webpack recognizes this path at build time, we implemented an alias in the Webpack configuration.
- We explicitly set `rootDir` to `'.'` in `tsconfig.json` to reference source files.

While the project now builds successfully, and all web components (including `ds-simple-footer`) render and behave as expected in Storybook, with all unit tests passing, there are pre-runtime TypeScript issues in web component test and story files.

**Error:** `'React' refers to a UMD global, but the current file is a module.` These issues can mostly be resolved by adding an explicit `import React from 'react';` to the affected files.

### Call for Feedback
To resolve these TypeScript issues, I’ve started adding `import React from 'react';` to web component files as needed—for example, in the modules for `ds-alert`, `ds-accordion`, and `ds-simple-footer`. However, I’m wondering if there’s a more efficient way to address this issue instead of modifying each file individually.

## How to Test

1. Run `yarn build` to ensure the project builds successfully without errors.
2. Start Storybook with `yarn storybook` and verify that the `ds-simple-footer` web component renders and behaves as expected.
3. Run unit tests: `yarn test:unit:wc`
4. Inspect the web component's Storybook stories and test files that haven’t been modified yet to observe the TS error (`'React' refers to a UMD global...`) as mentioned above.

## Checklist


tsconfig.json
```diff
@@ -10,5 +11,33 @@
     "baseUrl": ".",
     "skipLibCheck": true
   },
-  "include": ["packages/**/src/**/*", ".storybook/**/*"]
+  "include": ["packages/**/src/**/*", ".storybook/**/*"],
+  "exclude": [
+    "**/*.test.js",
+    "**/*.test.jsx",
+    "**/*.test.ts",
```
pwolfert, line 18: Did this end up working? I can't remember, was it failing at `compileTypescriptDefs` without this? I tried removing the excludes and rebuilding, and it worked.

packages/design-system/src/components/web-components/ds-accordion/ds-accordion.test.tsx
```diff
@@ -17,30 +18,34 @@ describe('ds-accordion', () => {
...
   it('does not render border classes when border prop is not set', () => {
     const { shadowRoot } = renderAccordion({ bordered: 'false' });
     const accordion = shadowRoot.querySelector('.ds-c-accordion');
     expect(accordion).not.toHaveClass('ds-c-accordion--bordered');
-    const items = accordion.querySelectorAll('ds-accordion-item');
-    for (const item of items) {
-      const shadowRoot = item.shadowRoot;
-      const contentEl = shadowRoot.querySelector('.ds-c-accordion__content');
-      expect(contentEl).not.toHaveClass('ds-c-accordion__content--bordered');
+    const items = accordion?.querySelectorAll('ds-accordion-item');
```
jack-ryan-nava-pbc, line 42: What's up with all these changes to the stories? They seem fine and innocuous, but out of scope?

tsconfig.json
tamara-corbalt: I just tried commenting this out and was able to build the project and run Storybook without any issues. It seems safe to remove this.

packages/design-system/src/components/web-components/ds-accordion/ds-accordion.test.tsx
```diff
@@ -1,3 +1,4 @@
+import React from 'react';
```
jack-ryan-nava-pbc, line 1: Is this because typescript is mad?

packages/design-system/src/components/web-components/ds-accordion/ds-accordion.test.tsx
tamara-corbalt: @jack-ryan-nava-pbc - Totally agree, these changes are out of scope. After committing these changes in ([eb414f917](https://github.com/CMSgov/design-system/pull/3405/commits/eb414f917af049dd332f650cdacb218dfe06835a)), I noticed TS pre-runtime issues in the test and story files across our entire web components library—not just in the new `ds-simple-footer`, but affecting components like `Accordion` and `Alert` as well. I referenced these issues in the PR description above but included these changes to illustrate what might be needed to resolve them. Oddly, even without resolving these issues, tests pass, the project builds successfully and components render in Storybook. This makes me suspect the recent changes to webpack and tsconfig might be contributing to these pre-runtime issues. cc: @pwolfert

packages/ds-medicare-gov/src/components/web-components/ds-simple-footer/ds-simple-footer.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+    // Adding custom event listeners to links prevents immediate navigation, allowing us to log and verify
+    // the `ds-click-link-analytics` event in storybook actions before the browser navigates away.
+    const links = footer?.querySelectorAll('a');
+    links?.forEach((link) => {
+      link.setAttribute('target', '_blank');
+      link.addEventListener('click', (e) => {
+        e.preventDefault();
+
+        setTimeout(() => {
+          window.open(link.href, '_blank');
```
jack-ryan-nava-pbc, line 71: I think we could leave out the timeout here, do you think it's necessary to actually do the navigation?

packages/ds-medicare-gov/src/components/web-components/ds-simple-footer/ds-simple-footer.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+          'For information about how and when to use this component, [refer to its full documentation page](https://design.cms.gov/components/footer/medicare-footer).',
+      },
+      componentEvents: {
+        'ds-link-click-analytics': {
+          description: 'A callback function triggered when the user clicks a link in the footer.',
+        },
+      },
+    },
+  },
+  argTypes: {
```
jack-ryan-nava-pbc, line 27: Why are you leaving out language? Do you think the addition would be confusing?

packages/ds-medicare-gov/src/components/web-components/ds-simple-footer/ds-simple-footer.stories.tsx
```diff
@@ -0,0 +1,95 @@
...
+    'about-medicare-label': {
+      description: 'Label for the "About" link.',
```
jack-ryan-nava-pbc, line 29: It'd be great to get the default values that are specified in the React component to display in the docs.

packages/ds-medicare-gov/src/components/web-components/ds-simple-footer/ds-simple-footer.stories.tsx
jack-ryan-nava-pbc: It's a little weird when you're in storybook. You click on a link and nothing happens, so you start doing other stuff, and then all of sudden another tab opens.

packages/design-system/src/components/web-components/ds-accordion/ds-accordion.test.tsx
jack-ryan-nava-pbc: The safe navigation operator isn't strictly necessary here, since we _know_ we are adding those elements to the DOM. If they don't get added to the DOM, something has gone wrong with Storybook. So I think you're correct in saying that the tsconfig might be contributing to these warnings.

tsconfig.json
jack-ryan-nava-pbc: > The build output for ds-simple-footer in the Medicare package appears the same whether or not the exclude list is present in tsconfig.json, so I think it's safe to remove. Is there something specific about the bundles you'd like me to investigate further?

No I think that is sufficient! Thanks for checking.

> That said, I noticed something curious: regardless of the exclude list, ds-simple-footer is included in all.js but doesn’t have its own individual bundle like other web components in the core design system.

We should figure out what's going on there. We should look at our a la carte bundling process for web components.

packages/ds-medicare-gov/src/components/web-components/ds-simple-footer/ds-simple-footer.stories.tsx
tamara-corbalt: No, I think I simply forgot to add it.

packages/ds-medicare-gov/src/components/web-components/ds-simple-footer/ds-simple-footer.stories.tsx
tamara-corbalt: Yeah, in hindsight, I think it's weird too. I'll change it

tsconfig.json
tamara-corbalt: @jack-ryan-nava-pbc  - The good news is, removing the exclude list from `tsconfig.json` appears to have resolved the TS pre-runtime issues I was seeing across web component modules. As a result, I was able to cleanly revert the out-of-scope changes in `ds-accordion` and `ds-alert`.

tamara-corbalt: > > That said, I noticed something curious: regardless of the exclude list, ds-simple-footer is included in all.js but doesn’t have its own individual bundle like other web components in the core design system.
> 
> We should figure out what's going on there. We should look at our a la carte bundling process for web components.

I think I diagnosed this issue. The individual web component bundles are still being generated from the core design system ([`coreEsm`](https://github.com/CMSgov/design-system/blob/main/webpack.config.js#L6-L8) in [`webpack.config.js`](https://github.com/CMSgov/design-system/blob/main/webpack.config.js#L55-L68)), even when building a child design system like `ds-medicare-gov`. That’s why `ds-simple-footer` is included in `all.js` but doesn’t have its own separate bundle. I wrote a follow-up ticket to address this: [WNMGDS-3221](https://jira.cms.gov/browse/WNMGDS-3221)

packages/ds-medicare-gov/src/components/web-components/ds-simple-footer/ds-simple-footer.tsx
```diff
@@ -0,0 +1,42 @@
...
+const attributes = [
+  'about-medicare-label',
+  'nondiscrimination-label',
+  'privacy-policy-label',
+  'privacy-setting-label',
+  'linking-policy-label',
+  'using-this-site-label',
+  'plain-writing-label',
+  'website-info',
+  'language',
```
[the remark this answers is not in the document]
tamara-corbalt: The original design of this was explicitly tied to the Privacy Setting modal per the Tealium Consent Preferences Manager documentation (see comment [here](https://github.com/CMSgov/design-system/blob/main/packages/ds-medicare-gov/src/components/SimpleFooter/SimpleFooter.tsx#L15-L16)). Since this is how it was in the react component, I kept it consistent. But I agree that it would be more intuitive for the modal to inherit the lang attribute from the html element instead of requiring a prop. I don’t know a whole lot about how Tealium works, but should we add a follow-up ticket to check if anything has changed in the Tealium documentation regarding this prop?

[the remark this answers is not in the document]
jack-ryan-nava-pbc: I think this is something we can wait on someone else bringing up. If the current api isn't causing medicare teams any problems, then we shouldn't change it.

packages/ds-medicare-gov/src/components/web-components/ds-simple-footer/ds-simple-footer.test.tsx
```diff
@@ -0,0 +1,41 @@
...
+
+    footer?.addEventListener('ds-click-link-analytics', mockAnalyticsCallback);
+
+    const link = screen.getByText(/About Medicare/i);
+    await userEvent.click(link);
+
+    expect(mockAnalyticsCallback).toHaveBeenCalledTimes(1);
+
+    const event = mockAnalyticsCallback.mock.calls[0][0];
+    expect(event.detail.url).toBe('https://www.medicare.gov/about-us');
```
jack-ryan-nava-pbc, line 37: Nice this is a good check!

jack-ryan-nava-pbc (approved):
Looks great! Storybook examples look good, and tests pass! Way to go on this one

PR #3424, opened 2025-02-04
[WNMGDS-3205] Adds analytics tracking to theme & version dialogs

tamara-corbalt:
## Summary

- Introduces a new analytics utility, `sendFilterAppliedEvent` and integrates it into the theme and version switcher dialogs on the docsite.

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3205)

## How to test

- Spin up the local server: `npm run start`
- Open http://localhost:8000/
- In the JavaScript console, execute `window.utag.link = console.log` to be able to intercept the analytics events and see them printed to the console.
- In your console settings, enable the preserving of logs between navigation requests. We'll be performing actions that both send analytics and go to a new page, so we don't want the page refresh to clear the console before we can see the event printed
- Try switching themes or versions.
  - For switching versions you should see something like this:
    Screenshot 2025-02-04 at 3 28 36 PM
    
  - For switching themes you should see something like this:
  Screenshot 2025-02-04 at 3 26 22 PM

## Checklist


packages/design-system/src/components/analytics/events.ts
```diff
@@ -19,7 +19,7 @@ export interface UtagContainer {
   utag?: UtagObject;
 }
 
-export type UtagEventType = 'link' | 'view';
+export type UtagEventType = 'link' | 'view' | 'filter';
```
jack-ryan-nava-pbc, line 22: So this event type corresponds to a function available on the Utag object provided to us by Blast. We use the link function on the utag object to send data over when an event happens, and we differentiate on the backend by the event type in the object we send via the link function, rather than in the function call. 

By adding 'filter' to the Event Types (confusing naming, I know) you're saying that you want to be able to reference the 'filter' function later in line 66 within the `sendAnalytics` function.

So all that to say, I don't think you want to add this type here, and I don't think you want to create a new function on line 81.

packages/design-system/src/components/analytics/events.ts
```diff
@@ -78,4 +78,8 @@ export function sendLinkEvent(event: AnalyticsEvent) {
   return sendAnalytics('link', event);
 }
 
+export function sendFilterEvent(event: AnalyticsEvent) {
+  return sendAnalytics('filter', event);
```
jack-ryan-nava-pbc, line 82: I have a big comment above, but I don't think there is a `filter` event on the utag object. Did you happen to find one? The tech spec only lists a `utag.link` function.

packages/design-system/src/components/analytics/events.ts
tamara-corbalt: Ah, I see—that makes sense! I misunderstood the code here and thought 'filter' was an event type we could add to the `utag` object, rather than a method. Doh! And then I mistakenly assumed we could extend `UtagEventType` to differentiate event types in the payload instead of just using 'link'. Thanks for the clarification—I’ll adjust accordingly!

packages/design-system/src/components/analytics/events.ts
tamara-corbalt: I'll remove this, see comment ⬆️

packages/docs/src/components/layout/SideNav/VersionDialog.tsx
```diff
@@ -15,6 +16,11 @@ export const ThemeVersionDialog = (props: ThemeVersionDialogProps) => {
 
   function handleUpdate() {
     if (version !== props.version) {
+      let filterCategoriesUsed = { theme: props.theme, version: version };
```
jack-ryan-nava-pbc, line 19: This is super close! In trying this out, there's one case that we won't capture which is if people select the latest version from the dropdown. We know that the latest version is the one shown by default, but users might not get that, and it could be helpful to see when users happen to select the latest version. I think moving these lines outside of the `if` statement to the beginning of the handle update would be a good way to handle this.

packages/docs/src/components/layout/SideNav/ThemeDialog.tsx
```diff
@@ -16,6 +17,10 @@ export const ThemeVersionDialog = (props: ThemeVersionDialogProps) => {
   function handleUpdate() {
     if (theme !== props.theme) {
       setQueryParam('theme', theme, true);
+
+      let filterCategoriesUsed = { theme: theme };
```
jack-ryan-nava-pbc, line 21: I think we want to move these added lines to outside of the if check. We want to capture when users stay on the currently selected theme. Also, is there a way to derive the version here too? Maybe by looking at the url and by using the `getVersionEquivalent` function?

packages/docs/src/components/layout/SideNav/VersionDialog.tsx
tamara-corbalt: Ah, I see what you mean. Will update!

packages/docs/src/components/layout/SideNav/ThemeDialog.tsx
tamara-corbalt: Love this idea! Instead of deriving the 'fromVersion' argument for getVersionEquivalent from the URL, I think it might make more sense to expand the props on the theme dialog component so that it receives the version prop from the parent—just like the version dialog. Let me know what you think!

kim-cmsds (approved):
Pulled this down locally and was able to follow the steps to verify the payload data using the `window.utag.link = console.log` step

jack-ryan-nava-pbc (approved):
👌🏻

PR #3428, opened 2025-02-07
[NO-TICKET] Adds error handling for missing alias case in `sync:from-figma` script

tamara-corbalt:
## Summary

- Adds error handling for Figma variables that reference a missing alias ID.
- Introduces a utility to log missing aliases for debugging purposes.

## How to test

1. Run `npm run sync:tokens-from-figma`
2. Verify that a text file, `missing_aliases.txt`, is generated with a list of variables referencing missing alias IDs.
3. Ensure that the script continues running without crashing and gracefully handles the following error currently seen on main:

    ```
    return { value: `{${aliasedVariable.name.replace(/\//g, '.')}}`, aliasedVariable };
                                              ^
    TypeError: Cannot read properties of undefined (reading 'name')
    ```

## Sample Output
```
// missing_aliases.txt

❌ Missing Alias: VariableID:894:86807 for variable theme/font/weight/heading/5xl. Mode: core
❌ Missing Alias: VariableID:894:86807 for variable theme/font/weight/heading/5xl. Mode: cmsgov
❌ Missing Alias: VariableID:894:86807 for variable theme/font/weight/heading/5xl. Mode: healthcare
❌ Missing Alias: VariableID:894:86806 for variable theme/font/weight/heading/5xl. Mode: medicare
❌ Missing Alias: VariableID:894:86807 for variable theme/font/weight/heading/md. Mode: core
❌ Missing Alias: VariableID:894:86807 for variable theme/font/weight/heading/md. Mode: cmsgov
❌ Missing Alias: VariableID:894:86807 for variable theme/font/weight/heading/md. Mode: healthcare
❌ Missing Alias: VariableID:894:86806 for variable theme/font/weight/heading/md. Mode: medicare
```
## Checklist


packages/design-system-tokens/src/figma/translateFigmaToTokens.ts
```diff
@@ -8,6 +8,13 @@ import { select } from '@inquirer/prompts';
 import { execSync } from 'child_process';
 import path from 'path';
 
+const missingAliasesLogPath = path.join(__dirname, 'missing_aliases.txt');
```
jack-ryan-nava-pbc, line 11: Will this error when someone doesn't have the missing_aliases.txt file locally?

packages/design-system-tokens/src/figma/translateFigmaToTokens.ts
```diff
@@ -221,15 +228,26 @@ function tokenValueFromVariable(
...
+      }
+
       return { value: `{${aliasedVariable.name.replace(/\//g, '.')}}`, aliasedVariable };
     } else if ('r' in value) {
       return { value: rgbToHex(value) };
     }
 
-    throw new Error(`Format of variable value is invalid: ${value}`);
+    console.warn(`⚠️ Unexpected variable format for ${variable.name}: ${JSON.stringify(value)}`);
+    return { value: `UNKNOWN_FORMAT(${JSON.stringify(value)})` };
```
jack-ryan-nava-pbc, line 250: Did you take the `UNKNOWN_FORMAT` name from somewhere? I ask because I want to make sure that this doesn't break anything downstream. To that end, do the built token files or CSS files change when you run this?

tamara-corbalt: @jack-ryan-nava-pbc This would introduce a new pattern—I added this prefix thinking of it as a fallback, but since reaching this point means we have an invalid case, it’s probably safer to leave it as-is rather than adding new handling. I’ll revert this.

packages/design-system-tokens/src/figma/translateFigmaToTokens.ts
tamara-corbalt: It shouldn’t error because we’re using `fs.appendFileSync` to log each entry, and it automatically creates the file at this path if it doesn’t exist.

jack-ryan-nava-pbc (commented):
This works great! I think one thing I would like to think about a little more is how we handle these downstream. Looking at the changes to the actual tokens, they're minimal. But when running this I am forced to select a new type for that variable, and I'm a bit wary of doing that for things with missing variable aliases. But maybe it doesn't matter because it's an alias and not the actual variable?

tamara-corbalt:
> This works great! I think one thing I would like to think about a little more is how we handle these downstream. Looking at the changes to the actual tokens, they're minimal. But when running this I am forced to select a new type for that variable, and I'm a bit wary of doing that for things with missing variable aliases. But maybe it doesn't matter because it's an alias and not the actual variable?

@jack-ryan-nava-pbc Agree completely! I don't see this as a finalized approach. When an alias is missing, we’re not just lacking a type—we’re also missing its true value, since we fail to resolve the reference in our local variables from Figma.`

That said, capturing how many variables have missing references is important for understanding the scope of the problem. For now, users can revert JSON changes while keeping `missing_aliases.txt` so that we can investigate further.

Maybe as a  next step, we could add an utility that uses Figma’s `getVariableById` to check for missing references. If that request returns `null`, should we default to manual type selection or handle it another way?

Either way, I think we need to understand which references are missing and why.

jack-ryan-nava-pbc (approved):
Smash the approve button!

kim-cmsds (approved):

PR #3458, opened 2025-02-27
[WNMGDS-3182] Updates two `@testing-library` dependencies

kim-cmsds:
## Summary

- [Jira Ticket](https://jira.cms.gov/browse/WNMGDS-3182)
- At a high-level, this PR upgrades `@testing-library/preact`, `@testing-library/react`, and `@testing-library/user-event` and removes `@testing-library/preact-hooks` and `@testing-library/react-hooks`.
- The user-event api changes from v13 to v14 means that any `userEvent.click` is now a promise that needs to be awaited and the [docs recommend invoking a single session](https://testing-library.com/docs/user-event/setup#starting-a-session-per-setup) that more closely replicates user behavior

## How to test

1. Pull down this branch locally and run `npm install`
2. Run `npm run test:unit`, `npm run test:unit:wc`, and `npm run test:unit:preact`
3. Note: there will most likely be two tests that are failing while running the preact tests (one is consistent with what's on `main` and another is a new issue with the `useClickOutsideHandler.test.tsx` test. I do not propose either should block this work but open to feedback.

## Checklist


.storybook/webpack.config.js
```diff
@@ -6,8 +6,7 @@ const usePreact = Boolean(JSON.parse(process.env.PREACT ?? 'true'));
 const preactAliases = usePreact
   ? {
       react: 'preact/compat',
-      'react-dom/test-utils': 'preact/test-utils',
```
jack-ryan-nava-pbc, line 9: Wait, why did these go away?

packages/design-system/src/components/DateField/SingleInputDateField.test.tsx
```diff
@@ -121,11 +124,11 @@ describe('SingleInputDateField', function () {
     // https://github.com/testing-library/dom-testing-library/issues/774#issuecomment-702574312
     // but it isn't working
     //
-    // it('selecting a day calls onChange', () => {
+    // it('selecting a day calls onChange', async () => {
```
jack-ryan-nava-pbc, line 127: This test doesn't work even with the new version updates?

packages/design-system/src/components/DateField/DateInput.test.tsx
```diff
@@ -24,6 +27,11 @@ function expectInvalid(textField: HTMLElement) {
 }
 
 describe('DateInput', () => {
+  afterEach(() => {
```
jack-ryan-nava-pbc, line 30: We need this clean up for anytime we use fakeTimers, right? I think there are two tests that use fake timers but don't do this clean up. Is that ok?

jack-ryan-nava-pbc (commented):
Some questions, but super close!

packages/design-system/src/components/DateField/DateInput.test.tsx
jack-ryan-nava-pbc: Seems like its: ChoiceList, ds-choice-list and IdleTimeout. Do those tests not need the cleanup for some reason?

packages/design-system/src/components/DateField/SingleInputDateField.test.tsx
kim-cmsds: @jack-ryan-nava-pbc fixed this in the latest commit; more to it than just swapping out the old for the new userEvent api

packages/design-system/src/components/DateField/DateInput.test.tsx
kim-cmsds: thanks for catching!

.storybook/webpack.config.js
kim-cmsds: early search and replace issue! Added back in

packages/design-system/src/components/DateField/SingleInputDateField.test.tsx
```diff
@@ -117,19 +117,18 @@ describe('SingleInputDateField', function () {
...
-    // but it isn't working
-    //
-    // it('selecting a day calls onChange', async () => {
-    //   const onChange = jest.fn();
-    //   const { user } = renderPicker({onChange});
-    //   await user.click(screen.getByRole('button'));
-    //   await user.click(screen.getByRole('button', {name: /9th/}))
-    //   expect(onChange).toHaveBeenCalledWith('01/09/2000', '01/09/2000');
-    // });
+    it('selecting a day calls onChange', async () => {
```
jack-ryan-nava-pbc, line 120: Yay!!!

jack-ryan-nava-pbc (approved):

PR #3471, opened 2025-03-11
[WNMGDS-3277] Added in default analytics value for empty string searches

jack-ryan-nava-pbc:
## Summary

- When someone searches for nothing, make sure we submit a default value of "no value" to our analytics handler

## How to test

2. Open the page locally
7. Search for nothing. Just click the search button, do not enter any string content
10. Confirm that search term is "no value"

## Checklist

### DIFF
--- a/packages/docs/src/components/content/SearchForm.tsx
+++ b/packages/docs/src/components/content/SearchForm.tsx
@@ -9,7 +9,9 @@ const SearchForm = ({ className }) => {
       method="GET"
       onSubmit={(event: React.FormEvent<HTMLFormElement>) => {
         const formData = new FormData(event.target as HTMLFormElement);
-        sendSearchInitiatedEvent(formData.get('query').toString());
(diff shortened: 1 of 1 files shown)


tamara-corbalt (approved):

packages/docs/src/components/content/SearchForm.tsx
```diff
@@ -9,7 +9,9 @@ const SearchForm = ({ className }) => {
       method="GET"
       onSubmit={(event: React.FormEvent<HTMLFormElement>) => {
         const formData = new FormData(event.target as HTMLFormElement);
-        sendSearchInitiatedEvent(formData.get('query').toString());
+        sendSearchInitiatedEvent(
+          formData.get('query').toString() === '' ? 'no value' : formData.get('query').toString()
```
kim-cmsds, line 13: Totally works!

NB: we could store `formData.get('query').toString()` as a variable to improve readability / intent.

PR #3478, opened 2025-03-17
[WNGDS-3283] Hide abbreviated month labels from screen readers in `MonthPicker`.

tamara-corbalt:
## Summary

- Adds functionality to hide abbreviated month labels from screen readers, ensuring that only the full month name is announced for each checkbox in the `MonthPicker` component.

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3283)

## How to test

1. Run Storybook: 
  - Inspect the `MonthPicker` while `VoiceOver` is enabled.
  - Each checkbox should announce only the full month name, not the abbreviated label.
2. Run unit tests - `npm run test:unit`

## Checklist

### DIFF
--- a/packages/design-system/src/components/Label/Label.tsx
+++ b/packages/design-system/src/components/Label/Label.tsx
@@ -61,6 +61,13 @@ export interface LabelProps {
    * Text showing the requirement ("Required", "Optional", etc.). See [Required and Optional Fields](https://design.cms.gov/patterns/Forms/forms/#required-and-optional-fields).
    */
   requirementLabel?: React.ReactNode;
+  /**
+   * Determines whether the label should be hidden from screen readers.
+   * Set to `true` to apply `aria-hidden="true"` to the label.
+   * This is useful in cases where the label does not provide valuable context for screen reader users,
+   * such as when the associated input already has an accessible name.
+   */
+  labelHidden?: boolean;
 }
 
 type LabelComponentProps = React.ComponentPropsWithRef<'label'> &
@@ -90,6 +97,8 @@ export const Label = (props: LabelComponentProps) => {
     requirementLabel,
     ...labelProps
   } = props;
+
+  const { labelHidden } = labelProps;
 
   if (process.env.NODE_ENV !== 'production' && (hint || hintId)) {
     console.warn(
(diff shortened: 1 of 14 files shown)


kim-cmsds (commented):
@tamara-corbalt nice! This is looking good. It looks like the preact snapshots need to be updated as well `npm run test:unit:preact -- -u`

kim-cmsds (approved):

packages/design-system/src/components/Label/Label.tsx
```diff
@@ -61,6 +61,13 @@ export interface LabelProps {
    * Text showing the requirement ("Required", "Optional", etc.). See [Required and Optional Fields](https://design.cms.gov/patterns/Forms/forms/#required-and-optional-fields).
    */
   requirementLabel?: React.ReactNode;
+  /**
+   * Determines whether the label should be hidden from screen readers.
+   * Set to `true` to apply `aria-hidden="true"` to the label.
+   * This is useful in cases where the label does not provide valuable context for screen reader users,
+   * such as when the associated input already has an accessible name.
```
jack-ryan-nava-pbc, line 68: Nice!

jack-ryan-nava-pbc (approved):
LGTM! What do you think about adding a story to our MonthPicker that demonstrates hiding the month labels? I'd like to be able to confirm that we don't get any a11y errors when we add this prop.

tamara-corbalt:
> LGTM! What do you think about adding a story to our MonthPicker that demonstrates hiding the month labels? I'd like to be able to confirm that we don't get any a11y errors when we add this prop.

Thanks for the feedback @jack-ryan-nava-pbc! I’m not sure about this since `labelHidden` isn’t currently exposed as a prop on `MonthPicker`, given that the labels should always hidden by default. Because of that, I’m not sure if adding a story specifically for this makes sense. But open to discussing if the team thinks we should expose `labelHidden` or take a different approach!

kim-cmsds (approved):
Took another look and verified it worked as expected. Nice job!

jack-ryan-nava-pbc:
> > LGTM! What do you think about adding a story to our MonthPicker that demonstrates hiding the month labels? I'd like to be able to confirm that we don't get any a11y errors when we add this prop.
> 
> Thanks for the feedback @jack-ryan-nava-pbc! I’m not sure about this since `labelHidden` isn’t currently exposed as a prop on `MonthPicker`, given that the labels should always hidden by default. Because of that, I’m not sure if adding a story specifically for this makes sense. But open to discussing if the team thinks we should expose `labelHidden` or take a different approach!

Yeah I think I did not realize at the time of writing that this was hardcoded in. I thought this was something that could be set by users.

jack-ryan-nava-pbc:
Tested with voiceover, works as expected! Great job

PR #3543, opened 2025-04-17
[wnmgds 3339] Fix `onChange` not firing for grouped items in `Autocomplete`

tamara-corbalt:
## Summary

Previously, when selecting a grouped item in the Autocomplete component, the `onChange` callback wasn't triggered. This was because the component only searched top-level items when resolving the selected ID and did not account for items nested within groups.

- This pull request updates the selection logic to correctly handle both standalone and grouped items.
- Adds unit tests to verify that:
  - `onChange` is triggered when selecting a grouped item (`Autocomplete`)
  - `ds-change` is dispatched when a grouped item is selected (`ds-autocomplete`)

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3339)

## How to test

- Run the unit tests: `npm run test:unit && npm run test:unit:wc`
- Open the `Autocomplete` story in Storybook. Select a grouped item and verify the `Actions` tab shows the appropriate `onChange` event.

## Checklist

### DIFF
--- a/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
@@ -12,6 +12,7 @@ import {
   getTextFieldChild,
   getActiveDescendant,
   removeUndefined,
+  findItemById,
 } from './utils';
 import { t } from '../i18n';
 import { useComboBox } from '../react-aria'; // from react-aria
@@ -232,7 +233,8 @@ export const Autocomplete = (props: AutocompleteProps) => {
       : undefined,
     onSelectionChange: onChange
       ? (selectedKey) => {
-          const selectedItem = items ? items.find((item) => selectedKey === item.id) : undefined;
+          const selectedItem =
+            items && selectedKey != null && findItemById(String(selectedKey), items);
           // We don't call onChange when the user deletes text, even though react-aria will call
           // this function with `null` if the input is cleared out. This is to maintain backwards
           // compatability, but we could consider changing this behavior in the future. If we
--- a/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
@@ -233,8 +233,7 @@ export const Autocomplete = (props: AutocompleteProps) => {
       : undefined,
     onSelectionChange: onChange
       ? (selectedKey) => {
-          const selectedItem =
-            items && selectedKey != null && findItemById(String(selectedKey), items);
+          const selectedItem = items ? findItemById(String(selectedKey), items) : undefined;
           // We don't call onChange when the user deletes text, even though react-aria will call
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -232,7 +233,8 @@ export const Autocomplete = (props: AutocompleteProps) => {
       : undefined,
     onSelectionChange: onChange
       ? (selectedKey) => {
-          const selectedItem = items ? items.find((item) => selectedKey === item.id) : undefined;
+          const selectedItem =
+            items && selectedKey != null && findItemById(String(selectedKey), items);
```
kim-cmsds, line 237:
```suggestion
            items ? findItemById(String(selectedKey), items) : undefined;
```
Do we need the `selectedKey != null` check here? Seems like we're only reaching this point if `selectedKey` is defined (L235)

packages/design-system/src/components/Autocomplete/utils.tsx
```diff
@@ -89,3 +89,28 @@ export function removeUndefined<T>(obj: T): T {
...
+    Array.isArray((item as AutocompleteItemGroup).items)
+  );
+};
+
+export const findItemById = (
+  id: string,
+  items: AutocompleteItems
+): AutocompleteItem | undefined => {
+  for (const item of items) {
+    if (isGroup(item)) {
```
kim-cmsds, line 107: I think it's possible to simplify this logic and remove the `isGroup` function if we swap this out to:
```suggestion
    if ('items' in item) {
```
since `items` only exists on the AutocompleteItemGroup type

packages/design-system/src/components/Autocomplete/utils.tsx
```diff
@@ -89,3 +89,28 @@ export function removeUndefined<T>(obj: T): T {
...
+      const match = item.items.find((child) => child.id === id);
+      if (match) return match;
+    } else {
+      if (item.id === id) return item;
+    }
+  }
+
+  return undefined;
```
kim-cmsds, line 115: If we change how this function is used by following this suggestion (https://github.com/CMSgov/design-system/pull/3543/files#r2049473499), then I don't think we need this return statement anymore. What do you think?

packages/design-system/src/components/web-components/ds-autocomplete/ds-autocomplete.test.tsx
```diff
@@ -364,6 +364,52 @@ describe('Autocomplete', () => {
     expect(autocompleteField.value).toBe('');
   });
 
+  it('should dispatch ds-change when a grouped item is selected', () => {
```
kim-cmsds, line 367: Nice! Thanks for adding this much test coverage!

kim-cmsds (approved):
This looks good to me. Test coverage is great and appreciated!

PR #3546, opened 2025-04-22
[WNMGDS-3302] Refactor custom Playwright reporter to isolate test results by config group

tamara-corbalt:
## Summary

- Previously, our custom Playwright reporter wrote all test results to a single JSON file, which caused results to be overwritten with each test run.
- This PR refactors the reporter to write results to separate subdirectories based on the test config name. This allows us to preserve results from each test group and review them in isolation.

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3302)

> **Note:** This change will impact the `test-nightly` Jenkins script. A corresponding pull request [(pr-94)](https://github.cms.gov/CMS-WDS/experience-toolkit/pull/94]) has been opened to support this new per-config test-result behavior.

## How to test

- Run `npm run test:browser:all`
- Verify that output is written to the following structure:

```
tests/
└── browser/
    └── test-results/
        ├── storybook-docs/
        │   └── test-report.json
        ├── examples/
        │   └── test-report.json
        ├── interaction/
        │   └── test-report.json
        └── <etc...>
```

## Checklist


tests/browser/custom-reporter.ts
```diff
@@ -9,6 +9,14 @@ import type {
 import { writeFileSync, mkdirSync } from 'fs';
 import path from 'path';
 
+function extractTestGroupFromConfigPath(configPath: string | undefined): string {
+  if (configPath === undefined) {
```
jack-ryan-nava-pbc, line 13: Is the config only going to be undefined during testing? If so, then maybe 'unknown' isn't the most helpful path name to include here?

tests/browser/custom-reporter.ts
```diff
@@ -18,6 +26,7 @@ class MyReporter implements Reporter {
   private isListMode = false;
   private totalTests = 0;
   private currentTestIndex = 0;
+  private testGroup: string | null = null;
```
jack-ryan-nava-pbc, line 29: This is a nit, but I find it confusing to have this be intitialized to null rather than "".  extractTestGroupFromConfigPath should always return a string, so the interpolation on line 115 won't ever include the word "null", but I'm just wondering why this needs to be null in the first place? Is it for debugging purposes?

jack-ryan-nava-pbc (commented):
Works great locally! Some small comments but overall looking good! Excited for the jenkins tests.

tests/browser/custom-reporter.ts
tamara-corbalt: Yep, it will be `undefined` when we run the [custom-reporter tests](https://github.com/CMSgov/design-system/blob/main/tests/browser/custom-reporter.test.ts)—can’t think of other scenarios where that’d be the case. Would something like `'testing'` be a clearer fallback than `'unknown'`?

tests/browser/custom-reporter.ts
tamara-corbalt: I think you're right — I originally set it to `null` to signal that it hadn’t been initialized with a legit value yet, but I can see how that’s not super clear here. Since we’re guaranteed to get a string from `extractTestGroupFromConfigPath`, initializing it as '' is cleaner. I’ll update it

kim-cmsds (approved):
Tested locally and things work as expected ✅

jack-ryan-nava-pbc (approved):
Thanks for making those changes!

PR #3552, opened 2025-04-29
[WNMGDS-3304] Removes default props from components

kim-cmsds:
## Summary

- This PR removes default props from components that are not overridden by a child design system (e.g. ds-healthcare-gov)
- [Jira ticket](https://jira.cms.gov/browse/WNMGDS-3304)

## How to test

1. Pull down the branch and run through the test suite:
2. `npm install`
3. `npm run test`
4. `npm run test:browser:all`

## Checklist

### DIFF
--- a/packages/design-system/src/components/Dropdown/Dropdown.test.tsx
+++ b/packages/design-system/src/components/Dropdown/Dropdown.test.tsx
@@ -128,6 +128,7 @@ describe('Dropdown', () => {
   });
 
   it('shows error message', async () => {
+    jest.useFakeTimers();
     const errorId = 'my-error';
     const { container, user } = makeDropdown(
       {
@@ -152,6 +153,7 @@ describe('Dropdown', () => {
   });
 
   it('supports bottom placed error', async () => {
+    jest.useFakeTimers();
     const errorId = 'my-error';
     const { container, user } = makeDropdown(
       {
--- a/packages/design-system/src/components/Tooltip/Tooltip.tsx
+++ b/packages/design-system/src/components/Tooltip/Tooltip.tsx
@@ -33,6 +33,7 @@ export interface BaseTooltipProps {
   closeButtonLabel?: string;
   /**
    * When provided, will render the passed in component for the tooltip trigger. Typically will be a `button`, `a`, or rarely an `input` element.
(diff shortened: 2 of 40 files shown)


packages/design-system/src/components/Dropdown/Dropdown.test.tsx
```diff
@@ -128,6 +128,7 @@ describe('Dropdown', () => {
   });
 
   it('shows error message', async () => {
+    jest.useFakeTimers();
```
jack-ryan-nava-pbc, line 131: Nice cleanup

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -246,7 +255,7 @@ export const Tooltip = (props: TooltipProps) => {
       ariaLabel,
       children,
       className,
-      component,
+      component = 'button',
```
jack-ryan-nava-pbc, line 258: There's a few props we're adding default values to in the renderContent function, like interactiveBorder and maxWidth that we aren't here. Is that intentional?

jack-ryan-nava-pbc (approved):
Approving! But I do have a question in the review that I'd like addressed prior to merging.

packages/design-system/src/components/Tooltip/Tooltip.tsx
kim-cmsds: Yes, that is. Those props aren't used in the `renderTooltip` function are being escaped. They are used in the `renderContent` function, which is where the default values are defined.

packages/design-system/src/components/Tooltip/Tooltip.tsx
```diff
@@ -141,7 +148,9 @@ export const placements: Placement[] = [
  * provide an `aria-label` on the `<Tooltip>` to ensure an accessible name for
  * the trigger.
  */
+
 export const Tooltip = (props: TooltipProps) => {
+  const { placement = 'top', offset = [0, 5] } = props;
```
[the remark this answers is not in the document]
kim-cmsds: Thanks, @jack-ryan-nava-pbc!

tamara-corbalt (approved):

PR #3559, opened 2025-05-01
[WNMGDS-3344] Add Utag data object

jack-ryan-nava-pbc:
## Summary

- Add new Tealium snippet to body tag via SSR script - this will fire page view events on every page now
- Update Tealium scripts in the Head

## How to test

2. Open the doc site locally
10. Confirm that the object contains a field that says "view", along with a "page_name" equal to the name of the page plus the theme name. 

## Checklist


jack-ryan-nava-pbc:
@kim-cmsds @tamara-corbalt This change will necessitate some changes to the Gatsby-Head component in the React 18 branch, but lets hold off doing those changes until Blast confirms these changes.

packages/docs/gatsby-ssr.tsx
```diff
@@ -0,0 +1,24 @@
...
+  const env = proc.env.NODE_ENV == 'production' ? 'prod' : 'dev';
+
+  const tealiumSnippet = `
+    (function(a,b,c,d){
+        a='https://tealium-tags.cms.gov/cms-design/${env}/utag.js';
+        b=document;c='script';d=b.createElement(c);d.src=a;d.type='text/javascript';d.async=true;
+        a=b.getElementsByTagName(c)[0];a.parentNode.insertBefore(d,a);
+    })();`;
+
+  const dangerousTealiumSnippet = makeDangerousHTML(tealiumSnippet);
```
tamara-corbalt, line 21: Non-blocking, but I’m wondering if we need the dedicated `makeDangerousHTML` function here. Since it’s only used once, would it be worth inlining it?
`dangerousTealiumSnippet = {  __html: tealiumSnippet  }`

jack-ryan-nava-pbc: yeah that's a good call. this is more an homage to all the work I did in a previous effort, only to delete. but I shouldn't be so sentimental. thank you!

jack-ryan-nava-pbc:
@tamara-corbalt did you get the video of the testing steps I shared via Google Docs?

tamara-corbalt:
> @tamara-corbalt did you get the video of the testing steps I shared via Google Docs?

Yes, and I was able to walk through the testing steps. Thanks for sharing the screencast! 

Everything looks like it's working as expected, so hitting approve!

tamara-corbalt (approved):

packages/docs/gatsby-ssr.tsx
```diff
@@ -0,0 +1,20 @@
+const proc = require('process');
+
+const HeadComponentsJs = (content) => (
+  <script type="text/javascript" dangerouslySetInnerHTML={content}></script>
+);
+
+exports.onRenderBody = ({ setPreBodyComponents }) => {
```
kim-cmsds, line 7: TIL (/ re-remembered) about this gatsby feature

packages/docs/src/components/layout/Layout.tsx
```diff
@@ -66,12 +69,45 @@ const Layout = ({
   theme,
   tableOfContentsData,
 }: LayoutProps) => {
-  const env = 'prod';
+  // Using a ref here because this value shouldn't change after the initial render.
```
kim-cmsds, line 72: thanks for the comment here

packages/docs/src/components/layout/Layout.tsx
```diff
@@ -66,12 +69,45 @@ const Layout = ({
...
+      }
+
+      const analyticsPayload = {
+        content_language: 'en',
+        content_type: 'html',
+        logged_in: 'false',
+        page_name: tabTitle,
+        page_type: tabTitle.includes('Page not found') ? 'true' : 'false', //If page is a 404 (error page) this is set to true, otherwise it is false
+        site_environment: env.current, //Used to include or exclude traffic from different testing environments. Ex: test, test0, imp, production
+        site_section: location.pathname == '/' ? 'index' : location.pathname, // Set the section to the pathname, except in the case of the index.
```
kim-cmsds, line 104: NB: I don't think this matters but why use two equals signs and not three for the "index" check?

kim-cmsds (approved):
Was able to verify the testing steps! And appreciate the context provided in the PR description and JIRA Ticket.

packages/docs/src/components/layout/Layout.tsx
jack-ryan-nava-pbc: So this is me trying out a sorta new thing for me: I've adopted the opinion that if you use `==` it means you know what the types are in the comparison you are making. The only difference between `===` and `==` is that `===` checks to see if the two items being compared are the same type at first. 
So using `==` here states that we know that `location.pathname` will only ever return a string so we don't need the type comparison to another string value.

PR #3563, opened 2025-05-05
[WNMGDS-3349] Add `status` to frontmatter for `.mdx` files.

tamara-corbalt:
## Summary

- Introduces a `status` field in component `.mdx` frontmatter:
```
---
title: Table
// ... other fields ...
status:
  level: use | caution | avoid
  note: // Optional context for "caution" and "avoid" cases. 
---
```
- Adds `status` to the frontmatter of each component `.mdx` file.
- Adds a type validation  for `status.level` using gatsby schema customization.

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3349)

## How to test
- Run `npm run start`. The site should build without errors.
- Open `http://localhost:8000/___graphql` and run the following query.
```
query mdxStatusQuery {
  allMdx(
    filter: { frontmatter: { title: { ne: null }, status: { level: { ne: null } } } }
    sort: { fields: frontmatter___title, order: ASC }
  ) {
    nodes {
      frontmatter {
        title
        status {
          level
          note
        }
      }
    }
  }
}
```
Verify that:
- Every component has a `status.level` (use, caution, or avoid)
  _Note:  When `status.level` is "caution" or "avoid", the `note` field is populated. For the `use` case, the `note` field is `null`._
To verify the `status.level` type validation:
- Open any component `.mdx` file.
- Change the `status.level` value to something invalid (e.g., "banana").
 - Run `npm run start` again — the build should fail with a graphQL schema error.
```
ERROR #85901  GRAPHQL

There was an error in your GraphQL query:

Enum "ComponentStatusLevel" cannot represent value: "banana"
```
## Checklist

### DIFF
--- a/packages/docs/content/components/accordion.mdx
+++ b/packages/docs/content/components/accordion.mdx
@@ -1,6 +1,8 @@
 ---
 title: Accordion
 intro: An accordion is a list of headers that hide or reveal additional content when selected.
+status:
+  level: use
 cmsgov:
   figmaNodeId: 6-79
 core:
--- a/packages/docs/content/components/alert.mdx
+++ b/packages/docs/content/components/alert.mdx
@@ -1,6 +1,8 @@
 ---
 title: Alert
 intro: Alerts give feedback to users, provide critical information, and/or offer key supplementary details about a task.
+status:
+  level: use
 cmsgov:
   figmaNodeId: 6-80
 core:
--- a/packages/docs/content/components/autocomplete.mdx
+++ b/packages/docs/content/components/autocomplete.mdx
@@ -1,6 +1,8 @@
 ---
 title: Autocomplete
 intro: Autocompletes allow users to enter any combination of letters, numbers, or symbols of their choosing (unless otherwise restricted), and receive one or more suggested matches in a list below the input.
+status: 
+  level: use
 cmsgov:
   figmaNodeId: 6-81
 core:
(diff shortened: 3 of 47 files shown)


jack-ryan-nava-pbc (approved):
This looks great! I learned so much looking into the APIs you call and the functions you implemented. 

I'm wondering, is the `note` value required when the status is not `use`? Has that been discussed? If it is required when the status is not `use`, I wonder if we can include some sort of validation in our schema definition for that? Just a thought, definitely non-blocking.

jack-ryan-nava-pbc (approved):
This looks great! I learned so much looking into the APIs you call and the functions you implemented. 

I'm wondering, is the `note` value required when the status is not `use`? Has that been discussed? If it is required when the status is not `use`, I wonder if we can include some sort of validation in our schema definition for that? Just a thought, definitely non-blocking.

tamara-corbalt:
> I'm wondering, is the `note` value required when the status is not `use`? Has that been discussed? If it is required when the status is not `use`, I wonder if we can include some sort of validation in our schema definition for that? Just a thought, definitely non-blocking.

Thanks for the feedback! No, we haven’t discussed validating these status fields yet. But as I was thinking ahead, I figured it would be helpful to validate `status.level` since the forthcoming `StatusIndicator` component will rely on it to render the correct `Badge` variant. Having that validation in place should keep the component logic simple and avoid a bunch of defensive checks.

I didn’t think it was necessary to validate `status.note`, since we can just conditionally render an `Alert` if it’s present. That said, I really like the idea of requiring context for "caution" or "avoid" cases.

jack-ryan-nava-pbc:
@tamara-corbalt thanks for looking into it! That's kinda what I suspected, but I appreciate you doing some digging. This is great as is!

kim-cmsds (approved):

PR #3566, opened 2025-05-06
[WNMGDS-3350] Remove form components table

jack-ryan-nava-pbc:
## Summary

- [Ticket](https://jira.cms.gov/browse/WNMGDS-3350)
- We attempt to render an additional styling table for form components, however there is no component named "form" in the Tokens lists (or in the design system, it is a pattern) so we are unable to render any additional form styling tables. 
- This ticket removes all mentions of `<ComponentThemeOptions componentname="form" />` in our MDX files, and the associated text providing context for those tables.
- I also took the liberty to add a little message for when no additional styling variables exist for a given component.

## How to test

1. Search for `<ComponentThemeOptions componentname="form" />` in the codebase. There should be 0 results.
2. Fire up the doc site locally, and add `<ComponentThemeOptions componentname="form" />` to a random MDX page. You should see the message `No variables available for component: "form".` Happy to discuss whether this is needed or not.

## Checklist

### DIFF
--- a/packages/docs/src/components/content/ComponentThemeOptions.tsx
+++ b/packages/docs/src/components/content/ComponentThemeOptions.tsx
@@ -72,7 +72,7 @@ const ComponentThemeOptions = ({ theme, componentname }: ComponentThemeOptionsPr
     </Table>
   );
 
-  if ((!componentVariables || componentVariables.length == 0) && !isProduction()) {
+  if (componentVariables.length == 0 && !isProduction()) {
     console.error(
       'You are trying to render componentVariables inside of a call to ComponentThemeOptions. The componentVariables array is either undefined or empty. This will prevent the table from rendering.'
     );
--- a/packages/docs/src/components/content/ComponentThemeOptions.tsx
+++ b/packages/docs/src/components/content/ComponentThemeOptions.tsx
@@ -75,7 +75,7 @@ const ComponentThemeOptions = ({ theme, componentname }: ComponentThemeOptionsPr
     <section className="c-configuration-options ds-u-padding-bottom--3">
       {componentVariables.length > 0
         ? componentOptions
-        : `No variables available for ${componentname}.`}
+        : `No variables available for component: "${componentname}".`}
     </section>
   );
 };
--- a/packages/docs/src/components/content/ComponentThemeOptions.tsx
+++ b/packages/docs/src/components/content/ComponentThemeOptions.tsx
@@ -10,6 +10,7 @@ import {
 } from '@cmsgov/design-system';
 import ColorSwatch from './ColorSwatch';
 import { getComponentVariables, ThemeName } from '../../helpers/themeTokens';
+import { isProduction } from '../../helpers/urlUtils';
 
 interface ComponentThemeOptionsProps {
   /**
@@ -71,8 +72,8 @@ const ComponentThemeOptions = ({ theme, componentname }: ComponentThemeOptionsPr
     </Table>
   );
 
-  if (!componentVariables || componentVariables.length == 0) {
-    console.warn(
+  if ((!componentVariables || componentVariables.length == 0) && !isProduction()) {
+    console.error(
       'You are trying to render componentVariables inside of a call to ComponentThemeOptions. The componentVariables array is either undefined or empty. This will prevent the table from rendering.'
     );
   }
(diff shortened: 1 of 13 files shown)


packages/docs/src/components/content/ComponentThemeOptions.tsx
```diff
@@ -72,7 +72,11 @@ const ComponentThemeOptions = ({ theme, componentname }: ComponentThemeOptionsPr
   );
 
   return (
-    <section className="c-configuration-options ds-u-padding-bottom--3">{componentOptions}</section>
+    <section className="c-configuration-options ds-u-padding-bottom--3">
+      {componentVariables.length > 0
+        ? componentOptions
+        : `No variables available for component: "${componentname}".`}
+    </section>
```
tamara-corbalt, line 79: This logic makes sense to me, although I'm not sure how much value the fallback message adds for the end user. Would it make more sense to return `null` early if `componentVariables.length === 0` to avoid rendering anything at all?

tamara-corbalt (approved):
Looks great! Just one non-blocking question

packages/docs/src/components/content/ComponentThemeOptions.tsx
tamara-corbalt: I hear you—that makes sense. I don’t have strong feelings about this, so feel free to disregard, but I wonder if there’s a way to split the difference? Maybe return early if it's empty, but issue a `console.warn` in development?

jack-ryan-nava-pbc: Ok, yeah that is the pattern we use for downstream packages. I guess it doesn't make sense to do it a different way for ourselves, but I'm just trying to make sure we don't miss anything, and burying it in the console feels like a way to miss it... I'll go with your suggestion for consistency and maybe I'm over thinking it. Thanks @tamara-corbalt !

packages/docs/src/components/content/ComponentThemeOptions.tsx
```diff
@@ -71,8 +71,18 @@ const ComponentThemeOptions = ({ theme, componentname }: ComponentThemeOptionsPr
     </Table>
   );
 
+  if (!componentVariables || componentVariables.length == 0) {
```
kim-cmsds, line 74: @jack-ryan-nava-pbc can you please add a check for whether the environment is production? I don't think we should be adding a console warning in that case.

packages/docs/src/components/content/ComponentThemeOptions.tsx
```diff
@@ -71,8 +72,8 @@ const ComponentThemeOptions = ({ theme, componentname }: ComponentThemeOptionsPr
     </Table>
   );
 
-  if (!componentVariables || componentVariables.length == 0) {
-    console.warn(
+  if ((!componentVariables || componentVariables.length == 0) && !isProduction()) {
+    console.error(
```
tamara-corbalt, line 76: Looks good! One small suggestion (non-blocking): since the type safety on `getComponentVariables()` guarantees an array, I think we can drop the `!componentVariables` check.

jack-ryan-nava-pbc: Good call, done

kim-cmsds: @jack-ryan-nava-pbc thanks!

The `urlUtils` helper function you wrote totally works or you could do `process.env.NODE_ENV !== 'production'` here.

jack-ryan-nava-pbc: @kim-cmsds sorry I'm slightly confused, I think this code only runs client side, so there is no "process" to refer to?

kim-cmsds: @jack-ryan-nava-pbc as part of the [build process](https://v4.gatsbyjs.com/docs/conceptual/overview-of-the-gatsby-build-process/#steps-of-the-build-phase), Gatsby is using node to build compile JS and CSS and build out the static HTML pages. As a result, it's possible to use that `process.env` variable to detect whether it's production or not. You can try it out by adding a `console.log(process.env.NODE_ENV`) to the `ComponentThemeOptions.tsx` file on L33. Run `npm run start` and see that it's `development` and then `npm run build:docs && npm run serve:docs` and it should return `production`.

PR #3580, opened 2025-05-12
[WNMGDS-3309] Make stylesheet polling stop after 30 seconds

jack-ryan-nava-pbc:
## Summary

- [Ticket](https://jira.cms.gov/browse/WNMGDS-3309)
- Our stylesheet polling script was originally set to just kinda, run. If users encountered an error with it, the error would happen every 10 ms or so, forever.
- This is a little bit of a combined approach. I set the polling to stop after 30 seconds, which I think is plenty of time to capture CSS load time. And I introduce some pretty minor error handling. 

### Rationale for 30-second cutoff.
- I did a teeny bit of research looking at the the loads times for CSS assets for medicare.gov, cms.gov, design.cms.gov and healthcare.gov
- Each site was throttled to run at 3G
- I took the average time of the CSS stylesheet that took the longest to load, and came up with the following table:

| medicare.gov  | cms.gov |  healthcare.gov
| ------------- | ------------- | ------------- |
| 17 s  | 10 s  | No CSS resources noted in the Network tab  |

The max value was 17 seconds, so 30 seconds is almost 2x the worst load time.

### Error Handling
Based on a reported error from one of our consuming teams, I added a try catch statement to the copyGlobalStyleSheets function in case either a. there are no associated cssRules with a document sheet or b. the user is using an older browser that does not support replaceSync.

## How to test

1. In the Angular example, pull in two external stylesheets via the `examples/angular/src/index.html` file (place in the <head> element). Maybe do Bootstrap? `<link href="https://cdn.jsdelivr.net/npm/bootstrap@5.0.2/dist/css/bootstrap.min.css" rel="stylesheet" integrity="sha384-EVSTQN3/azprG1Anm3QDgpJLIm9Nao0Yz1ztcQTwFspd3yD65VohhpuuCOmLASjC" crossorigin="anonymous">`
2. Remove the `crossorigin="anonymous"` attribute and the integrity attribute and sha: integrity="sha384-EVSTQN3/azprG1Anm3QDgpJLIm9Nao0Yz1ztcQTwFspd3yD65VohhpuuCOmLASjC"
3. Observe a console warning. 

We cannot write a spec yet for this functionality since [JSDom does not yet support the `stylesheet.replace()` or `stylesheet.replaceSync()` functions](https://github.com/jsdom/jsdom/issues/3766)

## Checklist


packages/design-system/src/components/web-components/preactement/globalStyles.ts
```diff
@@ -15,16 +19,23 @@ export function getGlobalStyleSheets() {
...
-      .join(' ');
-    sheet.replaceSync(css);
-    return sheet;
+    try {
+      const css = Array.from(documentSheet.cssRules)
+        .map((rule) => rule.cssText)
+        .join(' ');
+      sheet.replaceSync(css);
+    } catch (error) {
+      console.warn(
```
jack-ryan-nava-pbc, line 28: We can talk more about whether we want to check for whether folks are on production or not. Since we don't know where these components will be used, but they will always be in the browser, perhaps a check for an origin of localhost is reasonable. Happy to think more!

tamara-corbalt (commented):
Looks good! The changes make sense to me. Thanks for outlining the rationale behind the 30-second polling cutoff—that context is helpful.

One quick question: when you were testing the various sites, did you happen to try it in different browsers? Not sure if that would impact stylesheet load behavior or timing

jack-ryan-nava-pbc:
> Looks good! The changes make sense to me. Thanks for outlining the rationale behind the 30-second polling cutoff—that context is helpful.
> 
> One quick question: when you were testing the various sites, did you happen to try it in different browsers? Not sure if that would impact stylesheet load behavior or timing

Hey @tamara-corbalt good question! No I only tried in Chrome, as I am most familiar with the dev tools in that browser. It could impact loading times, as each browser has different ways of prioritizing resources. However, I did some more performance profiling last night. Still using Chrome I checked on Healthcare.gov. I throttled both the network speed to 3g, and then since a lot of the assets appear to be generated via JS on that site, I slowed down the CPU 6x. The whole page still managed to render in under 30 seconds. That includes time to execute JS and everything. Not sure if that helps a whole lot, but at least resource loading was likely not an issue here.

tamara-corbalt:
> Hey @tamara-corbalt good question! No I only tried in Chrome, as I am most familiar with the dev tools in that browser. It could impact loading times, as each browser has different ways of prioritizing resources. However, I did some more performance profiling last night. Still using Chrome I checked on Healthcare.gov. I throttled both the network speed to 3g, and then since a lot of the assets appear to be generated via JS on that site, I slowed down the CPU 6x. The whole page still managed to render in under 30 seconds. That includes time to execute JS and everything. Not sure if that helps a whole lot, but at least resource loading was likely not an issue here.

Awesome, thanks! Sounds like you ran a pretty conservative simulation—that’s probably more than sufficient for this. I’ll go ahead and approve ✅

tamara-corbalt (approved):

kim-cmsds (approved):
I was able to verify that this works locally! Thank you for the walkthrough and steps to test.

PR #3587, opened 2025-05-14
[NO-TICKET] Fix bad check for window

jack-ryan-nava-pbc:
## Summary

- We're checking for window on the server and it's breaking our deployment. Could confirm that we're on production via if (typeof window === 'undefined') but I'm just gonna inline this and save the lines of code.

## How to test

1. Nightly tests should deploy now

## Checklist


tamara-corbalt (approved):
Looks good!

PR #3590, opened 2025-05-14
[NO-TICKET] Fix ds-modal-dialog stories where the modals don't open

jack-ryan-nava-pbc:
## Summary

- Modal dialogs don't open on the main documentation page for the modal dialog! Ack!
- This is due to how we're initially querying the DOM in our useEffect, we're only grabbing the first modal, button and form when really we ought to be grabbing all of them and then tying the opening and closing of the modal to the supplied id. There's a bit of checking for whether we're using a Form element or the dialog to close things, but that's a pretty minor check.

## How to test

1. Pull this down locally
2. Build & run storybook
3. Go to the ds-modal-dialog story and open and close all of the modals. Only one modal should open at a time and all should be closable.
4. Confirm that the modals open and close as expected in the specific story pages as well.

## Checklist

### DIFF
--- a/packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
+++ b/packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
@@ -27,9 +27,9 @@ const meta: Meta = {
             <ds-button value="rstBtn" class-name="ds-u-margin-right--1">
               Reset
             </ds-button>
-            <ds-button type="submit" value="sbmBtn">
+            <button className="ds-c-button" type="submit" value="sbmBtn">
               Confirm
-            </ds-button>
+            </button>
           </form>
         </span>
       </>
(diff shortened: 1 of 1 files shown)


packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
```diff
@@ -120,32 +120,54 @@ export default meta;
...
+      const modal = document.getElementById(event.target.dataset.id);
+      modal?.setAttribute('is-open', 'true');
+    };
 
-    const handleExit = (event) => {
+    const closeModal = (event) => {
       action('ds-exit')(event);
-      closeModal();
+      let modal: Element;
+      if (event.target.tagName == 'FORM') {
```
tamara-corbalt, line 135: I know you mentioned ([here](https://github.com/CMSgov/design-system/pull/3559#discussion_r2089939741)) using `==` as a signal that the types are known — and just noting I haven’t dug fully into your follow-up explanation [here](https://github.com/CMSgov/design-system/pull/3559#discussion_r2091290939) yet — but in cases like this, especially when working with DOM APIs, do you think there’s a chance that looser comparisons could introduce subtle issues?

tamara-corbalt (commented):
Works great! Just one question — kind of a follow-up to [this discussion](https://github.com/CMSgov/design-system/pull/3559#discussion_r2089777311) about the use of == syntax.

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
tamara-corbalt: 🤓

tamara-corbalt: Love the nerd deep dive 😄 and appreciate the explanation! The usage of `==` here makes sense, since `tagName` should reliably be a string. I think from the jump I was mostly thinking about readability — I wasn’t super familiar with the nuances between loose and strict comparison. Now that I understand that subtlety better, my only lingering (and non-blocking) concern is around intention: when future-me reads `==`, I might automatically assume coercion is necessary, which could make me pause.

Also fully owning that I may have just crossed into overthinking territory 😅

That said, given the type guarantees in this context, using `==` here seems totally reasonable.

tamara-corbalt: > I should add some safe navigation operators to the event.target. 

Also, agree with this ^^

jack-ryan-nava-pbc: Yeah that's a fair read! I think I'm misapplying it here. Good news is I'm refactoring and might not even need that check!

packages/design-system/src/components/web-components/ds-modal-dialog/ds-modal-dialog.stories.tsx
```diff
@@ -130,7 +130,7 @@ const Template = (args) => {
 
     const closeModal = (event) => {
       action('ds-exit')(event);
-      const modal = event.target.closest('ds-modal-dialog');
+      const modal = event?.target.closest('ds-modal-dialog');
```
tamara-corbalt, line 133: Nice

tamara-corbalt (approved):
Refactor looks good to me! Probably need to update a snapshot, but approving in the meantime

PR #3591, opened 2025-05-15
[NO-TICKET] Remove useRef and useEffect from Layout.tsx to correct analytics functioning

jack-ryan-nava-pbc:
## Summary

The env variable is not being set properly on page load which is preventing our analytics script from working properly.

What I wanted to happen:
useEffect runs once on page load and updates the ref to the current environment, which then makes it’s way into the <head> element.

What is actually happening:
useEffect runs after page load and sets the ref value, but the page has already been rendered with null env variable from the initialized ref. Since it’s a ref, no rerender is scheduled and the updated value never makes it to the DOM.

I should've known this is what would happen. Anyway this change should fix it and will be released to the Doc site upon approval.

## How to test

1. Same as in https://github.com/CMSgov/design-system/pull/3559
2. Confirm that there is a line in the <head> element that reads: <script data-react-helmet="true">window.tealiumEnvironment = "dev";</script> when testing locally.

## Checklist

### DIFF
--- a/packages/docs/src/components/layout/Layout.tsx
+++ b/packages/docs/src/components/layout/Layout.tsx
@@ -1,5 +1,4 @@
 import type * as React from 'react';
-import { useEffect, useRef } from 'react';
 import Footer from './DocSiteFooter';
 import SideNav from './SideNav/SideNav';
 import PageHeader from './PageHeader';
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/layout/Layout.tsx
```diff
@@ -69,44 +68,41 @@ const Layout = ({
...
+        env = 'dev';
+        break;
+      case 'cmsgov.github.io':
+        env = 'github-demo';
+        break;
+      case 'design.cms.gov':
+        env = 'prod';
+        break;
+      default:
+        env = 'prod';
```
kim-cmsds, line 91: Do we want to default to `prod` here?

kim-cmsds (approved):

tamara-corbalt (commented):
The code looks good, but when I ran `npm run serve` (to simulate prod), I noticed the `tealiumEnvironment` was set to "undefined" in the script tag. That wasn’t what I expected — is there something I might be overlooking?

PR #3593, opened 2025-05-16
[NO-TICKET] Add in UTAG page view analytics

jack-ryan-nava-pbc:
## Summary

So this has been a little bit of a saga.

In #3559 I attempted to add some new analytics functionality for page views. This didn't really work because of my use of a ref to store the env variable.
In #3591 I tried to simplify things and get all the work done outside of a useEffect, just in the script. But since I didn't understand the fineries of how Gatsby does server rendering (and some of the distinctions between Gatsby's development and production build processes) we end up with an undefined variable on production (thank you @tamara-corbalt for catching).
Rolled #3591 back in #3592 and so now here we are going back to the original client side implementation with some more sophistication (thank you @kim-cmsds). 

Anyway, here we are.

## How to test

2. Open the doc site locally
10. Confirm that the object contains a field that says "view", along with a "page_name" equal to the name of the page plus the theme name. 
11. Confirm that the environment is correct in both development (should be 'dev') and when running `gatsby server` (should be 'prod')

## Checklist

### DIFF
--- a/packages/docs/src/components/layout/Layout.tsx
+++ b/packages/docs/src/components/layout/Layout.tsx
@@ -1,4 +1,5 @@
 import type * as React from 'react';
+import { useEffect, useState } from 'react';
 import Footer from './DocSiteFooter';
 import SideNav from './SideNav/SideNav';
 import PageHeader from './PageHeader';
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/layout/Layout.tsx
```diff
@@ -69,44 +69,51 @@ const Layout = ({
...
         page_type: tabTitle.includes('Page not found') ? 'true' : 'false', //If page is a 404 (error page) this is set to true, otherwise it is false
-        site_environment: env.current, //Used to include or exclude traffic from different testing environments. Ex: test, test0, imp, production
+        site_environment: env, //Used to include or exclude traffic from different testing environments. Ex: test, test0, imp, production
         site_section: location.pathname == '/' ? 'index' : location.pathname, // Set the section to the pathname, except in the case of the index.
       } as any;
 
       sendViewEvent(analyticsPayload);
     }
-  }, []);
+  }, [env, location.pathname, location.hostname, tabTitle]);
```
kim-cmsds, line 92: Looks like we're not using the `location.hostname` variable in the useEffect so we can remove that from the dep array

kim-cmsds (approved):
Testing this using `npm run build:docs && npm run serve:docs` and `npm run start` and was able to verify that the view event is sent once and 
`<script data-react-helmet="true">window.tealiumEnvironment = "dev";</script>` is present

packages/docs/src/components/layout/Layout.tsx
```diff
@@ -69,44 +69,51 @@ const Layout = ({
...
-          env.current = 'github-demo';
-          break;
-        case 'design.cms.gov':
-          env.current = 'prod';
-          break;
-        default:
-          env.current = 'prod';
-      }
-
+    if (env !== undefined) {
```
kim-cmsds, line 79: nit: we could simplify this to `if (env) {`

packages/docs/src/components/layout/Layout.tsx
```diff
@@ -69,44 +69,51 @@ const Layout = ({
...
+      case 'cmsgov.github.io':
+        setEnv('github-demo');
+        break;
+      case 'design.cms.gov':
+        setEnv('prod');
+        break;
+      default:
+        setEnv('prod');
+    }
+  }, [env, location.hostname]);
```
kim-cmsds, line 110: we're also not referencing `env` in this effect hook, so we could also remove that from the dep array

tamara-corbalt (approved):

PR #3595, opened 2025-05-19
[WNMGDS-3360] Add `StatusIndicator` component (full implementation)

tamara-corbalt:
## Summary
- Adds a new `StatusIndicator` component to render a status badge for each component page.
- Updates the `PageHeader` component to conditionally render:
  - A `StatusIndicator` badge (for all status levels)
  - An `Alert` for caution and avoid cases using `status.note`
  - An Alert for theme-specific guidance using new frontmatter fields: `targetTheme` and `targetThemeNote`.
 - Consolidates all status and theme-based alerts within the `<header>`.

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3360)

## How to test
- Run the doc site locally or check out the [deploy preview](https://cmsgov.github.io/design-system/branch/tamara/WNMGDS-3360/add-status-indicator-component).
- Visit various component pages and confirm that the page header includes a status badge that reflects the component’s `status.level`.
- For "caution" and "avoid" cases, confirm that an Alert with the corresponding `status.note` appears beneath the title.
- For theme-specific components like [Stars](https://cmsgov.github.io/design-system/branch/tamara/WNMGDS-3360/add-status-indicator-component/components/stars/?theme=core) or [Card](https://cmsgov.github.io/design-system/branch/tamara/WNMGDS-3360/add-status-indicator-component/components/card/?theme=core). If the current theme matches the `targetTheme`, no extra alert should appear.
  
> **Note:** Content approval may still be needed for the language of our badge component and status notes.  
> Here are some of the options we're considering:

| Status       | Option 1         | Option 2           | Option 3             | Option 4               | Option 5               |
|--------------|------------------|---------------------|-----------------------|-------------------------|-------------------------|
| Use          | Use Freely       | Fully Supported     | Recommended           | Standard                | Preferred               |
| Caution      | Use with caution | Limited Support     | Proceed with caution  | Use only if necessary   | —                       |
| Avoid        | Don’t use        | Not Supported       | Deprecated            | Unsupported             | Not recommended         |

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] Selected appropriate release milestone

PR #3597, opened 2025-05-20
[WNMGDS-3360] Add `StatusIndicator` component to docs. 

tamara-corbalt:
## Summary
- Adds `StatusIndicator` components from [pr-3595](https://github.com/CMSgov/design-system/pull/3595), omitting any logic that wires it into the `PageHeader` while we wait for content approval on status badge text and contextual alerts.

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3360)

## Testing Steps
- Confirm the doc site builds locally without errors.

## Checklist

### DIFF
--- a/packages/docs/src/components/layout/StatusIndicator.tsx
+++ b/packages/docs/src/components/layout/StatusIndicator.tsx
@@ -1,0 +1,54 @@
+import { Badge, AlertIcon, CheckIcon } from '@cmsgov/design-system';
+import { StatusInterface } from '../../helpers/graphQLTypes';
+import React from 'react';
+
+type StatusLevel = StatusInterface['level'];
+
+interface StatusIndicatorProps {
+  level: StatusLevel;
+}
+
+const statusConfig: Record<
+  StatusLevel,
+  {
+    variation: 'success' | 'warn' | 'alert';
+    label: string;
+    Icon: React.ElementType;
+    className: string;
+  }
+> = {
+  use: {
+    variation: 'success',
+    label: 'Use',
+    Icon: CheckIcon,
+    className: 'ds-u-margin-right--1',
+  },
+  caution: {
+    variation: 'warn',
+    label: 'Caution',
+    Icon: AlertIcon,
+    className: '',
+  },
+  avoid: {
+    variation: 'alert',
+    label: 'Avoid',
+    Icon: AlertIcon,
+    className: '',
+  },
+};
+
+export const StatusIndicator: React.FC<StatusIndicatorProps> = ({ level }) => {
+  const { variation, label, Icon, className } = statusConfig[level];
+
+  return (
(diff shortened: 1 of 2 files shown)


packages/docs/src/components/layout/StatusIndicator.tsx
```diff
@@ -0,0 +1,54 @@
...
+  level: StatusLevel;
+}
+
+const statusConfig: Record<
+  StatusLevel,
+  {
+    variation: 'success' | 'warn' | 'alert';
+    label: string;
+    Icon: React.ElementType;
+    className: string;
```
kim-cmsds, line 17: What do you think about changing this to "optional" so lines where `className: ''` are no longer needed?

Also, changing this to `iconClassName` might make it more clear that this class will be applied to the `Icon` component.

packages/docs/src/components/layout/StatusIndicator.tsx
```diff
@@ -0,0 +1,54 @@
...
+type StatusLevel = StatusInterface['level'];
+
+interface StatusIndicatorProps {
+  level: StatusLevel;
+}
+
+const statusConfig: Record<
+  StatusLevel,
+  {
+    variation: 'success' | 'warn' | 'alert';
```
kim-cmsds, line 14: These variations relate to the `Badge` component, is it possible to leverage that component's types here?

kim-cmsds (approved):
I left a couple of questions but this PR is ready to merge as is!

packages/docs/src/components/layout/StatusIndicator.tsx
```diff
@@ -0,0 +1,53 @@
+import { Badge, AlertIcon, CheckIcon } from '@cmsgov/design-system';
+import { BadgeVariation } from '@cmsgov/design-system/dist/react-components/types/Badge/Badge.js';
```
tamara-corbalt, line 2: @kim-cmsds  - To address your feedback [here](https://github.com/CMSgov/design-system/pull/3597#discussion_r2098951891), I went ahead and imported the `BadgeVariation` type. That said, I wasn’t sure how to do this without reaching into the `dist` folder, which feels a little off — I haven’t seen other examples of this in the doc site.
The package still builds and runs locally without issues, but let me know if you see any risks with this approach or would prefer I revert to inlining the type. Happy to adjust either way!

kim-cmsds: @tamara-corbalt thanks for digging in here! I would agree that we should inline the type and add a comment that it relates to the BadgeVariation type

tamara-corbalt: Ok awesome — I’ll make that change 😊 Thanks for taking another look!

PR #3602, opened 2025-05-21
[NO-TICKET] Doc Site Analytics - set environment at build time not on the client

jack-ryan-nava-pbc:
## Summary

- We now set the env variable once, at build time.
- It doesn't need to be in state, since it won't change.
- We no longer alter the env value in an effect, because it has been set already at build time. One small caveat of this is that we are not longer checking for other testing environments. We can handle that when we have a dev env set up.
- Another note: we will get `prod` as an env value if you run `gatsby serve` locally. 

## How to test

2. Open the doc site locally
11. Confirm that the object contains a field that says "view", along with a "page_name" equal to the name of the page plus the theme name.
12. Confirm that the environment is correct in both development (should be 'dev') and when running gatsby server (should be 'prod')

## Checklist

### DIFF
--- a/packages/docs/src/components/layout/Layout.tsx
+++ b/packages/docs/src/components/layout/Layout.tsx
@@ -68,51 +68,29 @@ const Layout = ({
   theme,
   tableOfContentsData,
 }: LayoutProps) => {
-  const [env, setEnv] = useState<'dev' | 'github-demo' | 'prod' | undefined>(undefined);
+  const env = process.env.NODE_ENV === 'production' ? 'prod' : 'dev';
   const baseTitle = theme === 'core' ? 'CMS Design System' : getThemeData(theme).longName;
   const tabTitle = frontmatter?.title ? `${frontmatter.title} - ${baseTitle}` : baseTitle;
 
   const pageId = slug ? `page--${slug.replace('/', '_')}` : null;
 
-  useEffect(() => {
-    if (env) {
-      const analyticsPayload = {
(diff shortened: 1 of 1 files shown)


packages/docs/src/components/layout/Layout.tsx
```diff
@@ -68,51 +68,29 @@ const Layout = ({
...
-      default:
-        setEnv('prod');
-    }
-  }, [location.hostname]);
+    const analyticsPayload = {
+      content_language: 'en',
+      content_type: 'html',
+      logged_in: 'false',
+      page_name: tabTitle,
+      page_type: tabTitle.includes('Page not found') ? 'true' : 'false', //If page is a 404 (error page) this is set to true, otherwise it is false
```
tamara-corbalt, line 87: Non-blocking super nit: maybe add a space after `//` for consistency?

tamara-corbalt (approved):
Verified the testing steps with no issues—looks great to me! Just two teeny-weeny, non-blocking nits. Approving now 😊

packages/docs/src/components/layout/Layout.tsx
```diff
@@ -68,51 +68,29 @@ const Layout = ({
...
+      site_environment: env, //Used to include or exclude traffic from different testing environments. Ex: test, test0, imp, production
```
tamara-corbalt, line 88: NB: (the teeeeeensiest of tiny nits) — same as above, just missing a space after `//` 😊

PR #3607, opened 2025-05-23
[NO-TICKET] Clean up ds-drawer story

jack-ryan-nava-pbc:
## Summary

Our Drawer story hides the demonstrations on the landing page via a `NoStoryDocTemplate.mdx` file. I initially tried to follow the procedure I did for the ds-modal stories, but realized that
a. The solution I had was more brittle as it relied on a particular DOM structure
b. The ds-drawer elements looked weird rendering in tine iFrames
c. The Drawer component stories are hidden on the doc page for similar display issues

So I followed that route and created a new template for this one. Open to suggestions!

## How to test

1. ds-drawer stories should not display on the doc page but should display and function normally on the subsequent stories

## Checklist

### DIFF
--- a/.storybook/docs/NoStoryWebComponentDocTemplate.mdx
+++ b/.storybook/docs/NoStoryWebComponentDocTemplate.mdx
@@ -1,0 +1,21 @@
+import { Meta, Title, Subtitle, Description, Primary, Stories } from '@storybook/blocks';
+import { WebComponentAttrs } from './WebComponentAttrs';
+import { WebComponentArgsTable } from './WebComponentArgsTable';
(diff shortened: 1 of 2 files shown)


tamara-corbalt (approved):
Looks great —this solution makes sense to me!

Non-blocking: Since the `ds-drawer` and `Drawer` docs pages don’t include embedded examples like most of our other components, do you think it’s worth adding a note directing users to the dedicated interactive stories?

jack-ryan-nava-pbc:
@tamara-corbalt I think that I don't want to add content at this moment. I'll open a ticket though for us to consider doing more devex interviews, maybe specifically related to our Storybook set up.

PR #3626, opened 2025-05-30
[WMNGDS-3373] Extend `VerticalNavItem` to conditionally render status icon

tamara-corbalt:
## Summary

- Adds a `StatusIcon` component that conditionally renders a `use`, `caution`, or `avoid` icon based on the `status` prop. This supports the status indicator work for the `SideNav` on the doc site.
- Updates `VerticalNavItem` to render a `StatusIcon` when a `status` prop is provided.
- Adds test coverage for conditional status icon rendering in `VerticalNavItem`.

[Jira ticket](https://jira.cms.gov/browse/WNMGDS-3373)

## How to test

- `npm run test:unit`

## Checklist


packages/design-system/src/components/VerticalNav/VerticalNavItem.test.tsx
```diff
@@ -29,6 +29,47 @@ describe('VerticalNavItem', () => {
     expect(labelEl).toMatchSnapshot();
   });
 
+  describe('Status icon', () => {
```
kim-cmsds, line 32: nice!

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
```diff
@@ -60,6 +61,11 @@ export interface VerticalNavItemProps {
    * If this item is currently selected
    */
   selected?: boolean;
+  /**
```
jack-ryan-nava-pbc, line 64: Since these are for internal doc site use only, I think they should be hidden on Storybook? What do you think @tamara-corbalt? They currently appear in the props table for the VerticalNavItem.

jack-ryan-nava-pbc (commented):
Looks great! Just one question and then happy to approve.

packages/design-system/src/components/VerticalNav/VerticalNavItem.tsx
tamara-corbalt: Yep good call out! Will update

jack-ryan-nava-pbc: TIL about @ignore!

PR #3632, opened 2025-06-05
[WNMGDS-3414] Fix Textfield regex patterns

jack-ryan-nava-pbc:
## Summary

- [Ticket](https://jira.cms.gov/browse/CMSDS-3414)
- Problem: Chrome appends a `/v` modifier to pattern regexes. This means we need to escape certain characters. This modifier leads to the following console error:
Pattern attribute value [0-9-]* is not a valid regular expression: Uncaught SyntaxError: Invalid regular expression: /[0-9-]*/v: Invalid character class

By escaping the `-` and `{}` characters we prevent this error. Note there is a double escape character because the regexes are turned into strings, and the original escape character is removed.

## How to test

1. Run locally
2. Go to the `components-textfield--all-masked-fields` story under TextField in Storybook
3. Confirm there is no console error.

## Checklist

### DIFF
--- a/packages/design-system/src/components/TextField/Mask.test.tsx
+++ b/packages/design-system/src/components/TextField/Mask.test.tsx
@@ -48,7 +48,7 @@ describe('Mask', function () {
     const input = screen.getByRole('textbox');
     expect(input.classList).toContain('ds-c-field--ssn');
     expect(input).toHaveAttribute('type', 'text');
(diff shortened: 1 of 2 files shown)


tamara-corbalt (approved):
Looks good! Ran it locally and didn't see the `Pattern attribute value [0-9-]* is not a valid regular expression:` console error.

PR #3618, opened 2025-05-28
DS-ISSUE-3616: Add analytics support to the DS Tooltip (React and web…

netboss:
## Summary

- [MSS-798](https://jira.cms.gov/browse/MSS-798): [DS-ISSUE-3616](https://github.com/CMSgov/design-system/issues/3616): Add analytics support to the DS Tooltip (React and web) component.
- Medicare.gov Jira: https://jira.cms.gov/browse/MSS-798

## How to test

1. Instructions on how to test the changes in this PR.

- [ ] Created or updated unit tests to cover any new or modified code
- [ ] Verified that running both `npm run test:unit` and `npm run test:browser:all` were each successful
- [ ] If necessary, updated unit-test snapshots (`npm run test:unit:update`) and browser-test snapshots (`npm run test:browser:all:update`)

Issue #3616, opened 2025-05-28
Tooltip React and web component analytics support


netboss:
## What

- Reviewing the Tooltip component in the Design System we noticed that it doesn't provide analytics support. 
- This issue is to propose a native implementation of analytics tracking for the Tooltip component in the Design System.

## Why

- In Medicare.gov there's a request to implement analytics tracking for the Tooltip user actions. 
- This solves analytics tracking of the Tooltip component.
- Checked the Design System and this feature doesn't exists.
- Yes, right now we're starting the Tooltip analytics tracking in Medicare.gov, and in the future there could be the need to reuse this in CMS.gov and Healthcare.gov, among other government web sites.

PR #3633, opened 2025-06-06
[WNMGDS-3412] Add Analytics Support to Tooltip

jack-ryan-nava-pbc:
## Summary

- [Ticket](https://jira.cms.gov/browse/CMSDS-3412)
- Added analytics support to Tooltip 
- Based on [this contribution](https://github.com/CMSgov/design-system/pull/3618)
- Address the React portion of [this ticket](https://github.com/CMSgov/design-system/issues/3616)
- Web component work will be forthcoming

## How to test

1. Run unit tests locally, confirm they work
2. Follow this procedure for examining analytics events:
3. The best way to test is using the debugger built into your Browser of choice, for Firefox:
4. Open the storybook site locally
5. Open dev tools and go to the Debugger tab
6. Press "Command + P" and search for "utag.js"
7. When that file is opened, search "Command + F" for track:
8. Place a break point at the line following the definition of the track function
9. Hover over the Tooltip
10. In the Browser console, type "a"
11. Look at the associated object, and expand it to see the data property on that object
12. Confirm that the object contains a field that says "link", along with the following object structure (or something similar) in the data field:

{
    "event_extension": "Design system integration",
    "event_name": "tooltip_viewed",
    "parent_component_heading": " ",
    "parent_component_type": " ",
    "text": "tooltip trigger",
  }

## Checklist

### DIFF
--- a/packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
+++ b/packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
@@ -1,0 +1,43 @@
+import { TooltipProps } from './Tooltip';
+import { config } from '../config';
+import { getAnalyticsContentFromRefs, eventExtensionText } from '../analytics';
+import { useRef } from 'react';
+
+export default function useTooltipAnalytics({
+  analytics,
+  analyticsLabelOverride,
+  analyticsParentHeading,
+  analyticsParentType,
+  onAnalyticsEvent = config().defaultAnalyticsFunction,
+  ariaLabel,
+  triggerAriaLabel,
+}: TooltipProps) {
+  const contentRef = useRef();
+
+  function sendTooltipEvent() {
+    if (analytics !== true && (!config().tooltipSendsAnalytics || analytics === false)) {
+      return;
+    }
+
+    // In the case where an icon is supplied there won't be any internal text of the tooltip trigger.
+    // So, getAnalyticsContentFromRefs will return undefined. In that case, grab the aria label for the trigger.
+    // In the case of webcomponents, we need to grab the triggerAriaLabel.
+    const tooltipText =
+      analyticsLabelOverride ??
+      getAnalyticsContentFromRefs([contentRef]) ??
+      ariaLabel ??
+      triggerAriaLabel;
+    const tooltipParentHeading = analyticsParentHeading ?? ' ';
+    const tooltipParentType = analyticsParentType ?? ' ';
+
+    return onAnalyticsEvent({
+      event_name: 'tooltip_viewed',
+      event_extension: eventExtensionText,
+      text: tooltipText,
+      parent_component_heading: tooltipParentHeading,
+      parent_component_type: tooltipParentType,
+    });
+  }
+
+  return { contentRef, sendTooltipEvent };
+}
--- a/packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
+++ b/packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
@@ -12,7 +12,7 @@ export default function useTooltipAnalytics({
   ariaLabel,
   triggerAriaLabel,
 }: TooltipProps) {
-  const contentRef = useRef();
+  const contentRef = useRef<HTMLElement>();
 
   function sendTooltipEvent() {
     if (analytics !== true && (!config().tooltipSendsAnalytics || analytics === false)) {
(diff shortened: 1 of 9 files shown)


packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
```diff
@@ -0,0 +1,43 @@
...
+    // In the case where an icon is supplied there won't be any internal text of the tooltip trigger.
+    // So, getAnalyticsContentFromRefs will return undefined. In that case, grab the aria label for the trigger.
+    // In the case of webcomponents, we need to grab the triggerAriaLabel.
+    const tooltipText =
+      analyticsLabelOverride ??
+      getAnalyticsContentFromRefs([contentRef]) ??
+      ariaLabel ??
+      triggerAriaLabel;
+    const tooltipParentHeading = analyticsParentHeading ?? ' ';
+    const tooltipParentType = analyticsParentType ?? ' ';
```
tamara-corbalt: Question: Does using a space (`' '`) here as a fallback (instead of an empty string) have a technical or semantic reason behind it?

packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
```diff
@@ -0,0 +1,43 @@
...
+export default function useTooltipAnalytics({
+  analytics,
+  analyticsLabelOverride,
+  analyticsParentHeading,
+  analyticsParentType,
+  onAnalyticsEvent = config().defaultAnalyticsFunction,
+  ariaLabel,
+  triggerAriaLabel,
+}: TooltipProps) {
+  const contentRef = useRef();
```
tamara-corbalt: Should this be typed for safety and clarity?

packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
jack-ryan-nava-pbc: I actually need to check this with Ivan Latorre who opened the original ticket. I need to confirm what Blast thinks a reasonable fallback should be.

packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
jack-ryan-nava-pbc: It can be, however the thing that is a bit tricky is that we don't know what kind of HTML element this ref will be attached to so we need to support any. So we can't be as specific as the `useThirdPartyExternalLinkAnalytics.ts` function where the type is limited to `useRef<HTMLAnchorElement>();`.

jack-ryan-nava-pbc: We could do something like \<HTMLElement\>

tamara-corbalt: Gotcha, makes sense that we can’t be too specific about the type. Sounds like `HTMLElement` might be a good middle ground

packages/design-system/src/components/Tooltip/useTooltipAnalytics.ts
```diff
@@ -0,0 +1,37 @@
...
+    // In the case where an icon is supplied there won't be any internal text of the tooltip trigger.
+    // So, getAnalyticsContentFromRefs will return undefined. In that case, grab the aria label for the trigger.
+    // In the case of webcomponents, we need to grab the triggerAriaLabel.
+    const tooltipText =
+      analyticsLabelOverride ??
+      getAnalyticsContentFromRefs([contentRef]) ??
+      ariaLabel ??
+      triggerAriaLabel;
+
+    return onAnalyticsEvent({
```
jack-ryan-nava-pbc, line 29: Matching the data layer specification from Blast: https://confluence.cms.gov/pages/viewpage.action?spaceKey=BLSTANALYT&title=AA+tooltip_viewed

tamara-corbalt: Makes sense! In the data layer spec, I’m seeing `event_name` and `text` fields—but not `event_extension`. Is that a more general analytics field that’s documented elsewhere?

tamara-corbalt: Never mind—I see now that `event_extension` is included in most (if not all) of our analytics event objects, so it looks like this is the intended pattern.

PR #3665, opened 2025-07-15
Update to use React's `useId` hook

tamara-corbalt:
## Summary

- Replaced our custom `useId`-like utility function that leverages `lodash/uniqueId` to use React's built in `useId`.
- Updated regex patterns in unit tests to reflect the new ID format generated by `useId`.
[Jira ticket](https://jira.cms.gov/browse/CMSDS-3403)

### Issues with removing `lodash/uniqueId`

In addition to replacing our existing `useId` utility with React’s built-in `useId` hook, another potential acceptance criterion in [this ticket](https://jira.cms.gov/browse/CMSDS-3403) was to remove our dependency on `lodash/uniqueId` entirely, if feasible. As a first step, I updated the utility to leverage React's `useId` (replacing the previous use of `uniqueId` and `useRef`)) while preserving the existing logic for ID prefixes (e.g., `autocomplete--`) in a centralized place. From there, I began debugging test failures. Most were straightforward and only required updates to the regex used in validating generated Ids.

However, as I debugged various failing web component tests, I found that in our web components—which are essentially our React components rendered via Preact—React’s `useId` did not generate reliably unique IDs across multiple instances of the same component, as seen in [this test case](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/web-components/ds-choice/ds-choice.test.tsx#L127-L157). My working hypothesis is that each web component is rendered in an isolated context—such as a `DocumentFragment` or shadow root—which prevents `useId` from coordinating unique IDs across these render contexts. In contrast, `lodash/uniqueId` relies on a global counter, which *does* appear to work across these boundaries.

For now, I’ve retained `lodash/uniqueId` in specific web components likely to have multiple instances like `ds-choice` and `ds-accordion-item`. 

**Additionally**, we can’t fully remove the `lodash/uniqueId` module yet, as it’s still used in our `DateInput` class component. Refactoring it to a functional component would be required before it could use hooks like `useId`.

## How to test

- `npm run test`
- `npm run test:unit:wc`

## Checklist

### DIFF
--- a/packages/design-system/src/components/Alert/Alert.test.tsx
+++ b/packages/design-system/src/components/Alert/Alert.test.tsx
@@ -118,16 +118,20 @@ describe('Alert', function () {
     it('points aria-labelledby to heading', () => {
       const heading = 'Elvis has left the building';
       renderAlert({ heading, variation: 'error' });
+
       const alert = screen.getByRole('alert');
       const id = alert.getAttribute('aria-labelledby');
-      expect(alert.querySelector(`#${id}`).textContent).toContain(`Alert: ${heading}`);
+      const labelEl = document.getElementById(id!);
+
+      expect(labelEl?.textContent).toContain(`Alert: ${heading}`);
     });
 
     it('falls back aria-labelledby to a11y label when no heading is provided', () => {
       renderAlert();
       const alert = screen.getByRole('region');
       const id = alert.getAttribute('aria-labelledby');
-      expect(alert.querySelector(`#${id}`).textContent).toContain('Notice');
+      const labelEl = document.getElementById(id!);
+      expect(labelEl.textContent).toContain('Notice');
     });
   });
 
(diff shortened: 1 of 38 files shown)


tamara-corbalt:
@jack-ryan-nava-pbc  I just discovered a potentially more critical issue after merging in the latest from `jryan/react-18-upgrade`.

While rebuilding the example projects to verify everything, I hit a build error in the `astro-react-17` example. It tries to import useId from react, but since react 17 doesn’t export `useId`, the build fails—which makes perfect sense.

My understanding was that we want to try and maintain backward compatibility for downstream applications that may still be using older versions of react. However, that might not be feasible with this change unless we introduce a wrapper that falls back to our previous implementation when `useId` isn’t available. But at that point, we’re still relying on `lodash/uniqueId`, and I’m not sure the added complexity is worth it.

PR #3463, opened 2025-03-06
[NO-TICKET] Add subcomponents table for web components

jack-ryan-nava-pbc:
## Summary

- So I thought it would be nice to have args tables for subcomponents with our web components like our React components. Here we are!

## How to test

1. Confirm the subcomponents table renders for the ds-accordion and ds-accordion-items

## Checklist

- [X] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [X] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [X] Selected appropriate `Impacts`, multiple can be selected.
- [X] Selected appropriate release milestone

PR #3686, opened 2025-07-31
[CMSDS-3442] Cleanup web component Stories

jack-ryan-nava-pbc:
## Summary

- Mostly this is minor cleanup, adding default options for certain radio buttons mostly. In some cases actions were added and in the case of the `ds-drawer` I removed unneeded state management. 
- I added a simple if check that filters out `undefined` from the attributes table, without it we see the word `undefined` listed as an available attribute, and while that's technically possible, it's not really the point of the documentation. 
- [Ticket](https://jira.cms.gov/browse/CMSDS-3442)

## How to test

1. Run browser tests
2. Visually review changes and make sure they make sense

## Checklist


.storybook/docs/WebComponentArgsTable.tsx
```diff
@@ -8,6 +8,10 @@ function optToCodeBlock(opt: undefined | string) {
 
 function getTypeLabel(argType: any) {
   const controlType = argType.control?.type ?? argType.control;
+  const options = argType?.options;
+  if (options) {
+    return options.filter((type) => type).map((type) => <code>{type}</code>);
```
jack-ryan-nava-pbc, line 13: The goal here is to remove undefined (or any other falsy value) from our options array before returning formatted text.

packages/design-system/src/components/web-components/ds-accordion/ds-accordion-item.stories.tsx
```diff
@@ -55,23 +53,3 @@ export default {
   },
   decorators: [webComponentDecorator],
 };
-
```
jack-ryan-nava-pbc, line 58: I removed this story since in this [pull request](https://github.com/CMSgov/design-system/pull/3463) introduced a subcomponent table using our tabs component. I think it feels redundant. Plus this better matches our React documentation.

packages/design-system/src/components/web-components/ds-accordion/ds-accordion.stories.tsx
```diff
@@ -36,23 +38,34 @@ export default {
...
-      </p>
-    </ds-accordion-item>
-  </ds-accordion>
-);
+const Template = (args) => {
+  useEffect(() => {
+    const accordionItems = Array.from(document.querySelectorAll('ds-accordion-item'));
+    accordionItems.forEach((item) => {
+      item.addEventListener('ds-change', (event) => {
+        action('ds-change')(event);
```
jack-ryan-nava-pbc, line 46: Wiring up a ds-change event handler.

packages/design-system/src/components/web-components/ds-button/ds-button.stories.tsx
```diff
@@ -55,7 +76,7 @@ export default {
   decorators: [webComponentDecorator],
 };
 
-const Template = ({ 'text content': text, args }) => {
```
jack-ryan-nava-pbc, line 58: Because of this destructuring the shape of the args object was slightly different than on other stories, so the button controls were not working.

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.stories.tsx
```diff
@@ -46,7 +46,7 @@ const meta: Meta = {
     },
     'close-button-variation': {
       description: 'Style variation for the close button ("solid" or "ghost").',
-      control: { type: 'select' },
+      control: { type: 'radio' },
```
jack-ryan-nava-pbc, line 49: This is me being nitpicky and possibly a little retentive, I like having all of the options listed out and visible, especially if there aren't a lot.

packages/design-system/src/components/web-components/ds-drawer/ds-drawer.stories.tsx
```diff
@@ -128,8 +128,6 @@ const drawerContent = (
 );
 
 const Template = (args) => {
-  const [drawerOpen, setDrawerOpen] = useState(args['is-open'] ?? false);
```
jack-ryan-nava-pbc, line 131: Got rid of the state management here since we don't need it and it doesn't reflect how we'd use this code in Real Life.

.storybook/docs/WebComponentArgsTable.tsx
tamara-corbalt: Cool! Quick question — if `options` are defined, we’ll return early and prioritize returning the formatted text before we get to checking the `controlType`. Will the type always be correctly inferred as a `radio`, or do we also use other control types like `select` with `options`? I can’t think of any other cases off the top of my head, but just calling this out in case it’s worth tightening the check.

jack-ryan-nava-pbc: Ah so the use of `type` is a bit confusing here. Type in the control object refers to the input type on the control table. The `type`s that are rendered in that table are actually drawn from the options property on whatever argType we're augmenting here.

For example:
```
variation: {
      options: ['default', 'info', 'success', 'warn', 'alert'],
      control: { type: 'radio' },
    },
```

The types that are listed in the args table will be: 'default, info, success, warn, alert' and they will be rendered using a series of radio button inputs. The types in the options array are what we're trying to display as Storybook handles selecting the correct input element.

kim-cmsds (approved):
This looks great! Verified this work against the ticket and ran storybook, plus tests locally.

PR #3688, opened 2025-08-01
[CMSDS-3512] Allow text overrides of screen reader content for the Badge component

jack-ryan-nava-pbc:
## Summary

- Overview of changes
- [Ticket](https://jira.cms.gov/browse/CMSDS-3512)

## How to test

1. Run specs locally
2. Try it out in storybook! The text should change within the `span.ds-u-visibility--screen-reader` element from whatever the variation is to the override text you provide via the control. It should go back to the variation provided text when the prop is undefined or just a " ".

## Checklist

### DIFF
--- a/packages/docs/content/components/badge.mdx
+++ b/packages/docs/content/components/badge.mdx
@@ -37,22 +37,6 @@ medicare:
 
 <StorybookExample componentName="badge" storyId="components-badge--informational" />
 
-## Code
-
-### React
-
-<SeeStorybookForGuidance storyId={'components-badge--docs'} />
-
-### Web Component
-
-<SeeStorybookForGuidance tech="wc" storyId={'web-components-badge--docs'} />
-
-### Style customization
-
-The following CSS variables can be overridden to customize Badge components:
-
-<ComponentThemeOptions componentname="badge" />
-
 ## Guidance
 
 ### When to use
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/Badge/Badge.test.tsx
```diff
@@ -32,4 +32,17 @@ describe('Badge', () => {
...
+    const srText = screen.getByText('barbaz', { exact: false });
+    expect(srText).toBeInTheDocument();
+  });
+
+  it('should not override the screen reader text when the prop is just a space', () => {
+    renderBadge({ variation: 'success', screenReaderText: ' ' });
+    // Note 'Success' is the value provided by the variation, not the screen reader text
+    const srText = screen.getByText('Success', { exact: false });
+    expect(srText).toBeInTheDocument();
+  });
```
jcq: This is the only bit where I have a pause... I'm not certain of how likely it might be, but I could at least imagine a scenario where the content of the badge is sufficient and one might not want any extra text read out to screen readers. With this implementation, there's no way to override it entirely (both empty string and `undefined` will result in the default being read out).

Likely not a huge deal, but just wanted to flag

jack-ryan-nava-pbc: @jcq Good callout! I've just pushed up some changes that introduce the option to hide this field completely via a separate prop. Take a look and let me know if those changes address your concern!

jcq: @jack-ryan-nava-pbc Yup yup, that seems like a good option to account for that edge case. TY!

kim-cmsds (approved):
Adding both props LGTM ✨

PR #3693, opened 2025-08-06
[CMSDS-3502] Adds accessibility guidance to the `LabelMask` documentation page.

tamara-corbalt:
## Summary
- Adds accessibility guidance to the `LabelMask` documentation page.
- Reorders sections on the page so that **Guidance**, **Accessibility**, and **Accessibility testing** appear before the **Code** section.
[Jira ticket](https://jira.cms.gov/browse/CMSDS-3502)

## How to test
- `npm run start` and navigate to `http://localhost:8000/components/text-field/label-masked-field/?theme=cmsgov`
## Checklist

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -9,7 +9,7 @@
 
 ## Checklist
 
-- [ ] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
+- [ ] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
 - [ ] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
 - [ ] Selected appropriate `Impacts`, multiple can be selected.
 - [ ] Selected appropriate release milestone


packages/docs/content/components/text-field/label-masked-field.mdx
```diff
@@ -32,6 +32,45 @@ The label-mask currently has five built-in implementations: _currency_, _phone n
...
+
+- When the input requires a free-form field that doesn't use a common input pattern, masking is not appropriate.
+- When the pattern is too complicated to allow for a valid label mask. A pattern like email, with many possible scenarios for input, is not a good candidate for masking. Allow the user to enter their email address (or other complicated data) and have your validation library confirm before form submission.
+
+### Usage
+
+- Only show error validation messages or stylings after a user has interacted with a particular field
+
+## Accessibility
+- Use a `<fieldset>` to group thematically related form controls, and provide a `<legend>` to label the group. This helps screen reader users navigate the form more easily.
```
jack-ryan-nava-pbc: I think all of this text can be removed. The accessibility concerns are the same as the Textfield. You can refer users to that guidance.

tamara-corbalt: Thanks @jack-ryan-nava-pbc ! I see the overlap with `TextField`. That said, this bullet is actually being removed from `text-field.mdx` in the new [proposed guidance](https://github.com/CMSgov/design-system/pull/3694/files#diff-30435fa7d2152c362168466b92efbeb5b84beba7a2968d062dc1feb7c9894e4a), so it wouldn’t be referenced anywhere anymore. Since grouping related fields with `<fieldset>` and `<legend>` feels like core a11y guidance, I’m wondering if we should add it back to `text-field.mdx` instead—and have `LabelMask` reference it there. What do you think?

jack-ryan-nava-pbc: I like that idea. I think this is also something we can add to our Form guidance. We should consider creating a better Form guidance page with patterns for demonstration purposes.

PR #3702, opened 2025-08-08
[CMSDS-3396] Added `headingClassName` attribute to `AccordionItem` and `heading-class-name` to `ds-accordion-item` and additional examples

kim-cmsds:
## Summary

- Added `headingClassName` to `<AccordionItem />`
- Added `heading-class-name` to `<ds-accordion-item />`
- Refactored Accordion Item example in the Angular example project and demonstrated how to apply consistent heading spacing because the global style css is not able to affect what is rendered within a shadow DOM by design. This is how the `ds-accordion-item`'s components are structured, also by design.
- [JIRA Ticket](https://jira.cms.gov/browse/CMSDS-3396)

## How to test

1. Run the Angular example project and notice that there is 8px of spacing between the top of the second accordion item and the bottom of the first accordion item and then another 8px of spacing between the bottom of second accordion item and the top of the third accordion items
2. Run `npm run test`
3. Run `npm run test:browser:all` to verify that the storybook docs have been updated accordingly

## Checklist


examples/angular/src/app/accordionItemWrapper/accordionItem.component.html
```diff
@@ -1,4 +1,14 @@
-<ds-accordion-item>
+<ds-accordion-item
+  [attr.button-class-name]="buttonClassName()"
+  [attr.content-class-name]="contentClassName()"
+  [attr.default-open]="defaultOpen()"
+  [attr.heading-class-name]="headingClassName()"
+  [attr.heading-level]="headingLevel()"
+  [attr.id]="id()"
+  [attr.is-controlled-open]="isControlledOpen()"
```
jack-ryan-nava-pbc: You know, I'm not sure this prop makes any sense for a web component. We also don't have it in our documentation on Storybook. Maybe remove it here unless you can think of a way to use it? I guess a question could be, how would we remotely open an accordion in the web component case?

ghost: It doesn't make sense for a web component, so I'll remove it. Thank you for flagging!

PR #3717, opened 2025-08-14
[NO-TICKET] Update "Focus Button" medicare/webkit snapshot.

tamara-corbalt:
## Summary
**Background**
In recent `test-nightly` builds, the Button focus interaction test has been extremely flaky in WebKit when using the Medicare theme
The test intentionally focuses the button and compares it to a stored snapshot. However, the Medicare snapshot did not include the expected focus ring, causing a mismatch.

After updating the Medicare snapshot to show the correct focused state, the same failure appeared for another theme (CMSgov, still WebKit).

While reviewing test results, I noticed that our `Textfield` component’s “Button focus” tests are stable and use `getByRole` to select the element, whereas the Button test was using `page.locator`.

**Pull summary**
- This pr updates the Button test to follow the Textfield pattern—using `getByRole` for element selection—which has proven more stable in local runs and on CI.
- It also updates the medicare/webkit snapshot so it reflects the intended focused state.

## How to test

`CI=true npm run test:browser:interaction -- --grep "Button focus" --repeat-each=10 `

## Checklist

- [ ] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.


packages/design-system/src/components/Button/Button.test.interaction.ts
```diff
@@ -14,7 +14,7 @@ describeByTheme((theme) => {
 
   test('Button focus', async ({ page }) => {
     await page.goto(storyUrl('components-button--default', theme));
-    await page.locator('text=Your button text here').focus();
+    await page.getByRole('button').focus();
```
jack-ryan-nava-pbc, line 17: Excellent catch!

PR #3745, opened 2025-09-08
[CMSDS-3492] Adds `ds-third-party-external-link` web component 

tamara-corbalt:
## Summary
- Adds `ds-third-party-externa-link` web component module which include
    - Preact wrapper component
    - Storybook file
    - Unit tests
    - New browser snapshots
    - Proper entry point for the new module so that it is included as an individual bundle as well as in main bundle, `all.js`.
- Adds an example implementation in our Angular example project.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-3492)

## How to test

1. Run unit tests: `npm run test:unit:wc`
2. Verify `ds-third-party-external-link` stories in Storybook
3. In the Angular example (`cd examples/angular && npm run start`), navigate to http://localhost:4200 and scroll to the bottom of the page.

## Checklist


packages/design-system/src/components/web-components/ds-third-party-external-link/ds-third-party-external-link.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+    children: {
+      control: 'text',
+      controlsOnly: true,
+    },
+    href: {
+      description: 'The URL that the link points to.',
+      control: 'text',
+    },
+    origin: {
+      description: 'A short label identifying the third-party origin.',
```
jack-ryan-nava-pbc: This is a bit confusing to me. Looking at the ThirdPartyExternalLink, we only use this prop to set the text in the header, not anything directly related to where the link's going to. I think these descriptions can be taken directly from the React story.

packages/design-system/src/components/web-components/ds-third-party-external-link/ds-third-party-external-link.stories.tsx
```diff
@@ -0,0 +1,71 @@
...
+      componentEvents: {
+        ...analyticsEventDocs,
+      },
+    },
+  },
+  decorators: [webComponentDecorator],
+};
+
+const Template = (args) => (
+  <ds-third-party-external-link {...args}>{args.children}</ds-third-party-external-link>
```
jack-ryan-nava-pbc: It's very strange, but the children arg is not reactive in the Storybook table. It does change, but it seems to require a repaint. I don't think this is something you need to fix. Just weird.

packages/design-system/src/components/web-components/ds-third-party-external-link/ds-third-party-external-link.tsx
```diff
@@ -0,0 +1,44 @@
...
+import { ThirdPartyExternalLink } from '../../ThirdPartyExternalLink';
+import { ThirdPartyExternalLinkProps } from '../../ThirdPartyExternalLink/ThirdPartyExternalLink';
+import { analyticsOverrideAttrs, analyticsParentDataAttrs } from '../shared-attributes/analytics';
+import { parseBooleanAttr } from '../wrapperUtils';
+import { onAnalyticsEvent } from '../analytics';
+
+const attributes = [
+  'aria-described-by',
+  'href',
+  'class-name',
```
jack-ryan-nava-pbc, line 11: I think we need a children attribute here so the link text can be changed. Probably the prop name on the React component is a bit confusing, as it actually targets the external link text rather than the dialog content.

packages/design-system/src/components/web-components/ds-third-party-external-link/ds-third-party-external-link.stories.tsx
jack-ryan-nava-pbc: Actually I'm not sure the children attribute is wired up properly in the web component definition?

packages/design-system/src/components/web-components/ds-third-party-external-link/ds-third-party-external-link.stories.tsx
tamara-corbalt: Thanks for the feedback! I meant to circle back to these descriptions, but forgot. I’ve updated them to match the React story.

packages/design-system/src/components/web-components/ds-third-party-external-link/ds-third-party-external-link.stories.tsx
tamara-corbalt: I’m still trying to recall whether defining children explicitly as an attribute is necessary. We don’t do this in `ds-button` or `ds-alert`, and both still accept `children` via the light DOM.

From what I can tell, the non-responsiveness here isn't necessarily due to the web component definition but the way story is implemented.

packages/design-system/src/components/web-components/ds-third-party-external-link/ds-third-party-external-link.tsx
tamara-corbalt: I take back everything I said above about the story implementation. I forgot about the shadow dom toggle.

tamara-corbalt: Ok, I dug into the [define](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/web-components/preactement/define.ts#L372-L376) code as a refresher and to check whether listing `children` as an attribute is necessary. Short answer is - I don't think so. Here's what I think is happening.

`ds-button` doesn’t list children among its [attributes](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/web-components/ds-button/ds-button.tsx#L6-L18), yet we can update its label in storybook via the children control. That works because shadow DOM is enabled: the browser projects light-DOM children into a `<slot>` and keeps them in sync automatically.

But for `ds-third-party-external-link`, we don’t set `shadow: true`, which means the current light DOM is snapshotted into a `<template>` , parsed into a vnode, and passed into the react component as `children`. Because it’s a snapshot — with the template acting as the source of truth — the label won’t live-update unless we explicitly rebuild the template and re-render.

packages/design-system/src/components/web-components/ds-third-party-external-link/ds-third-party-external-link.stories.tsx
jack-ryan-nava-pbc: As per our conversation, unless HTML is passed into the child attribute the mutation observer isn't going to catch the changes. And since this is just static text, we don't want to encourage people to make it dynamic by making it dynamic here.

PR #3763, opened 2025-09-24
[CMSDS-3547] Replace requiredLabel with hint on ChoiceList Story

### LINKED ISSUE #3726: ChoiceList Story appears to be using the wrong prop
https://github.com/CMSgov/design-system/blob/adcdaa5aefa0c1184966d5334168f8aa2973ed43/packages/design-system/src/components/ChoiceList/ChoiceList.stories.tsx#L19

On the ChoiceList example, the hint text is being rendered with the `requirementLabel` prop. The prop used here should be `hint` as `requirementLabel` is meant for that text that says "Required" or "Optional" before the hint text.

derek-cmsds:
## Summary

- Updated the ChoiceList example to use the `hint` prop instead of the `requirementLabel` prop to render the hint text. `requirementLabel` is meant for that text that says "Required" or "Optional" before the hint text.
- CMSDS-3547
- Closes #3726 

## How to test

1. Run `npm run storybook`
2. Navigate to the `ChoiceList` docs
3. Under the example, click "Show code"
4. Verify the second choice has a `hint` prop with the value "Choice hint text"

## Checklist

### DIFF
--- a/packages/design-system/src/components/ChoiceList/ChoiceList.stories.tsx
+++ b/packages/design-system/src/components/ChoiceList/ChoiceList.stories.tsx
@@ -26,6 +26,7 @@ const meta: Meta<typeof ChoiceList> = {
     errorMessage: { control: 'text' },
     hint: { control: 'text' },
     ref: { table: { disable: true } },
+    requirementLabel: { control: 'text' },
     size: {
       options: [undefined, 'small'],
       control: { type: 'radio' },
--- a/packages/design-system/src/components/ChoiceList/ChoiceList.stories.tsx
+++ b/packages/design-system/src/components/ChoiceList/ChoiceList.stories.tsx
@@ -16,7 +16,7 @@ const meta: Meta<typeof ChoiceList> = {
           console.log('I am a ref callback being called!');
         },
       },
-      { label: 'Choice 2', requirementLabel: 'Choice hint text', value: 'B' },
+      { label: 'Choice 2', hint: 'Choice hint text', value: 'B' },
     ],
     // https://github.com/CMSgov/design-system/pull/3003#discussion_r1545916584
     onBlur: action('onBlur'),
@@ -26,7 +26,6 @@ const meta: Meta<typeof ChoiceList> = {
     errorMessage: { control: 'text' },
     hint: { control: 'text' },
     ref: { table: { disable: true } },
-    requirementLabel: { control: 'text' },
     size: {
       options: [undefined, 'small'],
       control: { type: 'radio' },
@@ -140,7 +139,7 @@ export const ChoiceChildren: Story = {
       },
       {
         label: 'Choice 2',
-        requirementLabel: 'Choice hint text',
+        hint: 'Choice hint text',
         value: 'B',
         checkedChildren: (
           <div className="ds-c-choice__checkedChild">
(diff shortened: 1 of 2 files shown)


packages/design-system/src/components/ChoiceList/ChoiceList.stories.tsx
```diff
@@ -26,7 +26,6 @@ const meta: Meta<typeof ChoiceList> = {
     errorMessage: { control: 'text' },
     hint: { control: 'text' },
     ref: { table: { disable: true } },
-    requirementLabel: { control: 'text' },
```
jack-ryan-nava-pbc, line 29: I think we want to leave this in here. Otherwise it defaults to using a React Node which then requires users to set an object in the control. This prop can take strings, as our tests prove. React Nodes can be a String (as well as pretty much anything else).

packages/design-system/src/components/ChoiceList/ChoiceList.test.tsx
```diff
@@ -68,13 +68,19 @@ describe('ChoiceList', () => {
...
+        }),
+      });
       const choiceEls = screen.getAllByRole('radio');
       const choice = choiceEls[0] as HTMLInputElement;
+      const hintTexts = screen.getAllByText('Choice hint text');
 
       expect(choiceEls.length).toBe(numChoices);
       expect(choice.name).toBe('spec-field');
       expect(choice.value).toBe('1');
+      expect(hintTexts[0].tagName).toBe('P');
```
jack-ryan-nava-pbc, line 83: Are you trying to have a specific test case for rendering of the Hint Text? You can do that more simply if that's the case just by querying by the text you want to find, and expecting it to equal the prop you pass to the renderChoiceList object argument.

derek-cmsds: I was making sure that switching from `requirementLabel` to `hint` did not change the element outputted by the component.

packages/design-system/src/components/ChoiceList/ChoiceList.stories.tsx
derek-cmsds: This has been restored.

packages/design-system/src/components/ChoiceList/ChoiceList.stories.tsx
```diff
@@ -140,7 +140,7 @@ export const ChoiceChildren: Story = {
       },
       {
         label: 'Choice 2',
-        requirementLabel: 'Choice hint text',
```
jack-ryan-nava-pbc, line 143: Can we add in a requirement label for a story stating that a checkbox is required?

derek-cmsds: What is the use case for when an individual checkbox or radio button within a ChoiceList would be required instead of the ChoiceList itself?

derek-cmsds: In other words, when should we advise people to use the choice's `requirementLabel` prop instead of the ChoiceList's `requirementLabel` prop?

jack-ryan-nava-pbc: I'm not sure that's actually something we've spelled out before? I could imagine a situation where many options are not required, but perhaps only one is. I feel like cookie managers often use that technique.

derek-cmsds: It... kinda feels like an anti-pattern to me. I'm going to abstain from adding a story for it for now. If we come up with a compelling reason to add a story for it later, we can make that a separate ticket.

jack-ryan-nava-pbc: Sounds good! thanks for the feedback

PR #3769, opened 2025-09-26
[CMSDS-3527] Only render "Learn more" link when learnMoreUrl is defined

derek-cmsds:
## Summary

- The `learnMore` prop is optional, however the text/link that the prop applies was displaying whether the prop is applied or not. This PR updates the "Learn more" link to only display when `learnMoreUrl` is defined.
- [CMSDS-3527](https://jira.cms.gov/browse/CMSDS-3527)

## How to test

1. Pull down the branch and run `npm start`
2. Navigate to the [default `ThirdPartyExternalLink` story](http://localhost:8000/storybook/?path=/story/components-thirdpartyexternallink--default) in Storybook
3. Click the link and verify the modal appears
4. Verify the modal includes a link with the text "Learn more about links to third-party sites."
5. In the controls section, remove the `learnMoreUrl` control value.
6. Verify the "Learn more about links to third-party sites." link disappears.

## Checklist


packages/design-system/src/components/ThirdPartyExternalLink/ThirdPartyExternalLink.test.tsx
```diff
@@ -59,6 +59,17 @@ describe('ThirdPartyExternalLink', () => {
       expect(learnMoreLink.getAttribute('href')).toBe('https://www.google.com/');
     });
 
+    describe('when no learnMoreUrl is defined', () => {
```
tamara-corbalt, line 62: Nice test! I like that this is scoped inside a dedicated describe block—it adds clarity. We might want to consider more granular groupings like this for better readability across our test suites.

NB: If there are other tests related to this prop, maybe we can generalize the describe name a bit and move them into this block too?

tamara-corbalt: NB: Just a heads up: we may also want to add this test for the `ds-third-party-link` web component. Unfortunately, that component doesn’t exist on main yet, but something to keep in mind once it's merged.

tamara-corbalt (approved):
Looks great! Added two non-blocking suggestions

jack-ryan-nava-pbc (approved):
Tested locally with the Core version. This looks good to me.  Good work!

packages/design-system/src/components/ThirdPartyExternalLink/ThirdPartyExternalLink.test.tsx
derek-cmsds: > NB: If there are other tests related to this prop, maybe we can generalize the describe name a bit and move them into this block too?

My philosophy around `describe` blocks is that their primary purpose is to provide context to programmers. (In RSpec, there's a `context` alias for `describe` just for that purpose.) I want my tests to read like a sentence, both in the test readout and within the testing file. So, in this instance, that sentence is "ThirdPartyExternalLink's dialog (when no learnMoreUrl is defined) does not render the Learn More link." There may be other possible endings to that sentence we would want to add later, but I'm also fine with that being the only ending to that sentence if that's the only ending that makes sense. There's certainly the risk of run-on sentences and taking the conceit too far, but I am okay with a little superfluous nesting in test blocks if it makes the test easier for a human to comprehend at a glance.

jack-ryan-nava-pbc: Ah Ruby, a true _gem_ among programming languages

PR #3780, opened 2025-09-30
[CMSDS-3444] Remove aria-invalid attribute if ChoiceList is a checkbox group.

derek-cmsds:
## Summary

- The `aria-invalid` attribute is currently applied directly to the <fieldset> element on `ChoiceList` components. According to WAI-ARIA specifications, aria-invalid is a supported and valid attribute for elements with the "radiogroup" role, but not for fieldset elements that do not possess that role.
- This PR updates `ChoiceList` to only show the `aria-invalid` attribute when the `type` prop is set to 'radio'.
- [CMSDS-3444](https://jira.cms.gov/browse/CMSDS-3444)

## How to test

## Checklist

### DIFF
--- a/packages/design-system/src/components/ChoiceList/ChoiceList.tsx
+++ b/packages/design-system/src/components/ChoiceList/ChoiceList.tsx
@@ -82,11 +82,11 @@ export type ChoiceListProps = BaseChoiceListProps &
  * [HTML input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input).
  */
 export const ChoiceList = (props: ChoiceListProps) => {
-  const { onBlur, onComponentBlur, choices } = props;
+  const { onBlur, onComponentBlur, choices, type } = props;
   const id = useId('choice-list--', props.id);
 
   if (process.env.NODE_ENV !== 'production') {
-    if (props.type !== 'checkbox' && props.choices.length === 1) {
+    if (type !== 'checkbox' && choices.length === 1) {
       console.warn(
         `[Warning]: Use type="checkbox" for components with only one choice. A single radio button is disallowed because it prevents users from deselecting the field.`
       );
@@ -126,7 +126,7 @@ export const ChoiceList = (props: ChoiceListProps) => {
       onBlur: handleBlur,
       onChange: props.onChange,
       size: props.size,
-      type: props.type,
+      type,
       inputClassName: classNames(choiceProps.inputClassName, {
         'ds-c-choice--error': props.errorMessage,
       }),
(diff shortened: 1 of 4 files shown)


packages/design-system/src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -82,11 +82,11 @@ export type ChoiceListProps = BaseChoiceListProps &
  * [HTML input element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input).
  */
 export const ChoiceList = (props: ChoiceListProps) => {
-  const { onBlur, onComponentBlur, choices } = props;
+  const { onBlur, onComponentBlur, choices, type } = props;
```
derek-cmsds, line 85: We use `type` enough times in this file that it makes sense to extract it from `props`.

packages/design-system/src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -82,11 +82,11 @@ export type ChoiceListProps = BaseChoiceListProps &
...
   const id = useId('choice-list--', props.id);
 
   if (process.env.NODE_ENV !== 'production') {
-    if (props.type !== 'checkbox' && props.choices.length === 1) {
```
derek-cmsds, line 89: We already extract `choices` above, might as well use it here.

packages/design-system/src/components/ChoiceList/ChoiceList.tsx
```diff
@@ -149,12 +149,22 @@ export const ChoiceList = (props: ChoiceListProps) => {
     return <Choice key={choiceProps.value} {...completeChoiceProps} />;
   });
 
+  const addDynamicAttrs = () => {
+    const attrs: React.ComponentPropsWithRef<'fieldset'> = {};
+
+    if (type === 'radio') {
+      attrs['aria-invalid'] = invalid;
+      attrs.role = 'radiogroup';
```
derek-cmsds, line 159: Similar to `aria-invalid`, we should only explicitly include `role` if `type` is 'radio'.

packages/design-system/src/components/InlineError/useInlineError.tsx
```diff
@@ -76,7 +76,7 @@ export function useInlineError<T extends UseInlineErrorProps>(props: T) {
 
   // If the user has provided an `aria-invalid` attribute, use that as the source
   // of truth; otherwise, it's invalid if there's an error message.
-  const invalid = props['aria-invalid'] ?? !!errorMessage;
+  const invalid: React.AriaAttributes['aria-invalid'] = props['aria-invalid'] ?? !!errorMessage;
```
derek-cmsds, line 79: TypeScript threw a fit because `aria-invalid` accepts only a few specific values.

packages/design-system/src/components/ChoiceList/ChoiceList.tsx
jack-ryan-nava-pbc: Just checking here because you've looked more than I have, but there are no valid ARIA roles for groups of checkboxes right? I don't think there are, just double checking.

derek-cmsds: Not for _groups_ of checkboxes. Individual checkbox inputs have an implicit ARIA role of "checkbox" that can support an `aria-invalid` attribute, but adding support for that should probably be its own ticket as it would focus on the `Choice` component and not `ChoiceList`.

jack-ryan-nava-pbc: Cool. It seems we do not do that currently. I'll open a ticket for that fix.

PR #3783, opened 2025-09-30
[CMSDS-3467] Coerce field values to string and add `coerceToString` utility

tamara-corbalt:
## Summary
- Updates the `Mask` component to ensure field values are always treated as strings, in line with stricter React 19 type definitions.
- Adds a utility to coerce field values to a string type.
- Adds test coverage for the new utility

## How to test
- Run `npm run build`
- Run `npm run test`

## Notes
This is **PR 3 of 5** in the broader effort to upgrade React types to v19.  
Please review after [PR 3782](https://github.com/CMSgov/design-system/pull/3782) has been merged.

## Checklist

### DIFF
--- a/packages/design-system/src/components/TextField/LabelMask.tsx
+++ b/packages/design-system/src/components/TextField/LabelMask.tsx
@@ -1,6 +1,7 @@
 import { Children, cloneElement } from 'react';
 import type * as React from 'react';
 import { useLabelMask, MaskFunction } from './useLabelMask';
+import { TextInputProps } from './TextInput';
 
 export interface LabelMaskProps {
   /**
@@ -21,7 +22,7 @@ export interface LabelMaskProps {
 }
 
 const LabelMask = (props: LabelMaskProps) => {
-  const field = Children.only(props.children as React.ReactElement);
+  const field = Children.only(props.children) as React.ReactElement<TextInputProps>;
   const { labelMask, inputProps } = useLabelMask(props.labelMask, field.props);
   const input = cloneElement(field, inputProps);
 
--- a/packages/design-system/src/components/TextField/maskHelpers.ts
+++ b/packages/design-system/src/components/TextField/maskHelpers.ts
@@ -126,3 +126,16 @@ export function unmaskValue(value?: string, mask?: string): string {
 
   return value;
 }
+
+/**
+ * Coerce a string | number | undefined into a safe string.
+ * - string -> returned as-is
+ * - number -> converted with String()
+ * - undefined/null -> ''
+ */
+export function coerceToString(value: string | number | null | undefined): string {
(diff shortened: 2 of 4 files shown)


packages/design-system/src/components/TextField/LabelMask.tsx
```diff
@@ -21,7 +22,7 @@ export interface LabelMaskProps {
 }
 
 const LabelMask = (props: LabelMaskProps) => {
-  const field = Children.only(props.children as React.ReactElement);
+  const field = Children.only(props.children) as React.ReactElement<TextInputProps>;
```
jack-ryan-nava-pbc, line 25: Using big "C" Children seems to be on the outs with React. Looking [here](https://react.dev/reference/react/Children#children-only) it seems there is possibly another route to handling this particular verification that you're doing. Which, actually could maybe be handled by typescript? It seems that the logic here is to ensure that the children element only is a Text field at runtime? Does that scan to you?

packages/design-system/src/components/TextField/maskHelpers.ts
```diff
@@ -126,3 +126,16 @@ export function unmaskValue(value?: string, mask?: string): string {
...
 }
+
+/**
+ * Coerce a string | number | undefined into a safe string.
+ * - string -> returned as-is
+ * - number -> converted with String()
+ * - undefined/null -> ''
+ */
+export function coerceToString(value: string | number | null | undefined): string {
+  if (value === null || value === undefined) {
```
jack-ryan-nava-pbc, line 137: SUPER NIT: Feel free to ignore weird guy who cares about type coercion in Javascript (and also gets it wrong with some frequency) but if you wanted to be less verbose here you could just do:
```
if (!value) {
   return ""
} 
```
technically that'll return "" when value === "" but the benefit is you also catch cases where it's false?

jack-ryan-nava-pbc: Which should be handled by typescript. IDK, ignore me.

packages/design-system/src/components/TextField/LabelMask.tsx
jack-ryan-nava-pbc: Gonna spin up a separate ticket to remove Children in this file.

packages/design-system/src/components/TextField/maskHelpers.ts
jack-ryan-nava-pbc: Forgot about 0! Ignore me.

packages/design-system/src/components/TextField/LabelMask.tsx
tamara-corbalt: Just added this [ticket](https://jira.cms.gov/browse/CMSDS-3628)

PR #3784, opened 2025-09-30
[CMSDS-3467] Resolve remaining React 19 semantic errors in Table, IdleTimeout, Autocomplete, Tabs, and useDialog

tamara-corbalt:
## Summary
- React 19 semantic errors in `Table`, `IdleTimeout`, `Autocomplete`, `Tabs`, and `useDialog`

## How to test
- Run `npm run build`
- Run `npm run test`

## Notes
This is **PR 4 of 5** in the broader effort to upgrade React types to v19.  
Please review after [PR 3783](https://github.com/CMSgov/design-system/pull/3783) has been merged.

## Checklist

### DIFF
--- a/packages/design-system/src/components/Dialog/useDialog.ts
+++ b/packages/design-system/src/components/Dialog/useDialog.ts
@@ -24,7 +24,7 @@ export function useDialog<ReturnType = any>(
   renderDialog: (renderProps: UseDialogRenderProps<ReturnType>) => React.ReactElement
 ) {
   const [isOpen, setIsOpen] = useState(false);
-  const resolveRef = useRef<(returnValue: ReturnType) => any>();
+  const resolveRef = useRef<(returnValue: ReturnType) => any>(null);
 
   const resolveClose = (returnValue: ReturnType) => {
     setIsOpen(false);
--- a/packages/design-system/src/components/Table/TableRow.tsx
+++ b/packages/design-system/src/components/Table/TableRow.tsx
@@ -22,7 +22,7 @@ export const TableRow = ({ children, _isTableHeadChild, ...tableRowProps }: Tabl
     return Children.map(children, (child: React.ReactElement) => {
       // Extend props before rendering.
       if (child && child.props) {
-        return cloneElement(child, {
+        return cloneElement(child as React.ReactElement<any>, {
           _isTableHeadChild: _isTableHeadChild,
         });
       }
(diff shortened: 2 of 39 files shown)


packages/design-system/src/components/Dialog/useDialog.ts
```diff
@@ -24,7 +24,7 @@ export function useDialog<ReturnType = any>(
   renderDialog: (renderProps: UseDialogRenderProps<ReturnType>) => React.ReactElement
 ) {
   const [isOpen, setIsOpen] = useState(false);
-  const resolveRef = useRef<(returnValue: ReturnType) => any>();
+  const resolveRef = useRef<(returnValue: ReturnType) => any>(null);
```
jack-ryan-nava-pbc, line 27: I'm not super familiar with Return Types, is this saying that the type of the argument to the Ref is the same as the type it returns? In this case, "any"?

packages/design-system/src/components/Table/TableRow.tsx
```diff
@@ -22,7 +22,7 @@ export const TableRow = ({ children, _isTableHeadChild, ...tableRowProps }: Tabl
     return Children.map(children, (child: React.ReactElement) => {
       // Extend props before rendering.
       if (child && child.props) {
-        return cloneElement(child, {
+        return cloneElement(child as React.ReactElement<any>, {
```
jack-ryan-nava-pbc, line 25: Is this really type any? Not like `<TableRowProps>`?

jack-ryan-nava-pbc (approved):
I have a few non-blocking questions. Works well on my machine though!

packages/design-system/src/components/Table/TableRow.tsx
tamara-corbalt: Yeah good call — this could be typed more strictly. I think, in this case though, since TableRow's children are typically cells, it probably makes more sense to use `<TableCellProps>` here.

packages/design-system/src/components/Dialog/useDialog.ts
tamara-corbalt: Yeah, I’m not super familiar with this either. I think what's happening is that this ref will (at some point) hold a function that accepts a `ReturnType` param and returns `any` — so we're not making any constraints on what it returns.

packages/design-system/src/components/Table/TableRow.tsx
jack-ryan-nava-pbc: Actually, I think leave it. The type is very permissive, and looking at the HTML spec table rows can accommodate a number of different elements that might map on to our React component expectations.

jack-ryan-nava-pbc (approved):

PR #3794, opened 2025-10-02
[CMSDS-3467] Fix additional `JSX.Intrinsic` typing

tamara-corbalt:
## Summary

- Migrate additional `JSX.Intrinsic` references to `React.JSX.Intrinsic` as required by React 19 type definitions.
- These cases were missed in [pr-3782](https://github.com/CMSgov/design-system/pull/3782) because files compiled fine during the build process, but type-checking in CI flagged them once the React types package was bumped. Fixing these initializers is necessary to complete the version bump.

## How to test

1. `npm run build`
2. `npm run test:unit && npm run test:unit:wc`
3. `npm run type-check`

## Checklist

### DIFF
--- a/packages/docs/content/components/date-field/date-picker.mdx
+++ b/packages/docs/content/components/date-field/date-picker.mdx
@@ -2,24 +2,69 @@
 title: Calendar Picker
 intro: Date field (with calendar picker), also known as "date picker", provides single date entry with a text field or selection through a visual calendar element tied to the input.
 status:
-  level: use
+ level: use
 cmsgov:
-  figmaNodeId: 6-85
+ figmaNodeId: 6-85
 core:
-  figmaNodeId: 6-85
-  githubLink: design-system/src/components/DateField
+ figmaNodeId: 6-85
+ githubLink: design-system/src/components/DateField
 healthcare:
-  figmaNodeId: 6-85
+ figmaNodeId: 6-85
 medicare:
-  figmaNodeId: 6-85
+ figmaNodeId: 6-85
 ---
 
 ## Examples
 
 <StorybookExample
-  componentName="singleInputDateField"
-  storyId="components-singleinputdatefield--with-picker"
+ componentName="singleInputDateField"
+ storyId="components-singleinputdatefield--with-picker"
 />
+
+## Guidance
+
+### When to use
+
+- Use the date picker for dates that have happened within the last year, in the future, or is relevant to selecting the day of the week.
+
+### Guidelines
+
+Allow users to have flexibility in entering the date, such as allowing one-digit numbers with or without "0" prior to the number, or entering a date with or without slashes to separate month, day, and year.
+
+- Example: "01" for a month or day. Leading zeros for single digits are necessary if you do not include separating characters between the month, day and year values.
+- Example: "05/18/2022" as well as "05182022".
+
+### When to consider alternatives
+
+- Don't use if users are entering a memorable date. Instead, use either [single-](/components/single-input-date-field/) or [multi-input](/components/multi-input-date-field/) date fields.
+- Don't use if the date occurred in the distant past (like date of birth).
+
+### Accessibility
+
+#### Accessibility
+
+- Please see the [accessibility guidelines for all text inputs](/components/text-field/#guidance) for information specific to the input fields.
+- The calendar button uses `aria-describedby` to connect the required label text (supplied via the `label` prop) and the optional hint text (provided by the `hint` prop), to ensure screen readers announce contextual information when the open calendar button receives focus.
+- Users will be able to enter a variety of date formats that automatically format to the correct date format of MM/DD/YYYY. This change happens once the form field loses focus. For example, a user enters a date as "04122020" and it changes to "04/12/2020" once focus leaves the date input.
+- Focus automatically returns to the date input field after date selection.
+
+#### Accessibility Testing
+
+**Keyboard**:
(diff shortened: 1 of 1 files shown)


derek-cmsds (changes requested):
Running `npm run type-check` is currently giving me several errors similar to this:

```shell
packages/design-system/src/components/web-components/ds-accordion/ds-accordion-item.test.tsx:10:48 - error TS2345: Argument of type '"ds-accordion-item"' is not assignable to parameter of type 'keyof IntrinsicElements'.

10 const renderAccordionItem = createTestRenderer('ds-accordion-item', (attrs = {}) => (
```

packages/design-system/src/components/web-components/ds-alert/ds-alert.tsx
```diff
@@ -32,4 +32,14 @@ const Wrapper = ({ analytics, hideIcon, rootId, ...otherProps }: WrapperProps) =
   />
 );
 
+/* eslint-disable @typescript-eslint/no-namespace */
+declare global {
+  namespace React.JSX {
+    interface IntrinsicElements {
+      'ds-alert': any;
```
jack-ryan-nava-pbc: Can you add the `['div'] & {
        [K in (typeof attributes)[number]]?: string;`
incantation here?

tamara-corbalt:
> Running `npm run type-check` is currently giving me several errors similar to this:
> 
> ```shell
> packages/design-system/src/components/web-components/ds-accordion/ds-accordion-item.test.tsx:10:48 - error TS2345: Argument of type '"ds-accordion-item"' is not assignable to parameter of type 'keyof IntrinsicElements'.
> 
> 10 const renderAccordionItem = createTestRenderer('ds-accordion-item', (attrs = {}) => (
> ```

🧐 So strange, I'm wasn't seeing this locally when I opened the pr. Will keep investigating.

jack-ryan-nava-pbc:
In using CoPilot to better understand the typescript errors, it looks like it has to do with our type declarations for each web component. I'm looking into a solution now.

tamara-corbalt:
> In using CoPilot to better understand the typescript errors, it looks like it has to do with our type declarations for each web component. I'm looking into a solution now.

After I ran `npm run clean && npm install && npm run build` I saw the same errors that @derek-cmsds was seeing locally. 

Take a look at [this commit](https://github.com/CMSgov/design-system/pull/3794/commits/67b0ab9eab31ac5ed5d2c9e1fc152841f1b38cf5) — it shows how I handled the web component type declarations. Waiting to see if CI passes. Curious what you think: are there downsides to getting this specific with the types?

jack-ryan-nava-pbc:
> > In using CoPilot to better understand the typescript errors, it looks like it has to do with our type declarations for each web component. I'm looking into a solution now.
> 
> After I ran `npm run clean && npm install && npm run build` I saw the same errors that @derek-cmsds was seeing locally.
> 
> Take a look at [this commit](https://github.com/CMSgov/design-system/pull/3794/commits/67b0ab9eab31ac5ed5d2c9e1fc152841f1b38cf5) — it shows how I handled the web component type declarations. Waiting to see if CI passes. Curious what you think: are there downsides to getting this specific with the types?

The downsides that I see are possible type errors when/if we alter the React api. However, that's kind of the point of Typescipt. So I think it's ok. 

One thing I want to call out is that there are cases where we add the type for the component to the global namespace in both the component file and the test file. 

I was thinking (with the help of CoPilot) that it might make sense to pull all of the global type declarations out of the web component files and put them in a single global declaration in one type file: `web-component-types.d.ts`

We can then update our `tsconfig.json` file to include those declarations here: `"include": ["packages/**/src/**/*", ".storybook/**/*"]`

That would reduce the duplication of the type declarations in the global scope and we could delete the declarations in each component file and the tests. What do you think of that idea?

tamara-corbalt:
> One thing I want to call out is that there are cases where we add the type for the component to the global namespace in both the component file and the test file.
> 
> I was thinking (with the help of CoPilot) that it might make sense to pull all of the global type declarations out of the web component files and put them in a single global declaration in one type file: `web-component-types.d.ts`
> 
> We can then update our `tsconfig.json` file to include those declarations here: `"include": ["packages/**/src/**/*", ".storybook/**/*"]`
> 
> That would reduce the duplication of the type declarations in the global scope and we could delete the declarations in each component file and the tests. What do you think of that idea?

I like that idea. In the last commit I removed the test file declarations, so the component file is the single source of truth. But I see your point — if we consolidate all of the jsx global declarations into a central place, we’d reduce the scattering across components. I’m on board with consolidating and happy to take a pass at this.

jack-ryan-nava-pbc:
> > I like that idea. In the last commit I removed the test file declarations, so the component file is the single source of truth. But I see your point — if we consolidate all of the jsx global declarations into a central place, we’d reduce the scattering across components. I’m on board with consolidating and happy to take a pass at this.
> 
> The one caveat is that I'm about to create 30 new web components using a script and each one will need a JSX global declaration. It _might_ be easier to include that in the brand new component file instead of trying to append them to an existing global file. Either way, it might make sense to decide on this before I finish this script.

I'm pretty sure that Typescript will work with both solutions, meaning that they can coexist. Derek, you can generate in file, and then we can copy those declarations over. That's a little more manual work, but not complicated. What do you two think?

jack-ryan-nava-pbc (approved):

derek-cmsds (approved):
All good now.

PR #3795, opened 2025-10-02
[CMSDS-3575] Add a11y guidance for the Calendar Picker component

jack-ryan-nava-pbc:
## Summary

- Added a11y guidance for the calendar picker component
- [Ticket](https://jira.cms.gov/browse/CMSDS-3575)

## How to test

1. Run locally
2. Confirm that the text on /components/date-field/date-picker/ matches the text in the PDF in the ticket. Except for the small alteration on line 56:
Previous: Tab: Moves between the date input field and calendar button. When the calendar is initially opened, pressing tab will move focus first from a previous month selection button then to the month selection dropdown, next the year selection dropdown and finally **the initially selected date or the first of the month**
Revision: Tab: Moves between the date input field and calendar button. When the calendar is initially opened, pressing tab will move focus first from a previous month selection button then to the month selection dropdown, next the year selection dropdown and finally the **date entered in the text input or the first of the month if no date has been entered**.

## Checklist

### DIFF
--- a/packages/docs/content/components/date-field/date-picker.mdx
+++ b/packages/docs/content/components/date-field/date-picker.mdx
@@ -31,7 +31,7 @@ medicare:
 
 Allow users to have flexibility in entering the date, such as allowing one-digit numbers with or without "0" prior to the number, or entering a date with or without slashes to separate month, day, and year.
 
-- Example: "1" as well as "01" for a month or day.
+- Example: "01" for a month or day. Leading zeros for single digits are necessary if you do not include separating characters between the month, day and year values.
 - Example: "05/18/2022" as well as "05182022".
 
 ### When to consider alternatives
@@ -45,7 +45,7 @@ Allow users to have flexibility in entering the date, such as allowing one-digit
 
 - Please see the [accessibility guidelines for all text inputs](/components/text-field/#guidance) for information specific to the input fields.
 - The calendar button uses `aria-describedby` to connect the required label text (supplied via the `label` prop) and the optional hint text (provided by the `hint` prop), to ensure screen readers announce contextual information when the open calendar button receives focus.
-- Users will be able to enter a variety of date formats that automatically format to the correct date format of MM/DD/YYYY. This change happens once the form field loses focus. For example, a user enters a date as "4122020" and it changes to "04/01/2020" once focus leaves the date input.
+- Users will be able to enter a variety of date formats that automatically format to the correct date format of MM/DD/YYYY. This change happens once the form field loses focus. For example, a user enters a date as "04122020" and it changes to "04/12/2020" once focus leaves the date input.
 - Focus automatically returns to the date input field after date selection.
 
 #### Accessibility Testing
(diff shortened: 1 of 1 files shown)


packages/docs/content/components/date-field/date-picker.mdx
```diff
@@ -2,62 +2,83 @@
...
-### When to use
+### Accessibility
 
-- Use the date picker for dates that have happened within the last year, in the future, or is relevant to selecting the day of the week.
+#### Accessibility
 
-### When to consider alternatives
+- Please see the [accessibility guidelines for all text inputs](/components/text-field/#guidance) for information specific to the input fields.
+- The calendar button uses `aria-describedby` to connect the required label text (supplied via the `label` prop) and the optional hint text (provided by the `hint` prop), to ensure screen readers announce contextual information when the open calendar button receives focus.
+- Users will be able to enter a variety of date formats that automatically format to the correct date format of MM/DD/YYYY. This change happens once the form field loses focus. For example, a user enters a date as "4122020" and it changes to "04/01/2020" once focus leaves the date input.
```
derek-cmsds: > For example, a user enters a date as "4122020" and it changes to "04/01/2020" once focus leaves the date input.

Is that right? Wouldn't "4122020" correspond to "04/12/2020"?

derek-cmsds: This also does not function the way this copy says it does. You cannot leave out both a leading zero and the slashes. See the below screen recording:

https://github.com/user-attachments/assets/24a460c3-e968-42bc-9e77-0f83e81f2481

jack-ryan-nava-pbc: Good catch!

jack-ryan-nava-pbc: > > For example, a user enters a date as "4122020" and it changes to "04/01/2020" once focus leaves the date input.
> 
> Is that right? Wouldn't "4122020" correspond to "04/12/2020"?

Typo! Thank you

jack-ryan-nava-pbc: @derek-cmsds thanks for pointing this out. I'm updating the example in the sentence above but also editing the first example text under the Guidelines section so that it read:

- Example: "01" for a month or day. Leading zeros for single digits are necessary.


Does this seem like a reasonable change?

derek-cmsds: Well, you _can_ omit the leading zero _if_ you include slashes. "1/1/2025" is a valid input. So is "01012025". You just can't do "112025" because the component has no clue how to interpret that string.

jack-ryan-nava-pbc: Ok, how about this under "Guidelines"

- Example: "01" for a month or day. Leading zeros for single digits are necessary if you do not include separating characters between the month, day and year values.

derek-cmsds (approved):

PR #3801, opened 2025-10-06
[NO-TICKET] Ensure `Card` is exported from Core

tamara-corbalt:
## Summary
- Adds missing entry points for the new Core `Card` component.  
- Adds `ComponentCardExample` – tests that the Core `Card` component can be imported and used in an application.  

## How to test
1. `cd examples/astro-react-18` 
2. `npm run start`
3. Inspect Core Card component example.

## Checklist

### DIFF
--- a/examples/_shared/react/CardComponentExample.tsx
+++ b/examples/_shared/react/CardComponentExample.tsx
@@ -1,0 +1,17 @@
+import { Card } from '@cmsgov/design-system';
+
+export function CardComponentExample() {
+  return (
+    <>
+      <h2>Core Card Component Example</h2>
+      <Card className="ds-u-padding--2">
+        It was the best of times, it was the worst of times, it was the age of wisdom, it was the
+        age of foolishness, it was the epoch of belief, it was the epoch of incredulity, it was the
+        season of Light, it was the season of Darkness, it was the spring of hope, it was the winter
+        of despair...
+      </Card>
+    </>
+  );
+}
+
+export default CardComponentExample;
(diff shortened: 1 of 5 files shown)


packages/ds-medicare-gov/src/components/index.ts
```diff
@@ -12,6 +12,6 @@ export * from './MedicaregovHelpDrawer';
 export * from './MedicaregovLogo';
 export * from './MedicaregovThirdPartyExternalLink';
 export * from './SimpleFooter';
-export * from './Card';
+export { Card } from './Card';
```
tamara-corbalt, line 15: **Note:** Ran into a typeScript issue here:  

```
Module '@cmsgov/design-system' has already exported a member named 'Card'. 
Consider explicitly re-exporting to resolve the ambiguity.
```
So I updated this to an explicit export to ensure the medicare `Card` correctly overrides the Core `Card` at the typesScript level.

examples/_shared/react/CardComponentExample.tsx
```diff
@@ -0,0 +1,17 @@
+import { Card } from '@cmsgov/design-system';
+
+export function CardComponentExample() {
+  return (
+    <>
+      <h2>Core Card Component Example</h2>
+      <Card className="ds-u-padding--2">
```
jack-ryan-nava-pbc, line 7: Nice! Thanks for adding this example

packages/ds-medicare-gov/src/components/index.ts
jack-ryan-nava-pbc: Just putting this here for future me since I was confused:

In the medicare version of the Card we both have a default export and a named export for the Card component function. Here we are utilizing the named export to resolve ambiguity.

PR #3802, opened 2025-10-06
[CMSDS-3600] Remove Mgov `Card` component

tamara-corbalt:
## Summary

- Removes the Medicare-specific `Card` module while ensuring that the `m-c-card` CSS class remains available for use.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-3600)  

## How to test
- `npm run test:unit`
- I verified that the `m-c-card` class is still available for use by adding a temporary test component that renders a styled `<div>` using the `m-c-card` class.
- Updated `examples/astro-react-18/src/layouts/Layout.astro` to import the Medicare-specific styles instead of the core theme:
  ```ts
  import '@cmsgov/ds-medicare-gov/css/index.css';
  import '@cmsgov/ds-medicare-gov/css/medicare-theme.css';

- Ran the `astro-react-18` example site and confirmed in Chrome DevTools that the `m-c-card` class was applied to the test component.

## Checklist


examples/_shared/react/CssCardExample.tsx
```diff
@@ -0,0 +1,21 @@
+import { FunctionComponent } from 'react';
+
+interface CardProps {
+  className?: string;
+}
+
+export const CssCardExample: FunctionComponent<CardProps> = ({ className = 'ds-c-card' }) => {
```
jack-ryan-nava-pbc: Got it so this is confirming that the ds-c-card CSS class still exists?

jack-ryan-nava-pbc: Are we keeping the `m-c-card` style? I thought so.

jack-ryan-nava-pbc: Would this be better as a spec? I think the existence of this Example might get confusing after some time.

tamara-corbalt: @jack-ryan-nava-pbc, Yeah, I agree — this example isn’t very clear in its purpose. My intent was just to verify backward compatibility for teams that still use the `m-c-card` class.

I’m thinking I’ll remove this example and instead describe how I verified that `m-c-card` is still available in the testing steps. That way, reviewers can confirm it themselves if needed.

PR #3808, opened 2025-10-08
[CMSDS-3633] Convert icons to web components

derek-cmsds:
## Summary

- We have over 30 Icon components that we need to migrate into Web Components. Rather than manually creating our web component files, we determined that it made more sense to create a single script that generates new web component files from the imported React components. This PR contains both the generated files and the script that created them.

- [CMSDS-3633](https://jira.cms.gov/browse/CMSDS-3633)

### Test the web components
1. Clean install: `npm run clean && npm install && npm run build`
2. Remove the Angular cache: `rm -rf examples/angular/.angular`
3. Start the Angular example app: `cd examples/angular && npm run start`
4. Visit the Angular example app: http://localhost:4200/
5. Scroll down to the bottom of the page.
6. Verify it looks something like this:
Screenshot 2025-10-08 at 16 05 16

## Test the script
1. Run the script: `npm run convert:icons`
2. Verify no git diff as no files have been created or overwritten.
3. Remove all files from `packages/design-system/src/components/web-components/ds-icons` except `shared-attributes.ts` and the `templates` folder.
4. Run `npm run convert:icons` again.
5. Verify all files have returned. (There will be some differences compared to the committed files, but the number of files should be the same.)

## Checklist

### DIFF
--- a/packages/design-system/src/components/web-components/ds-icons/ds-add-icon.tsx
+++ b/packages/design-system/src/components/web-components/ds-icons/ds-add-icon.tsx
@@ -1,6 +1,7 @@
 import { define } from '../preactement/define';
 import { attributes } from './shared-attributes';
 import { AddIcon, IconCommonProps } from '../../Icons';
+import { parseBooleanAttr } from '../wrapperUtils';
 
 /* eslint-disable @typescript-eslint/no-namespace */
 declare global {
@@ -19,7 +20,7 @@ interface WrapperProps extends Omit<IconCommonProps, 'ariaHidden'> {
 }
 
 const Wrapper = ({ ariaHidden = 'true', ...otherProps }: WrapperProps) => (
-  <AddIcon ariaHidden={JSON.parse(ariaHidden)} {...otherProps} />
+  <AddIcon ariaHidden={parseBooleanAttr(ariaHidden)} {...otherProps} />
 );
 
 define('ds-add-icon', () => Wrapper, { attributes });
--- a/packages/design-system/src/components/web-components/ds-icons/templates/web-component-template.txt
+++ b/packages/design-system/src/components/web-components/ds-icons/templates/web-component-template.txt
@@ -1,6 +1,7 @@
 import { define } from '../preactement/define';
 import { attributes } from './shared-attributes';
 import { ICON_NAME, IconCommonProps } from '../../Icons';
+import { parseBooleanAttr } from '../wrapperUtils';
 
 /* eslint-disable @typescript-eslint/no-namespace */
 declare global {
@@ -20,7 +21,7 @@ interface WrapperProps extends Omit<IconCommonProps, 'ariaHidden'> {
 }
 
 const Wrapper = ({ ariaHidden = 'true', ...otherProps}: WrapperProps) => (
-  <ICON_NAME ariaHidden={JSON.parse(ariaHidden)} {...otherProps} />
+  <ICON_NAME ariaHidden={parseBooleanAttr(ariaHidden)} {...otherProps} />
 );
 
 define('WEB_COMPONENT_NAME', () => Wrapper, { attributes });
--- a/packages/design-system/src/components/web-components/ds-icons/ds-alert-circle-icon.tsx
+++ b/packages/design-system/src/components/web-components/ds-icons/ds-alert-circle-icon.tsx
@@ -1,6 +1,7 @@
 import { define } from '../preactement/define';
 import { attributes } from './shared-attributes';
 import { AlertCircleIcon, IconCommonProps } from '../../Icons';
+import { parseBooleanAttr } from '../wrapperUtils';
 
 /* eslint-disable @typescript-eslint/no-namespace */
 declare global {
@@ -19,7 +20,7 @@ interface WrapperProps extends Omit<IconCommonProps, 'ariaHidden'> {
 }
 
 const Wrapper = ({ ariaHidden = 'true', ...otherProps }: WrapperProps) => (
-  <AlertCircleIcon ariaHidden={JSON.parse(ariaHidden)} {...otherProps} />
+  <AlertCircleIcon ariaHidden={parseBooleanAttr(ariaHidden)} {...otherProps} />
 );
 
 define('ds-alert-circle-icon', () => Wrapper, { attributes });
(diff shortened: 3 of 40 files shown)


packages/design-system/src/components/utilities/convert-icons.test.ts
derek-cmsds, line 1: We may want to remove this file. I left it in for context and in case we wanted to update the script at a later time.

packages/design-system/src/components/utilities/convert-icons.ts
derek-cmsds, line 1: We don't _need_ to keep this file, but it might be useful in case we need to convert more icons later. It could also be generalized at some point to make the conversion of any React component into a web component easier.

package.json
```diff
@@ -25,6 +25,7 @@
     "clean:cache": "rm -rf packages/docs/.cache",
     "clean:builds": "find ./packages ./examples -name 'dist' -type d -depth 2 -exec rm -rf '{}' + && rm -rf ./packages/docs/public && rm -rf ./dist",
     "clean:modules": "find . -name 'node_modules' -type d -prune -exec rm -rf '{}' +",
+    "convert:icons": "tsx ./scripts/convert-icons.ts",
```
derek-cmsds, line 28: I left this command in for testing purposes. If we get rid of the script files, we need to make sure we delete this command.

scripts/convert-icons.ts
derek-cmsds, line 1: See above comment on the utilities file. We can easily remove the script, but it should also be safe to leave it. Running the script again _should_ not do any harm to the existing files.

packages/design-system/src/components/web-components/ds-icons/templates/web-component-template.txt
```diff
@@ -0,0 +1,26 @@
...
+}
+/* eslint-enable */
+
+interface WrapperProps extends Omit<IconCommonProps, 'ariaHidden'> {
+  ariaHidden?: string;
+}
+
+const Wrapper = ({ ariaHidden = 'true', ...otherProps}: WrapperProps) => (
+  <ICON_NAME ariaHidden={JSON.parse(ariaHidden)} {...otherProps} />
+);
```
derek-cmsds, line 25: `ariaHidden` had to be handled separately as it is treated like a string by the web component and like a boolean by the React component. There needs to be an explicit conversion during the hand-off in order for the `title` and `desc` elements to be rendered correctly.

examples/angular/src/app/app.component.html
derek-cmsds, line 1: I went ahead and added these examples to demonstrate the proper use of the web component attributes, since there may be a few things (like "class-name" and "view-box") that aren't intuitive.

packages/design-system/src/components/utilities/convert-icons.ts
```diff
@@ -0,0 +1,82 @@
+import { kebabCaseIt } from 'case-it/kebab';
+import { appendFile, open, readFile, writeFile } from 'node:fs/promises';
+import path from 'node:path';
+import * as Icons from '../Icons';
+
+const componentsPath = path.join(__dirname, '..');
+const webComponentsPath = path.join(componentsPath, 'web-components');
+const iconsPath = path.join(webComponentsPath, 'ds-icons');
+
+const icons = Object.keys(Icons);
```
jack-ryan-nava-pbc, line 10: This is cool. I did not know you could iterate over an import like this.

examples/angular/src/app/app.component.html
```diff
@@ -270,6 +271,19 @@ <h2 class="ds-text-heading--4xl ds-u-margin-top--2">Drawer</h2>
...
       Click to toggle drawer
     </app-button>
+
+    <h2 class="ds-text-heading--4xl ds-u-margin-top--2">Icons</h2>
+    <ds-star-icon is-filled="true" />
+    <ds-alert-icon aria-hidden="false" title="Testing a custom title" />
+    <ds-arrow-icon direction="up" />
+    <ds-check-icon class-name="ds-c-icon-color--error" />
+    <ds-pdf-icon class-name="ds-u-font-size--5xl" />
+    <ds-svg-icon title="Smiley face" view-box="0 0 496 512">
```
jack-ryan-nava-pbc, line 281: Nice. I like the demonstration of using a custom svg.

package.json
jack-ryan-nava-pbc: The harm of leaving this in for the moment seems minimal, besides potentially overwriting some content on a few of the components.

scripts/convert-icons.ts
jack-ryan-nava-pbc: Well, it will overwrite some of the content on a few of them right? But that's easy enough to discard. I think this is fine to leave for now.

packages/design-system/src/components/web-components/ds-icons/ds-add-icon.tsx
```diff
@@ -0,0 +1,25 @@
...
+  }
+}
+/* eslint-enable */
+
+interface WrapperProps extends Omit<IconCommonProps, 'ariaHidden'> {
+  ariaHidden?: string;
+}
+
+const Wrapper = ({ ariaHidden = 'true', ...otherProps }: WrapperProps) => (
+  <AddIcon ariaHidden={JSON.parse(ariaHidden)} {...otherProps} />
```
jack-ryan-nava-pbc: We have a helper utility already for parsing strings in a number of ways. I think you want to use this one: https://github.com/CMSgov/design-system/blob/4dd2eddd5cd75e27f35ea0b477a0f7955bbb9115/packages/design-system/src/components/web-components/wrapperUtils.ts#L1

scripts/convert-icons.ts
derek-cmsds: The script should not overwrite any files that currently exist. If you were to delete existing files and _then_ run the script, that would cause a diff.

packages/design-system/src/components/web-components/ds-icons/ds-add-icon.tsx
derek-cmsds: I really wish I knew that existed _before_ I ran the script. 😬

packages/design-system/src/components/web-components/ds-icons/templates/web-component-template.txt
```diff
@@ -0,0 +1,27 @@
...
+  }
+}
+/* eslint-enable */
+
+interface WrapperProps extends Omit<IconCommonProps, 'ariaHidden'> {
+  ariaHidden?: string;
+}
+
+const Wrapper = ({ ariaHidden = 'true', ...otherProps}: WrapperProps) => (
+  <ICON_NAME ariaHidden={parseBooleanAttr(ariaHidden)} {...otherProps} />
```
jack-ryan-nava-pbc, line 24: Perfecto!

PR #3811, opened 2025-10-15
[CMSDS-3634] Create Storybook story for Icon web components

derek-cmsds:
## Summary

- #3808 created a whole bunch of web components for our existing icons. This PR adds those to a Story so they are visible in our Storybook documentation.
- [CMSDS-3634](https://jira.cms.gov/browse/CMSDS-3634)

## How to test

1. Run storybook: `npm run storybook`
2. Scroll down to the "WEB COMPONENTS" section of the sidebar and click "ds-icons."
3. Verify a doc page for "ds-icons" loads with a table of icons. Inspect the icons and verify they are being rendered using web component tags.

## Checklist

- [x] Communicate the assigned milestone/release date with Design so they can communicate appropriately

### DIFF
--- a/tests/browser/snapshots/storybook-docs/Docs-Web-Components-ds-icons-Docs-prop-table-matches-snapshot-1.txt
+++ b/tests/browser/snapshots/storybook-docs/Docs-Web-Components-ds-icons-Docs-prop-table-matches-snapshot-1.txt
@@ -1,37 +1,37 @@
 [
   [
     "aria-hidden",
-    "",
+    "Describes the value of the aria-hidden attribute on the SVG. Defaulted to true with the assumption that most icons are decorative. If the icon does not have any associated label text, set this to false and ensure a title is provided for improved accessibility.",
     "\"true\"\"false\""
   ],
   [
     "class-name",
-    "",
+    "Additional CSS classes to be added to the svg element",
     "string"
   ],
   [
     "description",
-    "",
+    "Long-text description of any SVG. Use for complex icons, otherwise title prop will suffice.",
     "string"
   ],
   [
     "id",
-    "",
+    "A custom id attribute for the SVG",
     "string"
   ],
   [
     "inversed",
-    "",
+    "If true sets inverse fill color. Default is false.",
     "\"true\"\"false\""
   ],
   [
     "title",
-    "",
+    "The descriptive name for the SVG icon",
     "string"
   ],
   [
     "view-box",
-    "",
+    "A string describing the viewbox of the SVG. It is recommended that the icon is centered and fill up the default viewport size. See this blog post for further explanation on viewBox and how to use it.",
     "string"
   ]
 ]


packages/design-system/src/components/web-components/ds-icons/ds-icons.stories.tsx
```diff
@@ -0,0 +1,257 @@
+import { Meta, StoryObj } from '@storybook/react';
+import { webComponentDecorator } from '../storybook';
+import WebComponentDocTemplate from '../../../../../../.storybook/docs/WebComponentDocTemplate.mdx';
+import './index';
+
+const meta: Meta = {
+  title: 'Web Components/ds-icons',
+  decorators: [webComponentDecorator],
+  argTypes: {
+    'aria-hidden': { control: 'boolean' },
```
tamara-corbalt: I think we should maybe add `description` fields for each of these attributes so that our web component users get the same context that react users do.

tamara-corbalt (commented):
Looks great! Left one suggestion

packages/design-system/src/components/web-components/ds-icons/ds-icons.stories.tsx
```diff
@@ -0,0 +1,257 @@
...
+      <>
+        <ds-arrow-icon direction="up" />
+        <ds-arrow-icon direction="down" />
+        <ds-arrow-icon direction="left" />
+        <ds-arrow-icon direction="right" />
+      </>
+    ),
+    name: 'ds-arrow-icon',
+    notes:
+      'Component takes <code>direction</code> prop to determine if it is up, down, left or right.',
```
jack-ryan-nava-pbc, line 90: Would it be a good idea to put the directions in <code> blocks since these are the values that are acceptable as inputs to the direction prop?

packages/design-system/src/components/web-components/ds-icons/ds-icons.stories.tsx
jack-ryan-nava-pbc: Since none of these props are wired up to be reactive in the example, does it make sense to list these as controls or should we hide them?

packages/design-system/src/components/web-components/ds-icons/ds-icons.stories.tsx
```diff
@@ -0,0 +1,257 @@
...
+  {
+    defaultTitle: 'Star / Star Filled',
+    component: (
+      <>
+        <ds-star-icon />
+        <ds-star-icon is-filled="true" />
+      </>
+    ),
+    name: 'ds-star-icon',
+    notes: 'Component takes <code>isFilled</code> prop to determine star is filled or an outline.',
```
jack-ryan-nava-pbc, line 236: Same thing as the arrows above, should we denote the acceptable values to the `isFilled` prop via <code> blocks?

packages/design-system/src/components/web-components/ds-icons/ds-icons.stories.tsx
derek-cmsds: A `description` has been added to each field.

packages/design-system/src/components/web-components/ds-icons/ds-icons.stories.tsx
derek-cmsds: In order to maintain consistency with the React component documentation, this will remain as formatted.

packages/design-system/src/components/web-components/ds-icons/ds-icons.stories.tsx
derek-cmsds: The way the `WebComponentDocTemplate` page is set up, it uses the value of the `control` key to populate the "Accepted values" column. Setting `control` to `false` leaves that column empty. Which path makes more sense?

packages/design-system/src/components/web-components/ds-icons/ds-icons.stories.tsx
derek-cmsds: As with the `direction` prop, after discussing this in Slack I'm keeping the formatting as-is to maintain consistency with the React component's documentation.

PR #3863, opened 2025-12-18
[CMSDS-3417] Wire up `StatusIndicator` Badges & Theme-specific Alerts

tamara-corbalt:
## Summary

- Wires up `StatusIndicator` badges and Alerts. Some Alerts explain why a component has a **caution** or **avoid** status; others flag product-specific components (e.g., Medicare Footer) and prompt users to switch to the appropriate theme on the doc site.
- Updates select status messages to align with approved copy.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-3417)

> **Note**  
> This is the first of two PRs that will wire up all visual changes related to component status indication on the doc site.

## How to test

- Run `npm run start`.
- Verify that `StatusIndicator` Badges appear on all component pages.
- For components with **Caution** or **Avoid** status (e.g., Inset or Stars), verify that an Alert is displayed explaining the status.
- Verify that product-specific Alerts render correctly on the following pages (note: not all pages include these, only product-specific components):
  - [Healthcare Footer](http://localhost:8000/components/footer/healthcare-footer/?theme=core)
  - [Medicare Footer](http://localhost:8000/components/footer/medicare-footer/?theme=core)
  - [Healthcare Header](http://localhost:8000/components/header/healthcare-header/?theme=core)
  - [Medicare Header](http://localhost:8000/components/header/medicare-header/?theme=core)
  - [Inset (Healthcare)](http://localhost:8000/components/inset/?theme=core)
  - [Stars (Medicare)](http://localhost:8000/components/stars/?theme=core)
- Toggle between themes for each page. When the correct theme is selected for a product-specific component, the Alert should no longer appear.

## Checklist

### DIFF
--- a/packages/docs/src/components/layout/PageHeader.tsx
+++ b/packages/docs/src/components/layout/PageHeader.tsx
@@ -1,5 +1,8 @@
 import { FrontmatterInterface } from '../../helpers/graphQLTypes';
 import { linkAnalytics } from '../../helpers/analytics';
+import StatusIndicator from './StatusIndicator';
+import ThemeContent from '../content/ThemeContent';
+import { Alert } from '@cmsgov/design-system';
 import { withPrefix } from 'gatsby';
 import { makeFigmaUrl, makeGithubUrl, makeStorybookUrl } from '../../helpers/urlUtils';
 import GithubIcon from '../icons/GithubIcon';
@@ -16,7 +19,11 @@ type PageHeaderProps = {
  */
 const PageHeader = ({ frontmatter = { title: '' }, theme }: PageHeaderProps) => {
   const [themeLinks, setThemeLinks] = useState(undefined);
-  const { title, core, intro } = frontmatter;
+  const { title, core, intro, status } = frontmatter;
+  const level = status?.level;
+  const note = status?.note;
+  const targetTheme = status?.targetTheme;
+  const targetThemeNote = status?.targetThemeNote;
 
   const figmaNodeId = themeLinks?.figmaNodeId || core?.figmaNodeId || null;
   const figmaTheme = themeLinks?.figmaNodeId ? theme : 'core';
@@ -40,7 +47,17 @@ const PageHeader = ({ frontmatter = { title: '' }, theme }: PageHeaderProps) => 
 
   return (
     <header className={headerClassNames}>
-      <h1 className="ds-text-heading--4xl">{title}</h1>
+      <div className="ds-u-display--flex ds-u-align-items--baseline ds-u-flex-direction--row">
+        <h1 className="ds-text-heading--4xl">{title}</h1>
+        {status?.level && (
+          <div
+            className="ds-u-margin-left--2"
+            style={level === 'use' ? { position: 'relative', top: '-2px' } : undefined}
+          >
+            <StatusIndicator level={level} />
+          </div>
+        )}
+      </div>
       {intro && (
         <p className="ds-u-font-size--lg ds-u-measure--base ds-u-margin-top--1 ds-u-margin-bottom--1">
           {intro}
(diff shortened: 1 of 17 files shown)


packages/docs/src/components/layout/PageHeader.tsx
```diff
@@ -80,6 +97,22 @@ const PageHeader = ({ frontmatter = { title: '' }, theme }: PageHeaderProps) =>
           )}
         </div>
       )}
+      {note && level && (
```
jack-ryan-nava-pbc, line 100: Very good decision moving this into frontmatter and then escalating this to the template file!

jack-ryan-nava-pbc (commented):
Overall this looks great! Two things that may not be immediately relevant and one that is:
1. For the Inset component, our maturity checklist states that keyboard testing is "Not applicable" which I agree with. I'm not sure our note about keyboard testing really makes sense?
2. Aligning the Use Badge along the bottom of the H1 text just, doesn't look quite right to me? Did we decide against putting the badge under the H1? I'm happy to not worry about this if we already considered an alternative design and opted for this.
3. Do we want to override the screen reader text on the Badge component? Currently it announces the Variation (Info, Success, Warn, Alert) which may not be totally relevant for this information. Right now VoiceOver says "Success: Use" when the badge is selected. Is this what we want to happen?

tamara-corbalt:
> Overall this looks great! Two things that may not be immediately relevant and one that is:
> 
> 1. For the Inset component, our maturity checklist states that keyboard testing is "Not applicable" which I agree with. I'm not sure our note about keyboard testing really makes sense?
> 2. Aligning the Use Badge along the bottom of the H1 text just, doesn't look quite right to me? Did we decide against putting the badge under the H1? I'm happy to not worry about this if we already considered an alternative design and opted for this.
> 3. Do we want to override the screen reader text on the Badge component? Currently it announces the Variation (Info, Success, Warn, Alert) which may not be totally relevant for this information. Right now VoiceOver says "Success: Use" when the badge is selected. Is this what we want to happen?

Thanks for the feedback, @jack-ryan-nava-pbc!

1. Agreed! Happy to remove this note for now. The language for the status messages came from an [older audit](https://docs.google.com/spreadsheets/d/1_dZtf5PffrxfxBE03avpDt7eeVQifNbV_uy_1cWJzSM/edit?gid=520062201#gid=520062201) that identified gaps in our guidance. That was quite a while ago, so it’s not surprising that some of this reasoning is stale now. Would it make sense for me add a follow up with a ticket to draft a clearer a status note that reflects our current thinking about why `Inset` should be avoided?

2. I think I’ll defer to @sophia-cmsds here. The last discussion I remember about badge placement was in [this thread](https://github.com/CMSgov/design-system/pull/3595#issuecomment-2891972122). I implemented this based on the original Figma design, but I’m happy to update it if we land on a different placement.

3. Fair callout! Announcing the badge variation along with the badge text may add more noise than useful context for screen reader users. That said, this appears to be Badge’s current behavior and matches our [current guidance](https://design.cms.gov/components/badge/#accessibility-testing). If we decide we want to silence the variation announcement for this scenario, we’ll likely need to adjust the `Badge` API to make the behavior override-able.

tamara-corbalt:
> > > Re #3: we do actually already have the adjustments made to the Badge API. Clearly they need to be documented better. You can set hideScreenReaderText to hide all the screen reader specific text in case the text content is sufficient, maybe the way to go here. Or you can override the typical content by supplying your own via the screenReaderText prop.
> > 
> > 
> > DOH! You're right, I forgot we added these props 🤦🏼‍♀️. For this use case, I’m not sure that setting `hideScreenReaderText` to `true` is the right choice. If we only allow the variation to be read out, a non-sighted user would lose the badge label itself, which is the core message we’re trying to send (whether the component is recommended or not). So I'm kind of leaning toward setting the `screenReaderText` to the badge text (e.g. “Don't Use”), but I’m still persuadable.
> 
> I like your suggestion here of setting the screenReaderText value, however I want to point out that a non-sighted user should still have access to the text node of the Badge even if the screen reader text is hidden. The idea here is that sometimes the badge content is sufficient for screen readers and we don't need the additional text.

Ok, that makes sense to me, will push up the change :)

packages/docs/src/components/layout/StatusIndicator.tsx
```diff
@@ -41,7 +41,11 @@ export const StatusIndicator: React.FC<StatusIndicatorProps> = ({ level }) => {
   const { variation, label, Icon, iconClassName } = statusConfig[level];
 
   return (
-    <Badge variation={variation} className="ds-u-display--inline-flex ds-u-align-items--center">
+    <Badge
+      variation={variation}
+      hideScreenReaderText
```
jack-ryan-nava-pbc, line 46: Nice!

jack-ryan-nava-pbc (approved):
Approving! Regarding "Would it make sense for me add a follow up with a ticket to draft a clearer a status note that reflects our current thinking about why Inset should be avoided?" Yes, please! Make sure to note that we want folks to use the [NoteBox](https://design.cms.gov/components/note-box/?theme=core) instead

PR #3927, opened 2026-02-20
[NO-TICKET] Silence missing key warnings during build

jack-ryan-nava-pbc:
## Summary

- I'm sick of seeing an error at build time for a missing key on an element. 
- So I added a key that I should've added a while ago.

## How to test

1.`npm run start`
2. You shouldn't see any missing key warnings during build

## Checklist


derek-cmsds (approved):
Be gone, warnings!

PR #3985, opened 2026-03-25
[CMSDS-3954] Add hosted Markdown equivalents for documentation pages

tamara-corbalt:
## Summary

- Adds hosted Markdown equivalents for each documentation page.
- Converts `.mdx` content into pure Markdown output during build.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-3954)

### Run docs tests
`npm run test:unit:docs`

### Generate Hosted Markdown
1. `cd packages/docs`
2. `npm run build`
3. Inspect the `public` directory:
   - Each page directory should now include a `llms.txt` file alongside `index.html`.
4. Run `npm run start` (from within `packages/docs`):
   - Note: running `npm run start` from the repo root will run `gatsby clean && gatsby develop`, which clears the `public` directory and removes generated files.
5. Visit a page and its Markdown equivalent:
   - Example:
     - HTML: http://localhost:8000/components/accordion/?theme=core  
     - Markdown: http://localhost:8000/components/accordion/llms.txt
  

## Checklist


packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -2,8 +2,41 @@ import { normalizeSiteUrl } from './slug.mjs';
...
 
-export function buildLlmsTxt({ siteUrl, description, pages}) {
+export const buildMarkdownPage = ({ title, intro, body }) => {
+  return [
+    title ? `# ${title}` : '',
+    intro || '',
+    body || '',
+  ]
+    .filter(Boolean)
+    .join('\n\n');
```
jack-ryan-nava-pbc, line 20: If title, intro and body are all falsy, they'll be converted to empty strings, which then get filtered out and leaves an empty array, []. The join call then doesn't do anything, so you return empty string. Is that expected? Or is the idea to always return double new line?

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -0,0 +1,118 @@
...
+  let result = input;
+
+  for (const component of NON_CONTENT_COMPONENTS) {
+    result = result.replace(
+      new RegExp(`<${component}(\\s[^>]*)?>[\\s\\S]*?<\\/${component}>`, 'g'),
+      ''
+    );
+
+    result = result.replace(
+      new RegExp(`<${component}(\\s[^>]*)?\\s*\\/\\s*>`, 'g'),
```
jack-ryan-nava-pbc: Looks like this is duplicative with the self closing tags function? Why is this here?

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -2,8 +2,41 @@ import { normalizeSiteUrl } from './slug.mjs';
...
+};
+
+export const processMdxForHostedMarkdown = (body) => {
+  let result = body;
+
+  result = removeImportStatements(result);
```
jack-ryan-nava-pbc: I'd really love to see some simple tests for each of these functions.

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -2,8 +2,41 @@ import { normalizeSiteUrl } from './slug.mjs';
...
+  result = removeJsxComments(result);
+  result = stripComponentsWithChildren(result);
+  result = stripSelfClosingComponents(result);
+  result = convertThemeContentToAnnotations(result);
+  result = unwrapSimpleComponents(result);
+  result = normalizeMarkdownOutput(result);
```
derek-cmsds: Javascript needs a pipe operator so badly. I'm pretty sure there's a less verbose way to do this using `reduce`, but I like the legibility this maintains.

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -0,0 +1,118 @@
...
+  // Convert ThemeContent opening tags into readable annotations.
+  result = result.replace(
+    /<ThemeContent\s+onlyThemes=\{\[(.*?)\]\}\s*>/g,
+    (_, themes) => {
+      const themeList = themes
+        .split(',')
+        .map((t) => t.replace(/['"]/g, '').trim())
+        .filter(Boolean)
+        .join(', ');
+      return themeList ? `\n\n_Theme: ${themeList} only_\n\n` : '\n\n';
```
jack-ryan-nava-pbc: This is annoying of me, but I worry about the casing of "Theme" here. Since we're constructing the output of these functions for consumption by a non-human natural language processor, I think we need to be careful in probably pedantic ways. If themeList < 1, it should probably read "Themes:..." else it should say "Theme"

Am I being crazy?

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -0,0 +1,118 @@
...
+
+  result = result.replace(
+    /<ThemeContent\s+neverThemes=\{\[(.*?)\]\}\s*>/g,
+    (_, themes) => {
+      const themeList = themes
+        .split(',')
+        .map((t) => t.replace(/['"]/g, '').trim())
+        .filter(Boolean)
+        .join(', ');
+      return themeList ? `\n\n_Not for theme: ${themeList}_\n\n` : '\n\n';
```
jack-ryan-nava-pbc: Same thing here. "Not for themes: " if themeList > 1, else "Not for theme: "

packages/docs/scripts/llms-txt/index.mjs
jack-ryan-nava-pbc: Also, does the ordering of these function calls matter? Probably not, right?

derek-cmsds: You can also go the other direction and OOP this thing so it can be a chain of object methods. (ex: `result = mdxContent.removeImportStatements().removeJsxComments().stripComponentsWithChildren()....`) Again, I don't think we _should_, but you _could_.

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -0,0 +1,118 @@
...
+    result = result.replace(
+      new RegExp(`<${component}(\\s[^>]*)?\\s*\\/\\s*>`, 'g'),
+      ''
+    );
+  }
+
+  return result;
+}
+
+export function convertThemeContentToAnnotations(input) {
```
derek-cmsds: This might actually leave some confusing content in the LLMs.txt, as we appear to be stripping out the examples that the theme content is annotating.

jack-ryan-nava-pbc (commented):
Mostly minor stuff. I would like to see some specs added, and I'm not sure if we need error handling in place given that most of these functions are find and replace operations, but maybe worth thinking about...

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
derek-cmsds: Example from the Accordion page:
```markdown
# Accordion

An accordion is a list of headers that hide or reveal additional content when selected.

## Examples

_Not for theme: healthcare_

  

_Theme: healthcare only_
```

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -0,0 +1,118 @@
+const UNWRAP_TAGS = ['Alert', 'Badge'];
+
+const NON_CONTENT_COMPONENTS = [
```
derek-cmsds: This wipes out almost all of the content in the "Code" sections of the guidance. (See Accordion example below) Should we just remove that section entirely?

```markdown
## Code

### React

### Web Component

### Style customization

The following CSS variables can be overridden to customize Accordion components:

## Component maturity
```

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -0,0 +1,118 @@
...
+export function normalizeMarkdownOutput(input) {
+  return input
+    // Convert <br /> tags to blank lines.
+    .replace(/<br\s*\/?>/g, '\n\n')
+    // Remove JSX space expressions like {' '} or {" "}
+    .replace(/\{['"]\s*['"]\}/g, ' ')
+    // Remove any remaining self-closing PascalCase JSX tags.
+    .replace(/<[A-Z][A-Za-z0-9]*(\s[^>]*)?\s*\/>/g, '')
+    // Normalize excessive blank lines.
+    .replace(/\n{3,}/g, '\n\n');
```
derek-cmsds, line 137: I appreciate the code comments here.

packages/docs/scripts/llms-txt/renderMarkdown.mjs
```diff
@@ -38,7 +38,7 @@ function renderPagesList(pages, baseUrl) {
   let markdown = '';
 
   for (const { title, slug, intro } of sorted) {
-    markdown += `- [${title}](${baseUrl}${slug})${intro ? `: ${intro}` : ''}\n`;
+    markdown += `- [${title}](${baseUrl}${slug}/llms.txt)${intro ? `: ${intro}` : ''}\n`;
```
derek-cmsds, line 41: Is it best practice for these to be labeled `llms.txt` instead of something like `accordion.md`?

derek-cmsds (commented):
I left some questions about some of the Markdown generation. I have one other more general question though: Is there anything in the root `LLMs.txt` that references these other Markdown files? How do we prompt the chatbot to visit these other paths?

packages/docs/scripts/llms-txt/index.mjs
jack-ryan-nava-pbc: Even better make them all async and chain them using .then()!

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
jack-ryan-nava-pbc: Yeah I was just thinking this. On actually looking at the output I wonder if we couldn't simplify this by removing the Examples, Code, and Component Maturity sections in their entirety?

The theme information also doesn't really help, since it then doesn't say what is for that theme.

packages/docs/scripts/llms-txt/renderMarkdown.mjs
jack-ryan-nava-pbc: That's a good question. Mastra appears to label everything `llms.txt`: https://mastra.ai/docs/getting-started/studio/llms.txt

tamara-corbalt: Yeah I was emulating Mastra here - but the original [LLMs.txt proposal](https://llmstxt.org/#:~:text=We%20furthermore%20propose%20that%20pages%20on%20websites%20that%20have%20information%20that%20might%20be%20useful%20for%20LLMs%20to%20read%20provide%20a%20clean%20markdown%20version%20of%20those%20pages%20at%20the%20same%20URL%20as%20the%20original%20page%2C%20but%20with%20.md%20appended.%20(URLs%20without%20file%20names%20should%20append%20index.html.md%20instead.)) actually recommends providing `.md` equivalents for each HTML page. That said, I’m not sure the file extension itself matters as much as ensuring the content is pure Markdown.

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
tamara-corbalt: I think removing these sections entirely (for now) makes the most sense to me.

jack-ryan-nava-pbc: I think it would be good to think about how we can support the component maturity checklist later on. Maybe we can convert it to a table?

jack-ryan-nava-pbc: And I think for the Theme stuff we need to be more sophisticated about how we handle that. Like reformat the tag and keep child text, but if it's a Storybook link maybe include that link? IDK, I'm spitballing here.

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
tamara-corbalt: Yeah, I think a pretty clean approach would be to strip out the following sections:
```
## Examples
## Code
## Component maturity
```

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
tamara-corbalt: Sadly there are a lot of pages that use `ThemeContent` components that aren't neatly nested within `Examples`, `Code` or `Component Maturity`. Example: [headings.mdx](https://github.com/CMSgov/design-system/blob/main/packages/docs/content/foundation/typography/headings.mdx?plain=1#L31-L43)

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
tamara-corbalt: I'm going to removing this entire function in favor of stripping out whole heading sections.

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
tamara-corbalt: @jack-ryan-nava-pbc - I like the idea of preserving the child text and grouping it more clearly under a theme block. I was experimenting with something like:
```
**Theme: healthcare only**

Healthcare.gov uses Open Sans for all headings.

**Theme: medicare only**

Medicare.gov uses Montserrat for all headings.

```

This feels a bit clearer to me.

For the case where a StorybookExample might be nested inside ThemeContent: from what I’ve seen so far, most (if not all) StorybookExample instances live within ## Code sections, which we’re now stripping entirely. So that might simplify things.

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
tamara-corbalt: Not crazy at all!

packages/docs/scripts/llms-txt/index.mjs
tamara-corbalt: @jack-ryan-nava-pbc - Good question, I hadn’t considered this edge case since in practice all of our content pages really ought to have at least a title and body. Returning an empty string here seems like the right call for the no-content case.

packages/docs/scripts/llms-txt/index.mjs
tamara-corbalt: @jack-ryan-nava-pbc - I think the order actually does matter here. For example, we want to handle `ThemeContent` and preserve the content of Badges and Alerts before `normalizeMarkdownOutput` strips out any remaining JSX. I don’t love the brittleness of that, but I think that’s actually a good reason to keep the processing steps explicit rather than introducing async/.then() chaining. 

I think for now, I'm going to add comments to make this as clear as possible.

jack-ryan-nava-pbc: Ah gotcha. Yeah do not convert theses to async functions, I was being silly. 

Adding comments is a great idea! Thank you

packages/docs/scripts/llms-txt/index.mjs
tamara-corbalt: @jack-ryan-nava-pbc - agree, and I started the process of adding some tests for these functions. But I ran into some friction with jest not playing nice with the .mjs modules. Would you be okay with using node’s built-in test runner for now and revisiting the jest configuration later?

jack-ryan-nava-pbc: Yes I think so. We currently have no tests for anything in the /docs package, so maybe you can create a script local to that package that uses node.

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -2,8 +2,63 @@ import { normalizeSiteUrl } from './slug.mjs';
...
+    })
+    .join('');
+}
+
+export const processMdxForHostedMarkdown = (body) => {
+  let result = body;
+
+  // Run this first on the whole document, since ThemeContent
+  // may wrap across fenced code blocks.
+  result = normalizeThemeContent(result);
```
tamara-corbalt, line 41: With the new code-snippet-aware processing (above in `processMdxOutsideCodeFences`), the MDX body is split into fenced and non-fenced chunks so cleanup transforms only run on the non-fenced chunks.

Because `ThemeContent` can wrap both text and fenced code blocks, it must be normalized _before_ this split. Otherwise the opening and closing tags would end up in separate chunks.

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -2,8 +2,63 @@ import { normalizeSiteUrl } from './slug.mjs';
...
+  return [
+    title ? `# ${title}` : '',
+    intro || '',
+    body || '',
+  ]
+    .filter(Boolean)
+    .join('\n\n');
+};
+
+export function processMdxOutsideCodeFences(input, transform) {
```
tamara-corbalt, line 23: We want to preserve non-interactive code snippets in the hosted markdown output. Otherwise, large portions of code-heavy pages like `for-developers` become almost meaningless.

The limitation here is that we don't yet have a reliable way to preserve interactive code examples such as `<ThemeCodeBlock/>`, whose rendered output changes based on the theme switcher, like this [one](https://design.cms.gov/getting-started/for-developers/#example-loading-css-via-cdn).

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -2,8 +2,63 @@ import { normalizeSiteUrl } from './slug.mjs';
...
+  const regex = /(```(?:js|css|html)[\w-]*\n[\s\S]*?```)/g;;
```
jack-ryan-nava-pbc: Nit: double semi colon

jack-ryan-nava-pbc: To be annoying, but does our syntax highlighter plugin accept other languages like TS, or TSX or JSX?

jack-ryan-nava-pbc: This is a future thing if we ever indeed start supporting those syntaxes.

packages/docs/scripts/llms-txt/mdxFixtures.mjs
derek-cmsds, line 1: I love all of this.

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -2,8 +2,63 @@ import { normalizeSiteUrl } from './slug.mjs';
...
+export const processMdxForHostedMarkdown = (body) => {
+  let result = body;
+
+  // Run this first on the whole document, since ThemeContent
+  // may wrap across fenced code blocks.
+  result = normalizeThemeContent(result);
+
+  result = processMdxOutsideCodeFences(result, (input) => {
+    let chunk = input;
+    // ⚠️ NOTE: Order matters — do not rearrange these steps.
```
jack-ryan-nava-pbc, line 45: Very helpful comment!

packages/docs/gatsby-node.mjs
```diff
@@ -5,11 +5,12 @@ import path from 'path';
 import { fileURLToPath } from "url";
 import redirects from "./redirects.json" with { type: "json" };
 import fs from 'node:fs';
-import { buildLlmsTxt } from './scripts/llms-txt/index.mjs';
+import { buildRootLlmsTxt, processMdxForHostedMarkdown, buildMarkdownPage } from './scripts/llms-txt/index.mjs';
 // @ts-check
 
 const __dirname = path.dirname(fileURLToPath(import.meta.url))
 const require = createRequire(import.meta.url)
+const normalizePagePath = (slug) => slug.replace(/\d+_/g, '');
```
jack-ryan-nava-pbc, line 18: This function is actually duplicative of `removePositioning`, which is in the `packages/docs/src/helpers/casingUtils.ts` file. You should probably import and use that function instead of duplicating.

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -2,8 +2,63 @@ import { normalizeSiteUrl } from './slug.mjs';
...
+    // Earlier transforms operate on specific JSX structures,
+    // while later steps clean up and normalize the final Markdown output.
+    chunk = removeImportStatements(chunk);
+    chunk = stripMarkdownSections(chunk);
+    // Unwraps simple components (e.g., Alerts, Badges) but keeps their text 
+    chunk = unwrapSimpleComponents(chunk);
+    // Final cleanup: removes any remaining JSX.
```
derek-cmsds, line 52: I'm glad these comments are here. Would it make sense to document the functions in `mdxToMArkdown.mjs` with JSDoc comments so that devs can mouse-over them for more details?

packages/docs/scripts/llms-txt/mdxToMarkdown.test.mjs
derek-cmsds, line 1: I appreciate these tests.

package.json
```diff
@@ -55,6 +55,7 @@
     "test:unit:coverage:wc": "npm run test:unit:wc -- --collectCoverage --coverageReporters='json-summary'",
     "test:unit:coverage:all": "npm run test:unit:coverage:react && npm run test:unit:coverage:wc && tsx ./tests/unit/combineTestCoverage.ts",
     "test:unit": "jest --config=tests/unit/jest.config.js",
+    "test:unit:docs": "npm --prefix ./packages/docs run test",
```
derek-cmsds, line 58: Should the docs tests be run when running a broader test command like `npm run test` or `npm run test:unit`? Are we adding these tests to our CI pipeline?

derek-cmsds (approved):
No blockers, but I do have some questions about comments and the new doc site tests you added. Well done.

jack-ryan-nava-pbc (approved):
A few questions and observations, but over all this looks awesome! Well done and thank you!

packages/docs/scripts/llms-txt/index.mjs
tamara-corbalt: Looks like our [syntax highlighter](https://github.com/CMSgov/design-system/blob/main/packages/docs/src/helpers/syntaxHighlighting.ts#L6-L22) only supports `JSX` & `html`. Not seeing anything for `TS` or `TSX`

packages/docs/gatsby-node.mjs
tamara-corbalt: Agree this is duplicative but Node doesn’t love that `removePositioning` is being imported from a `.ts` file inside `gatsby-node.mjs`.

package.json
tamara-corbalt: Yep, I think that's a good idea to fold the new tests into the `npm run test:unit` script

packages/docs/gatsby-node.mjs
jack-ryan-nava-pbc: Does it throw an error or a type warning? I can live with a type warning (or even silence it).

tamara-corbalt: No this looks like a hard runtime failure; the build fails because node can't resolve the `.ts` file extension

jack-ryan-nava-pbc: Alrighty, let's duplicate the function for now. Thanks for checking! Can you add a brief `TODO` comment pointing to the other function?

package.json
tamara-corbalt: I may need to back out folding the docs tests into npm run test:unit and move that to a follow-up ticket. I forgot there are some downstream side effects: `test:unit:coverage` calls `test:unit`, and `combineTestCoverage.ts` expects a dedicated coverage artifact that no longer appears to be generated in CI.

tamara-corbalt: Added this follow up ticket: https://jira.cms.gov/browse/CMSDS-3973

PR #3997, opened 2026-03-30
[NO-TICKET] Convert Mask to functional component

jack-ryan-nava-pbc:
## Summary

- Refactored the Mask.tsx file to be a functional component rather than a Class based component
- Undertaken with the assistance of Copilot 
- This seemed like a solid starting point for investigating how to work with AI. 

**I'm looking for feedback on the following:**
- How to helpfully present and contextualize changes
- How to make sure that the changes done with AI are easy to review
- Should the PLAN.md file exist within our code base, or not?
- Please raise other concerns or criticisms around this

## How to test

1. All tests for the Mask component and TextField component should pass
2. All tests for the Textfield web component should pass
3. Check for code style that does not match best practices

## Checklist

- [X] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [X] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [X] Selected appropriate `Impacts`, multiple can be selected.
- [X] Selected appropriate release milestone

PR #3998, opened 2026-03-30
[NO-TICKET] CoPilot Instructions

jack-ryan-nava-pbc:
## Summary

- Alright, so if we're gonna use CoPilot we should try to do so safely. This is my initial attempt at fleshing out some instructions for the agent to follow during paired programming sessions. 
- These instructions came about during the development work of #3997 
- These are meant to be living, evolving instructions. They might be very restrictive currently, or maybe not.
- Recommended changes in tone and content are appreciated

## How to test

1. Find something small that you want to fix, a linter warning for example. See if you can have CoPilot fix that small thing using this instructions file. 

## Checklist


.github/copilot-instructions.md
```diff
@@ -0,0 +1,69 @@
...
+2. Refine the plan in a loop with the user. Prompt the user for feedback after the initial draft and do not proceed with the plan until the user explicitly states you may.
+3. After completing the work for a checkbox, meaning that a meaningful change to the code base has occurred. You must run the following commands:
+   1. npm run build
+   2. npm run test
+   3. npm run test:browser:smoke
+   4. npm run type-check
+4. Surface errors to the human you are working with. Do not propose fixes or attempt to fix at this time. If the human tells you to begin working on a test failure, you are allowed a maximum of 3 (three) iterations over that failure before you must return control to the human in charge. Be sure to summarize what you tried and what the outcomes were for each attempt.
+5. When a checkbox has been checked, meaning the work is done and all necessary npm scripts from step 3 have completed without issue, the AI must stop and prompt the human to perform the commit manually. The AI should provide the exact suggested commit command(s), a proposed commit message, and a summary of files expected to be committed. The AI must not attempt to run git commit itself.
```
tamara-corbalt, line 68: Since we’re being prescriptive that the human performs the commit manually (which is sensible!), it seems possible they could accidentally include `plan.md` in that commit. Are we sure this artifact needs to be created in the first place?
Could the plan instead live entirely in the interaction between the user and the AI, rather than being written to disk?

.github/copilot-instructions.md
```diff
@@ -0,0 +1,69 @@
...
+6. Upon completion of all steps in plan, create a pull request that follows the template in the .github folder. Include information from the plan.md file and make sure to ask the human in charge for the Milestone, and appropriate labels. Offer the changes for review to the human and do not proceed with creation of the pull request until the human has explicitly signed off. Do not commit the plan.md, but replicate the checkboxes from the plan.md file within the pull request template and note any deviations from those checkboxes.
```
tamara-corbalt: Nit: "... all steps in _the_ plan ..."

tamara-corbalt (commented):
Looks good! Left a question about necessity of creating the `plan.md` file and one teeny nit.

.github/copilot-instructions.md
jack-ryan-nava-pbc: So that's not a bad idea, here's why I thought to write it out to disk: context management. Let's say you work over a number of sessions with different Copilot instances, you'll need to reinitialize the context each time, having the plan written to disk makes that much easier. 

I can also just add a line to the .gitignore file to ignore any document named PLAN or plan.md in any directory, which might be the simplest thing to do.

.github/copilot-instructions.md
jack-ryan-nava-pbc: thank you!

.github/copilot-instructions.md
tamara-corbalt: Yep, that sounds like a good idea!

derek-cmsds: > I can also just add a line to the .gitignore file to ignore any document named PLAN or plan.md in any directory, which might be the simplest thing to do.

That sounds like a good thing to do regardless of anything else. Prevents other people from accidentally/intentionally pushing their junk files onto our repo.

derek-cmsds (approved):
Ship it.

PR #4006, opened 2026-04-08
[CMSDS-3988] Simple Footer code connect

jack-ryan-nava-pbc:
## Summary

- [Ticket](https://jira.cms.gov/browse/CMSDS-3988)
- So this one is very simple. Since there are no mappable props in Figma to the Simple Footer, we just have a shell basically. But this is helpful for Figma MCP because it associates that component with the design, so that's a win.
- I had to augment the figma.config.json file so we can scan for components within the child design systems
- I updated the Code Connect CLI package while I was at it

## How to test

1. Confirm that the `<SimpleFooter />` component is in the [Figma file](https://www.figma.com/design/OYkYP4pC9jwS7j2qafwmiv/Design-System-Library?m=dev&node-id=392-13685)

## Checklist


tamara-corbalt (commented):
The code changes look good to me! The one thing that’s giving me pause is how the import statement within the code snippet appears in Figma.
```
import SimpleFooter from "@cms/design-system"
```
But `SimpleFooter` is only available in Medicare. So, it should be:
```
import SimpleFooter from "@cmsgov/ds-medicare-gov"
```
Is there a way to update the generated snippet so it points to the correct package? If not, could we add a comment or note in the snippet to clarify that this component is Medicare-only?

jack-ryan-nava-pbc:
> The code changes look good to me! The one thing that’s giving me pause is how the import statement within the code snippet appears in Figma.
> 
> ```
> import SimpleFooter from "@cms/design-system"
> ```
> 
> But `SimpleFooter` is only available in Medicare. So, it should be:
> 
> ```
> import SimpleFooter from "@cmsgov/ds-medicare-gov"
> ```
> 
> Is there a way to update the generated snippet so it points to the correct package? If not, could we add a comment or note in the snippet to clarify that this component is Medicare-only?

There is, but I wasn't sure the juice was worth the squeeze. 

This gets set on [this line of the configuration file](https://github.com/CMSgov/design-system/pull/4006/changes#diff-5a693ae773c29dd4c357765c226444d46dd0285cb2786f1b1433a79ac786f3f4R4).

To make this work for all of our child design systems we'd need a separate Figma file for each system that overrides the import statement properly. So we'd need 4 configuration files that each get called in separate build and publish scripts.

The reason I'm not sure this is worth is that in my testing with Copilot having this import set didn't change anything. Copilot was able to import the SimpleFooter from the correct child design system.

There are a couple of options:
1. Add a note, some how
2. Just remove the import statements - I think this makes the most sense, as it seems the Agent is able to figure out where to import things from. I can do a test by removing the import statements and see if that degrades the agent and update back here.

What do you think @tamara-corbalt?

tamara-corbalt:
> >   
> > Is there a way to update the generated snippet so it points to the correct package? If not, could we add a comment or note in the snippet to clarify that this component is Medicare-only?
> 
> There is, but I wasn't sure the juice was worth the squeeze.
> 
> This gets set on [this line of the configuration file](https://github.com/CMSgov/design-system/pull/4006/changes#diff-5a693ae773c29dd4c357765c226444d46dd0285cb2786f1b1433a79ac786f3f4R4).
> 
> To make this work for all of our child design systems we'd need a separate Figma file for each system that overrides the import statement properly. So we'd need 4 configuration files that each get called in separate build and publish scripts.
> 
> The reason I'm not sure this is worth is that in my testing with Copilot having this import set didn't change anything. Copilot was able to import the SimpleFooter from the correct child design system.
> 
> There are a couple of options:
> 
> 1. Add a note, some how
> 2. Just remove the import statements - I think this makes the most sense, as it seems the Agent is able to figure out where to import things from. I can do a test by removing the import statements and see if that degrades the agent and update back here.
> 
> What do you think @tamara-corbalt?

Thanks for laying out the tradeoffs. Just documenting here that we talked this through offline, and I’m onboard with removing Core imports from the Figma config.

The current snippet pattern (without explicitly setting the import path in the config), `import SimpleFooter from './SimpleFooter'`, isn’t especially helpful for human users, but it also isn’t blatantly incorrect. Since this also doesn’t appear to degrade agentic output with Copilot, option 2 feels like the most pragmatic choice for now.

I did briefly wonder whether separate config files would truly be too heavy long term, since we already do this in other parts of the repo (for example, Playwright configs), and a small parameterized node script could maybe keep the build/publish side manageable. But for the sake of one small Medicare-specific snippet, I agree the simpler path makes sense.

Approving!

tamara-corbalt (approved):

jack-ryan-nava-pbc:
@tamara-corbalt that's a good point about the multiple configs. I have a separate ticket to explore that work.

PR #4011, opened 2026-04-10
[CMSDS-3970][POC] Explore `llms` manifest and distributed docs packaging flow

tamara-corbalt:
> **DO NOT MERGE**  
> Proof-of-concept branch for implementation discussion

## Summary
- Adds a proof of concept to explore how to distribute documentation markdown files into each package’s `dist/docs` folder
See: [Jira ticket](https://jira.cms.gov/browse/CMSDS-3970) comment.

## Implementation Notes
Below are notes outlining my current approach for this proof of concept, taking some inspiration from how Mastra.ai generates [embedded docs](https://github.com/mastra-ai/mastra/blob/main/scripts/generate-package-docs.ts).

The first step is to determine which documentation markdown files should be distributed into which themed design system package.

For example, we would not want documentation for the Medicare footer ending up in: 
```text
ds-healthcare-gov/dist/docs/components/
```
To support this, we can augment the GraphQL query that currently powers `llms.txt` generation so it also includes theme metadata from frontmatter. From there, we can add logic to define and build `docs-manifest.json`, which will help us map theme-specific content to target packages and determine how shared (theme-agnostic) content should be distributed. This manifest should be generated as part of the docs build process.

From there, we'll wan to add a `generate-docs` script that checks for both the built `public/llms.txt` artifact and `build/docs-manifest.json` within the docs package, reads the manifest, and creates `dist/docs` folders in each design system package. The script should copy only `.txt` assets into those folders according to the manifest entries.

Finally, we’ll need to decide where this script belongs in the broader build orchestration, since it depends heavily on artifacts generated during docs/build. 

## Suggested follow-up tickets
The implementation tickets coming out of this POC might look something like this:

1. **Determine the best schema for `docs-manifest.json`**
2. **Add the necessary logic to build the `docs-manifest.json**
3. **Add the root-level script that creates `dist/docs` in each package**
   - This script should copy the built markdown artifacts into the correct nested structure.
4. **Determine how this hooks into the general build process**
   - This currently depends on artifacts being generated ahead of time in `docs/public`.
   - Should this remain a standalone follow-up script after the general build command? Or should it be incorporated into the main build command.

## How to test
Start by running the following:
```
npm run freshen
cd packages/docs
npm run build
```
At this point, you should see an `docs-manifest.json` file within `packages/docs/public`.
Next, from the repo root, run:

```
npm run build:distributed-docs
```
This should create a docs folder in each core's dist. The docs folder should contain all the markdown files nested according to the hierarchy in the root `llms.txt`.

For example, the Alert markdown file should appear at: `packages/design-system/dist/docs/components/alert/llms.txt`

## Checklist


packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -58,18 +59,92 @@ export const processMdxForHostedMarkdown = (body) => {
...
+        manifest.packages['ds-medicare-gov'].push(pageEntry);
+        break;
+      case 'cmsgov':
+        manifest.packages['ds-cms-gov'].push(pageEntry);
+        break;
+      // Pages without a target theme are included in all packages.
+      case 'core':
+        manifest.packages['shared'].push(pageEntry);
+        break;
+      default:
```
jack-ryan-nava-pbc, line 131: Currently this function is called for all pages, or is it limited to just components?

jack-ryan-nava-pbc: If we have this set as an error, we now have a hard requirement for setting the theme value on all pages. If we miss that accidentally, that'll break our build both locally and on CI if it somehow made it that far. Do we want that? An alternative is to just make the 'shared' case the default case.

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -58,18 +59,92 @@ export const processMdxForHostedMarkdown = (body) => {
...
+ *
+ * Notes:
+ * - `shared` contains all core pages that should be copied into every package dist/docs folder
+ * - child package buckets contain only theme-specific pages that should be copied into their respective package dist/docs folder.
+ */
+export function buildDocsManifest(pages) {
+  const manifest = {
+    generatedAt: new Date().toISOString(),
+    packages: {
+      'shared': [],
```
jack-ryan-nava-pbc: Nit: shared is an accurate way to name this, but I think keeping it in alignment with our current naming schema (aka core) probably would reduce confusion?

scripts/generate-distributed-docs.ts
```diff
@@ -0,0 +1,81 @@
+import fs from 'fs';
+import path from 'path';
+
+const DOCS_PUBLIC_DIR = path.resolve(process.cwd(), 'packages/docs/public');
```
jack-ryan-nava-pbc: Just noting here that process.cwd will return the current working directory, which is likely to be the one the script is in, but you can in theory call the script from anywhere if you spell out the entire path to the script and then this would fail, methinks?

scripts/generate-distributed-docs.ts
```diff
@@ -0,0 +1,81 @@
...
+
+  // Look up each 'shared' entry in manifestJson within DOCS_PUBLIC_DIR/llms.txt and verify it exists.
+  // If it does, copy it to each package's dist/docs nested directory.
+  const manifestJson = JSON.parse(fs.readFileSync(docsManifestPath, 'utf-8'));
+  const sharedPages = manifestJson.packages.shared;
+  sharedPages.forEach((page: { path: string }) => {
+    const sanitizedPath = sanitizeManifestPath(page.path);
+    const pagePath = path.resolve(DOCS_PUBLIC_DIR, sanitizedPath, 'llms.txt');
+    console.log('Looking for shared page at:', pagePath);
+    if (!fs.existsSync(pagePath)) {
```
jack-ryan-nava-pbc, line 68: I don't know that we want a hard failure here. Again thinking about CI, do we want to halt the entire build process if we can't find a particular LLMs.txt file?

jack-ryan-nava-pbc:
Nice work! I need to read this again but some quick questions:
1. How did you settle on the /dist/docs/components/component-name folder convention? 
2. Is there some benefit to having a separate directory with subdirectories over colocating the llms.txt docs file with the component code? Colocating the docs with code would simplify the code you have here.

packages/docs/scripts/llms-txt/index.mjs
tamara-corbalt: Good call. It's not super obvious from this code block but earlier in the flow we call `normalizePages` where we force each page that isn't theme-specific to be labeled with 'core'. But, I think you're right, it’s cleaner to just make the shared path the default case here instead.

scripts/generate-distributed-docs.ts
tamara-corbalt: Ah, I think I see what you’re saying - this works as long as the script is invoked from a specific place (which it currently is via the root package script). Yeah, there’s probably a more robust way to tie this to the script location itself for the final version.

tamara-corbalt:
> 1. How did you settle on the /dist/docs/components/component-name folder convention?
> 2. Is there some benefit to having a separate directory with subdirectories over colocating the llms.txt docs file with the component code? Colocating the docs with code would simplify the code you have here.


Thanks for the early feedback @jack-ryan-nava-pbc! I have thoughts on the two questions.

For 1, I mostly landed on this convention because it loosely follows [Mastra's approach](https://github.com/mastra-ai/mastra/blob/main/scripts/generate-package-docs.ts#L334-L336) of having a dedicated docs tree within the `dist` folder. Also it mirrors the structure already present in `packages/docs/public`, just without the HTML artifacts.
So for example: `docs/public/components/alert/llms.txt`  becomes `<package>/dist/docs/components/alert/llms.txt`

This also naturally mirrors the hierarchical structure in the root `llms.txt`. On the first pass, that felt logical to me, but I'm open to adopting a different nesting structure.

For 2, I’m not sure how I feel about colocating yet. Some of our documentation maps nicely to components, but there are also plenty of markdown files here that don’t really belong to a component module at all (for example, `foundation/typography/headings/llms.txt`).

I’m glad this is coming up now, though, because it raises a fundamental product question: do we only want component docs included in our distributed assets?

jack-ryan-nava-pbc:
> > 1. How did you settle on the /dist/docs/components/component-name folder convention?
> > 2. Is there some benefit to having a separate directory with subdirectories over colocating the llms.txt docs file with the component code? Colocating the docs with code would simplify the code you have here.
> 
> Thanks for the early feedback @jack-ryan-nava-pbc! I have thoughts on the two questions.
> 
> For 1, I mostly landed on this convention because it loosely follows [Mastra's approach](https://github.com/mastra-ai/mastra/blob/main/scripts/generate-package-docs.ts#L334-L336) of having a dedicated docs tree within the `dist` folder. Also it mirrors the structure already present in `packages/docs/public`, just without the HTML artifacts. So for example: `docs/public/components/alert/llms.txt` becomes `<package>/dist/docs/components/alert/llms.txt`
> 
> This also naturally mirrors the hierarchical structure in the root `llms.txt`. On the first pass, that felt logical to me, but I'm open to adopting a different nesting structure.
> 
> For 2, I’m not sure how I feel about colocating yet. Some of our documentation maps nicely to components, but there are also plenty of markdown files here that don’t really belong to a component module at all (for example, `foundation/typography/headings/llms.txt`).
> 
> I’m glad this is coming up now, though, because it raises a fundamental product question: do we only want component docs included in our distributed assets?

Hmm, yeah I guess I had just assumed that we'd send over the llms.txt files for components only since that's what's in the codebase. Shipping everything is a good idea though... could actually be simpler because then we don't run into the situation where the agent can't find docs locally and has to go to the internet.

tamara-corbalt:
> Hmm, yeah I guess I had just assumed that we'd send over the llms.txt files for components only since that's what's in the codebase. Shipping everything is a good idea though... could actually be simpler because then we don't run into the situation where the agent can't find docs locally and has to go to the internet.

I don’t feel especially strongly either way, but I’ve been trying to think through some of the pros and cons of doing it this way.

Pros:
- Agents have local access to things like analytics implementation, typography and utility classes, and pattern-level guidance without needing to fetch from an external source.
- Everything would live aggregated in a clear `dist/docs` tree, with the root llms.txt still serving as that local entry point much like it does for our hosted docs.

Cons:
- Increased package size: although I’m not sure how significant this actually is for compressed markdown assets.
- Broader retrieval surface, more guidance to sift through for a narrow component-related question. We could probably reduce this a bit by being more intentional about what guidance belongs in the distributed bundle (for example, maybe excluding content like blog pages).

scripts/generate-distributed-docs.ts
tamara-corbalt: Heard! In the final version, we might consider collecting missing files and listing them in a wee report. For now, I'll update this log a warning and continue.

tamara-corbalt:
Going to close this pr and carry our discussion forward into the follow-up work in [CMSDS-4005](https://jira.cms.gov/browse/CMSDS-4005)

PR #4017, opened 2026-04-16
[CMSDS-3975] Remove mojibake sequences in `llms.txt` content

tamara-corbalt:
## Summary
- Mojibake sequences are appearing in the Markdown output for our `llms.txt` files. This PR adds a utility to remove them.
[Jira ticket](https://jira.cms.gov/browse/CMSDS-3975)

## How to test
```
cd packages/docs
npm run build
npm run start
```
Inspect some of the `llms.txt` files and compare them to the versions currently in production. The Accordion `llms.txt` file is particularly littered with these sequences. If you don’t see them right away, use Command+F  for `â€™`.

## Checklist

- [ ] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.


packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -135,4 +135,36 @@ export function normalizeMarkdownOutput(input) {
...
+    ['â€™', '’'],
+    ['â€˜', '‘'],
+    ['â€œ', '“'],
+    ['â€\u009d', '”'],
+    ['â€“`', '-'],
+    ['â€“', '-'],
+    ['â€”', '—'],
+    ['â€¦', '…'],
+    ['â€¢', '•'],
+    ['Â ', 'Â'],
```
tamara-corbalt, line 159: We do actually use `Â` in valid content, so I wanted to avoid removing it outright. The issue in the generated output is the stray trailing space which we remove right here.

This list isn’t exhaustive. If we notice additional mojibake in the future, we can simply add them here.

packages/docs/scripts/llms-txt/renderMarkdown.mjs
```diff
@@ -9,7 +10,9 @@ export function renderLlmsMarkdown({ title, description, baseUrl, tree }) {
   markdown += renderPagesList(tree.pages, baseUrl);
 
   markdown += renderNode(tree, baseUrl, 0);
-  return markdown;
+  
+  // In case any mojibake sequences made it through, fix them in the final output.
+  return fixMojibake(markdown);
```
tamara-corbalt, line 15: We also want to remove mojibake sequences in the root `llms.txt`, which is generated in a separate flow from the page-specific `llms.txt` files.

packages/docs/scripts/llms-txt/mdxFixtures.mjs
```diff
@@ -178,4 +178,10 @@ export const PROCESS_MDX_FOR_HOSTED_MARKDOWN_HTML_CODEBLOCK = `Before text
 </ds-alert>
 \`\`\`
 
-After text`;
\ No newline at end of file
+After text`;
+
+export const FIX_MOJIBAKE_INPUT =
```
jack-ryan-nava-pbc, line 183: well this just makes me so happy!

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -135,4 +135,36 @@ export function normalizeMarkdownOutput(input) {
...
+    // We can add more replacements here if needed.
+  ]);
+
+  let result = input;
+
+  for (const [bad, good] of replacements) {
+    result = result.split(bad).join(good);
```
jack-ryan-nava-pbc: I can't believe I never realized that split drops the element that it is split on...

jack-ryan-nava-pbc (approved):
LGTM! Nice work

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
tamara-corbalt: Yeah, this works nicely, but now I’m wondering if it might be clearer to use `replaceAll` instead

tamara-corbalt: Decided to go with `replaceAll` here. It’s logically equivalent (in this case) but more self-documenting than the `split/join`, which I think takes a bit more effort to reason about. @jack-ryan-nava-pbc - Let me know if this still looks good to ship.

packages/docs/scripts/llms-txt/mdxToMarkdown.mjs
```diff
@@ -135,4 +135,36 @@ export function normalizeMarkdownOutput(input) {
...
+    ['â€¦', '…'],
+    ['â€¢', '•'],
+    ['Â ', 'Â'],
+    // We can add more replacements here if needed.
+  ]);
+
+  let result = input;
+
+  for (const [bad, good] of replacements) {
+    result = result.replaceAll(bad, good);
```
jack-ryan-nava-pbc, line 166: Yeah this is way easier to reason about. Nice. Thank you for this change.

jack-ryan-nava-pbc (approved):
Nice!

PR #4027, opened 2026-04-20
[CMSDS-4005] Implement logic to generate `docs-manifest.json` for distributed docs

tamara-corbalt:
## Summary

- Adds logic to generate a `docs-manifest.json` artifact as part of the docs build process.  
  [Jira ticket](https://jira.cms.gov/browse/CMSDS-4005)

- Refactors page normalization and inclusion logic for both the root `llms.txt` and the manifest. This change:
  - separates page normalization from filtering
  - makes artifact-specific page inclusion rules more explicit
  - makes filtering logic easier to extend if needed

## How to test

- Run `npm run build:docs`
- Verify `docs-manifest.json` is generated in `packages/docs/public/` 
- Confirm that the root `llms.txt` is still being generated as before.

## Checklist


packages/docs/scripts/llms-txt/filters.mjs
```diff
@@ -15,3 +9,25 @@ export function getPageIntro(mdxNode) {
...
+export function shouldIncludeInRootLlms(page) {
+  return hasValidSlug(page) && isSidebarEligible(page);
+}
+
+export function shouldIncludeInManifest(page) {
+  return (
+    hasValidSlug(page) &&
+    isSidebarEligible(page) &&
+    !page.slug.includes('blog') &&
+    !page.slug.includes('migration-guides')
```
tamara-corbalt, line 30: Excluding for now. Ideally, we should only include the migration guide that corresponds to the currently distributed version (e.g. v12 -> v12 guide)

jack-ryan-nava-pbc: That makes sense to me.

packages/docs/scripts/llms-txt/filters.mjs
```diff
@@ -15,3 +9,25 @@ export function getPageIntro(mdxNode) {
   const trimmed = String(intro).trim();
   return trimmed;
 }
+
+export function hasValidSlug(page) {
```
jack-ryan-nava-pbc, line 13: I really appreciate how you broke down and renamed these individual steps.

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -82,4 +79,63 @@ export function buildRootLlmsTxt({ siteUrl, description, pages}) {
...
+
+  filteredPages.forEach((page) => {
+    // Add each page to the appropriate package based on its theme.
+    if (page.theme === 'healthcare') {
+      manifest.packages['ds-healthcare-gov'].push(page);
+    } else if (page.theme === 'medicare') {
+      manifest.packages['ds-medicare-gov'].push(page);
+    } else if (page.theme === 'cms') {
+      manifest.packages['ds-cms-gov'].push(page);
+    } else {
```
derek-cmsds, line 137: Do we have any pages that apply to more than one theme but not _all_ themes? Do we anticipate we will have any pages like this?

derek-cmsds (approved):
Looks good. I have one non-blocking question.

packages/docs/scripts/llms-txt/index.mjs
jack-ryan-nava-pbc: If we do, we should pause. That would mean we complicated theming rules and probably we should think through whether we want/need that.

packages/docs/scripts/llms-txt/index.mjs
```diff
@@ -82,4 +79,63 @@ export function buildRootLlmsTxt({ siteUrl, description, pages}) {
...
+  const filteredPages = pages.filter(shouldIncludeInManifest);
+
+  filteredPages.forEach((page) => {
+    // Add each page to the appropriate package based on its theme.
+    if (page.theme === 'healthcare') {
+      manifest.packages['ds-healthcare-gov'].push(page);
+    } else if (page.theme === 'medicare') {
+      manifest.packages['ds-medicare-gov'].push(page);
+    } else if (page.theme === 'cms') {
+      manifest.packages['ds-cms-gov'].push(page);
```
jack-ryan-nava-pbc, line 136: So this array ends up empty in the manifest. That basically means we have no CMSGov specific content, right? Do we have any insight into the instructions LLMs have when parsing these docs-manifest files?

jack-ryan-nava-pbc (approved):
Nicely done!

packages/docs/scripts/llms-txt/index.mjs
tamara-corbalt: We currently don’t have any pages that are geared toward more than one (but not all) themes. And I agree with Jack - if this ever comes up, we’d need to think it through carefully. But as of now, we don’t have any cases like this.

packages/docs/scripts/llms-txt/index.mjs
tamara-corbalt: I’m not envisioning LLMs parsing this artifact directly. The intent is for it to be used downstream in our build process so we can automatically determine which files get distributed to which packages.

jack-ryan-nava-pbc: ohhhhh ok cool. Yeah then no worries. Thanks for clarifying!

PR #4038, opened 2026-04-28
[CMSDS-3963] Choice Figma Code Connection

jack-ryan-nava-pbc:
## Summary

- Added two new code connect snippets: one for Checkbox and one for Radio
- [Ticket](https://jira.cms.gov/browse/CMSDS-3963)

## How to test

1. I've published both components to the production branches in Figma (sorry for deviating, we had to merge in changes that Sophia had into main) [Checkbox](https://www.figma.com/design/OYkYP4pC9jwS7j2qafwmiv/Design-System-Library?m=dev&node-id=18176-60557) and [Radio](https://www.figma.com/design/OYkYP4pC9jwS7j2qafwmiv/Design-System-Library?m=dev&node-id=25175-12575)
2. The functionality for the two components is identical.
3. In `Dev Mode`, navigate to the "Explore Component Behavior" button:
Screenshot 2026-02-19 at 9 42 47 PM

1. Click it.
2. Confirm that the following properties can be edited in Dev Mode and the changes are reflected in the Code Connect snippet: 
3. `checked`: toggled via `Selected` property in Figma
4. `disabled`: toggled via `Disabled` property in Figma
5. `errorMessage`: added via `Has Error` property in Figma
6. `hint`: added via `Has hint text` property in Figma
7. `inversed`: toggled via `Inversed` property in Figma
8. `label`: set via `Label text` property in Figma
9. `size`: set via `Size` property in Figma
10. `checkedChildren`: toggled via `Has child slot` property in Figma

## Checklist


packages/design-system/src/components/ChoiceList/ChoiceCheckBox.figma.tsx
```diff
@@ -0,0 +1,58 @@
...
+        Yes: true,
+        No: false,
+      }),
+      label: figma.nestedProps('.ChoiceLabel', {
+        text: figma.string('Label text'),
+      }),
+      name: 'accessible-name',
+      size: figma.enum('Size', {
+        Small: 'small',
+        Large: undefined,
```
tamara-corbalt, line 33: At first, I was a little confused by Large being mapped to `undefined` here. But then I realized that in our codebase, the `ChoiceSize` type is currently just 'small'. It seems odd that Figma exposes both small and large as options when our API doesn't support a large variant.

tamara-corbalt (commented):
Looks good overall! There are a couple of things in the component playground in Figma that felt a bit odd:
- There are two `inversed` toggles in both the radio and checkbox views.
- Changing the “Label text” input updates the visual component, but not the snippet, at least not consistently.
- I also left a comment about the size mapping: our `Choice` component API only supports 'small', but we expose 'small' and 'large' as options in Figma.

packages/design-system/src/components/ChoiceList/ChoiceCheckBox.figma.tsx
jack-ryan-nava-pbc: "Large" is effectively the default size. I would argue that from a design perspective it makes more sense to map the sizing to large and small rather than just include a setting for "small", but I can defer to @sophia-cmsds on that.

jack-ryan-nava-pbc:
@tamara-corbalt I think the two 'Inversed" controls are necessary, unfortunately. The .ChoiceLabel is effectively a separate reusable component (akin to the Label component) so any use of it inherits the 'Inversed' prop, and then the actual Checkbox button needs an "Inversed"  toggle to actually invert the color scheme of the checkbox//radio button itself. @sophia-cmsds is that correct?

jack-ryan-nava-pbc:
`Changing the “Label text” input updates the visual component, but not the snippet, at least not consistently`

yeah, I've gotten it to work several times in a row, but I've seen Code Connect become a bit flakier with more complex component mappings. I think the important thing is that the mapping exists for the MCP server to take advantage of, not necessarily that the toggles in the playground work 100% of the time.

packages/design-system/src/components/ChoiceList/ChoiceCheckBox.figma.tsx
tamara-corbalt: @jack-ryan-nava-pbc Ooooh, I see. Thanks for clarifying that!

tamara-corbalt:
> @tamara-corbalt I think the two 'Inversed" controls are necessary, unfortunately. The .ChoiceLabel is effectively a separate reusable component (akin to the Label component) so any use of it inherits the 'Inversed' prop, and then the actual Checkbox button needs an "Inversed" toggle to actually invert the color scheme of the checkbox//radio button itself. @sophia-cmsds is that correct?

Oh ok, gotcha! Thanks for explaining that. Assuming @sophia-cmsds confirms this, is there a way to make the relationship between these two controls clearer through naming?

jack-ryan-nava-pbc:
> > @tamara-corbalt I think the two 'Inversed" controls are necessary, unfortunately. The .ChoiceLabel is effectively a separate reusable component (akin to the Label component) so any use of it inherits the 'Inversed' prop, and then the actual Checkbox button needs an "Inversed" toggle to actually invert the color scheme of the checkbox//radio button itself. @sophia-cmsds is that correct?
> 
> Oh ok, gotcha! Thanks for explaining that. Assuming @sophia-cmsds confirms this, is there a way to make the relationship between these two controls clearer through naming?

Well, they are listed underneath their corresponding components:

Screenshot 2026-04-30 at 11 58 51 AM

Are you thinking of something different?

tamara-corbalt:
> > > @tamara-corbalt I think the two 'Inversed" controls are necessary, unfortunately. The .ChoiceLabel is effectively a separate reusable component (akin to the Label component) so any use of it inherits the 'Inversed' prop, and then the actual Checkbox button needs an "Inversed" toggle to actually invert the color scheme of the checkbox//radio button itself. @sophia-cmsds is that correct?
> > 
> > 
> > Oh ok, gotcha! Thanks for explaining that. Assuming @sophia-cmsds confirms this, is there a way to make the relationship between these two controls clearer through naming?
> 
> Well, they are listed underneath their corresponding components: Are you thinking of something different?

Ohhh ok, I see that now! I think I just missed how they’re grouped in the UI at first glance.

tamara-corbalt (approved):

PR #4036, opened 2026-04-24
[CMSDS-4032] Remove React CDN bundle

derek-cmsds:
## Summary

Currently we distribute our React components via our CDN in addition to NPM. The CDN allowed teams to download and run React in basic HTML pages without needing a complex build step.

The React team decided to remove support UDM, a file format the we use to enable our React CDN components. As a result, we can’t distribute our files via the CDN without considerable engineering work.

This PR removes the React bundle from the CDN, any documentation referencing it, and the associated example application.

- [CMSDS-4032](https://jira.cms.gov/browse/CMSDS-4032)

## How to test

1. Verify `npm run freshen` completes successfully.
2. Verify `npm run test:browser:examples` completes successfully.

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] Selected appropriate release milestone

### If this is a change to code:

- [x] Created or updated unit tests to cover any new or modified code
- [x] Verified that running both `npm run test:unit` and `npm run test:browser:all` were each successful
- [x] If necessary, updated unit-test snapshots (`npm run test:unit:update`) and browser-test snapshots (`npm run test:browser:all:update`)

PR #4035, opened 2026-04-23
[CMSDS-3154] Fix build errors

derek-cmsds:
## Summary

- The main purpose of this PR is to get the design system to build and the doc site and Storybook to successfully run after the React 19 upgrade. Mostly, this is done by fixing stray Typescript errors and removing reference to UDM files that React no longer generates.
- [CMSDS-3154](https://jira.cms.gov/browse/CMSDS-3154)

## How to test

1. Run `npm run freshen`
2. Verify all steps run succesfully.
3. Run `npm start`
4. Verify you can access the access the doc site at http://localhost:8000/ and from there navigate to Storybook.

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] Selected appropriate release milestone

PR #4039, opened 2026-04-29
[CMSDS-3301] Upgrade to React 19

derek-cmsds:
## Summary

- Various changes need to get the Design System updated to React 19.
- [CMSDS-3301](https://jira.cms.gov/browse/CMSDS-3301)
- PRs included:
   - #3943 
   - #3972 
   - #4035 
   - #4036 

## How to test

1. Run `npm run freshen` and ensure the Design System builds successfully.
2. Run `npm start` and verify the doc site and Storybook load successfully.
3. Run `npm test`, `npm run test:unit:wc` and `npm run test:browser:all` and verify all tests pass.

## Checklist


packages/design-system/src/components/Icons/ArrowsStackedIcon.tsx
```diff
@@ -7,7 +7,7 @@ const defaultProps = {
   viewBox: '0 0 10 14',
 };
 
-export function ArrowsStackedIcon(props: IconCommonProps): React.ReactElement {
+export function ArrowsStackedIcon(props: IconCommonProps): React.ReactElement<any> {
```
tamara-corbalt, line 10: Noticing a lot of these icons now have the return type `React.ReactElement<any>`. Curious why that’s necessary on these wrappers.

tamara-corbalt (commented):
Everything looks great overall, nice work! All of the testing steps succeeded. Just left one question about whether the type changes on the icon wrappers are necessary.

jack-ryan-nava-pbc:
@derek-cmsds can we update the name of this PR to include the Epic ticket number? https://jira.cms.gov/browse/CMSDS-3301

packages/design-system/src/components/Icons/ArrowsStackedIcon.tsx
derek-cmsds: Yeah, that's something React 19 introduced: https://react.dev/blog/2024/04/25/react-19-upgrade-guide#changes-to-the-reactelement-typescript-type
> The props of React elements now default to `unknown` instead of `any` if the element is typed as ReactElement.

derek-cmsds:
> @derek-cmsds can we update the name of this PR to include the Epic ticket number? https://jira.cms.gov/browse/CMSDS-3301

Updated!

packages/design-system/src/components/Icons/ArrowsStackedIcon.tsx
tamara-corbalt: Ok got it. So the goal is just to preserve the React 18 behavior where ReactElement defaulted to any, rather than allowing it to become unknown. That makes sense.

tamara-corbalt (approved):
LGTM 🎉

packages/design-system/src/components/Dropdown/DropdownMenu.tsx
```diff
@@ -13,7 +13,7 @@ interface DropdownMenuProps<T> extends AriaListBoxOptions<T> {
   children?: React.ReactNode;
   componentClass: string;
   heading?: React.ReactNode;
-  labelId: string;
+  labelId?: string;
```
jack-ryan-nava-pbc, line 16: Why did this become optional?  This gets used in the `aria-labelledby` attribute, and every other prop that composes the `aria-labelledby` attribute is optional so if all of them are omitted then we don't have a value for this attribute which means the menu would exist without an accessible label.

jack-ryan-nava-pbc (changes requested):
Overall this looks awesome! I have one question in a comment and one change request:
In `packages/design-system/package.json` please update the React version in the Peer Dependencies list: 
  "peerDependencies": {
    "react": ">=17.0.0",
    "react-dom": ">=17.0.0"
  },
to 
  "peerDependencies": {
    "react": ">=19.2.4",
    "react-dom": ">=19.2.4"
  },

packages/design-system/src/components/Dropdown/DropdownMenu.tsx
derek-cmsds: `LabelProps` is a type shared across multiple components and considers `id` to be an optional property, so Typescript throws a fit if we send `LabelProps.id` to `labelId` and it requires a string. There may be some Typescript jujitsu that can be done to make it behave, but this seemed to be the simplest solution that also forces us to think about what should be done when all of those props are missing.

derek-cmsds:
> In `packages/design-system/package.json` please update the React version in the Peer Dependencies list: "peerDependencies": { "react": ">=17.0.0", "react-dom": ">=17.0.0" }, to "peerDependencies": { "react": ">=19.2.4", "react-dom": ">=19.2.4" },

Peer dependencies have been updated.

packages/design-system/src/components/Dropdown/DropdownMenu.tsx
jack-ryan-nava-pbc: Ugh yeah I just stepped through and it's a very tangled up series of operations to get to that labelId value. Lots of places for undefined to creep in. 

We'd need to enforce a defined value at some point up in the component hierarchy but that seems non-trivial. 

On a very quick glance it seems most sensible to not allow the `id` prop from the `useLabelProps` function to be overwritten by a potentially null/undefined thing on line 253. Altering `useId` is an option, but we need to make sure that this id actually corresponds to the parent element of the DropdownMenu. 

What are your thoughts?

derek-cmsds: My thought is that refactoring the `DropdownMenu` component is outside the scope of this React 19 upgrade epic and we need a quick and easy solution to get this across the finish line.

jack-ryan-nava-pbc: Hmm, I am worried that this could cause an accessibility regression. In that the menu could in some circumstances not be labelled.

jack-ryan-nava-pbc: That said, I am not sure what the best path forward is.

derek-cmsds: No actual functionality has changed in the component code and we do not publicly expose DropdownMenu as a component, so in theory, we control whether `labelId` is ever null and are always providing a `labelId` through the Dropdown component. Typescript does not recognize that assurance, but there does not appear to be a true regression here.

jack-ryan-nava-pbc (approved):
Great job!

PR #4043, opened 2026-05-04
[CMSDS-3956] Create 404 Lite file

derek-cmsds:
## Summary

- Creates a self-contained HTML file of the [404 pattern](https://design.cms.gov/patterns/404-page-guidance/?theme=core) that is under 4 KB, can be deployed on static file servers, and can be served with no external dependencies.
- [CMSDS-3956](https://jira.cms.gov/browse/CMSDS-3956)

## How to test

1. Run `npm start`
2. Navigate to http://localhost:8000/patterns/404-page-guidance/404-lite.html
3. Verify the page that loads looks similar to the screenshot below. (Font family may vary depending on device.)
4. Verify the design responsively adapts to different screen resolutions.
5. Verify the pattern is adequately accessible.
6. Click the "Go to home" button.
7. Verify you are taken to https://www.cms.gov/

## Screenshots
Screenshot 2026-05-04 at 13 28 23

## Checklist

### DIFF
--- a/packages/docs/static/patterns/404-page-guidance/404-lite.html
+++ b/packages/docs/static/patterns/404-page-guidance/404-lite.html
@@ -1,0 +1,66 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Page Not Found</title>
+  <style>
+    :root {
+      --color-primary: #0071bc;
+      --font-family-heading: 'Open Sans', 'Inter', 'Roboto', 'Helvetica Neue', 'Arial Nova', 'Nimbus Sans', 'Arial', 'sans-serif';
+      --font-line-height-heading: 1.2999999523162842;
+      --font-size-3xl: 2.25rem;
+      --font-size-3xl--mobile: 1.875rem;
+      --font-weight-bold: 700;
+    }
+
+    * {
+      box-sizing: border-box;
+    }
+
+    h1 {
+      font: inherit;
+    }
+
+    .ds-text-heading--3xl {
+      font-family: var(--font-family-heading);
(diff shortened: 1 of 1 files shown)


sophia-cmsds (approved):
lgtm

packages/docs/static/patterns/404-page-guidance/404-lite.html
```diff
@@ -0,0 +1,122 @@
...
+
+    h1 {
+      font: inherit;
+    }
+
+    p {
+      margin-block: 1em 0;
+    }
+
+    .ds-c-button, .ds-c-button:visited {
```
jack-ryan-nava-pbc, line 45: Oh. I somehow did not fully get that we were going to be putting this inline.

jack-ryan-nava-pbc (commented):
So I think this looks good. The problem of inline styling is maybe worth grappling with a little. 

IIRC we need to ship less that 64 kb. According to my browser we're at 1.5kb. So we have some head room. 

Is it possible to copy all of the styling needed for the classes referenced without exceeding 64kb? If so then we could use a script to copy over those styles at build time (separate ticket) unless you have a better idea?

Also, we should get Sam Lerner's eyes on this just to make sure this will work for the teams directly impacted.

jack-ryan-nava-pbc (approved):
LGTM

jack-ryan-nava-pbc:
> So I think this looks good. The problem of inline styling is maybe worth grappling with a little.
> 
> IIRC we need to ship less that 64 kb. According to my browser we're at 1.5kb. So we have some head room.
> 
> Is it possible to copy all of the styling needed for the classes referenced without exceeding 64kb? If so then we could use a script to copy over those styles at build time (separate ticket) unless you have a better idea?
> 
> Also, we should get Sam Lerner's eyes on this just to make sure this will work for the teams directly impacted.

We opted to cross this bridge when we get there. We'll see how often these styles change.

PR #4049, opened 2026-05-07
[CMSDS-4007] Add script to embed documentation in distributed NPM packages.

tamara-corbalt:
## Summary

- Adds a `generate-distributed-docs` script which copies hosted Markdown (`llms.txt`) files from `packages/docs/public` into the appropriate package `dist/docs` directories according to the generated `docs-manifest.json` artifact.
- Integrates this script into the main build command so these Markdown files are included in published npm packages during release.
- Updates the README with documentation for the new distributed-docs build commands.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-4007)

## How to test

1. Run `npm run build`
2. Verify that `dist/docs` directories are generated for each design system package
3. Verify that:
   - shared documentation files are copied into all four design system packages
   - theme-specific documentation files are only copied into their appropriate package
4.  You can cross-reference copied files against `packages/docs/public/docs-manifest.json`

## Checklist


jack-ryan-nava-pbc (commented):
Question: currently the top level llms.txt file that is copied over has all remote urls in it. Is there a plan to make it so an agent is directed to the local directory/file location rather than our web based location?

tamara-corbalt:
> Question: currently the top level llms.txt file that is copied over has all remote urls in it. Is there a plan to make it so an agent is directed to the local directory/file location rather than our web based location?


Ohhh brother, this is a very good call! At first, I was thinking our root `llms.txt` could serve as an index file for the embedded docs since the structure roughly mirrors how files are grouped together. But of course, the urls still point to design.cms.gov, so that's not cool. 🙄

Also, I didn’t think about this initially, but we’re being more surgical about the correct destination for certain pages. For example, only healthcare-related files are copied into the healthcare package, whereas the hosted root llms.txt lists all documentation pages. 

I think we should remove the copied root llms.txt for now, and then address this properly during the skills.md follow-up work. What do you think @jack-ryan-nava-pbc

jack-ryan-nava-pbc:
> > Question: currently the top level llms.txt file that is copied over has all remote urls in it. Is there a plan to make it so an agent is directed to the local directory/file location rather than our web based location?
> 
> Ohhh brother, this is a very good call! At first, I was thinking our root `llms.txt` could serve as an index file for the embedded docs since the structure roughly mirrors how files are grouped together. But of course, the urls still point to design.cms.gov, so that's not cool. 🙄
> 
> Also, I didn’t think about this initially, but we’re being more surgical about the correct destination for certain pages. For example, only healthcare-related files are copied into the healthcare package, whereas the hosted root llms.txt lists all documentation pages.
> 
> I think we should remove the copied root llms.txt for now, and then address this properly during the skills.md follow-up work. What do you think @jack-ryan-nava-pbc

I think creating a skill that knows about the locations of all these files makes a lot of sense. Let's reconvene and get that work going. So I agree with your suggestion to remove the copied root llms.txt file for now.

jack-ryan-nava-pbc (approved):
LGTM!

PR #4065, opened 2026-05-20
[CMSDS-4051] Resolve item announcement issue on Autocomplete

jack-ryan-nava-pbc:
## Summary

- [Ticket](https://jira.cms.gov/browse/CMSDS-4051)
- The individual items in the Autocomplete element were not being read out JAWS or NVDA
- Component worked ok in Safari & Chrome, however the text content of the list item was not being read out when JAWS or NVDA stepped through using the DownArrow key. 

## How to test

Really, the reporting team will need to test once we cut the beta release, but putting testing steps here for posterity

1. Run locally - `npm run storybook`
2. Repeat for Chrome and Safari - if you have access to NVDA (via Browserstack or Windows VM) open in Edge
3. Open in Chrome - navigate to Autocomplete story
4. Start Voiceover - CMD + FN + F5 - or enable NVDA in Browserstack
5. Tab so that focus is on the TextField of the Autocomplete
6. Type a few letters so that the list options displays
7. The number of options should be read out
8. Press the down arrow key
9. Options should be visually highlighted
10. Text content of each item should be announced

## Checklist

### Chrome:
https://github.com/user-attachments/assets/cb5aa9e9-6253-40cc-9b4e-b253db916ded

### Safari:
https://github.com/user-attachments/assets/418ecdab-403e-47f3-9ed2-49fad2eba57b

### NVDA & Edge:
https://github.com/user-attachments/assets/fd0c77be-58f1-40bb-a00d-206290fcdc81

### DIFF
--- a/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
+++ b/packages/design-system/src/components/Autocomplete/Autocomplete.tsx
@@ -17,6 +17,7 @@ import {
 import { t } from '../i18n';
 import { useComboBox } from '../react-aria'; // from react-aria
 import { useComboBoxState } from '../react-aria'; // from react-stately
+import type { CollectionChildren } from '@react-types/shared';
 
 export interface AutocompleteItem extends Omit<React.HTMLAttributes<'option'>, 'name'> {
   /**
@@ -117,7 +118,7 @@ export interface AutocompleteProps {
   /**
    * Called when the user selects an item and the selected item has changed. Called with the item that was selected.
    */
-  onChange?: (selectedItem: AutocompleteItem) => void;
+  onChange?: (selectedItem: AutocompleteItem | null) => void;
   /**
    * Called when the child `TextField` value changes. Is called with a string representing the input value.
    */
@@ -168,15 +169,23 @@ export const Autocomplete = (props: AutocompleteProps) => {
     ...autocompleteProps
   } = props;
 
-  const hasValidStandaloneItems = items?.some((item) => !('items' in item));
-  const hasValidGroupedItems = items?.some((item) => 'items' in item && item.items?.length > 0);
+  function hasRenderableItems(items: AutocompleteItems | undefined): items is AutocompleteItems {
+    if (!items?.length) return false;
+
+    const hasValidStandaloneItems = items?.some((item) => !('items' in item));
+    const hasValidGroupedItems = items?.some((item) => 'items' in item && item.items?.length > 0);
+
+    return hasValidStandaloneItems || hasValidGroupedItems;
+  }
 
   // Determine what we'll show based on state
-  let reactStatelyItems = [];
+  let reactStatelyItems: React.ReactElement[] = [];
   let statusMessage;
 
-  if (hasValidStandaloneItems || hasValidGroupedItems) {
-    reactStatelyItems = renderReactStatelyItems(items);
+  if (hasRenderableItems(items)) {
+    // items can be "undefined" but because of the .some checks above we know
+    // items is defined and has a length, hence including the ! here to assert so.
+    reactStatelyItems = renderReactStatelyItems(items!);
   } else if (loading) {
     // If we're waiting for results to load, show the non-selected message
     statusMessage = renderStatusMessage(loadingMessage ?? t('autocomplete.loadingMessage'));
@@ -193,7 +202,7 @@ export const Autocomplete = (props: AutocompleteProps) => {
     ...autocompleteProps,
     allowsCustomValue: true,
     allowsEmptyCollection: true,
-    children: reactStatelyItems,
+    children: reactStatelyItems as CollectionChildren<object>,
     inputValue: textField.props.value,
     onInputChange: onInputValueChange
       ? (value) => {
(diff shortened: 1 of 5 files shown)


packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -117,7 +118,7 @@ export interface AutocompleteProps {
   /**
    * Called when the user selects an item and the selected item has changed. Called with the item that was selected.
    */
-  onChange?: (selectedItem: AutocompleteItem) => void;
+  onChange?: (selectedItem: AutocompleteItem | null) => void;
```
jack-ryan-nava-pbc, line 121: This onChange function is already set up to handle null items, this change just makes our documentation more honest.

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -168,15 +169,23 @@ export const Autocomplete = (props: AutocompleteProps) => {
     ...autocompleteProps
   } = props;
 
-  const hasValidStandaloneItems = items?.some((item) => !('items' in item));
-  const hasValidGroupedItems = items?.some((item) => 'items' in item && item.items?.length > 0);
+  function hasRenderableItems(items: AutocompleteItems | undefined): items is AutocompleteItems {
```
jack-ryan-nava-pbc, line 172: We were doing this check using two variables above, I just wrapped them in this function cuz it seemed like a nicer experience when doing the if check.

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -217,6 +226,7 @@ export const Autocomplete = (props: AutocompleteProps) => {
 
   const inputRef = useRef<HTMLInputElement>(null);
   const listBoxRef = useRef<HTMLElement>(null);
+  const popoverRef = useRef<HTMLDivElement>(null);
```
jack-ryan-nava-pbc, line 229: New ref for tracking the popover element in the DOM. Before the entire listbox was being hidden from the accessibility tree, which is why JAWS and NVDA missed the elements therein. VoiceOver obviously handles these things differently, maybe relying on the list semantics?

Anyway, the addition of this ref exposes the list box items (the DropdownMenu) to the accessibility tree by some mechanism in React aria.

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -226,7 +236,7 @@ export const Autocomplete = (props: AutocompleteProps) => {
       isDisabled: textField.props.disabled,
       inputRef,
       listBoxRef,
-      popoverRef: listBoxRef,
+      popoverRef,
```
jack-ryan-nava-pbc, line 239: Apparently the listbox ref needs to be different than the popoverRef or react aria excludes it from the accessibility tree and JAWS and NVDA miss it.

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -290,34 +300,31 @@ export const Autocomplete = (props: AutocompleteProps) => {
...
-          size={size}
-          state={state}
-          triggerRef={wrapperRef}
-          listBoxRef={listBoxRef}
-        >
-          {statusMessage}
-        </DropdownMenu>
-      )}
-
+      <div ref={popoverRef}>
```
jack-ryan-nava-pbc, line 303: The addition of this <div> is the core change here. It includes the ref that we setup before and feed into the useComboBox state handler so that this gets tracked in the accessibility tree and by react-aria's mutation observer.

packages/design-system/src/components/Autocomplete/Autocomplete.tsx
```diff
@@ -290,34 +300,31 @@ export const Autocomplete = (props: AutocompleteProps) => {
...
+          >
+            {statusMessage}
+          </DropdownMenu>
+        )}
+      </div>
       {clearSearchButton && (
         <Button
           className="ds-u-padding-right--0 ds-c-autocomplete__clear-btn"
           onClick={() => {
-            state.setSelectedKey(null);
```
jack-ryan-nava-pbc, line 314: `selectedKey` is deprecated, so I just removed it?

packages/design-system/src/components/Autocomplete/utils.tsx
```diff
@@ -69,7 +69,7 @@ export function getTextFieldChild(children: ReactNode): ReactElement<any> | unde
 export function getActiveDescendant(
   rootId: string,
   state: ComboBoxState<object>,
-  items: AutocompleteItem[]
+  items: AutocompleteItem[] | undefined
```
jack-ryan-nava-pbc, line 72: I allowed this to be undefined for convenience sake and we have a type check right in the first line with the nullish coalescing operator.

packages/design-system/src/components/utilities/mergeRefs.ts
```diff
@@ -4,7 +4,7 @@
  * Borrowed from https://github.com/gregberge/react-merge-refs/blob/main/src/index.tsx
  */
 export default function mergeRefs<T = any>(
-  refs: Array<React.RefObject<T | null> | React.Ref<T>>
+  refs: Array<React.RefObject<T | null> | React.Ref<T> | undefined>
```
jack-ryan-nava-pbc, line 7: Allowing this to be an array of undefined also is a convenience change. The type checking in this function should safely exclude undefined. The single equal sign check is safe here since null and undefined are considered equal in that case (triple equals would return true).

derek-cmsds (approved):
LGTM. I appreciate the comments and the added type hints.

tamara-corbalt (approved):
Looks great! Thank you for adding the helpful comments. 

There are some unit test snapshot failures on CI that I’m not seeing locally. Approving ahead of those being resolved.

PR #4089, opened 2026-06-08
[CMSDS-4090] Adds a skill for CMSDS component implementation

tamara-corbalt:
## Summary

- Adds the CMS Design System's first skill to assist with component implementation.
[Jira ticket](https://jira.cms.gov/browse/CMSDS-4090)

Skills follow the structure described in [Anthropic's documentation](https://resources.anthropic.com/hubfs/The-Complete-Guide-to-Building-Skill-for-Claude.pdf):

```text
skills/
  my-cool-skill/
    SKILL.md
    references/
      additional Markdown files loaded as needed
```

The YAML frontmatter in `SKILL.md` is especially important:

- `name` must be kebab-case and match the skill directory name.
- `description` should clearly describe the skill and include trigger phrases that help invoke it.

During skill discovery, only the frontmatter of SKILL.md is loaded. When a trigger phrase matches, the body of `SKILL.md` is loaded, and any referenced supporting documents are loaded on demand.

To keep the main skill concise, Healthcare and Medicare guidance has been separated into supporting documents under `references/`. I didn't do this for CMSgov since there are no CMSgov-specific components yet.

## How to test

1. Run `npm run start`
2. Navigate to: http://localhost:8000/cms-design-system-skills.zip
 3. Download the ZIP file and verify that it contains the expected skill directory and all supporting files under `references/`.

## Checklist


derek-cmsds (commented):
Two notes:

1. Do we foresee a need for people to only download a specific skill? It looks like right now, the zip file bundles up _all_ skills in the `skills` folder. That currently only includes one skill, but I believe others are soon to follow.
2. How do we feel about posting a zip file on the production server and asking our users to download and unzip it? Is this the common way to distribute skills? Do we need CMS permission to host this type of file? (I recall us getting attention for distributing `llms.txt` files.) @jack-ryan-nava-pbc, do you have an opinion on this?

packages/docs/package.json
```diff
@@ -9,6 +9,7 @@
   "scripts": {
     "copyThemes": "copyfiles -u 4 ../design-system-tokens/dist/css-vars/*-theme.css ./static/themes/",
     "build": "npm run copyThemes && gatsby build",
+    "build:skills-zip": "node ./scripts/skills/build-skills-zip.mjs",
```
jack-ryan-nava-pbc, line 12: So is the idea that we'll run this as a separate step from the main build process in CI? Just curious about the plan to wire this command up.

packages/docs/skills/implement-cmsds-react-components/SKILL.md
```diff
@@ -0,0 +1,31 @@
+---
+name: implement-cmsds-react-components
```
jack-ryan-nava-pbc, line 2: Will we want to extend this skill to support our Web Components as well? Or do you think that would be best served by a wholly separate skill?

packages/docs/skills/implement-cmsds-react-components/SKILL.md
```diff
@@ -0,0 +1,31 @@
...
+description: Implements, modifies, and troubleshoots CMS Design System (CMSDS) React components. Use when users ask for "CMSDS component guidance", "implement a CMSDS component", or "fix CMSDS component usage".
```
jack-ryan-nava-pbc, line 3: Do we think folks are gonna say "CMSDS" in their prompts? I guess we can start with this language, get some feedback and iterate. I'm just curious about how natural this description is compared to actual user prompts.

jack-ryan-nava-pbc (approved):
Looks good! Just questions. I'm very interested to see how this goes...

jack-ryan-nava-pbc:
> Two notes:
> 
> 1. Do we foresee a need for people to only download a specific skill? It looks like right now, the zip file bundles up _all_ skills in the `skills` folder. That currently only includes one skill, but I believe others are soon to follow.
> 2. How do we feel about posting a zip file on the production server and asking our users to download and unzip it? Is this the common way to distribute skills? Do we need CMS permission to host this type of file? (I recall us getting attention for distributing `llms.txt` files.) @jack-ryan-nava-pbc, do you have an opinion on this?

Since we control the server that hosts these files, I think we're ok but let's double check. I think our procedure will be: push to dev site, solicit feedback on the installation pattern & ask for security review on the Dev site. The reason I'm not _super_ worried is that we distribute .zips via our Github repo, and granted it's not a .gov domain, but it's still our property. 

I think getting something on the dev site for security to actually touch and work with will get us the answer we need.

I think we collected the alternative distribution patterns in confluence, but I can enumerate them here:
1. Doc Site
2. CDN
3. NPM Package - this could make a lot more sense to have a separate directory if we have lots of skills.

packages/docs/skills/implement-cmsds-react-components/SKILL.md
tamara-corbalt: Yes great question! I thought about this a bit.

Personally, I'd rather offer a dedicated Web Components skill than extend this one to include web components. I think keeping the scope of each skill tightly focused is the right approach, at least for now.

As it is, this skill is already fairly complicated in that it has to explain the relationship between the core design system and the theme-specific packages, including components that wrap other components from the core package. Adding the web component/Preact layer on top of that feels like a separate concern and would significantly complicate the logic for directing the agent to the appropriate guidance.

packages/docs/skills/implement-cmsds-react-components/SKILL.md
tamara-corbalt: Yes! ok, so my thinking here was that some downstream applications may have their own component libraries or skills. Including "CMSDS" in the trigger phrases should help the agent more confidently load _this_ skill instead of some more generic skill they might have in their environment. We can always adjust these triggers if folks find them unwieldy or the CMSDS prefix proves unnecessary.

I also think we'll want to make these trigger phrases explicit in our guidance on the docs site.

packages/docs/skills/implement-cmsds-react-components/SKILL.md
jack-ryan-nava-pbc: This sounds good to me! Thanks for clarifying!

packages/docs/skills/implement-cmsds-react-components/SKILL.md
jack-ryan-nava-pbc: That seems fair and reasonable.

packages/docs/package.json
tamara-corbalt: So I thought about this too, but my intention was for this to be a utility command to regenerate the skills zip (and commit the updated artifact) when the skills change, not necessarily a step in the normal build process. It would probably be helpful to document that in a README so folks know what it's for. I'm going to add that.

tamara-corbalt:
> > Two notes:
> > 
> > 1. Do we foresee a need for people to only download a specific skill? It looks like right now, the zip file bundles up _all_ skills in the `skills` folder. That currently only includes one skill, but I believe others are soon to follow.
> > 2. How do we feel about posting a zip file on the production server and asking our users to download and unzip it? Is this the common way to distribute skills? Do we need CMS permission to host this type of file? (I recall us getting attention for distributing `llms.txt` files.) @jack-ryan-nava-pbc, do you have an opinion on this?
> 
> Since we control the server that hosts these files, I think we're ok but let's double check. I think our procedure will be: push to dev site, solicit feedback on the installation pattern & ask for security review on the Dev site. The reason I'm not _super_ worried is that we distribute .zips via our Github repo, and granted it's not a .gov domain, but it's still our property.
> 
> I think getting something on the dev site for security to actually touch and work with will get us the answer we need.
> 
> I think we collected the alternative distribution patterns in confluence, but I can enumerate them here:
> 
> 1. Doc Site
> 2. CDN
> 3. NPM Package - this could make a lot more sense to have a separate directory if we have lots of skills.

- @derek-cmsds - Yes we we do have plans to add another skill in the coming weeks. I think for now though, it makes sense to keep them bundled in one place. 
- @jack-ryan-nava-pbc - I think pushing the skills zip to the Dev site and requesting security review makes sense. Added this ticket: https://jira.cms.gov/browse/CMSDS-4121

derek-cmsds (approved):

PR #4119, opened 2026-06-25
[CMSDS-4119] Guard ds-tabs parseChildren against undefined and non-element nodes

### LINKED ISSUE #4093: [Bug:] Tabs cannot be used in an Angular project
### Describe the Bug

I am wrapping Design System components in Angular Directives for use in a project. Attempting to use ds-tabs and ds-tab-panel results in one or two exceptions being thrown which prevents the tabs from rendering properly. From inspecting the code (ds-tabs.tsx), I believe the following changes to the parseChildren() function would help:
1. Depending on timing, the function may be called with nodes being undefined. A falsy check at the top would prevent the exception from the map call.
2. Inside the map function, it is possible for element to be undefined, since Angular inserts comment nodes. Ignoring these would prevent the exception from the subsequent attributes check.

### To reproduce

1. Using the existing Angular example, create simple Angular Directives wrapping the ds-tabs and ds-tab-panel selectors (I cannot attach files so I have pasted some code here).
2. In the app.component.html template, insert a <ds-tabs> block, using a for block to add ds-tab-panel elements.
3. Run the example app.
4. Observe the console - you will observe either "TypeError: Cannot read properties of undefined (reading 'map')" or "TypeError: Cannot read properties of null (reading 'props')".

```
import { Directive, input, output } from '@angular/core';

@Directive({
  selector: 'ds-tabs',
  host: {
    '[attr.defaultSelectedId]': 'initialTab()',
    '(ds-change)': 'onChange($event)',
  },
})
export class CmsDsTabsDirective {
  readonly initialTab = input<string>();
  readonly tabChanged = output<Event>();

  onChange(event: Event): void {
    this.tabChanged.emit(event);
  }
}
```
```
import { Directive, input } from '@angular/core';

@Directive({
  selector: 'ds-tab-panel',
  host: {
    '[attr.id]': 'tabId()',
    '[attr.tab]': 'tabLabel()',
  },
})
export class CmsDsTabPanelDirective {
  tabId = input.required<string>();
  tabLabel = input.required<string>();
}

```

### Expected behavior

No exceptions thrown.

### Severity/Impact

Low: Nice to have/UX improvement suggestion

### Desktop Operating System

Windows 11

### Desktop Browser Name

Chrome

### Desktop Browser Version

149.0.7827.102

### Device

_No response_

### Mobile Operating System

_No response_

### Mobile Browser Name

_No response_

### Mobile Browser Version

_No response_

### Additional context

_No response_

### Codebase/Repository URL

_No response_

### Design System Version

18.0.0

_No response_

### Team Name

_No response_

arpitjain099:
## Summary

- Closes #4093

`parseChildren` in the `ds-tabs` web component crashes in two ways when the component is driven by another framework. The issue reports it from Angular directives, but the root cause is generic.

First, depending on render timing the function can be called before children exist, so `nodes` is `undefined` and `nodes.map(...)` throws `TypeError: Cannot read properties of undefined (reading 'map')`. I reproduced this in the unit tests (rendering `ds-tabs` with no children) and confirmed the same TypeError before the fix.

Second, frameworks such as Angular insert comment/text nodes alongside the real panels. Those nodes have no `props`, so the existing `element.props` access throws. I could not actually trigger this one inside the Jest harness because Preact's JSX runtime filters out `null`/non-element children before they reach `parseChildren`, so the test for it exercises the happy path rather than the bad branch. The guard is there defensively for the real custom-element DOM path the issue describes, where those comment nodes do reach the function. Calling that out honestly rather than claiming a reproduction.

The fix adds a falsy check for `nodes` at the top of `parseChildren` and skips any child without `props` inside the map callback, matching the two changes suggested in the issue.

## How to test

1. `WC=true npx jest --config=tests/unit/jest.config.js ds-tabs.test.tsx`
2. All 16 tests pass, including the two added regression tests (`does not throw when rendered without children` and `ignores non-element child nodes such as comment nodes`).
3. To see the original crash, revert the change in `ds-tabs.tsx` and run the same command: the no-children test fails with `Cannot read properties of undefined (reading 'map')`.

## Checklist

- [x] Prefixed the PR title with [NO-TICKET] since this is unticketed external work.
- [x] Selected appropriate `Type` label for this PR.
- [x] Selected appropriate `Impacts`.
- [x] Selected appropriate release milestone.

- [x] Created or updated unit tests to cover the modified code.
- [x] Ran `WC=true` unit tests for `ds-tabs`; all pass. eslint and prettier are clean on the changed files (the two pre-existing `react/prop-types` warnings on the `Wrapper` component are unrelated to this change).


tamara-corbalt:
Jenkins: not a wolf

derek-cmsds (approved):
LGTM. Thank you very much for your contribution.

jack-ryan-nava-pbc (approved):
Thank you!

PR #4131, opened 2026-07-02
[CMSDS-2804] Improve keyboard navigation for `Tabs`

### LINKED ISSUE #3113: Improve tab keyboard interactions
_This is really a usability improvement or enhancement, less so a bug_ :) 

**Describe the bug**
The tab component's implementation of keyboard support is somewhat non-standard: 
* after you select a tab, you have to Tab key/arrow through all tabs before accessing the selected tab's content (expected: after selecting a tab, the next Tab press moves focus to that tab's content)
* you can use both arrows and the Tab key to navigate among tabs (expected: one or the other, preferably arrows so that Tab can be reserved for accessing content re: previous point)
* Space bar does not select a tab, but should

**To Reproduce**
1. Go to https://data.cms.gov/provider-data/dataset/0ba7-2cb0
2. Use the Tab key to access the tabs component (Data Table, Overview, API...)
3. Tab/arrow to traverse the tabs -- notice that arrows automatically select a tab, while the Tab key requires that you make a selection with Enter
4. Select the Overview tab
5. Use the Tab key again -- focus should move to "Download full dataset" button, but instead moves to the API tab

**Expected behavior**
Use a managed focus pattern for the tabs component, where the user can Tab key to the tab list and then use arrows to make a selection. The method for selection should probably be either auto-selection (as it does now with arrow keys) or manual selection (Enter or Space bar to choose a tab). Once a tab is selected, the next Tab key moves focus to the selected content; shift+Tab then moves focus back up to the tab list. Example: https://www.w3.org/WAI/ARIA/apg/patterns/tabs/examples/tabs-manual/

 - OS: Mac OS 14.5
 - Browser: Chrome
 - Version: 125.0.6422.142 (Official Build) (arm64)

**Additional context**
It would probably help to include interactive elements in the tab examples, too, to show how focus moves.

tamara-corbalt:
## Summary

- Updates the Tabs component to use a managed focus pattern for keyboard navigation.
- Removes non-selected tabs from the browser's tab order so arrow keys are the only method for navigating between tabs.
- Preserves the existing automatic activation behavior, where arrow navigation also selects the focused tab.
- Updates focus behavior so `Tab` moves from the selected tab into the corresponding `TabPanel`, and `Shift+Tab` returns focus back up to the selected tab.
- Adds tests and updates the storybook examples with focusable content to better demonstrate the expected behavior.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-2804)

- Closes #3113 

## How to test

1. Run `npm run test:unit` and `npm run test:unit:wc`.
2. Run Storybook and verify the keyboard behavior for both the `Tabs` and `ds-tabs` stories:
   - `Tab` moves focus to the selected tab.
   - Arrow keys navigate between tabs and automatically select them.
   - `Tab` moves focus from the selected tab into the selected `TabPanel`
   - `Shift+Tab` returns focus from the tab panel to the selected tab.

## Checklist


packages/design-system/src/components/Tabs/Tab.tsx
```diff
@@ -73,6 +73,7 @@ export const Tab = forwardRef((props: TabProps, ref: any) => {
       aria-selected={props.selected || defaultProps.selected}
       aria-controls={props.panelId}
       href={href}
+      tabIndex={props.selected ? 0 : -1}
```
tamara-corbalt, line 76: If not selected, remove from browser's tab order.

packages/design-system/src/components/Tabs/TabPanel.tsx
```diff
@@ -51,6 +51,7 @@ export const TabPanel = (props: TabPanelProps) => {
       className={classes}
       id={props.id}
       role="tabpanel"
+      tabIndex={0}
```
tamara-corbalt, line 54: Once this `TabPanel` is rendered, add it to the browser's tab order.

derek-cmsds:
I think this counts as a breaking change as we are drastically altering how the Tabs component is interacted with via keyboard. (Even if it is a better experience.) Product teams may need to add messaging or alert their users that navigating through tabs using the keyboard is different now.

derek-cmsds:
Can you rebase this off of the current `release/19.0` branch? There are changes here that we don't want merged in.

tamara-corbalt:
@derek-cmsds @jack-ryan-nava-pbc - Closing this in favor of https://github.com/CMSgov/design-system/pull/4139

PR #4140, opened 2026-07-08
[CMSDS-4053] Add `SvgIcon` dimension fallbacks

tamara-corbalt:
## Summary

- Fixes an issue where SVG icons (such as those used by the USA Banner) could render at unexpectedly large sizes when CSS is disabled or slow to load.
- Adds default `width` and `height` attributes to the `SvgIcon` component that match the dimensions defined by the `ds-c-icon` CSS class, providing a fallback when CSS is unavailable.
- Adds test coverage to verify the default `width` and `height` attributes and confirm that wrapper icon components (such as `AddIcon`) can override them.
- This updates a LOT of snapshots because the rendered icon markup now includes width and height attributes. The main change to inspect is in `SvgIcon.tsx` and the test coverage in `SvgIcon.test.tsx`

For more background, see: [Jira ticket](https://jira.cms.gov/browse/CMSDS-4053)

## How to test

1. `npm run test:unit`

## Checklist

### DIFF
--- a/packages/design-system/src/components/Accordion/__snapshots__/Accordion.test.tsx.snap
+++ b/packages/design-system/src/components/Accordion/__snapshots__/Accordion.test.tsx.snap
@@ -19,8 +19,10 @@ exports[`Accordion renders accordion 1`] = `
         <svg
           aria-hidden="true"
           class="ds-c-icon ds-c-icon--add ds-c-accordion__button-icon"
+          height="1.5em"
           id="1__icon"
           viewBox="3 3 18 18"
+          width="1.5em"
           xmlns="http://www.w3.org/2000/svg"
         >
           <path
@@ -52,8 +54,10 @@ exports[`Accordion renders accordion 1`] = `
         <svg
           aria-hidden="true"
           class="ds-c-icon ds-c-icon--add ds-c-accordion__button-icon"
+          height="1.5em"
           id="2__icon"
           viewBox="3 3 18 18"
+          width="1.5em"
           xmlns="http://www.w3.org/2000/svg"
         >
           <path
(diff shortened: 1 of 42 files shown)


derek-cmsds:
Do we consider this to be a breaking change? It adds two attributes to a component that did not previously exist, but in theory, it does not break any expectations of how icons should behave and should not break any overrides previously used. Is there something I'm missing?

tamara-corbalt:
> Do we consider this to be a breaking change? It adds two attributes to a component that did not previously exist, but in theory, it does not break any expectations of how icons should behave and should not break any overrides previously used. Is there something I'm missing?

No I don't think you're missing anything. I wasn't considering this a breaking change since it shouldn't alter the component's API or visual behavior. But it may require downstream snapshot updates because the rendered SVG markup now includes these new attributes.

derek-cmsds:
> No I don't think you're missing anything. I wasn't considering this a breaking change since it shouldn't alter the component's API or visual behavior. But it may require downstream snapshot updates because the rendered SVG markup now includes these new attributes.

You're probably not going to like this idea. If this isn't a breaking change, then I see no reason why it can't be part of the 18.1 release... which would mean rebasing this PR again back to `main` and changing the milestone. It also sounds like the Type is closer to "Type: Fixed" or "Type: Added" depending on whether we consider this a bug or not.

tamara-corbalt:
> > No I don't think you're missing anything. I wasn't considering this a breaking change since it shouldn't alter the component's API or visual behavior. But it may require downstream snapshot updates because the rendered SVG markup now includes these new attributes.
> 
> You're probably not going to like this idea. If this isn't a breaking change, then I see no reason why it can't be part of the 18.1 release... which would mean rebasing this PR again back to `main` and changing the milestone. It also sounds like the Type is closer to "Type: Fixed" or "Type: Added" depending on whether we consider this a bug or not.

Ok, I think this is a good point. Will rebase!

PR #4147, opened 2026-07-14
[CMSDS-4159] Update accessibility bug reporting

carolyn-crabbycrab:
## Summary

- Updated the accessibility bug reporting guidance on the **Contribution Guidance** and **Note Box** pages.
- Replaced outdated bug reporting language with the current approved guidance for both CMS internal and external users.
- Reviewed and updated the content using the Docsite Guidance Information Architecture to ensure consistency across the documentation site.
- Jira: CMSDS-4132 https://jira.cms.gov/browse/CMSDS-4159

## How to test

1. Navigate to the **Contribution Guidance** and **Note Box** pages.
2. Verify the updated bug reporting guidance appears on both pages.
3. Confirm the Confluence and GitHub links navigate to the correct reporting workflows.
4. Review the content for consistency with the rest of the documentation site.

## Checklist

- [x] Prefixed the PR title with the Jira ticket number as `[CMSDS-4159]`.
- [x] Selected appropriate `Type` label.
- [x] Selected appropriate `Impacts`.
- [x] Selected appropriate release milestone.

- [x] Checked for spelling and grammatical errors.
- [ ] Communicate the assigned milestone/release date with Design so they can communicate appropriately.


sophia-cmsds (approved):

packages/docs/content/getting-started/contribution-guidance.mdx
```diff
@@ -22,7 +22,7 @@ Use this guide to determine whether submitting a contribution proposal is the ri
...
 - The proposed design conflicts with accessibility standards, visual style, or content guidance set by the CMS Design System.
 - The design is not yet validated or has not shown repeat demand from other teams or users.
-- The request involves documentation updates, usage guidance, or content rather than a new visual or functional design element. These types of requests should be shared through one of our Slack channels or the bug-reporting workflow, which you can access in [#cms-design-system](https://cmsgov.slack.com/archives/CHH0381RD) or by opening a [GitHub issue](https://github.com/CMSgov/design-system/issues/new?template=propose-a-new-item-for-the-cms-design-system.md).
+- The request involves documentation updates, usage guidance, or content rather than a new visual or functional design element. Share these types of requests through one of our Slack channels, which you can access in [#cms-design-system](https://cmsgov.slack.com/archives/CHH0381RD). If you do not have access to CMS internal tools, use the [public GitHub issue template](https://github.com/CMSgov/design-system/issues/new?template=propose-a-new-item-for-the-cms-design-system.md) to submit a new issue.
```
tamara-corbalt: Looks good! Two nits:
- Disregard this if the content is already set in stone, but I found it a little odd to refer to "one of our Slack channels" and then link to only one.
- I wonder if the guidance for folks without access to CMS internal tools would be easier to scan if it were its own bullet, similar to the accessibility section on the Note Box page above.

tamara-corbalt (commented):
Looks good overall! Just two pieces of feedback -both disregardable if the content is already set in stone :)

packages/docs/content/components/note-box.mdx
```diff
@@ -139,7 +137,8 @@ What can be included in boxes:
 
 ### Reporting an accessibility bug
 
-If you encounter a bug with the accessibility of the alert component, [refer to our reporting confluence page](https://confluence.cms.gov/pages/viewpage.action?pageId=1138979323) and use CMSDS bug reporting workflow in our #cms-design-system-team slack channel.
+- If you have access to CMS internal tools, report accessibility issues through the [CMSDS bug reporting workflow in Confluence](https://confluence.cms.gov/spaces/HCDSG/pages/1138979323/Do+you+have+a+bug+to+report).
+- If you do not have access to CMS internal tools, report accessibility issues using the [public GitHub issue template](https://github.com/CMSgov/design-system/issues/new?template=propose-a-new-item-for-the-cms-design-system.md).
```
jack-ryan-nava-pbc: I don't think the new item is the correct template to use here. I think this should use the bug workflow: `https://github.com/CMSgov/design-system/issues/new?template=bug-report.md`

jack-ryan-nava-pbc (changes requested):
Agree with @tamara-corbalt's comment!

packages/docs/content/getting-started/contribution-guidance.mdx
carolyn-crabbycrab: Thanks Tamara! You actually validated my original thought. 😊 I initially had the guidance as its own bullet because I felt it was easier to scan but I changed it back to a sentence since that's how the original content was formatted. I agree that the bullet version is more readable so I'll switch it back.

For the Slack channel wording my thinking was that we do have multiple Slack channels but #cms-design-system is the relevant one for this particular scenario. That's why I kept "one of our Slack channels" while linking to the one people should use here. Hopefully that intent comes across but I'm happy to adjust the wording if you think it would be clearer.

packages/docs/content/components/note-box.mdx
carolyn-crabbycrab: Thanks Jack! I didn't know we had two different public GitHub issue templates. Good catch I'll update it to use the bug report template.

packages/docs/content/getting-started/contribution-guidance.mdx
tamara-corbalt: Oh ok, gotcha! I don't think it's a big deal either way. I think my reaction is just from a reader's perspective. If there's only one relevant Slack channel for this type of request, I'd probably just point directly to it. When I read "one of our Slack channels," I expected there to be a list of options.

carolyn-crabbycrab: Hmm I see. Keeping it simple is the better approach. I revised it to:

The request involves documentation updates, usage guidance, or content rather than a new visual or functional design element. Share these types of requests through the [#cms-design-system](https://cmsgov.slack.com/archives/CHH0381RD) Slack channel.

Thanks for the feedback!

tamara-corbalt (approved):
LGTM!

jack-ryan-nava-pbc (approved):

PR #4148, opened 2026-07-14
[DO NOT MERGE][TEST LEGACY VERSION BUMP] 7/14 Legacy version bump main

tamara-corbalt:
Adds the released legacy version(s) to versions.json on main.

PR #4149, opened 2026-07-14
[CMSDS-4152] Automate legacy version bump.

tamara-corbalt:
## Summary

- Previously, when releasing a supported version that was NOT the latest, we had to manually update `versions.json` on `main`.
- Adds an optional step to the release script to automate that process. The new workflow:
  * Creates a temporary branch from `main`.
  * Inserts each released package version into the appropriate array in `versions.json`, preserving descending semantic version order.
  * Commits the changes, pushes the branch, and opens a pull request against `main`.
  * Switches back to the original release branch to continue the release process.
  
**Note:** A good starting point for this review is `bumpLegacyVersionsOnMain` in `legacy-versions.ts`.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-4152)

## How to test
Since this change automates part of the release process, there isn't a great straightforward manual test. Here's how I verified it:
- Tested `insertVersionsIntoVersionsJson` independently with various version values to verify that package version arrays remained in descending semver order.
- Created a temporary test script to exercise the full workflow. For testing purposes, I replaced the call to `getPackageVersions()` with a constant object containing representative version values.
- Verified that the workflow created a temporary branch from main, updated versions.json, created the commit, pushed the branch, opened a pull request, and returned to the original branch.

Test pull request: https://github.com/CMSgov/design-system/pull/4148

## Checklist


scripts/legacy-versions.ts
```diff
@@ -0,0 +1,125 @@
...
+
+  for (const [themePackage, releasedVersion] of Object.entries(releasedVersions)) {
+    const existingVersions = versionsJson[themePackage];
+
+    if (!Array.isArray(existingVersions)) {
+      throw new Error(`Could not find a version array for "${themePackage}" in versions.json.`);
+    }
+
+    const updatedVersions = [...existingVersions, releasedVersion];
+    const uniqueVersions = [...new Set(updatedVersions)]; // In case of duplicates
```
tamara-corbalt, line 30: Duplicates probably aren't a huge risk here, but things happen, and this is just a small safeguard in case the same version gets processed more than once.

scripts/release.ts
```diff
@@ -2,12 +2,11 @@ import c from 'chalk';
 import { updateChildDSAndExamples } from './bump-child-deps-utils';
 import { confirm, select } from '@inquirer/prompts';
 import { hideBin } from 'yargs/helpers';
-import { sh, shI, verifyGhInstalled } from './utils';
+import { sh, shI, verifyGhInstalled, getCurrentBranch, REVIEWERS } from './utils';
 import { updateVersions } from './versions';
+import { bumpLegacyVersionsOnMain } from './legacy-versions';
 import yargs from 'yargs';
 
-const REVIEWERS = ['derek-cmsds', 'tamara-corbalt', 'jack-ryan-nava-pbc'];
```
tamara-corbalt, line 9: Moved a few things from this file into `utils.ts` so they can also be used by `legacy-versions.ts.`

scripts/legacy-versions.ts
```diff
@@ -0,0 +1,125 @@
...
+export const root = path.join(__dirname, '..');
+
+/**
+ * Inserts released versions into `versions.json`, preserving descending
+ * semantic version order for each design system.
+ *
+ * @param releasedVersions Mapping of theme package names to released versions.
+ */
+export async function insertVersionsIntoVersionsJson(releasedVersions: Record<string, string>) {
+  const versionsFileName = path.join(root, 'versions.json');
```
jack-ryan-nava-pbc, line 17: Lines 18, 20 and 21 might be replaced by simply importing the `versions.json` file at the top and working with the subsequent object: https://stackoverflow.com/questions/7163061/is-there-a-require-for-json-in-node-js

scripts/utils.ts
```diff
@@ -1,6 +1,7 @@
 import c from 'chalk';
 import { execSync, spawnSync, SpawnSyncOptionsWithBufferEncoding } from 'node:child_process';
 import { select } from '@inquirer/prompts';
+export const REVIEWERS = ['derek-cmsds', 'tamara-corbalt', 'jack-ryan-nava-pbc'];
```
derek-cmsds, line 4: It feels weird to have our GitHub usernames hard-coded into our repo. Seems like this should live in a config file or environment variable. Not a blocker though.

scripts/utils.ts
```diff
@@ -68,3 +69,12 @@ export async function chooseMilestone() {
 export function versionFromTag(tag: string): string {
   return tag.replace(/@cmsgov\/.*@(.*)$/, '$1');
 }
+
+export function getCurrentBranch() {
+  return sh('git rev-parse --abbrev-ref HEAD');
+}
+
+// TODO: Reuse this helper in other release workflows.
```
derek-cmsds, line 77: Nice.

scripts/release.ts
```diff
@@ -221,6 +216,7 @@ function printNextSteps() {
       await verifyNoUnstagedChanges();
       await bumpVersions();
       await bumpMain();
+      await bumpLegacyVersionsOnMain();
```
derek-cmsds: Why are we running both of these functions? Would it not be an either/or scenario?

derek-cmsds (commented):
I'm a bit confused by the approach. What is the scenario where we want to bump main and then bump main using this new legacy approach? Why run this function if we aren't releasing a legacy version?

scripts/release.ts
tamara-corbalt: That's a really good point. My first naive thought was that adding another opt-in step would be the simplest way to preserve the existing flow, but calling both functions makes it look like both apply to every release.

Maybe something like this instead?

```
const isLegacyRelease = await confirm({
  message: 'Are you releasing a version that is not the latest?',
});

if (isLegacyRelease) {
  await bumpLegacyVersionsOnMain();
} else {
  await bumpMain();
}
```

derek-cmsds: Let's talk about it more in today's pairing session, but I like this approach.

tamara-corbalt:
> I'm a bit confused by the approach. What is the scenario where we want to bump main and then bump main using this new legacy approach? Why run this function if we aren't releasing a legacy version?

Great point! I updated this in the [latest commit](https://github.com/CMSgov/design-system/pull/4149/changes/4964befbb05ce4d40a76afcf2c0dc7727c07f80a). It now uses if/else branching so we invoke either `bumpLegacyVersionsOnMain()` for legacy releases or `bumpMain()` for standard releases that are the latest, based on our discussion in eng huddle.

scripts/legacy-versions.ts
tamara-corbalt: I may be a little paranoid here, but I’m wary of importing `versions.json` once at module load time because the release flow updates that file at different points and switches branches along the way. We may also end up reordering parts of the release flow in the future, and I'm not sure how that would affect whether we're working with a fresh copy of `versions.json` or one that was loaded earlier.

derek-cmsds: Where else is the release flow updating `versions.json`?

derek-cmsds: I see: https://github.com/CMSgov/design-system/blob/f65b09c3822a52c4bf1d74bdf0f7ec65ab2f0b99/scripts/versions.ts#L39
Boy, this code is a mess.

tamara-corbalt: In [release.ts](https://github.com/CMSgov/design-system/blob/main/scripts/release.ts#L100C1-L100C53) we call [`updateVersions`](https://github.com/CMSgov/design-system/blob/main/scripts/versions.ts#L39-L67) which is in charge of updating `versions.json` on the release branch (not `main`).

tamara-corbalt: Yep. Added this ticket in jira: https://jira.cms.gov/browse/CMSDS-4181 
cc: @jack-ryan-nava-pbc

tamara-corbalt:
Based on our discussion in engineering huddle, we'd like to refactor the release script to simplify the workflow and improve its unit testability. I created a follow-up spike to capture that work: https://jira.cms.gov/browse/CMSDS-4181.
cc @jack-ryan-nava-pbc @derek-cmsds

derek-cmsds (approved):
Our whole release script needs to be rewritten to be more modular. But that is outside of the scope of this ticket. I'm approving this solution in order to make legacy versions work with the release script we have now. Good work.

scripts/legacy-versions.ts
jack-ryan-nava-pbc: Ok, cool. Thanks for the explanation @tamara-corbalt. Sounds like a synchronous and very explicit read of the json makes a lot of sense here.

jack-ryan-nava-pbc (approved):
Excellent work! I'm glad we're fixing this pipeline and building our understanding so we can make it better and/or replace it. Thanks @tamara-corbalt !

PR #4067, opened 2026-05-22
[CMSDS-4042] Complete DatePicker localization

jack-ryan-nava-pbc:
## Summary

- Months now localize to Spanish
- Day abbreviations localize to Spanish as well
- Masking localizes by switching between MM/DD/YYYY and DD/MM/YYYY
- [Ticket](https://jira.cms.gov/browse/CMSDS-4042)
- Closes #4044

## How to test

1. Run locally and open SingleInputDateField story in Storybook
2. Select the "With Picker" story
3. Change the localization to Spanish
Screenshot 2026-05-22 at 4 35 13 PM

4. Months are translated
5. Days are translated
6. Mask is localized

## Checklist

- [X] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [X] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [X] Selected appropriate `Impacts`, multiple can be selected.
- [X] Selected appropriate release milestone

PR #4155, opened 2026-07-17
[CMSDS-4165] Fix `SingleInputDateField` duplicate date selection with picker.

### LINKED ISSUE #4141: [Bug:] CalendarPicker stays open when a previously selected date is selected
### Describe the Bug

When a date is selected in the calendar picker that has already been selected the picker stays open and an error is logged to the console.

### To reproduce

1. Go to [the calendar picker](https://design.cms.gov/storybook/?path=/story/components-singleinputdatefield--with-picker&globals=theme:healthcare)
2. Select any date with the picker
3. Re-open the picker and select the same date
4. Note the picker stays open and an error is logged to the console

This is a video of it happening on a recent commit on main (note the error doesn't happen but the picker still stays open on a duplicate date selection)

https://github.com/user-attachments/assets/6669b375-f9f8-4d48-956d-16f27c8c244d

### Expected behavior

When a duplicate date is selected the date picker closes and there is no error to the console.

### Severity/Impact

Medium: Product teams can use workaround but still blocker

### Desktop Operating System

Mac OS

### Desktop Browser Name

Chrome

### Desktop Browser Version

_No response_

### Device

_No response_

### Mobile Operating System

_No response_

### Mobile Browser Name

_No response_

### Mobile Browser Version

_No response_

### Additional context

[Slack thread](https://cmsgov.slack.com/archives/C0111BVM1LZ/p1783523636031469)

### Codebase/Repository URL

_No response_

### Design System Version

17.2.1

_No response_

### Team Name

Accounts

tamara-corbalt:
## Summary
**Background**
- When a user selects a date in the calendar picker that is already selected, the `onSelect` handler is invoked with `undefined`. Then,`handlePickerChange` attempts to process this value, resulting in a TypeError when `computeDateValue` calls `getMonth()` on `undefined`. You can see this behavior in our [deployed storybook](https://design.cms.gov/storybook/?path=/story/components-singleinputdatefield--with-picker).
- In [pr-4067](https://github.com/CMSgov/design-system/pull/4067) (targeted for v18.1.0 and not yet released), we updated the `onSelect` handler to guard against undefined values ([link to line](https://github.com/CMSgov/design-system/blob/main/packages/design-system/src/components/DateField/SingleInputDateField.tsx#L305)). This prevents the console error that we see currently in prod, but it also means `handlePickerChange` never runs when the selected date is clicked again. Because `handlePickerChange` is responsible for closing the calendar dialog, the picker remains open after reselecting the same date.

**Fix**
- Enables the `required` prop on `CustomDayPicker`. This is a `react-day-picker` prop (`DayPickerSingleProps`) that makes date selection required, ensuring `onSelect` receives the selected `Date` instead of `undefined`.
- Moves the `undefined` check from the `onSelect` handler into `handlePickerChange`. If `handlePickerChange` is ever invoked with `undefined`, it now exits early after closing the calendar dialog. This shouldn't occur with `required` enabled, but it provides a fallback if `required` is removed in the future.
- Adds test coverage for `SingleInputDateField`

[Jira ticket](https://jira.cms.gov/browse/CMSDS-4165)

- Closes #4141 

## How to test
**Reproduce the bug**
1. Go to [the calendar picker](https://design.cms.gov/storybook/?path=/story/components-singleinputdatefield--with-picker&globals=theme:healthcare)
2. Select any date with the picker
3. Re-open the picker and select the same date
4. Note the picker stays open and an error is logged to the console

**Verify the fix**
- Pull down this branch and start storybook
- Navigate to [SingleInputDateField with picker](http://localhost:6006/?path=/story/components-singleinputdatefield--with-picker)  
- Repeat steps 2 & 3 (above).
- Picker should close and we shouldn't see any console errors.
- Repeat the same verification for the [web component](http://localhost:6006/?path=/story/web-components-ds-date-field--with-picker).
- Run unit tests: `npm run test:unit`

## Checklist

### DIFF
--- a/packages/design-system/src/components/DateField/SingleInputDateField.tsx
+++ b/packages/design-system/src/components/DateField/SingleInputDateField.tsx
@@ -199,7 +199,12 @@ const SingleInputDateField = (props: SingleInputDateFieldProps) => {
     }
   };
 
-  const handlePickerChange = (date: Date) => {
+  const handlePickerChange = (date?: Date) => {
+    if (!date) {
+      setPickerVisible(false);
+      inputRef.current?.focus();
+      return;
+    }
     const updatedValue = computeDateValue(date, lang);
     const maskedValue = dateMask(updatedValue);
     const validDate = getValidDate(updatedValue);
@@ -302,7 +307,8 @@ const SingleInputDateField = (props: SingleInputDateFieldProps) => {
         <div ref={dayPickerRef} role="dialog" onKeyDown={handleDayPickerKeyDown}>
           <CustomDayPicker
             selected={date}
-            onSelect={(day) => day && handlePickerChange(day)}
+            required
+            onSelect={handlePickerChange}
             defaultMonth={date ?? defaultMonth}
             {...{
               fromDate,
--- a/packages/design-system/src/components/DateField/CustomDayPicker.tsx
+++ b/packages/design-system/src/components/DateField/CustomDayPicker.tsx
@@ -17,6 +17,7 @@ type CustomDayPickerProps = Pick<
   DayPickerSingleProps,
   | 'selected'
   | 'onSelect'
+  | 'required'
   | 'defaultMonth'
   | 'fromDate'
   | 'fromMonth'
--- a/packages/design-system/src/components/DateField/SingleInputDateField.test.tsx
+++ b/packages/design-system/src/components/DateField/SingleInputDateField.test.tsx
@@ -345,28 +345,25 @@ describe('SingleInputDateField', function () {
       expect(onChange).toHaveBeenCalledWith('01/19/2000', '01/19/2000', expectedDate);
     });
 
-    it('does not deselect the date when the selected day is selected again', async () => {
+    it('does not deselect the selected date when it is selected again', async () => {
       const onChange = jest.fn();
-      const { user } = renderPicker({ onChange });
+      const { user } = renderPicker({
+        value: '01/19/2000',
+        onChange,
+      });
 
-      const calendarButton = screen.getByRole('button');
-      const selectedDay = () => screen.getByRole('gridcell', { name: /19th/ });
+      await user.click(screen.getByRole('button'));
 
-      await user.click(calendarButton);
-      await user.click(selectedDay());
+      // Reset the mock so this assertion only covers reselecting the date.
+      onChange.mockClear();
+
+      await user.click(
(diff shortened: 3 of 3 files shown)


packages/design-system/src/components/DateField/SingleInputDateField.tsx
```diff
@@ -302,7 +307,8 @@ const SingleInputDateField = (props: SingleInputDateFieldProps) => {
         <div ref={dayPickerRef} role="dialog" onKeyDown={handleDayPickerKeyDown}>
           <CustomDayPicker
             selected={date}
-            onSelect={(day) => day && handlePickerChange(day)}
+            required
```
jack-ryan-nava-pbc, line 310: Is this attribute necessary? Why did you add it?

jack-ryan-nava-pbc (commented):
Tests are passing and everything looks good, I'm just curious about the addition of the required attribute? I couldn't find it in the renderer DOM tree, so I wonder if it gets removed by the CaldendarPicker internals anyway?

tamara-corbalt:
> Tests are passing and everything looks good, I'm just curious about the addition of the required attribute? I couldn't find it in the renderer DOM tree, so I wonder if it gets removed by the CaldendarPicker internals anyway?

Good question! `required` is a react prop on react-day-picker's `DayPickerSingleProps`, not an HTML attribute, so I wouldn't expect it to appear in the DOM.

I probably should have explained this better in the PR description. In my approach, the primary fix is adding the `required` prop, with the early return in `handlePickerChange` acting as a defensive fallback in case `required` is inadvertently removed in the future. If I disable `required` now, the unit test I added fails. According to the `react-day-picker` [docs](https://daypicker.dev/v8/using-daypicker/selection-modes#required-selection), enabling `required` prevents the selected date from being deselected.

I think this is more consistent with our current API. If the user clicks the selected date again, we don't propagate a "no date selected" state back to the input. If we relied only on the guard in `handlePickerChange`, we'd silently deselect the date in the calendar without communicating this to user.

packages/design-system/src/components/DateField/SingleInputDateField.tsx
tamara-corbalt: Good question, left my reasoning [here](https://github.com/CMSgov/design-system/pull/4155#issuecomment-5006905161)

jack-ryan-nava-pbc (approved):
Nice! Thank you for the explanation!

PR #4158, opened 2026-07-21
[CMSDS-3937] Add Tab Iconography example to Storybook and Doc Site

jack-ryan-nava-pbc:
## Summary

- Our Tabs component can accept a React Node as a child, so that makes it easy to add complex markup like <svg> + <p> or whatever. 
- Most of the changes here are a new Storybook story and it's associated snapshot updates
- New documentation mostly copied from the [Button component with Icon section] (https://design.cms.gov/components/button/#adding-icons)
- There is a single styling change however. We previously had a rule that only styled SVGs if they were direct children of the tabs component. I changed it so the inline icon styling is applied to svgs anywhere in the child hierarhcy.
- [Ticket](https://jira.cms.gov/browse/CMSDS-3937)

## How to test

1. `npm run freshen`
2. `npm run storybook`
3. Make sure the stories display

## Checklist


packages/docs/content/components/tabs.mdx
```diff
@@ -14,12 +14,36 @@ healthcare:
...
 
 ## Examples
 
 <StorybookExample componentName="tabs" storyId="components-tabs--default" minHeight={400} />
 
+### Adding icons
+
+- When adding an inline SVG icon, it will appear in the same color as the tab text.
+- For the crispest icon rendering, the icon must have a square `viewBox` with values that are multiples of 8 (e.g., 24x24).
+- Use the margin utility class to add spacing between the icon and tab text.
```
tamara-corbalt: Maybe give an example of which utility class to use, or link to our utility classes?

packages/docs/content/components/tabs.mdx
```diff
@@ -70,6 +94,7 @@ When implemented correctly, screen readers announce:
 - Its `role` (e.g., "tab").
 - Its position within the set (e.g., "Tab 1 of 3").
 - Whether the tab is selected.
+- An element's name, or accessible name, is how it is identified by browsers and the [accessibility tree](https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree). Accessible names provide information for assistive technology, such as screen readers, and so icon tab must include an accessible name. If there is no text beside the icon then an `aria-label` or `aria-labelledby` attribute can be used to give the tab its accessible name.
```
tamara-corbalt: Can we simplify the first sentence? Something like: "An element's accessible name is how browsers and the [accessibility tree](https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree) identify it."

I think the last sentence can be simplified as well: "If there is no visible text beside the icon, use an `aria-label` or `aria-labelledby` attribute to provide one."

tamara-corbalt (commented):
Looks great to me overall! Left a few minor wording suggestions for the added documentation. 

Also, do we want to add a icon example for `ds-tabs`?

packages/design-system/src/styles/components/_Tabs.scss
```diff
@@ -120,7 +120,7 @@
   }
 
   // SVG icons
-  > svg {
```
jack-ryan-nava-pbc, line 123: This is one of the more important changes.

packages/docs/content/components/tabs.mdx
jack-ryan-nava-pbc: I tested and we actually need to expand a bit on this guidance. So I'm incorporating your changes and adding that we need to set `ariaHidden` to false on the icon component.

jack-ryan-nava-pbc:
> Looks great to me overall! Left a few minor wording suggestions for the added documentation.
> 
> Also, do we want to add a icon example for `ds-tabs`?

So, funny story about the web component version: we can only pass in strings as arguments. I can imagine a world where someone passes in a stringified svg icon, but I don't think I'm that person. If you strongly disagree, I can still do the example, but I think I want to pass for now.

tamara-corbalt:
> > Looks great to me overall! Left a few minor wording suggestions for the added documentation.
> > Also, do we want to add a icon example for `ds-tabs`?
> 
> So, funny story about the web component version: we can only pass in strings as arguments. I can imagine a world where someone passes in a stringified svg icon, but I don't think I'm that person. If you strongly disagree, I can still do the example, but I think I want to pass for now.

I support your decision to not become the "stringified SVG" person. 😄

packages/docs/content/components/tabs.mdx
```diff
@@ -70,6 +94,7 @@ When implemented correctly, screen readers announce:
 - Its `role` (e.g., "tab").
 - Its position within the set (e.g., "Tab 1 of 3").
 - Whether the tab is selected.
+- An accessible name is how browsers and the [accessibility tree](https://developer.mozilla.org/en-US/docs/Glossary/Accessibility_tree) identify it. Accessible names provide information for assistive technology, such as screen readers, and so icon tab must include an accessible name. If there is no text beside the icon then you must set the `ariaHidden` prop (or `aria-hidden` attribute) to `false`. Icons have built in title attributes which are the default accessible name, however you can set your own accessible name for the icon by passing in a `string` to the `description` prop on the icon component.
```
tamara-corbalt: I think the expanded guidance here looks great! One nit: the last sentence feels like it should be split into two:
``` 
Icons have built in title attributes which are the default accessible name. However, you can set your own accessible name for the icon by passing in a `string` to the `description` prop on the icon component.
```

jack-ryan-nava-pbc: Agreed! It is done.

tamara-corbalt (approved):
LGTM! Thanks for addressing my feedback 😸

PR #4165, opened 2026-07-27
[NO-TICKET] Fix failing `ds-choice` snapshot

tamara-corbalt:
## Summary

* PR #4158 included a few snapshot updates for components that use icons.
* It looks like the snapshot for `ds-choice` was generated with a different font, which was likely not intended and is causing the nightly VRTs to fail.

## How to test

`CI=true npm run test:browser -- --grep "Choice"`

## Checklist


derek-cmsds (approved):
Any idea why the failing snapshot had this different font?

tamara-corbalt:
> Any idea why the failing snapshot had this different font?

Not really sure. The only styling change introduced in that pr was the rule that applies styling to SVGs only when they're direct children of `Tabs`

jack-ryan-nava-pbc:
Yeah this is weird. I wonder if the font didn't load in time or something?

tamara-corbalt:
> Yeah this is weird. I wonder if the font didn't load in time or something?

Possibly. We've seen font-loading issues in CI before, but I'd expect to see more snapshot failures sprinkled across our suite if that were the case. This has been the only one failing for three nights.

PR #4198, opened 2026-08-13
[NO-TICKET] Prettier should ignore Astro files

derek-cmsds:
## Summary

- If you run npm run apply-example-templates right now, you'll see the following errors:
```shell
[error] No parser could be inferred for file: examples/astro-themes/src/components/WebComponentExamples.astro
[error] No parser could be inferred for file: examples/astro/src/pages/web-components.astro
```
- This is because we don't currently have an Astro plugin for Prettier installed in our repo. I tried installing one to see what would happen, and it wiped the files completely. We probably don't want that.
- The easiest way to get rid of these errors is for Prettier to ignore all Astro files by updating the `.prettierignore` file. This PR does that.

## How to test

1. Run `npm run apply-example-templates`
2. Verify the above errors no longer appear in the output.

## Checklist

## AI Usage

- [ ] USED_AI: I attest that I have read, understood, and take ownership of all AI-generated code in this PR.
- [x] NO_AI: I did not use AI.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,3 +1,4 @@
+**/*.astro
 **/node_modules/**
 **/dist/**
 tmp/**


.prettierignore
```diff
@@ -1,3 +1,4 @@
+**/*.astro
```
jack-ryan-nava-pbc, line 1: Yes, yes it should ignore astro files.

jack-ryan-nava-pbc (approved):
LGTM!!!!!!!!!!!!!!!

PR #4216, opened 2026-08-26
[CMSDS-4576] Resolve noImplicitAny errors in shipped source

derek-cmsds:
## Summary

Stage 4a of the staged `strict: true` rollout. Resolves every `noImplicitAny` error in
shipped source — `packages/design-system/src`, excluding tests and stories — **without
enabling the flag**. `noImplicitAny` itself is turned on in stage 4c, once tests, stories,
docs, and the child packages are also clean. `npm run type-check` stays at 0 errors
throughout.

- [CMSDS-4576](https://jira.cms.gov/browse/CMSDS-4576)
- Follows [CMSDS-4573](https://jira.cms.gov/browse/CMSDS-4573) — Enable `strictFunctionTypes` (#4215)

Repo-wide `noImplicitAny` errors: **396 → 328**. Shipped source: **53 → 0**.

## What changed

**Missing third-party types.** Every `TS7016` in the repo is resolved by installing
`@types/lodash`, `@types/prismjs`, `@types/lodash.sortby`, and `@types/url-join`, plus local
shims for `ev-emitter` and `react-use-flexsearch`, which publish no types anywhere. The
shims live in a new root `@types/` directory registered via `typeRoots`, mirroring the
existing `packages/docs/@types` naming.

They are deliberately **not** under a package's `src/`. `gulp-typescript` builds declarations
from its own glob rather than the tsconfig `include`, so a shim under `src/` would be
invisible to the declaration build — and anything under `src/` risks being emitted into
`dist`, which would ship a global `declare module` into consumer projects.

Note for anyone adding a shim later: because `typeRoots` names the directory explicitly, the
`@types` name is convention rather than a requirement, but the *layout inside* it is load
bearing. Each subdirectory must be a type package, and a scoped package has to be flattened
— `@types/scope__name/index.d.ts`, not `@types/@scope/name/index.d.ts`. Both shims here are
unscoped, so this doesn't bite yet.

**Event handler parameters.** Annotated from the attachment site, not from the props they
forward to. The public `onChange` / `onSelectAll` / `onClearAll` / `onEditClick` / `exit`
props are all declared `(...args: any[]) => any` and are left alone — those are explicit
anys on public signatures, out of scope here.

**Index lookups into untyped objects.** Accumulators and lookup tables got the shapes their
enclosing functions already declared: `maskDeliminatedRegex` is `Record<string, RegExp>`,
`maskOverlayContent` is `Partial<Record<MaskMask, string>>`, `tabsRef` is
`Record<string, HTMLAnchorElement>`, and the two `preactement` attribute accumulators say
`IProps`, which their functions already returned.

**`preactement/define.ts`** was the densest file at 9 errors. `TheCustomElement` declares
`implements CustomElement` but left every member untyped, so the class had no usable type
and `__root` tripped `TS7022` through its own initializer. The members now take their types
from the interface they already claim to implement.

**Two small correctness improvements** fell out of the work rather than being sought:

- `Tabs.isTabPanel` is now a real type predicate, which is what lets `getPanelChildren`'s
  `filter` match its declared return type. The old version returned `child && (...)`, so it
  could return a falsy non-boolean despite its `: boolean` annotation. Every caller used it
  in a boolean position, so this is not a behavior change.
- `TableCell`'s `let Component` was an evolving any, meaning the JSX spread onto it was
  never type-checked at all. It is now `TableCellComponent`.

## Deviations from the plan

**1. Error count.** Planned 57, actual 53 — 28 react/core and 20 web-components against a
planned 33 and 24. Same baseline drift stage 3 saw; the plan's table predates stage 3
landing.

**2. `@types/url-join` pinned to `^4`.** The plan said to install it; the current 5.x line
is a stub package with no declarations, and installing it fails the *entire* compile with
`TS2688` rather than resolving the import. `url-join` in this repo is 4.0.1, so the types
must match at `^4`.

**3. A `packages/docs` fix landed in this stage.** Installing `@types/prismjs` gave
`Prism.highlight` its real three-argument signature and turned three existing two-argument
call sites into `TS2554`, taking `type-check` from 0 to 3. `packages/docs` is nominally
stage 4c's scope, but this stage's bar is that `type-check` stays green with the flag off,
so the regression is repaired here rather than carried forward.

*Behavior note:* only one of the three call sites is reachable —
`ContentRenderer.tsx:53`, the code-block renderer. The other two are in
`helpers/syntaxHighlighting.ts`, which has no callers at all (see "Found but not fixed"
below).

For the live one: Prism assigns the third argument to `env.language` and passes it through
its `before-tokenize` / `after-tokenize` hooks. The call previously left it `undefined`, so
any language-specific hook was silently skipped; supplying it means those hooks now fire.
Nothing in this repo registers such a hook today, so no change in output is expected — a
glance at docs-site code blocks is enough to confirm.

**4. Seven commits, not one.** The plan calls for one commit per stage. This is split by
concern — dependency install, shims, prism fix, mockTime, react/core, web-components, and a
directory rename — which should make review easier. Squash on merge if you'd rather the
history match the plan.

**5. Released as v19, not 18.2.0.** The plan targeted 18.2.0, in line with stages 1–3. The
`useId` narrowing below is a breaking TypeScript-API change, so this PR and every later
stage in the rollout now target the **19.0.0** milestone. Stages 1–3 are already merged and
keep their 18.2.0 milestone, which records where that work landed. The knock-on for stages
4b–6: a public `.d.ts` narrowing is now an expected outcome rather than an exception to be
argued — each still gets its own PR-body section, but none of them blocks.

## Declaration output — 4 lines changed

Stages 1–4 were supposed to leave `dist/react-components/types` byte-identical. Four lines
moved, every one an `any` narrowing to the truth:

```diff
  IdleTimeout/utilities/mockTime.d.ts
- export function mockTime(timestamp: any): void;
+ export declare function mockTime(timestamp: number): void;

  TextField/useLabelMask.d.ts
-         defaultValue: any;
+         defaultValue: undefined;

  web-components/storybook.d.ts
- export declare function webComponentDecorator(Story: any, context: any): …
+ export declare function webComponentDecorator(Story: React.ComponentType, context: DecoratorContext): …

  utilities/useId.d.ts
- export default function useId(prefix?: string, providedId?: string): any;
+ export default function useId(prefix?: string, providedId?: string): string;
```

The first three are reachable only through the `./dist/*` deep path.

> [!CAUTION]
> **BREAKING CHANGE — `useId`'s public return type narrows from `any` to `string`.**
>
> `useId` is re-exported from `components/index.ts`, so this is part of the public API.
> Runtime behavior is unchanged; the TypeScript API is not. Consumer code that assigned the
> result somewhere non-`string` compiled before and will now fail to compile.
>
> It was not an intentional edit: it falls out of `@types/lodash` giving `uniqueId` a
> `string` return, which propagates to `useId`'s inferred type. It is correct on every path
> — `providedId ?? uniqueId(prefix)` is always a string.
>
> Per @tamara-corbalt's review, this is being treated as breaking rather than as an
> accepted `any` narrowing: **milestone retargeted 18.2.0 → 19.0.0**, and the `Type` label
> changed to `Type: Breaking` per the checklist rule that a breaking change carries that
> label alone. See deviation 5.

Stage 3 (#4215) set a precedent for accepting an `any` narrowing — `createSlotVNode(): VNode
→ VNode<any>` — on the grounds that preserving the byte-identical bar would have meant
asserting away the exact inaccuracy the flag caught. That precedent still holds for the
three deep-path lines above. It does **not** stretch to cover `useId`, which is reachable
from the package root.

## Found but not fixed

None of these is in scope for a `noImplicitAny` stage — each is either an *explicit* `any`
rather than an implicit one, or a packaging/dead-code question rather than a typing one.
All four are ticketed under the Tech Debt epic and none blocks this PR:

- **[CMSDS-4592](https://jira.cms.gov/browse/CMSDS-4592)** — Stop publishing the `mockTime` Jest helper
- **[CMSDS-4593](https://jira.cms.gov/browse/CMSDS-4593)** — Remove the dead `@mdx-js/react` type shim in `packages/docs`
- **[CMSDS-4594](https://jira.cms.gov/browse/CMSDS-4594)** — Give `Tab`'s `forwardRef` a real element type
- **[CMSDS-4595](https://jira.cms.gov/browse/CMSDS-4595)** — Remove the dead `syntaxHighlighting` helpers in `packages/docs`

1. **[CMSDS-4592](https://jira.cms.gov/browse/CMSDS-4592) — `mockTime` is a Jest-only helper that is being published.** `compileTypescriptDefs`
   runs with `allowJs: true` and `jsSrcGlob` does not exclude it, so it lands in
   `dist/react-components/{cjs,esm,types}` — and the emitted CJS calls the global `jest`,
   which would be a `ReferenceError` for any consumer who imported it. Its only caller is
   the colocated test. Moving it removes files from the published package, which is a
   packaging decision.
2. **[CMSDS-4594](https://jira.cms.gov/browse/CMSDS-4594) — `Tab` is declared `forwardRef((props: TabProps, ref: any) => …)`**, so its ref type
   erases to `{}` and `Tabs` needs a cast to store the anchor element (commented at the
   cast site). The fix is `forwardRef<HTMLAnchorElement, TabProps>`, but that changes
   `Tab.d.ts`, and the offending `any` is explicit rather than implicit.
3. **[CMSDS-4595](https://jira.cms.gov/browse/CMSDS-4595) — `packages/docs/src/helpers/syntaxHighlighting.ts` is dead code.** Neither
   `highlightHtmlSyntax` nor `highlightJsxSyntax` has a caller anywhere in `packages/docs`.
   The JSX one could not work if it did: the docs site imports `prismjs` bare, which
   registers only `markup`/`html`/`css`/`clike`/`javascript`, so `Prism.languages.jsx` is
   `undefined` and `Prism.highlight` throws `The language "jsx" has no grammar.` That throw
   is pre-existing and not introduced here — the old two-argument call threw the same way,
   with the less useful message `The language "undefined" has no grammar.` Since nothing
   calls the function, neither version has ever run. The fix is most likely deletion, so it
   is left for its own ticket.

   Worth stating plainly for review: **the two-line change to `syntaxHighlighting.ts` in
   this PR edits unreachable code.** It was still needed to get `type-check` back to 0 after
   `@types/prismjs` landed. The `ContentRenderer.tsx` change in the same commit is the one
   that affects rendered output.
4. **[CMSDS-4593](https://jira.cms.gov/browse/CMSDS-4593) — `packages/docs/@types/@mdx-js/react/index.d.ts` is dead.** A `typeRoots`
   entry expects scoped packages flattened as `mdx-js__react/`, so TypeScript looks for
   `@mdx-js/index.d.ts` and finds nothing — the shim never loads as a type root. It is also
   no longer needed: `@mdx-js/react` ships its own `index.d.ts`. Left alone here.

## How to test

1. `npm install` — four new `@types/*` devDependencies.
2. `npm run build`
3. `npm run type-check` — expect **0 errors**.
4. Turn the flag on repo-wide and confirm it is **not** yet clean — this is what stages 4b
   and 4c are for, and it is why the flag stays off in `tsconfig.json`:
   ```bash
   npx tsc --noEmit --noImplicitAny 2>&1 | grep -c 'error TS'
   ```
   Expect **328**. Every one should be in tests, stories, `packages/docs`, `.storybook`, or
   the child packages — none in shipped `packages/design-system/src`, which step 5 checks.
5. Turn the flag on and narrow to the area this stage covers — shipped source, excluding
   tests and stories. This is the stage's actual goal, and the default `type-check` does not
   cover it:
   ```bash
   npx tsc --noEmit --noImplicitAny 2>&1 \
     | grep '^packages/design-system/src' \
     | grep -vE '\.test\.|\.stories\.|__tests__|__mocks__'
   ```
   Expect **no output**.
6. `npm run test:unit`, `npm run test:unit:wc`
7. Diff `packages/design-system/dist/react-components/types` against a build of `main`; the
   only differences should be the four lines above.
8. Spot-check code blocks on the docs site — that exercises `ContentRenderer.tsx`, the only
   reachable Prism call site of the three this PR touches.

## Verification run

| Gate | Result |
| --- | --- |
| `npm run build` (all packages) | clean |
| `npm run type-check` | **0 errors** |
| `npm run test:unit` | 909 passed, 8 skipped, 195 snapshots |
| `npm run test:unit:wc` | 407 passed, 6 skipped, 71 suites |
| `npm run test:browser:smoke` | **804 passed, 0 failed** |
| `npm run lint` | 0 errors; warnings 93 → 89 |

Lint warnings dropped because typing the three web-component wrappers cleared four
`react/prop-types` warnings.

**`test:browser:smoke` ran as `--no-docker --ignore-snapshots`** — Docker was not available
on this machine. The a11y assertions executed; **VRT image comparison was skipped**, so CI
is the real check on visual regressions here.

**`test:unit:preact` was compared against a stashed baseline** rather than read on its own,
because several of the failing suites are files this PR edits. Clean tree: 8 suites /
14 tests / 6 snapshots failing. This branch: 7 / 14 / 6. Identical test and snapshot counts,
one fewer failing suite — `DateField/SingleInputDateField` fails at baseline and passed
here, which looks like flake rather than a fix. **No new failures.**

## Checklist

## AI Usage

- [ ] NO_AI: I did not use AI.

#### Type of assistance:

- [ ] Code generation
- [ ] Documentation
- [ ] Debugging
- [ ] Testing
- [x] Refactoring
- [ ] Other

#### AI System used:

- [ ] ChatGPT
- [x] Claude
- [ ] Gemini
- [ ] GitHub Copilot

#### Level of modification:

- [ ] As-is (AI generated all code in this submission)
- [ ] Used as inspiration

No test changes: this stage is type annotations on existing behavior, and the existing
suites are what prove the behavior is unchanged. No snapshots needed updating — 195 unit
snapshots and 71 web-component snapshots all passed as-is.

`test:browser:all` was **not** run in full; `test:browser:smoke` was, without Docker and
therefore without VRT comparison. Flagging that rather than checking the box silently.


derek-cmsds:
`useId` is interesting. We expose it publicly so that child themes can use it for their own components. See below:
https://github.com/CMSgov/design-system/blob/8093ca3d902ee1cd9379215013403c079eca537b/packages/ds-medicare-gov/src/components/Icons/RoundedStarEmptyIcon.tsx#L2
I highly doubt any of our client teams use it or even know it is there, as we don't document it on https://design.cms.gov/. But the change we make here _could_ potentially break a client team's build if they were calling `useId` and expecting it to return `any` instead of `string`. @jack-ryan-nava-pbc and @tamara-corbalt, would you consider this a breaking change?

packages/design-system/src/components/Tabs/Tabs.tsx
```diff
@@ -49,11 +49,15 @@ const RIGHT_ARROW = 'ArrowRight';
  * @param {React.Node} child - a React component
  * @return {Boolean} Is this a TabPanel component?
  */
-const isTabPanel = (child): boolean => {
+const isTabPanel = (child: React.ReactNode): child is React.ReactElement<TabPanelProps> => {
+  if (!isValidElement(child)) {
+    return false;
+  }
```
tamara-corbalt, line 55: 👍🏻

packages/design-system/src/components/Tabs/Tabs.tsx
```diff
@@ -214,7 +218,9 @@ export const Tabs = (props: TabsProps) => {
           onKeyDown={handleTabKeyDown}
           panelId={panel.props.id}
           ref={(tab) => {
-            tabsRef.current[panel.props.id] = tab;
+            // `Tab` is declared with an untyped `ref`, so this callback's parameter
+            // widens to `{}`. It always receives the `<a>` that `Tab` renders.
+            tabsRef.current[panel.props.id] = tab as HTMLAnchorElement;
```
tamara-corbalt: Can callback refs receive `null` on dismount?

tamara-corbalt (commented):
Loos great! Thanks for all the helpful comments. It helped a lot!
Left one question.

packages/design-system/src/components/Tabs/Tabs.tsx
derek-cmsds: From https://react.dev/learn/referencing-values-with-refs#refs-and-the-dom:
> When you pass a ref to a ref attribute in JSX, like `<div ref={myRef}>`, React will put the corresponding DOM element into myRef.current. Once the element is removed from the DOM, React will update myRef.current to be null.

As far as I can tell, if the `Tab` were to be removed from the DOM, the ref associated with the `ref` attribute would be updated to `null`, but the value of `tabsRef` would remain the same. But React doesn't dismount a component on re-render, so I don't believe this situation would arise in normal usage. We do not allow a `Tab` to be removed dynamically from a `Tabs` component.

derek-cmsds: **This was written by Claude:**

Yes — good catch, and my comment was wrong on two counts.

React calls a callback ref with `null` on detach, and only skips that if the callback returns a cleanup function (React 19); this one has a block body, so the `null` call still happens. Two situations:

1. **Unmount** — a `TabPanel` removed from `children` (or `Tabs` itself unmounting) leaves `tabsRef.current[id]` set to `null` permanently, and `handleSelectedTabChange` dereferences it unguarded.
2. **Every re-render** — the inline arrow is a new function each render, so React calls the old callback with `null` and then the new one with the element. Both happen in the same commit, so nothing observes the gap.

The second is harmless; the first means the `as HTMLAnchorElement` cast was laundering a `null` past the compiler.

My claim that it "always receives the `<a>`" was also inaccurate — `Tab` renders a `<span aria-disabled="true">` when `disabled`, so a disabled tab stores an `HTMLSpanElement`. Unreachable today only because arrow-key navigation skips disabled tabs.

Updated to drop the entry on detach instead, which keeps the non-nullable `Record` honest and avoids storing stale nodes, with no change to any currently reachable runtime path:

```tsx
ref={(tab) => {
  // `Tab` declares an untyped `ref`, so this callback's parameter widens
  // to `{}`. React passes the rendered element on attach and `null` on
  // detach, so drop the entry rather than storing a `null` under it.
  if (tab) {
    tabsRef.current[panel.props.id] = tab as HTMLAnchorElement;
  } else {
    delete tabsRef.current[panel.props.id];
  }
}}
```

Widening to `Record<string, HTMLAnchorElement | null>` and guarding the `.focus()`/`.href` calls would be the stricter fix, but that changes behavior (silent no-op instead of a throw) beyond this PR's scope — happy to open a follow-up ticket if you'd prefer that.

derek-cmsds: **This is Derek**

I went ahead and pushed up the new commit. It should handle the `null` case.

packages/design-system/src/components/Tabs/Tabs.tsx
```diff
@@ -218,9 +218,14 @@ export const Tabs = (props: TabsProps) => {
...
-            // widens to `{}`. It always receives the `<a>` that `Tab` renders.
-            tabsRef.current[panel.props.id] = tab as HTMLAnchorElement;
+            // `Tab` declares an untyped `ref`, so this callback's parameter widens
+            // to `{}`. React passes the rendered element on attach and `null` on
+            // detach, so drop the entry rather than storing a `null` under it.
+            if (tab) {
+              tabsRef.current[panel.props.id] = tab as HTMLAnchorElement;
+            } else {
+              delete tabsRef.current[panel.props.id];
+            }
```
tamara-corbalt, line 228: Nice!

tamara-corbalt:
> ```diff
>   utilities/useId.d.ts
> - export default function useId(prefix?: string, providedId?: string): any;
> + export default function useId(prefix?: string, providedId?: string): string;
> ```

I think this should be considered a breaking change. The runtime behavior hasn’t changed, but the typescript API has, so it’s probably something we should flag.

derek-cmsds:
> > ```diff
> >   utilities/useId.d.ts
> > - export default function useId(prefix?: string, providedId?: string): any;
> > + export default function useId(prefix?: string, providedId?: string): string;
> > ```
> 
> I think this should be considered a breaking change. The runtime behavior hasn’t changed, but the typescript API has, so it’s probably something we should flag.

Makes sense. That means this PR and all PRs in the TypeScript `strict: true` rollout moving forward (as well as the Storybook v10 ticket) need to move to the v19 milestone branch. I'll update accordingly.

derek-cmsds:
Agreed — a narrowed public type is a compile-time break for consumers even with runtime behavior untouched. `useId` is re-exported from `components/index.ts`, so it is reachable from the package root, which puts it outside the `any`-narrowing precedent #4215 set for deep-path-only declarations.

Actioned on this PR:

- **Milestone 18.2.0 → 19.0.0.**
- **`Type: Internal` → `Type: Breaking`**, per the checklist rule that a breaking change carries that label alone.
- The `useId` note in the PR body is now a `[!CAUTION]` BREAKING CHANGE callout rather than a request for sign-off, and a new deviation 5 records the retarget.

The rest of the strict-mode chain (CMSDS-4571, 4575, 4570, 4568, 4569, 4572) will target 19.0.0 as well. Stages 1–3, already merged as #4210/#4214/#4215, keep their 18.2.0 milestone since that records where the work landed — say the word if the release plan for 18.2.0 changes and they should move.

— posted by Claude on behalf of @derek-cmsds

derek-cmsds:
As a breaking change, this PR has been rebased onto the v19 milestone branch. It should be ready for review once more.

tsconfig.json
```diff
@@ -8,6 +8,7 @@
     "allowSyntheticDefaultImports": true,
     "esModuleInterop": true,
     "skipLibCheck": true,
+    "typeRoots": ["./@types", "./node_modules/@types"],
```
jack-ryan-nava-pbc, line 11: Just checking but we're not missing any type declarations anywhere because our repo is a monorepo and Node hoists all installed packages to the top level node_modules directory, right?

packages/docs/src/helpers/syntaxHighlighting.ts
```diff
@@ -9,7 +9,7 @@ export function highlightHtmlSyntax(code) {
     parser: 'html',
```
jack-ryan-nava-pbc, line 9: Should we just delete this file? Maybe in a later PR, but still.

packages/design-system/src/components/web-components/preactement/define.ts
```diff
@@ -159,7 +160,7 @@ function proxyEvents(props, events: IOptions['events'], CustomElement) {
     // Convert the event name to a kebab-case format and replace 'on' with 'ds'
     // This prevents the custom events from conflicting with the native events
     const customName = kebabCaseIt(name.replace('on', 'ds'));
-    let existingCallback = () => null;
+    let existingCallback: () => void = () => null;
```
jack-ryan-nava-pbc, line 163: I've not seen this before. What's going on with ` () => void = () => null`?

packages/design-system/src/components/TextField/Mask.tsx
```diff
@@ -19,7 +19,8 @@ const maskPattern = {
   currency: '[0-9.,\\-]*',
 };
 
-const maskOverlayContent = {
+// Only some masks render an overlay, so every lookup here is guarded.
```
jack-ryan-nava-pbc: What is meant by "every lookup here is guarded"? My understanding of Partial is that all of the fields get set to optional... which doesn't feel like a guard? Source: https://www.typescriptlang.org/docs/handbook/utility-types.html#partialtype

jack-ryan-nava-pbc (commented):
I have a few questions I'd like answered before fully approving. These are minor. This is great work though, and very close.

jack-ryan-nava-pbc:
> > > ```diff
> > >   utilities/useId.d.ts
> > > - export default function useId(prefix?: string, providedId?: string): any;
> > > + export default function useId(prefix?: string, providedId?: string): string;
> > > ```
> > 
> > 
> > I think this should be considered a breaking change. The runtime behavior hasn’t changed, but the typescript API has, so it’s probably something we should flag.
> 
> Makes sense. That means this PR and all PRs in the TypeScript `strict: true` rollout moving forward (as well as the Storybook v10 ticket) need to move to the v19 milestone branch. I'll update accordingly.

@derek-cmsds @tamara-corbalt Maybe I'm confused, but I'm trying to think through the consequences of narrowing the return type of this function.

Let's say I have a variable, and I've already been assigning the output of `useId` to that variable, what actually happens when the return type goes from any -> string? In theory, the type of the variable should be set to the return type of the function, which would already be any. If it's inferred at compile time, that's also ok, because now it'll just narrow that var to a string, and if `useId` was already returning strings only, then that inference was already happening. 

So I'm just wondering if this will actually have an impact?

derek-cmsds:
> @derek-cmsds @tamara-corbalt Maybe I'm confused, but I'm trying to think through the consequences of narrowing the return type of this function.
> 
> Let's say I have a variable, and I've already been assigning the output of `useId` to that variable, what actually happens when the return type goes from any -> string? In theory, the type of the variable should be set to the return type of the function, which would already be any. If it's inferred at compile time, that's also ok, because now it'll just narrow that var to a string, and if `useId` was already returning strings only, then that inference was already happening.
> 
> So I'm just wondering if this will actually have an impact?

Let's say for some reason, one of our clients wrote this code: `const id: number => useId();`

Before this change, that code compiles and TypeScript is happy. After this change, TypeScript cries foul and their code no longer compiles, all because they updated to the latest Design System version.

Do I think any of our clients have code like that anywhere? No, almost definitely not.... But they _could_, and therefore I think we have to call this a breaking change.

tsconfig.json
derek-cmsds: These are the only places in our repo we are currently declaring global types. Imported packages also still follow normal module resolution and walk up directories regardless of these settings. The only time this would ever become a problem is if a package had a version mis-match with the rest of the repo that forced it to install deeper in the tree. But that's not happening right now and we'd probably want to resolve it if it did.

packages/docs/src/helpers/syntaxHighlighting.ts
derek-cmsds: Yes, and I created https://jira.cms.gov/browse/CMSDS-4595 to do just that.

jack-ryan-nava-pbc:
@derek-cmsds do you mean `const id = (): number => useId()` instead of `const id: number => useId();`?

packages/design-system/src/components/web-components/preactement/define.ts
derek-cmsds: `() => void` is the type of the function `() => null`. It says the function does not return anything, as opposed to returning strictly `null`. This is correct because `existingCallback` gets reassigned later on in the code, so calling it a function that returns `null` would not line up with how it actually gets used.

derek-cmsds: The return type is `void` instead of `any` because we do not care about the return value. We execute the callback on line 187 but do nothing with what it returns.

derek-cmsds:
> @derek-cmsds do you mean `const id = (): number => useId()` instead of `const id: number => useId();`?

`const id: number = useId();`
If you execute `useId` and assign it's return value to a variable typed as a number, that would compile fine in previous versions of the Design System but would return an type error after this change.
Apparently, I have coding dyslexia. :-P

packages/design-system/src/components/TextField/Mask.tsx
derek-cmsds: This was in regards to the `maskOverlay` around line 158, but it's poorly worded. I'll update the comment.

packages/design-system/src/components/web-components/preactement/define.ts
jack-ryan-nava-pbc: OK, so this is just me not being able to read typescript code. Cool, thanks for bearing with me!

jack-ryan-nava-pbc:
> > @derek-cmsds do you mean `const id = (): number => useId()` instead of `const id: number => useId();`?
> 
> `const id: number = useId();` If you execute `useId` and assign it's return value to a variable typed as a number, that would compile fine in previous versions of the Design System but would return an type error after this change. Apparently, I have coding dyslexia. :-P

You and me both

jack-ryan-nava-pbc (approved):
Thanks for walking me through my questions. I ran the steps locally and didn't encounter any errors.

PR #4213, opened 2026-08-21
[CMSDS-3959] Generates a list of `llms.txt` urls

tamara-corbalt:
## Summary
Generates a list of hosted `llms.txt` URLs as part of the docs build. This provides an easy way to produce an up-to-date list of assets for usage tracking as the documentation IA changes.

The generated list is stored in `packages/docs/build-artifacts/ ` and is not tracked by git.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-3959)

## How to test

- `npm run build:docs`
- Inspect `packages/docs/build-artifacts/llms-urls.txt`

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] Selected appropriate release milestone

## AI Usage

- [x] USED_AI: I attest that I have read, understood, and take ownership of all AI-generated code in this PR.
- [ ] NO_AI: I did not use AI.

#### Type of assistance:

- [ ] Code generation
- [ ] Documentation
- [ ] Debugging
- [ ] Testing
- [ ] Refactoring
- [x] Other

#### AI System used:

- [x] ChatGPT
- [ ] Claude
- [ ] Gemini
- [ ] GitHub Copilot

#### Level of modification:

- [ ] As-is (AI generated all code in this submission)
- [ ] Modified (You made some changes to the AI's generation)
- [x] Used as inspiration

PR #4206, opened 2026-08-18
[CMSDS-4564] Modernize root tsconfig compiler options

derek-cmsds:
## Summary

Modernizes the compiler options in the root `tsconfig.json`. This is groundwork for enabling
`strict` mode — it changes no source code and no published output.

- [CMSDS-4564](https://jira.cms.gov/browse/CMSDS-4564)

Four changes, one per commit:

| Change | Why |
| --- | --- |
| Removed `baseUrl` | Deprecated in TS 5.x; nothing in the repo relied on it for resolution |
| `target`: `es2015` → `es2020` | `es2015` predates every browser and Node version the DS supports; Babel, not `tsc`, controls what actually ships |
| `module`: `esnext` → `nodenext` | Type-checks the repo the way Node actually resolves it, so `require`/`import` interop errors surface here instead of in consumer builds |
| Dropped the explicit `moduleResolution: bundler` | `module: nodenext` implies `moduleResolution: nodenext`; keeping `bundler` alongside it is contradictory |

`nodenext` rather than `node18` (the intermediate step in commit `eefc600c2`): `node18` pins
resolution to a frozen Node 18 profile, which reported ESM-interop errors on
correctly-written imports. `nodenext` tracks current Node semantics and reports 0 errors.

## How to test

1. `npm run build`
2. `npm run type-check` — expect 0 errors
3. `npm run test`
4. Confirm `packages/design-system/dist/react-components/types` is unchanged from a `main`
   build. `gulpfile.js` feeds this `tsconfig.json` into `ts.createProject` for `.d.ts` emit,
   so a compiler-option change here can alter published type declarations. This one should
   not.

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking` — `Type: Internal`
- [x] Selected appropriate `Impacts`, multiple can be selected. — `Impacts: Core`
- [x] Selected appropriate release milestone — 18.2.0

## AI Usage

- [x] USED_AI: I attest that I have read, understood, and take ownership of all AI-generated code in this PR.

#### Type of assistance:

- [x] Debugging
- [x] Refactoring

#### AI System used:

- [x] Claude

#### Level of modification:

- [x] Modified (You made some changes to the AI's generation)

### If this is a change to code:

[...]

PR #4205, opened 2026-08-18
[CMSDS-3454] Remove `$extensions` field from all JSON tokens

tamara-corbalt:
## Summary
* Removes:
  - `$extensions` field from all JSON tokens and mock token data.
  - `$extensions` from the token type definition.

[Jira ticket](https://jira.cms.gov/browse/CMSDS-3454)

## How to test
* Run `npm run freshen` and verify that all packages build successfully.
* Run the tests related to token parsing and translation:
  * `npx jest packages/design-system-tokens/src/css/translate.test.ts`
  * `npx jest packages/design-system-tokens/src/lib/readTokenFiles.test.ts`
* Verify all other unit and visual regression tests pass.
* Run storybook and the docs site and verify they build and render as expected.

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] Selected appropriate release milestone

## AI Usage

- [x] USED_AI: I attest that I have read, understood, and take ownership of all AI-generated code in this PR.
- [ ] NO_AI: I did not use AI.

#### Type of assistance:

- [x] Code generation (temporary node script to recursively remove `$extensions` fields from JSON token files)
- [ ] Documentation
- [ ] Debugging
- [ ] Testing
- [ ] Refactoring
- [ ] Other

#### AI System used:

- [x] ChatGPT
- [ ] Claude
- [ ] Gemini
- [ ] GitHub Copilot

#### Level of modification:

- [ ] As-is (AI generated all code in this submission)
- [x] Modified (You made some changes to the AI's generation)
- [ ] Used as inspiration

PR #4201, opened 2026-08-14
[NO-TICKET] Revert `Tabs` VRT snapshots to match `main`

tamara-corbalt:
## Summary

* Since the changes from PRs #4091 and #4144 are no longer in `main`, reverts the `Tabs` and `ds-tab` VRT snapshots to reflect the current versions of those components in `main`.

## How to test

1. `npm run test:browser`

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/CMSDS/) as `[CMSDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] Selected appropriate release milestone

## AI Usage

- [ ] USED_AI: I attest that I have read, understood, and take ownership of all AI-generated code in this PR.
- [x] NO_AI: I did not use AI.

PR #4217, opened 2026-08-27
[NO-TICKET] Merge `main` into `milestone-19-0-0`

derek-cmsds:
> [!IMPORTANT]
> **Do not squash this PR — it should land as a merge commit.**
>
> Squashing flattens the two-parent merge (`bde2b96b2` + `8093ca3d9`) into a single commit,
> so git no longer records that `main` has been merged into `milestone-19-0-0`. The next
> sync would then replay all 11 commits and raise the same `tabs` token conflict again.
>
> Note that repo settings currently permit **squash only** — `allow_merge_commit: false`,
> `allow_rebase_merge: false` — which is why the previous sync (#4144) has a single parent.
> Landing this as a real merge needs an admin to enable "Allow merge commits" for it, or to
> push the merge commit directly with branch protection briefly lifted. If it does get
> squashed, that is workable, but expect this conflict to return on the next sync.

## Summary

Brings `milestone-19-0-0` up to date with `main`. The two branches had diverged at
`3a57c9cb0` (2026-08-13): 11 commits on `main`, 7 on the milestone branch.

This is a plumbing merge with no new work of its own — it unblocks stage 4a of the
`strict: true` rollout ([CMSDS-4576](https://jira.cms.gov/browse/CMSDS-4576), #4216), which
was retargeted from 18.2.0 to 19.0.0 after review found the `useId` return-type narrowing to
be a breaking change. That PR builds on stages 1–3, which exist only on `main`, so it cannot
rebase onto the milestone branch until those commits are here.

Incoming from `main` (11): strict-mode stages 1–3 (#4210, #4214, #4215), the root tsconfig
modernization (#4206), `$extensions` removal from token JSON (#4205), `llms.txt` URL
generation (#4213), the Figma sync pipeline removal (#4199), and four smaller fixes.

### One conflict, resolved deliberately

Six hunks across `Theme.core.json`, `Theme.healthcare.json`, and `Theme.medicare.json`, all
inside the `tabs` block. The milestone branch changed two tab token **values** in "Update
tabs background and hover styling" (#4091); `main` removed every `$extensions` block in
#4205 while keeping the old values. Git could not see that the correct result is the new
value *with* the `$extensions` removal applied.

Resolved as milestone value + no `$extensions`, for all three themes:

```
component.tabs.background-color--hover:  {theme.color.primary-lightest}
component.tabs.color--hover:             {theme.color.primary-darker}
```

Verified two ways. Parsing every `$value` in the three files: **0** differences against
`origin/milestone-19-0-0` — no milestone work was dropped — and exactly **2** against
`origin/main`, the #4091 restyle. `$extensions` occurrences remaining: **0**. Then
`npm run build:tokens` was run on both this merge and a clean `origin/main` checkout and the
emitted output diffed; every difference is those same two tokens, across `css-vars` and
`scss` for all three themes:

```diff
---tabs__background-color--hover: var(--color-background);
+--tabs__background-color--hover: var(--color-primary-lightest);
---tabs__color--hover: var(--color-primary);
+--tabs__color--hover: var(--color-primary-darker);
```

### Four Tabs VRT references, caught and corrected

An earlier revision of this branch silently took `main`'s copies of four Tabs `--default`
snapshot references:

```
tests/browser/snapshots/stories/components-tabs--default-cmsgov--firefox.png
tests/browser/snapshots/stories/components-tabs--default-medicare--webkit.png
tests/browser/snapshots/stories/web-components-ds-tabs--default-healthcare--firefox.png
tests/browser/snapshots/stories/web-components-ds-tabs--default-healthcare--webkit.png
```

Git had no reason to flag it. Those four changed only on `main` since the merge base
`3a57c9cb0` — in #4201, which reverted them to match `main`'s rendering — and never on the
milestone side, so they resolved cleanly to `main`'s version with no conflict shown. The other
28 differing Tabs references changed only on the milestone side, via #4144, and were kept. The
result would have been 28 milestone-style references sitting beside 4 `main`-style ones.

Root cause: `bf422a070` (#4145, 2026-07-13) is on **both** branches and set those four to
release/19.0's Tabs rendering, `main` included — where that styling does not exist. #4201
undid it on `main`, correct for `main` and wrong to carry onto this branch.

The earlier claim in this description that `--default` stories are unaffected by two
hover-only tokens, and that no VRT failure was expected, was wrong. The difference is not
color: the tab panel content and the button beneath it sit at a different vertical offset.

Confirmed in Docker with `npm run test:browser -- --grep "Tabs"`:

| Branch state | Result |
| --- | --- |
| Merge as first pushed (`968b46b3f`) | **4 of 200 failed** — exactly the four files above |
| Milestone references restored (`6f91f6e75`) | **200 of 200 passed** |

`6f91f6e75` restores the `milestone-19-0-0` copies, so this branch now changes **zero** VRT
reference images relative to `milestone-19-0-0`.

Caught by Tamara in review.

## How to test

1. `npm install` — `package.json` differs between the two branches.
2. `npm run build:tokens`, then confirm the only tab-token differences against `main` are
   `--tabs__background-color--hover` and `--tabs__color--hover` in all three themes.
3. `npm run build && npm run type-check` — expect **0 errors**.
4. `npm run test:unit`
5. `npm run test:browser -- --grep "Tabs"` in Docker — expect **200 of 200 passing**.
6. Let CI run the rest of the browser suite.

## Checklist

`Type: Internal` — the merge itself introduces no new behavior. The breaking changes it
carries are labeled on their own PRs.

## AI Usage

- [ ] NO_AI: I did not use AI.

#### Type of assistance:

- [x] Other — conflict resolution and verification

#### AI System used:

- [x] Claude

#### Level of modification:

- [x] As-is (AI generated all code in this submission)

No VRT reference image differs between this branch and `milestone-19-0-0`, so there is no
visual change for design to review. The four discussed above were restored to the milestone
branch's existing images rather than regenerated.

No test changes: a branch-sync merge adds no behavior to cover.

- [ ] Verified that running both `npm run test:unit` and `npm run test:browser:all` were each successful
- [x] If necessary, updated unit-test snapshots and browser-test snapshots

Partially run locally. `npm install` and `npm run build` were re-run against this merge, then
the Tabs browser suite in Docker: `npm run test:browser -- --grep "Tabs"`, 200 of 200 passing.
The full `npm run test:browser:all` and `npm run test:unit` were not run here, so the box stays
unchecked; CI covers both.


tamara-corbalt (commented):
Everything looks good except, I'm a little confused about the Tab-related snapshots. By bringing these snapshot changes into `milestone-19-0-0`, we’d overwrite the snapshots that currently match the Tab stories on this branch. We’d then have to update them again later when the milestone branch is merged back into `main`.

derek-cmsds:
> Everything looks good except, I'm a little confused about the Tab-related snapshots. By bringing these snapshot changes into `milestone-19-0-0`, we’d overwrite the snapshots that currently match the Tab stories on this branch. We’d then have to update them again later when the milestone branch is merged back into `main`.

Good catch! Looks like this was a side effect of us accidentally putting breaking changes into `main` when we thought we were going straight to v19. (You might've even brought up these snapshots during that period.) We want to keep the updated snapshots from the v19 milestone branch. Should now be updated.

tamara-corbalt (approved):
LGTM! Approving ahead of resolved CI

PR #4232, opened 2026-09-09
[CMSDS-4571] Resolve noImplicitAny errors in design-system tests and stories

derek-cmsds:
## Summary

Stage 4b of the staged `strict: true` rollout. Resolves every `noImplicitAny` error in `packages/design-system`'s tests and stories — **without enabling the flag**, which happens in [CMSDS-4575](https://jira.cms.gov/browse/CMSDS-4575) once docs and the child packages are also clean. `npm run type-check` stays at 0 errors throughout, and no shipped code changed, so `dist/react-components/types` is byte-identical.

- [CMSDS-4571](https://jira.cms.gov/browse/CMSDS-4571)
- Follows [CMSDS-4576](https://jira.cms.gov/browse/CMSDS-4576) — Resolve `noImplicitAny` errors in shipped source (#4216)

Repo-wide `noImplicitAny` errors: **326 → 95**. Tests and stories: **231 → 0**. The remaining 95 belong to [CMSDS-4575](https://jira.cms.gov/browse/CMSDS-4575) (`packages/docs`, `.storybook`, the child packages).

Three changes are not pure annotations, each called out in its commit: the multi-header `Table` story's cell `key` looked rows up by `displayName` where they are keyed by `propName`, so every key interpolated `undefined`; `web-components/__tests__/rendering.tsx`'s `rerenderTest` could not chain, because `rerender` returns void and it passed that on as the next render result; and `ds-badge`'s default children fallback is a string rather than a fragment. Two findings were filed rather than fixed here — [CMSDS-4613](https://jira.cms.gov/browse/CMSDS-4613) (a `Table` test that asserts nothing) and [CMSDS-4614](https://jira.cms.gov/browse/CMSDS-4614) (`attributes` arrays missing `as const`, which leaves 27 wrappers' JSX attribute types unenforced). Seven story files carry a TODO pointing at that ticket.

## How to test

1. `npm run type-check` — 0 errors.
2. `npx tsc --noEmit --allowImportingTsExtensions --noImplicitAny | grep '^packages/design-system/'` — no output.
3. `npm run test:unit` and `npm run test:unit:wc` — green. `npm run test:unit:preact` fails 8 suites / 15 tests / 6 snapshots, identical to this branch's base.
4. `npm run test:browser:all` — green, in Docker.

## Checklist

## AI Usage

#### Type of assistance:

- [x] Code generation
- [x] Refactoring

#### AI System used:

- [x] Claude

#### Level of modification:

- [x] Created or updated unit tests to cover any new or modified code — no new tests; every change is a type annotation in an existing test or story.
- [x] If necessary, updated unit-test snapshots (`npm run test:unit:update`) and browser-test snapshots (`npm run test:browser:all:update`) — no snapshot changed.


packages/design-system/src/components/Autocomplete/Autocomplete.stories.tsx
```diff
@@ -40,20 +40,20 @@ const meta: Meta<typeof Autocomplete> = {
...
+  const onInputValueChange: AutocompleteProps['onInputValueChange'] = (...args) => {
     action('onInputValueChange')(args);
     setInput(args[0]);
   };
   let filteredItems = null;
   if (input.length > 0) {
     filteredItems = items
       .map((item) => {
-        if (item.label && item.items) {
+        if ('items' in item) {
```
jack-ryan-nava-pbc, line 56: Why was the check for the label key here in the first place?

packages/design-system/src/components/Table/Table.stories.tsx
```diff
@@ -26,7 +27,32 @@ export default meta;
 
 type Story = StoryObj<typeof Table>;
 
-const tableTemplateData = {
+/**
+ * The story tables render generically, looking each row's cells up by the
+ * `propName` of the corresponding heading, so a fixture's headings can only name
+ * keys its rows actually have.
+ */
+interface TableFixture<Row> {
```
jack-ryan-nava-pbc, line 35: @derek-cmsds I'm having trouble understanding exactly what is going on here. Can we do a huddle?

tamara-corbalt (approved):
Ran through the test steps, and everything behaved as expected. Thanks for carefully documenting the follow-up work to add `as const` to the WC attribute arrays and to fix the Table test. LGTM!

jack-ryan-nava-pbc (commented):
Overall this looks solid. Nice work! I want to understand what's going on with the Table more and double check a few things on a call.

packages/design-system/src/components/Autocomplete/Autocomplete.stories.tsx
jack-ryan-nava-pbc: Looks like it's not used anyway. Not sure. Moving on.

jack-ryan-nava-pbc (approved):
Spoke with Derek about outstanding questions, all good! Thanks Derek! Approving after unit tests passed and tsc compiled clean for the Core package.

PR #4238, opened 2026-09-13
fix(Pagination): re-render page labels when the language changes

### LINKED ISSUE #4094: [Bug:] i18n appears to be broken
### Describe the Bug

MCT recently adopted the CMSDS Pagination component. It contains overridable translations for Previous and Next links, but the mobile-viewport-width content `Page {X} of {XX}` is not controlled by a prop, and only uses the CMSDS internal translation helper, `t`.

MCT uses both the `setLanguage` helper from CMSDS and also updates the `html` element's `lang` attribute anytime a user changes language using the switch in the MAX global header. But the component is not updating its content to reflect the language set using the helper or derived from the `html` `lang` attribute.

See code [here](https://github.com/CMSgov/oc-mct-frontend/blob/176af9a3bd673707f2f96b278d0693979b181333/src/helpers/queryParamsHook.ts#L122) and observe the value of the `html` `lang` attribute when switching from English to Spanish and back.

### To reproduce

- Go to https://imp.medicare.gov/plan-compare
- Enter 21224
- Select MA plans
- Choose no extra help (low-income subsidy)
- Choose to add no drugs
- Choose to add no providers
- On the plans results page, open devtools, narrow the viewport down to mobile-device width, scroll down to where the pagination component is rendered and showing `Page 1 of 3` (or similar)
- Scroll to the top, open the mobile menu using the hamburger icon, and choose "Cambiar a Espanol"
- Scroll back to the bottom and observe that `Page 1 of 3` has not changed to Spanish

### Expected behavior

Either changing the `html` `lang` attribute (from `en` to `es` or vice-versa) or calling CMSDS's `setLanguage` helper with either `en` or `es`, or both, should update CMSDS components' internal translations.

### Severity/Impact

High: Affects PI objectives, breaks 508 compliance, breaks application

### Desktop Operating System

Mac OS

### Desktop Browser Name

Chrome

### Desktop Browser Version

149.0.7827.54

### Device

N/A

### Mobile Operating System

N/A

### Mobile Browser Name

N/A

### Mobile Browser Version

N/A

### Additional context

I have tried all variations of using `html` `lang` or the CMSDS language helpers (one, not the other, both, choosing to update the URL with a `lang` query parameter or not -- we do), and no combination of changes trigger the Pagination component to update its internally translated content.

### Codebase/Repository URL

https://github.com/CMSgov/oc-mct-frontend

### Design System Version

Medicare 14.4.5

N/A

### Team Name

MCT

soroush5:
# fix(Pagination): re-render page labels when the language changes

Fixes #4094

Hey! Found that the Pagination component keeps showing English labels
("Previous", "Next", page numbers' aria-labels) after switching the
site language — the labels are read once and never update, so screen
reader users hear the wrong language until a full page reload.

## What I changed

- `Pagination.tsx` — subscribe to the i18n language change (via the
  existing `useTranslation` hook) so labels re-render on `setLanguage`

## How I verified

Added a test that renders Pagination, calls `setLanguage`, and asserts
the labels update — red before the fix (stale English labels), green
after. Ran the neighboring component suites too, all green.


packages/design-system/src/components/utilities/useTranslation.ts
```diff
@@ -0,0 +1,19 @@
...
+import type { Language, TFunction } from '../i18n';
+import { getLanguage, subscribeToLanguage, translate } from '../i18n';
+
+/**
+ * Reactive translation helper. Unlike the module-level `t`, the returned
+ * function re-renders the calling component when `setLanguage` is called,
+ * so translated output (e.g. Pagination's "Page X of Y") follows runtime
+ * language switches.
+ */
+export function useTranslation(): { t: TFunction; language: Language } {
```
jack-ryan-nava-pbc, line 11: @soroush5 this is a great suggestion, and I like this approach, however I wonder if we can make something that is more portable so that it can be used in all of our translations. Are you able to research that? If not, our team can take this work as a starting point and later refactor our main translation function. 

cc @tamara-corbalt @derek-cmsds

packages/design-system/src/components/utilities/useTranslation.ts
soroush5, line 1: Thanks for taking a look, and for the kind words. You make a good point about portability, and it is something I weighed up before putting this up. I had already gone through the other components beforehand: most of them (Alert, UsaBanner, Dialog and the rest) use the module-level `t()` directly, so they all run into the same stale labels after `setLanguage`. The fully portable answer would be making `t()` itself reactive, but that turns it from a plain function into something hook-like, and that felt like your team's call rather than something to slip into a one-component fix. So what I put up here is deliberately just a starting point that breaks nothing: any component can switch over with a one-line import change. Happy for your team to run with it from here, and I would be glad to help carry it over to more components or reshape it to fit the core function, whatever works best on your end.

soroush5:
Closing as stale, no activity for a while. Happy to rebase and reopen if there is interest. Thanks for your time.

jack-ryan-nava-pbc:
Hey @soroush5 thanks for reminding me of this. I think this is a bigger effort that warrants greater attention from our team. We're gonna spin up some tickets internally and reference this work. We'll make sure to give you coauthor credit. Thanks so much for working on this!

cc @derek-cmsds @tamara-corbalt

soroush5:
Glad it helped. Happy to review or test your team's take whenever it is ready, just ping me. Feel free to close this one out or keep it as reference, whichever suits you.

PR #2442, opened 2023-03-27
[WNMGDS-2058] Setting up DS scripts to distribute ds-cms-gov assets via CDN

forrestbaer:
## Summary

- Removed 'incomplete' key for CMS.gov in themes.json
- Adding snapshots for CMS.gov to VRT

## How to test

1. [Manually Uploaded](https://design.cms.gov/cdn/ds-cms-gov/6.0.1) - CDN page
2. Since the themes.json value is no longer `incomplete`, future releases will work just like the others to the CDN

## Checklist

- [x] Prefixed the PR title with the [Jira ticket number](https://jira.cms.gov/projects/WNMGDS/) as `[WNMGDS-####] Title` or [NO-TICKET] if this is unticketed work.
- [x] Selected appropriate `Type` (only one) label for this PR, if it is a breaking change, label should only be `Type: Breaking`
- [x] Selected appropriate `Impacts`, multiple can be selected.
- [x] If necessary, updated unit-test snapshots (`yarn test:unit:update`) and browser-test snapshots (`yarn test:browser:update`)

PR #4239, opened 2026-09-14
[CMSDS-4575] Resolve the remaining noImplicitAny errors and enable the flag

derek-cmsds:
## Summary

Stage 4c of the staged `strict: true` rollout, and the last one for `noImplicitAny`. Resolves the remaining 95 errors — `packages/docs` 59, `.storybook` 19, `ds-healthcare-gov` 12, `ds-medicare-gov` 3, `tests/browser` 2 — and enables the flag in the root `tsconfig.json`.

- [CMSDS-4575](https://jira.cms.gov/browse/CMSDS-4575)
- Follows [CMSDS-4571](https://jira.cms.gov/browse/CMSDS-4571) — Resolve `noImplicitAny` errors in design-system tests and stories (#4232)

Four emitted declarations narrow from `any` to their real types: `useHint`, `useInlineError`, `getMonthNames` and `ds-healthcare-gov`'s `ActionMenu`. The flag causes all four, not the source fixes — three are in `packages/design-system`, which this stage never touches, and building the branch with the flag removed reproduces the previous declarations byte for byte. None is re-exported from a package entry point.

Two changes go past annotation and are worth a look. `DocSiteSeo`'s `location` prop was declared `Location`, the DOM interface, which Gatsby never passes and no caller ever satisfied; it now takes `LocationInterface` like the rest of the site. And `describeByTheme` skipped any theme marked `incomplete` in `themes.json`, a key removed from cmsgov in March 2023 (#2442), so the dead guard is gone.

Follow-ups, none of them fixed here:

- [CMSDS-4595](https://jira.cms.gov/browse/CMSDS-4595) — three unreachable docs modules, added to the existing ticket rather than deleted in this PR.
- [CMSDS-4614](https://jira.cms.gov/browse/CMSDS-4614) — gained the one `ds-*` wrapper outside core, whose story keeps the `Record<string, string>` workaround.
- [CMSDS-4622](https://jira.cms.gov/browse/CMSDS-4622) — the child packages publish their tests and snapshots to npm.
- [CMSDS-4623](https://jira.cms.gov/browse/CMSDS-4623) — two more dead `incomplete` filters, in files outside the tsconfig `include`.

## How to test

1. `npm run type-check` — 0 errors.
2. `npx tsc --noEmit --allowImportingTsExtensions --noImplicitAny` — no output.
3. `npm run test:unit` and `npm run test:unit:wc` — green. `npm run test:unit:preact` fails 8 suites / 15 tests / 6 snapshots, identical to this branch's base.
4. `npm run test:browser:smoke` and `npm run test:browser:interaction` in Docker — 804/804 and 309/309.

## Checklist

- [x] Confirmed that the base branch appropriately matches the selected release milestone.

## AI Usage

#### Type of assistance:

- [x] Code generation
- [x] Refactoring

#### AI System used:

- [x] Claude

#### Level of modification:

- [x] Created or updated unit tests to cover any new or modified code — no new tests; every change is a type annotation or a typed assertion in an existing file.
- [x] Verified that running both `npm run test:unit` and `npm run test:browser:all` were each successful — `test:unit` and `test:unit:wc` green; browser tests run in Docker, smoke 804/804 and interaction 309/309.
- [x] If necessary, updated unit-test snapshots and browser-test snapshots — none changed.


packages/ds-healthcare-gov/src/components/Header/ActionMenu.test.tsx
```diff
@@ -100,7 +100,7 @@ describe('ActionMenu', function () {
       });
       fireEvent.click(screen.getByRole('button'));
       expect(mock).toHaveBeenCalled();
-      expect(window['utag'].link.mock.calls[0][0]).toMatchSnapshot();
+      expect(mock.mock.calls[0][0]).toMatchSnapshot();
```
jack-ryan-nava-pbc: Can we call the mocked window.link object something other than "mock"? mock.mock is a bit confusing to me.

jack-ryan-nava-pbc (approved):
Everything passes fine locally, examples build and look ok. Thanks for doing this! One step closer.

packages/ds-healthcare-gov/src/components/Header/ActionMenu.test.tsx
derek-cmsds: What name would you prefer to use? We use `tealiumMock` elsewhere in the test suite, but that name relies on knowledge that is not derived from the code itself. `utagLinkMock` would be more accurate, but it's kinda clunky and not used anywhere else. `analyticsMock` might convey more semantic meaning, but it is also not used anywhere else right now.

jack-ryan-nava-pbc: I vote for `tealiumMock` since it's consistent and an easy search can explain what it's from. It's also what we're testing here.

derek-cmsds: `mock` renamed to `tealiumMock`.

